PR #173, opened 2014-07-30
Sphinx2

Alberth289346:
Except for the last 2 commits, these are all changes / improvements to the ags.tex file. I cannot build the output from it, since I cannot compile tex2rtf at my system, so I my have broken the .tex file. Please verify.

Commit 55602ec50a2ca5360f1c59ce1eb33633e1ca2719 adds the same manual, but in Sphinx format. This is a text format, with markup in plain ascii. The tool can produce a multitude of output format, including html and windows help files. It is an industral-strength tool, docs.python.org is generated with it, containing all documentation for every Python version. The .RST files are sources, other things are build files. I added a short README.txt as well explaining how to build HTML and Windows HTML Help files.

Commit cf902739fd37ab469049992c020799ab166aa7a adds a Python 3 conversion script, and a patch file for the last changes that the conversion script missed. You don't actually need these files, but they may be useful perhaps.

### DIFF
--- a/Manual/README.txt
+++ b/Manual/README.txt
@@ -21,3 +21,8 @@ Ultra short manual for Sphinx:
 
 - I kept all existing files here to not disturb the currently working solution.
   Obviously, most of them are obsolete if you use Sphinx instead.
+
+
+- The 'process.py' file is a custom Python 3 script to convert the ags.tex file
+  to the *.rst files. It works almost completely, the final changes are listed in
+  all_changes.patch .
--- a/Manual/all_changes.patch
+++ b/Manual/all_changes.patch
@@ -1,0 +1,220 @@
+--- other-features.rst	2014-07-30 21:08:46.077779102 +0200
++++ other-features.rst	2014-07-30 21:11:50.764427104 +0200
+@@ -490,7 +490,7 @@
+ *NOTE:* It is not possible to load the exe file back into the AGS Editor. This
+ means two things when only the EXE file is available: (1) other people can't
+ edit your game's data, and (2) you can't either. Always keep a backup of the
+-other files produced (*.CRM, GAME.AGF, etc) as they are what the Editor
++other files produced (\*.CRM, GAME.AGF, etc) as they are what the Editor
+ needs to be able to load your game for editing.
+ 
+ *TIP:* You can make a "Loading..." style splash screen to be displayed while your game starts
+@@ -576,13 +576,13 @@
+   all of the game settings. Without it, your game is lost.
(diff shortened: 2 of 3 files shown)


ivan-mogilko:
Having other priorities, I was not able to thoroughly check all the changes proposed here at a time, and this pull request was "freezed" undefinitely. I guess it makes little sense to discuss it now..., but the reasons I did not want to merge it as it is in the past were that it changed too much at a time (beautification of ags.tex is one thing, but adding Sphynx files is another), and I doubted I would be able to review the results of those changes on my own.

Now when I have little more time, I think I could try to apply *some* of these patches at least to improve the looks of existing manual.

Alberth289346:
Thanks for the update.

Yeah, I stopped hoping anything would be done with it after a few years. Apparently the step was too big. I don't know how to convert to a different documentation system in small steps though.

If you want to discuss something about the patches, I might be able to help you, if I still remember it :)
I won't do the conversion again, though.

Maybe sphinx is not the right answer nowadays. wiki markup languages have become more popular, so you might want to consider switching to markdown or asciidoc instead.

sonneveld:
Perhaps we could only merge in @Alberth289346 's conversion and build scripts for now. Our teamcity user manual build task could build the manual from tex as per usual but also run the rST conversion/build to test. 
The primary source would be .tex until we're happy with the conversion and then migrate to only using rST.

ivan-mogilko:
I do not really see a reason to put conversion script to the main repository, because it is supposed to be used only once, after that it becomes redundant.

sonneveld:
But that's my point, we could run it for every build until we're happy with the conversion.

ivan-mogilko:
Cannot the script be placed somewhere else instead and still be usable by server?

BTW, do I understand correct that this is basically LaTeX to rST conversion, that does not have anything specific to AGS?
Thing is that I searched for converter before, and saw this: http://pandoc.org/

sonneveld:
It can be placed on the build server but it makes it hard for other people to test it locally and I anticipate that we'd probably want to adjust the script as we continue to add content to primary .tex source.

I haven't looked at pandoc. I wonder what the output is like for converted documents. We'd still need to keep track of the settings used for conversion though and it would be easier to keep track in the repo.

ivan-mogilko:
OK.
But I guess it is better to first try out available converters and compare what they produce.

ivan-mogilko:
The manual is now remade into MD source and available in the separate repository: https://github.com/adventuregamestudio/ags-manual

PR #221, opened 2014-12-31
Game.agf: Wrote faulty timestamps to audio files

persn:
Timestamps are written to the file Game.agf with the UTC time standard, which adds an 'Z' at the end of the time stamp. However the read from Game.agf did not account for this and was thrown off by the 'Z' character, adding a couple of hours to the time stamp every time the editor was booted up. This should now be fixed.


monkey0506:
I remember running into this when committing AGF files to private repos.
Never got around to looking into it. Thanks for this!

PR #224, opened 2015-01-07
Feature editor color theme

persn:


tzachshabtay:
Added a few comments inside (preferring an interface over the abstract class and generalizing more behaviors in that interface) which I think should be addressed before merging.

ivan-mogilko:
@tzachshabtay , thank you for commenting :). Actually I was about to close this pull request, but probably forgot to.
We had a conversation with @persn on AGS forums regarding this feature, and I think we came to conclusion that this should be redesigned first.

PR #227, opened 2015-01-13
Editor: Close all other tabs

persn:
Addressed the following issue in the bug tracker http://www.adventuregamestudio.co.uk/forums/index.php?issue=621.0

While I was in the code area I added an option for closing all tabs as I've been missing this feature myself.

### DIFF
--- a/Editor/AGS.Editor/GUI/TabbedDocumentManager.cs
+++ b/Editor/AGS.Editor/GUI/TabbedDocumentManager.cs
@@ -23,6 +23,7 @@ namespace AGS.Editor
 
         private const int TAB_HEIGHT = 18;
         private const string MENU_ITEM_CLOSE = "Close";
+        private const string MENU_ITEM_CLOSE_ALL = "CloseAll";
         private const string MENU_ITEM_CLOSE_ALL_BUT_THIS = "CloseAllOthers";
         private const string MENU_ITEM_NAVIGATE = "Navigate";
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
This will be fine addition, but, although I do not know this area of the Editor code, I have a natural question: is it really necessary to call these two functions in a sequence:
<pre>
RemoveAllDocumentsExcept(document);
RemoveDocument(document);
</pre>
Is not there a function that just removes them all at once? :)

persn:
I couldn't find one no, you can see for yourself here https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/TabbedDocumentManager.cs

I could have made a new function that does what you describe, but then that function would be mostly duplicate code from the already existing "RemoveAllDocumentsExcept". I could maybe have modified said function to do what you describe, but in my experience that only contributes to make the function more clustered.

Calling the two already existing functions seemed the best way to make a clean code implementation, however if you have any ideas on how to solve this better I'd be happy to make the switch.

ivan-mogilko:
Hmm, how about let RemoveAllDocumentsExcept take 'null' value?
This would require two minor changes:
At line 254:
<pre>if (pane != null)
    _panes.Add(pane);</pre>
At line 260 :
<pre>if (pane != null && pane != _currentPane)</pre>

ivan-mogilko:
As I mentioned earlier, I am not acquainted with this part of the code, so some things here seem not obvious to me.
For instance, what is the need to create a new list and assign it to _panes after? Cannot you just remove the document from _panes directly (since the function already using a copy of array in "foreach")?
I just tried doing that, and it worked well (tested couple of cases); and the code became tad simplier too.

Something in the lines of:
<pre>
        public void RemoveAllDocuments(ContentDocument except)
        {
            ContentDocument[] copyOfPaneList = _panes.ToArray();
            foreach (ContentDocument doc in copyOfPaneList)
            {
                if (doc != except)
                {
                    bool cancelled = false;
                    doc.Control.PanelClosing(true, ref cancelled);
                    if (!cancelled)
                    {
                        doc.Control.DockingContainer.Hide();
                        doc.Visible = false;
                        _panes.Remove(doc);
                        _panesInOrderUsed.Remove(doc);
                    }
                }
            }

            if (except != null && except != _currentPane)
            {
                SetActiveDocument(except);
            }
            else
            {
                RefreshWindowsMenu();
            }
        }
</pre>

NOTE: I specifically checked that SetActiveDocument adds an item to _panes if it is not contained there.

I am not designing Editor code, so my view on it may be different from the people who do. Is what I suggest a valid idea?

persn:
That function looks very neat! Much better than the previous one in terms in readability. I'll make the switch, it should be perfectly valid I think.

ivan-mogilko:
I just realized I had a logical mistake there, we still have to call RefreshWindowsMenu() if no new active document is set, because the menu is to be updated always.

So:
<pre>
if (except != null)
{
    SetActiveDocument(except);
}
else
{
    RefreshWindowsMenu()
}
</pre>

persn:
Is there a preferred way to upload revised branches? Do I have to make a new pull request?

ivan-mogilko:
No, just force push modified branch to remote and GitHub will detect changes automatically.

persn:
That should do it

tzachshabtay:
Just a small nitpick regarding the question of the difference between the two methods: The original implementation is probably much faster than the more readable version (not that it really matters in this scenario). Each call to Remove actually iterates the entire collection, and you do it (twice) for each document, increasing the complexity. Since the most popular scenario is probably remove all documents but one, creating the list makes more sense performance-wise. Maybe if the user has something like 100 open documents there will be a noticeable difference but I doubt it.

PR #228, opened 2015-01-15
Editor: Wrong default item gets focus in General Settings

persn:
Addressed the following issue in the bug tracker
http://www.adventuregamestudio.co.uk/forums/index.php?issue=606.0

### DIFF
--- a/Editor/AGS.Types/AudioClip.cs
+++ b/Editor/AGS.Types/AudioClip.cs
@@ -5,6 +5,7 @@ using System.Xml;
 
 namespace AGS.Types
 {
+    [DefaultProperty("BundlingType")]
     public class AudioClip : IToXml, IComparable<AudioClip>
     {
         private const string COMPILED_AUDIO_FILENAME_PREFIX = "au";
--- a/Editor/AGS.Types/AudioClipFolder.cs
+++ b/Editor/AGS.Types/AudioClipFolder.cs
@@ -16,6 +16,7 @@ namespace AGS.Types
             base(new AudioClipFolder(node)) { }
     }
 
+    [DefaultProperty("DefaultBundlingType")]
     public class AudioClipFolder : BaseFolderCollection<AudioClip, AudioClipFolder>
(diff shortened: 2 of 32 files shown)


ivan-mogilko:
That issue you mention was made as a quick note (because I tested creating new games a lot, this behavior became annoying). It should be expanded to a more general rule, because there are more than one PropertyGrid in the Editor, and same problem may occur on others too. And there should be a consistency here.

I am not quite sure how default Property grid works in the regard to initial item selection. When an object is opened for editing from the Project Settings usually the first item in alphabetic order is selected, but with some objects the "Name" property gets selection (one I remember is AudioClipType), although it is not the first in alphabetic order. At some occasion I tried to find out why it happens so, but failed (do not know much about these things).

For the reference, the MSVS property pane chooses different items depending on edited object, e.g. when editing UI controls with label (Labels, Panels, Buttons) the "Text" property gets selection by default, CheckBoxes get "Checked" selected, and so forth.

Is it possible to implement similar system in AGS Editor? May such thing be done with attributes (to define a default property)?

ivan-mogilko:
Oops. After little search, I think I found it :).

The attribute is called "DefaultPropertyAttribute", when applied to the *class* it makes specified property selected when the property pane is opened.

We need to test how well this would work for AGS objects.

persn:
Your idea is interesting, I'll need to do some experiments to see what is possible.

persn:
I added "DefaultProperty" to a bunch of classes, it works like a charm as far as I can tell.

ivan-mogilko:
Wait, I would suggest to discuss this on forums first to gather other devs & user suggestions on what they see a property they would like to see/change first.

persn:
That's not a bad idea. My last update was mostly a proof of concept really, some feedback to what properties should be default would be helpful.

persn:
Update: http://www.adventuregamestudio.co.uk/forums/index.php?topic=51594.0

PR #229, opened 2015-01-19
Editor: Papagayo .dat File Support for Lip-syncing

gurok:
This issue was originally submitted by AGD2. The AGS bug tracker entry can be found here:
http://www.adventuregamestudio.co.uk/forums/index.php?issue=628.0

As stated in the bug tracker entry, Pamela has long been superseded as a method for generating lip sync files. It's buggy and creating the lip sync data is a very manual process. The authors of the software that Pamela was built for have created their own tool called Papagayo. This patch adds support for its .dat file format (which is extremely similar and even a bit simpler than the Pamela format). As AGD2 is the guy who originally suggested Pamela, I think we can trust his research that Papagayo is a worthwhile replacement tool.

AGD2 gave me a test project that can be used to verify support:
http://www.himalayastudios.com/scratch/PapagayoTest.rar

Note 1: The test project comes with the disclaimer that nobody uses the dialog portrait images or audio in any other projects/games, etc.

Note 2: The Pamela lip sync looks slightly different to the Papagayo lip sync in the test project because the data points are actually different. The Pamela data was created by hand and the Papagayo data was created automatically.

### DIFF
--- a/Editor/AGS.Editor/Components/SpeechComponent.cs
+++ b/Editor/AGS.Editor/Components/SpeechComponent.cs
@@ -67,7 +67,7 @@ namespace AGS.Editor.Components
             get { return ComponentIDs.Speech; }
         }
 
-        private int FindFrameNumberForPhenome(string phenomeCode)
+        private int FindFrameNumberForPhoneme(string phonemeCode)
         {
             int frameID = -1;
             for (int i = 0; i < (_agsEditor.CurrentGame.LipSync.CharactersPerFrame.Length) && (frameID < 0); i++)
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
So, this compiles the lipsync data native to AGS, and no extra changes to Engine are necessary?

gurok:
Yes. This is just a secondary format. There's a new function in SpeechComponent to read the Papagayo format with minor differences from the Pamela one. It is processed and compiled into syncdata.dat just like the Pamela data.

PR #231, opened 2015-01-31
Scripting: Allow variable declarations in for loop initialisers

gurok:
This allows syntax like `for(int i = 0; i < 10; i++)` where `i` is a variable local to the for loop. I started this a long time ago when someone complained about for loops not really being useful in their current state. I thought it would be good for completeness.

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -377,7 +377,20 @@ int deal_with_end_of_ifelse (char*nested_type,long*nested_info,long*nested_start
          return 1;
      }
      else
+     {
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
This works, but I have one question; there is obviously a duplicated code, copied from plain variable initialization case. Is there a way to avoid this duplication by refactoring the code a bit?

gurok:
Yes there is, but I am just going to go with your advice to leave a note there for now.

ivan-mogilko:
Okay, since there is a free time issue, I will merge this; the code duplication has to be dealt with later.

PR #232, opened 2015-01-31
Scripting: Remove 10,000 symbol limit

gurok:
This replaces all of the fixed-size arrays in the symbol table with vectors. Between 3.3.3 and 3.4.0, some 75 (approximately) new symbols were added to the default environment. This includes new operators, new statements and many new API methods and brings the total to about 2,100 symbols without any user code. It surfaced as a problem recently with one particular game that compiled under 3.3.3 but broke in 3.4.0. The game defined 9980 symbols in a script and the new changes pushed it over the 10,000 symbol limit.

### DIFF
--- a/Compiler/script/cc_symboltable.cpp
+++ b/Compiler/script/cc_symboltable.cpp
@@ -47,6 +47,18 @@ int symbolTable::get_propset(int symb) {
 void symbolTable::reset() {
     int rr;
     for (rr=0;rr<numsymbols;rr++) free(sname[rr]);
+    sname.resize(0);
+    stype.resize(0);
+    flags.resize(0);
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Since you are replacing const-sized arrays with vectors, may you also replace
<pre>
 std::vector<unsigned long *> funcparamtypes;
 std::vector<short *> funcParamDefaultValues;
</pre>
with
<pre>
 std::vector< std::vector<unsigned long> > funcparamtypes;
 std::vector< std::vector<short> > funcParamDefaultValues;
</pre>
?

gurok:
Sorry about the delay. I think the data types are correct now.

PR #238, opened 2015-03-25
Fixes file name to prevent issues when cloning

monkey0506:
...to networked locations on Windows.

As far as I can tell, this file is not referenced
anywhere in the Editor or Engine code. It may be
automatically generated, in which case it is
probably best to ignore it instead of renaming it.


ivan-mogilko:
The filename should be config_types.h: https://github.com/adventuregamestudio/ags/blob/master/Common/libinclude/OGG/MAKEFILE.IN#L48

Fixed with 3c0f207.

PR #243, opened 2015-05-13
Editor: Fixed help window multi-monitor size

cellproductions:
The HTML Help API's HtmlHelp function forces the help window to stretch across multiple monitors if it opens at a negative location (-x, -y), which can be quite inconvenient.
I added a method that gets the size of the monitor that the help window was created in and resizes the help window to that size, as well as move the window to the top left corner of the monitor.

### DIFF
--- a/Editor/AGS.Editor/Components/HelpCommandsComponent.cs
+++ b/Editor/AGS.Editor/Components/HelpCommandsComponent.cs
@@ -5,11 +5,40 @@ using System.IO;
 using System.Text;
 using System.Windows.Forms;
 using System.Xml;
+using System.Runtime.InteropServices;
+using System.Diagnostics;
 
 namespace AGS.Editor.Components
 {
     class HelpCommandsComponent : BaseComponent
     {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@terrorcell , there was a plan to make Editor Mono-compatible over time. This change adds dependency on Win32 dlls. Can you discuss this with other Editor developers first? Maybe there is a better alternative. I would like to have their opinion first.

terrorcell:
@ivan-mogilko that's fair enough. It's more me being nit-picky than anything serious, so I don't think it really warrants much discussion. Regardless, I'll add it as an issue instead in case the same behaviours exist in Mono (as it turns out, I don't think Help has been finished in Mono, so perhaps that does warrant some discussion).

tzachshabtay:
As cellproduction said, Mono doesn't currently support Help (and I don't know if they are planning to), so it doesn't matter much. If we ever do support Mono we'll probably need to implement an alternative solution for showing help. So I'll merge this now.

PR #253, opened 2015-07-13
Editor: Ignore lip sync files with invalid data

gurok:
Any lip sync data files that contain a non-number where a number would be expected will now be silently ignored. It is assumed that these are files not related to Pamela or Papagayo.


ivan-mogilko:
It appears we will have a problem with this change.
Editor does not rebuild the lipsync data unless any lipsync file changes. But some of the errors do not refer to only file content, but also to lipsync configuration inside the game project ("No frame found to match phoneme code"). If user fixes the lipsync properties after trying to build once, the Editor will not detect this as a change and won't rebuild lipsync data.

gurok:
Okay, this should be safer. Generates errors for invalid offsets and unmatched phonemes are back to throwing errors, as requested.

PR #254, opened 2015-07-13
Find / Replace Dialogue Issues

persn:
Reference: http://www.adventuregamestudio.co.uk/forums/index.php?issue=672.0

1. "Case sensitive" check box does not remember its value.
    Rewrote the handling for remembering checked value of "Case Sensitive". Mimicking the behaviour of the same feature in Visual Studio it will now remember the value when checked, regardless if a search / replace was made or not.

2. The order of cycling (tabbing) though controls is wrong: "Search for" field is followed by "<< Find" button; must be followed by "Replace with" field instead ("<< Find" button should probably go after all setup fields).
    Implemented exactly as described. Setup fields first., then "<<Find/Replace" button, and finally "Find", "Replace" and "Cancel" buttons.

### DIFF
--- a/Editor/AGS.Editor/GUI/FindReplaceDialog.Designer.cs
+++ b/Editor/AGS.Editor/GUI/FindReplaceDialog.Designer.cs
@@ -88,7 +88,7 @@ namespace AGS.Editor
             this.cmdToggleReplace.Location = new System.Drawing.Point(298, 15);
             this.cmdToggleReplace.Name = "cmdToggleReplace";
             this.cmdToggleReplace.Size = new System.Drawing.Size(75, 21);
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
@persn , could you merge this into master branch instead? We are planning a hotfixed release, and this change would be very useful.

persn:
@ivan-mogilko I'll do that, but I'm going to wait a little. After I made the request I saw that @gurok made a comment in the bug tracker, I would like to include that in the merge if I can figure out how to fix it.

ivan-mogilko:
Alright.

persn:
New pull request in https://github.com/adventuregamestudio/ags/pull/255

PR #257, opened 2015-07-14
Absolute Paths for Resources

persn:
Reference: http://www.adventuregamestudio.co.uk/forums/index.php?issue=520.0

Importing sprites, audio and fonts would always store absolute file path in the property "SourceFileName". It now stores relative path if the file is found inside the Game Project, or absolute path if it is from outside of the game project. This is useful if the the game project is moved to another place on a local computer, or if the game project is shared between multiple collaberators on each their private computer.

Examples:
  Relative Path: Resources/Sprites/MySprite.bmp
  Absloute Path: D:/MyGameProject/Resources/Sprites/MySprite.bmp

### DIFF
--- a/Editor/AGS.Editor/ApplicationController.cs
+++ b/Editor/AGS.Editor/ApplicationController.cs
@@ -28,6 +28,7 @@ namespace AGS.Editor
             _pluginEditorController = new AGSEditorController(_componentController, _agsEditor, _guiController);
 
             _events.GameLoad += new EditorEvents.GameLoadHandler(_events_GameLoad);
+            _events.GamePostLoad += new EditorEvents.GamePostLoadHandler(_events_GamePostLoad);
             _events.GameSettingsChanged += new EditorEvents.ParameterlessDelegate(_events_GameSettingsChanged);
             _events.ImportedOldGame += new EditorEvents.ParameterlessDelegate(_events_ImportedOldGame);
             _events.RefreshAllComponentsFromGame += new EditorEvents.ParameterlessDelegate(_events_RefreshAllComponentsFromGame);
@@ -80,6 +81,26 @@ namespace AGS.Editor
             foreach (string componentID in missingComponents)
             {
                 _guiController.ShowMessage("This game contains data from a plugin or component '" + componentID + "' which you do not have installed. If you save the game, this data will be lost.", AGS.Types.MessageBoxIconType.Warning);
+            }
+        }
+
+        private void _events_GamePostLoad()
+        {
+            Game game = Factory.AGSEditor.CurrentGame;
+
+            // Convert absolute paths to relative paths. This is an automatic fixup from when the
+            // editor stored absolute paths only
+            foreach (Sprite sprite in game.RootSpriteFolder.GetAllSpritesFromAllSubFolders())
+            {
+                sprite.SourceFile = Utilities.GetRelativeToProjectPath(sprite.SourceFile);
+            }
+            foreach (Types.Font font in game.Fonts)
+            {
+                font.SourceFilename = Utilities.GetRelativeToProjectPath(font.SourceFilename);
+            }
+            foreach (AudioClip audio in game.RootAudioClipFolder.GetAllAudioClipsFromAllSubFolders())
+            {
+                audio.SourceFileName = Utilities.GetRelativeToProjectPath(audio.SourceFileName);
             }
         }
 
--- a/Editor/AGS.Editor/Components/AudioComponent.cs
+++ b/Editor/AGS.Editor/Components/AudioComponent.cs
@@ -304,7 +304,7 @@ namespace AGS.Editor.Components
             {
                 string newScriptName = EnsureScriptNameIsUnique(Path.GetFileNameWithoutExtension(sourceFileName));
                 AudioClip newClip = new AudioClip(newScriptName, _agsEditor.CurrentGame.GetNextAudioIndex());
-                newClip.SourceFileName = sourceFileName;
+                newClip.SourceFileName = Utilities.GetRelativeToProjectPath(sourceFileName);
                 newClip.FileType = _fileTypeMappings[fileExtension];
                 newClip.FileLastModifiedDate = File.GetLastWriteTimeUtc(sourceFileName);
                 Utilities.CopyFileAndSetDestinationWritable(sourceFileName, newClip.CacheFileName);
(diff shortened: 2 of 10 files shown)


ivan-mogilko:
This might be nitpicking, but I want to point out that this implementation imposes an additional restriction on these functions - like CreateAudioClipForFile etc - that their filename argument must be an absolute path, otherwise this replacement will not work correctly.

Is this something to worry about, and isn't there any method that assuredly converts any random path into relative to given directory?

persn:
Hmm, I think I see what you're worried about. The Uri class should have some magic functions that solves this. I'll take a look later.

ivan-mogilko:
@persn , do you have plans to finish this pull request?
I think I've found the proper Uri method, so if you do not have time, I could do this myself.

persn:
What a coincidence that you should ask of this now, I just resumed work on this a couple of hours ago.

Using the following code block instead of the previous solution I can create a relative path from the project directory to a referenced file. If I understood you correctly that is what you wanted?

https://msdn.microsoft.com/en-us/library/system.uri.makerelativeuri(v=vs.110).aspx

```C#
Uri currentProjectUri = new Uri(Factory.AGSEditor.CurrentGame.DirectoryPath + Path.DirectorySeparatorChar);
newClip.SourceFileName = Uri.UnescapeDataString(currentProjectUri.MakeRelativeUri(new Uri(sourceFileName)).OriginalString);
```

If we're on the same page I'll update the pull request.

ivan-mogilko:
Yes, I think it was that one.

persn:
Pull Request is now updated to use Uri.MakeRelativePath instead.

ivan-mogilko:
@persn , the idea of the pull request was changed with the latest commit?
Originally, it was to keep absolute paths if the file is outside of the project folder, which makes sense, but now it stores relative paths always.

I believe this is not a good thing.

Real situation:
* I have a special folder to keep gfx and audio for my AGS projects, and a project that imports them. I move the project to the folder with different nesting level, or move project to another disk, and all references break.

* I may keep common resources (icons, sample sounds for quick tests) in the same absolute path on two computers, like "My Documents\AGS\CommonRes", but may place AGS project on different  nesting levels. If I copy project from one PC to another, references will break.

sonneveld:
In regards to the failing builds, rebasing with develop-3.4.0 should fix that issue.

ivan-mogilko:
Maybe the dev-3.3.5 needs to use OpenGL and DirectX SDK enviroment variables too?

sonneveld:
Looks like it. What are the current active branches? Are they develop-3.3.5, develop-3.4.0 and master ?

ivan-mogilko:
Yes.

sonneveld:
@ivan-mogilko  Cool. I take it master doesn't get merged into develop-3.3.5 .   I'll make two pull requests for master and 3.3.5 then to graft the same changes in 3.4.0

What gets considered for adding into 3.3.5 branch?  Perhaps this pull request should be for develop-3.4.0 instead.

ivan-mogilko:
@sonneveld 
> What gets considered for adding into 3.3.5 branch

Basically, fixing most annoying problems in AGS that are not hard to fix, without changing big parts of code.

persn:
@ivan-mogilko 

I'm not sure if I agree with your verdict to that storing relative paths as a constant rule is not a good thing. This is after all how every other any kind of Editor does it, going away from that commonality would add unexpected behavior to AGS that the end-user would need to get familiar with.

Also, storing relative paths outside of the project folder is useful for source control. If I have two folders one containing an AGS project and a separate one containing music and graphics. With relative paths I know the game project will successfully reference the assets on any fork.

I'm not really that sure about the relevance of your "real-situation" either. If I did such a thing the references would indeed break, but references would also break if I started moving projects in Visual Studio, Eclipse, PHPStorm and so on. Moving important individual folders around is not recommended practice, and it should be the responsibility of the game developer to make sure that such move-changes are kept to a minimal, and to repair them if needed.

ivan-mogilko:
I need to ask people on forums. I do not want them to start complaining that Editor started to act differently with no way to change that.

persn:
Agreed, let's collect some feedback on how this feature should work before appying a solution.

Regarding the automatic fixup, I'll have to look into it.

persn:
I'm not sure, where in the code would it be prudent to implement the automatic fixup

ivan-mogilko:
I would need to examine the code, but I guess some kind of "After Load" step added to be performed after project deserialization has finished.

persn:
Work In Progress. This is WIP since I haven't been able to test the changes made, I keep getting run-time errors with AGS on my laptop for reasons I don't have time to figure out, so I'll have to test the changes when I get some free time on my desktop.

In the meantime, I've divided the work into 3 commits:

The 1st commit fixes indentations and code formatting on all the files relevant for this commit, this was solved with VS's format function.

The 2nd commit implements the convert to relative paths, it converts to relative path only if the file is found inside the project folder like we agreed on the AGS forums. To avoid code duplication I put the actual converting code inside "AGS.Editor/Utils/Utilities.cs". Additionally as a difference from the last time I uploaded the branch, I neglected to check all usages of "source file" properties. I should now have covered every case.

The 3rd commit is the implementation of the automatic fix-up that was requested by @ivan-mogilko. This is the biggest change from the previous upload and should be the focus on code-review, I'm not entirely certain if this solution is the best one so I'm open to feedback. My primary concerns is if the way it was implemented into the code base is adequate, also I wonder if there is a better way to solve this entirely. For instance if the code is loaded into another computer than where the assets were originally loaded it won't work. On the other hand I'm thinking there are only so many scenarios we CAN cover.

ivan-mogilko:
I suggest to split indentation cleanup to another pull request, because it is completely unrelated to this task.

persn:
I'll post a new merge request shortly

persn:
Indent cleanups is now in another branch. I still need to test if my changes actually work once I have some spare time with my desktop computer.

persn:
Changes have now been tested. I tested that importing assets is done correctly, as well as testing the automatic fix-ups on project load. All values are stored correctly whether the imported asset was located inside the project folder, outside it, or on another disk completely.

Only thing I noticed is that Uri stores the directory separator in the opposite way than Path. That is "Assets/hero_76.bmp" instead of "Assets\hero_76.bmp". It still works combining different directory separators in one path, it just looks weird. I'm not sure if there exists a fix trivial enough to worth bother putting in.

PR #263, opened 2015-08-18
Refactor and introduce tests for cc_treemap and cc_symboltable.

sonneveld:
Refactor and introduce tests for cc_treemap and cc_symboltable.  No behavioral changes were made to other classes.

* cc_treemap now uses std::map
* cc_symboltable uses a single vector for all entries.
* WIP test infrastructure


sonneveld:
Engine is currently broken. I didn't notice that treemap was also used by the engine.  Will fix!

ivan-mogilko:
@sonneveld , I actually have a patch that removes ccTreeMap from Engine.
I haven't commited yet, because it was a part of one experimental feature.
I may actually do this now (it is a separate change), and then we may remove that tree class completely.

ivan-mogilko:
This commit 4220cb9f0d5021c5a155215abc2c2d487caf895c removed ccTreeMap from the Engine.

sonneveld:
Looks like it builds correctly *if* it's merged into develop- :)  I'll rebase when I have time.

PR #267, opened 2015-08-31
Windows Installer build scripts.

sonneveld:
These are not perfect but hopefully enough to start getting an installer building through teamcity. Comments welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Licensed under the Artistic License 2.0, see License.txt.
 To get started building the AGS engine, see the platform specific instructions or forum threads:
 
 -    [Linux](debian/README.md) ([Forum thread](http://www.adventuregamestudio.co.uk/forums/index.php?topic=46152.0))
--    Windows ([Forum thread](http://www.adventuregamestudio.co.uk/forums/index.php?topic=46847.0))
+-    [Windows](Windows/README.md) ([Forum thread](http://www.adventuregamestudio.co.uk/forums/index.php?topic=46847.0)) ([wiki](http://www.adventuregamestudio.co.uk/wiki/Compiling_AGS))
 -    [Android](Android/README.md) ([Forum thread](http://www.adventuregamestudio.co.uk/forums/index.php?topic=44768.0))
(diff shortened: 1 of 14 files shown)


sonneveld:
Oh TeamCity won't include the manual until I add those artifacts, but that requires the user manual pull request to go in.  I think both pull requests can go in at the same time.. the installers just won't have the help file in the first couple of builds.

sonneveld:
Yay, okay, now this one works in Teamcity and the user manual one fails (specifically the installer job).  Anyway, they should both be okay to merge in.

ivan-mogilko:
I tried to run this script on my PC and I found following:
* I got syntax error in build.py at "except Exception**,** e". I learnt that modern Python requires "Exception **as** e" syntax. Which version of python should be used to run the script?
* This script requires pywin32 module.

May the requirements be mentioned somewhere?

ivan-mogilko:
Since this is Windows installer anyway, I learnt there is a way to make it choose python version explicitly by using python launcher: https://docs.python.org/3/using/windows.html#python-launcher-for-windows.

Will this method be suitable for our case?

sonneveld:
Actually you raise a good point. I'm used to just writing all my python code for python2.  The version I use (and is installed on build server) is activepython 2.7 from http://www.activestate.com/activepython/downloads . It includes pywin32 by default.  The py launcher doesn't seem to come with python2 , at least not the version I have installed.

I'm happy to fix so it works in python3 instead.

sonneveld:
As of commit 9819a8d, there is a README with the minimum windows build requirements.

sonneveld:
Ah hangon, I've just realised other ports have their own README. I'll move my additions into a Windows README

sonneveld:
Issue: vcredist_x86.exe isn't installed into tmp.

sonneveld:
Ensure correct version of vcredist is installed. (changed teamcity config to download right one) In 436f9b6

sonneveld:
Tested current build. After uninstalling the VS2008 redist, it will detect it isn't installed and give you an option to install it.  Reinstalling again, it will not show the option.

PR #270, opened 2015-09-05
Fix a couple of compiler issues and introduce testing via googletest

sonneveld:
Not quite ready for merging but thought you might be interested in looking at the code.

I'm not happy with how I added googletest (it requires GOOGLETEST_PREFIX env var to be set) since visual studio has some issues with env vars in relative paths.  I might look into just checking in the small amount of googletest library code into ags repo (like how llvm does here https://github.com/llvm-mirror/llvm/tree/master/utils/unittest/googletest )

Still need to write a couple of tests for the code I modified but seems to work.


ivan-mogilko:
Can you split bug fixing and tests into two pull requests?
Also, I suggest to find a convention on commit prefix, we never used "Compiler.Lib" before, we used "Scripting". What about using just "Compiler", for its just shorter?

sonneveld:
No worries with the commit titles.

With the bug fixes, it makes sense to keep the tests with the bug fix (as they usually go hand in hand).  However I do agree that the initial googletest framework should be one pull request and additional bug fixes and another pull request.

sonneveld:
re commit titles.  I don't want to lose current comments so I'll rebase at the end to fix those up.

sonneveld:
I'll wait for the new google test pullrequest to go in first and then I'll rebase this branch onto that.

sonneveld:
I'm closing this pull request as I have split into separate patches.

PR #271, opened 2015-09-07
Add google-test and initial dummy test for compiler

sonneveld:


tobihan:
Hi,

is it really necessary to add more external code to the repository?
That's what git submodules are for.

Best,
Tobias

sonneveld:
@tobihan This is a good point but I was hoping to filter the files. No matter, included as a submodule to make you happy.

PR #273, opened 2015-09-08
Compiler int fixes

sonneveld:
* parse integers with overflow checks
* allow default parameters up to INT_MIN/INT_MAX
* support negative enums


sonneveld:
Marked as wip so I can fix some error messages after the symbol-fixes merge goes in.

ivan-mogilko:
One thing I noticed is that if I type a *negative* constant bigger than an 'int' can handle, then error message referes to positive constant instead (or maybe absolute value of one).
Do you think it matters?

sonneveld:
I think it does.. otherwise someone will ask why the overflow.  Will fix that.

PR #281, opened 2015-09-28
PSP: Fix build scripts for building in linux.

sonneveld:
Changes to assist in building PSP builds in teamcity.

(against master so it can be merged into develop-* branches)


sonneveld:
(note: Installer/manual doesn't build in 'master')

PR #283, opened 2015-10-04
Android: Update and fix project files

sonneveld:
* update to latest sdk template files
* fix building with ant


sonneveld:
Android project fixes are complete and ready for review.

sonneveld:
One possibly controversial change is that I removed all the nativelibs since they were now out of date. The build server will now rebuild those as well. I could potentially make them available to download but archive would be quite large!

PR #284, opened 2015-10-06
Read and remove Zone Identifier on the help file.

morganwillcock:
Add utility class to read and delete NTFS alternate streams. If the help
file exists check it's zone identifier, remove if it's potentially
restrictive.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -328,6 +328,7 @@
       <DesignTimeSharedInput>True</DesignTimeSharedInput>
     </Compile>
     <Compile Include="Resources\ResourceManager.cs" />
+    <Compile Include="Utils\AlternateStreams.cs" />
     <Compile Include="Utils\SourceControlProvider.cs" />
     <Compile Include="Enums\SpriteImportMethod.cs" />
     <Compile Include="GUI\TabbedDocumentManager.cs">
--- a/Editor/AGS.Editor/Components/HelpCommandsComponent.cs
+++ b/Editor/AGS.Editor/Components/HelpCommandsComponent.cs
@@ -67,27 +67,36 @@ namespace AGS.Editor.Components
 
         public override void CommandClick(string controlID)
         {
-            if (controlID == ABOUT_AGS_COMMAND)
+            if (controlID == CRASH_EDITOR_COMMAND)
+            {
+                throw new AGSEditorException("Crash test");
+            }
+            else if (controlID == ABOUT_AGS_COMMAND)
             {
                 AboutDialog dialog = new AboutDialog();
                 dialog.ShowDialog();
                 dialog.Dispose();
+                return;
             }
             else if (controlID == VISIT_AGS_WEBSITE)
             {
                 LaunchBrowserAtAGSWebsite();
+                return;
             }
             else if (controlID == VISIT_AGS_FORUMS)
             {
                 LaunchBrowserAtAGSForums();
+                return;
             }
             else if (controlID == CHECK_FOR_UPDATES)
             {
                 CheckForUpdates();
+                return;
             }
             else if (!File.Exists(_helpFileName))
             {
                 _guiController.ShowMessage("The help file '" + _helpFileName + "' is missing. You may need to reinstall AGS.", MessageBoxIcon.Warning);
+                return;
             }
             else if (Utils.AlternateStreams.GetZoneIdentifier(_helpFileName) > Utils.AlternateStreams.URLZONE_LOCAL_MACHINE)
             {
@@ -100,10 +109,15 @@ namespace AGS.Editor.Components
                     else
                     {
                         _guiController.ShowMessage("The help file couldn't be unblocked. Please try running the AGS editor using 'Run as administrator' and try to unblock the file again.", MessageBoxIcon.Warning);
+                        return;
                     }
-                }  
+                }
+                else
+                {
+                    return;
+                }
             }
-            else if (controlID == LAUNCH_HELP_COMMAND)
+            if (controlID == LAUNCH_HELP_COMMAND)
             {
                 string keyword = string.Empty;
                 if (_guiController.ActivePane != null)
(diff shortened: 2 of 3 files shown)


ivan-mogilko:
I apologize for this delay.

There are two problems with this code, first made it unable to open the stream at all, at least for me.
In GetSafeFileHandle() you request GENERIC_ALL permissions, that did not work. GENERIC_READ worked though. Since reading the stream is all you need, I think GENERIC_READ should be used instead.

Another minor thing is that help file is not opened automatically after removing the blocking stream, thus requiring user to select menu command for the second time. This is simply because of how the conditional blocks are set up in the code. I think it could be modified a little to let it continue into launching help viewer right after file got unblocked.


Regarding the unblocking technique; upon occasional search I found this code snippet, which dealt with the problem differently:
http://blogs.msdn.com/b/oldnewthing/archive/2013/11/04/10463035.aspx

Instead of manually opening and parsing the file stream, it takes advantage of IZoneIdentifier COM object (https://msdn.microsoft.com/en-us/library/ms537130(v=vs.85).aspx). Please, give this example a look; I suppose it may be a better way, relying on operating system doing all the low-level job; if it is possible to to make it work from .NET program.

*NOTE:* the sample code above actually has couple of syntax mistakes, author being too lazy to test it. Also it uses ATL, which is not really necessary.

ivan-mogilko:
Oh, one more thing. I wonder if the zone check should be that restrictive, skipping only "Local Computer". Maybe let "Intranet" and "Trusted" zones too?

morganwillcock:
I'll take a look at doing it with the COM object and getting the help file to open.

On the subject of determining what is restrictive, "Local Machine" is the default for all content on the machine which makes it very unlikely that any applications would get restricted by this (it's likely the computer would be unusable when someone tried to login if content in this zone is ever deemed to be questionable). For all greater zones it is application dependant on what is deemed to be restrictive, so I checked for zone Ids which are _potentially_ restrictive. What the Help Viewer determines to be safe looks to be configured in the registry, but I didn't attempt to read it because:

1. Help Viewer settings are present in 32 bit and 64 bit registry keys. Without building against a newer .NET Framework version, to access the 64 bit view would require additional use of p/invoke

2. Any other software on the machine could deem that any zone ID greater than “Local Machine” is unsafe and interfere before the Help Viewer is launched. E.g. An overly paranoid anti-virus product could block access based on Zone ID with file system filter driver (on-access scanner).

ivan-mogilko:
More interesting details.

The caching problem you mentioned before: I found that culprit could actually be COM object (I guess .NET classes may be a wrappers over IInternetSecurityManager and its siblings).

For me it worked slightly different though (both .NET and COM): it detected when I manually unblock file, but it did not detect when I replaced the file by copying blocked/unblocked version over it while program was still running.
Are you sure that you clicked OK or Apply after unblocking the file in your tests? That would be interesting to re-check to be sure.

I think I found how the cache can be controlled. IInternetSecurityManager calls this "mapping". I found that as soon as you check the zone first time, it maps the URL to the zone in its internal dictionary.
Thing is that you may actually remove that mapping with IInternetSecurityManager::SetZoneMapping(),

I wonder if .NET class has a similar workaround for this.



BTW, I found someone else made a C# library that works with Alternate Data Streams similarly to your code: https://github.com/hubkey/Trinet.Core.IO.Ntfs

morganwillcock:
I'm pretty sure that when I was testing I didn't copy files over the top of each to change the Zone Identifier, I saw the problem when deleting the stream with DeleteFile and by manually unblocking. I was testing on Windows 10, so potentially the behaviour may have altered on a different version of Windows.

I think the reason why other solutions resort to working with the filesystem rather than with the Internet Security components is because behaviour can change based on the IE version that is installed. I've found these flags:

https://msdn.microsoft.com/en-us/library/dd759042(v=vs.85).aspx

...including the promisingly named MUTZ_DONT_USE_CACHE, but it looks like you can't rely on these working if the IE version it too low. Perhaps the differences in behaviour we are seeing is because of differences in IE components rather than differences in Windows?

tzachshabtay:
I'm not too familiar with this issue, but can't we just embed the help file as a resource in the solution? I think windows won't have any issues with opening the file if it's embedded (but maybe I'm way off due to misunderstanding the issue).
Any file can be embedded by "adding existing file" from the solution, then in the file properties selecting "embedded resource". Loading the resource to a stream is then performed by Assembly.GetManifestResourceStream.

sonneveld:
Does it make it tricky since the manual is usually built after the editor.

tzachshabtay:
Mmm, yes, it does sound tricky in that case.
Is there a reason for the manual to be built after the editor?

sonneveld:
The user manual is an artifact of all the user manual build process. It's only done during release. I don't think everything should be built through visual studio.

tzachshabtay:
The fact that the user manual build is only done during release does not dictate that it should build after the editor (or from within visual studio).
When not releasing, the embedded help file will contain the last built help file. When releasing, user manual can be built first (overriding the last help file), followed by the editor.

sonneveld:
I'm still not convinced embedding the help file is a great idea. Adding the "last built help file" isn't a great idea, you need to ensure that you're adding the most up to date artifact.  If you feel strongly, please move this discussion to a new issue.

ivan-mogilko:
>  If you feel strongly, please move this discussion to a new issue.
But if the manual is embedded, this pull request becomes redundant. I think this option should be considered here too, since it's already mentioned.

Are there any other arguments against embedding the manual, except for build-time complications?
IMHO the building step should not impose restrictions on how program works.

ivan-mogilko:
I must confess thought, I was saying that without actually trying embedded help file.
I guess we should first try out if that will work, and whether it will solve the zone identifier issue.
For instance, Editor executable will have zone identifier as well; won't that cause any problems when running help viewer? This needs to be tested out.

ivan-mogilko:
I apologize again, upon thinking I guess it is actually better to move the discussion about embedding manual to another issue, as @sonneveld suggested, because, although providing alternate solution, it is not directly related to this pull request, neither to this code.

Perhaps make a task issue and test embedded solution out before making decision regarding this pull request?

sonneveld:
I still think it's a good idea to keep the build dependencies as simple as
possible. I don't see why we should complicate that when  there are other
solutions. And the normal case is that the user installs through the
installer which doesn't have this problem.

Have we actually checked that if you extract a chm file from an exe to view
it, that the zone identifier is not carried over?

ivan-mogilko:
I added an "experiment" issue (link above), let's discuss embedding option there. This pull request may be delayed until experiment has a result.

Issue #104, opened 2013-12-21
Unified configuration-reading routines


ghost:
There are several routines that read config file, one in the main engine code, others made for each mobile platform separately, and they read from differently named config file.
These files and functions should be unified as much as possible to avoid code duplication.

ivan-mogilko:
Somehow I missed the comment above.
To elaborate, this issue is not about making single config for multiple games, it is about making single code for reading config file.
I will rename the issue to avoid further confusion.

sonneveld:
Related to https://github.com/adventuregamestudio/ags/issues/106 in that the different ports could pass in a config tree instead of reading/writing psp_ variables.

ivan-mogilko:
As an update: years later engine gathers all config options into ConfigTree map before applying it. Filling the ConfigTtree is done in multiple steps, and may include reading key/values from file or adding hard-coded values per platform, but IMHO the goal should be that in the end engine only applies values that are in this map.

ericoporto:
Should `clear_cache_on_room_change` be added as a config under `misc` in acsetup.cfg?

ivan-mogilko:
> Should clear_cache_on_room_change be added as a config under misc in acsetup.cfg?

Do you know what "cache" is this refering to? I suspect it might be the sound cache which is no longer present in the current code. If so, this option may be removed, until a cache is reintroduced.

ericoporto:
ok, looks like this: https://github.com/ericoporto/ags/commit/f61566c859b608580475da54c2fdf2cc4aac4322

now, the missing things are
- **`[misc][rotation]`** : should we actually add this or just let it automatically figure out (find the better fit game resolution -> screen)
  - pros: if we ever do dynamic game resolution to screen size a thing it could be useful
  - cons: it's one more thing the game developer has to remember to properly configure
- **`[sound][cache_size]`** : we haven't implemented this yet, so it's easier to just remove it
- **`[midi][preload_patches]`** : I have no idea what this should do, currently does nothing, so will just remove
- **`[debug][logging]`** : originally made to write to Android logcat or not, I think we can assume the stdout in newer log config is logcat.
- **`[controls][mouse_method]`** : already currently implemented as `[mouse][control_enabled]` in acsetup, sets relative or direct mouse movement.
[...]

PR #286, opened 2015-10-18
IniFile class and reading game config from the generic key-value map

ivan-mogilko:
Few planned future updates requires writing certain options back to config file on exit.
Unfortunately, config write functions provided by Allegro 4 library proved to not be fully reliable: they fail if the config file format is slightly different from their expectations; besides there is a chance we will be moving away from Allegro 4 in the future.

This pull request introduces IniFile class that constructs a description of INI either by parsing a file or by user command, allows to iterate sections and items, and write the contents back into file. IniFile supports merging changes into existing file; this is implemented by remembering precise locations of items in the text. If a single existing item value was changed by user, it is replaced while keeping the rest of the contents unchanged.

Also provided utility functions which exchange data between IniFile and a key-value tree (a map of maps of strings) in both directions. The named tree is then used during actual game setup routine.
The advantage of separating the configuration tree from IniFile is that the config tree may be easily filled from other sources as well, for example - from command line arguments.

Advanced changes that would follow this request:
* Saving necessary config options back to the config file on game exit;
* Combining options from multiple sources (general config, platform-specific config, command line parameters) into the single configuration map.

### DIFF
--- a/Common/util/ini_util.cpp
+++ b/Common/util/ini_util.cpp
@@ -1,0 +1,175 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+
+#include <memory>
+#include "util/file.h"
+#include "util/ini_util.h"
+#include "util/inifile.h"
+#include "util/stream.h"
+#include "util/textstreamwriter.h"
+
+namespace AGS
+{
+namespace Common
+{
+
+typedef std::auto_ptr<Stream>         AStream;
+typedef IniFile::SectionIterator      SectionIterator;
+typedef IniFile::ConstSectionIterator CSectionIterator;
+typedef IniFile::ItemIterator         ItemIterator;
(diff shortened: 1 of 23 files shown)


ivan-mogilko:
On a side note, I would need to recreate engine tests as a separate automatic test project, as was done for Compiler library. I guess I will do so a bit later, as another pull request.

sonneveld:
How does combining settings from multiple sources work with saving back to ini files?

ivan-mogilko:
@sonneveld 
> How does combining settings from multiple sources work with saving back to ini files?

Combining options from sources is done with the use of ConfigTree, which is simply a two-level map (map of string maps). It does not really matter how this map is filled.
IniUtil::Write and IniUtil::Merge functions simply save the resulting tree in INI format; the latter function uses IniFile class as a medium, which helps substituting new values into existing file without clearing existing content.

ivan-mogilko:
Android and PSP ports compiling now (they were using some of the functions I rewrote).
May someone test that they actually work? I cannot do this right now myself.

On a side-note, before someone asks why there is a code duplication between these ports, we already have a related ticket: https://github.com/adventuregamestudio/ags/issues/104

ivan-mogilko:
It seems I need to add new cpp references into iOS project. How do I do that?

sonneveld:
Try this patch!

[ini-xcode.patch.txt](https://github.com/adventuregamestudio/ags/files/20169/ini-xcode.patch.txt)

ivan-mogilko:
@sonneveld 
I cannot apply this patch, the project.pbxproj contents does not match... did you based this patch from another branch?

ivan-mogilko:
Okay, I got it, I squashed fixes and rebased whole pull request on top of recent dev-3.3.5.

PR #288, opened 2015-10-23
iOS project fixes

sonneveld:
Fix native library builds and xcode project to build within TeamCity.


rofl0r:
since svn repos and git repos are moving targets, you should pick the commit id your patches are made against, and supply a download script that fetches just that version (for example git clone foo ; cd foo; git checkout 9fcafebabe). otherwise your patches will break as soon as upstream touches one of the files. this is also necessary to achieve a reproducible build.

sonneveld:
This pull request actually downloads archives of particular revisions now so that hopefully shouldn't be a problem. But still a good point!

PR #289, opened 2015-10-25
Scripting: Improve switch() statement

gurok:
This is an improved implementation of switch(). I found multiple issues with my previous implementation. Using non-local variables inside the switch() part caused problems and it was just way more complicated than it needed to be. It also recalculated the switch part for each case statement, which was incorrect. The new implementation stores the result of the switch() statement on the stack and loads it into a register for comparison when necessary.


gurok:
Just a quick note: I used 3.4.0.6 (and all new language features) extensively when working on a MAGS entry in July. This was the only new statement I found to be buggy (necessitating some workarounds). I meant to make this pull request a lot sooner, but work got in the way.

ivan-mogilko:
A note on automatic tests: mobile ports, compiler tests and installer won't work for 3.4.0 at this point.

PR #292, opened 2015-11-05
Mouse lock option

ivan-mogilko:
This pull request implements an optional mouse lock (aka mouse grab) in game window.

This option was requested by numerous players, because most of AGS games are mouse controlled, and if the game object is located close to the edge of the screen, it is possible to occasionally click outside of the window and loose control over game.
This is even more essential for games where character, or other entity is controlled by cursor movement in real time.

* Mouse lock only activates in windowed mode (fullscreen mode grabs mouse by definition).
* Manual mouse lock is toggled with Alt + Ctrl key combination anytime game has input focus.
* Automatic lock is enabled in config: section "mouse", entry "auto_lock". Automatic lock is performed whenever game receives input focus or user clicked inside window (if lock was previously removed with Alt + Ctrl combo).
* Mouse lock is always removed when game window looses input focus, and (obviously) when game exits.

The pull requests contains lock implementation for Windows and Linux.


sonneveld:
Okay, try this patch for linux libs 

[ags-libs.patch.txt](https://github.com/adventuregamestudio/ags/files/34783/ags-libs.patch.txt)

ivan-mogilko:
Rebased to the head of develop-3.3.5 again.

PR #293, opened 2015-11-18
Editor: Mouse flickering in the code editor

persn:
In reference to: http://www.adventuregamestudio.co.uk/forums/index.php?topic=52805.0

The Scintilla Control uses the text edit mouse cursor, but when WM_SETCURSOR was sent back to WinForms it was overriden and set to the default arrow cursor. This commit makes sure that WM_SETCURSOR is not sent back to WinForms.

### DIFF
--- a/Editor/AGS.Controls/ScintillaNET/ScintillaControl.cs
+++ b/Editor/AGS.Controls/ScintillaNET/ScintillaControl.cs
@@ -640,12 +640,18 @@ namespace Scintilla
         #region Event Dispatch Mechanism
         protected override void WndProc(ref Message m)
         {
-            //	Uh-oh. Code based on undocumented unsupported .NET behavior coming up!
-            //	Windows Forms Sends Notify messages back to the originating
-            //	control ORed with 0x2000. This is way cool becuase we can listen for
-            //	WM_NOTIFY messages originating form our own hWnd (from Scintilla)
-            if ((m.Msg ^ 0x2000) != WinAPI.WM_NOTIFY)
+            if (m.Msg == WinAPI.WM_SETCURSOR)
             {
+                base.DefWndProc(ref m); // Make sure message is sent to Scintilla
+                return;
+            }
+            else if ((m.Msg ^ 0x2000) != WinAPI.WM_NOTIFY)
+            {
+                //	Uh-oh. Code based on undocumented unsupported .NET behavior coming up!
+                //	Windows Forms Sends Notify messages back to the originating
+                //	control ORed with 0x2000. This is way cool becuase we can listen for
+                //	WM_NOTIFY messages originating form our own hWnd (from Scintilla)
+
                 base.WndProc(ref m);
                 return;
             }
--- a/Editor/AGS.Controls/ScintillaNET/WinAPI.cs
+++ b/Editor/AGS.Controls/ScintillaNET/WinAPI.cs
@@ -16,6 +16,7 @@ namespace Scintilla
     public class WinAPI
     {
         public const int WM_NOTIFY = 0x004e;
+        public const int WM_SETCURSOR = 0x0020;
 
         [DllImport("user32.dll", SetLastError = true)]
         public static extern bool MessageBeep(BeepType type);


ivan-mogilko:
@persn , this looks more like a hack. There might be a reason (probably some mistake) for this cursor switch to happen, and "stealing" message from control only hides that mistake.

Since full Stintilla code is located in the repository, I searched for all occurences of SetCursor WinAPI function, and found the culprit.

It is Window::SetCursor() function in the Editor\scintilla\src\PlatWin.cxx:
https://github.com/adventuregamestudio/ags/blob/master/Editor/scintilla/src/PlatWin.cxx#L1041

Commenting the call to SetCursor stops flickering as well.
However, it does not directly answer why this function receives "cursorInvalid". I would like to investigate this more.

persn:
I only did a quick google search where someone else reported that Scintilla conflicted with the hosting .NET control, this inspired me try this specific solution https://groups.google.com/forum/#!topic/scintilla-interest/hw8bjiw9LIs

I see now that the bleeding edge build of ScintillaNET does something similar, but instead of aborting the WndProc it instead forwards it to DefWndProc, I tried this however and it didn't do anything. https://github.com/jacobslusser/ScintillaNET/blob/master/src/ScintillaNET/Scintilla.cs#L2690

Some more investigation into why this is happening is probably wise.

persn:
I'm probably not going to do much more tonight so I'll post today's findings for later reference and convenience.

I think there is some merit to the idea that Scintilla WndProc is being overrode by the hosting control(System.Windows.Forms.Control). A quick debugger-round tells me that calling "base.WndProc" usually forwards the execution into ScintillaWin.cxx, but not always. In some cases the debugger skips gently over the line and don't forward the execution anywhere, which presumably means that it went to Windows.Forms?

By some visual observation and debugging I think I've found that there are 2 situations that trigger WM_SETCURSOR

  1. Mouse enter/leave the scintilla control
  2. Mouse moves around within the borders of the scintilla control

The mouse flickering occurs when moving the cursor inside the scintilla control so situation 2. is the point of interest until other information should present itself. More debugging shows that the LParam property is the only one that appears to be different for the listed scenarios, when casted to int.

 1. 33554439
 2. 33554433

Some quick experiment code confirms my findings, mouse does not flicker

```csharp
protected override void WndProc(ref Message m)
{
    if (m.Msg == WinAPI.WM_SETCURSOR)
    {
        // Mouse moving around in the scintilla control
        if ((int)m.LParam == 33554433)
        {
            //base.WndProc(ref m); // Does not enter ScintillaWin.cxx
            base.DefWndProc(ref m); // Now it enters ScintillaWin.cxx
            return; // Without return future calls of WndProc will reintroduce the mouse flicker
        }
        // Mouse enter/leave the scintilla control
        else
        {
            base.WndProc(ref m);
        }
    }
    more code...
}
```

Now I just need to figure out why...

persn:
Sigh.. Taking a 30-minute break and looking over my previous post again almost everything becomes self-evident. My initial solution was:
  1. Intercept the message
  2. Dump it

When what I really should do is:
  1. Intercept the message
  2. Forward it to the correct receiver (Scintilla instead of WinForms)

So to answer the previous question "Do you have any clue why the cursor type is overriden?". Well we can't really say that it was overridden in any way, since it was never forwarded for processing in the first place. In fact, that is the problem.

This works:
```csharp
protected override void WndProc(ref Message m)
{
    if (m.Msg == WinAPI.WM_SETCURSOR)
    {
        base.DefWndProc(ref m); // Force WM_SETCURSOR messages to be sent to Scintilla
        return;
    }
    else if ((m.Msg ^ 0x2000) != WinAPI.WM_NOTIFY)
    {
        //	Uh-oh. Code based on undocumented unsupported .NET behavior coming up!
        //	Windows Forms Sends Notify messages back to the originating
        //	control ORed with 0x2000. This is way cool becuase we can listen for
        //	WM_NOTIFY messages originating form our own hWnd (from Scintilla)

        base.WndProc(ref m);
        return;
    }
    mode core...
}
```

I don't straight away see any problems with this solution, it's fairly simple, follows the structure already laid out in the function, and is pretty much identical to how WM_SETCURSOR was handled in the current ScintillaNET source, which I boldly assume is a good thing. I'm still a little uncertain however, why does "DefWndProc" work but not "WndProc"? I also don't understand why those calls are sent to "ScintillaWin.cxx" when the class is set to inherit from WinForms Controls, what causes this?

persn:
I updated the pull request to use the solution in my previous message. I haven't been able to find more information on exactly why this is happening but I believe this solution is okay as I am no longer "stealing" the message which was the initial concern.

If there are still some concerns with the fix this pull request will have to "stew" until Christmas/New Years as I will be busy studying for exams.

PR #297, opened 2015-12-17
Restrict savegame & script file paths to locations with assured write access

ivan-mogilko:
This changeset enforces AGS to only create and modify files at locations with assured write access. 

Previous versions still had possibility that some files could still be written to game installation directory (at least under MS Win), but, for example, under MS Windows it could be a subdir of "C:/Program Files", which require special permissions to write into. Namely following items could cause the problem:

1) User configuration file is still located in the game directory. Under MS Windows WinSetup utility writes to that file, and also the 3.3.5 engine now writes particular options to this file upon quitting the game.
2) Script may create custom files in game directory.
3) It is still allowed to create saved games in game directory if game author did not provide custom directory name for saves (this is true for 100% of the games made with AGS 2.72 and earlier).

The solution used is this:

1. **For the old games** the file paths supplied by script are remapped when required. For instance, if script does not tell to write file either in "Saved Games" or "App Data" special directory, but provides a relative path (assuming game directory being parent), then the path is converted into path relative to "App Data".
**For the new games** trying to write to game installation dir will result in script error.

2. For backwards compatibility, when **old game** reads a custom file using path relative to current dir, it first check "App Data" directory; only if file is missing it will also check game installation dir.

3. For savedgames, the custom save directory name is always created; if it was not specified by game author, it is constructed from game title or other game parameters (if title is not present for any obscure reason). The save path is composed of "Saved Games" special directory path and custom game's subdirectory.

4. Optionally, end-user (player) may assign custom location to write all game files (saves, custom files created at runtime) to. This is done by adding following config setting:
<pre>
[misc]
user_data_dir = .
</pre>

5. When game reads configuration, it first reads the config file found in installation directory, then reads the one in "special folder". This way the newer file in "special" location overrides the "default" config file in the installation dir (if there is any).

As a reminder, the platform "driver" code now specifies following "special directories":
- Saved games directory (personal user path);
- Application data directory (assumed to be shared between all users, but it is up to platform driver);
- Configuration directory - added with this pull request.

Currently this pull request specifies configuration directory only for MS Windows, and it is equal to "Saved Games". If required we may add one for other platforms. I am particularly interested to know if Linux needs one specified (as opposed to writing over config file in the game's location).


ivan-mogilko:
Fixed up commits in the branch, and clarified the rule of remapping "unsafe" paths.
Remapping is only done for old games to keep them running; newer games created with AGS are required to use safe paths only.

PR #298, opened 2016-01-04
Minor improvements related to logging to standard output

ivan-mogilko:
This picks out message formatting from platform-dependant code and keeps only simple string output in the respective platform "drivers".

Additionally made Windows version write to stdout if not run under debugger (prior it was always writing to the debugger output, even if it was not present).


ivan-mogilko:
I am still not happy about this; it may be a minor thing, but it makes 1 extra function call when printing out a formatted string.
I will think a little more.

PR #300, opened 2016-01-09
Fixed debian rules: version no longer has -1 at the end

monkey0506:
With the addition of the debian/git-archive-all.sh script (3a234c8), I noticed that the make_ags+libraries.sh script stopped working. As we are no longer appending the "-1" to the version string (360015a), there is no longer a need to check for it. Removing this allows make_ags+libraries to run properly, and AFAICT doesn't affect anything else in the debian build process.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -3,7 +3,7 @@
 # Keep this list of folders that are not included in the
 # orig syncronized with the one in debian/source/local-options!
 ORIG_CONTENT := $(shell git ls-tree --name-only HEAD | grep -Ev '^(Android|Editor|iOS|OSX|PSP|Windows|debian|\.git.*)')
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Should not this be applied to lower branch as well (master, develop-3.3.5)? They have same version format change, and similar contents in debian/rules.

ivan-mogilko:
Also, a small note: according to [Debian conventions](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version) the version string *may* contain '-N' part in the end (i.e. it is optional).

monkey0506:
Sorry, I'm not fully up-to-date on the state of things. Which branch should
I submit a pull request against? I will make the -N an optional part of the
regex instead.

ivan-mogilko:
Yes, the current state could be confusing, because we still have both ongoing updates to 3.3 version while having 3.4 in WIP state.

I guess since the version format change is already in master branch, this fix should be merged to master.

PR #301, opened 2016-01-10
Fixed debian rules: make debian revision optional

monkey0506:
Rebased #300 on master branch, and made debian revision optional instead of removing it. Version info returned by sed will include debian revision, but this does not prevent build scripts from running.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -3,7 +3,7 @@
 # Keep this list of folders that are not included in the
 # orig syncronized with the one in debian/source/local-options!
 ORIG_CONTENT := $(shell git ls-tree --name-only HEAD | grep -Ev '^(Android|Editor|iOS|OSX|PSP|Windows|debian|\.git.*)')
-VERSION := $(shell dpkg-parsechangelog | grep -x "Version:.*" | sed 's@Version: \(.\+\)-.\+@\1@')
+VERSION := $(shell dpkg-parsechangelog | grep -x "Version:.*" | sed 's@Version: \(.\+\)@\1@')
 MAKE = make --directory=Engine
 GENCONTROL_SUBSTVARS = -Vbuildinfo:Date="$(shell date -R)" -Vbuildinfo:Git-Object="$(shell git show HEAD | head -n1)"
 


monkey0506:
Actually the whole bit with the hyphen seems to already be included in sed's regex rules. Deleting it entirely still permits the debian revision, upon further testing.

monkey0506:
I'm not familiar with the commands being used, so pardon my ignorance, but correcting what was said above...

`dpkg-parsechangelog` prints a human-readable report of the information in the changelog.

`grep -x "Version:.*"` searches that report for a **whole-line only** matching the pattern, resulting in output such as "Version: 3.3.4.2" *or* "Version: 3.3.4.2-1" *or* "Version: 3.3.4.2-hotfix".

So the "-.\+" bit in the sed regex pattern was **always** superfluous! I'm going to force an update to correct the pattern now to reflect what I've found.

monkey0506:
The intent may have been to omit the debian revision from the resultant version string, but that isn't necessary, and I see no practical reason to do so (and also it fails if the debian revision is omitted, of course).

monkey0506:
I did some further research and the regex engine used by sed is rather simple. It does not permit a non-greedy quantifier, nor does it support look-aheads or look-behinds. In short, it really isn't possible to omit the debian revision without switching to a more powerful regex engine. Again, I see no need to do this, as the debian revision should not have any illegal path characters or whitespace. As long as the debian revision (if any) is correct, then the other scripts should no longer fail.

rofl0r:
this discussion makes me wonder why ags needs debian-specific code at all.
usually one provides just the source code, and the debian package maintainer maintains the debian-specific infrastructure.
s/debian/linux distro XY/g

monkey0506:
If I sound naive, it's only because I am, but I'm learning when and where I can. We don't *have* a "Debian package maintainer". The entire Linux port is sorely lacking in attention, but we try to make due with those who do have the time, skill, and motivation to get things done. That's why we have these build scripts. It may be less than ideal, but it allows those, like myself, who lack experience with Linux to build AGS and get games running.

ivan-mogilko:
That's actually what I was thinking earlier, because I have to update this changelog, although it is not done correctly anyway (it is not a proper debian changelog, as far as I can tell), and I feel uneasy doing things that I do not understand well.

tobihan:
Since AGS is not in Debian and we want to build deb packages anyway,
it's required to maintain the debian folder here. Having this deb
package stuff is required to build AGS with cowbuilder which is the
easiest way to build it for an arbitrary version of Debian or Ubuntu
without having to install that version. The changelog just happens to be
the place where the version of a deb package is defined, although it
does not make sense to keep a real changelog here in this use case.

ivan-mogilko:
Would it be a viable alternative if we create a separate repository with only Debian-related files, and some volunteer will maintain it? It could have ags source as dependency, or submodule.

tobihan:
The only thing that needs to be changed regularly is the version number and it makes most sense if that is done by the person who changes the ags version. I won't keep track of ags versions in multiple branches just to change it in the debian changelog every time.

PR #302, opened 2016-01-11
Fixing the cases of redundant consequent vsprintfs and text translation in the Script API functions

ivan-mogilko:
This fixes my oversight made long ago when I was rewriting Script interpreter, introducing safer argument list.

A number of engine functions kept redundant variadic parameter list and string printf-to-buffer operations, while they were always called with "%s" format string in every practical case, and required just a constant string.

Additionally, *get_translation()* was being called for the "%s" string in all those functions.

Aside from doing excess work, this caused a bug. A *source_text_length* global variable is set inside *get_translation()* to remember pre-translated text length, which in turn could be used to calculate time required to display the text message on screen. Of course, since get_translation was being called for the second time for "%s" string, this length was always considered to be 2 symbols; hence messages were displayed on screen for a minimal time, making it hard to read them.
(This behavior depends on game parameters, and does not take effect in every game.)

In this pull request these redundant operations are removed, and bug fixed.

### DIFF
--- a/Engine/ac/global_api.cpp
+++ b/Engine/ac/global_api.cpp
@@ -74,7 +74,7 @@ extern ScriptString myScriptStringImpl;
 RuntimeScriptValue Sc_sc_AbortGame(const RuntimeScriptValue *params, int32_t param_count)
 {
     API_SCALL_SCRIPT_SPRINTF(_sc_AbortGame, 1);
-    _sc_AbortGame("%s", scsf_buffer);
+    _sc_AbortGame(scsf_buffer);
     return RuntimeScriptValue();
 }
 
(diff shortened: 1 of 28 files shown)


Engine/ac/global_api.cpp
```diff
@@ -2196,101 +2196,71 @@ RuntimeScriptValue Sc_WaitMouseKey(const RuntimeScriptValue *params, int32_t par
 // void (char*texx, ...)
 void ScPl_sc_AbortGame(const char *texx, ...)
```
rofl0r: why do these funcs keep the variadic prototype ?

ivan-mogilko: They are still part of plugin API.

PR #303, opened 2016-01-12
Dev 3.3.5  tidy font

jclee:
Just trying to tidy up some code in the font renderers:

* Eliminate some globals
* Limit some globals to static scope
* Prefer RAII to malloc
* Use const more
* Eliminate some unnecessary casts

Please let me know if this sort of thing is helpful.

### DIFF
--- a/Common/font/ttffontrenderer.h
+++ b/Common/font/ttffontrenderer.h
@@ -36,6 +36,4 @@ private:
     std::map<int, ALFONT_FONT*> _fontData;
 };
 
-extern TTFFontRenderer ttfRenderer;
-
 #endif // __AC_TTFFONTRENDERER_H
--- a/Common/font/ttffontrenderer.h
+++ b/Common/font/ttffontrenderer.h
@@ -17,6 +17,10 @@
 
 #include "font/agsfontrenderer.h"
 
+#include <map>
+
+struct ALFONT_FONT;
+
 class TTFFontRenderer : public IAGSFontRenderer {
 public:
     virtual bool LoadFromDisk(int fontNumber, int fontSize) override;
@@ -27,6 +31,9 @@ public:
     virtual void RenderText(const char *text, int fontNumber, BITMAP *destination, int x, int y, int colour) override;
     virtual void AdjustYCoordinateForFont(int *ycoord, int fontNumber) override;
     virtual void EnsureTextValidForFont(char *text, int fontNumber) override;
+
+private:
+    std::map<int, ALFONT_FONT*> _fontData;
 };
 
 extern TTFFontRenderer ttfRenderer;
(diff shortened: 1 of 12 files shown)


Common/font/ttffontrenderer.h
```diff
@@ -17,18 +17,23 @@
...
-  virtual bool LoadFromDisk(int fontNumber, int fontSize);
-  virtual void FreeMemory(int fontNumber);
-  virtual bool SupportsExtendedCharacters(int fontNumber) { return true; }
-  virtual int GetTextWidth(const char *text, int fontNumber);
-  virtual int GetTextHeight(const char *text, int fontNumber);
-  virtual void RenderText(const char *text, int fontNumber, BITMAP *destination, int x, int y, int colour) ;
-  virtual void AdjustYCoordinateForFont(int *ycoord, int fontNumber);
-  virtual void EnsureTextValidForFont(char *text, int fontNumber);
-};
+    virtual bool LoadFromDisk(int fontNumber, int fontSize) override;
```
monkey0506: The project files for Windows are still restricted to Visual Studio 2008, which doesn't support the override keyword. Until the project files can be updated it may make sense to use a macro for this.

ivan-mogilko:
Greetings, @jclee , I really appreciate your will to help improve the code, but there is one thing about current minor update branch (3.3.5): we are close to make a release, and have a convention to not  change anything that is not directly required for planned features or bug fixes.

There is another reason: we have code refactored in many places in the following WIP version (develop-3.4.0), and the font code was changed as well. If any more refactoring is to be applied, it should be applied over these changes.

I suggest you check develop-3.4.0 branch and see if your proposed changes may be applied there.

ivan-mogilko:
Also, like @monkey0506 already mentioned, we do not use C++11 features in our code yet, because  some of the compilers used to build the engine and ports do not fully support them.
It is probably possible to work around this with the use of macros; for example create an "override_" macro which would be defined as "override" for compilers that support it, and as empty symbol for those that don't (I was considering this some time ago too).

jclee:
Sure, I can move the work to the develop-3.4.0 branch.  I wasn't sure where development should be taking place; it looked like more recent commits were going into develop-3.3.5 than develop-3.4.0.

And actually, "override" *is* available in Visual Studio 2008.  I assumed that VS2008 was the compiler that constrained the set of language features in play...  However, looking at the build output, the PSP compiler is also using something prior to gcc 4.7, which is unfortunate.

I can certainly omit C++11-specific code from pull requests.  However, I was planning to use C++11 in some personal experiments with AGS code.  In light of the compiler constraints, I guess this may limit their relevance to AGS at large.

ivan-mogilko:
AFAIK an "override" was available in MSVC2008 simply because it was Microsoft-specific keyword before C++11.

Latest versions of GCC compiler (usually used for Linux port) potentially support C++11, but we cannot restrict the code to C++11, because Linux systems are varied, and some may not have updated version of compiler yet. Using macros, like in case with "override", seems like a possible way to go, but it is not clear if they may help with more complex C++11 features.

PR #305, opened 2016-01-19
Remove changelog dependency from debian/make_ags+libraries script

monkey0506:
As noted in #301 the debian/changelog file is actually a *fake* changelog which only exists to allow the debian package tools to function correctly. With this change, the make_ags+libraries script will now read the version info from Common/core/def_version.h, update the changelog file so the debian package tools can run, then reset the changelog file to prevent unnecessary commits.

This also removes the use of dpkg-parsechangelog from debian/rules, because I found that it requires loading several external libraries just to parse the version info. This could be made as a separate commit, if necessary.

It may additionally be worth noting in the debian readme that the changelog is not meant to be changed or updated.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-ags (3.3.4.2) UNRELEASED; urgency=low
+ags (3~git-1) UNRELEASED; urgency=low
 
   * some release
 
--- a/debian/make_ags+libraries.sh
+++ b/debian/make_ags+libraries.sh
@@ -46,7 +46,9 @@ sed -i -r "5s/.*/BIT=32/" $BASEPATH/debian/ags+libraries/hooks/B00_copy_libs.sh
 
 # Build the source package:
 cd $BASEPATH
-VERSION=$(dpkg-parsechangelog | grep -x "Version:.*" | sed 's@Version: \(.\+\)@\1@')
+CHANGELOG_VERSION=$(head -1 debian/changelog | sed 's/.*(\(.*\)).*/\1/')
+VERSION=$(grep '#define ACI_VERSION_STR' Common/core/def_version.h | sed 's/.*"\(.*\)".*/\1/')
+sed -i -- "s/$CHANGELOG_VERSION/$VERSION/" debian/changelog
 debian/rules get-orig-source
 debuild -us -uc -S
 
@@ -94,3 +96,6 @@ cd $BASEPATH && tar -cf - ags+libraries/ | xz -9 -c - > ags+libraries_$(date +%Y
 # Delete the ags+libraries folder, because some of the subfolders and files are owned by root.
 # Extracting the .tar.xz yields the folder with proper owner/permissions.
 sudo rm -rf $BASEPATH/ags+libraries/
+
+# Restore changelog version to prevent unnecessary commits
+sed -i -- "s/$VERSION/$CHANGELOG_VERSION/" $BASEPATH/debian/changelog
--- a/debian/rules
+++ b/debian/rules
@@ -3,7 +3,7 @@
 # Keep this list of folders that are not included in the
 # orig syncronized with the one in debian/source/local-options!
 ORIG_CONTENT := $(shell git ls-tree --name-only HEAD | grep -Ev '^(Android|Editor|iOS|OSX|PSP|Windows|debian|\.git.*)')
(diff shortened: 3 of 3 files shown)


monkey0506:
The 'g' in the sed commands in make_ags+libraries should not have been part of the commit, and add unnecessary overhead. I will fix that prior to merging.

tobihan:
This assumes that Debian packages are only built using the
make_ags+libraries.sh script. This is not true, there are also
instructions how to build a Debian package in debian/README.md and for
example at
https://wiki.ubuntuusers.de/Adventure_Game_Studio/
and it's generally the best way for end users to install AGS on Debian
and Ubuntu.

monkey0506:
That's mostly true, but this doesn't remove the changelog file. It simply removes the dependency of the make_ags+libraries script on the changelog's version info. Perhaps the name of the request should be changed, but this doesn't prevent anyone from updating or using the changelog file for other purposes.

monkey0506:
It has already been proposed that the debian/changelog should not be maintained as part of the main repository because it adds an additional location which must be updated by hand whenever the version changes. This allows the debian/changelog to be maintained however is deemed necessary or fit without having to touch it at every point version of the editor or engine. Additionally I will reiterate that the file is already a fake changelog that is only present to assist in building, so I greatly doubt that it places undue burden on anyone building manually to have to update the package version by hand.

tobihan:
As long as the normal package build works, I'm fine. If the version in the changelog is not updated that means that a working package with a wrong version is built.

monkey0506:
Well arguably the debian package version is not intrinsically part of the AGS source code anyhow, so I would say that it's absolutely up to the package maintainer to update the version prior to building. Let me be totally clear:  this does **nothing** to the build process, and **only** allows make_ags+libraries to do a just-in-time update of the debian package version that it produces.

tobihan:
Yes, I saw that the package build should not be affected.

There is no package maintainer in the Linux distribution sense, because AGS is not part of the distributions. There are however people who want to build and install AGS themselves. If you are on Debian/Ubuntu, building the package is better than "make install", because this way one makes sure that the dependencies stay installed and it can be cleanly removed later. For this use case having a wrong version number on the package can be confusing, but most people probably won't care.

tobihan:
How about this: We set the version in the changelog to "3~git". This way the packages people build for themselves do not have a wrong version number. I used ~ instead of . because that way the version is lower than anything starting with "3." so the package can later be replaced by something like 3.4. "3.git" would be a higher version than "3.4" according to the Debian tools.

monkey0506:
Perhaps I'm using the wrong terminology here, but when I said "package maintainer", I literally meant whoever is building **that particular package**, whether it's part of some larger distribution channel, distributed by itself, or meant only for local installation. If *you* are building the package yourself for installation on your system, then I think it is not unreasonable for *you* to have to update the package version. That's my thoughts on the matter. Others have voiced similar opinions, at least in respect to the AGS source code's git repo.

I don't know enough about Linux or what is considered the norm here, but perhaps it would be possible for other build methods to take a similar approach to ensure that the package version is up-to-date? This also raises the question, should the debian package version always match the AGS version? I really don't know.

tobihan:
On Linux the person building the package is the end user, so it should be as simple as possible. Copy four commands from the readme (install dependencies, clone git repo, build package, install package) and that's it. I don't want to require people to edit files. But in this case it's ok if the package does not have the precise version. Since it's too much hassle to keep the version up to date, that's the best solution. We can change the version to 3~git after this pull request is merged.

tobihan:
Actually we have to change it to 3~git-1, because building the package with dpkg-buildpackage is otherwise broken. "fakeroot debian/rules binary" still works.

monkey0506:
I'm not saying your inherently wrong about keeping the build process simple, but this is more about keeping the repo clear of unnecessary changes. Keeping the version up-to-date in multiple places is objectionable, especially when this case deals with a "package version" that isn't intrinsically the same thing as the "AGS version". I need to split and fix-up this commit before it's pulled in anyway. I can also update the changelog version to "3~git-1" if you feel that's the appropriate thing to do here.

tobihan:
Yes please do, that should solve the problem.

monkey0506:
Unless anyone else has further feedback, this should be good to go, I believe.

PR #306, opened 2016-01-25
Fixed output directory for Windows plugins

monkey0506:
This was previously overlooked when the build target interface was added.

Although this only affects building for Windows, this code should not be moved to BuildTargetWindows (for the time being) because BuildTargetWindows is not invoked if using the "legacy compiler" setting. It also doesn't make sense to move this into BuildTargetDataFile (which is invoked, respecting the legacy setting) as this has nothing to do with building the data file.

If the "legacy compiler" setting is later removed, then it makes sense for this to be moved into BuildTargetWindows at that time.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1224,7 +1224,14 @@ namespace AGS.Editor
 
                 foreach (Plugin plugin in _game.Plugins)
                 {
-                    File.Copy(Path.Combine(this.EditorDirectory, plugin.FileName), Path.Combine(OUTPUT_DIRECTORY, plugin.FileName), true);
+                    string outputDir = OUTPUT_DIRECTORY;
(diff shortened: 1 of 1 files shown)


monkey0506:
I said this shouldn't be moved, but if it were refactored into a separate function then there probably wouldn't be harm in moving it into BuildTargetWindows. I was looking back over the build process and rather forgot how much of a mess it really all is.

In preserving the legacy build settings alongside the new build settings, I invoke NativeProxy methods in some cases where the Native methods are no longer being used. That really doesn't make sense, and is confusing at best. This should be further refactored itself, and the point of invocation should check whether the legacy builder is being used, rather than doing so in the NativeProxy methods. Not fully understanding/remembering how this was implemented was part of the rationale of my previous statement, though I think actually I should instead fix up the calls to NativeProxy and refactor this out to a separate function that can be invoked as needed by the legacy builder. Similar code already exists to update the Windows EXE by invoking methods from BuildTargetWindows directly.

PR #307, opened 2016-01-25
Editor: Fixed index out of range exception

gurok:
This fixes a bug in auto-complete, the simplest test case for which is a new script file with:
```
{
//
```
Insert a new line *before* the `{` and start typing `int`. The editor will encounter an index out of range exception. The `IncrementIndexToSkipAnyComments` function checked for a multiline comment immediately after a single line comment, but did not ensure that the index was within the bounds of the script. It also only advanced past one comment at a time (or maybe two if you had a multiline following a
single line). I rewrote the function to do what I thought it should do: advance past all comments in succession and return true if the end of the script was reached.


ivan-mogilko:
Another way to cause this error:
* start with the same script sample;
* save the script

The worst thing is that project cannot be reopened after this (Editor raises an exception on load).

PR #311, opened 2016-02-04
Compiler: Better parsing of negative numbers

### LINKED ISSUE #295: Incorrect parsing of INT_MIN value in script
In develop-3.3.5 branch we had fixes related to parsing large integer values, however script compiler still cannot parse INT_MIN value (-2147483648) properly.

Example of script:
<pre>
int a = -2147483648;
</pre>

Instead it reports following error:

<pre>
Error while parsing integer symbol '2147483648'.
</pre>

Notice: it also does not include sign when reporting a symbol it cannot parse.

Values higher than INT_MIN are parsed properly. For example following script produces correct expected result:
<pre>
int a = -2147483647;
a--;
Display("a = %d", a); // this displays "a = -2147483648"
</pre>

gurok:
This fixes #295 

The tokeniser will now recognise negative numbers as single symbols. `-` continues to be the unary negation operator for situations where the sign and number are separated by a space, e.g. `int x = - 4;`. The specific case of someone negating a negative constant is disallowed.

Some interesting edge cases:
```
int a = 3 - -4; // Legal, previously not allowed
int a = a - -2147483648; // Ditto
int a = a - 2147483648; // Not legal, +2147483648 is out of range
int a = - 2147483648; // Legal (unary negation)
```

Note: This behaviour is in line with C#.

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -1702,28 +1702,6 @@ int call_property_func(ccCompiledScript *scrip, int propSym, int isWrite) {
   return 0;
 }
 
-int accept_literal_value(int &value, int symidx) {
-    errno = 0;
-    char *endptr = 0;
-    const long longValue = strtol(sym.get_name(symidx), &endptr, 10);
-
-    if ((longValue == LONG_MIN || longValue == LONG_MAX) && errno == ERANGE) {
-        cc_error("Could not parse integer symbol '%s' because of overflow.", friendly_int_symbol(symidx, false).c_str());
-        return -1;
-    }
-    if (endptr[0] != 0) {
-        cc_error("Could not parse integer symbol '%s' because the whole buffer wasn't converted.", friendly_int_symbol(symidx, false).c_str());
-        return -1;
-    }
-    if (longValue > INT_MAX || longValue < INT_MIN) {
-        cc_error("Could not parse integer symbol '%s' because of overflow.", friendly_int_symbol(symidx, false).c_str());
-        return -1;
-    }
-
-    value = static_cast<int>(longValue);
-    return 0;
-}
-
 int do_variable_memory_access(ccCompiledScript *scrip, int variableSym,
                               int variableSymType, bool isProperty,
                               int writing, int mustBeWritable,
@@ -1731,7 +1709,7 @@ int do_variable_memory_access(ccCompiledScript *scrip, int variableSym,
                               int soffset, bool isPointer,
                               bool wholePointerAccess,
                               int mainVariableSym, int mainVariableType,
-                              bool isDynamicArray) {
+                              bool isDynamicArray, bool negateLiteral) {
   int gotValType = 0;
   int readcmd = get_readcmd_for_size(sym.ssize[variableSym], writing);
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Please pay attention to Compiler Tests. They all must be successful if this issue is fixed properly.

ivan-mogilko:
Seeing your comments in code, I also suggest to add more compiler tests as a part of this pull request.
For example, a test for double negation symbol, and for "A-B" case (to ensure it parses this as subtraction operation, and not as negative literal).

gurok:
I see that I was too hasty to make a pull request, yet again. Sorry. I really didn't mean to waste your time. I have been accustomed (wrongly) to ignoring the test results because when making pull requests on 3.4.0, the tests routinely fail. I should've checked the tests results in this case and I didn't check the second commit thoroughly enough. I'll try to correct this in future.

As I've previously ignored them, I don't actually know how to run the tests without making a pull request. I'm sure there's a way, but I've just never discovered it. This is something I'll have to look into too. I am guessing there's a run configuration for doing Compiler Tests locally, but I'd like to find out if there's a way to invoke TeamCity tests on an arbitrary branch via GitHub, as that seems more thorough and convenient.

Regarding the failures themselves:

The failed `Compile.ParsingIntLimits` test is something I missed. I will look into it tonight and over the weekend.

I'm pretty sure the other two failures are due to differences in error messages, as you mentioned. The messages have changed because the compiler is using `accept_literal_or_constant_value` in all cases now. The old error message was more of a special case associated with `accept_literal_value`. I will provide a commit that updates those tests accordingly.

Lastly, I agree, those commented areas should have tests associated with them. I will add the tests you mentioned and any other important things I think need testing in a further commit to this pull request.

ivan-mogilko:
To run tests locally, just use Compiler.Lib.sln solution in the Solutions folder.
You may need to update googletest submodule to make it work (we are getting headers from that repository).

gurok:
I spent a long time looking over this, and I think I've found a better solution than the tokeniser modification.

The expression parser now treats negated literals on the left-hand side of an expression as a special case, passing this information through to the point of variable access, where the common `accept_literal_or_constant_value` can be flagged to negate the value in the token. For other negated values, the existing generalised negation code (subtract from zero) is used.

Furthermore, the expression parsing function now recognises negation of a value on the right-hand side of an expression. Note: With the way expression parsing works, the right-hand side will eventually become the left-hand side, and so negated *literals* on the right of an operator need only be flagged when they're on the left in order to be parsed correctly.

Handling the negation operator this way means there's no need to make a special case in the tokeniser for `-4` over `- 4`. They're both a `SCMD_SUBREG` operator followed by a `SYM_LITERALVALUE` and are both treated the same way.

I did not make a separate commit for the merging of `accept_literal_value` with `accept_literal_or_constant_value` as `accept_literal_value` did not allow for negation in a separate symbol.

Aaaaand all tests pass. Thank you for your guidance on using the test suite, CW. Also, what a difference a clear mind and not trying to do this after work make.

ivan-mogilko:
@gurok , this seem working well, but could you bring indentation to same kind (spaces)?

gurok:
Sorry about that. Removed the errant tabs.

PR #314, opened 2016-02-28
Installer changes for release

sonneveld:
- Versions are stored in version.json
- Source files can be updated with push_version.py
- Installer uses version.json to set version and appid
- Installer allows selecting engine without mp3 support
- small unicode fix to ags.tex


ivan-mogilko:
Just to clarify, the GUID in the version.json is already a new one?

sonneveld:
Corrrect.. I generated a new GUID.

PR #316, opened 2016-03-05
Packaging: script to create release packages with checksums

sonneveld:
The included script ensures that the Teamcity Release job includes all files needed for release (including installer) with appropriate names and checksums.

### DIFF
--- a/Script/build_packages.py
+++ b/Script/build_packages.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+from collections import namedtuple
+import codecs
+import hashlib
+import json
+import os
+import os.path
+import re
+import shutil
+import time
+import zipfile
+
+WINDOWS_IS_STUPID_RETRIES=8
+WINDOWS_IS_STUPID_DELAY=2
+
+
+SCRIPT_LOC=os.path.dirname(os.path.realpath(__file__))
+WORKSPACE_DIR=os.path.join(SCRIPT_LOC, "..")
+
+def workspace_dir():
+    assert os.path.exists(os.path.join(WORKSPACE_DIR, "version.json"))
+    assert os.path.exists(os.path.join(WORKSPACE_DIR, "Copyright.txt"))
+    return WORKSPACE_DIR
+
+def workspace_rel(path):
+    return os.path.join(workspace_dir(), path)
+
+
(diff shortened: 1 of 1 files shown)


sonneveld:
One possibly weird thing: Only AGS-3.3.5.zip has a directory in it.  The -PDB and -noMP3 zips files just contain the files.  My reasoning was that people would unzip AGS-3.3.5 to get the install dir plus files then unzip PDB and noMP3 inside that if needed.

sonneveld:
Actually, thinking about it, maybe all archives should have the same directory "AGS-3.3.5/" so if you unzip them up they all end up in the right spot

monkey0506:
I would **very** much prefer that the archive **not** have a directory in it. It is a very simple step to create a directory from any extraction tool. Putting a directory into the archive with the wrong name (e.g., one that doesn't match the naming convention I personally am using on my HDD) means that I have to take an additional step and rename the directory after extraction.

sonneveld:
How often are you unzipping this? Whether apps, that are distributed via .zip files, include a directory or not seems to vary. It seems to be used more often in source tarballs.  I prefer the directory in archive format because it means you never accidentally unzip a selection of random files in your directory if you forget to specify "unzip into directory". It ensures that the application is bundled in a directory with an explicit version. Plus you could argue that previous release archives also had a directory so the decision has already been made.

I feel this topic could get very bikesheddy.

rofl0r:
usually one makes such packages on linux where sha512sum, md5sum etc are readily available and the whole thing can be put in a 20 line shell script rather than introduce a new burden for the maintainer in form of a yet another new scripting language.
imagine a project with 5 contributors where each one adds scripts in his preferred language: perl, python2, python3, ruby, javascript.... that will end up in a huge mess.

ivan-mogilko:
Nick, I am too not sure why do we need this right in the repository. I thought the idea was to make build server do this?
Could you elaborate?

sonneveld:
It's easier to keep track of these scripts if they are part of the repository.  I picked Python because it's a scripting language I knew and it works on Windows where most of the developers are.

ivan-mogilko:
Nick, I have just one small request, could you clean the pull request branch so that it did not contain commits that first add particular code and then remove one?

sonneveld:
No worries.. I just pushed a squashed version.

PR #317, opened 2016-03-07
Linux: Revert mouse behaviour for testing

morganwillcock:
This reverts the commit for fixing X11 mouse lag which seems to cause with input polling (very long delay ~20 seconds) and enables mouse control. It seems that, without mouse control enabled, cursor movement restrictions are not aligned with game display when running fullscreen  - the issue may only show where game display does not fill the screen, i.e. running fullscreen with borders.

Provides opportunity for anyone using the Linux build to build it and report issues.

### DIFF
--- a/Engine/platform/linux/acpllnx.cpp
+++ b/Engine/platform/linux/acpllnx.cpp
@@ -49,6 +49,7 @@ struct AGSLinux : AGSPlatformDriver {
   virtual const char *GetAppOutputDirectory();
   virtual unsigned long GetDiskFreeSpaceMB();
   virtual const char* GetNoMouseErrorString();
+  virtual bool IsMouseControlSupported(bool windowed);
   virtual const char* GetAllegroFailUserHint();
   virtual eScriptSystemOSID GetSystemOSID();
   virtual int  InitializeCDPlayer();
@@ -137,6 +138,11 @@ unsigned long AGSLinux::GetDiskFreeSpaceMB() {
 
 const char* AGSLinux::GetNoMouseErrorString() {
   return "This game requires a mouse. You need to configure and setup your mouse to play this game.\n";
+}
+
+bool AGSLinux::IsMouseControlSupported(bool windowed)
+{
+  return true; // supported for both fullscreen and windowed modes
 }
 
 const char* AGSLinux::GetAllegroFailUserHint()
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
From testing, there is some mistake in code that breaks cursor functionality under following conditions:
* Fullscreen mode;
* Mouse speed control is disabled;
* Game frame is not stretched to whole window (letteboxed/pillarboxed mode), e.g. has certain offset.

This is reproducible regardless of platform.

ivan-mogilko:
This ^ issue should be fixed with bc146fd7622235c9fed2cb7a6ad3d9c7dd07945a

ivan-mogilko:
Since the virtual machine has its own issues, I would like to clarify how that lag patch acts on your side.
For me:
* The issue happens only when I have no scaling filter. At least this is when it becomes noticeable.
* It looks like the game is redrawing only every 5 seconds or so. I was testing this on game that has scrolling background during the intro, and the background scrolls, which means that game actually updates, but screen is redrawn with long pauses.

morganwillcock:
The fix for the cursor boundaries works for me.

For the lag patch, I see the same results irrespective of being in a window or fullscreen:
- no scaling - input lag
- StdScale2 - no input lag
- Hq2x - input lag

Also noteworthy, when running the game in a window and experiencing lag (so using either 'no scaling' or 'Hq2x') there is also a delay of around 8 seconds when trying to close the engine window (clicking on the window decorator). This would suggest that something must be fundamentally different about how the event queue is processed when StdScale2 is used.

Which game are you using to test? If I use the same one it would be interesting to see if we both see the same length of delay, for input and redraw.

ivan-mogilko:
I used "Heroine's Quest".

morganwillcock:
Okay I see the same. I get about 4 or 5 frames of animation, then an 8 second delay, then 4 or 5 frames of animation, then an 8 second delay, etc...

ivan-mogilko:
The patch author confirmed that this breaks the game when the engine is patched, but works well (for him) when Allegro library is patched, so I am thinking to accept reverting this commit.

morganwillcock:
How do you feel about the new mouse control routine? For me the cursor speed control works but I think it needs further testing at high resolution and where users have customised their system mouse speed. Unfortunately I've only got older equipment to test with but perhaps someone on the forums would be running something newer, perhaps with HighDPI scaling.

ivan-mogilko:
In theory, there should not be any difference in mouse movement if the cursor speed is remaining at 1.0. But since I found that weird behavior on virtual machine, I think I could add an optional config option that would disable mouse control.
I had such option in the original variant of the code, but then I thought it won't be required...

PR #321, opened 2016-04-09
Scripting: Clarify the unclosed brace error message

gurok:
Addresses the issue reported here:
http://www.adventuregamestudio.co.uk/forums/index.php?issue=718.0

AGS will now distinguish between runtime errors and errors where the line number is not known. Additionally, the line number is now reported for "Function still open, missing }" compilation errors.

Note: I understand there is a big merge from 3.3.x -> 3.4.0 happening right now, so I don't expect this to be merged right away.

### DIFF
--- a/Common/script/cc_error.cpp
+++ b/Common/script/cc_error.cpp
@@ -19,6 +19,7 @@
 #include "script/script_common.h"  // current_line
 
 extern void cc_error_at_line(char *buffer, const char *error_msg);
+extern void cc_error_without_line(char *buffer, const char *error_msg);
 
 int ccError = 0;
 int ccErrorLine = 0;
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
@gurok, the merge was completed; could you please rebase your three pull requests (#321, #322, #323) on top of latest develop-3.4.0?

This will also make use of up-to-date compiler tests.

gurok:
I think these should be fine to merge now. I think I did the rebasing correctly. I looked at the history of all of these branches, and they all look right.
I am assuming the Engine PSP test is broken. I get an error from it in every pull request.
One of the automatic tests for #323 timed out too. I'm not sure what to do about that.

PR #322, opened 2016-04-09
Scripting: Switch statement improvements

gurok:
A couple of things left out of the previous implementation. Continue statements inside switch blocks were being incorrectly rejected as outside a loop. A `case` statement immediately following a `default` label would be evaluated.

Note: I understand there is a big merge from 3.3.x -> 3.4.0 happening right now, so I don't expect this to be merged right away.

### DIFF
--- a/Compiler/script/cc_compiledscript.cpp
+++ b/Compiler/script/cc_compiledscript.cpp
@@ -235,52 +235,6 @@ void ccCompiledScript::write_code(intptr_t byy) {
     codesize++;
 }
 
-intptr_t ccCompiledScript::yank_chunk(int32_t start, intptr_t **nested_chunk, int index) {
-    intptr_t chunk_size;
-    intptr_t chunk_index;
-
-    if (start < codesize) {
-        chunk_index = chunk_size = codesize - start;
-        nested_chunk[index] = (intptr_t *) malloc(chunk_size * sizeof(intptr_t));
-        while(chunk_index-- > 0) // speed up with memcpy?
-            nested_chunk[index][chunk_index] = code[start + chunk_index];
-        codesize = start;
-    }
-    else
-        chunk_size = 0;
-
-    return chunk_size;
-}
-
-void ccCompiledScript::write_chunk(intptr_t **nested_chunk, int index, intptr_t chunk_size, bool dispose, int fixup_start, int fixup_stop, int32_t adjust) {
-    intptr_t chunk_index;
-    if(chunk_size > 0)
-    {
-        if(codesize + chunk_size >= codeallocated) {
-            codeallocated += chunk_size + 500;
-            code = (intptr_t *) realloc(code, codeallocated * sizeof(intptr_t));
-        }
-        chunk_index = chunk_size;
-        while(chunk_index-- > 0) // speed up with memcpy?
-            code[codesize + chunk_index] = nested_chunk[index][chunk_index];
-        codesize += chunk_size;
-        if(dispose)
-            free(nested_chunk[index]);
-    }
-    if(dispose)
-        while(fixup_start < fixup_stop)
-        {
-            fixups[fixup_start] += adjust;
-            fixup_start++;
-        }
-    else
-        while(fixup_start < fixup_stop)
-        {
-            add_fixup(fixups[fixup_start] + adjust, fixuptypes[fixup_start]);
-            fixup_start++;
-        }
-}
-
 const char* ccCompiledScript::start_new_section(const char *name) {
 
     if ((numSections == 0) ||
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
With the second change following script won't work properly:

<pre>
int test;

  test = 3;
  switch(test)
  {
  case 1:
    Display("test 1");
    break;
  default:
  case 2:
    Display("test 2");
    break;
  case 3:
    Display("test 3");
    break;
  }
</pre>

This script will perform case 2, instead of case 3.

gurok:
A little too optimised. I'll fix this, sorry.

gurok:
@ivan-mogilko I think I need to rewrite this. default: labels can't work with the way it's written now. I'm going to implement a lookup table that handles all cases. Something like:

```
// Switch part
Load switch value into the BX register
Jump to :lookuptable
Jump 2
Jump to :end // Used as point to jump to for break statement
// Various entry points and case statements
// ... :entrypoint1
// ... some code
// ... :entrypoint2
// ... some code
// ... :defaultentrypoint
// ... some code
// :lookuptable
Load case value 1 in to AX
If AX == BX, jump to :entrypoint1
If AX == BX, jump to :entrypoint2
... etc.
// Only emitted if default: is present
Jump to :defaultentrypoint
// :end
```

With this, I think there will be no need for pushing/popping the expression on the stack. I will also write a check to see whether default has been used twice (oversight in current implementation).

ivan-mogilko:
Well, it is better to take time and not make any hasty changes. The 3.4.0 still requires certain improvements before final release.

gurok:
I think this implementation is ready now. I tested it with many switch statement forms, including some pretty esoteric ones:

```
int j = 3;

function testFunc(int a)
{
	return(a);
}

function Gauntlet()
{
	int i;

	i = 2;
	for(int a = 0; a < 100; a++)
	{
		switch(true)
		{
			case a > 0:
				continue;
		}
		if(a == 0)
			Display("%d", a);
		else
			Display("X");
	}
	switch(2)
	{
		case 1:
			Display("A");
			break;
		case 2:
			Display("1");
			break;
		default:
			Display("B");
			break;
	}
	Display("2");
	switch(false)
	{
	}
	Display("3");
	switch(true)
	{
		default:
			Display("4");
	}
	Display("5");
	switch(i)
	{
		case 1:
			Display("C");
			break;
		case 2:
			Display("6");
			break;
		default:
			Display("D");
			break;
	}
	switch(i)
	{
		default:
		case 1:
			Display("7");
			break;
	}
	switch(testFunc(2))
	{
		case 1:
			Display("E");
			break;
		default:
		case 2:
			Display("8");
			break;
	}
	switch(player.x)
	{
		case 1:
			Display("F");
			break;
		case 2:
			Display("G");
			break;
		default:
			Display("9");
			break;
	}
	switch(j)
	{
		case 1:
			Display("H");
			break;
		case 2:
			Display("I");
			break;
		default:
			Display("10");
			break;
	}
	switch(player)
	{
		case character[0]:
			Display("11");
			break;
	}
	switch(true)
	{
		case j * 22 == 66:
			Display("12");
			break;
	}

	return;
}
```

It's not formal, but the above compiles and runs with the numbers going from 0-12, in order.

ivan-mogilko:
With the recent code you cannot modify a variable used in the switch condition inside the "case":

<pre>
    int test = 1;
    switch(test)
    {
    case 1:
      test = 0;
      break;
    default:
      test = 2;
    }
    Display("%d", test);
</pre>

This will display "1", although it should "0".

gurok:
This should be fixed now. The stack was in an incorrect state in the middle of the switch block. No local variable assignment/reading would work because it's all relative to the stack pointer. I can't believe I missed that. Thanks, CW.

ivan-mogilko:
I found another issue. Any commands added on the same line after "case x:" are either ignored (this happened when I wrote "case 1: return;" or, more usually cause runtime crashes in the engine.

For example:
<pre>
int F()
{
    int test = 1;
    switch (test)
    {
        case 1: return 1;
    }
    return 2;
}
</pre>
Function F() will always return 2;

Another example:
<pre>
void F()
{
    int test = 1;
    switch (test)
    {
        case 1: Display("word");
    }
}
</pre>
Game stops at runtime with the following error:

> Error: Error running function 'game_start':
> Error: invalid instruction 174 found in code stream

Another example:
<pre>
int global_var; // this HAS to be global variable
void F()
{
    int test = 1;
    switch (test)
    {
        case 1: global_var = 2;
    }
}
</pre>
This causes Access violation exception at runtime (something related to incorrect opcodes, I am investigating this now).

gurok:
This is fixed now too. The switch table jumped to an incorrect point (2 off where it should have), but in most cases, this was harmless as it just skipped two bytes for the line number (SCMD_LINENUM). I thought I'd tested every permutation, but obviously, I made some assumptions. I have gone through and tested with even more examples. It *should* be right now.

Also, I saw your commit for the engine. That's good. Should be more friendly there.

PR #323, opened 2016-04-10
Scripting: Struct namespacing

gurok:
Continuing the work here:
https://github.com/adventuregamestudio/ags/pull/154

And addressing concerns here:
http://www.adventuregamestudio.co.uk/forums/index.php?issue=361.0

I see two other situations where global names might conflict with struct member names.
1. A struct member has the same name as a global variable, e.g. a GUI
2. A struct member has the same name as a type, e.g. `function MyStruct::Object()`

This is an attempt to tidy up the namespacing of struct members and allow for both of these cases. 

Allowing global variables was easy to achieve. Most namespacing is done at the tokenise step. We just needed to allow tokens of that type to be prefixed with the struct name.

Allowing type names had to happen later in the process (during parsing). It's not known at the point of tokenisation whether someone means `Object` as a return type or `Object` as a method name. Furthermore, it's not clear in some cases what the purpose of such a token is until after we've determined that it's an extender method.

Even though it's technically possible, for the purposes of language clarity, I've left language elements like `for` and `if` disallowed as member names. Similarly, primitive types like `int` and `float` are rejected.

These are the only remaining namespacing issues I could think of. If there are any more that come to mind, please mention them here and I'll amend this pull request to allow them.

Also, general disclaimer, I don't think this is best merged until after the 3.3.x->3.4.0 merge is complete.

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -682,6 +682,23 @@ int process_function_declaration(ccInternalList &targ, ccCompiledScript*scrip,
 		  targ.getnext();
 	  }
   }
+  else if(sym.get_type(funcsym) == SYM_VARTYPE) {
+      // Attempt to prefix a vartype with the struct name
+      // e.g. MyStruct::Object
+      if(isMemberFunction)
+      {
+          sprintf(functionName, "%s::%s", sym.get_name(isMemberFunction), sym.get_name(funcsym));
+          funcsym = sym.find(functionName);
+          if (funcsym < 0)
+              funcsym = sym.add(functionName);
+          *funcsymptr = funcsym;
+      }
+      if (sym.entries[funcsym].stype != 0) 
+      {
+          cc_error("function '%s' is already defined", functionName);
+          return -1;
+      }
+  }
 
   sym.entries[funcsym].stype = SYM_FUNCTION;
   sym.entries[funcsym].ssize = varsize;  // save return type size
@@ -3690,6 +3707,7 @@ int __cc_compile_file(const char*inpl,ccCompiledScript*scrip) {
                 do {
                     int vname = targ.getnext();
                     bool isDynamicArray = false;
+                    bool isFunction = sym.get_type(targ.peeknext()) == SYM_OPENPARENTHESIS;
                     if (sym.get_type(vname) == SYM_COMMA)
                         vname = targ.getnext();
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@gurok , I have an "unusual" kind of request, but could you rewrite the title and comments of those two commits?

For instance, one the title currently reads: "Allow global variables as struct members". What is that suppose to mean? How global variable can be member of the struct?
I do know the backstory, so I can correctly "decypher" this as "allow using global names for struct members", but other users of the code may get very confused.

Similarily, I believe that there is no need and sense mentioning forum user names and their particular problems. This reference may be clear to you, but not to any random person who would research the history of commits in our repository.
I suggest explaining the problem in general, e.g. "This allows to have struct members named identically to global symbols", or something like that.

gurok:
Yes, I see your point and I think it's very valid. I was trying to be terse because my Git client warns me if I go over a certain length. I also didn't really considering the meaning of what I wrote there as much as I should have. I will remove the reference to a specific forum user.

gurok:
Are these still awkward?

ivan-mogilko:
I guess these are okay, thank you.

gurok:
@ivan-mogilko Engine Debug failed here because of an execution timeout, but you can see in the logs that it was not hung and busily churning out warnings from compilation. Can we increase the execution timeout for the builds? I'm not sure why they run so slowly. Presumably shared processor time.

ivan-mogilko:
Build logs can be accessed by everyone, you may just login to build server as a guest.

@sonneveld controls the build server agents, it's up to him to fix this problem though.

gurok:
Hrmmm... help me understand what's going on. This is how I read it:
The build agent detected that it had been running for too long, but took > 2 min to kill the threads. During which time, the compilation process outputted one of its warnings. Indicates no stalling to me, just a timeout.

ivan-mogilko:
No, I guess I misunderstood what those times mean.

ivan-mogilko:
I finally had a chance to try this out.

I found following cases when using global names does not work:
* Local variables and function parameters: I guess AGS script simply does not support local variable names overriding global names.
* Extender functions cannot be named identically to global symbol; probably this is because they are treated as being in global scope themselves?
* Struct variable members *can* be named after global variable but cannot be named after user type, although member *function* can be, e.g.:
<pre>
int GlobalVar;

struct Test
{
  import Object* Object(); // this works
  
  int GlobalVar; // this works
  Character* Character; // this does not work
};
</pre>

gurok:
Extender functions and non-function struct members should both be doable.
I think local variables and function parameters are a bit out of scope. It's more of a structural limitation of the compiler (symbol redefinition) than an artificial one. I will have a think about it and see what can be done.
I have tagged this as WIP now.

ivan-mogilko:
I propose to keep this change as simple as possible for now, if you wish to have this in the 3.4.0 release, even though some of the limitations remain. I was planning to freeze 3.4.0 feature-wise very soon.

gurok:
Okay, fair enough. If you are happy with these for now, I'm happy.

On a side note, still working on the rewrite of switch/case stuff for the other pull request, but all that remains is testing.

ivan-mogilko:
How difficult it will be to fix the member variable issue? I think only this case breaks consistency in this pull request.

gurok:
It's probably the same or very similar to fixing member functions that share a name with a type. I will look into just this one thing then.

gurok:
Okay, the member variable thing is now fixed.

I looked into the extender function task. Just to clarify, extender functions can be named after existing variable types currently, but not existing global functions. This is because a new extender function needs a new import and when that import is created, the compiler uses the base name. I would like to fix it, but it's probably something that would require restructuring of the process_function_declaration function.

Local variables / parameters seems to be as we both covered before. Pretty involved too.

PR #329, opened 2016-05-15
Fixed build on bigendian systems like PowerPC

susnux:


rofl0r:
what was the error you were getting and why do you think its endian-related ?

susnux:
@rofl0r The error is:

    [ 143s] ../Common/util/mutifilelib.cpp:347:18: error: 'AGS::Common::BitByteOperations' has not been declared
    [ 143s] AGS::Common::BitByteOperations::SwapBytesInt32(val);

And BitByteOperations Swap is declared when endian is big, from bbop.h:

    #if defined(AGS_BIG_ENDIAN)
        AGS::Common::BitByteOperations::SwapBytesInt32(val);
    #endif

rofl0r:
looks legit. however imo the proper way to solve this would be to compile this code unconditionally so it's always tested on the developer machines:
```c
//instead of
#ifdef BIGENDIAN
x = SwapBytesInt32(x);
#endif
// at some header
#ifdef BIGENDIAN
#define BIGENDIAN 1
#else
#define BIGENDIAN 0
#endif
```
and then in the code:
if(BIGENDIAN) {
x = SwapBytesInt32(x);
}
```
if the swap is not needed the compiler will optimize the call away (due to if(0) -> unreachable), but it is always checked and will throw an error if the method is missing.

ivan-mogilko:
@rofl0r , I will look into reimplementing the bigendian code the way you suggested.

PR #330, opened 2016-05-18
Add method to fix MIDI playback

dakillaz98:
I am proposing this change as I have a Galaxy S5 running Cyanogenmod 13 (Android 6.0.1) and couldn't get AGS Android to play MIDI files despite having the patches.dat file downloaded from this link: http://alleg.sourceforge.net/digmid.html The game I tried was Yahtzee's Five Days a Stranger

I found out that the actual patches.dat is compressed, I opened the file with 7-Zip and found a patches[1] (I assume compressed name derives from container filename.)

So I extracted this patches[1] file, copied it to ags, deleted old patches.dat, renamed patches[1] to patches.dat. From there, I tried running Five Days a Stranger again, and sound worked perfectly. The original patches.dat file is 19MB, but after extracting, it grew to 25MB.

### DIFF
--- a/Android/README.md
+++ b/Android/README.md
@@ -63,6 +63,11 @@ http://www.eglebbk.dds.nl/program/download/digmid.dat
 Rename that file to **patches.dat** and place it into the **ags**
 directory alongside your games.
 
+*If MIDI music fails to play, attempt using an archive manager such as 7-Zip to extract the patches.dat file.
+    Extract the patches[1] file from patches.dat into the AGS directory
+    Remove patches.dat from the AGS directory
+    Rename the patches[1] file to patches.dat
+This method should fix MIDI playback due to compression on the raw files.
 
 
 #Building the engine


ivan-mogilko:
This is very interesting. I was not really tracking the whole issue with MIDI patches, but I recall we had several reports about them not working on mobile systems. I may suggest these people to try your solution and see if it works for them.

ivan-mogilko:
Alright, I did little research, and also got confirmation from the few more people that this solution works.

I cannot accept this pull request as it is though, IMHO the instructions should be worded a bit differently; also the formatting of the steps is incorrect. Besides, we would need to add this fix to several readmes, related to different ports (because they have same misinformation about MIDI in them).

But, regardless, I thank you for finding out a correct solution for the patches installation.

dakillaz98:
I am glad I could help, thank you for considering my process.                                                                                                                                                                        Sent from my BlackBerry - the most secure mobile device                                                                                                                                                                                                                                                          From:notifications@github.comSent:June 2, 2016 7:52 AMTo:ags@noreply.github.comReply-to:reply@reply.github.comCc:austenhuffman1@gmail.com; author@noreply.github.comSubject:Re: [adventuregamestudio/ags] Add method to fix MIDI playback (#330)  Alright, I did little research, and also got confirmation from the few more people that this solution works.

I cannot accept this pull request as it is though, IMHO the instructions should be worded a bit differently; also the formatting of the steps is incorrect. Besides, we would need to add this fix to several readmes, related to different ports (because they have same misinformation about MIDI in them).

But, regardless, I thank you for finding out a correct solution for the patches installation.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.

PR #332, opened 2016-06-06
New Editor splash screen

ivan-mogilko:
This is an implementation of a new Editor's splash screen (displayed both at the Editor launch and in the "About" dialog), based on the original design by abstauber (@dkrey).

**Historical note:** Because the source (non-flattened graphic document) of the old splash screen was lost long time ago, a community contest was held and the winner was chosen as a new splash design.
(Reference image: http://shatten.sonores.de/wp-content/uploads/2013/06/ags_splash3.png)

The splash is no longer implemented as a seamless image resource, the only binary resource used is a tiny PNG file, depicting a "Cup" logo. The rest is created and arranged programmatically.

Example as seen at runtime: http://i.imgur.com/14wZxrd.png


gurok:
Changes in your second commit have break spacing consistency. There are tabs where there should be spaces and vice versa.

The size of picCup should be 73 x 73. It's currently 73 x 75, which stretches the image ever so slightly and has the effect of blurring the image.

This is only a suggestion. I can see why you added the 1 pixel border:
https://github.com/ivan-mogilko/ags-refactoring/blob/814db7584954cf940013922940f7924e8a78580d/Editor/AGS.Editor/GUI/SplashScreen.cs#L84-88

But it does not match Metro-style splash screens:
http://www.howtogeek.com/wp-content/uploads/2012/08/image162.png

What about using the same colour as the background? #3b82d2

Also, if there's a 1-pixel gap, the SplashPage for the splash screen should be 618x298. That way the overall size is 620x300.

A couple of other suggestions:
 - If you give lblMainVersion padding of (4, 0, 0, 4), it will even out the spacing, but I think this might be font dependent :/
 - If you use #b1cded for the background colour of lblSeparator, it will more closely match ~1.5 pixel border of the mockup

ivan-mogilko:
> Changes in your second commit have break spacing consistency. There are tabs where there should be spaces and vice versa.

What is the spacing rule for Editor project (C# modules)? So far I've seen people use spaces when adding new code, but some of the old code has tabs (it was originally non consistent).

gurok:
I don't know. I use whatever surrounds the code I'm working on. I know there are some files with spaces in one part and tabs in the other. It is a mess and it annoys me too. I have always lined things up with the aim of making the commits look clean. Maybe we should just have spaces everywhere, going forward.

ivan-mogilko:
We had people making commits that bring multiple files to consistent identation before; for now I guess I will keep it same as surrounding code.

ivan-mogilko:
What was changed:
* fixed picCup size (actually, set SizeMode to AutoSize);

gurok:
I was going to suggest that you move the version label like that! I thought the border code looked a bit clunky -- the position of the SplashPage in the SplashScreenDesigner.cs didn't seem to matter because it would be set again later on. Yes, this makes more sense.

Looks good to me now.

PR #335, opened 2016-07-02
OSX Engine Port

sonneveld:
This pull requests provides a mostly working AGS engine for OSX. 90% of the work was done by Edward Rudd (@urkle). I'm just tidying it up.

Includes
- library build scripts
- allegro patch for OSX
- AGSKit shared framework and example ags app
- mac platform and engine changes.
- built in plugin registration for osx apps


ivan-mogilko:
I have a question: is it possible to have library sources and/or patches put into separate repository?

I recall we had some discussion a while ago, and it was suggested to make repository for Allegro or other libraries with necessary patches applied (if there is no way to get these patches into official library, or those patches have some AGS specific hacks).

Engine/plugin/agsplugin.cpp
```diff
@@ -922,6 +931,19 @@ bool pl_use_builtin_plugin(EnginePlugin* apl)
 #endif // IOS_VERSION
 #endif // BUILTIN_PLUGINS
 
+    for(std::vector<InbuiltPluginDetails>::iterator it = _registered_builtin_plugins.begin(); it != _registered_builtin_plugins.end(); ++it) {
+        if (strncmp(apl->filename, it->filename, strlen(it->filename)) == 0) {
```
ivan-mogilko: Why is the comparison limited to length of the name in '_registered_builtin_plugins'? This means that if game requires plugin "MyPlugin", and someone registered plugin "My", that plugin will be used. Is this really a wanted behavior?

Engine/plugin/agsplugin.h
```diff
@@ -582,4 +582,18 @@ void pl_run_plugin_init_gfx_hooks(const char *driverName, void *data);
...
+    char      filename[50];
+    void      (*engineStartup) (IAGSEngine *);
+    void      (*engineShutdown) ();
+    int       (*onEvent) (int, int);
+    void      (*initGfxHook) (const char *driverName, void *data);
+    int       (*debugHook) (const char * whichscript, int lineNumber, int reserved);
+};
+
+// Register a builtin plugin.
+int pl_register_builtin_plugin(InbuiltPluginDetails details);
```
ivan-mogilko: I think pl_register_builtin_plugin could take a (const) reference to InbuiltPluginDetails.

Engine/util/library_posix.h
```diff
@@ -66,6 +70,11 @@ class PosixLibrary : BaseLibrary
   {
     Unload();
 
+    // Try rpath first
+    _library = dlopen(BuildPath(NULL, libraryName).GetCStr(), RTLD_LAZY);
+    AGS::Common::Out::FPrint("dlopen returned: %s", dlerror());
+    if (_library != NULL) return true;
```
ivan-mogilko: Please, put statement on another line, below condition.

Engine/util/library_posix.h
```diff
@@ -48,8 +48,12 @@ class PosixLibrary : BaseLibrary
 
   AGS::Common::String BuildPath(char *path, AGS::Common::String libraryName)
   {
-    AGS::Common::String platformLibraryName = path;
-    platformLibraryName.Append("/lib");
+    AGS::Common::String platformLibraryName = "";
+    if (path) {
+      platformLibraryName = path;
```
ivan-mogilko: The code style for our project suggests to put curvy braces on separate lines (you may refer to how this is done in the rest of the file).

Engine/platform/osx/acplmac.cpp
```diff
@@ -113,6 +151,31 @@ void AGSMac::ShutdownCDPlayer() {
...
+}
+
+const char *AGSMac::GetAppOutputDirectory()
+{
+  return commonDataPath;
+}
+
+const char *AGSMac::GetIllegalFileChars()
+{
+  return "\\/:?\"<>|*"; // keep same as Windows so we can sync.
```
ivan-mogilko: I think we need elaboration on why this has to be synced with Windows.
I suspect there may be misunderstanding in how this function is used, because I do not see any immediate reasons to replicate Windows behavior here.

sonneveld:
In regards to the library. I agree. I've been putting it off because we have multiple patches all from the same point. There's windows, ios, osx, psp, android and possible sdl2 ports.

OSX/xcode/ags/plugin_registration.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "agsplugin.h"
+//#include "example.h"
+
+typedef void (*t_engine_pre_init_callback)(void);
+extern void engine_set_pre_init_callback(t_engine_pre_init_callback callback);
+
+static void pl_register_builtin_plugin() {
+
+//  pl_register_builtin_plugin((InbuiltPluginDetails){
+//    .filename = "example",
```
ivan-mogilko: This may be something specific to xcode (or objective-c maybe?) which I do not know about, but from C++ perspective this example is incorrect, because "filename" is declared as an array of chars, not a string object.

Engine/plugin/agsplugin.cpp
sonneveld: I was copying the behaviour of pl_use_builtin_plugin in agsplugin.cpp.  I assumed it was so we could ignore filename extensions.

Engine/plugin/agsplugin.h
sonneveld: It could! I didn't feel that one off registration required particular optimisation but I could change it to a reference.

Engine/util/library_posix.h
sonneveld: Will do.

.gitignore
```diff
@@ -68,3 +67,36 @@ iOS/nativelibs/arm64
 iOS/nativelibs/i386
 iOS/nativelibs/x86_64
 Solutions/Engine.App/test.ini
+
+libsrc/*.tar*
+libsrc/*.zip
+
+ac2game.dat
+audio.vox
+speech.vox
```
ivan-mogilko: Ok, this may be a dumb question, but why those files got ignored? Are they produced by some process?

Engine/util/library_posix.h
sonneveld: Is that true though? Apart from this file, a large amount of code seems to have braces on the same line. I'm more partial to only having braces on new line for function definitions and same line for everything else.  The code feels a bit spread out otherwise.

Engine/platform/osx/acplmac.cpp
sonneveld: Ah, the reasoning here was to ensure that Steam (or even just a user manually copying) could sync saves between operating systems.  It's no good writing a file in Windows to a path that can't be read in OSX.

.gitignore
sonneveld: libsrc/ files are downloaded as part of building native libraries for OSX, iOS and Android. 

The game files are sometimes in the project for testing (as they get added to app bundles)

Engine/util/library_posix.h
ivan-mogilko: AGS has a very old code, which has several styles used by its author (and contributors).
Our coding convention: http://www.adventuregamestudio.co.uk/wiki/AGS_Engine_Coding_Conventions

If you check most of my additions for the last few years tried to correspond to that convention, except for cases when doing small fixes having different style in surrounding code.

OSX/xcode/ags/plugin_registration.cpp
sonneveld: This is actually standard C behaviour.  You can initialise a char array with a string literal.  http://stackoverflow.com/questions/8244453/initializing-char-array-with-smaller-string-literal

Engine/util/library_posix.h
sonneveld: No worries. I will mimic the current file brace style. I wouldn't mind amending the AGS coding conventions in the future though. :)

OSX/xcode/ags/plugin_registration.cpp
ivan-mogilko: Oh, my mistake, I guess I did not use this kind of initialization for a long time.

Engine/platform/osx/acplmac.cpp
ivan-mogilko: Hmm, I guess I am still missing something.
Can you give a practical example of what and how the user will copy?

Engine/device/mousew32.cpp
```diff
@@ -306,7 +309,16 @@ int mgetbutton()
     return NONE;  // don't allow holding button down
 
   if (butis & 1)
+  {
     toret = LEFT;
+#if defined(MAC_VERSION)
+    // j Ctrl-left click should be right-click
+    if (IsKeyPressed(405) || IsKeyPressed(406))
```
ivan-mogilko: I think I understand why this is needed, but I believe we need to write TODO about making this configurable later.
Some games may script Ctrl+LMB for particular in-game action.

OSX/xcode/ags/plugin_registration.cpp
ivan-mogilko: Um, Nick, you really tricked me here (so I deleted my previous answer).

The example you referenced is about **initializing** char[] variable, which is ofcourse valid.
But in this case you suggest assigning new value to char[] variable. In C++ that won't work.

Engine/gfx/ali3dsw.cpp
```diff
@@ -376,6 +376,10 @@ bool ALSoftwareGraphicsDriver::Init(int virtualWidth, int virtualHeight, int rea
 
     virtualScreen = BitmapHelper::GetScreenBitmap();
 
+    // [ER] 2014-03-13
+    // Hide the system cursor via allegro
+    show_os_cursor(MOUSE_CURSOR_NONE);
```
ivan-mogilko: I need to clarify, is it elaborate decision to put this into ALSoftwareGraphicsDriver initialization?
I mean, is this specific to using this renderer?

Engine/platform/osx/acplmac.cpp
ivan-mogilko: To elaborate my confusion: why does it matter for, for example, OSX port which paths are used by Windows version?

Engine/plugin/agsplugin.cpp
ivan-mogilko: Oh, right, I noticed this now...
Actually, this means the previously existing code is also not correct.

sonneveld: Engine now splits string at extension and changes buffer length. Fixed in b1c31123061079d454748ea82dd4103d8c6d9fff

Engine/plugin/agsplugin.h
sonneveld: Changed to const ref in b1c31123061079d454748ea82dd4103d8c6d9fff

Engine/util/library_posix.h
sonneveld: Fixed in 408757b822ec5be9a14e89229add71c838baa3c5

Engine/util/library_posix.h
sonneveld: Fixed in 408757b822ec5be9a14e89229add71c838baa3c5

Engine/platform/osx/acplmac.cpp
sonneveld: If a user saves a file with the characters "\\/:?\"<>|*" on OSX or Linux, Steam will not be able to copy that file over to the Windows machine when syncing save games.

OSX/xcode/ags/plugin_registration.cpp
sonneveld: Ah but in this case I am initialising the struct itself!

.gitignore
ivan-mogilko: Oh, they still use "ac2game.dat" as game file name?

I recall we had a big misunderstanding in the past, when Jochen (@jjsat) just made his ports, he used this as default file name for ags games. Later I had to do a research to find out the origins of this name, and I learnt that this is a very old convention. I think he made a mistake by choosing it, because it is no longer used anywhere in modern code, nor as a compilation output.

Current convention is to call game file as "GameName.ags". The 3.4.0 Editor also produces this file when configured to deploy game for Linux.

Engine/gfx/ali3dsw.cpp
sonneveld: Edward Rudd's specific commit comment about this was "hide the stupid mouse cursor all the time on ags graphics startup". :)  However it does hide the mouse earlier while windows in OSX and are animating to fullscreen. (there may be a reason for it in Linux too)

Engine/platform/osx/acplmac.cpp
ivan-mogilko: File names for AGS saves are never edited by user, they are always agssave.XXX, where XXX is some number.

This function is currently used only in 1 place in the code, to fix the name of the save folder, because it may be edited in project settings.

.gitignore
sonneveld: Ahh, that makes sense.  And correct, Wadjet Eye still use "ac2game.dat" for their current games.  They might just be following the convention set by @jjsat though.

Engine/platform/osx/acplmac.cpp
ivan-mogilko: On other hand, you addressed an interesting problem... since AGS can create custom files by  script command.

sonneveld: You are correct about the savegame names being fixed. I was concerned about other files created by scripts (and the game name itself being used as a folder name)

Engine/device/mousew32.cpp
sonneveld: Added task AGS-11

OSX/xcode/ags/plugin_registration.cpp
ivan-mogilko: Ohhhhh.... this is initializing struct in the function parameter list.

For some reason I misread it as interns of a function where you set up struct members.

Ok, nevermind then :-).

sonneveld:
I've added AGS-12 for tracking the library repositories.

Engine/platform/osx/acplmac.cpp
ivan-mogilko: Alright, I understand now.

I *think* we might need two functions in the platform class then: one for fixing portable file names (such as files that are created and shared by game on all systems) and one for fixing names that are not supposed to be portable (such as temporary files, and platform-specific locations).

ivan-mogilko: No, actually, since the rule for portable names is supposed to be always same, then such function should not be overriden in a platform class at all!

sonneveld: Would you like to do that in this pull request or create a task? I'm not entirely familiar with the filename code so I would need to spend some time on that.

.gitignore
```diff
@@ -33,7 +33,6 @@ Engine/resource/version.aps
 *.o
 *.d
 common.a
-ags
```
ivan-mogilko: We need to have Engine/ags in ignore list, because it is produced by making Makefile directly (at least on Linux).

Engine/platform/osx/acplmac.cpp
sonneveld: Ok, if we need to pull it out of the platform class, I believe that should be a separate pull request. It would require me to modify multiple platforms and test them.  Created task AGS-13 to track it.

.gitignore
sonneveld: Ah! It was hiding all changes in OSX/xcode/ags and driving me mad.  I readded and made more specific in b3e0f7e0573340322e68178373fbcb3d855920c1

Engine/platform/osx/acplmac.cpp
ivan-mogilko: I think we may leave the function that already exists, just in case we would need it for fixing non-portable names (according to my idea I explained above).
BTW, if I remember correctly, this function was overridden only in Windows platform so far. Other platforms used default implementation, that likely is linux-compatible.

The portable file name cases deserve a separate discussion. For starters we would need to define where we do have to apply this syncronizing rule and where not.

Engine/gfx/ali3dsw.cpp
ivan-mogilko: I mean, if this is a universal problem, perhaps there could be a better place to put this call, instead of gfx driver's Init method.

Maybe somewhere above the line where "graphics_mode_init" is called. But I cannot tell for certain, because I do not see how this works at runtime.

sonneveld:
Ok, I think I've addressed all the comments.. Is there anything else you'd like addressed before merging?

rofl0r:
minor nitpicking: i noticed that the patches are not cleaned up, they contain stuff like:
```
only in foo/ags: foo.o
```

Engine/gfx/ali3dsw.cpp
sonneveld: Moved it to a more relevant location in 454ecdd50076ffd5984bf7f33c95c2bf1d8d0723

sonneveld:
Ok, I think that's it!

Engine/plugin/agsplugin.cpp
```diff
@@ -977,11 +977,26 @@ void pl_read_plugins_from_disk (Stream *in) {
...
+        {
+            char *ext = strchr(buffer, '.');
+            if (ext) {
+                *ext = 0;
+            }
+        }
+        
+        if (strlen(buffer) > PLUGIN_FILENAME_MAX) {
+            sprintf(buffer, "Plugin '%s' is not a valid AGS plugin because the filename is invalid.", buffer);
+            quit(buffer);
```
ivan-mogilko: Just a note for the future, there is a "quitprintf" function for sending formatted quit message.

ivan-mogilko: Actually, I did not pay attention at first, but is not it that when you use output buffer as an input argument, the results are undefined by standard?
I would really suggest to replace this with quitprintf just to be safe.

ivan-mogilko: Yeah, with that sprintf MS runtime produces following message:
<pre>Plugin 'Plugin 'Plugin 'Plugin 'Plugin 'Plugin 'Plugin 'Plugin 'Plugin ' is not a valid AGS plugin because the filename is invalid.</pre>

ivan-mogilko:
RE efa24b2e879df6c9842fe29c9fb85d4a2a72f5da
**bealleg** is not a Windows header... it is BeOS header, probably ended up in the sources by mistake, and should be de removed.

sonneveld:
RE efa24b2 / bealleg
yeah, I know, but there's a bunch of other allegro os files in the same directory.  I suggest we worry about deleting files when have proper thirdparty library management

Engine/plugin/agsplugin.cpp
sonneveld: Ok that's kind of hilarious.  I'll fix that.

ivan-mogilko:
I do not understand your reasoning! Instead of deleting the file that is obviously there by mistake, you move it to the subdirectory of the very particular purpose, that does not even match this file?

Also, I do not see any other "os" files in the Common/libinclude. Are we speaking of same thing here?

ivan-mogilko:
Although, I honestly do not know why these files are kept there... this could be an oversight made at the times of restructuring old SVN repository contents.

sonneveld:
Windows/include has all the standard allegro headers.  It is a direct copy of allegro's include directory. It includes headers for osx, linux, qnx, etc, which are also not windows related. I figured I would keep the original layout instead of deleting files

ivan-mogilko:
Well, the original layout was broken long time ago, and, as I said, these files were not even required. I honestly do not see the point of restoring it, because it is useless.

ivan-mogilko:
I'd really wish the commit description explained all that, because when I saw this I was baffled about what's going on.

Engine/plugin/agsplugin.cpp
sonneveld: Fixed in b416474

ivan-mogilko:
I guess it is okay now.

sonneveld:
I can do that if you like. However, I don't think it's too bad here as these changesets represent a clear progression of work. It would be different if I was just changing random things to try and fix something.

ivan-mogilko:
No, I do not refer to progression of work, I rather meant the kind of commits you described as -
>  It would be different if I was just changing random things to try and fix something.

But I may be nitpicky about that sometimes...

PR #337, opened 2016-07-17
Xcode build

sonneveld:
This pull request fixes builds for ios and osx. It also attempts to wrap the tr1 namespace.


sonneveld:
I haven't tested on windows or linux yet, I'm hoping the build server will help me out there.

ivan-mogilko:
This thing with stdtr1compat namespace is something I was considering too, it should be a good thing to put to use.

I was wondering if we could move those repeating declarations in a special header. So far it were just 2 or 3 places where they were required, but that may change soon.
What I mean, for example, make something like "Common/utils/unordered_map.h" and put this ifdef/else construct there.

sonneveld:
Actually, it turns out the tr1compat wrapper is similar to what boost.tr1 provides already (only better implemented).  How would you feel if we just included boost as a dependency?  It uses the system libraries if available, otherwise it uses boost implementation.

sonneveld:
Rendering on OSX is a little weird, for some reason we get a really large window with the background surface rendered in a tiny box in the middle.  I'm happy to leave this pull request as just fixing xcode builds.  Fixing rendering could be done in another pull request.

ivan-mogilko:
We could move away from vc2008, but I think it was discussed before that it may be not quite the time to move away from C++03.

ivan-mogilko:
Regarding rendering, display mode configuration has changed in 3.4.0, and I do not know whether OpenGL renderer is catching up to that.

PR #339, opened 2016-07-20
Installer: support for optional extra version tag

ivan-mogilko:
This adds an optional version tag which is appended to the package/installer file names (as well as app name the Editor is registered on Windows with when installed). This tag can be used, for example, to specify beta version, or patch number to make it easier for end users to distinguish updates of same public release.

### DIFF
--- a/Script/build_packages.py
+++ b/Script/build_packages.py
@@ -29,7 +29,7 @@ def workspace_rel(path):
     return os.path.join(workspace_dir(), path)
 
 
-AgsVersion = namedtuple('AgsVersion', ['version', 'version_friendly', 'app_id'])
+AgsVersion = namedtuple('AgsVersion', ['version', 'version_friendly', 'version_sp', 'app_id'])
 
 def load_version(path):
     with open(path, "r") as f:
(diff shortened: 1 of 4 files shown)


Script/build_packages.py
```diff
@@ -75,7 +77,7 @@ def refresh_staging():
     return staging_path
 
 def extract_into(destdir, zippath):
-    with zipfile.ZipFile(zippath, "r") as z:
+    with contextlib.closing(zipfile.ZipFile(zippath, "r")) as z:
```
sonneveld: contextlib.closing on zipfiles shouldn't be necessary for recent versions of python. (2.7 and above)

ivan-mogilko: Well, this is why the commit is called "python 2.6 compatibility".

ivan-mogilko: Alright, I will remove this commit.

sonneveld: oh right! I was only looking at the diff. What systems were you hoping to support that were stuck with python 2.6?

ivan-mogilko:
I made one mistake when reorganizing pull branch, which resulted in incorrect script line. Now it should be fixed.

Script/build_packages.py
ivan-mogilko: Nevermind, I was just getting annoyed with older version of python interfering when I run the script without specifying correct interpreter version.

PR #344, opened 2016-08-03
Refactor: picked out and cleaned up code surrounding reading and writing of game save

ivan-mogilko:
This picks out data processing that precedes and follows writing and restoring game data of the game's save. General purpose is to separate complemetary tasks from reading and writing actual data. Currently they are spread all around reading/writing code.
These tasks include:
* Preparing game for writing;
* Opening savegame file and checking its header information (format, description);
* Preparing game for reading (releasing particular runtime objects mostly);
* Preparing game for running again after reading save (recreating runtime objects with new data, loading new room, and so on).

Additionally introduced savegame error enumeration and made reading function return those error codes instead of quitting the game.
Merged several different pieces of code into common function that deduces what to do if restoration failed, depending on how far engine went (display on-screen message or quit).

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -15,6 +15,7 @@
 #include <errno.h>
 #include <stdlib.h>
 #include "gui/guidefines.h"
+#include "util/math.h"
 #include "util/string_utils.h"
 #include "util/stream.h"
 
@@ -201,6 +202,11 @@ void StrUtil::WriteString(const String &s, Stream *out)
     out->WriteInt32(len);
     if (len > 0)
         out->Write(s.GetCStr(), len);
+}
+
+void StrUtil::WriteCStr(const String &s, Stream *out)
+{
+    out->Write(s.GetCStr(), s.GetLength() + 1);
 }
 
 } // namespace Common
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -64,6 +64,7 @@
 #include "device/mousew32.h"
 #include "font/fonts.h"
 #include "game/savegame.h"
+#include "game/savegame_internal.h"
 #include "gui/animatingguibutton.h"
 #include "gfx/graphicsdriver.h"
 #include "gfx/gfxfilter.h"
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1085,8 +1085,6 @@ long write_screen_shot_for_vista(Stream *out, Bitmap *screenshot)
 
 void save_game_head_dynamic_values(Stream *out)
 {
-    out->WriteInt32(play.viewport.GetHeight());
-    out->WriteInt32(ScreenResolution.ColorDepth);
     out->WriteInt32(frames_per_second);
     out->WriteInt32(cur_mode);
     out->WriteInt32(cur_cursor);
@@ -1529,14 +1527,6 @@ char rbuffer[200];
 
 SavegameError restore_game_head_dynamic_values(Stream *in, RestoredData &r_data)
 {
-    in->ReadInt32(); // gamescrnhit, was used to check display resolution
-
-    // CHECKME: is this still essential? if yes, is there possible workaround?
-    if (in->ReadInt32() != ScreenResolution.ColorDepth) {
-        Display("This game was saved with the engine running at a different colour depth. It cannot be restored.");
-        return kSvgErr_DifferentColorDepth;
-    }
-
     r_data.FPS = in->ReadInt32();
     r_data.CursorMode = in->ReadInt32();
     r_data.CursorID = in->ReadInt32();
@@ -2098,12 +2088,16 @@ SavegameError load_game(const String &path, int slotNumber, bool &data_overwritt
 
     our_eip = 2051;
 
-    // saved in different game
+    // saved in incompatible enviroment
     if (err != kSvgErr_NoError)
         return err;
+    // CHECKME: is this color depth test still essential? if yes, is there possible workaround?
+    else if (desc.ColorDepth != ScreenResolution.ColorDepth)
+        return kSvgErr_DifferentColorDepth;
     else if (!src.InputStream.get())
         return kSvgErr_NoStream;
 
+    // saved with different game file
     if (Path::ComparePaths(desc.MainDataFilename, usetup.main_data_filename))
     {
         // [IKM] 2012-11-26: this is a workaround, indeed.
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -453,7 +453,7 @@ void restore_game_dialog() {
     int toload=loadgamedialog();
     restore_after_dialog();
     if (toload>=0) {
-        load_game_and_print_error(toload);
+        try_restore_save(toload);
     }
 }
 
@@ -2096,8 +2096,9 @@ bool read_savedgame_screenshot(const String &savedgame, int &want_shot)
     return true;
 }
 
-SavegameError load_game(const String &path, int slotNumber)
+SavegameError load_game(const String &path, int slotNumber, bool &data_overwritten)
 {
+    data_overwritten = false;
     gameHasBeenRestored++;
 
     oldeip = our_eip;
@@ -2134,6 +2135,7 @@ SavegameError load_game(const String &path, int slotNumber)
 
     // do the actual restore
     err = RestoreGameState(src.InputStream.get(), src.Version);
+    data_overwritten = true;
     if (err != kSvgErr_NoError)
         return err;
     src.InputStream.reset();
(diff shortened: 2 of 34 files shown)


Engine/game/savegame.h
```diff
@@ -0,0 +1,202 @@
...
+//
+//=============================================================================
+
+#ifndef __AGS_EE_GAME__SAVEGAME_H
+#define __AGS_EE_GAME__SAVEGAME_H
+
+#if defined (_MSC_VER)
+#include <memory>
+#elif defined (__GNUC__)
+#include <tr1/memory>
```
rofl0r: isn't this pulling in a requirement of stdc++11 support ?

ivan-mogilko: no, it is std::tr1 namespace, which we are using for a about year in the new version. Should compile with latest gcc 4.

rofl0r: ok, thanks for clarification.

gurok:
I understand that this has no milestone, but are you releasing 3.4.0 before or after this gets merged?

ivan-mogilko:
3.4.0 is secured in its corresponding branch now, release-3.4.0, so it is safe to put completely new stuff to master now. This is where our development continues from now on.

As for 3.4.0, any hotfixes should go to release-3.4.0. I will add final release tag at that branch too when it's time.

gurok:
Ah, okay, sorry. I missed the fact that a branch had been made.

ivan-mogilko:
Hmm, I maybe need to find a way to announce such things more explicitly on github.

Common/util/string_utils.cpp
```diff
@@ -203,5 +204,12 @@ void StrUtil::WriteString(const String &s, Stream *out)
         out->Write(s.GetCStr(), len);
 }
 
+void StrUtil::WriteCStr(const String &s, Stream *out, int max_len)
+{
+    max_len = Math::Min(max_len, s.GetLength());
```
sonneveld: Instead of silently truncating, would it be better to raise an error if it doesn't fix within max_len?

Engine/ac/game.cpp
```diff
@@ -2443,177 +2323,134 @@ int restore_game_data (Stream *in, const char *nametouse, SavedGameVersion svg_v
         cachedQueuedMusic = load_music_from_disk(play.music_queue[0], 0);
     }
 
-    return 0;
+    return kSvgErr_NoError;
 }
 
-int restore_game_data (Common::Stream *in, const char *nametouse)
+bool restore_game_data(Common::Stream *in, const char *nametouse)
```
sonneveld: Are we converting all methods from errorcodes to bools?  Bools don't really let you have more than one error state :)

Common/util/string_utils.cpp
ivan-mogilko: I can make it return actually written length probably?

This function is meant to forcefully truncate string, it emulates how AGS was writing limited fields originally, so it is not like an erroneous situation.

ivan-mogilko: Well, I just found a bug in this function... parameter's default value -1 will make nothing to be written, instead of whole string, as was probably planned.

Engine/ac/game.cpp
ivan-mogilko: Ok, I will change it to errcode.

ivan-mogilko: Ok, I will change to errcode.

Common/util/string_utils.cpp
sonneveld: hah, didn't even notice that.  Clearly more tests needed :)  Returning the written size (ignoring null char so you can compare with strlen) would be a good start I think.

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+
+
+extern GameSetupStruct game;
+extern Bitmap **guibg;
+extern AGS::Engine::IDriverDependantBitmap **guibgbmp;
+extern AGS::Engine::IGraphicsDriver *gfxDriver;
+extern Bitmap *dynamicallyCreatedSurfaces[MAX_DYNAMIC_SURFACES];
+extern Bitmap *raw_saved_screen;
+extern RoomStatus troom;
+extern RoomStatus *croom;
```
sonneveld: Why do we still use externs for globals?

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+    Mouse::SetMoveLimit(Rect(oldx1, oldy1, oldx2, oldy2));
+
+    set_cursor_mode(r_data.CursorMode);
+    set_mouse_cursor(r_data.CursorID);
+    if (r_data.CursorMode == MODE_USE)
+        SetActiveInventory(playerchar->activeinv);
+    // ensure that the current cursor is locked
+    spriteset.precache(game.mcurs[r_data.CursorID].pic);
+
+#if (ALLEGRO_DATE > 19990103)
```
sonneveld: Are we still supporting allegro versions from before 1999? :)

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+        }
+    }
+    if ((cf_in_chan > 0) && (channels[cf_in_chan] != NULL))
+        play.crossfading_in_channel = cf_in_chan;
+    if ((cf_out_chan > 0) && (channels[cf_out_chan] != NULL))
+        play.crossfading_out_channel = cf_out_chan;
+
+    // If there were synced audio tracks, the time taken to load in the
+    // different channels will have thrown them out of sync, so re-time it
+    for (int i = 0; i <= MAX_SOUND_CHANNELS; ++i)
```
sonneveld: did you mean to iterate from 0 .. MAX_SOUND_CHANNELS?

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+    {
+        int pos = r_data.AudioChans[i].Pos;
+        if ((pos > 0) && (channels[i] != NULL) && (channels[i]->done == 0))
+        {
+            channels[i]->seek(pos);
+        }
+    }
+
+    for (int i = 1; i < MAX_SOUND_CHANNELS; ++i)
```
sonneveld: i.e, this is now 1 .. MAX_SOUND_CHANNELS

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+    troom.FreeProperties();
+    free_do_once_tokens();
+
+    // unregister gui controls from API exports
+    for (int i = 0; i < game.numgui; ++i)
+    {
+        unexport_gui_controls(i);
+    }
+
+    for (int i = 0; i <= MAX_SOUND_CHANNELS; ++i)
```
sonneveld: Another instance of 0 .. MAX_SOUND_CHANNELS

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+    if (play.digital_master_volume >= 0)
+        System_SetVolume(play.digital_master_volume);
+
+    // Run audio clips on channels
+    // these two crossfading parameters have to be temporarily reset
+    const int cf_in_chan = play.crossfading_in_channel;
+    const int cf_out_chan = play.crossfading_out_channel;
+    play.crossfading_in_channel = 0;
+    play.crossfading_out_channel = 0;
+    for (int i = 0; i <= MAX_SOUND_CHANNELS; ++i)
```
sonneveld: 0 .. MAX_SOUND_CHANNELS

Engine/game/savegame.cpp
```diff
@@ -0,0 +1,596 @@
...
+
+    unload_old_room();
+    delete raw_saved_screen;
+    raw_saved_screen = NULL;
+    remove_screen_overlay(-1);
+    is_complete_overlay = 0;
+    is_text_overlay = 0;
+
+    // cleanup dynamic sprites
+    for (int i = 1; i < spriteset.elements; ++i)
```
sonneveld: Ok, last for loop to comment on :)  is sprite 0 special case? I can't remember.

sonneveld:
Reviewed. Looks good. Mostly straightforward extraction I think?

Engine/game/savegame.cpp
ivan-mogilko: I did not mean anything, I plain moved this code from another place.

Engine/game/savegame.cpp
ivan-mogilko: Okay, I will move these into some headers.

Engine/game/savegame.cpp
ivan-mogilko: If I remember correctly sprite 0 was reserved to keep "invalid sprite" graphic. I do not remember how it is used though.

Engine/game/savegame.cpp
sonneveld: Actually, could these globals be passed in as references to the save functions?  It would make it more explicit that we're changing global state. (maybe a different pull request)

It's messy but I like that we're being explicit at the top which globals we're using.

Engine/game/savegame.cpp
ivan-mogilko: These are iterations over different arrays. They may have different sizes too. I cannot tell right away why it is done so.

sonneveld: Would it be worth adding a couple of comments as a warning for the next guy? Maybe it's fine.

Engine/game/savegame.cpp
ivan-mogilko: Loigcally they have to be gathered in a class, or few classes referencing each other. If I start to pass references to every object I use in savegame function now, I will end up with dozens of references.

sonneveld: hrm you're right.  Perhaps session classes for different lifetimes, an application session, a per game session, etc.  Sorry, not suggesting you do that in this story. :)

Engine/game/savegame.cpp
ivan-mogilko: I would, but what do I write in these comments?
This code needs investigation, because I do not remember (and possibly never knew) why does it look this way.
I honestly cannot even say if that is not a bug. I can only check that it was originally same way.

sonneveld: How about "TODO: beware this strange loop range" ?  Or do we just track it  as a potential bug in youtrack?

ivan-mogilko: Ok, a note to investigate later may be a good idea.

ivan-mogilko:
This needs additional fixing after String's pull request was merged.

Engine/ac/game.cpp
ivan-mogilko: This function is supposed to be called only in replay loading code that is factually not used in the engine write now. That code is not interested in precise error. Also, I do not want to expose savegame error codes to other modules, because they don't need to know it. Making this properly will require reorganizing code more, which is not worthy for this issue.
So I will leave this function having just a bool; probably rename it to indicate that it is used just for very specific case.

ivan-mogilko: Ok, nevermind, I found a better way.

ivan-mogilko:
Rewrote branch.
Following @sonneveld 's comment https://github.com/adventuregamestudio/ags/pull/344#discussion_r73779708 I finally decided to refactor more, and hide couple of structs and functions that are only used in a proper sequence.

Regarding WriteCStr, I investigated bit more and realized I did a mistaske by truncating the saved field at the time of writing, while originally it was supposed to be truncated at the time of *reading*, simply because the input buffer was limited. Now when this code uses dynamic strings, this is no longer required, so I changed that and removed constants defining lengths of those fields.

ivan-mogilko:
Had to fix one oversight, plus added little more refactoring.

Tested saving and loading games with this branch, restoring saves made by older engine and restoring saves made by this branch in older engine. Now this pull request is done, I think.

PR #345, opened 2016-08-04
Changed String class to use size_t for lengths and indexes instead of int

ivan-mogilko:
Initially I wrote String class using ints for buffer length and indexes for some reasons, now changing to size_t.
This is long overdue, but something distracted me all the time from committing.

NOTE: We currently have string tests in the engine, which are only built in Debug configuration, and run at program startup. I was referencing to them to check that this conversion is correct.

### DIFF
--- a/Common/util/mutifilelib.cpp
+++ b/Common/util/mutifilelib.cpp
@@ -112,14 +112,14 @@ MFLUtil::MFLError MFLUtil::ReadSigsAndVersion(Stream *in, int *p_lib_version, lo
     if (HeadSig.Compare(sig) != 0)
     {
         // signature not found, check signature at the end of file
-        in->Seek(-TailSig.GetLength(), kSeekEnd);
+        in->Seek(-(int)TailSig.GetLength(), kSeekEnd);
         sig.ReadCount(in, TailSig.GetLength());
         // signature not found, return error code
         if (TailSig.Compare(sig) != 0)
             return kMFLErrNoLibSig;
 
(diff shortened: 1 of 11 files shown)


Common/util/mutifilelib.cpp
```diff
@@ -112,14 +112,14 @@ MFLUtil::MFLError MFLUtil::ReadSigsAndVersion(Stream *in, int *p_lib_version, lo
     if (HeadSig.Compare(sig) != 0)
     {
         // signature not found, check signature at the end of file
-        in->Seek(-TailSig.GetLength(), kSeekEnd);
+        in->Seek(-(int)TailSig.GetLength(), kSeekEnd);
```
sonneveld: I'm worried that just casting to int without checking if it fits will cause some interesting issues for very large files.

ivan-mogilko: TailSig is a very short constant string (around 10 characters), if it is what you are refering to?

sonneveld:
Reviewed. In general looks fine. In the future, would it be worth splitting up the string class into different formatting modules?  Have we considered just using a normal c++ string as backing?

ivan-mogilko:
Do you mean replacing this class with std::string completely, or making this class a reference counted wrapping over std::string?

Common/util/mutifilelib.cpp
sonneveld: Ah I didn't look at definition of TailSig.. I guess as long as we know it's always around 10.

PR #346, opened 2016-08-06
Windows: strip engine and editor of precompiled libraries

ivan-mogilko:
Windows version of AGS historically used precompiled libraries built by original author; some of these libraries were actually _modified_, either to achieve backwards-compatibility with older games or other reasons.

This changeset removes those binaries and provides patches instead to let developers build Windows version fully from source.
The content of patches, when not available, was either reverse-engineered or otherwise deduced. Notably, _alfont_ patch is based on reverse-engineering by @jjsat (we also use that work for other ports).

Windows readme is updated with explanation on building with library sources.

I need other people check this out to make sure everyone can actually build working engine and editor using these instructions and patches.

**NOTE**: of course, the build server won't be able to build this version - at all - until reconfigured.


monkey0506:
The Allegro patch seems to rename "pack_fopen" to "__old_pack_fopen" but doesn't replace "pack_fopen", so I am getting an unresolved symbol with the patch. Renaming that back to "pack_fopen" allows it to build, but I don't understand why it was renamed if it wasn't supposed to be replaced.

sonneveld:
pack_fopen is currently implemented in Engine/ac/file.cpp .

monkey0506:
I think the Allegro MSVC solution file is set up to build both static and dynamic linking. I'll double check that I'm building the correct one and make sure I can get everything else to build. Subsequent problems stemmed from this. Thanks for clarifying.

monkey0506:
I made sure that all of the _static_ libs were built with /MT and the _dynamic_ libs were built with /MD. The engine now builds, but the AGS.Native project is failing to link with unresolved external "_pack_fopen". From what I can tell, AGS.Native doesn't even reference pack_fopen, but it also isn't defined anywhere in that compilation unit. What am I missing?

monkey0506:
I got the engine to build, but I'm getting an unresolved symbol for pack_fopen ("_pack_fopen" to be precise) when building AGS.Native. From what I can see, Common.Lib, Compiler.Lib, and AGS.Native do not provide an implementation of pack_fopen, though it seems to be needed at some point in that compilation (not directly referenced, but failing to link without it). What am I missing here?

ivan-mogilko:
I will be looking into this shortly.

ivan-mogilko:
Aha, I overlooked the fact that `pack_fopen` was overridden only for the Engine; apparently the prebuilt libraries meant for AGS.Native did not have this change.

Having two versions of Allegro for building both Engine & Editor seems too much, so I guess the better solution would be to rename all `pack_fopen` calls in AGS.Native to `old_pack_fopen`. I will test this out and make another commit if that works.

monkey0506:
I was pretty sure that the linker flags weren't the problem, but I was just making absolutely certain that I was following the instructions properly. Thanks for looking into that, it was driving me crazy. I look forward to being able to compile everything.

ivan-mogilko:
For information, in relation to pack_fopen implementation, I found this patch suggested to Allegro team by Bernhard Rosenkraenzer (I think he was author of the first Linux port).
https://sourceforge.net/p/alleg/feature-requests/42/

This link is mentioned in comments to AGS code.

ivan-mogilko:
@monkey0506 , I pushed a fix, please test now.

monkey0506:
That got everything to build, and basic testing seems to show everything working as usual. Thanks!

rofl0r:
/me wonders why the pack_fopen hack is needed at all. most likely there's a way to achieve the same effect without symbol name hacks (i don't know the background though).

ivan-mogilko:
> /me wonders why the pack_fopen hack is needed at all. most likely there's a way to achieve the same effect without symbol name hacks

We are going to find this out; but I'd prefer to keep things as close as possible to original for this pull request.

ivan-mogilko:
> That got everything to build, and basic testing seems to show everything working as usual.

Since we got this confirmation, I guess the pull is working; but we need to resolve build server issue somehow before merging this. Will it build one particular branch (master) using new rules? Need to get clarification from @sonneveld .

sonneveld:
Ideally there would be a script that downloads the library source, patches, builds and makes available to the ags project. That script could be called in a single step in teamcity jobs.  I would rather not manually patch and copy to sources to the build server as that would be difficult to maintain and doesn't track newer versions.

monkey0506:
Aside from the build server, is there any plan/strategy to reintegrate the various library sources as a streamlined part of the build process (e.g., git submodules or such)? Locating the appropriate sources then patching and rebuilding them is somewhat jarring as it stands... for example, to build with VS2008 I had to "upgrade" the Allegro and Alfont solution files and modify the build settings for all of the libraries, then update the build paths in the AGS solution files. I understand why we would want/need to strip these prebuilt libraries, but I'm wondering if there's a plan for this moving forward.

ivan-mogilko:
Yes, I think @sonneveld wanted to make library sources either downloaded with scripts or having them as submodules, because that would make it much easier for the build server as well.
We already have a script for downloading libraries for linux port that server uses AFAIK.

ivan-mogilko:
Additionally, some of these patches may become unnecessary over time, when we find better solutions.

ivan-mogilko:
Rebased to the top of the latest master.

PR #348, opened 2016-08-15
Fixed missing standard header for std::shared_ptr

monkey0506:
std::shared_ptr/std::tr1::shared_ptr is defined only in the `<memory>`
header file, which is not guaranteed to be included by `<unordered_map>`.
On MSVC2008, `<unordered_map>` does include `<memory>` but this is an
implementation detail and not a standard inclusion.


monkey0506:
A better fix is implemented with #349

PR #349, opened 2016-08-15
Moved stdtr1compat to a header file

monkey0506:
Also checks for VS2015, which is the earliest version of Visual Studio which doesn't need the TR1 namespace for these headers (AFAICT).

### DIFF
--- a/Common/game/customproperties.h
+++ b/Common/game/customproperties.h
@@ -25,17 +25,8 @@
 #ifndef __AGS_CN_GAME__CUSTOMPROPERTIES_H
 #define __AGS_CN_GAME__CUSTOMPROPERTIES_H
 
-#if __cplusplus >= 201103L
-#include <unordered_map>
-namespace stdtr1compat = std;
-#else
-#if defined (_MSC_VER)
-#include <unordered_map>
-#else
-#include <tr1/unordered_map>
-#endif
-namespace stdtr1compat = std::tr1;
-#endif
+#include "util/stdtr1compat.h"
+#include TR1INCLUDE(unordered_map)
 
 #include "util/string.h"
 #include "util/string_types.h"
--- a/Common/util/stdtr1compat.h
+++ b/Common/util/stdtr1compat.h
@@ -1,0 +1,45 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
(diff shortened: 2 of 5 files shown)


sonneveld:
Is this actually tidying it up or is this just adding complication via macros?  Could we leave it messier for now given that all this would go away if we force everyone to use a recent compiler?

monkey0506:
I would argue that this header removes complications and makes the code more readable. I also don't believe there are any plans to move away from VS2008 as the default Windows compiler in the near future, unless that has changed due to @ivan-mogilko stripping the pre-compiled libs. In any case, I think keeping the option open of compiling without C++11 as a strict enforcement is preferable for some (and Microsoft has refused thus far to implement the entire C++11 standard). Other files may also wish to include similar headers before the project files are migrated to a newer Visual Studio, and I think it would be best to not redefine "stdtr1compat" in every single file that needs it.

monkey0506:
Furthermore, something of this nature is required if someone _wants_ to use VS2015 to build, as the `__cplusplus` macro does not report C++11 (due to missing core features from the standard), so the existing conditions are incorrect for VS2015.

sonneveld:
Hrmm, fair enough.  I wonder if we could combine your work with this stackoverflow answer: http://stackoverflow.com/a/5956576/84262 (specifically TR1INCLUDE) or just replace the whole thing with boost.TR1 :)

monkey0506:
How's this? Note that this also invalidates #348 by pulling in the missing header.

sonneveld:
That does look pretty tidy!  Looks like linux and android builds are failing. ios/osx seem to work (they just fail when merged with latest master)

monkey0506:
Looks like I forgot to include the AGS_NEEDS_TR1 macro in that branch of the condition. Pushing a fix now.

ivan-mogilko:
>  Note that this also invalidates #348 by pulling in the missing header.

To clarify, we may just close #348 if #349 is accepted?

ivan-mogilko:
I think I know why do you need these changes right now, are you probably using some kind of custom branch that has #346 and master merged in, as well as MSVS 2015 solution.

But there is a different problem with these changes that I can foresee: at the moment we do not have MSVS 2015 built in the build server, which means that there will be no automatic checks made for repository branches. If anyone includes these headers "old way" again, this will break 2015th build once more and won't let them notice. This will also require a person who has MSVS 2015 to periodically check if master builds for him/her and possibly append fixes.

To fully close this issue we would need to properly support MSVS 2015 solution in the repository and have server build it as a dedicated task (along with MSVS 2008 or instead - this is something to consider).

ivan-mogilko:
Of course what I said above is not a criticism of the proposed code change on its own, but rather the warning if you assumed master to be compatible with 2015 from now on.

monkey0506:
I definitely understand that this isn't a _final_ solution to building with VS2015, but I still see that as tangential to this PR. Ideally we shouldn't accept any (new) use of TR1 headers that aren't passed through this compat layer or else other _existing_ target platforms will likely fail to build anyway.

sonneveld:
I'm happy if VS2015 support breaks every now and again until we add VS2015 to the build server. Hopefully the breakages will be infrequent if at all.

PR #351, opened 2016-08-15
C99-compliant snprintf replacement for MSVC

monkey0506:
MSVC's "_snprintf" is **not** C99-compliant as it does not truncate the buffer on overflow (correction: it does limit how much it writes, but does not null-terminate the buffer). As such it is not a safe replacement for snprintf (as is currently implemented). This provides an alternative that is C99-compliant. Additionally, VS2015 implements a C99-compliant snprintf which causes the existing macro to fail with a compile error. This does not supply the macro if not building with MSVC or if using VS2015 or later.

### DIFF
--- a/Common/util/c99_snprintf.h
+++ b/Common/util/c99_snprintf.h
@@ -1,0 +1,60 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+
+// borrowed and modified from [Stack Overflow](http://stackoverflow.com/q/34912145/1136311)
+//
+// Visual Studio's _snprintf is not C99-compliant as it does not
+// null-terminate the buffer, which may result in unexpected bad memory
+// access. This provides a C99-compliant alternative. From VS2015 this is
+// no longer needed as a compliant snprintf function is implemented.
+
+#ifndef __AGS_CN_UTIL__C99_SNPRINTF_H
+#define __AGS_CN_UTIL__C99_SNPRINTF_H
+
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdio.h>
+
+#define snprintf c99_snprintf
+#define vsnprintf c99_vsnprintf
+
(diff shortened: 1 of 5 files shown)


monkey0506:
Sorry guys, got a bit heavy-handed with copy-pasta. This should be ready to go now.

ivan-mogilko:
I have just one small request, could you put the explanation also in comments to the new header you made?

Common/util/c99_snprintf.h
```diff
@@ -0,0 +1,55 @@
...
+    if (count == -1)
+        count = _vscprintf(format, ap);
+
+    return count;
+}
+
+__inline int c99_snprintf(char *outBuf, size_t size, const char *format, ...)
+{
+    int count;
+	va_list ap;
```
ivan-mogilko: This is the only line that uses tab as indentation; really stands out, this is how I noticed it.

monkey0506: Odd. I'll look at that.

monkey0506:
Not a problem. I'll push a new version in a moment.

Engine/libsrc/apeg-1.2.1/audio/mpg123.c
```diff
@@ -10,7 +10,7 @@
 #endif
 
 #ifndef __GNUC__
-#define snprintf _snprintf
+#include "util/c99_snprintf.h"
```
rofl0r: isn't this one of the files scheduled for deletion in @ivan-mogilko 's cleanup PR ?

ivan-mogilko: I believe you mean #346? No, it is not, at least not for the moment.

ivan-mogilko:
Okay, I just want to clarify this, the comment mentions stack overflow article as the origin of the code (at least that was my impression), but following that link I could not find anything resembling the code, only the description of a problem alone.

Am I missing something, or the link is incorrect?

ivan-mogilko:
Here's the problem. The engine has a number of embedded "semi-automatic" tests that are run at startup  only when the program is built in debug mode. I have a high priority task in my todo list for moving these tests into separate project that could be automatically run by build server, similarily to how Compiler Tests are run. But at the moment you have to check that by running the engine in debug mode yourself.

So, running these tests with your branch causes some assertions to fail around the functions that use snprintf, which means there is now difference in their behavior. We need to find out whether the new snprintf implementation has mistakes, or tests have mistakes.

ivan-mogilko:
Yeah, apparently those tests of mine were assuming that snprintf does not bother about null terminator and fills all the existing "buffer length" with characters. We would need to adjust those tests after pulling this request.

Also, it is curious how many similar assumptions are made in code, since AGS was originally Windows-only program. Searching for "snprintf" shows that on several occasion it is called with (BUFFER_SIZE - 1) instead of BUFFER_SIZE.

PR #353, opened 2016-08-26
Refactor: picked out and cleaned up main game data reading

ivan-mogilko:
This pull request is in some ways similar in concept to the one I did for savegame code: https://github.com/adventuregamestudio/ags/pull/344

Here I picked out all the code that opens, checks, reads and converts (upgrades) main game data from the binary file, and moved it to one unit in Common library (Common/game/main_game_file.h + cpp).
Similarily, decoupled the code strictly related to initialization of runtime game state and moved it to separate code unit (Engine/game/game_init.h + cpp).
Introduced separate enumeration of data reading and engine initialization errors and made corresponding functions return those error codes instead of just quitting the game.

As a bonus I was also able to share this game reading function with Editor (in AGS.Native library) which makes Editor capable of importing any known format version of the compiled game, at least to certain degree.
To elaborate: since we do not have an actual decompiler yet, it is still impossible to fully recover game project if script sources are missing. However, it will now be possible to automatically import game data not only from 2.72 games (which is presently supported by AGS 3.* editors), but from any range of games,  from latest version and down to 2.50.


sonneveld:
Nice.  Happy to review.. but might not be able to until weekend.

ivan-mogilko:
Honestly speaking, this might require more testing, to make sure games of different versions can still be loaded. Most of the loading code was moved without any changes, but relative order of reading/conversion operations changed.

ivan-mogilko:
I had to rebase the branch on top of latest master again because of certain conflicts.

Also, ran multiple tests loading games of the whole range of versions and applied couple fixes. I think new code is generally working.

ivan-mogilko:
Judging by OSX and iOS build logs they fail because of the project file structure change (again), which is going to be fixed separately.

PR #354, opened 2016-08-29
Redirected main output files to Compiled/Data

monkey0506:
This is equal parts clean-up and a feature-fix for the AndroidBuilder plugin I am working on. Everything appears to build and run properly, including audio and speech VOX files and debug.

Particularly, the `jobb` tool used in Android deployment takes an entire directory (recursively) as input, so the data files need to be in a separate directory.

### DIFF
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetWindows.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetWindows.cs
@@ -191,7 +191,7 @@ namespace AGS.Editor
         public override bool Build(CompileMessages errors, bool forceRebuild)
         {
             if (!base.Build(errors, forceRebuild)) return false;
-            string compiledDir = AGSEditor.OUTPUT_DIRECTORY;
+            string compiledDataDir = Path.Combine(AGSEditor.OUTPUT_DIRECTORY, AGSEditor.DATA_OUTPUT_DIRECTORY);
             string baseGameFileName = Factory.AGSEditor.BaseGameFileName;
             string newExeName = GetCompiledPath(baseGameFileName + ".exe");
             string sourceEXE = Path.Combine(Factory.AGSEditor.EditorDirectory, AGSEditor.ENGINE_EXE_FILE_NAME);
@@ -199,7 +199,7 @@ namespace AGS.Editor
             UpdateWindowsEXE(newExeName, errors);
             CreateCompiledSetupProgram();
             Environment.CurrentDirectory = Factory.AGSEditor.CurrentGame.DirectoryPath;
-            foreach (string fileName in Utilities.GetDirectoryFileList(compiledDir, "*"))
+            foreach (string fileName in Utilities.GetDirectoryFileList(compiledDataDir, "*"))
             {
                 if (fileName.EndsWith(".ags"))
                 {
--- a/Editor/AGS.Editor/Panes/GeneralSettings.cs
+++ b/Editor/AGS.Editor/Panes/GeneralSettings.cs
@@ -35,7 +35,8 @@ namespace AGS.Editor
         {
             foreach (Translation translation in Factory.AGSEditor.CurrentGame.Translations)
             {
-                string compiledPath = Path.Combine(AGSEditor.OUTPUT_DIRECTORY, translation.CompiledFileName);
+                string compiledPath = Path.Combine(AGSEditor.OUTPUT_DIRECTORY,
+                    Path.Combine(AGSEditor.DATA_OUTPUT_DIRECTORY, translation.CompiledFileName));
                 if (File.Exists(compiledPath))
                 {
                     File.Delete(compiledPath);
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -34,6 +34,7 @@ namespace AGS.Editor
 
 		public const string BUILT_IN_HEADER_FILE_NAME = "_BuiltInScriptHeader.ash";
         public const string OUTPUT_DIRECTORY = "Compiled";
+        public const string DATA_OUTPUT_DIRECTORY = ""; // subfolder in OUTPUT_DIRECTORY for data file outputs (TODO: change to "Data")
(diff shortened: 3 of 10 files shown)


ivan-mogilko:
@monkey0506 , the concept that editor components (speech, translation) know which build targets exist and derive an actual output path using one of them pretty dubios. Would not that be a better option to tell those components where to compile their files instead?

monkey0506:
The data file "build target" is technically, as you've pointed out, the
*only* build target. The rest just simply deploy those files for other
platforms. The way I've implemented it uses the common interface that is
shared between the data file build target class and the platform classes,
but this is an implementation detail. The VOX files **should** be built to
the Data folder as **part of** the data file builder. These things are core
components of each other, not unrelated items.

ivan-mogilko:
What I mean is, that I do not think that single component should be aware of what should be built and where. I believe they need to only be aware of their own function, and what data should be created for them. It is up to their user, whether that is an Editor, or build target, to know which components should be put into game.

monkey0506:
So are you proposing that I add a global constant which has no meaning
outside of the data file builder, or that I recreate this value in
triplicate? I just don't understand what alternative you're suggesting.
Unless you mean to change the meaning of those constants to only include
their file name, and remove the path from them altogether?

ivan-mogilko:
Yes, I am suggesting for components to keep their output file names only, and make the destination paths to be passed to them by their owner (editor, or data target).

Sorry if I was unclear. I will try to explain again.
What I mean is that this is a question of component hierarchy. In this pull request implementation it seems that all the components are siblings, and every single component, like translation component, for example, knows that it has to be part of the game package, knows where to put the output files.
I suggest that there should be some kind of controller (general Editor class, or build target class) that tells components when to build and where to.


This is actually one of the cases which illustrate the need of design plan for the Editor, like I was repeatedly mentioning on forums... Regardless of the chosen way, there has to be a very clear explanation of how the build process should look like in the end from the point of view of component ties.

ivan-mogilko:
Ah, actually, I now remembered that it was always like this, that these components made their paths themselves using some global constant (OUTPUT_DIRECTORY). I guess you just altered the way OUTPUT_DIRECTORY is received.

Maybe it is possible to use this as a chance to change this? I would like to find a good solution here that would be acceptable.

monkey0506:
Now that I better understand what you meant, I agree. I'll be able to amend
this in a way I think should meet what you've described.

monkey0506:
Ultimately the VOX files, which are built by the components, need to be placed in the data folder, whether that's "Compiled" or "Compiled/Data". Since the VOX files are built by the components **after** the BuildTargetDataFile class has been built, I don't actually see a way around making this subdirectory a global constant (in the AGSEditor class). I'll push up a commit with **only** that change, and if you think it's okay ( @ivan-mogilko ) then I'll push up the other changes once again.

ivan-mogilko:
Not implying changing anything right now, but there is some kind of discrepancy in BuildTargetDataFile only dealing with the main file. AFAIK the other targets deal with all files (vox too). I think we might need to review build target functionality at some point. For example, BuildTargetDataFile could run all component compilation, not only main data file.

Regarding current situation, I was hoping there is a way to pass the path as an argument into component building functions, having this path taken from BuildTargetDataFile, as you initially wanted to have it there. Now I am not certain if this is easy to achieve, because apparently there is a multitude of different methods being called (PreCompileGame, ExtraCompilationStep, and so on), and trying to get an extra argument into every place where it is used may require to rewrite larger part of code.

So maybe my proposal was too early to implement. If you think that using that variable will work, so be it.

monkey0506:
The build process was about as far away from "generic" as it could have been when I started. I'm not saying everything in my approach was perfect, but until the entire process can be more full abstracted, then I don't think that there's a better solution than this. Abstracting that process will require much more planning, however, and this PR is meant as a feature-fix for my previous work, and is needed by the Android plugin I am working on (as an aside, the deployment process is already functional, I'm finishing now adding a UI to collect the necessary/relevant inputs).

Editor/AGS.Editor/Panes/GeneralSettings.cs
```diff
@@ -35,7 +35,7 @@ private void DeleteAllCompiledTranslations()
         {
             foreach (Translation translation in Factory.AGSEditor.CurrentGame.Translations)
             {
-                string compiledPath = Path.Combine(AGSEditor.OUTPUT_DIRECTORY, translation.CompiledFileName);
+                string compiledPath = BuildTargetsInfo.FindBuildTargetByName(BuildTargetsInfo.DATAFILE_TARGET_NAME).GetCompiledPath(translation.CompiledFileName);
```
monkey0506: Missed this. Will resubmit.

Editor/AGS.Editor/BuildTargets/BuildTargetWindows.cs
```diff
@@ -191,15 +191,15 @@ public void UpdateWindowsEXE(string filename, CompileMessages errors)
         public override bool Build(CompileMessages errors, bool forceRebuild)
         {
             if (!base.Build(errors, forceRebuild)) return false;
-            string compiledDir = AGSEditor.OUTPUT_DIRECTORY;
+            string compiledDataDir = BuildTargetsInfo.FindBuildTargetByName(BuildTargetsInfo.DATAFILE_TARGET_NAME).GetCompiledPath();
```
monkey0506: Missed this one as well, sorry. Did a search against `IBuildTarget` and forgot these one-liners.

ivan-mogilko:
Okay, I have just double-checked that all compilation works with this pull. Is it ready to be merged, or you are still adding some commits?

As a side note, the engine should be fixed to not explicitly checking "Compiled" folder when being run under debugger. I am thinking about implementing new command line arguments for that.

monkey0506:
@ivan-mogilko I've removed the WIP label, but to clarify does the fix you mentioned need to be a part of this? I'd forgotten that the debugger looks in the Compiled folder for File access and such. I only tested that VOX files worked.

ivan-mogilko:
No, I meant that it does not have to be.

PR #355, opened 2016-09-10
[WIP] Build step interface

monkey0506:
Creating this PR early for peer review.

As @ivan-mogilko has pointed out, the `BuildTarget` classes are primarily used in deploying the already "_built_" game files, which leads to some oddities in the way I implemented those classes (for example, the `ExtraOutputStep` event is raised _before_ any platform-specific builds). Admittedly, mistakes were made in both understanding of the existing code and in creating a more extensible approach. Ivan suggested that the editor should have a more incremental approach to building and deploying, and that indeed part of that would be to allow more verbose logging during the build process.

The initial commit here adds a new class to generate _informational_ compile messages. Note that the existing `CompileMessage` class in AGS.Types is an `ApplicationException` subclass. I decided against renaming that, as it may cause confusion to introduce a new class by the same name, but the `CompileInfoMessage` class is **not** an `Exception`, so it is not a `CompileMessage`. To allow `CompileMessages` (container class) to accept a `CompileInfoMessage`, then, the `CompileMessages` class had to be changed as well. Therefore I introduced the `ICompileMessage` interface to allow `CompileMessages` to accept both `CompileMessage`s (including `CompileError`s and `CompileWarning`s) and `CompileInfoMessage`. These changes make it possible to generate output during building without raising warnings or errors. A `CompileInfoMessage` may be shown with or without an information icon.

This initial commit should also rename various function parameters which refer to `CompileMessages errors`. I have omitted those changes from this commit, for now, to prevent clutter, since I am asking for review of these changes.


ivan-mogilko:
Now, this should be interesting :-).
I will check this out when I have time, but there is no need to wait after each commit (just saying).

PR #357, opened 2016-09-21
Engine: Fix return value for GetVoiceMode

gurok:
Ensure that GetVoiceMode (the Speech.VoiceMode getter) returns the correct value when a vox file isn't available. Operations need to be done in reverse order of those in SetVoiceMode.

### DIFF
--- a/Engine/ac/global_audio.cpp
+++ b/Engine/ac/global_audio.cpp
@@ -439,7 +439,7 @@ void SetVoiceMode (int newmod) {
 
 int GetVoiceMode()
 {
-    return play.want_speech >= 0 ? play.want_speech : (-play.want_speech + 1);
+    return play.want_speech >= 0 ? play.want_speech : -(play.want_speech + 1);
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Is there a bug that exists in recently released version(s) of AGS?

gurok:
It's not a recent regression. From what I gather, this bug has been present since the introduction of the Speech class (3.3.0).

Here's a simple test case:

In a new project, with no vox file...

```
function game_start()
{
	Speech.VoiceMode = 2;
	Display("%d", Speech.VoiceMode);
}
```

Displays "4".

I think this means it qualifies for being pushed to master. To be honest, going forward, I thought the general approach was to push to master and then branch master for a new release as necessary. I didn't think there'd even be 3.x.x-specific commits anymore. At least, not for very long.

ivan-mogilko:
Ok, I got it, I had to remember how those modes are stored.

PR #358, opened 2016-09-25
Engine: fix AdjustVolumeWithScaling

AlanDrake:
The adjusted volume was never actually applied to the audio channel

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -2076,7 +2076,7 @@ void animate_character(CharacterInfo *chap, int loopn,int sppd,int rept, int noi
 
 void CheckViewFrameForCharacter(CharacterInfo *chi) {
 
-    int soundVolumeWas = play.sound_volume;
+    int soundVolume = SCR_NO_VALUE;
 
     if (chi->flags & CHF_SCALEVOLUME) {
         // adjust the sound volume using the character's zoom level
@@ -2084,17 +2084,15 @@ void CheckViewFrameForCharacter(CharacterInfo *chi) {
         if (zoom_level == 0)
             zoom_level = 100;
 
-        play.sound_volume = (play.sound_volume * zoom_level) / 100;
+        soundVolume = zoom_level;
 
-        if (play.sound_volume < 0)
-            play.sound_volume = 0;
-        if (play.sound_volume > 255)
-            play.sound_volume = 255;
+        if (soundVolume < 0)
+            soundVolume = 0;
+        if (soundVolume > 100)
+            soundVolume = 100;
     }
 
-    CheckViewFrame(chi->view, chi->loop, chi->frame);
-
-    play.sound_volume = soundVolumeWas;
+    CheckViewFrame(chi->view, chi->loop, chi->frame, soundVolume);
 }
 
 Bitmap *GetCharacterImage(int charid, int *isFlipped) 
--- a/Engine/ac/viewframe.cpp
+++ b/Engine/ac/viewframe.cpp
@@ -16,6 +16,7 @@
 #include "ac/viewframe.h"
 #include "debug/debug_log.h"
 #include "media/audio/audio.h"
+#include "media/audio/soundclip.h"
 #include "ac/spritecache.h"
 #include "gfx/bitmap.h"
 #include "script/runtimescriptvalue.h"
@@ -30,6 +31,7 @@ extern GameSetupStruct game;
 extern ViewStruct*views;
 extern SpriteCache spriteset;
 extern CCAudioClip ccDynamicAudioClip;
+extern SOUNDCLIP *channels[MAX_SOUND_CHANNELS+1];
 
 
 int ViewFrame_GetFlipped(ScriptViewFrame *svf) {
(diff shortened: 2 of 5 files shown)


ivan-mogilko:
Hi, Alan.
Can you give an example of a test case to look how it works/does not work?

AlanDrake:
Sure, let me make a quick demo project.

ivan-mogilko:
Instructions to make one would be sufficient, but if you can do actual project, that would do even better.

AlanDrake:
Here it is:
https://dl.dropboxusercontent.com/u/32465342/Test%20adjustvolumewithscaling.zip

It has been broken since the beginning, I had fixed it on my old edition, but the fix didn't make it here.
Basically a character with AdjustVolumeWithScaling to true should scale the volume of the sounds played by their view. This does not happen because while the correct volume is set on the global variable, it is never set to the channel playing the clip.

ivan-mogilko:
I think I understand what you mean, but I am worried that play_audio_clip_on_channel() function might not be the good place to apply this volume. After all, this is a generic function that has no direct relation to characters.

Also, the fact that volume that should be used at certain character at certain location is applied to global variable sounds very weird on its own... I believe this should be investigated.

ivan-mogilko:
To make it clear, are you speaking of function CheckViewFrameForCharacter() ?
I can see that engine basically does this:
1) stores old play.sound_volume;
2) adjusts play.sound_volume;
3) calls some function to start a sound;
4) puts old play.sound_volume value back.

This all looks like some terrible hack to me. Maybe it won't be too hard to refactor this procedure?

AlanDrake:
Exactly that.
I'm not sure how to refactor, it's been years since I last handled AGS source, there's a lot that has changed since then.

ivan-mogilko:
I am very much hesistant to apply the fix like this, because it applies an arbitrary global variable inside a function that can be called for multitude of reasons, increasing complexity of code. It is not immediately obvious whether it won't break other things (now or in future).

I'd rather have this problem written as an issue in the tracker, with your explanation (or bumped up, if there is already one), so that we could investigate it and find a good solution without haste.

AlanDrake:
Issue #359

ivan-mogilko:
Thank you.
I will reply tomorrow probably, I need to refresh my memories about changing audio volume.

AlanDrake:
Pull request updated.
Let me know if this is good enough.

ivan-mogilko:
Alan, I think this is still incorrect calculation.

Original was:
> play.sound_volume = (play.sound_volume * zoom_level) / 100;
Now it is:
> soundVolume = zoom_level;

As you may notice, the original code applied *relative* volume. In your new code you set *absolute* value.
I think you need to apply zoom factor to the volume this sound clip already has after it started, because it may have different volume, depending on Editor settings. This volume can be retrieved using clip->get_volume().

ivan-mogilko:
Another minor thing, regarding the code style, could you please put statement under "if" on the next line? (brackets are not enforced if there is just one statement)

AlanDrake:
Amended. I thought there was some magic further down the line to adjust to the global volume.
**But wait**, now that I think about it, the new audio clip can have a default volume, should I take that too into consideration ? The original didn't seem to do it, probably because it was written before the new audio system...

ivan-mogilko:
Default volume is already taken into account when the clip starts in load_sound_clip.

AlanDrake:
It is, but at the moment it's being replaced by the new volume from _CheckViewFrame_.

I might get the current volume from the channel and scale that value with the new volume.
`channels[channel->id]->set_volume_percent(channels[channel->id]->get_volume() * sound_volume / 100);`

I've tested the change and updated the pull request.

ivan-mogilko:
Yes, this is what I was noting few comments above :).

ivan-mogilko:
Alan, wait, you keep changing the code into different directions... now it is different again, and it does not make any sense now!

You are making soundVolume serve as a factor for the clip volume, but at the same time making soundVolume itself calculated from play.sound_volume ? Why?
It should be simply equal to zoom, and play.sound_volume does not play any role in current audio system.

AlanDrake:
When you pointed out I had to to a relative volume I wrongly assumed you meant I was supposed to use play.sound_volume. I thought maybe that was the global volume or something...

Amending the commit again...

ivan-mogilko:
Ok, sorry if I was not clear.

What I am trying to say is, that it is not really important for your case, what variable has global volume, and how it is assigned. Because when you get AudioChannel pointer, all the correct volume should *already* be applied.
So you do not have to worry about these things, and simply do your custom relative adjustment.

AlanDrake:
This bugfix was trickier than expected. I hope you'll stick around for code reviews when (if) you leave.

ivan-mogilko:
> I hope you'll stick around for code reviews when (if) you leave.
Ok :).

Re the fix, it is working!
You can also try changing your clip's default value in the editor, and you can notice it is actually taken into account.

AlanDrake:
My first successful pull request :)

I already tested by changing the clip's default volume and subfolders clip volume. It sounds fine to me.

PR #362, opened 2016-10-02
Make stopping movement optional in (Un)LockView

monkey0506:
Character::UnlockView currently always stops any character movement. This preserves that as the default behavior, but gives the option to continue character movement while releasing the locked view.

Related to the [QueuedSpeech module bug](http://www.adventuregamestudio.co.uk/forums/index.php?topic=23806.msg636534826#msg636534826) reported by cat. The module script line in question was an incomplete attempt at doing the same, and was ~~not meant for production release.~~ probably not ~~thoroughly~~ tested, because the released code was in a completely impossible state.


gurok:
Can I suggest a few things?
 - Use an enum instead of a boolean: eStopMoving, eKeepMoving
 - Add supporting documentation in the online help for this, stating that eStopMoving is the default
 - Perhaps think about adding a similar switch for LockView, which I believe stops movement as well

ivan-mogilko:
There are number of issues with current commit.
Most important is that you **must not** remove old exported function signatures, because that will break all the game that used those until now.
For example of how to add new parameters to existing functions, please look into Character.ChangeRoom nad Region.Tint (they were added new parameters during last couple of years).

ivan-mogilko:
Additionally, you change ReleaseCharacterView declaration in the Editor, but I do not see it modified in the engine.
Honestly though, I believe it should not be changed, because that is an obsoleted function that is not supposed to be used in new scripts at all.

Also, you are using indentation mismatching our convention.

Editor/AGS.Editor/Resources/agsdefns.sh
```diff
@@ -2080,13 +2088,29 @@ builtin managed struct Character {
   /// Checks whether this character is in collision with the object.
   import function IsCollidingWithObject(Object* );
   /// Locks the character to this view, ready for doing animations.
-  import function LockView(int view);
+  import function LockView(int view
+#ifdef SCRIPT_API_v341
```
monkey0506: Note that this commit is affected by #364

ivan-mogilko:
Ouch... hmmm somehow I did not realize this SCRIPT_API macros system would require declaring new version number everywhere.
Not that it is critical, but it's just that we do not know what version we are making right now, it may be 3.4.1, or 3.5... it should be deduced only when we decide to actually prepare release, depending on number and quality of changes we will have at that point.

I need to think a bit about this...

ivan-mogilko:
To elaborate, what bothers me - is that if the next version released would be different from the one declared here (e.g. 3.5.0), then IMO there should not be any traces left of this "temporary" version in the script api selector nor in the script api header, because that will only confuse users.

monkey0506:
We should only increment the version to 3.5.0 if there are major changes to the engine. Currently, that is not the case. If during the pre-release phase major changes are made to the engine, then there is **no** harm in incrementing from, say, 3.4.1.9 to 3.5.0.0. The same thing happened when 2.63 grew too large and was updated to 2.7. There was no *official* 2.63 release, only pre-releases, and by the time 2.7 release was official, no traces were left of 2.63.

We should use 3.4.1 as the version unless major changes are made to the engine, at which point we can change the pre-release version. Changing the version from 3.4.1 to 3.5.0 would **include** changing the script API enums, macros, etc. They would only exist in the git history, they would not be left behind in the source, because there would be no *official* 3.4.1 release, only pre-releases. Any code dependent on `SCRIPT_API_v341` should be forced to update to `SCRIPT_API_v350` or break. That is the **nature** of pre-releases.

ivan-mogilko:
Yes, this is what I mean, the only issue is to make sure whoever does the release in future wont forget about all this....

monkey0506:
This is why I included the new enum value (from which the macros are generated) as part of the build version commit. Updating the script API is now part of updating the build version.

monkey0506:
That is, whenever a new major or minor version is **added** (rather than only the build/revision increasing), a new script API version must also be added -- unless the change affects the editor only. Perhaps this should be documented somewhere, but it's a necessary part of the versioning system.

ivan-mogilko:
Yes, I see what you mean and agree with that.

However, I have a small idea I would like to discuss.

This ScriptAPILevel enum is not directly (programmatically) connected with program version. Moreover it techincally does not have to include actual version numbers. For example, it should be possible to declare enum item like:

> SCRIPT_API_Post340Dev

This will clearly indicate that we are dealing with non-finalized API version.

Again, technically it won't require actual program version to be rised, but since we already have some new code in master branch, that might be a good idea to increase it too.

monkey0506:
Is there any actual benefit behind doing things this way? Utilizing a pre-release version like 3.4.1 which may change before the final release allows using things like `#ifver`, and leaves behind no artifacts unless the final version *actually* changes. Utilizing a non-deterministic API level like `Post340Dev` leaves behind artifacts which must be cleaned up later, regardless of whether the pre-release and final release version numbers are the same.

ivan-mogilko:
Hmmm... ok, that's a good point too.

ivan-mogilko:
The sole benefit from having special enum is that it stands out, so it is less likely to be missed.
With the version number in macros, someone can forget that this version was never released, because there is nothing that indicates that in the code.

monkey0506:
Or, putting it a different way, what is the *actual* harm of calling this `version 3.4.1`? **If** major changes are implemented, we may **then** need to call it something else, but if that isn't the case then it would end up as 3.4.1 *anyway*. It just seems like you're worried about labeling this version *too early*, while it's actually impossible to build a *functioning* pre-release without giving it a version.

ivan-mogilko:
The sole benefit from having special enum is that it stands out, so it is less likely to be missed.
With the version number in macros, someone can forget that this version was never released, because there is nothing that indicates that in the code.

monkey0506:
> With the version number in macros, someone can forget that this version was never released

The only way that I can perceive this **actually** causing a problem is if a script module or plugin is released based on a pre-release version of AGS, and **not** subsequently updated to match the final release of that version. This, IMO, isn't something that (AGS developer hat on) we should be worrying about. If anything, perhaps the recommended idiom for module/plugin developers should be:

    #ifdef SCRIPT_API_v341
    // blah
    #else
    #error This module depends on AGS 3.4.1+.
    #endif

Which is already possible, and as a module and plugin developer, I have used the equivalent with `#ifver`.

ivan-mogilko:
No, what I mean is simply that if an AGS developer will forget to check if such version was released, they may add a new enum constant instead of replacing old one, and thus there will be listed a version that was never released.

monkey0506:
Even though the script API is not required to change with the engine version, it should only be updated in relation to the engine version being updated. So I stand by what I said earlier, that new script API enums should only be added if a new major/minor version is being added, and should be done as part of updating the build version. There is no plausible reason for 3.4.1 to be replaced with 3.5.0 in all places, yet the v341 script API enum left behind. It's feasible that it could be forgotten, but this should be amended, and there is no reason to expect it to happen.

monkey0506:
Even if the v341 enum were forgotten and a 3.5.0 pre-release (or even final release) were published, the breaking change of removing the enum would be by-design, and any developers who had used the v341 API should be expected to update their scripts or be unable to update to a newer engine.

There is little to no cause to believe that a game developer would select v341 API if using AGS 3.5.0, while no AGS 3.4.1 release exists. So the only way that a game would feasibly be published with v341 API in lieu of an AGS 3.4.1 release is if the game were published using a pre-release of AGS. Again, breaking changes are in the nature of pre-releases, so even if AGS 3.5.0 final release was published without removing the v341 API, it would still be appropriate to make a breaking patch to remove that enum.

I contend that it is much worse to create artifacts in the source which **must** be updated prior to release than to **hypothetically** create a scenario in which someone's game *might* break (due to **user error**).

monkey0506:
You're describing a scenario in which a maintainer:

* Is adding a new build version (a prerequisite for adding a new script API version) to a pre-release version (potentially upgrading to final release status)
* Forgets that they, themselves, are updating the build version from 3.4.1 to 3.5.0
* Fails to remove the v341 script API enum
* Publishes a (pre-)release

**AND** a game developer:

* Has published a final game using the former pre-release version
* Utilized the v341 script API
* Loses the source to the game **OR** does not publish the source, but disappears from contact
* Wishes the game to be usable by future versions of AGS

I don't see the merit to this argument.

ivan-mogilko:
Firstly, I am not counting game developer in this scenario at all. Stop bringing that here.
Secondly where does this come from?
> Forgets that they, themselves, are updating the build version from 3.4.1 to 3.5.0

I was about to say it's okay and end this discussion, but you bring more stuff that basically ridicules me.

monkey0506:
I didn't mean to ridicule you, but I just honestly have no idea what else you could have possibly meant by:

> No, what I mean is simply that if an AGS developer will forget to check if such version was released, they may add a new enum constant instead of replacing old one, and thus there will be listed a version that was never released.

It was clearly a misunderstanding, but it's not important. Are you wanting to devise some alternate strategy to what I've done with f71a4b2? Otherwise, are there any other issues with this PR?

ivan-mogilko:
No, I think it is okay.

monkey0506:
> 1&#8203;.  I am worrying about MovementStyle name not suiting this enum very well...

I agree, but I was trying to keep it generic. I think `StopMovementStyle` makes it more clear, while keeping it generic for possible reuse elsewhere.

> 2&#8203;.  You have enum in script, but in the engine you have an integer parameter...

I reviewed how other script enums are handled, and it seems they are always parsed as `int` parameters. I added a couple of `#define` macros in `runtime_defines.h` to match the behavior of other script enums.

ivan-mogilko:
> I agree, but I was trying to keep it generic. I think StopMovementStyle makes it more clear, while keeping it generic for possible reuse elsewhere.

I apologize greatly for keeping you busy with this, what do you think about ChangeMovement(Style)? I think it would imply that it does not always stop. But StopMovementStyle would work too.

> I reviewed how other script enums are handled, and it seems they are always parsed as int parameters. I added a couple of #define macros in runtime_defines.h to match the behavior of other script enums.

Just a small note: having these macros declared you are still doing comparison with literal 0 in the Character_LockViewEx and Character_UnlockViewEx().

monkey0506:
The closest thing we have for comparison in way of naming convention for this kind of script enum would be `BlockingStyle` and `RepeatStyle`, which could be seen as asking the question, "Do you want to block?" or "Do you want to repeat?". This is kind of my justification for `StopMovementStyle` ("Do you want to stop movement?"). `ChangeMovementStyle` seems weaker in this respect, IMO, but if anyone else chimes in, I'm willing to go with whatever is deemed the more fitting name.

And you're right that I forgot to use the `KEEP_MOVING` macro in those two relevant lines. I'll push up a quick fix for that.

ivan-mogilko:
Well, with the Script API version switch we now have a potential workaround of renaming enums in the future, if a strong need arise.

Issue #366, opened 2016-10-14
ags crashes on startup when trying to play "A Tale of Two Kingdoms" when built with gcc 6.2


berolinux:
Trying to run
www.adventuregamestudio.co.uk/site/games/game/905/
in ags (3.4.0.12 tag) on Linux x86_64 results in a crash on startup.

$ ags ATOTK.exe
Adventure Game Studio v3.4 Interpreter
Copyright (c) 1999-2011 Chris Jones and 2011-2016 others
ACI version 3.4.0.12

Initializing allegro
Reading configuration
Looking for game data file
Game data file: /home/bero/abf/twokingdoms/BUILD/twokingdoms/ATOTK.exe

Game data version: 32
Requested engine version: 2.7.2.0
Setting up window
Initializing TTF renderer
Initializing mouse: number of buttons reported is 3
Checking memory
ci_find_file: cannot change to directory: Compiled
ci_find_file: cannot change to directory: Compiled
Initializing keyboard
Install timer
Initialize sound drivers
Trying digital driver ID: 'Auto' (0xffffffff), MIDI driver ID: 'Auto' (0xffffffff)
Installed digital driver ID: 'ALSA' (0x414c5341), MIDI driver ID: 'AMID' (0x414d4944)
Install exit handler
Initialize path finder library
Load game data
Game data version: 32
Requested engine version: 2.7.2.0
Game GUI version: 115
A Tale of Two Kingdoms
Checking for disk space
Initializing resolution settings
Game native resolution: 320 x 200 (16 bit)
Device display resolution: 2560 x 1440
Game settings: windowed = no, screen def: max, screen size: 0 x 0, match device ratio: yes, game scale: max_round
Using graphics factory: DX5
Created graphics driver: Allegro/DX5
Supported gfx modes (16-bit): 
        640x480;640x480;640x480;640x480;640x480;720x400;720x480;720x480;
        720x576;800x600;800x600;800x600;800x600;832x624;1024x768;1024x768;
        1024x768;1152x864;1280x720;1280x720;1280x720;1280x800;1280x960;1280x1024;
        1280x1024;1440x900;1600x1200;1680x1050;1920x1080;1920x1080;1920x1080;1920x1200;
        2560x1440;
Supported gfx modes (15-bit): 
        640x480;640x480;640x480;640x480;640x480;720x400;720x480;720x480;
        720x576;800x600;800x600;800x600;800x600;832x624;1024x768;1024x768;
[...]

PR #368, opened 2016-10-15
Fix check for CAFEBEEF magic value

berolinux:
gcc 6.2 and clang 4.0 optimize away the previous check because an
unsigned variable (size_t newlen) is compared to an signed
negative constant (0xCAFEBEEF).

Solves part of issue #366.

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -340,7 +340,7 @@ void ReadDialogs(DialogTopic *&dialog,
         while (1)
         {
             size_t newlen = in->ReadInt32();
-            if (newlen == 0xCAFEBEEF)  // GUI magic
+            if (static_cast<int32_t>(newlen) == 0xCAFEBEEF)  // GUI magic
             {
                 in->Seek(-4);
                 break;


ivan-mogilko:
I am probably missing something here, size_t is unsigned, and uint32_t is still unsigned, how does this fix work then?

monkey0506:
@ivan-mogilko you're right about them both being unsigned types. I'm guessing that was a typo. As for why it "works", my guess is that clang isn't optimizing it because of the cast, even with the signed/unsigned comparison.

berolinux:
True, was a typo but nevertheless works. I've just confirmed that casting it to int32_t works as well, that may be the more obvious variant... Will send an updated request.

berolinux:
Patch fixed to cast to int32_t instead of uint32_t

berolinux:
The real explanation why casting it to uint32_t works as well is that uint32_t is a 32-bit value.

32-bit signed CAFEBEEF == 32-bit unsigned CAFEBEEF
But the unsigned 64-bit variable (size_t is 64-bit) doesn't convert to 0xCAFEBEEF even if it's implicitly cast to signed

ivan-mogilko:
Ah, alright; just wanted to make sure this code is really what you intended to do.

PR #369, opened 2016-10-25
ScriptAPI: Character and Object properties for getting tint values

ivan-mogilko:
Added **properties for getting Tint component values** from Characters and room Objects.
Added **LightLevel readonly property** (SetLightLevel is still used to explicitly set light level).
Added **HasExplicitLight** to Character to complement **HasExplicitTint**, and add both of same properties to the Object.
This makes Character and Object's tint API compliant to the Region API in ways to read currently applied light & tint parameteres, and complementary to each other.

An explanation is due, why I propose to still have SetLightLevel to set light for characters and objects, as opposed to Region.LightLevel property serving as both setter and getter of a light level.

Unlike Region, which always has **some** light level set (even if default 0, meaning no change in lighting), Character and Object has a distiction between "explicit light level" and "no individual light level". The former overrides regions and ambient tint, the latter does not. Explicit light level of 0 is a thing, so setting character's LightLevel to 0 would not turn individual lighting off, instead it would still override area lighting with default value.

Therefore, in my opinion, having SetLightLevel function to change char's and object's light would better indicate that we do not only change some value, but also toggling lighting mode (even though at a cost of people get slightly annoyed that they cannot assign values to LightLevel property).

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -88,6 +88,9 @@ struct CharacterInfo {
     int get_blocking_top();    // return Y - BlockingHeight/2
     int get_blocking_bottom(); // return Y + BlockingHeight/2
 
+    inline bool has_explicit_light() const { return (flags & CHF_HASLIGHT) != 0; }
+    inline bool has_explicit_tint()  const { return (flags & CHF_HASTINT) != 0; }
+
 	// [IKM] 2012-06-28: I still have to pass char_index to some of those functions
 	// either because they use it to set some variables with it,
 	// or because they pass it further to other functions, that are called from various places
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
I am putting this back into WIP state for now, because there were some existing API peculiarities I did not pay attention to (like Characters already have HasExplicitTint property).

ivan-mogilko:
Another note: I changed the meaning of Character.HasExplicitTint. Earlier it meant to check whether character has *either* tint or lightlevel, now it checks only for the tint, and light level is checked using HasExplicitLight.
It works as before for games compiled with previous script api level.

PR #370, opened 2016-10-30
Feature  scalenativeresolution

AlanDrake:
With this branch I've aligned the D3D rendering to DD,
A game property ScaleNativeResolution (default to false) has been added to re-enable the legacy behaviour.
This option can be configured both from the editor's general settings and winsetup.

The full list of changes involves the following:
- the render routine now uses a surface the size of game resolution by default
- added ScaleNativeResolution property to switch between scaling mode
- added the option to winsetup and game settings
- removed redundant proportional calculations between D3DMATRIX and _renderSprite. Now _renderSprite just uses the x,y coordinates and the translation is left to the matrix
- GetCopyOfScreenIntoBitmap() now forcibly uses the native game sized surface, both for legacy and non-legacy scaling. This fixes all the consistency issues for screen reading (eg: badly resized crossfade transition between rooms when resolution multiplier is not rounded, etc).

Note: room crossfading with FlipScreen() is still broken, as it has always been. This baffles me, because if the acquired screen surface is properly flipped, which I've tested by saving a screenshot, then how the heck is this bug still happening ? Hmm, perhaps it's a case of double flipping, I'll need to investigate.


ivan-mogilko:
There are number of issues here, related to different parts of the program, so I will split my review in parts.

First of all I would like to make several notes on the game settings.

I see that you added an option to the Editor, but it still being used as a default value for config file (when creating config), not as the actual game setting. You added new constant (OPT_IGNORENATIVERES) to the game.options array, but never use it in the engine, which defeats its purpose.
I believe there might be a misunderstanding on how the game option should work, and this has to be elaborated.

When we discussed this scaling mode on forums, we eventually came to agreement, that the scaling mode should be controlled by TWO options: game setting, and config option. Game setting is something that is set by game author at design time, and config option is set by players.
Idea was that since this scaling mode *might* be part of game design, the game setting is allowed to disable reading of config. In other words, the game setting should have following values:
* Enabled (e.g. use legacy scaling, depending on how you make this option)
* Disabled
* Read from user config

(Text could be altered if you find better wording.)

Or, alternatively, instead of just "Read from config":
* Read from user config (default: on)
* Read from user config (default: off)

The latter is not *that* essential, because game author can prepare default config file by other means (by hand or with setup application) before packaging the game.


When the config file is created by the Editor, this option is written there only if the game setting has "Read from user config".

When the engine reads config, it should check this game setting from **game.options[OPT_IGNORENATIVERES]**, and either use author's choice (Enabled/Disabled), or get the value from config tree, if the **OPT_IGNORENATIVERES** sais "Read from user config".

Additionally, since setup application now preloads some options from the game data, I think that may even be possible to disable checkbox in WinSetup if the **,OPT_IGNORENATIVERES** is not "Read from config". (You can see how this is done to send certain game parameters to setup in engine_check_run_setup()).

AlanDrake:
I had temporarily given up using OPT_IGNORENATIVERES because I couldn't access game.option. I ended up having some trouble importing the extern where I wanted to read it, so I went the second best option, saving the default on new configs. I'll try sending the settings through engine_check_run_setup(), when I glanced it I thought those INIwriteint() were writing directly to file. 

"Read from config" could be "User defined"

Will look into it

ivan-mogilko:
If you need any tip on how to access game objects, I'll be glad to help.

Regarding INIwrite* and INIread functions, they no longer write directly to file anymore. Instead they are working with the key/value tree (dictionary), which may be read from and written to file if wanted; but in the case I was referring to it is just being passed to setup application, since it's integrated into the engine.

AlanDrake:
Be sure to rebase or reset to fb499eb before pulling.

ivan-mogilko:
I have to say, that even knowing the background of this feature does not help me when I am trying to understand how to use this switch, either from game designer perspective, or engine's coder perspective. Both the setting name and descriptions/comments are very confusing.

The fact that in many places this option is called "ScaleNativeResolution", but in one place it is called "IGNORENATIVERES" only worsens the situation. Do those settings have same meaning, or opposite ones? That is, if I have ScaleNativeResolution = true, should IGNORENATIVERES be true, or false? I cannot know that without investigating the code.

For example, the comment to OPT_IGNORENATIVERES sais:
> use the legacy D3D scaling that ignores the native game resolution

This is simply baffling. I would not understand what "ignores native resolution" means if I did not know what the "legacy D3D scaling" is.

ivan-mogilko:
By the way, regarding ScaleNativeResolution and OPT_IGNORENATIVERES relation. Judging by comments, they should have opposite meanings ("ignoring the actual game resolution" versus "ignores the native game resolution").

But in the code they are directly assigned without any conversion, e.g.:
<pre>game->Settings->ScaleNativeResolution = (ScaleNativeResolution)thisgame.options[OPT_IGNORENATIVERES];</pre>

ivan-mogilko:
For example, something that has to be fixed or elaborated, is the difference in terms: "actual game resolution" and "native game resolution". Is "actual game resolution" = "native" or these are different things?

ivan-mogilko:
I would like to take couple of days and think of possibly more elaborate descriptions.

ivan-mogilko:
Feature itself is working properly (I tested all variants).

Perhaps I could merge this request and work on enhancing descriptions and comments later, unless you have a good idea how to improve them right away.

AlanDrake:
I'm conflicted about the names, but I can't come up with anything better.
This is about disregarding the intended resolution things are rendered at, like emulators, who usually give a "native resolution" option one can change at will. 
I'm just not good at naming things.

Want me to rename IGNORENATIVERES to SCALENATIVERES for consistency sake ?

ivan-mogilko:
> Want me to rename IGNORENATIVERES to SCALENATIVERES for consistency sake ?

If they are same thing, then yes please rename.

AlanDrake:
Done.
But, take note that the game option and the property have different values, the former has the three values discussed (user-defined, enabled, disabled) while the latter is just true/false.

Maybe that was the reason I felt compelled to give it a different name... oh well.

ivan-mogilko:
> But, take note that the game option and the property have different values, the former has the three values discussed (user-defined, enabled, disabled) while the latter is just true/false.

Wait, this is not how it should be... also I did not notice that be the case judging by the code.
OPT_* constant should be able to have all 3 values, otherwise it looses required functionality.

AlanDrake:
Yes, the OPT is the one with all the 3 values, otherwise I wouldn't be able to see if I need to force or not the setting.

ivan-mogilko:
@AlanDrake , I was considering some variants.

What about "*Subpixel sprite scaling*"? Although I know that "subpixel rendering" usually means different thing, the idea behind this render mode is to use pixels that are smaller than scaled-up  native "pixels" are.

Regardless of whether it is a good name or not, I've also came with following property description:

*When drawing zoomed character and object sprites, AGS will take advantage of higher runtime resolution to possibly give scaled images more detail, than it would be possible if the game was displayed in its native resolution*

AlanDrake:
The property description sounds good.
The subpixel thing still sounds misleading, maybe _FreeScaling_... BrokenScaling, HiresScaling. IncreaseResolution, IgnoreNativeResolution, LegacyScaling, RenderToScreenResolution...

ivan-mogilko:
Hmm... erm, "Use high-res pixels for sprite scaling"?

AlanDrake:
We could ask the forums, probably useless, but you never know.

ivan-mogilko:
What about *Scale sprites in display resolution* or *Render sprites in display resolution*?
Because this is technically what the renderer is doing.

AlanDrake:
_"Render sprites at screen resolution"_, yeah. That doesn't sound bad.

PR #371, opened 2016-11-11
Provide means for the alternate versions of AGS to indicate their games

ivan-mogilko:
There is an issue with alternate versions of AGS Editor that they have no means to tag their games so that original engine could at least detect that the game does not work as expected before trying to load and run it. General data format number is not suitable for this purpose, because there is no guarantee that it won't be duplicated by "official" engine, or another alternate editor.

While there are not so many alternate versions, sometimes users try to run such games with regular engine or its port, and they receive no elaborate message why had this happened.

This change adds simply an *optional* list of strings at the game data header, called "extended capabilities". When the game is opened engine tests the required list of caps with the caps it supports.

It is true that there is no way for us to force alternate version use these to mark their, but we may at least provide means for them to tag their games with custom strings, and recommend to do it.
This set of strings may also come handy to mark any experimental builds, or perhaps for other unusual purposes in the future.

### DIFF
--- a/Editor/AGS.Editor/DataFileWriter.cs
+++ b/Editor/AGS.Editor/DataFileWriter.cs
@@ -1282,6 +1282,13 @@ namespace AGS.Editor
             writer.Write(NativeConstants.GAME_DATA_VERSION_CURRENT);
             writer.Write(AGS.Types.Version.AGS_EDITOR_VERSION.Length);
             WriteString(AGS.Types.Version.AGS_EDITOR_VERSION, AGS.Types.Version.AGS_EDITOR_VERSION.Length, writer);
+            // Write extended engine caps; none for this version
+            writer.Write((int)0);
+            // An example of writing caps (pseduo-code):
+            //   writer.Write(caps.Count);
+            //   foreach (cap in caps)
+            //       FilePutString(cap.Name);
+            //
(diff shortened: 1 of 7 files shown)


Editor/AGS.Editor/DataFileWriter.cs
```diff
@@ -1282,6 +1282,13 @@ public static bool SaveThisGameToFile(string fileName, Game game, CompileMessage
             writer.Write(NativeConstants.GAME_DATA_VERSION_CURRENT);
             writer.Write(AGS.Types.Version.AGS_EDITOR_VERSION.Length);
             WriteString(AGS.Types.Version.AGS_EDITOR_VERSION, AGS.Types.Version.AGS_EDITOR_VERSION.Length, writer);
+            // Write extended engine caps; none for this version
```
sonneveld: Would it be worth setting it up so all non-release builds have a special CAP flag?

ivan-mogilko: If I understand your idea right, this is to make release builds refuse to run games made with non-release builds?
Considering we often have breaking changes in the process of development, this may have certain sense.
What are the negative consequences:
* we must enforce the rule that we remove this flag as soon as we make first public release.
* if someone makes a game using nightly build from the build server, that game won't be runnable by final release even if technically they are compatible.

PR #372, opened 2016-11-13
Font vertical offset property

ivan-mogilko:
The ALFont library currently used by AGS to draw TTF is built around very old version of the FreeType font, and also have some additional tweaks done by Chris Jones. It has an issue that it ignores font's descent parameter while keeping its real height, thus aligning font's baseline to the bottom line. This results in some fonts rendered lower than expected.

This pull request adds an optional vertical offset property to fonts, which is applied when a text line is drawn. It works for fine relative positioning of the line of text.

A note should be made that even after we upgrade or fix fonts library, this property may still come handy for dealing with backwards-compatibility cases and imperfectly designed fonts.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -92,7 +92,7 @@ Custom properties changed at runtime.
 Ambient lighting
 
 48 : 3.4.1
-OPT_RENDERATSCREENRES, extended engine caps check.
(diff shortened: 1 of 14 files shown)


sonneveld:
Reviewed.

ivan-mogilko:
Had to rebase to master, because of #371.

PR #374, opened 2016-11-22
Split graphics mode initialization into three distinct steps

ivan-mogilko:
Although the free display mode support was previously implemented into AGS engine, the code still retained older trait of being straitforward one-way initialization of everything at once, with little support to change anything during the game.

This pull request refactors the related code and breaks gfx mode init routine into three distinct steps, which (in theory) could be called again at runtime separately or altogether:
* Setting screen (window) resolution;
* Setting render frame (the final rectangle game is stretched to);
* Setting scaling filter (which is used at stretching when needed).

### DIFF
--- a/Common/util/geometry.h
+++ b/Common/util/geometry.h
@@ -157,6 +157,8 @@ struct Size
     }
 };
 
+// TODO: consider making Rect have right-bottom coordinate with +1 offset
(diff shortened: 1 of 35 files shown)


sonneveld:
Looks pretty straightforward. Nothing obvious leapt out.

PR #375, opened 2016-12-02
Add mouse utility methods

monkey0506:
Includes Mouse.SelectPreviousMode and Mouse.IsModeEnabled. These functions are trivial in the engine source, but less trivial for end-user devs to implement.

### DIFF
--- a/Engine/ac/mouse.cpp
+++ b/Engine/ac/mouse.cpp
@@ -201,6 +201,10 @@ void SetNextCursor () {
     set_cursor_mode (find_next_enabled_cursor(cur_mode + 1));
 }
 
+void SetPreviousCursor() {
+    set_cursor_mode(find_previous_enabled_cursor(cur_mode - 1));
+}
+
 // set_cursor_mode: changes mode and appearance
 void set_cursor_mode(int newmode) {
     if ((newmode < 0) || (newmode >= game.numcursors))
@@ -291,6 +295,12 @@ int IsButtonDown(int which) {
     return 0;
 }
 
+int IsModeEnabled(int which) {
+    return (which < 0) || (which >= game.numcursors) ? 0 :
+        which == MODE_USE ? playerchar->activeinv > 0 :
+        (game.mcurs[which].flags & MCF_DISABLED) == 0;
+}
+
 //=============================================================================
 
 int GetMouseCursor() {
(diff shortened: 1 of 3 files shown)


Engine/ac/mouse.cpp
```diff
@@ -535,8 +581,10 @@ void RegisterMouseAPI()
     ccAddExternalStaticFunction("Mouse::EnableMode^1",              Sc_enable_cursor_mode);
     ccAddExternalStaticFunction("Mouse::GetModeGraphic^1",          Sc_Mouse_GetModeGraphic);
     ccAddExternalStaticFunction("Mouse::IsButtonDown^1",            Sc_IsButtonDown);
+    ccAddExternalStaticFunction("Mouse::IsModeEnabled^1",           Sc_IsModeEnabled);
     ccAddExternalStaticFunction("Mouse::SaveCursorUntilItLeaves^0", Sc_SaveCursorForLocationChange);
     ccAddExternalStaticFunction("Mouse::SelectNextMode^0",          Sc_SetNextCursor);
+    ccAddExternalStaticFunction("Mouse::SelectPreviousMode^0",      Sc_SetPreviousCursor);
```
rofl0r: why is the enduser func called "SelectNextMode" and the internal func "SetNextCursor" ? these 2 terms seem to have an entirely distinct meaning.

ivan-mogilko: This is a legacy issue, in AGS "cursors" and "modes" are strictly tied components of same entity (the graphical and logical ones), the type is even called "CursorMode", which can refer both to chosen interaction mode and its cursor.
The cursor has sprite attached to it, which may be swapped, but cursor is always same for the given mode, so in practice saying "next cursor" is same as "next mode".

ivan-mogilko:
@monkey0506 , I would really like to ask you to move the duplicating code from **find_next_enabled_cursor** and **find_previous_enabled_cursor** into the separate function. I am talking about the code that tests if the mouse cursor can be selected:

<pre>
        if ((game.mcurs[testing].flags & MCF_DISABLED)==0) {
            // inventory cursor, and they have an active item
            if (testing == MODE_USE) 
            {
                if (playerchar->activeinv > 0)
                    break;
            }
            // standard cursor that's not disabled, go with it
            else if (game.mcurs[testing].flags & MCF_STANDARD)
                break;
        }
</pre>

monkey0506:
@ivan-mogilko Done. I also slightly modified my implementation of `IsModeEnabled` to reflect that `eModeUseinv` is disabled while the player has no active inventory.

Issue #287, opened 2015-10-21
Using one config file for many games


ivan-mogilko:
This was requested by user in the [comment to another issue](https://github.com/adventuregamestudio/ags/issues/104#issuecomment-41592817):

@JPRuehmann

> Hello
> Me to I thing at least the possibility of an central config  such as an ~/.acsetup or /etc/acsetup file would be great.
> I´ve got fifty or so AGS games and all of them should run with the same Configuration.
> Thanks,
> JPR

Judging by the comments in code, AGS should actually first try to find a config in the "current directory". Need to check how that actually works (because it may change current directory to game dir early).
OTOH, it is perhaps viable to pass path to config as a new command line argument.

rofl0r:
how about a global/user-specific "override config":
engine first reads config from game dir, then overrides its values with the users one
so user can set for example all games to fullscreen, but leave the default resolution to the game's config.

ivan-mogilko:
*bump*

With the configuration map implemented in the recent releases, this should be easier to do. In fact we already potentially read two config files (at least on Windows), where one in the game-related "user's directory" overrides one in the installation directory.

Adding one extra source in all-games shared path that overrides map again should not be a problem. Only problem is to decide where this should be located (on every supported system).

ivan-mogilko:
This may also be a chance to decide whether Linux needs a separate path for a game-specific config.

Since AGS 3.3.5 there are couple of config options that may be altered during the game, and engine writes them back into config file.
On Windows there was an issue that game may be installed into "C:/Program Files", and program is not allowed to write back there unless it is run with administrator's privileges.
This is why AGS now reads config from the game installation directory first, and then overrides it with config read from users directory (and saves config into latter when quitting the game).

No one has reported similar issue on Linux yet (probably everyone is putting games into user's home dir?), but in may that be a theoretical possibility?

rofl0r:
how about this:
game first reads config from game dir, then overwrites it with values from $HOME/ags/gamename/config, and finally with values from $HOME/ags/global.conf
for saving, game always tries to save into $HOME/ags/gamename/config.

ivan-mogilko:
I opened pull request: #378, but I am suggesting different order for global and game config: "$HOME/ags/global" first and "$HOME/ags/gamename/config" second, because otherwise global config may override options that were written by game itself (which could be chosen by player from some game menu).

JPRuehmann:
That is exactly what is intended by that.

[...]

PR #378, opened 2017-01-06
Global (all-games) user configuration

ivan-mogilko:
As of (rather old) feature request (#287), this adds support for user's global configuration file, that is - a single file that is applied for every game.

The suggested rule is that config files are read in following order, every next one overridding previous ones (only options that are included in the current file are overridden, options that are not included keep their values from previous configs):
* config file inside game installation directory (aka "default game config");
* user's global, all-games config file;
* user's game-specific config file.

The latest is also one that is allowed to be modified by the engine (for example, to write config options that could be changed at runtime, such as game translation).

Actual config location is different for every platform. At the moment only two supported desktop platforms have such locations specified, but they could be easily added for other OSes if needed.

*Linux*
* user's global config: $XDG_DATA_HOME/ags/acsetup.cfg
* user's game config: $XDG_DATA_HOME/ags/GAMENAME/acsetup.cfg

NOTE: if $XDG_DATA_HOME is not defined, then "$HOME/.local/share" is used instead, similar as for saved games.

*Windows*
* user's global config: %USERPROFILE%/Saved Games/.ags/acsetup.cfg
* user's game config: %USERPROFILE%/Saved Games/GAMENAME/acsetup.cfg

### DIFF
--- a/Engine/main/config.cpp
+++ b/Engine/main/config.cpp
@@ -51,11 +51,11 @@ extern char psp_translation[];
 extern char replayfile[MAX_PATH];
 extern GameState play;
 
-//char datname[80]="ac.clb";
-const char *ac_conf_file_defname = "acsetup.cfg";
-String ac_config_file;
+// Filename of the default config file, the one found in the game installation
+const String DefaultConfigFileName = "acsetup.cfg";
 
 // Replace the filename part of complete path WASGV with INIFIL
+// TODO: get rid of this and use proper lib path function instead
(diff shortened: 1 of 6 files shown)


rofl0r:
nice one. is there some kind of documentation project where this information could be added?

ivan-mogilko:
No, we do not have any documentation project, but I could explain this in OPTIONS.md for now, where the config options are listed.

PR #384, opened 2017-02-15
Scripting: Do not push 0-size structs onto the stack

gurok:
If a function contained an instance of a struct with no fields, exiting the function would cause a stack underflow. This bug has been present since at least 3.2.1, possibly earlier.

Note: I am specifically talking about non-managed structs.

It could be argued that the engine's stack pointer handling should know how to properly deal with 0-size structs, but I think not pushing them to the stack is the simplest solution.

Here is a quick example of what I mean:
```
struct MyStruct {};

function OffendingFunction() { MyStruct a; }
```

Also, I am still working on the switch/case string comparison bug. This was in a stash and annoying me, so I thought I'd commit it.

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -3358,8 +3358,10 @@ int parse_variable_declaration(long cursym,int *next_type,int isglobal,
       sym.entries[cursym].flags |= SFLG_STRBUFFER;
       scrip->fixup_previous(FIXUP_STACK);
     }
-    scrip->cur_sp += varsize;
-    scrip->write_cmd2(SCMD_ADD,SREG_SP,varsize);
+    if(varsize > 0) {
+      scrip->cur_sp += varsize;
+      scrip->write_cmd2(SCMD_ADD,SREG_SP,varsize);
+    }
   }
   if ((getsvalue != &lbuffer) && (getsvalue != NULL))
     free(getsvalue);


ivan-mogilko:
I have a possibly related question. Did you try what happens when you make a global variable of that struct? This was something that I planned to test for a while.

The reason I ask, there seem to be an ancient bug which we only met once with one game (titled "Metal Dead"). After I rewrote script interpreter and added more checks around the code, that game began to stop with error "global variable refers to data beyond allocated buffer". I was not able to find a workaround, so I changed error to warning. You may find it here:
https://github.com/adventuregamestudio/ags/blob/master/Engine/script/cc_instance.cpp#L1872

I emailed game authors to see if I can get any information about their scripts, but they never replied back.

I wonder if this may somehow be related to the empty global variables.

ivan-mogilko:
Regarding this error and your fix. I was very curious how it could happen, and also worrying that this could be engine's fault for not deinitializing something properly, so I ran some debugging on two scripts:
**First**:
<pre>
struct MyStruct {};
function game_start() { MyStruct a; }
</pre>
**Second:**
<pre>
struct MyStruct { int a; };
function game_start() { MyStruct a; }
</pre>

What I found is that when the struct has positive size, script does (SCMD_ADD, SREG_SP, 4) on variable allocation, and (SCMD_SUB, SREG_SP, 4) on deallocation.

But when the struct has zero size, script does (SCMD_ADD, SREG_SP, 0)  on allocation, but no SCMD_SUB whatsoever. In other words, SCMD_ADD/SCMD_SUB pair is not closed.

---

A bit on how stack works now in AGS. There are two counters: stack entry pointer, which counts "elements" allocated on stack, and stack bytes pointer, which counts actual bytes allocated. When something is added to stack, the element counter is incremented by 1, and some bytes are saved, associated with that element. When something is removed from stack, element counter is decremented by 1, and associated number of bytes is cleared.

Now, the thing is, that in the original AGS engine stack worked bit differently: it did not have any explicit "element" counter, it had only bytes counter, and simply advanced stack pointer by number of bytes forth and went back after.

I believe this is the reason why original AGS engine simply did not notice this script mistake: since element size was 0, there was no difference whether it was subtracted or not.

Now there is difference, because it also counts number of allocated elements (variables). So it demands explicit matching deallocation.

---

From the first glance, there are two solutions here:
1) Do not do SCMD_ADD for empty structs, this is the one you chose.
2) Find out why there is no matching SCMD_SUB for this case and add one. Maybe compiler does not find any bytes left in stack and thinks that it removed everything?

Currently my concern is, whether the presence of empty variable, which won't get registered on stack, may cause any miscalculations if other variables are allocated prior or after it?
What do you think?

ivan-mogilko:
By the way, strictly speaking, we might need to add a workaround for running old games that had empty local variables, if such game will ever be found. Something like checking remaining stack elements and removing them on function end if these elements have size = 0.

gurok:
This matches my analysis pretty closely.

I considered option 2, but I believe it would be a case of modifying code here:

https://github.com/adventuregamestudio/ags/blob/master/Compiler/script/cs_parser.cpp#L3540

and changing:

```
            if (totalsub > 0) {
                scrip->cur_sp -= totalsub;
                scrip->write_cmd2(SCMD_SUB,SREG_SP,totalsub);
            }
```

to:

```
            if (totalsub >= 0) {
                scrip->cur_sp -= totalsub;
                scrip->write_cmd2(SCMD_SUB,SREG_SP,totalsub);
            }
```

But this is frought with danger. We want to account for 0-size structs, but we don't want to write a SCMD_SUB if there were no local variables. I chose option 1 as the better option. It seems the only way option 2 could work at a compiler level is to have two arrays when compiling (number of elements and number of bytes).

But as you noted, is not just a mismatch of SCMD_SUB and SCMD_ADD. My analysis was "It seems that there is a problem allocating or deallocating 0-byte variables on the stack. Not having 0-byte structs solves the problem." Your analysis is much better though:

> E: Curiously, engine actually does that if there is any valid variable declared before the empty one, because it is being told to deallocate N bytes, and removes all elements one by one until N bytes got removed from stack.

Thank you for elucidating this. Also, technically, it would be a variable declared after, like so:

```
struct MyStruct {};
function game_start() { MyStruct a; int b; }
```

Regarding your first comment, I have not found any similarities or apparent problems with empty structs when testing. I will keep an eye out, but as you've already closed this, I'm guessing you also concluded that they're unrelated bugs.

Also, sorry for the delay in reply. I was busy at work all week and meant to come back to this sooner. Thank you for your in-depth reply too. It is basically everything I understood with one of the gaps filled in.

PR #386, opened 2017-02-19
Scripting: Fix the switch statement's handling of strings

gurok:
String comparisons in switch statements haven't been working properly since 3.4.0 final. Instead of comparing the value of strings, the switch statement was comparing the pointers. I have fixed this issue and some incorrect offsets I found when diagnosing the problem.

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -29,6 +29,7 @@ int  check_operator_valid_for_type(int *vcpuOp, int type1, int type2);
 void yank_chunk(ccCompiledScript *scrip, std::vector<ccChunk> *list, int codeoffset, int fixupoffset);
 void write_chunk(ccCompiledScript *scrip, ccChunk item);
 void clear_chunk_list(std::vector<ccChunk> *list);
+int  is_any_type_of_string(int symtype);
 
 void yank_chunk(ccCompiledScript *scrip, std::vector<ccChunk> *list, int codeoffset, int fixupoffset) {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I have a question. There was a big test you referenced in one of the past pull requests: 
https://github.com/adventuregamestudio/ags/pull/322#issuecomment-222319556

I was using it to double check these changes, and I found that the first test case does not work:
<pre>
    for(int a = 0; a < 100; a++)
    {
        switch(true)
        {
            case a > 0:
                continue;
        }
        if(a == 0)
            Display("%d", a);
        else
            Display("X");
    }
</pre>

That does not work before your current changes too, and I do not remember if that worked before.

ivan-mogilko:
RE previous comment. I think I narrowed the problem down. The mentioned cases do **not** work when local variable is used in "case" expression calculation. But global variables **do** work.

ivan-mogilko:
Yes, basically local variables are not properly compared in case statements.
For example, if I have this:
<pre>
    Character *pc;
    pc = player;
    switch (player)
    {
      case pc:
        Display("pc matches");
        break;
    }
</pre>

If "pc" is a global variable this works properly, and if "pc" is a local variable it does not work.

I think this is a separate and pretty distinct issue. so I think I'll merge your fixes to string comparison now.

gurok:
I have an idea of what might be causing this. I will run it against my old tests before committing again (actually forgot about writing those). I'd also like to commit something that will prevent switching on a literal null.

PR #388, opened 2017-02-26
Windows: strip engine and editor of precompiled libraries (compatible with Allegro 4.4.2)

ivan-mogilko:
This pull request is a remake of https://github.com/adventuregamestudio/ags/pull/346.

---

Windows version of AGS historically used precompiled libraries built by original author; some of these libraries were actually modified, either to achieve backwards-compatibility with older games or other reasons.

This changeset removes those binaries and provides patches instead to let developers build Windows version fully from source.
The content of patches, when not available, was either reverse-engineered or otherwise deduced. Notably, alfont patch is based on reverse-engineering by @jjsat (we also use that work for other ports).

Windows readme is updated with explanation on building with library sources.

---

Since the last pull request I modified the engine code and it no longer requires patching pack_fopen function of Allegro. Also, I decided to remove a code dealing with certain Windows keys from the patch too and propose linking against Allegro 4.4.2 instead (which supposedly has few compatibility fixes related to contemporary versions of Windows).

After those changes, the patch only makes several Allegro variables non-static, so that AGS could refer to them. They are:

1) DirectX interface pointers for sound and input; these are only required for non-portable plugin support which probably could be dropped later.
2) hw_to_mycode variable used for some kind of keyboard hack (Windows version only). I did not have time to look into that yet, so I do not know if there is better solution than using Allegro's internal symbol.


ivan-mogilko:
I am currently trying to set up our own Allegro's fork (https://github.com/adventuregamestudio/lib-allegro) to have CMake scripts to create static libraries, as needed for the Windows version of AGS.
After that is done, we could hopefully fix the build server work with these changes.

ivan-mogilko:
Ah, disregard the previous comment, I found which  parameters to set in cmake to get a static lib solution.

ivan-mogilko:
On the other hand, Android build seem to be broken with this pull request. I suspect that may be because it also had "pack_fopen" renamed to "old_pack_fopen" in its allegro patch file.

ivan-mogilko:
The build server is owned by @sonneveld , and as far as I know he does not have much free time now, so it is unknown when he will be able to adjust automatic building.

Since Windows versions are built very often by developers, that won't be critical if they won't on the server for a while.
(Apparently, automatic tests still work, IIRC they do not require Allegro and other libraries)

I would like to ask those who work with Windows version to try if they can build this branch using instructions I added to README.md. If it will be possible, I will merge this pull request already.

PR #390, opened 2017-03-08
Custom font line spacing

ivan-mogilko:
This pull request does two things:

1) Introduces new customizable Font property LineSpacing, which defines an automatic height of step between two lines of text. Default value is 0, which actually means "use font's height".

2) Reworks the internal engine's algorithm of text drawing, using line spacing where font height was previously used for calculating steps between lines. Since default value for line spacing is "font height", old games and new games which did not have custom line spacing set should keep identical text looks.

Also added two script functions: GetFontHeight(int font) and GetFontLineSpacing(int font), which are self explanatory.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -94,6 +94,9 @@ Ambient lighting
 48 : 3.4.1
 OPT_RENDERATSCREENRES, extended engine caps check, font vertical offset.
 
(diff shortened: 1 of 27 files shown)


ivan-mogilko:
Rebased to latest master, after some refactoring was made there.

PR #391, opened 2017-04-01
Scripting: Fix switch on local variables

gurok:
Ensure the stack size is the same when parsing a case expression and when it's executed. I'm now including the stack operations in the chunk associated with a case expression.

The imbalance that previously existed was to blame for local variable comparisons not working in switch statements (MAR was wrong).


gurok:
Just a quick note, I tested this against the "gauntlet" in #322

ivan-mogilko:
As a formal reminder, automatic server not building Windows versions is normal, since it still not configured for a new compilation process.

Compiler Tests pass though, which is a good sign.

PR #392, opened 2017-04-01
Fix headings in README

torkleyy:


ivan-mogilko:
Hm, since then these stopped working? Did they change mardown rendering rules?
Anyway, thanks for pointing out!

PS. Few other readme files need similar fix too.

tzachshabtay:
They changed it about two weeks ago:
https://githubengineering.com/a-formal-spec-for-github-markdown/

PR #395, opened 2017-04-16
Ran the VS upgrader on solution "AGS.Editor.NoNative"

persn:
It was bothering me.

### DIFF
--- a/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
+++ b/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
@@ -1,4 +1,4 @@
-﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="3.5">
+﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
@@ -19,9 +19,10 @@
     </SccProvider>
     <FileUpgradeFlags>
     </FileUpgradeFlags>
-    <OldToolsVersion>2.0</OldToolsVersion>
+    <OldToolsVersion>3.5</OldToolsVersion>
     <UpgradeBackupLocation>
     </UpgradeBackupLocation>
+    <TargetFrameworkVersion>v2.0</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
--- a/Editor/AGS.Controls/AGS.Controls.csproj
+++ b/Editor/AGS.Controls/AGS.Controls.csproj
@@ -1,4 +1,4 @@
-﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="3.5">
+﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
   <PropertyGroup>
     <ProjectType>Local</ProjectType>
     <ProductVersion>9.0.30729</ProductVersion>
@@ -34,7 +34,8 @@
     </SccAuxPath>
     <SccProvider>
     </SccProvider>
-    <OldToolsVersion>2.0</OldToolsVersion>
+    <OldToolsVersion>3.5</OldToolsVersion>
+    <TargetFrameworkVersion>v2.0</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <OutputPath>bin\Debug\</OutputPath>
--- a/Solutions/AGS.Editor.NoNative.sln
+++ b/Solutions/AGS.Editor.NoNative.sln
@@ -1,6 +1,8 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 10.00
-# Visual Studio 2008
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26403.3
+MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AGS.Controls", "..\Editor\AGS.Controls\AGS.Controls.csproj", "{D937E1F4-1B9F-4567-B28E-EED24AA31C9C}"
 EndProject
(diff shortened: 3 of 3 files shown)


ivan-mogilko:
Does it upgrade solution to VS2015? Out latest version is set up to be built in VS2008 still.

persn:
I upgraded with VS2017, the more recent VS starting from 2010 are fairly compatible across each other so it matters little which you use to upgrade. Upgrading the solution will result in Minimum Visual Studio required to VS2010 for that particular solution, and yes, that will be inconsistent with the other solutions.

I guess this PR is just as much to gauge what we can and cannot do about this as the VS2008 legacy solution is really tedious and disruptive, I can't do something as simple as change the branches without having to reset the entire upgrade, and then do the upgrade all over again when I have checked out the new working branch. Just doing this entire process as I was switching back and forth between with my testing with making the editor build for Linux in #396 and other work I haven't uploaded yet resulted in me running the upgrader 10 times in just a couple of hours. There are of course stuff I can do to circumvent the problem, but not without adding other inconveniences in the process.

Is it really important if the solution files are consistent in this manner? We're talking about a 4 lines worth of header in the top of the solution files which work in most cases completely independent from the rest of the solution file. On top of that, the solution files work completely independent of each other. Doesn't the "AGS.Editor.NoNative.sln" exist for the purpose of letting C# developers develop away without being stopped by the C++ code portion and its limitations? As it currently stands it feels like people working in C# is getting punished for whatever limitations is in place by the engine, compiler and whatever else is made in C++.

The best possible compromise here I can think of is to simply upgrade the C# projects, omit the solution file upgrade, and then I can make my own copy "AGS.Editor.NoNative.VS2017.sln" or whatever which I'll add to my local git exclude. However it would be really stupid if AGS goes another 3, 6, 9 or 12 months with solution files made for VS2008.

ivan-mogilko:
> Is it really important if the solution files are consistent in this manner?

This was my question... because I do not know, but I may check.

ivan-mogilko:
Alright, I tried to open Editor.Full solution in your branch, and it gave me following error:

> Unable to read the project file 'AGS.Controls.csproj'. 
> <...>\Editor\AGS.Controls\AGS.Controls.csproj: The tools version "12.0" is unrecognized.

The problem is that your change also modifies shared project headers. These projects are used by both Editor.Full and Editor.NoNative solutions.

I know it is my fault that these projects are still not upgraded. I had the plan for removing VS2008 dependency about a year ago, but then decided to wait for build server to be updated before doing upgrade. I guess main problem is that I very rarely get reminded of these things, that I decided that they are not urgent.

If you give me just couple of days, I can upgrade all projects in repository to VS2015 (I have that one installed) and let you continue from that point?

persn:
Alright, I'll wait for you to do the upgrade.

ivan-mogilko:
This looks somewhat more complicated than I first thought.

AGS.Native.dll is a C++/CLI project, and when upgraded to VS2015 it implicitly targets .NET 4.0 (other projects are still 2.0).

I can upgrade all other projects to 4.0 (otherwise Editor cannot reference AGS.Native). But I need to clarify, because I am a bit confused, is .NET 4.0 supported by Windows XP? We are still supporting XP, and I would not like to drop that support just like that...

I tried other ways, like editing project file by hand and making it target 2.0 framework, also setting "Visual Studio 2015 - Windows XP (v140_xp)" toolset version, but all of those options caused various compilation & linking issues, which, I fear, would require additional research to solve (with no guarantee of success).

PS. With .NET 4.0 Editor builds fine and seem to work well (rebuilt couple of AGS games with it).

ivan-mogilko:
Another note, I'd rather split out branch for v3.4.1 release before pushing in this upgrade; the release is almost ready, and I do not feel confident enough about this update to add it there.

Further development in master could continue in VS2015.

persn:
> I tried other ways, like editing project file by hand and making it target 2.0 framework, but that caused linking issues with CRT. I fear going that way would require additional research (with no guarantee of success).

[So I researched this](http://stackoverflow.com/questions/2921702/change-c-cli-project-to-another-framework-than-4-0-with-vs2010), and what it looks like is that Microsoft has tied in the .NET version for C++ into the Native Toolset version, the reasoning for this being that in order for C++/CLI to work, the native toolset must also have some awareness to the .NET framework in use. What that means is that adjusting AGS.Native framework version to v3.5 or lower won't work unless you also have VS2008 and possibly VS2010 installed.

> I can upgrade all other projects to 4.0 (otherwise Editor cannot reference AGS.Native). But I need to clarify, in case I am missing anything, is NET 4.0 supported by Windows XP? We are still supporting XP, and I would not like to drop that support without a warning...

[It sure is.](https://msdn.microsoft.com/en-us/library/8z6watww(v=vs.110).aspx)

persn:
So I've done some more research.. Turns out upgrading to .NET 4.0 may not be a high enough of a upgrade  unfortunately, at least if there's going to be any development towards bringing the AGS Editor to *nix systems. The problem is that Mono has dropped support for any compatibility with .NET below minimum v4.5. So if I or anyone else is going to pursue to work I recently began in #396 it is required that we at least upgrade to v4.5, or alternatively v4.6.1 which is the highest supported version by Mono.

By upgrading to v4.6.1 we would lose Windows Vista users(is that a problem? Is there anyone using Vista?) and minimum required VS will be 2015.

By upgrading to v4.6 we would keep the Vista support, however minimum required VS will still be 2015.

By upgrading to v4.5 we would keep the Vista support, however minimum required VS will be 2012.

https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#History
https://msdn.microsoft.com/en-us/library/8z6watww(v=vs.110).aspx

You'll notice on my screenshot that when I attempt to run AGS it tries to circumvent the issue by running the .NET 2.0 application in a default environment, what you can't see on my screenshot is that this freezes the entire OS when it attempts to show the splash screen, so it is stuck on that transparent rectangle.. Updating the framework version resolves the issue.

screenshot 2017-04-16 19 22 47

ivan-mogilko:
Sorry, but I won't update anywhere beyond 4.0 while I am on it. That's maximal supported on Windows XP SP3, and last time I heard there are still users who run AGS on Windows XP. But that seem to be enough to at least make AGS be fully built in VS 2015, so that looks like a acceptable move for now.

Dropping WinXP in favor of potentially being able to run editor on Unix is something that has to be discussed with community first, I believe.

persn:
Alright that's fair enough, I'll settle with 4.0

persn:
> Dropping WinXP in favor of potentially being able to run editor on Unix is something that has to be thought through, and discussed with community first, I believe.
Perhaps there is an alternate solution, like using projects with different settings, but I would leave that for you or anyone else who will be working on the editor in the future.

There may be a solution here that avoids any conflicts. Shared Projects for C# was introduced in VS2015. How it works is that there is one project type Shared project which acts as a project container holding all the code files. It will be the job of a regular C# project to reference the shared project and resolve information like assembly references, target framework, precompiler directives and other project config. If we hold all our code files AGS.Editor in a shared project, and use regular .NET projects with different framework versions we should be able to provide an environment where *nix development and Windows XP development can co-exist.

ivan-mogilko:
> There may be a solution here that avoids any conflicts. Shared Projects for C# was introduced in VS2015. How it works is that there is one project type Shared project which acts as a project container holding all the code files.

A very funny coincidence, just todays morning I was thinking that a system like that would make a great addition to MSVS projects :).

ivan-mogilko:
I opened #401 to upgrade whole thing to vs2015.

ivan-mogilko:
I believe this pull request may be closed after #401 was merged?

PR #396, opened 2017-04-16
Fixed C# project module "AGS.Types" so that it successfully builds in xBuild on *nix systems

persn:
Like the title says, I have more details on the individual commit messages so make sure to read those.

### DIFF
--- a/Editor/AGS.Types/DialogOption.cs
+++ b/Editor/AGS.Types/DialogOption.cs
@@ -7,13 +7,13 @@ using System.Xml;
 namespace AGS.Types
 {
     [DefaultProperty("Text")]
+    [DeserializeIgnore("EntryPointOffset")]
     public class DialogOption
     {
         private int _id;
         private string _text = string.Empty;
         private bool _show;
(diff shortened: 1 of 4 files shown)


sonneveld:
Thanks for this and other pull requests.. I'll have to get the build server working again before I can accept them though.

persn:
No rush.. I'm still experimenting with the other C# modules.

ivan-mogilko:
Regarding EntryPointOffset which references otherwise unused _entryPointOffset.
Comment there sais:
> Pre-3.1.1 wrote this into the Game.agf XML, so we have to keep this dummy property to allow it to deserialise.

This is a caveat of AGS project serialization system, which I do not really like. Something could be done with it later, but for now you could remove this property if you add following attribute to the DialogOption class:

<pre>[DeserializeIgnore("EntryPointOffset")]</pre>

That's an attribute I created while ago for the purpose of deprecating properties without breaking deserialization.

persn:
Updated according to feedback

PR #397, opened 2017-04-16
New icon for the Editor

persn:
Based off of the splash screen

https://www.adventuregamestudio.co.uk/forums/index.php?topic=53510.msg636537102#msg636537102


persn:
Here's a screen shot showing off what it looks like.

screenshot 2017-04-16 18 19 14

ivan-mogilko:
Might be nitpicking, but I am using classic theme in Windows, the one that has bluish title, and since the icon is bluish too, it looks not very elaborate:

classic theme

Don't know how they solve things like that, making icons suitable for generic theme. Maybe add a white outline?

persn:
Nice catch, I'll look into it

persn:
Hrrmm, I was hoping or assuming there would be an easy way to add outline to a shape in GIMP but I was mistaken. Image editing isn't really my skill set either way, maybe it is better we outsource this to the community? If someone makes some stylish editing to my 256x256 image I can make an icon out of it and implement it.

ags_cup

ivan-mogilko:
> maybe it is better we outsource this to the community?

Certainly. This is how we got new splash screen :).

monkey0506:
The way I generally add an outline (using Photoshop, but I assume GIMP has
similar tools) is to use the magic wand to select the transparent area
(non-contiguous, anti-aliased), invert the selection, expand the selection
by a few pixels, (optionally) feather the selection by 1 pixel to smooth
any aliasing, and finally flood-fill the selected area on a layer beneath
the object.

persn:
I tried what @monkey0506 suggested but I didn't have the skills to avoid making the edges look awful.

https://www.adventuregamestudio.co.uk/forums/index.php?topic=54726.0

persn:
I uploaded the most recent proposal, it looks much better than before

persn:
Updated.

PR #398, opened 2017-04-18
Environment.OSVersion output correct OS

persn:
Environemnet.OSVersion wouldn't output correct OS version on Windows 8.1 and Windows 10. This is solved by a adding a manifest file with metadata that tells the application at build time that the OS in question are supported.

Microsoft do not actually recommend using Environemnet.OSVersion, but it works just fine  and they haven't provided alternatives that doesn't involve invoking native code or doing registry lookups, which I'm trying to avoid.

### DIFF
--- a/Editor/AGS.Editor/app.manifest
+++ b/Editor/AGS.Editor/app.manifest
@@ -1,0 +1,76 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="3.4.1.3" name="AGSEditor"/>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <!-- UAC Manifest Options
+             If you want to change the Windows User Account Control level replace the 
+             requestedExecutionLevel node with one of the following.
+
+        <requestedExecutionLevel  level="asInvoker" uiAccess="false" />
+        <requestedExecutionLevel  level="requireAdministrator" uiAccess="false" />
+        <requestedExecutionLevel  level="highestAvailable" uiAccess="false" />
+
+            Specifying requestedExecutionLevel element will disable file and registry virtualization. 
+            Remove this element if your application requires this virtualization for backwards
+            compatibility.
+        -->
+        <requestedExecutionLevel level="asInvoker" uiAccess="false" />
+      </requestedPrivileges>
+    </security>
+  </trustInfo>
+
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- A list of the Windows versions that this application has been tested on and is
+           is designed to work with. Uncomment the appropriate elements and Windows will 
+           automatically selected the most compatible environment. -->
+
+      <!-- Windows Vista -->
+      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}" />
+
+      <!-- Windows 7 -->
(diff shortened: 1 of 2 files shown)


Editor/AGS.Editor/app.manifest
```diff
@@ -0,0 +1,76 @@
...
+
+        <requestedExecutionLevel  level="asInvoker" uiAccess="false" />
+        <requestedExecutionLevel  level="requireAdministrator" uiAccess="false" />
+        <requestedExecutionLevel  level="highestAvailable" uiAccess="false" />
+
+            Specifying requestedExecutionLevel element will disable file and registry virtualization. 
+            Remove this element if your application requires this virtualization for backwards
+            compatibility.
+        -->
+        <requestedExecutionLevel level="asInvoker" uiAccess="false" />
```
persn: I'm honestly a little unsure what this line does, it was like that by default when the file was generated. I guess it needs more research.

ivan-mogilko: As far as I remember, this is the user permission level required to run the program. AsInvoker is a lowest one.
You need to change this if your program needs access to some administration tasks.

persn: So it's fine as it is? I don't have to change anything?

ivan-mogilko: Yes, I think it should be fine, as Editor is not supposed to have any advanced rights.

ivan-mogilko: BTW, speaking of which, there is a related issue: game templates are still saved in the program directory, which means AGS cannot create one if installed in C:/Program Files/, unless run "as administrator".

I believe that this should be changed at some point.

ivan-mogilko:
To clarify, was this tested under OS in question? I won't be able to do this at the moment.

persn:
It was tested on Windows 10, but I don't have Windows 8.1 available for testing. I'm also not sure if this affect any other OS but it shouldn't. I can later do a check to verify if Ubuntu has been affected or not.

ivan-mogilko:
No, I just wanted to be sure that this actually returns correct string on Windows 10.

Editor/AGS.Editor/app.manifest
```diff
@@ -0,0 +1,76 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
```
ivan-mogilko: Something I missed at first; should this string actually have AGS Editor name and version in it?

persn: Mmm, that's probably for the best, I don't know if there is some way we can automate updating version number however. I imagine doing it manually gets old really fast.

ivan-mogilko: We have a script for that. See version.json and Script/push_version.py. Later must be modified to also update manifest, but that may be done as a separate task.

ivan-mogilko:
Say, if you have time, could you also check the code that gets information about Framework version?

Right now it queries registry for that, and also does not detect anything above .NET 4.0. Do you know any portable way instead?

Editor/AGS.Editor/app.manifest
persn: I'll update this PR so that assemblyIdentity shows correct initial values then

persn:
Yes I did actually notice while I was working on this that the function checking for .NET version was filled with funky hard coding for reg values. I'll write it down on a post-it note and see if I can find a fix for it sometime later. I don't think it needs or should be included in this Pull Request however.

Without having researched it I'm guessing there's also a good chance I can't do it right now anyways until #395 and related issues are resolved. A lot of the Windows specific code in the editor exists purely because there doesn't exist a cross platform solution in the .NET v2.0 API.

persn:
Updated according to feedback

PR #399, opened 2017-04-18
Hotfix/get-dotnet-runtime-version

persn:
Detailed description in individual commit messages

### DIFF
--- a/Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
+++ b/Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
@@ -11,9 +11,11 @@ namespace AGS.Editor.Components
     {
         private const string STATS_REPORT_URL = @"http://www.adventuregamestudio.co.uk/clientstats.php";
         private const int STATS_SEND_INTERVAL_IN_DAYS = 30;
+
+        private readonly int _screenWidth;
+        private readonly int _screenHeight;
+
         private Timer _timer;
-        private int _screenWidth;
-        private int _screenHeight;
 
         public StatisticsSenderComponent(GUIController guiController, AGSEditor agsEditor)
             : base(guiController, agsEditor)
@@ -22,7 +24,7 @@ namespace AGS.Editor.Components
             // retrieve them while the game is being tested full-screen
             _screenWidth = System.Windows.Forms.SystemInformation.PrimaryMonitorSize.Width;
             _screenHeight = System.Windows.Forms.SystemInformation.PrimaryMonitorSize.Height;
-            _timer = new Timer(new TimerCallback(timer_Callback), null, 120000, 240000);
+            _timer = new Timer(Timer_Callback, null, 120000, 240000);
         }
 
         public override string ComponentID
@@ -30,7 +32,7 @@ namespace AGS.Editor.Components
             get { return ComponentIDs.StatisticsSender; }
         }
 
-        private void timer_Callback(object parameter)
+        private void Timer_Callback(object parameter)
         {
             try
             {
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
The code:

> string resolution = $"{_screenWidth}x{_screenHeight}";

Is not recognized by VS2008 (or perhaps .NET 2.0.). Could you change this to lower standard for now so that I could use this fix in nearest editor release?

persn:
Updated from feedback

ivan-mogilko:
Sorry, there is another occurance of same thing I did not immediately notice:

> return $"{runtime} {version}";

persn:
Updated

ivan-mogilko:
I am terribly sorry again. It looks like there are lots of features that 2.0 was lacking, because I found another one, which is a lambda expression if I am not mistaken?

> public override string ComponentID => ComponentIDs.StatisticsSender;

After fixing it to 

> { get { return ComponentIDs.StatisticsSender; } }

(I hope I got right idea)

I was finally able to compile this.

ivan-mogilko:
There is something else, actually, besides the compilation issues.

Original code was intended to know which *maximal* .NET version is installed in system. If I remember correctly, this was to let developer(s) know whether it may be safe to upgrade to higher frameworks.

The new code instead returns the one used to run the program.

persn:
> Original code was intended to know which maximal .NET version is installed in system. If I remember correctly, this was to let developer(s) know whether it may be safe to upgrade to higher frameworks.

Well that complicates things an awful lot. No I don't think there's a clean way to do this. The only thing I can think of would be to do registry checks on Windows, if we want our code to work on Linux and OS X 
we would have to introduce multiple project configs and preprocessor directives to avoid using conflicts and whatever else.

That being said though, I don't think it matters. The real question for me becomes, "**Should** we obtain the maximal .NET version?" And I don't see a reason to do that, what's the point? Which .NET framework is used to run AGS should only be decided on which OS we want to support. The end user doesn't care if they have to upgrade the .NET framework, and there's a good reason for that, because the end user should barely be aware that the .NET framework was installed in the first place. In most software that require the .NET runtime it silently installs at pre-install, I don't know if AGS does that but it should.

persn:
Updated.

This time I set the compiler to fail on C# Language Level > 3.0 so I'll be flabbergasted if it doesn't work. Apologies for the inconvenience, I knew very well I was introducing these features as I was writing them, I just assumed they would be merged after the upgrade.

ivan-mogilko:
In such case, does it make any use to send framework version in the statistics then? (Since we know what .NET is required by the Editor)

I do not know much about how >4.0 frameworks work, can there be a case when a higher version will run Editor that requires lower version?

persn:
> In such case, does it make any use to send framework version in the statistics at all? (Since we know what minimal .NET is required by the Editor)

That's a good point actually, I can't think of a reason to provide this data. I guess it might have made sense whenever this file was originally created.

We could just simply remove it.

> I do not know much about how >4.0 frameworks work, can there be a case when a higher version will run Editor that requires lower version? At the moment it reports 2.0 all the time, since Editor requires 2.0, and higher frameworks cannot run 2.0 assemblies.

Like I wrote in the commit message, .NET 4.0 and greater are all bundled together in some strange fashion, a side effect of this is exactly that. When I try to target 4.0 or higher on the AGSEditor project on my computer it will run on 4.6.1 as that is the highest runtime I have available. For any .NET version lower than 4.0 however it will do as you say because all those versions are distributed separately.

ivan-mogilko:
> Like I wrote in the commit message, .NET 4.0 and greater are all bundled together in some strange fashion, a side effect of this is exactly that. When I try to target 4.0 or higher on the AGSEditor project on my computer it will run on 4.6.1 as that is the highest runtime I have available. For any .NET version lower than 4.0 however it will do as you say because all those versions are distributed separately.

Well, I guess in such case this information may eventually be useful for statistics purposes.

persn:
Based on what I just said I guess we could say this code DOES return highest framework version installed, but only for v4.0 and above, so that problem will resolve itself when we upgrade to 4.0.

PR #401, opened 2017-04-18
Upgrade Windows project to VS2015

ivan-mogilko:
This upgrades all solutions and projects in AGS from VS2008 to VS2015, and also changes .NET Framework version editor uses from 2.0 to 4.0.

Other platforms should not be affected by this change.

I think we will also need to create VS2015 projects in our allegro and alfont patched branches to comply with this change, to make it easier for potential collaborators.
Just in case someone wants to test compiling it without having to deal with libraries, here is an archive with precompiled libs that AGS needs to link to. They are required only if you are building C++ components of Editor and engine:
https://www.dropbox.com/s/4p6nw6waqwat6co/ags-prebuilt-libs.zip?dl=0

You would still need to have all library headers, and install proper DirectX SDK, of course. [Please refer to the readme](https://github.com/adventuregamestudio/ags/blob/master/Windows/README.md).

**Important note:** I would like to first split out a branch for **v3.4.1** release before merging this change, that will be the last Windows version to be built in VS2008.


persn:
You'll have to forgive me for my lack of knowledge in C++, but I thought I'd take a crack at trying to compile "AGS.Editor.Full" in VS2017. I didn't go great to be honest, and I can't make heads or tails of the error log, what am I doing wrong? Target Platform version is v10.0 and Platform Toolset is v140.

[VS2017.txt](https://github.com/adventuregamestudio/ags/files/935275/VS2017.txt)

ivan-mogilko:
Ohhh. That is something I experienced recently myself.

This is what happens if you have old DirectX SDK headers included prior to newest Windows SDK. I had this when trying to build Allegro library, which I will need to fix for VS2015 upgrade,, but your case looks different, because you are compiling AGS.Native.

I'd check the C++ include paths list and make sure DirectX is included after Windows SDK.

persn:
There's definitely something funky with my setup because I adjusted it a little and now I have the same problems originating from Common.Lib and AGS.Native at the same time.

Just for clarification, does this look wrong?

ivan-mogilko:
Yes, it does look like a source of problem, in the first list your have DirectX SDK prior to Windows Kits, I think it should go after it, although I do not know how exactly did it get into that list.

ivan-mogilko:
Alright, I found what's going on. Modern Windows SDK already necessary contains DirectX headers in it, so you need to completely remove DirectX includes directory from the path (but not the library path!).

persn:
Ah, it worked I'm making progress. Depending on your findings the Windows build instructions should probably be updated.

I have a new problem however : /

ivan-mogilko:
Hmm, in the past kind of error occured because AGS.Native required very specific CRT version installed. But since you are building it yourself now... and probably AGS works well on your computer?

Are you absolutely sure that AGS.Native was built successfully?

persn:
It builds, produces a file that I can see and interact with, and VS says it was successful. I don't know what else to gleen from that.

[VS2017-2.txt](https://github.com/adventuregamestudio/ags/files/937302/VS2017-2.txt)

ivan-mogilko:
Could you download this tool, open AGS.Native.dll with it and check if it reports any DLL requirements missing:
http://www.dependencywalker.com/

persn:
Event Viewer Application Log had more detailed info

> Activation context generation failed for "E:\Programming\AdventureGameStudio\ags\Solutions\.build\Debug\AGS.Native.dll". Dependent Assembly Microsoft.VC90.DebugCRT,processorArchitecture="x86",publicKeyToken="1fc8b3b9a1e18e3b",type="win32",version="9.0.21022.8" could not be found

ivan-mogilko:
Oh, weird, it still asks for VC90 libraries... I will check what may cause that.

persn:
> What if you try building Release config?

It works.

Solutions/Engine.sln
```diff
@@ -1,12 +1,11 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 10.00
-# Visual Studio 2008
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
+MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Common.Lib", "Common.Lib\Common.Lib.vcproj", "{463A715C-3EF3-47C7-AC7F-D97660149C49}"
```
persn: I think this was missed in the upgrade process, Common.Lib.vcproj doesn't exist anymore

ivan-mogilko: Right, there is vcxproj instead, but solution update was not accurate perhaps.

AlanDrake:
I managed to get everything to compile, but the steps are not always clear:
- our allegro fork should have the patched branch as the default branch, I was fooled into downloading allegro5 for my wrong assumptions;
- allegro's headers will be missing "alplatf.h" if one is just using the precompiled libraries, there should be a mention about that;
- You mention to add paths to allegro headers and all the libs. Coming from the old system I find that somewhat annoying, I'd rather there were predetermined locations so I don't have to make changes to the solutions and just drop the required files into the designated places. (or were you already planning to do it with some submodules magic ?)

A starter kit, with all the libs+headers ready to extract over some folder, would greatly help newcomers.

P.S: what's with the Engine.sln, I thought you guys had it fixed with "vcxproj" instead of "vcproj", all else looks fine so far.

ivan-mogilko:
> allegro's headers will be missing "alplatf.h" if one is just using the precompiled libraries, there should be a mention about that;

Since Allegro 4.4 its authors suppose you to use Cmake to build an MSVS solution. When you do, it also generates alplatf.h for the chosen system.
Because few ppl already mentioned that is confusing, I was planning to create MSVS solution for the library right in the patched branch, which should solve the trouble.

> You mention to add paths to allegro headers and all the libs. Coming from the old system I find that somewhat annoying, I'd rather there were predetermined locations so I don't have to make changes to the solutions and just drop the required files into the designated places.

There is a controversy on this. Personally, I prefer the solution to not have a predesignated path in it, but instead let you to setup your enviroment, either by adding to VC paths in IDE, or define enviroment variables, like $(ALLEGRO_PATH), etc.

I like this approach becayse one may have multiple projects and even IDEs to use same headers and libs, so I try to keep clones of libs to minimal. Also this lets me have a central place for all libs on disk.

> what's with the Engine.sln, I thought you guys had it fixed with "vcxproj" instead of "vcproj", 

This is a mistake made during upgrade, I overlooked it. I will be fixing this an other things found soon.

ivan-mogilko:
Update:
* Rebased the branch;
* Fixed Engine.sln
* Set default branches in ags/allegro and ags/alfont repositories to patched branches.
* Fixed alfont's output name setting for one config.
* Reuploaded prebuilt libraries. I *think* there might have been a mistake with debug version of alfont library, that could cause the dependency issue that @persn had, but I am not fully certain, because I could not reproduce that problem myself.
https://www.dropbox.com/s/4p6nw6waqwat6co/ags-prebuilt-libs.zip?dl=0

TODO:
* Create MSVS solutions for Allegro lib (probably both VS2008 and 2015 just in case)
* Create VS2015 solutions for alfont lib.
* Update Windows/Readme.md instructions.

persn:
They new pre-built libraries did the trick. I was able to build everything without much effort.

persn:
You seemed to have rebased this branch on f1550d7 which is a WIP in #397, was this intentional?

VS2015 throws all user files in a directory ".vs", consider including that into the .gitignore

ivan-mogilko:
> You seemed to have rebased this branch on f1550d7 which is a WIP in #397, was this intentional?

No, I was testing that new icon and occasionally pushed it to "master" on my personal repository. I will rebase again shortly.

ivan-mogilko:
Upd:
* rebased again;
* cleaned up .gitignore (new commit);
* added VS2008 and 2015 solutions to our Allegro and Alfont forks;

AlanDrake:
Now that the editor is compiled with 4.0, in order to play audio, we should adjust the following:
- remove "irrKlang.NET2.0.dll", replace with "irrKlang.NET4.dll" and update references
- replace "ikpMP3.dll" with the 4.0 compiled lib
The 4.0 compiled libraries can be found [http://www.ambiera.at/downloads/irrKlang-32bit-1.5.0.zip](here)

ivan-mogilko:
I'd leave that for other pull request maybe...

persn:
I wouldn't tamper with that stuff in this Pull Request unless the audio playback is acting weird. Note also that if we ever want to have a cross platform Editor irrKlang needs to be replaced with something entirely different, as the dotnet wrapper is made with the C++/CLI project type which Mono doesn't like.

AlanDrake:
As you wish, but note that you'll get a "Mixed mode assembly" error trying to play audio on the editor, since the current ones are built for 2.0. There is a workaround to make them work anyway, but I see no reason not to update.

persn:
Yes, I tried it now, it throws an exception when attempting to load the audio file, so this needs to be addressed in this Pull Request after all.

ivan-mogilko:
TBH I do not like putting binaries into repository. Could we just provide download links?

persn:
Yes, that should be alright, for now. I'm not a fan of binaries in the repository either.

ivan-mogilko:
Update:
* removed irrKlang for .NET 2 binaries, replaced assembly references with irrKlang for .NET 4, plus added a small readme into References folder with hints which libs you need.
* updated Windows/Readme.md

persn:
There's a new file "AGSEditor.exe.config" in the .build folder which should be included in the installer.

ivan-mogilko:
I found installer script includes "Source\*", which makes "Source\AGSEditor.exe" unneeded, but should also include anything else there.

persn:
Alright, then this PR looks good to me

ivan-mogilko:
Well... the Rubicon crossing, as they say.

PR #402, opened 2017-04-19
Editor: Show modern Windows versions in About

gurok:
Added support for Windows 8, Windows 8.1 and Windows 10 (and onwards?) detection to the about dialogue box.

I wanted to post about this on #398 , but it got merged before I got a chance to comment. The addition of a manifest file allows us to get the correct values for the about dialogue box. I have updated the code to populate the relevant field according to this table:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832%28v=vs.85%29.aspx

### DIFF
--- a/Editor/AGS.Editor/GUI/AboutDialog.cs
+++ b/Editor/AGS.Editor/GUI/AboutDialog.cs
@@ -59,16 +59,25 @@ namespace AGS.Editor
                 }
                 else if (Environment.OSVersion.Version.Major == 6)
                 {
-                    osName = "Windows Vista";
-
-                    if (Environment.OSVersion.Version.Minor > 0)
+                    switch (Environment.OSVersion.Version.Minor)
                     {
-                        osName = "Windows 7";
+                        case 3:
+                            osName = "Windows 8.1";
+                            break;
+                        case 2:
+                            osName = "Windows 8";
+                            break;
+                        case 1:
+                            osName = "Windows 7";
+                            break;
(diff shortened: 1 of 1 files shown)


persn:
Maybe consider adding case for Linux and OS X while you're in there. https://msdn.microsoft.com/en-us/library/3a8hyw88(v=vs.110).aspx

ivan-mogilko:
Why there is no way to get a string from the framework? That seems silly  :-/.

persn:
¯\_(ツ)_/¯

ivan-mogilko:
I could actually merge this pull request as it is, since, if I got that right, Editor does not work on Unix systems yet. This fix will be used in AGS 3.4.1 then, and later updates could be made for versions that support running under Unix (whenever you make that happen).

persn:
Sure

gurok:
I'm fine with leaving it till there is an actual Unix port of the editor -- I think that makes the most sense.

I was ready to accommodate persn's request, but I am doing so around my work commitments right now. Perhaps next time I'll post in the pull request "I will do this", just to let you guys know I'm looking at it. There seem to be large spans of time between when I can look at AGS-related dev things.

ivan-mogilko:
We have a label "pull: wip" made for situation, you just put it on your request, and that will indicate that it's in progress.

gurok:
Ah yes, sorry. I keep forgetting. I suppose I should've done that.

PR #404, opened 2017-04-27
Refactor GUI Control classes

ivan-mogilko:
I am awfully sorry for doing this (sort of), but it was too hard to "let it go" when you have an almost ready piece of work that took you many days to do. So, I decided to recreate the (almost 2 years old) branch which refactors GUI Control classes.

Luckily, there were only several conflicts to resolve (the related code did not change much since the old branch was first made), and I found and fixed only couple of minor mistakes so far.

As this is refactoring, the new code is supposed to be functionally similar to an old one. Classes are tidied, and brought to general code style; some optimization is done along the way.

Notable addition is an introduction of generic FrameAlignment flags which substitute varied alignment types in Buttons, Labels and ListBox.

The code in this branch was written with the idea to keep compatibility in all known ways; but I guess it won't be hard to remove if wanted. (Adjustments to how Editor saves GUI would be required too in such case)

### DIFF
--- a/Common/ac/interfacebutton.h
+++ b/Common/ac/interfacebutton.h
@@ -16,8 +16,6 @@
 #define __AC_INTERFACEBUTTON_H
 
 #define MAXBUTTON       20
-#define IBACT_SETMODE   1
-#define IBACT_SCRIPT    2
 #define IBFLG_ENABLED   1
 #define IBFLG_INVBOX    2
 struct InterfaceButton {
--- a/Common/core/types.h
+++ b/Common/core/types.h
@@ -82,6 +82,13 @@
 #endif // WINDOWS_VERSION
 
 
+// Suppress override keyword for compilers that do not support it
+// TODO: this should be reviewed if project would demand C++11 or higher
+#if !(defined (_MSC_VER) && _MSC_VER >= 1700 || __cplusplus >= 201103L)
+#define override
+#endif
+
+
 #define fixed_t int32_t // fixed point type
 #define color_t int32_t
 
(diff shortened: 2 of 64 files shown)


AlanDrake:
Nice, I'll give it a pull.

AlanDrake:
Looks good to me.

AlanDrake:
Sorry, I found something amiss. If you try to reset the game, you get a "Error: Error: Too many control events, need newer version"
Saving loading normally does not reproduce the error, only RestartGame()

ivan-mogilko:
That's GUIButton::Read/Write using non-matching methods to seralize text. I need to double check them and also other controls for same problem.

AlanDrake:
Yeah, just got around to fix it.
It was the Text and the missing reserved in32 on guibutton.cpp
```
    Text.WriteCount(out, GUIBUTTON_TEXTLENGTH);
    out->WriteInt32(TextAlignment);
    out->WriteInt32(0);
```

ivan-mogilko:
There's couple of more mistakes I found, I will be fixing them.

ivan-mogilko:
Hmm.. btw I have a strike of deja vu while fixing these bugs. Probably I've already tried to ressurect this branch before but did not finish for some reason.

ivan-mogilko:
Updated:
* Fixed button serialization
* Fixed list box serialization when savedgames are filled into it.
* Fixed graphic buttons with unremoved "New Button" text still displaying it in some cases (AGS has a compatibility requirement that this text is not displayed for graphic buttons).
* Corrected code style for non-public member variables.

Also, I realized that we cannot change alignment properties right away without also changing them in Editor, and hence modifying data format. So I made the conversion into generic type to be performed at the moment of drawing, and added TODO notes in engine and editor, suggesting moving to universal text alignment type for all GUI controls.

PR #405, opened 2017-04-29
Multi process compilation speeds up the build

AlanDrake:
I just enabled the option for multi process compilation,
It cut down the engine rebuild time from 1m32s to just 40s.

### DIFF
--- a/Editor/AGS.Native/NativeDLL.vcxproj
+++ b/Editor/AGS.Native/NativeDLL.vcxproj
@@ -82,6 +82,7 @@
       <PrecompiledHeader />
       <WarningLevel>Level3</WarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <MultiProcessorCompilation>true</MultiProcessorCompilation>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Since this looks more like an option which relies on particular machine you are building on, I would rather look for a way to not make this a global project setting, but a per-user setting, using propery sheets.

AlanDrake:
It just launches more build processes if the CPU allows. It will auto turn off if an incompatible flag is enabled, like /Gm in debug mode.

rofl0r:
i was about to say the same, but since this is windows stuff that's probably the way to do it.
on linux, it would be up to the user to just run make -jX where X corresponds to the number of cores he wants to dedicate to the build.

ivan-mogilko:
> It just launches more build processes if the CPU allows. It will auto turn off if an incompatible flag is enabled, like /Gm in debug mode.

Yes, I understand, and it builds fine for me, but I was concerned that this may cause extra use of machine resources which may be unwanted for some users.

AlanDrake:
There's very little to research: https://msdn.microsoft.com/en-us/library/bb385193.aspx

ivan-mogilko:
I've read that. I meant reasearch ways to set it up on per-user basis instead of making it a global project setting.

rofl0r:
the article indeed talks about command line usage using "cl.exe", rather than hardcodign it into the sln file...

ivan-mogilko:
> the article indeed talks about command line usage using "cl.exe", rather than hardcodign it into the sln file...

I think that's just how MSVS documentation is built, it points to this article even if you command context help in the solution properties window.

ivan-mogilko:
As for hardcoding, there is project's configuration, and user configuration that overrides it. I was thinking that maybe it's better to leave this for user to set up instead of forcing this option into project file.

On the other hand, symmetrically there are ways to disable that option on user side too. So we may have this as a default for the project, and user may turn it off if it does not work well for them.

AlanDrake:
Well, I vote for having it as default. Faster compile times means happier developers.

ivan-mogilko:
Alright; I also found there is a separate setting in MSVS that allows to limit number of CPU cores for projects with /MP.

PR #406, opened 2017-05-05
Fixed C# project module "AGS.CScript.Compiler" so that it successfully builds in xBuild on *nix systems

persn:
Needs #396 to be resolved in order to test properly since AGS.Cscript.Compiler references AGS.Types

### DIFF
--- a/Editor/AGS.CScript.Compiler/CodeBlockCompiler.cs
+++ b/Editor/AGS.CScript.Compiler/CodeBlockCompiler.cs
@@ -154,7 +154,7 @@ namespace AGS.CScript.Compiler
 
             if (_source.NextIsKeyword(PredefinedSymbol.SetEqual))
             {
-                Expression setEqualTo = ReadExpression(false, PredefinedSymbol.Semicolon, PredefinedSymbol.Comma);
+                ReadExpression(false, PredefinedSymbol.Semicolon, PredefinedSymbol.Comma);
                 // TODO: check types, assign result to variable
             }
         }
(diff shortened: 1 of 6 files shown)


Editor/AGS.CScript.Compiler/CodeBlockCompiler.cs
```diff
@@ -269,7 +268,7 @@ private void GenerateCodeForExpressionWithoutOperator(Expression expression)
                 {
                     throw new CompilerMessage(ErrorCode.ParentIsNotAStruct, "'" + firstToken.Name + "' is not a struct");
                 }
-                memberName = reader.ReadNextToken();
+                ////Token memberName = reader.ReadNextToken();
```
ivan-mogilko: I cannot tell if this is correct. ReadNextToken may advance some iterator, thus commenting its call out may cause further  operations break.

I suggest leaving original variable declaration (above) instead, and doing something like this on line 271:

> reader.ReadNextToken(); // memberName

Editor/AGS.Editor/Components/HelpCommandsComponent.cs
```diff
@@ -325,10 +325,6 @@ private VersionCheckStatus CompareSoftwareVersions(string serverVersionText)
         private bool AdjustHelpWindowSize()
         {
             IntPtr helpHandle = Process.GetCurrentProcess().MainWindowHandle; // get the Help window handle
-            if (helpHandle == null)
-            {
-                return false;
-            }
```
ivan-mogilko: I wonder if Chris Jones meant to compare this with 0 (zero handle) instead?

ivan-mogilko:
I checked this out, and found just two changes that need to be reworked in my opinion.

Regarding unused _isLegacyString flag, I **do** know what that means, but since this compiler is not complete, I cannot tell how it was supposed to be used. If anyone would finish this in the future, they would have to address this issue anyway (assuming legacy strings still would be supported, which is questionable).

ivan-mogilko:
E: actually, I rejected one of my comments, noticing that it only affects local reader object.

persn:
Updated according to feedback and rebased on master.

PR #408, opened 2017-05-17
Fixed C# project module "AGS.Editor" so that it successfully builds in xBuild on *nix systems

persn:
Needs #406 to be resolved since "AGS.CScript.Compiler" is a requirement of "AGS.Editor"

### DIFF
--- a/Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
+++ b/Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
@@ -15,7 +15,9 @@ namespace AGS.Editor.Components
         private readonly int _screenWidth;
         private readonly int _screenHeight;
 
+#pragma warning disable 0414
         private Timer _timer;
+#pragma warning restore 0414
 
         public StatisticsSenderComponent(GUIController guiController, AGSEditor agsEditor)
             : base(guiController, agsEditor)
--- a/Editor/AGS.Editor/ImportExport.cs
+++ b/Editor/AGS.Editor/ImportExport.cs
@@ -277,8 +277,8 @@ namespace AGS.Editor
                         new Script("Module" + i + ".asc", moduleScript, name, description, author, moduleVersion, uniqueKey, false));
                 game.RootScriptFolder.Items.Add(scripts);
 
-                int permissions = reader.ReadInt32();
-                int weAreOwner = reader.ReadInt32();
+                reader.ReadInt32(); ////int permissions = reader.ReadInt32();
+                reader.ReadInt32(); ////int weAreOwner = reader.ReadInt32();
             }
 
             game.RootScriptFolder.Items.Add(new ScriptAndHeader(scriptHeader, globalScript));            
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -234,6 +234,7 @@ namespace AGS.Editor
 
 		public bool ApplicationStarted
 		{
+            get { return _applicationStarted; }
 			set { _applicationStarted = value; }
 		}
 
--- a/Editor/AGS.Editor/Audio/MidiPlayer.cs
+++ b/Editor/AGS.Editor/Audio/MidiPlayer.cs
@@ -24,7 +24,7 @@ namespace AGS.Editor
                 return false;
             }
             StringBuilder sb = new StringBuilder(100);
-            uint lRet = mciSendString(string.Format("open \"{0}\" alias track", clip.CacheFileName), sb, 0, IntPtr.Zero);
+            mciSendString(string.Format("open \"{0}\" alias track", clip.CacheFileName), sb, 0, IntPtr.Zero);
             mciSendString("set track time format ms", sb, 0, IntPtr.Zero);
             mciSendString("play track", sb, 0, IntPtr.Zero);
             _playingClip = clip;
--- a/Editor/AGS.Editor/Panes/DialogOptionEditor.cs
+++ b/Editor/AGS.Editor/Panes/DialogOptionEditor.cs
@@ -11,12 +11,9 @@ namespace AGS.Editor
 {
     public partial class DialogOptionEditor : UserControl
     {
-        private DialogOption _option;
-
         public DialogOptionEditor(DialogOption option)
         {
             InitializeComponent();
-            _option = option;
             chkSay.DataBindings.Add("Checked", option, "Say", false, DataSourceUpdateMode.OnPropertyChanged);
             chkShow.DataBindings.Add("Checked", option, "Show", false, DataSourceUpdateMode.OnPropertyChanged);
             txtOptionText.DataBindings.Add("Text", option, "Text", false, DataSourceUpdateMode.OnPropertyChanged);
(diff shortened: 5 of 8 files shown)


rofl0r:
why are unused variables errors ? should be warnings.

persn:
Because the .csproj config is enabled to treat warnings as error.

rofl0r:
well, maybe it's better than to turn that off ? -Werror is an anti-pattern http://blog.schmorp.de/2016-02-27-tidbits-for-the-love-of-god-dont-use-werror.html

persn:
That'll be an executive decision for someone else than me to make. @ivan-mogilko?

ivan-mogilko:
Hah. And I was relying on your judgement in regards to the C# project.

ivan-mogilko:
That's fine by me.
Warnings like unused variables are good to be fixed, but frankly they are not *that* critical in the Editor, because many of these seem to have been added "just in case" for the future.

persn:
Sure thing, I'll make a PR later tonight unless someone beats me to it

tzachshabtay:
I think "treat warnings as errors" is a very good configuration, actually. What I usually see with projects that don't treat warnings as errors is that the warnings start to pile up, and eventually you can't even see that your code created new warnings because they're hidden in all the mess.

That article posted above is less relevant in c# world in my opinion:
1. Compilers are much less fragmented.
2. Only CSC warnings are promoted to errors, MSBuild warnings stay as warnings. And all CSC warnings can be suppressed with `#pragma`s if needed.

rofl0r:
but apparently the compiler @persn uses is much more strict, so the effect is that he needs to go on a unused-variable-killing-spree to make the code compile, and that may actually be counter-productive (the variables may have been left there for a reason by the original author, sometimes they're the result of code thats only used with some #ifdefs).
Werror is a thing that makes sense if you turn it on temporarily to hunt down all possible bugs, but not something that should be permanently enabled.
/me paints the bikeshed blue

tzachshabtay:
Unused variables are bugs 99% of the time, usually harmless but sometimes can point to real bugs, so hunting for those is not counter-productive imo. And anyway, there's also the option of suppressing specific warning ids (i.e without doing pragmas, just marking "unused variable" as a warning that won't be promoted at the project level) without removing Werror.

rofl0r:
well, unused variables are sometimes just leftovers of previous code (in which case its fine to remove them, but usually only the author that did the change knows for sure), but sometimes are misdetected since the code using it was not compiled (due to an ifdef, in which case removing it just to shut up the compiler will break your compilation once somebody tries to compile with that define on) or there due to a not-yet or half-implemented feature; in which case removing it actively sabotages the author of the half-finished feature. in every case, never is an unused variable actually problematic or a "bug" as you label it.

rofl0r:
also, in case of @persn and other contributors this behaviour is simply annoying and unfriendly: i suppose he just wants to get the code compiling on his platform, not wasting hours to go hunting down unused variables, and filing pull requests for them just because someone insisted that Werror must be on for everybody. why don't **you** enable Werror, and then "fix" all the unused variables instead ?

persn:
Maybe continue the discussion of Werror in #409, there's a lot of posts here now that is only vaguely related to this pull request at best.

Editor/AGS.Editor/ImportExport.cs
```diff
@@ -277,8 +277,10 @@ public static CompileMessages ImportOldEditorDatFile(string fileName, Game game,
                         new Script("Module" + i + ".asc", moduleScript, name, description, author, moduleVersion, uniqueKey, false));
                 game.RootScriptFolder.Items.Add(scripts);
 
-                int permissions = reader.ReadInt32();
-                int weAreOwner = reader.ReadInt32();
+                ////int permissions = reader.ReadInt32();
+                ////int weAreOwner = reader.ReadInt32();
+                reader.ReadInt32();
+                reader.ReadInt32();
```
ivan-mogilko: Could you appened commented name of variable to the actual working code instead like -
<code>
reader.ReadInt32(); // int permissions
</code>

Right now it is not immediately obvious whether commented ReadInt calls correspond to following ones, or supposed to be 2 separate reads.

Editor/AGS.Editor/DataFileWriter.cs
```diff
@@ -64,7 +64,6 @@ private static byte[] GetAnsiBytes(string text, int min_size, int max_size, int
             //
             // The output array will be at least min_size in size.
             byte[] bytes = new byte[Math.Max(min_size, Encoding.Default.GetByteCount(text) + reserve_bytes)];
-            int data_length = Encoding.Default.GetBytes(text, 0, text.Length, bytes, 0);
             if (max_size <= 0 || bytes.Length <= max_size)
```
ivan-mogilko: This is not right, the call to GetBytes fills in bytes array, that's the core functionality of the function.

Why data_length is not used is a curious case, may be a mistake. I will look into this.

Editor/AGS.Editor/DataFileWriter.cs
```diff
@@ -64,7 +64,6 @@ private static byte[] GetAnsiBytes(string text, int min_size, int max_size, int
             //
             // The output array will be at least min_size in size.
             byte[] bytes = new byte[Math.Max(min_size, Encoding.Default.GetByteCount(text) + reserve_bytes)];
-            int data_length = Encoding.Default.GetBytes(text, 0, text.Length, bytes, 0);
```
ivan-mogilko: This is not right, GetBytes fills in "bytes" array and is core call in this function, it must not be removed.

Why "data_length" is not used is an interesting question, could be a mistake. I will look into this.

Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
```diff
@@ -15,16 +15,14 @@ internal class StatisticsSenderComponent : BaseComponent
...
-
         public StatisticsSenderComponent(GUIController guiController, AGSEditor agsEditor)
             : base(guiController, agsEditor)
         {
             // Screen width x height must be cached here, otherwise we could attempt to
             // retrieve them while the game is being tested full-screen
             _screenWidth = System.Windows.Forms.SystemInformation.PrimaryMonitorSize.Width;
             _screenHeight = System.Windows.Forms.SystemInformation.PrimaryMonitorSize.Height;
-            _timer = new Timer(Timer_Callback, null, 120000, 240000);
+            new Timer(Timer_Callback, null, 120000, 240000);
```
ivan-mogilko: My knowledge of C# is not enough for this; is Timer guaranteed to complete and call callback function before it is disposed by garbage collector?

ivan-mogilko:
Added few comments.
The code in DataFileWriter.cs is mine, I'd rather check that out myself, because it looks like an actual bug to me.

Editor/AGS.Editor/Components/StatisticsSenderComponent.cs
persn: That's a really good question, I don't know enough of the nitty gritty details of the automatic garbage collector, in my mind it makes sense that it wouldn't dispose a value in memory that is still being used. Just to be safe I suppose we could disable the error throw here with a pragma statement instead.

persn:
Updated according to feedback and rebased on master

ivan-mogilko:
Sorry for delay, merged it now.

PR #409, opened 2017-05-17
Disable Werror on C# projects

persn:
As discussed in #408. The Werror property seems to have beends randomly set across projects and build configs so it was good we looked into this

### DIFF
--- a/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
+++ b/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
@@ -34,7 +34,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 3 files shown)


tzachshabtay:
@rofl0r continuing the discussion from #408 here. An alternative suggestion: remove werror as you suggest from the projects, but keep using it in the CI build. So you wouldn't need to worry about fixing other people warnings as you load the project using your favorite compiler, but you wouldn't be able to create new warnings for everyone.

rofl0r:
@tzachshabtay that sounds like a great idea!

persn:
Just out of curiosity, how do you configure the repo so that the C# projects uses Werror only on the build server? If this is more complex than originally imagined we could simply just toggle Werror so that it's active for Release builds only.

rofl0r:
maybe like this:
in CI script before build:
```sh
sed -i 's/TreatWarningsAsErrors>false/TreatWarningsAsErrors>true/g' foo.csproj
```

ivan-mogilko:
I apologize for delay, was busy with other matters.

Was there any conclusion on this?

persn:
If what @rofl0r says works, we can do that assuming no one disagrees.

PR #411, opened 2017-06-08
Linux: use nanosleep for CPU yield

morganwillcock:
This replaces the sleep function in the Linux platform driver, which is called by platform->YieldCPU() in the main game loop. In testing I'm seeing a CPU usage reduction of between 5% and 10% and no other side effects, having built against the AGS Allegro fork. The kernel has supported this for a long time so I think it's pretty safe to switch.


ivan-mogilko:
Sorry for delay, merged it now.

PR #413, opened 2017-06-18
New save format

ivan-mogilko:
Restored another of my older branches... this one creates a new save format.

**WARNING:** needs more testing.

The two main reasons for this change are:
1) existing save format is very badly organized, with related data often separated across the file, which makes it difficult to track what is saved and make changes.
2) existing save format partially reuses seralization routines from game loading/saving, and also some of the runtime structs were written awhole even if that was not required; because of this quite a lot of unnecessary data is written to savegame.

Accordingly, for the new format the primary goals are:
1) make it cleaner and have data grouped better;
2) save only objects and variables that may change at runtime (if I remember correctly, my older tests showed 20-40% less size of saves, depending on circumstances).
3) add extra checks to help with debugging faulty serialization.

Originally I had intention to create a universal storage format and spent some time on this, but later realized that I won't be able to implement that properly even if come with something good. Game objects are still very much unorganized in the code, and "perfect" serialization would require big refactory and overhaul of the game data structs, something that I was not intended to do now.
Considering this, and also the fact that I no longer have plans for the project, I decided to step back and make a smaller simplier change for the time being.

Aside from writing game data, the proposed serialization routine safeguards each part of a save with special tags, containing name of the component (and its individual format version). There are also additional tags splitting larger parts of save. Upon loading these tags are used to determine whether engine can read such component, and also to detect reading mistakes (tags must be found in certain places).

This branch is still capable of reading old save format. I left this opportunity just in case, but it may be easily removed, because old routines are completely decoupled from new ones.


AlanDrake:
What to do with this PR ? Did you plan to merge it to the final 3.4.1 release ? Is it safe enough to be pulled into the cleanup branch ?

If it's good enough and can help the cleanup process I should merge it soon, because on a later stage it might be too late.

ivan-mogilko:
The plan was to merge it to master, not to 3.4.1 (that's a too big change for that), but then I was off to vacation and forgot about this.

The code was restored from older WIP branch, but this still needs testing. Perhaps I could do this on the weekend.

ivan-mogilko:
Well, did some tests, fixed mistakes I found. Unless you want to give it a try yourself, I could merge it.

rofl0r:
@fuzzie's scummvm port could use a new save format too, because so far it has none...

AlanDrake:
I think you can proceed with the merge. 👍

AlanDrake:
I have found something amiss, savegame screenshots break the save loading.
It looks like it's expecting to find another tag while there's only the raw bitmap right after the </Plugin Data>.

ivan-mogilko:
> It looks like it's expecting to find another tag while there's only the raw bitmap right after the </Plugin Data>.

Oh, I think I know what's wrong. I'll be fixing that soon!

ivan-mogilko:
@AlanDrake, made pull request #421, please see if that works for you.

PR #415, opened 2017-06-24
Prevent crashes when dealing with unprintable symbols

gurok:
This relates to the bug reported here:

http://www.adventuregamestudio.co.uk/forums/index.php?issue=825.0

The behaviour of get_friendly_name did not exactly match the get_name it replaced.

There are some areas in the compiler where it assumes the next symbol is valid to print. These are errors in the compiler's logic, but whereas get_name would give the reasonable output "const (null)", get_friendly_name causes a crash.

Note: This was likely because sprintf dealt with a null c-style string pointer and output "(null)".

This new behaviour isn't perfect, but it's better than hard crashes when there are errors in the compiler logic. When there is a problem getting the name of a symbol, it will now be shown as "(invalid symbol)", and we can fix these issues as we find them.

I also added two EOF checks to address the specific cases mentioned on the bug report page.

### DIFF
--- a/Compiler/script/cc_symboltable.cpp
+++ b/Compiler/script/cc_symboltable.cpp
@@ -149,19 +149,27 @@ int symbolTable::find(const char*ntf) {
 }
 
 std::string symbolTable::get_friendly_name(int idx) {
-    std::string result = entries[idx & STYPE_MASK].sname;
+    std::string result;
 
-    if (idx & STYPE_POINTER) {
-        result = result + std::string("*");
-    }
+    idx &= STYPE_MASK;
+    if(idx > -1 && idx < entries.size()) {
+        result = entries[idx].sname;
 
-    if (idx & STYPE_DYNARRAY) {
-        result = result + std::string("[]");
-    }
+        if (idx & STYPE_POINTER) {
+            result = result + std::string("*");
+        }
 
-    if (idx & STYPE_CONST) {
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
I would like to clarify this just in case: your branch is called 341_eof_checks. I split 3.4.1 remaining development into release-3.4.1 branch, because master branch contains upgrade to VS2015 and some larger code refactoring, which I do not want to risk putting into 3.4.1 which is in late beta stage.

Just want to make sure you know that if this fix is merged to master it won't get into 3.4.1.

gurok:
I think I did this the wrong way. I will remake it against 3.4.1. Sorry. And thank you for the catch.

ivan-mogilko:
Wait, I just found github now supports changing target branch, under "Edit" button. I have some warning displayed when I switch it, but I will give it a try.

ivan-mogilko:
Wait, I just found github now supports changing target branch, under "Edit" button. I have some warning displayed when doing it, but I will give it a try. I guess it depends on what your branch was based on.

ivan-mogilko:
Hmm... look like it's working.

gurok:
Thanks! I don't know how I screwed up and ended up basing it on master, but my intention all along was to commit to 3.4.1. I'm working on an old copy because I haven't upgraded to VS2015 yet.

Issue #417, opened 2017-08-07
[Windows 10] Unable to get My Documents dir: GetShortPathNameW failed


rocifier:
In Windows 10, a message box displays every time I run a game created with AGS. This includes games I make from the editor 3.4.0 and exes. The error message is:

    Unable to get My Documents dir: GetShortPathNameW failed

This is related to saved games as seen [in this code](https://github.com/adventuregamestudio/ags/blob/19cab6c05af56624a257b2a456e226fce777cb51/Engine/platform/windows/acplwin.cpp#L578).

My use account has a My Documents directory. This error message doesn't seem to affect the game, and clicking OK resumes the loading of the game, and all the saved games work perfectly in each game.

rocifier:
Update: I have been looking through my file system, and I actually don't have a "My Documents" folder. The closest equivalent I can find is in C:\Users\XXX\Documents however it doesn't contain much of anything interesting so I wouldn't expect saved games to go there.

From my experience, app / game data usually should get stored locally under C:\Users\XXX\AppData\Local\AppName

Can anyone second this?

ivan-mogilko:
AGS saves to "My Documents" only on Windows XP, on Windows 7 and higher it saves games to %USERPROFILE%/Saved Games/<Name of the game>, or correspondent alternative. This is the path returned by FOLDERID_SAVEDGAMES, I think.

The problem could originate from CSIDL_PERSONAL returning something unexpected on Windows 10. Although, to be honest, this is the first time such problem is reported for Windows 10.

I will see if I can change the code so that it won't check for paths compatible with old Windows versions unless newer version was not detected.

rocifier:
I'm not sure what my alternative would be since I don't have a saved games folder. I will see what these paths return when I get home.

rocifier:
Looking at the code after your comment it is not doing precisely what you said. First it is getting My Documents dir every time as a fallback, and then trying to get the Saved Games dir after that.

monkey0506:
@ivan-mogilko stated, "I will see if I can change the code". Clearly the code hasn't been magically fixed retroactively.

monkey0506:
And "%USERPROFILE%/Saved Games" is a standard system path.

rocifier:
@ivan-mogilko are you still working on AGS then? I'd be happy to make a contribution. I just ask because I saw your announcement from April

ivan-mogilko:
[...]

PR #419, opened 2017-08-09
Fixed issue #417

rocifier:
"My Documents" folder was being retrieved on every version of Windows. Now it is only retrieved on XP and earlier, with "Saved Games" retrieved on later versions of Windows. There is also a fallback to store saved games within the data folder in case of failure.

### DIFF
--- a/Engine/platform/windows/acplwin.cpp
+++ b/Engine/platform/windows/acplwin.cpp
@@ -54,6 +54,7 @@ extern Bitmap *virtual_screen;
 #include <shlobj.h>
 #include <time.h>
 #include <shlwapi.h>
+#include <windows.h>
 #ifdef VS2005
 #include <rpcsal.h>
 #endif
@@ -538,12 +539,39 @@ void AGSWin32::PostAllegroInit(bool windowed)
 
 typedef UINT (CALLBACK* Dynamic_SHGetKnownFolderPathType) (GUID& rfid, DWORD dwFlags, HANDLE hToken, PWSTR *ppszPath); 
 GUID FOLDERID_SAVEDGAMES = {0x4C5C32FF, 0xBB9D, 0x43b0, {0xB5, 0xB4, 0x2D, 0x72, 0xE5, 0x4E, 0xAA, 0xA4}}; 
+#define _WIN32_WINNT_VISTA              0x0600
+#define VER_MINORVERSION                0x0000001
+#define VER_MAJORVERSION                0x0000002
+#define VER_SERVICEPACKMAJOR            0x0000020
+#define VER_GREATER_EQUAL               3
+
+// These helpers copied from VersionHelpers.h in the Windows 8.1 SDK
+bool IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WORD wServicePackMajor)
+{
+  OSVERSIONINFOEXW osvi = { sizeof(osvi), 0, 0, 0, 0,{ 0 }, 0, 0 };
+  DWORDLONG        const dwlConditionMask = VerSetConditionMask(
+    VerSetConditionMask(
+      VerSetConditionMask(
+        0, VER_MAJORVERSION, VER_GREATER_EQUAL),
+      VER_MINORVERSION, VER_GREATER_EQUAL),
+    VER_SERVICEPACKMAJOR, VER_GREATER_EQUAL);
+
+  osvi.dwMajorVersion = wMajorVersion;
+  osvi.dwMinorVersion = wMinorVersion;
+  osvi.wServicePackMajor = wServicePackMajor;
+
+  return VerifyVersionInfoW(&osvi, VER_MAJORVERSION | VER_MINORVERSION | VER_SERVICEPACKMAJOR, dwlConditionMask) != FALSE;
+}
+
+bool IsWindowsVistaOrGreater() {
+  return IsWindowsVersionOrGreater(HIBYTE(_WIN32_WINNT_VISTA), LOBYTE(_WIN32_WINNT_VISTA), 0);
+}
 
 void determine_app_data_folder()
 {
   if (win32AppDataDirectory[0] != 0) 
   {
-    //already worked it out
+    // already discovered
     return;
   }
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
First things first, using Windows 8.1 SDK is still not acceptable for us, because we still aim to support Windows XP (with latest SP3 update) for engine, and Windows 8,1 SDK does not work with XP.

There is already a similar check made in the same source file, using GetVersionEx. It may be picked out as a separate function, like:
<pre>
bool IsWindowsVistaOrGreater()
{
    OSVERSIONINFO OS;
    OS.dwOSVersionInfoSize = sizeof (OSVERSIONINFO);
    GetVersionEx(&OS);
    return OS.dwMajorVersion >= 6;
}
</pre>

[I've read that this function is getting deprecated](https://stackoverflow.com/questions/1963992/check-windows-version), and it returns incorrect values on some newer versions of Windows **with major version 6 or above**, but for current case it seems to work, because we only need to check whether major version is less than 6 or not. If in the future engine would need more precise detection, it may be substituted with contemporary method.


Other than that, the general idea of the patch looks fine. Few more points though.

1. There is an inconsistency in checking GetShortPathNameW return value, I think there may be a mistake in the second (non-vista) case.
2. Since you are completely rewriting function anyway, I suggest to not use functions like strcpy and strcat as unsafe, and either use snprintf or our engine's string class (String, with capital S).
3. Last, please use spaces as indentation. Because of all the old code the style varies across the program, but we were using spaces as a style for the new code in last several years.

rocifier:
Hi @ivan-mogilko if Windows 8.1 SDK is not compatible with Windows XP then why does it have support in that header for back to Windows 95? Actually we are just using a helper method in that header file which calls a long-standing windows api - we aren't linking to anything. I think it's much better to use these official functions instead of rely on a code snippet you know has problems with newer windows versions.

Unless you mean you want it to compile on Windows XP? In that case I'm at a loss, I don't see why that would be supported.

rofl0r:
> Unless you mean you want it to compile on Windows XP?

yeah, that's exactly the point @ivan-mogilko was raising.

ivan-mogilko:
I guess I was misinformed. The VerifyVersionInfoW it calls is supported by Windows XP, as noted in documentation.

rocifier:
When I loaded the solution for the first time, I was asked to select which Windows SDK and which platform tools version to use. Are you sure we aren't using them already elseware? It's pretty common developers will have some version of the SDK installed already.

If not, I guess we can extract the function from the header, though it's pretty low level code which isn't easily understandable.

ivan-mogilko:
> When I loaded the solution for the first time, I was asked to select which Windows SDK and which platform tools version to use.

Did you load it up in MSVS 2008 or upgraded to higher version? Solution in this branch is for VS2008.

TBH I am not sure if I had not installed this SDK on my computer before that, since I have VS2015 too, and it may be a part of its installation, but I am sure it is not used in the project at least this branch's version, because solution exists since before that SDK was released, and I had to add path to 8.1 SDK's include folder to VS2008's settings.

> If not, I guess we can extract the function from the header, though it's pretty low level code which isn't easily understandable.

Ah, well, there's some other low level code in there, frankly, this won't change much.
Since the master branch there was a big update, it uses VS2015 already, and may also use other contemporary SDKs if required (maybe it already does implicitly, as you supposed); and low-level code may be gradually removed from it if better alternatives exist.

For now, I am just about to create release candidate of AGS 3.4.1, and I really won't like to add new dependency to project at this stage.

rocifier:
I guess I upgraded my copy from VS2015 to VS2017, which was a trivial update. That must be why it asked me about the versions of those, but I can still go to the property pages for the project and see them set. As to whether they are being used, I also guess they are not. I am going to do another commit which includes that helper function in our code.

rocifier:
@ivan-mogilko I'm getting an odd build error in the 3.4.1 branch. Two errors:

Error (active)	E2474	user-defined literal operator not found	Engine.App	c:\Repositories\ags\Engine\test\test_inifile.cpp	33	

Error	C3688	invalid literal suffix 'ENDL'; literal operator or literal operator template 'operator ""ENDL' not found	Engine.App	c:\repositories\ags\engine\test\test_inifile.cpp	49	

Any ideas what's going on there?

monkey0506:
I don't know the specifics, but as a guess, C++11 introduced user-defined literals, which seems to be what the error is referring to. 'ENDL' on the other hand is probably supposed to be a macro defined as a string-literal. So:

#define ENDL "\n"
"Some text" ENDL

Would be mutated into "Some text\n" by the preprocessor. I'm guessing the definition for ENDL is missing (missing #include?)

rocifier:
It seems to be defined directly above in that file as "\r\n" due to building on windows. Hmm I'm not sure what is wrong with my compiler, I even tried redefining it as:

`constexpr const char* operator"" _ENDL(const char* val) { return "\r\n"; }`

But I still get the same error on those lines.

rocifier:
While I get this compile working, I have pushed my code changes in-case someone is able to test them

rocifier:
After excluding that particular test, I had to modify the debug configuration under Linker -> Input for Engine project to link to debug versions of all libraries (a bunch of them are trying to link to release versions):

    Common_d.lib
    odbc32.lib
    odbccp32.lib
    libcpmtd.lib
    alleg-debug-static-mt.lib
    alfont_mt_d.lib
    ddraw.lib
    dinput.lib
    dsound.lib
    dxguid.lib
    d3d9.lib
    quartz.lib
    strmiids.lib
    winmm.lib
    shlwapi.lib
    amstrmid.lib
    opengl32.lib
    libogg_static_d.lib
    libvorbis_static_d.lib
    libvorbisfile_static_d.lib
    libtheora_static_d.lib

ivan-mogilko:
I had a feeling that there is some misunderstanding going on. Now I seem to got it. Are you building release-3.4.1 in VS2017?

Take the error with ENDL you mentioned above, this is one of the issues I had to fix when we moved to VS2015. Obviously, in release-3.4.1 this is not fixed yet. Actually, I thought there might be more problems, but since you are using precompiled allegro etc libraries, you could have avoid them.

Anyway, I am going to test PR now, I'll be back in a short while.

ivan-mogilko:
On the PR itself. Could you rebase your commits on top of branch instead of merging them? It is a practice we use here, to not include "merge" commits in the PRs.

Also, could you check the two additional points I mentioned in my first comment?

1. There is an inconsistency in checking GetShortPathNameW return value, I think there may be a mistake in the second (non-vista) case.
2. Please use spaces as indentation. Because of all the old code the style varies weirdly across the program, but we were using spaces as a style for the new code in last several years.

rocifier:
That's fine if you've fixed that elsewhere then. I've just commented it out; those files won't be staged. The compile issues were mostly a matter of me learning and getting my head around things - I feel much more confident about those types of issues (CLR) now. I didn't see your other points, I'll make those changes now and rebase / squash my commits.

rocifier:
Done, let me know if I've missed anything. Good spotting on the mistake - I reversed an if statement there and forgot to reverse my conditional inside.

ivan-mogilko:
Thank you, this looks good now.

PR #421, opened 2017-10-25
Encase whole savedgame's component list in tags

ivan-mogilko:
Added tags which explicitly mark the beginning and the end of the components data list in the savedgame.
Components are now read until the end of list is found, as opposed to end of file.

This should resolve the error caused by screenshots with the new save format (https://github.com/adventuregamestudio/ags/pull/413#issuecomment-339395427), which are appended to the end of file.


AlanDrake:
It seems to be working, will do better testing tomorrow!

AlanDrake:
Alright, it works for all the cases I can try. 👍

PR #422, opened 2017-12-03
Stop reseting filesystem permissions when compiling

morganwillcock:
This removes the code which recursively gives every user access to the contents of the compilation folder, which should fix errors where `takeown.exe` is not installed (this is not a standard utility) and prevent other users and processes from modifying game data (I would class this as a security risk). If you do not have permission to write to this folder, this is really outside the scope of AGS, and the editor already correctly catches no write / read-only access to the destination.


ivan-mogilko:
I believe this makes sense.

I could be mistaken, but I guess this was an attempt to force Editor create hardlinks for non-administrator users. The problem is that some of the hardlinked files are located outside of game directory (in Editor's installation dir), so this solution did not work for all cases anyway.

PR #427, opened 2018-01-08
Custom color theme

persn:
This is my 3rd annual attempt at making color themes for AGS, hooray. It should be mostly self explanatory.

This pull request introduces a 3rd party framework by NuGet, Newtonsoft.JSON. XML was way too verbose for this job.

If anyone wants to test this they'll need the .json file I've been working on, I'll attach it to the post [VisualStudioDark.zip](https://github.com/adventuregamestudio/ags/files/1613027/VisualStudioDark.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,16 @@ Manual/ags-help.chm
 # Tools output and temporary files
 libsrc/*.tar*
 libsrc/*.zip
+
+
+# NuGet
+*.nupkg
+# The packages folder can be ignored because of Package Restore
+**/[Pp]ackages/*
+# except build/, which is used as an MSBuild target.
+!**/[Pp]ackages/build/
+# Uncomment if necessary however generally it will be regenerated when needed
+#!**/[Pp]ackages/repositories.config
+# NuGet v3's project.json files produces more ignorable files
+*.nuget.props
+*.nuget.targets
--- a/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
+++ b/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
@@ -117,6 +117,16 @@ namespace AddressBarExt.Controls
             set { this.selectedStyle = value; }
         }
 
+        public override Color BackColor
+        {
+            get { return ts_bar.BackColor; }
+            set { this.ts_bar.BackColor = value; }
+        }
+
+        public Color DropDownBackColor { get; set; }
+
+        public Color DropDownForeColor { get; set; }
+
         /// <summary>
         /// Gets/Sets the currently selected node. Validates upon set and updates the bar
         /// </summary>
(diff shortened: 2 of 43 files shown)


ivan-mogilko:
The first thing happened when I tan it was the unhandled System.IO.DirectoryNotFound exception when %USERPROFILE%/AppData/Local/AGS/Themes folder was not found.

I created directory and will continue testing.

ivan-mogilko:
A question: is there a reason which require Editor to be restarted to find newly added themes if they are put into Themes directory by hand?

If not, maybe reload the list of themes when Preferences dialog is open, or even when combo box is open (if that's not too time consuming).

persn:
Try now

ivan-mogilko:
Sorry for the delay.

Don't think I won't be able to provide much input regarding the code, because I don't have any experience in skinning UI. But it looks like generally a proper approach, and pretty customizable one.

The biggest doubts I currently have is that -
1) Every panel has to explicitly call "Factory.GUIController.ColorThemes.Load(LoadColorTheme);" in its constructor; IDK why is it necessary, but this may be bug-prone, because if anyone implements new panel they would have to not forget to do the same.
2) Every single panel having its own individual settings for literally every control.

This may do for starters, but perhaps there could be way to implement "default setting" for the control **type** in the future (i.e - default style for a Panel, for a Button, for a ListView, etc).

I'll test this a little more, and also would ask other ppl to give it a try.

ivan-mogilko:
I did a bit of testing with unexpected conditions, and found following may cause unhandled exceptions (including locking editor at splash screen):

1. There will be unhandled System.IO.IOException if you try to import file from "Themes" folder (file already exists).
2. Having syntax errors in theme files causes unhandled JsonReaderException at the editor's launch. Easiest way to reproduce is to put an empty .json file into Themes folder.
3. Having some of the values missing, e.g. some of the color component values not present, results in "System.ArgumentNullException: value cannot be null" at the editor's launch.
Way to reproduce: for example, delete 1-3 colour components from "background" parameter group.

persn:
Updated.

I believe we should think about where color themes are hosted? I'm not sure if uploading it to the main repo is very nice, maybe there could be a separate repo where users could contribute color themes.

ivan-mogilko:
This seem to work, but could you make it not report errors everytime user opens combo box?
What about adding some tag to the ones that have mistakes in the list, like "ERROR" or something.

persn:
To solve the problem I just made it so that Load and Init became two distinct steps. The user will get a error message when trying to start the editor with an invalid json file, but not anytime else.

There's no exception handling in the load step anymore however, shout out if you can think of a reason this would be a problem

ivan-mogilko:
@persn , I've just merged tzach's breadcrumbs GUI for the room editor, and it is now conflicting with your PR, could you rebase again please?

persn:
Rebased

ivan-mogilko:
Works now!

PR #428, opened 2018-02-10
Breadcrumb navigation bar for the Room editor

ivan-mogilko:
This adds a breadcrumb navigation bar in the Room editor, and support for locking and showing/hiding individual room items (objects, regions) or whole item "layers" at once.

This pull request is based on @tzachshabtay 's work done back in 2014, which was suspended for some reasons.
Original commit: https://github.com/tzachshabtay/ags/commit/45010f27e6b66ecf92c30787fb7d07ed6b636946

The examples of their look and usage may be seen on these screenshots (linking instead of embedding because of their large size):
http://i.imgur.com/dSAbGb3.png
http://i.imgur.com/Fk4YLkw.png
http://i.imgur.com/CPkUOwA.png

I authored new commit with Tzach's name, hopefully that's allright.

The code is practically the same, I did some cleanup, and notably removed buttons and the code intended for adding and deleting regions, because these operations are not very well devised in the current Editor.

If wanted these could be restored later, but there are problems to keep in mind:
1) AGS engine still has a fixed limit of room regions and does not have a concept of actual "region count", only "max regions". In theory, it is possible to change the way Editor works with the regions without changing the engine: let users to add/remove areas within the supported range, and fill in missing slots with dummy areas when compiling game data. Best to keep engine and editor consistent though.
2) Regions are based on a bitmap mask. Adding and deleting regions at will seem inconsistent with the mask importing, which may contain different number of regions.

Perhaps I am overthinking this, but as I said, that should be possible to restore add/delete operations later, or design more convenient ways to work with regions.

The primary purpose of this new GUI control is to provide quicker navigation and ability to show/hide particular items in any combination.


ivan-mogilko:
I gave it a test, and **technically** it is working. Might need some polishing and fixing usability issues later though. Guess I'll merge it now.

Issue #35, opened 2012-11-29
Discussion: Status of the master branch


tobihan:
Hi,

I did some testing with the master branch on 64 bit and it seems like some games don't crash anymore at all. There still seem to be some regressions though, especially with older games, see issues 33 and 34.

The detection of game files when only providing the folder does not work reliably yet, for instance when the folder name is given in quotes because it contains spaces. Barn Runner 1 Eclair 1 is not found even when the full path does not contain spaces and no quotes are used.

PR #430, opened 2018-02-14
updates the Android port to use ndk-bundle/platforms/android-14

ericoporto:
pull request for building AGS Engine for Android 

Forum Thread here: http://www.adventuregamestudio.co.uk/forums/index.php?topic=44768.msg636581163#msg636581163

My environment is Ubuntu 16.04 and I used the Android Sdk and ndk installed through Android Studio 3.0.1.

    uname -a
    Linux  4.13.0-32-generic #35~16.04.1-Ubuntu SMP Thu Jan 25 10:13:43 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

details of my Android Studio : 

    Android Studio 3.0.1
    Build #AI-171.4443003, built on November 9, 2017
    JRE: 1.8.0_152-release-915-b01 amd64
    JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
    Linux 4.13.0-32-generic

Commands I used to build:

    export PATH=$PATH:~/Android/Sdk/ndk-bundle/
    cd git/ags/Android/buildlibs/
    ./makestandalones.sh 
    ./buildall.sh 
    cd ../library/
    ndk-build

### DIFF
--- a/Android/buildlibs/armeabi-v7a/ndkenv
+++ b/Android/buildlibs/armeabi-v7a/ndkenv
@@ -5,7 +5,7 @@ get_abs_path() {
 
 unset DEVROOT SDKROOT CFLAGS CC LD CPP CXX AR AS NM CXXCPP RANLIB LDFLAGS CPPFLAGS CXXFLAGS
 
-export NDK_PLATFORM_ROOT=~/Android/Sdk/ndk-bundle/platforms/android-14/arm
+export NDK_PLATFORM_ROOT=$NDK_STANDALONE/arm
 export NDK_PLATFORM_NAME=armeabi-v7a
 export NDK_HOST_NAME=arm-linux-androideabi
 export NDK_ADDITIONAL_LIBRARY_PATH=$(get_abs_path "../../nativelibs/$NDK_PLATFORM_NAME")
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
On thing that I want to clarify, since I know little about developing for Android, does using higher version of NDK affect the range of supported devices?

ericoporto:
Yes. The platform-9 supported previously added support from Android-2.3 onwards. Going to platform-14 supports Android-4.0 onwards. (I can't find better references than [this doc](https://developer.android.com/about/dashboards/index.html) and [this stackoverflow](https://stackoverflow.com/questions/45685227/android-ndk-support-version-limited#) 
 )

ivan-mogilko:
Ok, you also have following modification:
<pre>
 # android-9 is minimum to support mips and x86
-PLATFORM=android-9
+PLATFORM=android-14
</pre>

So, this will disable running engine on some older devices?

I'd rather find out if that's possible to build these without changing platform. It would also be the good thing to get someone who actually knows a thing about Android to check this out too.

ivan-mogilko:
Alright, I was able to complete nativelibs compilation.
In the light of the information discussed in #431, we should be switching to android-14 platform, so that's okay.

I would like to propose number of changes to this pull request though.

### For the makestandalones.sh

* remove duplicate of hard-coded path when declaring NDK_STANDALONE, do $NDK_HOME/platforms/$PLATFORM instead.
* add **--verbose** flag to the make-standalone-toolchain.sh call, because it is not possible otherwise to understand why making toolchain failed.
* fix comments in the script to comply with the changes. For instance, you change PLATFORM=android-14, but the comment above it still sais "android-9 is minimum to support mips and x86".

### For the library building scripts

* First of all, I suggest exporting NDK_STANDALONE in the buildall.sh, similarily to how it is exported in makestandalones.sh, then use $NDK_STANDALONE everywhere in ndkenv scripts instead of hard-coding the actual paths everywhere.

What I mean, in **buildall.sh**
<pre>
$NDK_HOME=< ndk path >
$PLATFORM=android-14
export $NDK_STANDALONE=$NDK_HOME/platforms/$PLATFORM
</pre>

Now in every ndkenv you could use $NDK_STANDALONE, like:
<pre>
NDK_PLATFORM_ROOT=$NDK_STANDALONE/arm ... mips .. x86 etc
</pre>

* Also, in the ndkenv, let's avoid duplicating hard-coded paths even more by using $NDK_PLATFORM_ROOT when declaring CC and CXX variables, e.g.:
<pre>
export CC=$NDK_PLATFORM_ROOT/bin/arm-linux-androideabi-gcc
</pre>

* There is an inconsistency with how __ANDROID_API__ compilation flag is declared. For ogg you add the flag into ogg.sh scripts, but for lua - into liblua.patch, although you could probably add it to lua.sh too, into MYCFLAGS.

Alternatively, I found it's possible to add this flag into NDK_CFLAGS variable exported in ndkenv scripts. I frankly do not know if that's the best place to add it, but it worked for me (you just need to have it declared in 4 ndkenv scripts, and that will work for both ogg and liblua, and anything else that requires this API flag).

Would you be able to ammend these changes to your commit? If not I could make second patch over/instead yours myself.

ericoporto:
Hey @ivan-mogilko , I tried to comply with your proposed changes. The flag **`-D__ANDROID_API__=\$API`** , I moved to the NDK_CFLAGS in ndkenv scripts. Please take a look.

ivan-mogilko:
Alright, merged!

ivan-mogilko:
I just realized that this pr possibly had a mistake, it changes "Android/library/project.properties" setting target platform to android-14, but this file and setting has nothing to do with compiling native libraries. It is related to building java app. "Android/library/AndroidManifest.xml" also has targetSDKVersion set to 16.
We need to change project.properties back.

PR #433, opened 2018-03-10
Apply Editor message box preferences to 'successful compilation'

morganwillcock:
Previously the Editor message boxes preference didn't apply to the 'successful compilation' message box.
This makes the following changes:

- adds a default enum 'Always' (previous default 'WarningsAndErrors' was already allowing success messages)
- suppress the 'successful compilation' message box unless using 'Always'
- write the successful compilation message to the output pane

The main benefit being, if you set the option to 'Never', you now get faster testing. As an added bonus, this stops the Windows sound effect that accompany message boxes being played, every time the game is compiled.

Potential problem: With preference set to 'Never' and then reverting to an older AGS version, the index is out of range. I'm not sure if there is any safer way to extend a preference item with a new value? At the moment it seems existing preferences are not validated.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1020,9 +1020,9 @@ namespace AGS.Editor
 			}
 			else if (errors.Count > 0)
 			{
-				if (_applicationSettings.MessageBoxOnCompile == MessageBoxOnCompile.WarningsAndErrors)
-				{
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
> Potential problem: With preference set to 'Never' and then reverting to an older AGS version, the index is out of range. I'm not sure if there is any safer way to extend a preference item with a new value? At the moment it seems existing preferences are not validated.

I think the only way to make it work is to save under new key (e.g. "MessageBoxOnCompileErrors2").

morganwillcock:
Now rebased for newer application settings.

PR #435, opened 2018-03-14
Reset Enum preferences which are undefined

morganwillcock:
This fixes where a successfully cast enum preference is out-of-range.


ivan-mogilko:
If I understand correctly, this is related to problem mentioned in #433 (when preference file has a value unknown by current version)?

I am thinking, is there a way to make this fixup more automated and noticeable at the same time, so that future contributors, if any, could see and use it?

For example, if there were a method that would combine a call to Convert.ToInt32(key.GetValue(...)) with setting default value in case conversion resulted in undefined enum value.
E.g. (pseudo code, since I do not remember generic syntax 100% well) -
<pre>
T ReadEnum<T>(object value, T default_value)
{
     T e = (T)Convert.ToInt32(value);
     if (!Enum.IsDefined(typeof(T), e))
          return default_value;
     return e;
}
</pre>


**PS**. On a side note, is that Registry actually a Windows registry (big binary thing)? Never paid attention, and thought preferences are saved to XML similar to other settings...

morganwillcock:
I had given a quick go to running everything through a function, but it isn't entirely straightforward as all the registry entries have been stored as strings (REG_SZ). So it probably needs a load of exception handling or TryParse-ing around all of initial conversions (at the moment it will throw exceptions where what is in the registry doesn't convert or is too large for Int32).

Would it be easier to define some sort of structure at the top, which maps the preference name, variable type, and fallback value? At the moment the defaults are repeated everywhere, and it is probably easier to just rewrite the whole thing. Migrating settings to something else (like XML) would solve the issue of incorrect registry values crashing older versions; perhaps there is already a type safe XML loader as part of the editor?

ivan-mogilko:
> I had given a quick go to running everything through a function, but it isn't entirely straightforward as all the registry entries have been stored as strings

Well, my proposal was simply to merge Convert.ToInt32 and undefined enum check, for now, in case future contributors would be adding new preferences of enum types, so that they could keep things consistent.

But if you really want to go that way, saving game project and additional project-related user settings are serialized using SerializeUtils static class that provides methods for writing and reading XML. It uses reflection to parse public object's properties. Not fully safe atm either, as for some reason it does not implicitly skip obsolete variables, and I think it also throws exception if it cannot convert enum value (so not really forward-compatible either).
We were adding additional attributes to work around some things for backward compatibility, for example DeserializeConvertValueAttribute that allows to rename enum value on deserialization (if old value is no longer supported). I believe similar method could be added to support forward compatibility (it just that forward compatibility was not needed for game projects so far, since they usually contain whole new features).

morganwillcock:
Any objection to trying it with the built in [.NET user settings](https://docs.microsoft.com/en-us/dotnet/framework/winforms/advanced/using-application-settings-and-user-settings)?

ivan-mogilko:
> Any objection to trying it with the built in .NET user settings?

I cannot say anything, because never used these.

But that's just a way to store settings, or will it change anything in regards to converting values?

morganwillcock:
I think that anything which can be serialised can be stored, and the serialise and de-serialise implemented with a custom TypeConverter where you need custom checks. With the default behaviour it looks to be strongly typed.

ivan-mogilko:
Alright, what are your intentions in regards to this pull request, would you like it merged as is for now, or you're going to remake the settings?
This question also relevant to #433, if I understood right it may break loading settings by previous versions of an Editor?

morganwillcock:
I've had a go at switching to the built in application settings, but it is currently missing the bits to migrate registry settings across. By the look of what is done so far, does the direction look okay to you? If you think this is a mistake, I'll look at another way.

Yes, #433 would make older versions crash, so that needs to be on hold for the moment.

morganwillcock:
I should probably also mention, you can 'upgrade' the settings with the program version, and also request settings from previous versions. Probably not useful now, but hopefully the next times settings are changed this mechanism gives some more flexibility. 

[https://msdn.microsoft.com/en-us/library/system.configuration.applicationsettingsbase.upgrade.aspx](https://msdn.microsoft.com/en-us/library/system.configuration.applicationsettingsbase.upgrade.aspx)

The settings go into local appdata as XML, named after the company and product, which is why I changed 'AssemblyCompany' to 'AGS Team'.

ivan-mogilko:
> By the look of what is done so far, does the direction look okay to you?

I'd really rather leave the judgement to you, or other people who has expertise in .NET applications.
Guess it's all right so long as it works.

My only concern is that there is a panel layout config saved to "Environment.SpecialFolder.LocalApplicationData / AGS". Also, skin system made by @persn is also saved there.
Is this same location as these preferences would be saved to? I think it would be nice to have all Editor user config save in one place.

morganwillcock:
Okay, I'll have a look at those extra settings and see how they work.

morganwillcock:
The loading and saving for panel layouts looks like it is handled externally, but changing AssemblyCompany to 'AGS' gets the Editor preferences into the same place as the layout file and the theme directory. I'll try to add the registry migration over the next few days and test it out.

morganwillcock:
@ivan-mogilko, I think this should migrate all current settings and any bad registry data will just end up as the default value. It probably needs some other people to try it out, but everything seems to be working for me.

There are a couple of changes which I later reverted (mainly adding a TypeConverter and removing it again), do you want me to rebase to clean it up or is it okay as-is?

ivan-mogilko:
> There are a couple of changes which I later reverted (mainly adding a TypeConverter and removing it again), do you want me to rebase to clean it up or is it okay as-is?

I decided to not enforce this convention anymore, but you may do it if you like.
Anyway, this pr now needs additional fix after merging #437.

morganwillcock:
Yes, using 'switch' was probably a bad idea, and using a default case, another bad idea. I've made it more specific.

ivan-mogilko:
Ok, following option is not imported from the old registry:

Editor appearance:
- Prompt dialog on closing multiple tabs

Script editor:
- Indent: spaces/tabs (indent width is loaded correctly)

- New game directory.

Also, may I suggest to increase maximal recent games to at least 10? Previous versions did not have any limit, which probably did not make much sense, but in practice, for example, I am using around 8-10 recent links for current game project, test games, demo games for WIP test modules etc.

ivan-mogilko:
Found couple of more bugs:

* In recent games same entries get duplicated as you open/close a project.
* Not critical, but related to the "recent search" limit: the limit only gets applied when you search something, but not right after old registry settings are imported (unlike "recent games" which is applied immediately).

morganwillcock:
> Prompt dialog on closing multiple tabs

When I tried it, this was working okay (perhaps 'migrated' was already set to true?).
The other issues should be fixed.

ivan-mogilko:
> When I tried it, this was working okay (perhaps 'migrated' was already set to true?).

I tested unchecking it in "registry" version, but it started checked in the new version.

Also, the settings are saved in a subfolder named with editor's version number. If the version number changes, will it know to read existing settings from previous version?

morganwillcock:
>I tested unchecking it in "registry" version, but it started checked in the new version.

I think I must have tried the other way around which happened to match the default value, as it was a typo which then got auto-completed everywhere. It should be fixed now.

There is a built-in upgrade mechanism which migrates settings between versions, so you can override to add behavior, or roll backwards on purpose if you wish. At the moment when the settings are first loaded there is a check that they have been upgraded, if not it calls [Upgrade](https://msdn.microsoft.com/en-us/library/system.configuration.applicationsettingsbase.upgrade) and then stores that as one of the settings (so it only runs once per version).

<pre>if (!UpgradedSettings)
{
    Upgrade();
    UpgradedSettings = true;
}</pre>

ivan-mogilko:
It just occured to me, does this means that every new release of AGS will have its own separate preferences values?

morganwillcock:
Yes, upgrading creates new settings which are separate. If you want to read previous settings after upgrade you can use [GetPreviousVersion](https://msdn.microsoft.com/en-us/library/system.configuration.applicationsettingsbase.getpreviousversion) (so going forwards is flexible). Going backwards will just use the settings for the matching version (so using newer settings would have to be implemented by manually parsing the XML for settings which are known to be safe).

ivan-mogilko:
Well, my concern was that before this change the preferences you set will affect any version of AGS, so if you install older version, compatible values will be applied there too.
Now in that case you will have to setup older versions of the editor separately, if they are installed later than the newer ones.

Idk if that might be an issue, guess regular users do not switch between versions often. What do you think?

morganwillcock:
I think if people are going backwards it would only be to versions where they need to work with existing builds of ports (e.g. where they are distributing games based around a working iOS port). Those versions would be old enough to just be using the registry settings again. I wouldn't have thought people would be downgrading the more recent versions.

ivan-mogilko:
Alright, so, this may be merged now, right?

PR #438, opened 2018-04-03
new jump point search routing with path refinement

kmar:
Hi guys, I forked AGS to try my jump point search A* study and it seems to work well so far.

What's the benefit:
+ faster (up to much faster) pathfinding
+ somewhat nicer paths (better refinement especially at path end)
+ when clicking unreachable walls the player moves somewhat closer to destination using the new algorithm
+ amortized heap allocation elimination when pathfinding
+ eliminating old pathfinding code

What's the price to pay:
- 4 times more memory scratch memory required (4xwall bitmap dims), for 3kx1k bitmap, original AGS used 3 megs of extra RAM while my new pathfinder uses 12 megs; probably not a bit deal even on mobiles
- not thoroughly tested (3 games only - King's Quest II, Resonance Demo and Eternally us), may contain bugs

Performance of original AGS (runtime polling disabled):

King's Quest II
routing: 2828.739462 usec
routing: 3104.942280 usec
routing: 1496.928715 usec
routing: 8.753969 usec
routing: 1469.157503 usec
routing: 2791.610558 usec

Eternally us
routing: 44596.944623 usec
routing: 46470.897732 usec
routing: 44568.871550 usec
routing: 60039.247968 usec
routing: 44669.994986 usec
routing: 43908.701536 usec

Performance of my new JPS refined pathfinder (doesn't poll at all):
(note that first request is outlier due to scratch memory allocation)

King's Quest II
routing: 1969.039324 usec
routing: 527.954895 usec
routing: 1580.846074 usec
routing: 0.301861 usec
routing: 324.198718 usec
routing: 1015.762276 usec

Eternally Us:
routing: 6435.374729 usec
routing: 283.447482 usec
routing: 565.083798 usec
routing: 521.917675 usec
routing: 308.501945 usec

So we can see the new pathfinder is 1.5 times faster on simple path and almost two orders of magnitude faster on large bitmaps.

The patch still needs more testings but I think it'd be worth it.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -97,9 +97,6 @@ OPT_RENDERATSCREENRES, extended engine caps check, font vertical offset.
 49 : 3.4.1.2
 Font custom line spacing.
 
-50 : 3.4.1.3
-New movelist format.
-
 */
 
 enum GameDataVersion
@@ -132,8 +129,7 @@ enum GameDataVersion
     kGameVersion_340_4          = 47,
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
-    kGameVersion_341_3          = 50,
-    kGameVersion_Current        = kGameVersion_341_3
+    kGameVersion_Current        = kGameVersion_341_2
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -97,6 +97,9 @@ OPT_RENDERATSCREENRES, extended engine caps check, font vertical offset.
 49 : 3.4.1.2
 Font custom line spacing.
 
+50 : 3.4.1.3
+New movelist format.
+
 */
 
 enum GameDataVersion
@@ -129,7 +132,8 @@ enum GameDataVersion
     kGameVersion_340_4          = 47,
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
-    kGameVersion_Current        = kGameVersion_341_2
+    kGameVersion_341_3          = 50,
+    kGameVersion_Current        = kGameVersion_341_3
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1343,10 +1343,7 @@ void ReadMoveList_Aligned(Stream *in)
     AlignedStream align_s(in, Common::kAligned_Read);
     for (int i = 0; i < game.numcharacters + MAX_INIT_SPR + 1; ++i)
     {
-        if (loaded_game_file_version >= kGameVersion_341_3)
-            mls[i].ReadFromFile(&align_s);
-        else
-            mls[i].ReadFromFile_Legacy(&align_s);
+        mls[i].ReadFromFile(&align_s);
 
         align_s.Reset();
     }
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1343,7 +1343,7 @@ void ReadMoveList_Aligned(Stream *in)
     AlignedStream align_s(in, Common::kAligned_Read);
     for (int i = 0; i < game.numcharacters + MAX_INIT_SPR + 1; ++i)
     {
-        mls[i].ReadFromFile(&align_s);
+        mls[i].ReadFromFile_Legacy(&align_s);
 
         align_s.Reset();
     }
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1343,7 +1343,11 @@ void ReadMoveList_Aligned(Stream *in)
     AlignedStream align_s(in, Common::kAligned_Read);
     for (int i = 0; i < game.numcharacters + MAX_INIT_SPR + 1; ++i)
     {
-        mls[i].ReadFromFile(&align_s);
+        if (loaded_game_file_version >= kGameVersion_341_3)
+            mls[i].ReadFromFile(&align_s);
+        else
+            mls[i].ReadFromFile_Legacy(&align_s);
+
         align_s.Reset();
     }
 }
--- a/Engine/ac/movelist.cpp
+++ b/Engine/ac/movelist.cpp
@@ -13,16 +13,17 @@
 //=============================================================================
 
 #include "ac/movelist.h"
+#include "ac/common.h"
 #include "util/stream.h"
 
 using AGS::Common::Stream;
 
 void MoveList::ReadFromFile(Stream *in)
 {
-    in->ReadArrayOfInt32(pos, MAXNEEDSTAGES);
+    in->ReadArrayOfInt32(pos, MAXNEEDSTAGES_LEGACY);
     numstage = in->ReadInt32();
-    in->ReadArrayOfInt32(xpermove, MAXNEEDSTAGES);
-    in->ReadArrayOfInt32(ypermove, MAXNEEDSTAGES);
+    in->ReadArrayOfInt32(xpermove, MAXNEEDSTAGES_LEGACY);
+    in->ReadArrayOfInt32(ypermove, MAXNEEDSTAGES_LEGACY);
     fromx = in->ReadInt32();
     fromy = in->ReadInt32();
     onstage = in->ReadInt32();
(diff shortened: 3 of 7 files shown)


ivan-mogilko:
Hello.
This sounds interesting. For the testing I may recommend creating test games in addition to using "real" ones, that will allow setting up some unusual cases with complicated walkable areas.

rofl0r:
:+1:

check this post about one of the games that is tricky with regards to pathfinding: https://github.com/adventuregamestudio/ags/pull/160#issuecomment-47511917

also maybe it'd make more sense to do development on the scummvm port https://github.com/adventuregamestudio/scummvm/tree/ags since AFAIK the engine development here is on hold.

kmar:
thanks for the link, unfortunately Duty and beyond crashes for me when I try to load a saved game (not related to my changes), so I can't really use that for testing

as for scummvm port, I'm not quite sure people who still use AGS to publish their games want a 5 times bigger executable for packaging, anyway my navigation routines are fully independent on AGS code so it should be trivial to port

maybe the scummvm port would work better on Linux - what I find sad is that the AGS Linux port runs like a slug (got like 3 fps for a certain game, running in lower res I got a whopping 10 fps or so)

now back to routing: seems to work fine so far, I pushed some refactoring update to make the external API easier
it's a pity that AGS uses a hardwired maximum of 40 waypoints; changing this would be trivial but since it's serialized it would break savegames

kmar:
I made some experiments in AGS editor and created a 1-pixel wide non-walkable spiral (worst-case)
because my pathfinder is pixel-perfect, this is enough to navigate around the spiral
however, the original pathfinder happily walks around the map freely

I investigated the old code and it seems to jump (in dijkstra mode) in 2 or 3 pixel units (3 pixels in 99.9% cases) probably in an attempt to improve performance

this explains why it ignores 1-pixel wide non-walkable pixels
it most likely also explains why "Duty and beyond" worked with the old code despite having non-connected walkable ares

so I think the old behavior is not desired anyway because it's not what one would expect;
I can live with that - broken games (Duty and beyond - disconnected walkable areas) wouldn't work anymore
but they can still run with old binaries which they were packaged with

ivan-mogilko:
> it's a pity that AGS uses a hardwired maximum of 40 waypoints; changing this would be trivial but since it's serialized it would break savegames

Savegame incompatibility is not an issue currently, because we recently pushed a completely new savegame format to master branch, that was not used in any new official release yet.

And even without that, it is always possible to make backward compatible switch in savegame loading, using format version (you may notice these in old savegame code).

kmar:
ok I changed max number of waypoints to 256 without breaking backward compatibility (numstages is saved and can be used for this)
tests look good so far, even complicated spiral path can be fully resolved now

ivan-mogilko:
Is there a reason to actually set _any_ limit to number of waypoints, maybe just use std::vector instead?

kmar:
yes, I thought about vectors as well, but the code involved in changing movelist data involves multiple places

I think 256 waypoints is more than generous for now
my concern actually wasn't the limit itself but extra memory usage (because of PSP),
but since AGS game has implied limit of 1000 characters, 256 * 12 * 1000 = ~3 megabytes
that's ok I think and I don't expect any game to use more than 100 characters (correct me if I'm wrong)
from the code I deduced that move lists are preallocated per character (a fixed number for each game) => 1 move list per character
this is necessary because one can save the game at any time and paths have to be remembered (plus multiple characters can be moving at the same time, if that's what you meant)

for now I'll stop pushing changes (except for potential bugfixes) and I'll test the pathfinding a bit more;
(performance looks good, no problems so far, complex navigation works well in synthetic tests)
plus I don't want to make any more changes to make it easier for review;
of course future improvements/clean up are possible (if necessary)

the only thing I noticed is that sometimes AGS has problems to interpolate animation/movement smoothly across waypoints,
but I double-checked and the new pathfinding code doesn't generate anything weird like double waypoints in the same place,
so my guess is this is simply a feature of AGS (or perhaps it's actually game-related)

ivan-mogilko:
Ok, I will probably be able to review and test this more closely on this weekend.

PS. PSP port may be thought as non-supported currently. It has been ages since anyone mentioned using it, it was last tested like 2 years ago (with only partial success),

ivan-mogilko:
So, I made this test game with 8 rooms: https://www.dropbox.com/s/0gwez1u3qt59660/ags-pathfinding-test.zip?dl=0
Changing rooms may be done using debug command (ctrl + X).
Game features "free look" ability activated by holding down right mouse button. Moving mouse around will make viewport move. This may be useful when testing scrolling rooms (if you want to order walking to the larger distance). Pressing '1' key will reset camera to the character.
I changed walking mode cursor graphic to the "pointer", to make it easier choosing where to go.


Room 1: a spiral made of rectangular shapes.

Old pathfinding: stand in top-left corner, order move into the center of spiral: path not found. Also, if you click fast enough while moving mouse, you will notice that cursor freezes for a moment, indicating long process.
New pathfinding: walks all the way, and freezes are only noticeable with Debug build (STL is pretty slow in debug configuration).


Room 2: a round spiral.

Old pathfinding: same thing, cannot walk all the way from starting point to the center of spiral. Have to give two separate orders, first walking half way in.
New pathfinding: walks all the way.


Room 3: a grid made of wide passages.

Old pathfinding: no visible problems
New pathfinding: there are issues with the character walking around the grid cells in a zig-zag fashion instead of choosing obvious route.
I've made pictures to illustrate the problem: https://i.imgur.com/dgvXLgb.png
**Green line** is the _old pathfinder_ work, **red line** is _new pathfinder_.
It is interesting to note that this reproduces 100% on the bottom grid row, but walking around any other rows or columns works better (still happens sometimes).


Room 4: a cross-like walkable area.

Old pathfinding: no visible problems
New pathfinding: no visible problems


Room 5: big walkable blob with some irregularly shaped obstacles.

Old pathfinding: no visible problems
New pathfinding: no visible problems


Room 6: grid of thin lines.

Old pathfinding: Although character generally walks where told to when you click exactly on walkable area, there are often mistakes when you click on black (non-walkable) areas: instead of walking to the closest point character sometimes takes weird roundabout routes and ends up in the unexpected position. E.g. when standing on lowest line then clicking on a black space above it sometimes makes character go several "levels" upwards.
New pathfinding: seem to work well.


Room 7: a scrolling room with a long bending walkable area, and additional curved route in the middle.

Old pathfinding: when standing at the bottom of the room, clicking on the middle area causes pathfinding to fail. Also, standing in the middle area and click at bottom caused character to take non-optimal route (could randomly reproduce only once).
New pathfinding: seem to work well.


Room 8: a larger scrolling room with random walkable areas.

Old pathfinding: couple of times I thought that taken route is non optimal, but not completely sure. Did not notice any other issues.
New pathfinding: no visible problems

kmar:
thanks for the test cases
I fixed the issue (I won't go into details of what went wrong - let's say I tried to reduce memory usage and broke something along the way)
so now it utilizes 2x more scratch memory for routing than before, namely 8 times the dimensions of routing bitmap (W*H*8), so for Eternally Us (~3000x1000), that would be 24 megabytes
I think it may be ok (for 640x480 games that's only 2 megabytes)
I pushed the fix so you can try it yourself now

kmar:
ok I refactored movelist serialization, but I had to bump up game version
I hope I didn't miss any other places

ivan-mogilko:
No...
There is literally no need to check for version. No new format version is necessary.

Just this:
<pre>
AlignedStream align_s(in, Common::kAligned_Read);
     for (int i = 0; i < game.numcharacters + MAX_INIT_SPR + 1; ++i)
     {
         mls[i].ReadFromFile_Legacy(&align_s);
         align_s.Reset();
     }
</pre>
And that's it.

kmar:
but that would break savegames written before this change, I don't follow

kmar:
let me be more specific: let's say I'm using a current engine version to play a game and I have a bunch of savegames.
now when I update to the new engine binary, I won't be able to load older savegames, because ReadFromFile_Legacy would never be called (because current engine version uses the new savegame format, right?)

ivan-mogilko:
Oh, you mean you were actually using engine from the master branch to play games... TBH I did not think about that.

kmar:
yes, that's exactly what I meant
so bumping up the version will actually work for current engine, last stable engine, this new format (and all future engines if you accept this pull request) and also for acient legacy savegames

funny thing is I actually spent more time with this new movelist format than fixing the routing problems you found :)

ivan-mogilko:
Ok, I see now, sorry about that.
Ok, frankly I hoped this won't happen so fast. This complicates the issue a bit.

So, I need to explain few things then.

First of all, the way you did it by bumping up "game data version" won't work on its own, because "loaded_game_file_version" is the version of ... well, game data files, created by AGS Editor. Which means that the game needs to have it bumped also, otherwise the old save route will be used. Which you obviously cannot do with the already existing games.


This has to be done differently.

The new save system is implemented in savegame_components.cpp module, and uses completely separate set of format versions now, which refer strictly to the runtime save format and don't directly depend on game data (meaning you may serialize stuff differently for existing games).

Savegame serialization is split into "components", each component having its own version number.
If you are changing something in the new save reading/writing, you would need to add new version number to corresponding component, e.g. "Characters" or "Current room".

The "current save version" is defined in an array called ComponentHandlers. It's a number that is 0 for all components now.

The functions that read savegames have similar prototype, which takes "int32_t cmp_ver" argument. This is the component's version that may be checked.

So what have to be done is this:
* character component save version increased to 1;
* "current room" save version increased to 1;

In ReadCharacters function you need to check the save version and call appropriate function to read character's movelist:

<pre>
    // character movement path cache
    if (cmp_ver == 0)
       mls[CHMLSOFFS + i].ReadFromFile_Legacy(in.get());
    else
        mls[CHMLSOFFS + i].ReadFromFile(in.get());
</pre>

Similar change made in ReadThisRoom function where object's movelists are loaded.

Naturally, better is to introduce enums for version number, but this was not done yet.

ivan-mogilko:
Just to note, if you think all this is too confusing for you, I could do that instead after merging your pull request.

ivan-mogilko:
You know what, frankly, as I see the situation better now, I believe your very first serialization variant was working better, as it did not require too many changes, and could be left for the time being.

Sorry for taking so much of your time.

ivan-mogilko:
Alright, tested reading legacy saves, new saves before this pr, and made by this pr, everything works.

PR #441, opened 2018-05-02
Scrollbar fixes for the view editor

morganwillcock:
This should fix issues with loop positions in the view editor.
[http://www.adventuregamestudio.co.uk/forums/index.php?topic=55992.0](http://www.adventuregamestudio.co.uk/forums/index.php?topic=55992.0)


morganwillcock:
I've looked at trying to keep the add and remove buttons in view as well, but it looks like a lot of this is fighting against the default focus & autoscroll behaviour. Firing extra resize events can do it, but then other quirks start to show up. Long term solution is probably to move the buttons into the toolbar.

Adding a breakpoint actually keeps the buttons in view, because of the extra events from the focus change :/

ivan-mogilko:
> Long term solution is probably to move the buttons into the toolbar.

Maybe make a panel or toolbar under the loops and move it down / resize loops area are more loops are added/removed?

PR #444, opened 2018-05-12
Check GUI type is NormalGUI when deleting controls

### LINKED ISSUE #443: Text Window GUI crashes engine when border images are deleted
When a text windows is displayed (e.g. `Display("Hello")`), if any of the border images have been deleted, the game will crash. I think this makes it impossible to create a text window GUI that does not have borders, although the editor also appears to incorrectly crop the editing controls, so perhaps I am missing something. 

Also, once deleted, I can't see any way to put the border images back.

morganwillcock:
This should fix #443 by removing the delete option for Text Window GUIs.

### DIFF
--- a/Editor/AGS.Editor/Panes/GUIEditor.cs
+++ b/Editor/AGS.Editor/Panes/GUIEditor.cs
@@ -784,7 +784,7 @@ namespace AGS.Editor
 
         private void DeleteControlClick(object sender, EventArgs e)
         {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Oh, by the way, when testing this out, I noticed that you also can move these "buttons" with keyboard. Maybe there is something else.

Guess the proper solution would be to create a new class for these, which has its own handling.

PR #452, opened 2018-06-10
Remove internal engine type declarations from the plugin API

ivan-mogilko:
This disables plugin API functions that return addresses to the internal engine's data (characters, room objects, etc).

Their use may be only partially substituted with calls to script API from plugin, but I suppose that in the end plugins will have full and safer access to all script commands.

There are still few things left, like GetPalette and GetSpriteGraphic, but I decided to leave them in for now, because they do not impose much restriction on engine data. Something should be done with these later though.

~Also disabled non-portable API, that returns DirectX interfaces on Windows.~

I say "disabled", because just made them all throw errors, but did not remove function declarations because that would break existing plugins.

NOTE: this is only the first step to make these structs' layout unrestricted. Next we should deal with the script API that does practically same thing as plugins.

***

**EDIT** The purpose of this pr has changed, as I found that simply disabling functions that return engine objects may prevent plugins from using necessary script functions. So instead of completely turning these off, I decided to change object type declarations into incomplete (abstract) types, which do not describe their contents.

Naturally, this cannot prevent plugins from accessing engine memory, on mistake, or with actual purpose, but at least discourages from doing so, while we may officially drop support of using engine object memory directly.

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -47,7 +47,7 @@ using namespace AGS; // FIXME later
 #define FOLLOW_ALWAYSONTOP  0x7ffe
 
 struct CharacterExtras; // forward declaration
-// remember - if change this struct, also change plugin header file struct
+
 struct CharacterInfo {
     int   defview;
     int   talkview;
--- a/Common/ac/mousecursor.h
+++ b/Common/ac/mousecursor.h
@@ -22,7 +22,7 @@ using namespace AGS; // FIXME later
 #define MCF_DISABLED 2
 #define MCF_STANDARD 4
 #define MCF_HOTSPOT  8  // only animate when over hotspot
-// this struct is also in the plugin header file
+
 struct MouseCursor {
     int   pic;
     short hotx, hoty;
--- a/Engine/ac/gamestate.h
+++ b/Engine/ac/gamestate.h
@@ -62,7 +62,6 @@ struct GameState {
     int  fast_forward;           // player has elected to skip cutscene
     int  room_width;      // width of current room (320-res co-ordinates)
     int  room_height;     // height of current room (320-res co-ordinates)
-    // ** up to here is referenced in the plugin interface
     int  game_speed_modifier;
     int  score_sound;
     int  takeover_data;  // value passed to RunAGSGame in previous game
--- a/Engine/ac/roomobject.h
+++ b/Engine/ac/roomobject.h
@@ -23,8 +23,6 @@
 namespace AGS { namespace Common { class Stream; }}
 using namespace AGS; // FIXME later
 
-// This struct is only used in save games and by plugins
-// [IKM] Not really.... used in update loop
 struct RoomObject {
     int   x,y;
     int   transparent;    // current transparency setting
--- a/Engine/plugin/agsplugin.cpp
+++ b/Engine/plugin/agsplugin.cpp
@@ -46,6 +46,7 @@
 #include "media/audio/audio.h"
 #include "media/audio/sound.h"
 #include "plugin/agsplugin.h"
+#include "plugin/plugin_builtin.h"
 #include "plugin/plugin_engine.h"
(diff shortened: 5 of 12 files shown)


ivan-mogilko:
Need to check if existing built-in plugins are using some of these functions (Flashlight plugin seem to do).

ivan-mogilko:
From the looks of it, the Flashlight plugin requires IAGSPlugin::GetCharacter, and SnowRain needs IAGSPlugin::GetViewFrame, but in both cases it should be possible to replace raw data use with calls to the few exported engine functions (using GetScriptFunctionAddress). That will slow them down a bit relatively, but should not affect perfomance.

AGSLua plugin, on another hand, will be completely impossible to maintain. It uses the internal engine data extensively to read and write variables right in there.

ivan-mogilko:
Hm, I realized that completely disabling aforementioned plugin functions cannot be done, because most script functions also take pointer to struct, not some ID. If plugin would need to call, for example, Character's script function, it would need to pass *some* pointer, therefore the engine functions that return pointers cannot throw errors.

Looks like only solution currently is to use incomplete types in C-style: only declare returned pointer types (AGSCharacter, AGSViewFrame, etc) in plugin API, but don't define their contents.
In such case, plugin code will store returned pointers and pass them to other engine functions (such as Character_GetX), but won't be able to directly access their contents.

ivan-mogilko:
Alright, I modified the branch.

The example of how plugins may work now:
"Importing" script API:
<pre>
typedef int (*SCAPI_CHARACTER_GETX)(AGSCharacter *ch);
typedef int (*SCAPI_CHARACTER_GETY)(AGSCharacter *ch);
typedef int (*SCAPI_CHARACTER_GETLOOP)(AGSCharacter *ch);

Character_GetX = (SCAPI_CHARACTER_GETX)engine->GetScriptFunctionAddress("Character::get_X");
Character_GetY = (SCAPI_CHARACTER_GETY)engine->GetScriptFunctionAddress("Character::get_Y");
Character_GetLoop = (SCAPI_CHARACTER_GETLOOP)engine->GetScriptFunctionAddress("Character::get_Loop");
</pre>
Using script API with abstract pointer type, instead of accessing internal data directly:
<pre>
int char_x = Character_GetX(g_FollowCharacter);
int char_y = Character_GetY(g_FollowCharacter);
</pre>

Of course, that would be more convenient if plugin API header defined all those function typedefs, but we need only these few to keep supported plugins working.
Probably this is also about how plugin API should have looked like from the beginning, that would save some trouble.

Not sure if this is the way we want to keep it, but perhaps one of the options.

rofl0r:
rather than handing out pointer to real engine memory, the plugin API could copy the actual contents into plugin-user-supplied space, e.g:
```c
struct Character char;
pluginapi_get_character_data(&char);
printf("character locations: %d, %d\n", char.x, char.y);
```

ivan-mogilko:
> rather than handing out pointer to real engine memory, the plugin API could copy the actual contents into plugin-user-supplied space

There are two issues here:
- plugins usually need to know up-to-date values on each update, meaning you'd have to copy full struct every frame
- some plugins were allowed to change these values as well (and some did as far as I know).

ivan-mogilko:
Anyhow, the intent of this pull request was to remove restriction from the engine and let us refactor all engine's types at will. I tried to do so with minimal changes, and without adding anything to existing plugin API.

I have just finished testing new plugin code and it's working well, so if there aren't any objections I'd like to merge this soon.

Redesigning plugin API is a separate task, and to be honest, personally I am not even ready to discuss it at the moment. If someone has any thoughts regarding that, please open a new issue.

ivan-mogilko:
Only a small comment to previous discussion: of course current implementation still gives an abstract pointer to engine's memory to plugin's author, which may in theory be used for hacks or malicious purposes. If that is a concern, one alternative that comes to mind is to return some ID instead of pointer.

PR #455, opened 2018-06-12
Add builtin stubs for AGSteam and AGSGalaxy

monkey0506:
The AGSteam and AGSGalaxy client plugins are beyond the scope of this project, but function stubs for them should not require much maintenance as their APIs are not expected to change much. This could also serve as a complete replacement for [ags2clientstub](https://github.com/monkey0506/ags2clientstub).

I'm not sure if I knew plugins could be stubbed this way, but this seems appropriate based on the other existing stubs. I had started writing ags2clientstub as a builtin plugin when I found this file.

### DIFF
--- a/Engine/plugin/global_plugin.cpp
+++ b/Engine/plugin/global_plugin.cpp
@@ -50,9 +50,19 @@ RuntimeScriptValue Sc_PluginStub_IntNeg1(const RuntimeScriptValue *params, int32
     return RuntimeScriptValue().SetInt32(-1);
 }
 
+RuntimeScriptValue Sc_PluginStub_NullStr(const RuntimeScriptValue *params, int32_t param_count)
+{
+	return RuntimeScriptValue().SetStringLiteral(NULL);
+}
+
 bool RegisterPluginStubs(const char* name)
 {
   // Stubs for plugin functions.
+
+  bool is_agsteam = (stricmp(name, "agsteam") == 0) || (stricmp(name, "agsteam-unified") == 0) ||
+    (stricmp(name, "agsteam-disjoint") == 0);
+  bool is_agsgalaxy = (stricmp(name, "agsgalaxy") == 0) || (stricmp(name, "agsgalaxy-unified") == 0) ||
+    (stricmp(name, "agsgalaxy-disjoint") == 0);
 
   if (stricmp(name, "ags_shell") == 0)
   {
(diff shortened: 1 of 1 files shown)


monkey0506:
I did leave out the `FindLeaderboard` and `Initialize^2` stubs from `AGS2Client`. That was a mistake. Each of those functions is specific to AGSteam and AGSGalaxy (respectively), but they are defined in `AGS2Client` in the script if using a ["unified" build of the plugin](https://github.com/monkey0506/ags2client/wiki/Unified-builds). I will correct that and recommit; no need for a separate commit for that.

monkey0506:
For those not aware of what "unified" and "disjoint" mean here, AGSteam and AGSGalaxy have near-identical function calls if the Steamworks/GOG Galaxy developer sets up similar achievements and stats on both backend websites. "unified" means that both will be represented in game scripts using the `AGS2Client` struct (e.g., `AGS2Client.IsAchievementAchieved(...)` for both AGSteam and AGSGalaxy), and "disjoint" means the opposite (e.g., `AGSteam.IsAchievementAchieved(...)` and `AGSGalaxy.IsAchievementAchieved(...)`).

"agsteam-unified.dll" and "agsgalaxy-unified.dll" were names used during the implementation of the AGS2Client interface, but unified builds later became the default (introducing the "-disjoint" suffix instead). Notably, Wadjet Eye Games did make use of "agsteam-unified.dll". This is the reason why it's necessary to check several possible names for each plugin. There are no foreseeable reasons for naming forks such as this in the future.

ivan-mogilko:
First of all, only to make sure that you are targetting wanted version, the master branch is currently the one with removed backwards compatibility, if you want to have these in the version that supports older games (made with 3.4.1 and earlier), then better make pull request against "ags3" branch.

Then, what is the actual reasoning for having a builtin stub vs providing a stub plugin? The plugin stubs were kind of compelled measure, and personally I hoped to get rid of these at some point if better solution is found.
AFAIK @sonneveld has implemented a way to add built-in plugins externally, using **pl_register_builtin_plugin()** function. It is used in Wadjet Eye's OSX games. Since mac port requires to have some additional wrapper around engine anyway, they make builtin plugins in the game code instead, and register using that engine API.

ivan-mogilko:
@monkey0506 , you've closed this pull request without any comment, I hope my reply did not sound as "brush off"?
My intention was to find out whether adding a stub is actually a necessity, because if there is a critical need to have it (e.g. if it's too complicated to have a stub plugin), then this could still be used as a temporary workaround. After all, quite a lot of games use your steam plugin now, and I've seen complaints about missing script links popping up from time to time on forums.

On the other hand, if it's merely an option, then having an actual plugin seem like a cleaner variant.

Right now I am trying to refresh my memories on issues we have with plugins on different platforms. We had ideas how to "automate" plugin stubbing in the past, maybe that idea could be revisited.

PR #460, opened 2018-06-30
3.4.1: Separate native and displayed image formats; support actual 16-bit native mode again

ivan-mogilko:
Before 3.4.1 AGS had certain issues when running 16-bit games in Direct3D and OpenGL renderers (loss of color hue precision, and something else), and also sometimes software renderer had trouble running in 16-bit display mode on latest Windows systems (where it uses DirectDraw 5 interface).

For 3.4.1 we've made a decision to force all renderers to run in 32-bit display mode always (except for 8-bit games in software mode, but these are very special case anyway), and to convert all game bitmaps to 32-bit ones on load, for simplicity (and for simplier conversion from bitmap to texture).

While the first half of this decision (32-bit display mode) is rational on its own, the second half was a lazy thinking on my part. The problem here is that if the game was developed and scripted as 16-bit, having its resources converted to 32-bit changes the way drawing operations work, although slightly, but causing occasional mistakes and differences, notably when it comes to sprite transparency and scripted raw drawing (since color range is different). Unfortunately, actual bugs were not found until after 3.4.1 was already released.

This pull request suggests different and, I hope, more logical approach to the problem. Instead of converting game resources to the bitness of display mode, it keeps them in native game's color depth (converting varying bitmaps to **game's** bitness when necessary), and hardware-accelerated renderers are taught to correctly convert three main bitmap types (8, 16 and 32 bit) to the 32-bit RGBA texture.

**NOTE 1:** This change mainly affects how 16-bit games work. But my belief is that another big benefit is in keeping native game format clearly separated from the renderer mode.

**NOTE 2:** There may be a difference in how 8-bit games will be rendered with Direct3D and OpenGL with this change. But they did not quite work before, and they still won't quite work, only visual issues may be different.

**PS.** The main problem with running 8-bit games with Direct3D and OpenGL seem to be that the engine cannot correctly tell *when to update the texture*, and large parts of the game may get rendered using uninitialized or non-matching palette (usually this results in parts of the game scene being completely black).

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -207,9 +207,6 @@ void setpal() {
     set_palette_range(palette, 0, 255, 0);
 }
 
-
-#ifdef USE_15BIT_FIX
-
 int _places_r = 3, _places_g = 2, _places_b = 3;
 
 // convert RGB to BGR for strange graphics cards
@@ -238,9 +235,6 @@ Bitmap *convert_16_to_16bgr(Bitmap *tempbl) {
 
     return tempbl;
 }
-#endif
-
-
 
 // PSP: convert 32 bit RGB to BGR.
 Bitmap *convert_32_to_32bgr(Bitmap *tempbl) {
@@ -324,13 +318,11 @@ Bitmap *AdjustBitmapForUseWithDisplayMode(Bitmap* bitmap, bool has_alpha)
         else // else simply convert bitmap
             new_bitmap = BitmapHelper::CreateBitmapCopy(bitmap, game_col_depth);
     }
-#ifdef USE_15BIT_FIX
     // Special case when we must convert 16-bit RGB to BGR
     else if (convert_16bit_bgr == 1 && bmp_col_depth == 16)
     {
         new_bitmap = convert_16_to_16bgr(bitmap);
     }
-#endif
     return new_bitmap;
 }
 
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -221,14 +221,14 @@ Bitmap *convert_16_to_15(Bitmap *iii) {
 
     if (iii->GetColorDepth() > 16) {
         // we want a 32-to-24 conversion
-        Bitmap *tempbl = BitmapHelper::CreateBitmap(iwid,ihit,System_GetColorDepth());
-        if (System_GetColorDepth() < 24) {
+        Bitmap *tempbl = BitmapHelper::CreateBitmap(iwid,ihit,game.GetColorDepth());
+        if (game.GetColorDepth() < 24) {
             // 32-to-16
             tempbl->Blit(iii, 0, 0, 0, 0, iwid, ihit);
             return tempbl;
         }
 
-        GFX_VTABLE *vtable = _get_vtable(System_GetColorDepth());
+        GFX_VTABLE *vtable = _get_vtable(game.GetColorDepth());
         if (vtable == NULL) {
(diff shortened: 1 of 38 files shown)


Engine/ac/draw.cpp
```diff
@@ -210,87 +210,6 @@ void setpal() {
 
 #ifdef USE_15BIT_FIX
```
sonneveld: probably worth getting rid of any use of USe_15BIT_FIX.

sonneveld:
What games were having issues with the 16bit to 32-bit conversion?  Were there scripts that were relying on pixel colour values?

Engine/gfx/ali3dogl.cpp
```diff
@@ -1397,166 +1396,40 @@ void OGLGraphicsDriver::DestroyDDB(IDriverDependantBitmap* bitmap)
...
-    if (y == tile->height)
-    {
-      unsigned int* memPtrLong = (unsigned int*)memPtr;
-      unsigned int* memPtrLong_previous = (unsigned int*)(memPtr - tileWidth * 4);
+    unsigned int* memPtrLong = (unsigned int*)(memPtr + pitch * tile->height);
+    unsigned int* memPtrLong_previous = (unsigned int*)(memPtr + pitch * (tile->height - 1));
 
       for (int x = 0; x < tileWidth; x++)
         memPtrLong[x] = memPtrLong_previous[x] & 0x00FFFFFF;
-
```
sonneveld: Is it possible to revert some of these refactorings until later? I believe some of these changes that are simply refactoring/tidying up will conflict with the sdl2 port.

ivan-mogilko:
@sonneveld 
> What games were having issues with the 16bit to 32-bit conversion?

This one was reported:
http://www.adventuregamestudio.co.uk/site/games/game/2182/

The main menu is on the GUI. With the 3.4.1 the GUI becomes a pink rectangle. This pull request fixes it.

> Were there scripts that were relying on pixel colour values?

I will be honest, this is theoretical possibility that I do not have any practical proof for.


I believe there is a need for more elaboration. 16->32 bit conversion of the game resources still could be tried, but it may require additional attention, to make sure that conversion is done right and that drawing operations are performed using specific variants of functions.
For example, the issue in the game linked above could be occuring because of incorrectly assigned transparency flags.

But what I did not like, and what IMHO was a biggest mistake in my previous work, that *game bitmaps were strictly tied to the renderer's mode* (which could be anything). With this pr that's reverted, and now game bitmap format is defined by game's setting rather than renderer's setting. In my opinion that is specifically important if we are planning to implement a "mixed" renderer for 8-bit games (the one that only converts final image to 32-bit). Again, in theory that might be possible to reimplement 16->32 bit conversion of resources later, only this time we would be a separate engine's option that does this, not dictated by renderer's mode.
Of course I may got confused here, I've changed my opinion on this several times in the past. Does what I propose make any sense?

Engine/ac/draw.cpp
ivan-mogilko: I will remove it, but just to make things clear, USE_15BIT_FIX is always defined now, it was introduced when Chris Jones switched from the older version of Allegro (there is a comment in this source file above).
I am not sure whether the code under this @ifdef has any relation to the macro's original meaning though.

Engine/gfx/ali3dogl.cpp
ivan-mogilko: That's possible, but what's the nature of conflict? I looked into the latest code in your sdl2 branch and it seem identical to what is in 3.4.1 now.

PR #482, opened 2018-07-20
AGS 3: improvements to the new Room Editor's interface

ivan-mogilko:
### Problem

Adding new breadcrumb navigation bar in the Room Editor (#428) introduced several issues.

Two most noteable of these are:
* All the design-time item properties (such as visibility in the editor) are getting reset as soon as list of items is refreshed.
* Layer's visibility and locked state is also reset under same condition.
* Navigation bar fails to react to the adding and deleting of Characters, as well as changing Character's starting room.

The reason of the first problem was simply a lack of corresponding event handling (in case of Characters - the lack of events in general).
The reason of the second problem was that these Locked and Visible flags were implemented as lists of item names, and the presence of the item name was intepreted as True state of the flag, while abscense as False. With each item list refresh these lists were reset to the default states (all items visible, and all unlocked).
As for the layers, they simply did not have persistent Visible/Locked properties at all, which made them fully depend on the navbar button state.

Trying to fix these I decided to reimplement certain things in the room editor filters, but was not able to achieve particulary good looking code, only relatively better one (that's my hope at least). Unfortunately, the Room Editor code appears to be messy, with duplicated functionality and confusing ties with the rest of the program. I wasn't ready for massive refactoring and decided to limit my changes to only achieve what's necessary for the new UI.

---

### Changes

1) First of all, I replaced two lists VisibleItems and LockedItems with a class named DesignTimeProperties which contains two boolean values. Instances of that class are stored in a dictionary with item ID as a key. I believe such class is easier to ammend and work with. Also, this dictionary does not get recreated when the navbar refreshes itself.

For the layers, I added persistent Visible and Locked properties to the IRoomEditorFilter interface. When navbar is initialized or refreshed, it reads these, and then user clicks on navbar buttons, Filter class remembers that choice.

2) In the original implementation the breadcrumb bar was addressing objects using their display name. Room Filters did not store these names anywhere, but instead generated them every time on demand. Probably this was made so to prevent desynchronization errors between navbar and game content.
Instead of this I added a persistent lookup table which maps room item IDs to the item instances, and is updated by events to keep in sync with the game.
Also, for item ID navbar and filters now use strings created as formatted integer ID of room item or Character.

The latter may need explanation. The problem of AGS now is that there are no real unique identifiers. The ScriptName could be thought as a first candidate for one, but unfortunately not every game entity has got a script name, and secondly this may sound paradoxical, but even for those that have script names are *not mandatory*. That's true: you may have a Character or Room Object without a script name (in fact, Room Objects are created without one by default).

This is why the only real way to make an ID is to use the numeric index in the list of items.
Of course this index will **not** be globally unique, but it will be *unique in the scope of objects of same type*, and that seem to be enough for the room editor at this time.
(I was thinking that perhaps if necessary they could be made generally unique with a prefix indicating their type.)

3) Finally, I made Room Editor correctly react to Room Objects and Characters being added, removed and their starting room changed (other types of items cannot be added/deleted at the moment).

**NOTE:** I had plans to also work on serializing these design-time settings, but then decided to prepare that as a separate pull request, because current changes are tough already.

---

**PS.** To be honest, I was very much dissapointed with the Room Editor's code structure. There seem to be several core issues, and the worst IMHO is that there is no parent interface for the game entity. This latter makes it rather difficult to invent a generic editor code. Another notable problem is that there does not seem to be convenient notification mechanism between game contents and the editor UI. Property Grid sends notifications for edited values, but only to the active pane, while another pane may also need it.

My belief is that if someone would be improving this UI further later, or even rewriting Room Editor from scratch, the first thing they'd have to do is improve the game data management.


ivan-mogilko:
Ammended with also fixing whole layer's Visible and Locked states being reset when the item list changes.
Now everything seem to work correctly and remembers their state.

PR #484, opened 2018-07-31
AGS 4: refactor RoomStruct class and room saving/loading

ivan-mogilko:
This does two things:

1) Room serialization tidied and grouped in one code unit (Common/roomfile.h and cpp).
2) RoomStruct class cleaned of few remaining unused data, and brought to proper code style.

The latter caused a lot of code changed throughout the engine, but that's mostly because of variable renames.

Note, that in regards to read/write I follow same "philosophy" as with the previous refactoring of the main game data and savegame code (as well as the one used in the Editor with XML serializer and DataFileWriter class) that tells that the class that implements game functionality does not have to know the file format it is saved to. This is why the serialization code is separated from the RoomStruct class.


ivan-mogilko:
Rebased on latest master and fixed bugs in room save function.

PR #493, opened 2018-08-17
Large file support

ivan-mogilko:
This implements large file (64-bit offsets) support in the AGS engine.

I used SDL2 rwops code for the reference to define different variants of function names. This works on MS Windows but I'd need help setting this up for ports. For example, is it safe to declare HAVE_FSEEKO64 in the Linux Makefile as a default?

### DIFF
--- a/Common/util/filestream.cpp
+++ b/Common/util/filestream.cpp
@@ -50,6 +50,11 @@ FileStream::~FileStream()
     Close();
 }
 
+bool FileStream::HasErrors() const
+{
+    return IsValid() && ferror(_file) != 0;
+}
+
 void FileStream::Close()
 {
     if (_file)
--- a/Common/util/filestream.cpp
+++ b/Common/util/filestream.cpp
@@ -16,7 +16,18 @@
 #include "util/filestream.h"
 #include "util/math.h"
 
+#if defined(HAVE_FSEEKO) // Contemporary POSIX libc
+    #define file_off_t  off_t
+    #define fseek       fseeko
+    #define ftell       ftello
+#elif defined(_MSC_VER) // MSVC
+    #define file_off_t  __int64
+    #define fseek       _fseeki64
+    #define ftell       _ftelli64
+#else // No distinct interface with off_t
+    #define file_off_t  long
 #endif
+
 
 namespace AGS
 {
@@ -71,13 +82,13 @@ bool FileStream::EOS() const
     return !IsValid() || feof(_file) != 0;
 }
 
-size_t FileStream::GetLength() const
+soff_t FileStream::GetLength() const
 {
     if (IsValid())
     {
-        size_t pos = (size_t)ftell(_file);
+        soff_t pos = (soff_t)ftell(_file);
         fseek(_file, 0, SEEK_END);
-        size_t end = (size_t)ftell(_file);
+        soff_t end = (soff_t)ftell(_file);
         fseek(_file, pos, SEEK_SET);
         return end;
     }
@@ -85,11 +96,11 @@ size_t FileStream::GetLength() const
     return 0;
 }
 
-size_t FileStream::GetPosition() const
+soff_t FileStream::GetPosition() const
 {
     if (IsValid())
     {
-        return (size_t) ftell(_file);
+        return (soff_t) ftell(_file);
     }
     return -1;
 }
@@ -145,7 +156,7 @@ int32_t FileStream::WriteByte(uint8_t val)
     return -1;
 }
 
-size_t FileStream::Seek(int offset, StreamSeek origin)
+soff_t FileStream::Seek(soff_t offset, StreamSeek origin)
 {
     if (!_file)
     {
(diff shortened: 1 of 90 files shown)


rofl0r:
nope sorry, you're doing this wrong.

quote from https://wiki.musl-libc.org/faq.html


> Do I need to define _LARGEFILE64_SOURCE to get 64bit off_t?
>
>    -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE have nothing to do with supporting large files properly but with exposing the idiotic legacy interfaces with 64 on the end of their names (like open64, lseek64, etc.). only -D_FILE_OFFSET_BITS=64 should ever be used, anywhere, even on glibc.

that means you should only use `-D_FILE_OFFSET_BITS=64` in your CFLAGS but **NOT** use `_LARGEFILE_*SOURCE` and any of the *64 interfaces. this gives the desired result that the standard LIBC functions like stat() and friends work with 64-bit off_t, without having to use any hacks.

ivan-mogilko:
Okay, if I understood correctly -
1. Remove _LARGEFILE64_SOURCE, and instead expect that '_FILE_OFFSET_BITS=64' is defined in flags when necessary.
2. Don't use fseeko64 and other *64 functions.

About fseek**o**/ftell**o**, are they supposed to be declared always in standard library on Linux (regardless of OFFSET_BITS value)?

rofl0r:
almost :)
1. Remove _LARGEFILE64_SOURCE
2. Add -D_FILE_OFFSET_BITS=64 to CFLAGS in Makefile
3. Don't use fseeko64 and other *64 functions.

Common/util/filestream.cpp
```diff
@@ -12,24 +12,23 @@
...
-#if !defined (WINDOWS_VERSION)
-#include <sys/stat.h>
-long int filelength(int fhandle)
-{
-    struct stat statbuf;
-    fstat(fhandle, &statbuf);
-    return statbuf.st_size;
-}
+#if defined(HAVE_FSEEKO) // Contemporary POSIX libc
+    #define file_off_t  off_t
```
rofl0r: does your build system support checking things ?
if so, the best thing would be to check whether the off_t type is defined
like by trying to compile this here
```c
#include <unistd.h>
int main() { off_t foo; }
```

ivan-mogilko: > does your build system support checking things ?

You mean like CMake? frankly that's the only thing I know that can do that.
No, we don't have anything like that set up right now. Maybe we should do at some point.

ivan-mogilko:
Added -D_FILE_OFFSET_BITS=64 to makefiles.

Common/util/filestream.cpp
rofl0r: hell no, [cmake sucks](https://github.com/libtom/libtomcrypt/issues/391#issuecomment-390434739). autoconf could do it, or even plain Makefiles with some tricks.

Engine/Makefile-defs.linux
```diff
@@ -1,6 +1,6 @@
 INCDIR = ../Engine ../Common ../Common/libinclude ../Plugins
 LIBDIR =
-CFLAGS := -O2 -g -fsigned-char -Wfatal-errors -DNDEBUG -DAGS_RUNTIME_PATCH_ALLEGRO -DAGS_HAS_CD_AUDIO -DAGS_CASE_SENSITIVE_FILESYSTEM -DALLEGRO_STATICLINK -DLINUX_VERSION -DDISABLE_MPEG_AUDIO -DBUILTIN_PLUGINS -DRTLD_NEXT $(shell pkg-config --cflags freetype2) $(CFLAGS)
+CFLAGS := -O2 -g -fsigned-char -Wfatal-errors -DNDEBUG -DAGS_RUNTIME_PATCH_ALLEGRO -DAGS_HAS_CD_AUDIO -DAGS_CASE_SENSITIVE_FILESYSTEM -D_FILE_OFFSET_BITS=64 -DHAVE_FSEEKO -DALLEGRO_STATICLINK -DLINUX_VERSION -DDISABLE_MPEG_AUDIO -DBUILTIN_PLUGINS -DRTLD_NEXT $(shell pkg-config --cflags freetype2) $(CFLAGS)
 CXXFLAGS := -std=c++11 -fno-rtti -Wno-write-strings $(CXXFLAGS)
```
rofl0r: CFLAGS is not used in CXXFLAGS ?

the MacOS makefile below does

ivan-mogilko: Hmm.

Makefile.linux has this:
`CXXFLAGS += $(CFLAGS)`

but Makefile.osx has this:
`CXXFLAGS := -std=c++11 -fno-rtti -Wno-write-strings $(CXXFLAGS)

then later -

CXXFLAGS := $(CFLAGS) $(CXXFLAGS)`

which means that in Mac's makefile CFLAGS are duplicated in CXXFLAGS?

ivan-mogilko:
Sorry, had to rebase again (github could not resolve diff in macos makefile).

Engine/Makefile-defs.linux
rofl0r: oh, seems i overlooked the CXXFLAGS += line.

Common/util/filestream.cpp
sonneveld: Now implemented in current cmake build system.

rofl0r: nice try trolling me

PR #497, opened 2018-08-24
Sign assemblies to generate a strongname

### LINKED ISSUE #481: Editor: new preferences are serialized separately for separate installations of the same version of application
Since I am running Editor from different locations on my computer, such as normal installation, and couple of source code locations, I noticed that user preferences in these do not match each other, even though the program versions do.

I checked the directory where these settings are now saved, and it looks like Editor creates *separate* group of setting files for each program location (URL).

Is this meant to be like this, and also is it actually preferred way for same program to not share config if it's run from another place?

morganwillcock:
Standardizes settings path, regardless of assembly location (fixes #481)
Minimum version of irrKlang is now 1.6 (earlier versions were not signed).


rofl0r:
"strongname" ?

morganwillcock:
I like to think so.

ivan-mogilko:
Something that I would like to understand, if I keep using older version if irrKlang the editor builds and launches, and merely displays a warning ("referenced assembly does not have a strong name").
Earlier I've had an impression that you mean it won't work with non-sign assembly, but maybe I did not got what you meant right?

ivan-mogilko:
Oh, nevermind, I realized irrKlang only gets accessed when you try to actually play the sound in the editor.

ivan-mogilko:
Ah, did not notice this at first, but AGS.Native does not get signed in the Release configuration for some reason, what has to be changed to fix this?

morganwillcock:
Sorry about that, I didn't check both types of build.

Yes that should do it.
Project properties -> Linker -> Advanced -> Key File = "AGS.Native.snk"

Do you want me to do it with a new PR?

ivan-mogilko:
No, I fixed it accordingly 9c0169684d5ca4509aa5976685ea667bc2836002

PR #498, opened 2018-08-25
Hide recent games which are currently missing

morganwillcock:
Hides recent game references where the game path doesn't exist.

- When a game folder is removed or renamed, it now won't appear in the recent game list
- The entry remains in the settings file, so that if a game is on a volume which isn't mounted it will be listed again when the game folder returns

### DIFF
--- a/Editor/AGS.Editor/GUI/WelcomeScreen.cs
+++ b/Editor/AGS.Editor/GUI/WelcomeScreen.cs
@@ -1,9 +1,5 @@
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Drawing;
-using System.Text;
+using System.IO;
 using System.Windows.Forms;
 using AGS.Editor.Preferences;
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I understand the purpose, but when you mentioned unmounted volume I tried to imagine myself in the similar situation, and think that I'd be very puzzled as a user if the project I was working on is missing in the recent list. Could there be alternative options, like marking missing projects with color in the list?

morganwillcock:
It is more a byproduct of not tracking a unique ID and modification time for the game, that unless the old entries get rotated out you may be seeing multiple entries for the same game, and then are unable to tell the difference between any other game using the same name or a previously used path. For games on removable drives (e.g. USB stick), there isn't a guarantee that the drive letter is going to be the same, so after weighing it up I deemed that presenting a cleaner list (i.e. you can't have two games at the same path) was more beneficial that showing multiple projects and letting someone try to work out which one is relevant.

Presumably the GUID in the .agf file is unique per game given a single computer, but potentially someone else could have made a game with a duplicate GUID (if it was based on the time + random seed, I'm not sure how globally unique they are in reality)?

ivan-mogilko:
> Presumably the GUID in the .agf file is unique per game given a single computer, but potentially someone else could have made a game with a duplicate GUID (if it was based on the time + random seed, I'm not sure how globally unique they are in reality)?

Game GUID is generated using .NET's Guid.NewGuid: https://docs.microsoft.com/en-us/dotnet/api/system.guid?view=netframework-4.7.2
which is said to have "very low probability of being duplicated".

There is still a problem that copies of same game keep one GUID and AGS probably has no means to check that (at least it does not try currently).

PR #499, opened 2018-08-29
AGS 3: removed sprite number limit in the engine

ivan-mogilko:
This cleans up SpriteCache code and replaces const sized sprite arrays with vector containers.
Also sprite file format was updated a little again to make sprite count a 32-bit integer.

Theoretically this makes sprite numbers unlimited from the engine's perspective (well, limited by INT32_MAX for now). In practice though I kept a limit of the "static" (created at design time) sprites in the Editor. This limit is 30k as before, and may be raised simply by changing one constant in the editor code (or even made an option in the game settings). The dynamic sprites (created at runtime) are now not restricted and may exceed it.

The reason why I still kept some limit to static sprites is one problem which I explain below. I spent a while trying to figure out best way to deal with it, but could not make a decision. This is quite possible that I've had a moment of "obsessive perfectionism" or was missing an obvious solution, so did not want to waste more time and left it as it is for now.

---

The sprite list in AGS has following properties:
1. The list elements contains only brief sprite descriptions, like width, height, address in the data package, extra flags etc. The bitmap is allocated separately and the process of allocation/deletion of bitmap is controlled by sprite cache mechanism (so the main list only stores pointer to bitmap).
2. List may contain quite large number of items (currently 30k max, but in the past users were asking to increase it by 2-3 times at least);
3. the items are addressed by integer key;
4. the key may be customized by the user in the provided limits.
5. at the same time in the most games majority of sprites are probably arranged in big continious sequences with little to no gaps.
6. for most games majority of entries are added into the list only once (at the game load) and never removed; exceptions are games with heavy use of dynamic sprites.

With the random key **p4** the map or even better hash map seems like obvious choice. If the index is limited only by variable capacity one may have e.g. game with two sprites: '1' and '1000000', for which vector/array would have 999998 unused elements.

However, because of **p5** and **p6** hashmap's memory overhead becomes particulary silly; you won't have much benefit from hashmap if the game has 50k sprites with indexes going strictly from 0 to 49999.

Dynamic sprites make hashmap look even worse, because their indexes are assigned automatically in strict order (currently it first fills in the gaps in static sprites, but that's not necessary).

In this situation keeping array/vector but imposing some reasonable upper limit may be a good compromise.

In the past I was also considering more sophisticated solution: a type of container with the interface corresponding to hashmap, which internally stores elements in a list of arrays.
The size of arrays would depend on statistical spread of element indices. The hash would contain a pair of numbers combined: the number of array and the index of element inside array. For example, if the game has sprite 0 - 9999 and then 1000000 - 1009999, such container would create two 10k arrays and have zero memory overhead per element.
Unfortunately, I do not know if such type already exists and how it is called. Maybe there are similar specialized implementations of hashmap.

### DIFF
--- a/Common/ac/common_defines.h
+++ b/Common/ac/common_defines.h
@@ -94,7 +94,8 @@
 #define LEGACY_MAXOBJNAMELEN 30
 #define MAX_BSCENE    5   // max number of frames in animating bg scene
 
-#define MAX_SPRITES         30000
+#define LEGACY_MAX_SPRITES_V25  6000
+#define LEGACY_MAX_SPRITES      30000
 #define MAX_CURSOR          20
 
 #ifndef int32
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -77,21 +77,6 @@ void GameSetupStruct::read_font_flags(Common::Stream *in, GameDataVersion data_v
     }
 }
 
-HGameFileError GameSetupStruct::read_sprite_flags(Common::Stream *in, GameDataVersion data_ver)
-{
-    int numToRead;
-    if (data_ver < kGameVersion_256)
-        numToRead = 6000; // Fixed number of sprites on < 2.56
-    else
-        numToRead = in->ReadInt32();
-
-    if (numToRead > MAX_SPRITES)
-        return new MainGameFileError(kMGFErr_TooManySprites, String::FromFormat("Count: %d, max: %d", numToRead, MAX_SPRITES));
-    in->Read(&spriteflags[0], numToRead);
-    memset(spriteflags + numToRead, 0, MAX_SPRITES - numToRead);
-    return HGameFileError::None();
-}
-
 void GameSetupStruct::ReadInvInfo_Aligned(Stream *in)
 {
     AlignedStream align_s(in, Common::kAligned_Read);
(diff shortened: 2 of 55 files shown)


jeancallisti:
What I usually do for such allocation/disallocation tasks that require to quickly find a free cell is design a solution similar to what you explained: on the one hand there's the container, which is really an interface transforming any arbitrary index (0 to 1,000,000,000,000 or whatever) into the actual index of the items as it appears in the real storage object hidden inside of it. On the other hand, you have that sorage object(s) which can be anything you want as long as you may access their contents through indices. It can really be anything : a simple array or stack where you fill cells as they are needed, or a series of arrays (inside each of them you can keep continuous sequences of indices depending on your needs, but inbetween each array there are gaps in indices) or hashmaps with reasonable collision ratios.

The side question to this is : how do you spot a "free slot" in all of that?

 If you went for hashmaps as storage structures and/or if the sprite numbers can only go incrementally, then it sorts itself out.

However, if you went with series of arrays, or hashmaps i.e. the indices can be completely arbitrary (i.e. you need to fill holes and find them reasonably fast) then I usually keep an **index of free slots**.

This index is just a stack where I push the indices of the brand new or recently freed slots, and pop them when I need one. The plus side is that you may store any value there, not necessarily in an orderly fashion, and the values need not be continuous (for example you may store values 0 to 9 and 10,000,000 to 10,000,009 if these are the storage arrays that you created). The access is immediate. The downside is that the index is as big as the number of free slots. But then again if you've made free slots available in a non-continuous fashion (i.e. there are physical gaps inbetween them) then it's not much.

SpeechCenter:
> In the past I was also considering more sophisticated solution: a type of container with the interface corresponding to hashmap, which internally stores elements in a list of arrays.

The closest thing I know that does this is Hash Array Mapped Trie (HAMT). In a quick search I found [this C++ implementation](https://github.com/chaelim/HAMT). Looking at the code it seems the author has gone through the effort of optimizing it, but I didn't try it or compared it to alternatives.

rofl0r:
an important thing to consider nowadays is cache locality and memory access speed. sony has published an excellent guide about the topic: http://harmful.cat-v.org/software/OO_programming/_pdf/Pitfalls_of_Object_Oriented_Programming_GCAP_09.pdf
afaik, a C++ vector is internally implemented like a linked list, or at least an array with a pointer to the actual data, rather than having the data in a continuous array, which make memory lookups to often used items a lot slower. from that pov it's probably the best design performance-wise to have sprindex and spritedata in continuous array allocated at once (or optimally mmap()ed).
as for the issue with potentially huge gaps between used items, i could imagine a solution where we have a single array that translates spritenumber as seen from script to an index to a re-ordered array with gaps removed.
so if game requests e.g. sprite 2000, engine would do something like `load_sprite(spritedata[sprite_index_map[2000]])` instead of `load_sprite(spritedata[2000])`.

ivan-mogilko:
> afaik, a C++ vector is internally implemented like a linked list, or at least an array with a pointer to the actual data, rather than having the data in a continuous array

As far as I know, std::vector must be continious array by standard, so that if you get address of the first first element, you should be able to iterate that pointer yourself and achieve correct results.

PR #503, opened 2018-09-04
EditorConfig is awesome: http://EditorConfig.org

persn:
EditorConfig is the new trendy and hip config file that is reusable across various IDEs and text editors (check their homepage for a list)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+# Any files
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# C# code files
+[*.cs]
+indent_style = space
+indent_size = 4


ivan-mogilko:
> new trendy and hip config file
this pull request looks like its a spambot's doing :/

persn:
Yes, sorry I was being funny

ivan-mogilko:
Oh ok, :) that was a little different from usual.

ivan-mogilko:
I have two questions:
1. This won't automatically modify existing code's style, only the new one which you type?
2. Not sure how well the line ending definition will work; for example right now my git repository is configured with autocrlf = true (checkout Windows endings / commit Unix endings), won't there be conflicts with that? Perhaps leave line endings out and let git do them instead?

persn:
> This won't automatically modify existing code's style, only apply to the new one which you type?

Yes.

> Not sure how well the line ending definition will work; for example right now my git repository is configured with autocrlf = true (checkout Windows endings / commit Unix endings), won't there be conflicts with that? Perhaps leave line endings out and let git do them instead?

I haven't noticed any issues with this previously and I've been using editorconfig for a couple of years now. My git is set up the same as yours, I work in projects where developers are using Windows, Linux and Mac.

Issue #365, opened 2016-10-04
startgame script and directory names


SilasMortimer:
I was trying to troubleshoot an AGS game that I have on Steam, because it wasn't starting. What it boiled down to was that they were using the default `startgame` script (renamed), which has this line at the beginning:

`SCRIPTPATH="$(dirname "$(readlink -f $0)")"`

The game's directory has spaces in it and for some reason I still am not sure about, that line does not parse the path name with quotes. However, when the outer quotes are removed, like this:

`SCRIPTPATH=$(dirname "$(readlink -f $0)")`

... it works just fine. Steam user flesk pointed out that this line would be more robust:

`SCRIPTPATH=$(realpath "$(dirname "$0")")`

Which I've also tested with the game and it works. I should also note that other games based on AGS that didn't have spaces in their path work just fine with the original version of the line; only when there are spaces (and presumably other characters that require an escape) does it not work.

SilasMortimer:
I forgot to say that this issue exists in Linux.

ivan-mogilko:
I believe you are speaking about this file:
https://github.com/adventuregamestudio/ags/blob/master/debian/ags%2Blibraries/startgame

@tobihan and @monkey0506 were working on that kind of installation before, I think it would be better if either of them took a look into this.

SilasMortimer:
Yes, that is the file I'm talking about. Sorry about that - I forgot to link.

monkey0506:
Honestly, my experience with Linux is pretty localized to the use-cases I've had with building and deploying AGS games. Is the "more robust" script in some way preferable to removing the outer quotes? My personal thought would be that those outer quotes would actually *prevent* this problem, rather than *creating* it.

The `startgame` script has really only been touched by @tobihan, and it seems that this was his intention with 3960672.

monkey0506:
[This article](http://unix.stackexchange.com/questions/136494/whats-the-difference-between-realpath-and-readlink-f) (dated more than 2 years ago) suggests that `readlink -f` is better (and more consistently) implemented across Linux flavors, particularly noting that Debian has its own `realpath` that is inconsistent with the GNU implementation. I have no context for whether this has changed over the last two years.

SilasMortimer:
I think I might understand why removing the quotes works. It's because of this line: 

`ALLEGRO_MODULES="$SCRIPTPATH/data/lib64" "$SCRIPTPATH/data/ags64" "$@" "$SCRIPTPATH/data/"`

[...]

PR #518, opened 2018-10-10
Fixes for Linux launch wrapper (fixes issues with itch.io client)

morganwillcock:
- quoted $0 in launch wrapper
This fixes launching the game when there is a space in the path to launch wrapper (the itch.io launch failures are because `readlink -f` could return multiple lines from the sub-shell)

- ~~added exit 1 when printing usage~~
~~For some reason, it was still trying to start the game after printing the instructions~~

- `-h` is not a valid switch
Using this would have printed the usage header but still start the game

- Fix broken if statements and argument checks, when using arguments
   e.g. `./MyGame: 4: test: x--help: unexpected operator`

### DIFF
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetLinux.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetLinux.cs
@@ -104,11 +104,11 @@ namespace AGS.Editor
             {
                 results +=
 @"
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
@morganwillcock ,
I cannot tell because never investigated how this thing works, but there was this old issue regarding startgame script: #365, could you check it out to see if it's related?

PR #522, opened 2018-10-20
Ignore game name for recent game comparison

### LINKED ISSUE #521: AGS3: Editor adds two "recent games" entries when creating game from template
When you create a new game from template, editor seem to add two entries into "recent games" list, first with default template's name then with the actual name you gave it. This was never an issue before, which means this bug was added in 3.4.2/3.5.0 branch.

Steps to reproduce:
* Run AGS Editor and choose "Start a new game".
* When in "Start New Game" wizard choose any template and type any name. Press Finish.
* Save project and quit.
* Launch Editor again.
* Observe two new entries in the Recent games list, pointing to the same location.

morganwillcock:
Fixes #521 by comparing recent games using only their path on disk.

### DIFF
--- a/Editor/AGS.Editor/Entities/EditorPreferences.cs
+++ b/Editor/AGS.Editor/Entities/EditorPreferences.cs
@@ -57,7 +57,9 @@ namespace AGS.Editor.Preferences
 
         public bool Equals(RecentGame otherGame)
         {
-            return Name == otherGame.Name && Path == otherGame.Path;
+            // for lack of an identifier, if the path on disk matches
+            // this is considered to be the same game
+            return Path == otherGame.Path;
         }
 
         public string Name { get; set; }


ivan-mogilko:
Ignoring duplicates is one thing, but why does it get registered in the "recent" list twice at all?

ivan-mogilko:
I found why it was getting into recent list twice for the new game, the recent game list is updated _both_ when the game is loaded and saved, and so when the new game is created from template it first gets loaded and registered in recent list with Template's name, and then saved and tries to get registered again with the actual name user gave it:
https://github.com/adventuregamestudio/ags/blob/ags3/Editor/AGS.Editor/GUI/GUIController.cs#L973

ivan-mogilko:
Also, "Fixes" keyword no longer works... is it github issue or something needs to be set in our repository?

PR #526, opened 2018-11-03
Unified sprite import + reload from source

### LINKED ISSUE #500: Importing tiled sprites interface is a pain
I tried to work with it a little bit, it is really hard and inconvenient. Ended splitting the sprites manually and then importing one by one.

The problesm I see are:

1. Zoom level is limited and not always sufficient.
2. Can't pinpoint the top left corner (cant see which pixel the mouse is hovering over)
3. Can't see where is the left size.
4. Can't edit selection size after it was set.
5. It would be nice If there was a pixel grid to guide selection.
6. Need to repeat for each tile.

Many game engines allow importing grid images by specifying the width and height of each tile and a pixel offset between each two tiles. This might be a mice addition as well.

image

morganwillcock:
This is to help fix #500 and help facilitate changes to the sprite management.

Main changes:
- heavily modified sprite import UI
   - use the mouse to draw a single tile
   - use the tiling controls to specify how those dimensions are used
   - cropping an image is just a tiled import with the tile count set to 1
   - 'import whole image' is just not using the tiled import option
- removal of GIF and quick import options because everything is now processed as a potential image sequence (from multiple files, GIF frames, or both at once)
- Sprite class is now tracking extra import parameters, frame number, tiling offset, palette options, etc...

Good things:
- setting up a tiled import is easier (hopefully)
- you can specify a margin for tiled imports, so spritesheets can now have a bleed area between frames
- no prompts for "this image has an alpha channel, do you want to use it?", so a quick or GIF import is still quick, since you are just clicking import instead of answering this question
- no blind imports where you are importing with options that you can't see (Editor preferences are treated as a default, but you still get the UI to confirm them)
- all imports are potentially multi-frame, so you can you can do a tiled import of multiple files, or of an animated GIF, or both at the same time
- reload of any source that came from a file (a tile within an image, a frame within a GIF sequence)

What isn't so great:
- a lot of duplicate code in the sprite manager for the moment, related to getting / setting the new Sprite class data
- sprite manager performance is not good, so if you mistakenly do a tiled import of a GIF animation and generate 1000s of frames, it'll hang the Editor for a substantial period of time
- I can't understand how the Sprite class parameter for locking a palette to a room is used, or even how a room number is set. This is still hard-coded to `false` for reload purposes.
- new import UI probably needs tweaking a bit

Since this does change the sprite workflow significantly I was hoping that someone can check that the class / XML changes are okay and then this can get built for people to try out; I don't want to accidentally have removed some critical function, but testing every type of format with every type of import method has been tricky.


morganwillcock:
I've rebased this to back out some changes for sprite folders. It looks like the majority of performance issue was calling sprite folder functions 1000s of times rather than the sprite folder functions themselves causing the delay.

ivan-mogilko:
I started actually testing this now, and there is one big question I have prior to anything else:
Is it intentional that the sprite source properties may be freely edited by user in the table? Is it suppose to update the sprite whenever user changes them?

Couple of other things I quickly noticed:
* Is the "Frame" property '1' by default? All the sprites that do not come from GIF are now said to have Frame 1 (I would expect 0, unless there's misunderstanding, or "None").
* Perhaps "ImportMethod" field may be called "TransparencyMethod", "TransparencyOption" or simply "Transparency"? IMHO "Import method" sounds way too "general" for this.


The new import dialog is overwhelming, so I'll test it thoroughly a bit later. One thing though: I wonder if it would be possible to drag large pictures around along with using scrollbars (holding one of the mouse buttons).

morganwillcock:
> Is it intentional that the sprite source properties may be freely edited by user in the table? Does the sprite manager supposed to update the sprite whenever user changes these?

It is intentional, although it is reliant on the user understanding that these are import options rather than editing the live sprite (which is why I placed them under the heading 'import'). So this would be where you can manually repair sprite cache links (edit a sprite source) or actively use on the options to modify the sprite (fix an offset, or get frame 2 in place of frame 3, etc). So whether this is a good idea probably depends on how the sprite cache rebuild would be handled, when sprites are missing. If you wanted to manually fix some sprites by mapping them back against a spritesheet, you'd need these things to be editable.

> Is the "Frame" property '1' by default? All the sprites that do not come from GIF are now said to have Frame 1 (I would expect 0, unless there's misunderstanding, or "None").

That is the default yes, but more a by-product of the import process handling multiple frames by default, plus a sprite cache repair possibly needing to have the frame number altered. Starting from 1 was for two reasons, firstly because it lets you change the source filename whilst keeping a valid frame number, secondly I think the Editor is already pretty inconsistent in it's use of 0 (I'm fairly sure in some places sprite 0 means the actual sprite in slot 0, whilst in others places sprite 0 means 'no sprite'). Changing it be 0 is probably pretty easy though.

> Perhaps "ImportMethod" field may be called "TransparencyMethod", "TransparencyOption" or simply "Transparency"? IMHO "Import method" sounds way too "general" for this.

It is tricky, since this options seems to relate to handling of transparent colours and has nothing to do with the alpha channel. As long as it is clear that the alpha channel is separate, anything is fine by me. Pixel0 is also a really bad name for the colour in index 0, since it sounds more like you are taking the colour of the first pixel.

> One thing though: I wonder if it would be possible to drag large pictures around along with using scrollbars (holding one of the mouse buttons).

In theory, yes. The right mouse button is now entirely free, and just duplicating the action of the left, for the moment.

ivan-mogilko:
Another big question I wanted to ask, how intertwined is the new sprite import properties code and the new import GUI? As in, would it be complicated to make them two separate pull requests?
The reason I ask is that logically they are separate things, and the testing would be easier separately; perhaps, first the import properties, then the new GUI. Besides, import requires mostly a code review while editor would benefit from real user feedback, which may take more time.


> > Perhaps "ImportMethod" field may be called "TransparencyMethod", "TransparencyOption" or simply "Transparency"? IMHO "Import method" sounds way too "general" for this.

> It is tricky, since this options seems to relate to handling of transparent colours and has nothing to do with the alpha channel. As long as it is clear that the alpha channel is separate, anything is fine by me. 

Hmm, idk, what about "TransparentColorPick"?

> Pixel0 is also a really bad name for the colour in index 0, since it sounds more like you are taking the colour of the first pixel.

Perhaps, Palette0, or PaletteSlot0?


---


Unfortunately I still was not able to dedicate much time testing this further yet, but I may note couple of more things:

Regarding import settings:
- BUG: after doing tiled import *every* resulting sprite has the same OffsetX/Y values.
- RemapToGamePallete is always set to True for some reason, although the game nor sprite are not 8-bit. Is that supposed to be so?
- Do you think that may be a good idea to disable SourceFile field direct editing, but have something like "..." button opening file selection dialog? I do not remember if AGS has such fields anywhere, but there are fields that open custom dialogs, such as Game Resolution, Custom Properties, etc.

Regarding import dialog:
- Just a note: the color rectangles showing transparent color preview is an interesting addition, but they may be misleading when doing tiled import (since each tile has its own corner pixels).
- Perhaps let user to drag inverse rectangles too (from bottom-right and up-left)?
- TBH I like it how AGS let you define number of tiles by moving mouse around.

morganwillcock:
> Another big question I wanted to ask, how intertwined is the new sprite import properties code and the new import GUI? As in, would it be complicated to make them two separate pull requests?

It would be pretty tricky, since there were multiple ways before where you can bypass the UI or import sprites using values you can't see (either assumed based on the colour depth, or related to the last time you used the UI). When I started I only wanted to re-do the tiling controls, but this mandated storing more data.

>Hmm, idk, what about "Trans [parent] ColorPick"?
>Perhaps, Palette0, or PaletteSlot0 or PaletteIndex0 (I think you are already using latter in the GUI)?

I'll try and change the names.

>BUG: after doing tiled import every resulting sprite has the same OffsetX/Y values.

Do you know if this was for every type of import, or just from a certain source?

>RemapToGamePallete is always set to True for some reason, although the game nor sprite are not 8-bit. Is that supposed to be so?

I think I copied what the original import was doing, but would have to go back through to check.

>Do you think that may be a good idea to disable SourceFile field direct editing

I did consider it. Adding a 'browse' button that opens a dialog would be useful, but locking the field to read-only would mean you wouldn't be able to copy and paste the values. So for a 'cache repair', it is probably a lot faster to edit the fields directly rather than restrict it to 'browse only'.

>Just an observation: the color rectangles showing transparent color preview is an interesting addition, but they may be misleading when doing tiled import (since each tile has its own corner pixels).

Yes, I think I've currently missed setting the colour value for Index 0 too. Maybe moving the mouse over a tile could pop-up a display of the corner colours for that tile. I'll have a think.

>Perhaps let user to drag inverse rectangles too (from bottom-right and up-left)?

I can give it a go, but I was having a bit of an ordeal with zooming and the scroll bars. I thought there would be a better built-in way to display a bitmap with some basic editing controls, but apparently not. It would probably need a different approach, because the mouse control is effectively just a shortcut to filling the values directly (see below)

>TBH I liked it how AGS did let you define number of tiles by moving mouse around.

Personally I thought that 'Importing tiled sprites interface is a pain' (#500) was putting it very politely, and how it worked before was akin to trying to cast some magic spell with the mouse. I had a look at the import UIs in some other software, before I started, and the emphasis seems seems to be on allowing the user to enter the data for their spritesheet (so the drawing of a tile is just a quick way to do this, if the visible part of the image allows you to do it).

spriteslicepanelgrid

Plus, if the reloading of spritesheet components works, the number of times this interface is visited should be greatly reduced (emphasis on accurately setting the values once, rather than a fast way to keep redefining the tiles with the mouse, every time the spritesheet source is updated externally). I think this is why the UI is so bare-bones in most other programs, since you only use them once per spritesheet, rather than continually keep redefining the sizes.

ivan-mogilko:
> > BUG: after doing tiled import every resulting sprite has the same OffsetX/Y values.

> Do you know if this was for every type of import, or just from a certain source?

Hmm, this is an interesting question...
This happens with PNG and BMP files. I draw a tile rectangle, type in a number of tiles, click Import. Every tile has OffsetX/Y equal to the top-leftmost one. Reimporting them using "Replace from source" naturally causes them all to have top-leftmost tile content.


Occasionally, I found that importing from the animated GIF (which I opened by mistake) will create number of sprites equal to number of tiles * number of frames (probably). Not sure if this is intententional or not.

morganwillcock:
> This happens with PNG and BMP files. I draw a tile rectangle, type in a number of tiles, click Import. Every tile has OffsetX/Y equal to the top-leftmost one. Reimporting them using "Replace from source" naturally causes them all to have top-leftmost tile content.

This should be fixed now.

> Occasionally, I found that importing from the animated GIF (which I opened by mistake) will create number of sprites equal to number of tiles * number of frames (probably). Not sure if this is intentional or not.

That it is intentional, as the GIF file is potentially processed as multiple images, so it is doing the same thing as if you had selected two or more non-GIF images and also defined a tile.

ivan-mogilko:
Sorry for delay, I will be testing this soon again. Meanwhile I found the old ticket on the forum's tracker, which suggests to hide/disable "Palette index 0" option when not working with 8-bit gfx:
http://www.adventuregamestudio.co.uk/forums/index.php?issue=841.0

morganwillcock:
Do you know if the native function that performs the import would easily be converted to being given an actual colour, rather than an instruction on where to find the colour? It seems the colour representation in the engine is only 16 bit (i.e. the colour number is an unsigned 16bit integer), is this why you cannot just specify the colour you want to make transparent?

morganwillcock:
Thanks. I'm trying to clean it up a bit and get all the import functions out of the sprite manager at the moment, which will hopefully allow the import window to also process individual images (which makes more sense you choose to import some files that are different formats). Once I've got it cleaned up enough, I'll check back with you about the import colour.

morganwillcock:
This is rebased to merge, but I still can't see how the 8 bit palette can be locked to a particular room, or how this was ever set previously. The data type is storing an integer and describing it as a room number, the native proxy is taking a boolean. @ivan-mogilko, any ideas?

I think the rest of the interface is as good as I can get it, re-using the existing functionality.

ivan-mogilko:
> This is rebased to merge, but I still can't see how the 8 bit palette can be locked to a particular room, or how this was ever set previously. The data type is storing an integer and describing it as a room number, the native proxy is taking a boolean. @ivan-mogilko, any ideas?

I never investigated this, but will do now. Might also consult Scavenger on how he uses this setting in his 8-bit games.

ivan-mogilko:
Alright, from what I found:
- when you load up any room in the editor, editor remembers its palette;
- if you import any sprite and check the "use background palette" it will remap the sprite colours to this active room palette.
This means you must have a room loaded to use this option.

Editor saves room index in the sprite for the reference, but native library does not seem to care, it just needs to know whether to do remapping or not, because it matters only at the moment of import, and what room is loaded at the moment of import (right now this may be only one room). Hence - it passes the boolean (either remap to background or not).

I suspect that in order to support full sprite reimport one would have to either -
1) load actual rooms for every such sprite (dumb solution);
2) or make a function to extract palette data for a necessary room from compiled file; which should be doable.
for optimisation it could somehow pass the sprites in the order of the rooms, or keep room palettes in a dictionary, but I guess that's a question of implementation.

morganwillcock:
Thanks, I think I've found the relevant bits and it looks like the palette information is a side-effect of the room loading, so no function takes the palette directly.

Did you mean to apply the room palette to the image in the Editor before it gets to the native function, and modify the native function to just store the room number rather than alter the image?

ivan-mogilko:
> Did you mean to apply the room palette to the image in the Editor before it gets to the native function, and modify the native function to just store the room number rather than alter the image?

Hm, no, I think  native function must alter the image because this is how correct bitmap gets into compiled sprite pack, and otherwise it won't look properly in the sprite manager itself. Also, that won't make much sense to keep room number with native sprite data because when in game they just rely on current room for palette anyway.

What I mean is to keep sprite data as it is now, but write a special processing for 8-bit images with this "room number" != 0 when you do reimport from source. Native function may actually be rewritten to take palette as an argument and so become independent from the active room.

It seems that reimport from source file was never working correctly for 8-bit sprites so this is going to be something completely new. Maybe do this as a separate task? If you feel that this pull request is becoming too large already.

morganwillcock:
I've added the option to use the room background palette to the sprite class, so as far as I know, all of the hard-coded import settings are now removed and the only options used will come from the import window (or back from the sprite, when reloading).

Since I don't think I've removed any functionality, I'd probably prefer to get this merged and give it to some people to try, as long as you think the data / serialization changes are okay?

ivan-mogilko:
I see that you write both ColoursLockedToRoom (room number) and RemapToRoomPalette (boolean). Is not it enough to store only ColoursLockedToRoom? Unless I missed something in theory the presence of room number is indicating that remap to room palette was on.

ivan-mogilko:
On second thought, is your idea to keep all options enabled at all times because for example the source image may be manually changed between 8bit and non-8bit?

ivan-mogilko:
I have an idea, although I can't tell if it's the good one.

What if you remove this separate boolean RemapToRoomPalette property and instead move ColoursLockedToRoom to the "Import" category and make it editable. Then, if user does "reimport from source" and it's a wrong room the editor will refuse to reimport this particular sprite and tell to open correct room first.
Later when (if) we implement automatic palette extract from the room this behavior may be adjusted.

What do you think about this?

ivan-mogilko:
On the other hand, we may simply start ignoring the "RemapToRoomPalette" field (don't read and write it) if it won't be needed after 8-bit reimport is fixed, so I guess it's not a big deal?

PR #528, opened 2018-11-04
Fix number dialog for numbers greater than control default

morganwillcock:
This was crashing for numbers greater than 100, because the value was set before the max value was adjusted.


ivan-mogilko:
I could swear this was fixed couple of months ago, are you sure your ags3 branch is updated?
udNumber.Value = initialValue;

PR #529, opened 2018-11-05
Update colour message for palette editor

morganwillcock:
Presumably this text predated 32 bit colour games?


ivan-mogilko:
> "This palette information will only be used for drawing any 8-bit graphics that you may have imported."

Yes, I think that makes it clear.

morganwillcock:
Okay, I've re-done it.

PR #538, opened 2018-12-06
Add user templates folder and remove template wizard

morganwillcock:
* Since the installation directory may not be writable, this adds a second 'Templates' folder in the user profile and checks it for room and game templates. The export of templates now just uses a save file dialog that defaults to the new 'user' templates directory
* The editor isn't forced closed if the templates folder isn't present - I don't really know why it was doing this, since you can still open an existing game, and crashes from it not being there were just because there was limited exception handling on the directory search for template files
* I also updated the theme and layout paths to reference the common local appdata property

Basically, this makes templates work without the need for admin rights, if you've installed the Editor into somewhere that you cannot write as a regular user.


ivan-mogilko:
This is long due, but I have concerns about two things in current implementation:

1. Is it okay to have templates in a user data folder as opposed to public data folder? This changes "privacy level" of templates, because previously the templates you save could be used by any user on the computer and now these are only accessible to their creator (unless he/she moves them to program directory).

2. Is it okay to put all templates in 1 folder meaning all installed versions of AGS will see same list of templates? Actually, this may have both good advantage (no need to copy created templates over to each new version of AGS) and possible disadvantage.

PR #545, opened 2018-12-20
Spritesheet import: inverted mouse selection and fixes

morganwillcock:
- Allows selection to invert when dragging, so you can now make a tile selection with the left mouse button from bottom-right right to top-left, as well as top-left to bottom-right. The inversion is inclusive of the first pixel, which feels correct in practice.

select

- Clamp selection size to image edges. Previously cursor position was still tracked, even after leaving the control.
- Fix crashes where the selection height or width was set to 0.
- Don't updated selection sizes when switching images. Invalid selection sizes are handled, and it is more likely that someone with a spritesheet will want to work from the top-left corner.
- Fix copy sprite to clipboard (regression)

### DIFF
--- a/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
+++ b/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
@@ -334,8 +334,8 @@ namespace AGS.Editor
             // Draw dragging indicator
             if (dragging)
             {
-                numSizeX.Value = position.X / zoomLevel - numOffsetX.Value + 1;
-                numSizeY.Value = position.Y / zoomLevel - numOffsetY.Value + 1;
+                numSizeX.Value = Math.Min(image.Width - numOffsetX.Value, position.X / zoomLevel - numOffsetX.Value + 1);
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Not directly related to this pr, I noticed that there are few operations that often fail with exception. They are one-liners and instead of testing the condition preliminary they are wrapped in a try/catch block letting exception to happen.

Examples:
<code>
            try
            {
                frames = SpriteTools.GetFrameCountEstimateFromFile(imageLookup[cmbFilenames.SelectedIndex]);
            }
            catch
            {
                frames = 1;
            }
</code>
<code>
            try
            {
                panelIndex0.BackColor = image.Palette.Entries[0];
            }
            catch (IndexOutOfRangeException)
            {
                // not an indexed palette
            }
</code>

Why is this considered a better way compared to e.g.
<pre>
if (image.Palette.Entries.Count > 0)
     panelIndex0.BackColor = image.Palette.Entries[0];
</pre>
?

morganwillcock:
The second was probably left over from seeing which exceptions would get generated, so if only interested in IndexOutOfRange just checking the count would be better.

The first one uses the GIF decoder internally, and the control may be missing the value, and I just needed to guarantee a fallback value of one.

PR #547, opened 2018-12-21
AGS 3: clone RoomStruct class refactor from ags4 branch

ivan-mogilko:
Previously I committed room class refactor (#484) but I made a mistake merging it only to the ags4 branch (probably still did not think ags3 will be developed for long? idk).

Now I am "backporting" it to ags3 (with added legacy feature support bits ofcourse). In my opinion this will benefit this branch too since cleaner code is always good, and room loading functions had some very messy code, as well as some of the member variables had misleading names (like walk-behinds called "objects" and so on).

As previously I note that although a lot of files are changed in this pull request, most of these changes beyond roomstruct.cpp and room_file.cpp are actually just class members renamed.


ivan-mogilko:
Fixed some mistakes, also fixed them in the master branch (probably no one was using it since September).

PR #549, opened 2018-12-25
Script API: helper functions for Viewport and Camera

ivan-mogilko:
This adds a bunch of script functions to work with Viewport and Camera.

But first of all, I added Screen struct to place static screen-related properties and functions there. At the moment its declaration looks like this:
<pre>
builtin struct Screen {
  /// Gets the width of the game resolution.
  import static readonly attribute int Width;
  /// Gets the height of the game resolution.
  import static readonly attribute int Height;
  /// Gets/sets whether the viewport should automatically adjust itself and camera to the new room's background size
  import static attribute bool AutoSizeViewportOnRoomLoad;
  /// Gets the primary room viewport
  import static readonly attribute Viewport *Viewport;

  /// Returns the point in room which is displayed at the given screen coordinates
  import static Point *ScreenToRoomPoint(int sx, int sy);
};
</pre>

You may notice there's now Point struct in the builtin API. It's declaration is trivial:
<pre>
managed struct Point {
	int x, y;
};
</pre>
So far this is the only managed struct that is intentionally not declared as "builtin", because I thought it would be good to let users also spawn its instances in script.

* Moved RoomViewport and AutoSizeViewportOnRoomLoad properties from Game to Screen struct.
* Deprecated System.ScreenWidth and System.ScreenHeight, as well as System.ViewportWidth and System.ViewportHeight in favor of Screen.Width/Height.

Added following to the Viewport:
<pre>
  /// Returns the viewport at the specified screen location.
  import static Viewport *GetAtScreenXY(int x, int y);
  /// Changes viewport's position on the screen
  import void SetPosition(int x, int y, int width, int height);
  /// Returns the point in room corresponding to the given screen coordinates if seen through this viewport.
  import Point *ScreenToRoomPoint(int scrx, int scry, bool clipViewport = true);
  /// Returns the point on screen corresponding to the given room coordinates if seen through this viewport.
  import Point *RoomToScreenPoint(int roomx, int roomy, bool clipViewport = true);
</pre>

**Removed ScaleX and ScaleY from Camera.**

Added following to the Camera:
<pre>
  /// Changes camera position in the room and disables automatic tracking of the player character.
  import void SetAt(int x, int y);
  /// Changes camera's capture dimensions in room coordinates.
  import void SetSize(int width, int height);
</pre>

**NOTES:**
* Screen.ScreenToRoomPoint returns null if there is no room Viewport under these coordinates. If there will be multiple viewports supported in the future engine simply won't know which to choose if there's none directly there.
* Both Viewport.ScreenToRoomPoint and Viewport.RoomToScreenPoint have clipViewport parameter which tells whether they should return null if the coordinates lie outside of the visible part of the room.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -365,6 +365,12 @@ enum eKeyCode
   eKeyF12 = 434
 };
 
+#ifdef SCRIPT_API_v350
+managed struct Point {
+	int x, y;
+};
+#endif
+
 #define CHARID int  // $AUTOCOMPLETEIGNORE$
 builtin struct ColorType {
   char r,g,b;
@@ -650,12 +656,6 @@ builtin managed struct Game {
   readonly import static attribute int AudioClipCount;
   /// Accesses the audio clips collection.
   readonly import static attribute AudioClip *AudioClips[];
-#endif
-#ifdef SCRIPT_API_v350
-  /// Gets/sets whether the viewport should automatically adjust itself and camera to the new room's background size
-  import static attribute bool AutoSizeViewportOnRoomLoad;
-  /// Gets the room viewport
-  import static readonly attribute Viewport *RoomViewport;
 #endif
 };
 
@@ -2042,6 +2042,7 @@ builtin managed struct AudioClip {
 };
 
 builtin struct System {
+#ifdef SCRIPT_COMPAT_v341
   readonly int  screen_width,screen_height;
   readonly int  color_depth;
   readonly int  os;
@@ -2050,6 +2051,7 @@ builtin struct System {
   readonly int  viewport_width, viewport_height;
 #ifndef STRICT_STRINGS
   readonly char version[10];
+#endif
 #endif
   /// Gets whether Caps Lock is currently on.
   readonly import static attribute bool CapsLock;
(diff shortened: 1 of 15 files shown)


ivan-mogilko:
So, the recent conversation with @ericoporto on forums led me to a thought to completely remove the Scale* properties from the camera.

Initially I added these properties as means of automatic camera resize. But thinking this over, this currently lead to more problems. In AGS right now the camera's position is defined as a top-left of a rectangle, and, for example, when tracking the player character camera is placed at character's location minus half-size of camera. If the camera's size is not fixed (as when defined by autoscale) we need to put both camera and viewport into calculation. Which is frankly a hack and may cause troubles when (if) multiple viewports/cameras are supported. Even now camera's true position values become undetermined without knowing viewport. So this was something not thought through very well in the first place.

ivan-mogilko:
On other topic, what do you think about GetRoomLocation/GetScreenLocation function names?
I have doubts because there are other script functions with term "location": GetLocationName and GetLocationType, where "location" means some entity inside the room (object, hotspot, etc).
What could be alternate good variants for functions that convert coordinates?

ericoporto:
If we consider the Room as our game world, we have a RoomPoint, and a ScreenPoint, where transforming from RoomPoint implies knowing the X,Y RoomPoint, the Camera, and the Viewport to findout the x,y ScreenPoint. I think the X and Y case has similar meaning in AGS too.

ivan-mogilko:
Maybe ScreenToRoomPoint/RoomToScreenPoint, without "From"? Just trying to shorten them a bit :).

ericoporto:
Yeah, that works! One could also use coordinates, but point is shorter!

Issue #542, opened 2018-12-17
AGS 3: support recreating sprites that were imported from clipboard


ghost:
This is related to the problem of sprites (among few other things) being stored in a binary blob (#463) which makes it difficult to treat them individually under source control.
#526 has made it possible to recreate/reimport tiled sprites from the source. Now AFAIK the only problem remaining is sprites imported from the clipboard. Such sprites do not leave any actual source behind so if compiled sprite pack is lost they also cannot be restored.

The only solution that comes to my mind is to generate bitmaps in a certain directory inside the game project and link these as import sources to the new sprites. Or maybe even create them first and import them normally as if they were chosen in the import dialog.

What's required for this:
1. Define directory name and bitmap naming convention.
2. Run bitmap export after sprite is imported from clipboard.
3. After bitmap(s) are successfully created, link them to the new sprites.

Questions:
* If user does tiled import from clipboard, should we save full image or only individual tiles?

See Also: #1281

morganwillcock:
> 1. Define directory name and bitmap naming convention.
> 2. Run bitmap export after sprite is imported from clipboard.
> 3. After bitmap(s) are successfully created, link them to the new sprites.

You could potentially do this manually after import. If there will be a function to export the entire sprite cache to disk, one of the options on it could be 'export all sprite which have no source file', with a second option 'set exported files as source'. This way you don't need to pick a specific directory and everything would align with how the other sprites are managed, you would only need pick a filename format for the export process itself. Tokens for all the data retrievable from the spritecache would probably do:

e.g. "%spritenumber%_%width%x%height%.%format%" exports as "120_320x200.bmp" for sprite number 120 that is 320px wide and 200px tall.

Also, those tokens could then be used anywhere in the path (so sprite number could be created as a directory, for example).

> If user does tiled import from clipboard, should we save full image or only individual tiles?

The reload for tiles would expect the full image as the logic is just a rectangle generator, unless there was metadata available that described the image. But if storing metadata, it is probably better to define the spritesheet selections for the full image rather than identify a piece of an image.

ivan-mogilko:
> You could potentially do this manually after import. If there will be a function to export the entire sprite cache to disk, one of the options on it could be 'export all sprite which have no source file', with a second option 'set exported files as source'.

What I wish the editor to do is a uniform sprite handling where every sprite is stored similarily no matter where it came from. The idea was to gradually move to having sprites on disk as a primary source rather than the packed sprite cache.

[...]

PR #551, opened 2018-12-25
Tokenised sprite export

morganwillcock:
Possible solution to #542, where sprite properties can be used as tokens in the export path. This adds a sprite dialog option, where you can edit the tokens, skip sprites that have an existing source file, and rewrite the source file to use the exported file.

The tokens would probably need to be listed somewhere, but they are just the property names with '%' at either end, case-insensitive. So '%NUMBER%', '%number%', and '%NuMbEr%' would resolve to the sprite number, for example.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -150,6 +150,12 @@
     <Compile Include="GUI\SplashScreen.Designer.cs">
       <DependentUpon>SplashScreen.cs</DependentUpon>
     </Compile>
+    <Compile Include="GUI\SpriteExportDialog.cs">
+      <SubType>Form</SubType>
+    </Compile>
+    <Compile Include="GUI\SpriteExportDialog.Designer.cs">
+      <DependentUpon>SpriteExportDialog.cs</DependentUpon>
+    </Compile>
     <Compile Include="GUI\TextEntryDialog.cs">
       <SubType>Form</SubType>
     </Compile>
@@ -320,6 +326,9 @@
     </EmbeddedResource>
     <EmbeddedResource Include="GUI\Progress.resx">
       <DependentUpon>Progress.cs</DependentUpon>
+    </EmbeddedResource>
+    <EmbeddedResource Include="GUI\SpriteExportDialog.resx">
+      <DependentUpon>SpriteExportDialog.cs</DependentUpon>
     </EmbeddedResource>
     <EmbeddedResource Include="Panes\BigPropertySheet.resx">
       <SubType>Designer</SubType>
--- a/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
+++ b/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
@@ -1,0 +1,241 @@
+﻿namespace AGS.Editor
+{
+    partial class SpriteExportDialog
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Windows Form Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            this.groupBoxExport = new System.Windows.Forms.GroupBox();
+            this.chkSkipValidSpriteSource = new System.Windows.Forms.CheckBox();
+            this.chkRecurse = new System.Windows.Forms.CheckBox();
+            this.radRootFolder = new System.Windows.Forms.RadioButton();
+            this.radThisFolder = new System.Windows.Forms.RadioButton();
+            this.groupBoxSaveAs = new System.Windows.Forms.GroupBox();
+            this.btnBrowse = new System.Windows.Forms.Button();
+            this.txtFolder = new System.Windows.Forms.TextBox();
+            this.txtFilename = new System.Windows.Forms.TextBox();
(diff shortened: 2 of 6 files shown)


ivan-mogilko:
Some bugs I found:

1. Context menu works very unreliably. It worked the first time I opened this dialog, yet it stopped working at some point.
UPD: debugging shows that contextMenuStripExport.SourceControl may be null, so it does not know where to insert the token.
Pasting usually works (not 100%), and sometimes I was able to insert tokens right after pasting but then it broke again.
2. Is this intentional that insert token command also works in the "Folder" text box?


Usability issues:

1. Dialog does not remember last settings.
2. Since with this dialog export requires a little more clicking, and browsing for the destination is something that has to be done anyway, I wonder if that will work if the browse dialog would open automatically as soon as this window is on screen?
This way user will first choose a path, then is able to modify export settings and confirm.
3. I think that "Folder" label may be confusing since there are also sprite folders in this context. Maybe rename it to "Path" or "Export to"?


Since this is a new dialog window, I also have some nitpicking regarding the UI because it looks inconsistent with other dialogs in the editor:

1. The order (and sizes) of Close/Export buttons is inconsistent with other dialogs in AGS where they have OK/Cancel (e.g. compare with Preferences window or New room dialog).
2. Why the "Set exported file as source" checkbox is located separately and not in the same group with other options ("Skip sprites..." etc)?
3. Everywhere else the Browse (files) command seem to be made as a small "..." button to the right of the path, here it is made as a larger button below the path.
4. Usually the labels before text boxes are ending with ":", but not in this window.
5. Idk if resizing ability is important for this dialog, but being able to resize it vertically does not give any benefit (there are no controls that are scaled vertically, only ones that scale horizontally).
Also, it could be maximized and minimized.

ivan-mogilko:
> Are you right-clicking anywhere in particular to get a null SourceControl, or do you have any utility running that might be stealing focus?

Oh, I was able to pinpoint this. The focus is lost if I **click on** the "Insert token" submenu item instead of just hovering over it.

BTW, is the dialog made "always topmost" now? That's pretty annoying to be honest.

morganwillcock:
> Oh, I was able to pinpoint this. The focus is lost if I click on the "Insert token" submenu item instead of just hovering over it.

Thanks for tracking it down, I'll have another look later today.

> BTW, is the dialog made "always topmost" now? That's rather inconvenient to be honest.

I wasn't sure if it should have that set or not, but I'll put it back if you think it shouldn't be set.

ivan-mogilko:
> I wasn't sure if it should have that set or not, but I'll put it back if you think it shouldn't be set.

The thing is that the form is considered separate application window (I forgot what was the property that does that) and probably because of that it becomes topmost among *all applications* and stays on screen even when you switch out of the AGS.

morganwillcock:
> Oh, I was able to pinpoint this. The focus is lost if I click on the "Insert token" submenu item instead of just hovering over it.

Maybe it is something to do with the difference in text encoding and/or .NET framework version, but I'm clicking everything I can and I can't get it to go wrong. Do you think you could record your screen, just so I know for sure which bit you are clicking on? I've just redone the window behaviour/styling, so perhaps that will make a difference.

ivan-mogilko:
Hm, no it's not fixed yet on my side. I made a video, but not sure if there's anything useful there:
https://www.dropbox.com/s/tmt4z89p6gur4m9/ags-test-spriteexportdialog.flv?dl=0
It works when I simply hover over the "Insert token" to let the submenu open, but if I click on this item then it breaks.

morganwillcock:
> It works when I simply hover over the "Insert token" to let the submenu open, but if I click on this item then it breaks.

It seems the behaviour is changed depending on .NET framework version, so I guess it was eventually treated as a bug and fixed:

[https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/retargeting/4.7.1-4.7.2](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/retargeting/4.7.1-4.7.2l)

"In the .NET Framework 4.7.1 and previous versions, the ContextMenuStrip.SourceControl property incorrectly returns null when the user opens the menu from nested ToolStripMenuItem controls. In the .NET Framework 4.7.2 and later, SourceControl property is always set to the actual source control."

> Also, there was probably a misunderstanding regarding the "browse" button. I meant that the "..." buttons are usually placed to the right of the text box, but if you like to place it below for some reason then better leave it with "Browse" text (otherwise it looks strange).

I couldn't find a way to put the button on the right and also have the window grow to the right, whilst keeping alignment with the text field above, so eventually I put it back underneath. I'll change it back to "browse" if it is an issue.

ivan-mogilko:
> I couldn't find a way to put the button on the right and also have the window grow to the right, whilst keeping alignment with the text field above,

Oh! I gave it a try and found following combination after some experimentation:

Filename textbox: anchors: left, top, right;
Directory textbox: anchors: left, top, right;
The "..." button: anchors: top, right (no left!)

morganwillcock:
How does that look?

ivan-mogilko:
Yes, it's okay now.

morganwillcock:
> Is this intentional that insert token command also works in the "Folder" text box?

> Since with this dialog export requires a little more clicking, and browsing for the destination is something that has to be done anyway, I wonder if that may work if the browse dialog will be opened automatically as soon as this window is on screen? This way user will first choose a path, then will be able to modify export settings and confirm.

> Dialog does not remember last settings.

These issues are sort of connected, since the token can work anywhere in the path, and using the browse button will give you a path that won't have any tokens in it. Unless you have used the dialog before, you wouldn't know the use for the path you are choosing if it automatically opened the directory chooser, whereas launching it manually at least gives you a chance to see which field you are setting before it is set. Technically, by being blank it is defaulting to the project folder everytime it opens, but I could make an Editor preference for it if you like? If this was an HTML text field I would set "placeholder" text to indicate it defaults to the project directory, but I couldn't find an equivalent for a Winform TextBox.

> Why the "Set exported file as source" checkbox is located separately and not in the same group with other options ("Skip sprites..." etc)?

This was to separate sprite selection from modifying the sprite. Everything in the "Export from" group box is only concerned with which sprites you will export, whereas the option "Set exported file as source" is acting upon that selection.

ivan-mogilko:
Alright then, am I right to assume this may be merged in current state?

morganwillcock:
I wasn't planning on adding anything extra, so I think so.

PR #576, opened 2019-01-04
Refactor audio player

### LINKED ISSUE #567: Audiofiles: Always Show Length in Player
at the moment the length is only displayed when play is clicked.

ALSO: when clicking stop the position/ length is back to 0 / 0 and not 0 / length!

morganwillcock:
Fixes #567

- Specifically request the software driver for irrKlang sound engine (on my computer at least, using DirectSound seems to sometimes cause invalid values to come back from Sound data on all but the first play through)
- Show AudioClip length without having to press play (this also splits the position and length text between two labels, and makes them bigger)
- Fix MIDI polling / stop events confusing the stop and pause buttons of the GUI
- Fix MIDI pause and resume resetting all MIDI instruments to channel 0 (if you try this currently, every instrument will become a piano when resuming)
- Throw errors for playback and loading problems

### DIFF
--- a/Editor/AGS.Editor/Audio/IAudioPreviewer.cs
+++ b/Editor/AGS.Editor/Audio/IAudioPreviewer.cs
@@ -7,12 +7,11 @@ namespace AGS.Editor
 {
     public delegate void PlayFinishedHandler(AudioClip clip);
 
-    internal interface IAudioPreviewer
+    internal interface IAudioPreviewer: IDisposable
     {
         event PlayFinishedHandler PlayFinished;
 
-        bool RequiresPoll { get; }
-        bool Play(AudioClip clip);
+        void Play();
         bool IsPlaying();
         void Poll();
         int GetLengthMs();
--- a/Editor/AGS.Editor/Audio/IrrklangPlayer.cs
+++ b/Editor/AGS.Editor/Audio/IrrklangPlayer.cs
@@ -11,41 +11,53 @@ namespace AGS.Editor
     {
         public event PlayFinishedHandler PlayFinished;
 
-        private ISoundEngine _soundEngine = new ISoundEngine();
+        private ISoundEngine _soundEngine = null;
         private ISound _soundPlaying = null;
         private AudioClip _audioClip = null;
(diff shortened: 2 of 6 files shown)


rofl0r:
imo if more than one thing is being fixed, it should be one commit per fix...

morganwillcock:
It is all interconnected, except the labels being resized.

morganwillcock:
Problem solved.

rofl0r:
no worries, just my 2 shekels.
btw another minor nitpick: if one day the ags source is no longer hosted on github, the text you added to your PR description will be lost. so imo it would be better as part of the commit message.

ivan-mogilko:
> Fix MIDI pause and resume resetting all MIDI instruments to channel 0 (if you try this currently, every instrument will become a piano when resuming)

Ha, never noticed that before.


It seems that after these changes opening MIDI files in the editor preview became noticeably slower. Before it took a fraction of second, now I can count to 2 or 3 seconds before the pane opens. At first I thought it is strictly proportional to the file length, but some of the longer files open faster so maybe format type counts too.

morganwillcock:
MIDI playback does seem quite slow to initialize, and the previewer is still constructed as the GUI is drawn but now also incorporates loading the source. Previously nothing was setup until pressing play, so for irrKlang this was probably quite slow as it was actually setting up the audio engine at that point.

Unfortunately the two 'previewers' behind the interface don't actually share much in common, and irrKlang is not thread safe (unless I misunderstood how to use it), so I'm thinking the setup cannot be moved of the GUI thread?

ivan-mogilko:
> irrKlang is not thread safe (unless I misunderstood how to use it), so I'm thinking the setup cannot be moved off of the GUI thread?

I don't know anything about that myself atm.
Well, if you think that's not trivial we could merge the pr and open a new issue regarding preview optimisation for the future.

morganwillcock:
That would be fine with me. I don't think it is trivial to fix without some persistent audio system that is outside of the GUI, whereas the changes here are just based on fixing the original implementation.

PR #581, opened 2019-01-13
Use Magick.Net as a GIF loader

### LINKED ISSUE #543: Issues with GIF decoder
Relevant for #541 as well as general use, it seems the GIF decoder does not correctly handle importing 'optimised' GIF frames, where subsequent frames rely on information from the previous frame. ~~It also seems to create resource leaks when in use, so re-importing a sequence of images from a large GIF file can potentially crash the editor (you can generally get away with it when not importing in bulk, or when image dimensions are fairly small).~~

I did look at a few alternatives to the existing GIF decoder, the easiest solution would seem to be using the 8-bit version of Magick.NET, although a completely different image loading library for GIFs would seem to be overkill. Everything touched by GDI will become 32-bit but I tried out some alternatives from here:

[https://blogs.msdn.microsoft.com/dotnet/2017/01/19/net-core-image-processing/](https://blogs.msdn.microsoft.com/dotnet/2017/01/19/net-core-image-processing/)

...and I couldn't find options on SkiaSharp, ImageSharp, or Magick.NET (> 8-bit) to preserve an 8-bit image. I also looked at wrapping the WPF GIF decoder, but it has no built-in support for rebuilding an optimised frame, so it would have to manually place frame components in a similar fashion to the current decoder.

morganwillcock:
This removes the custom GIF decoder and uses the Nuget packaged Q8-x86 version of [Magick.NET](https://github.com/dlemstra/Magick.NET) to load GIF images. I've been looking at this for a few weeks and haven't found a smaller library that will correctly decode an optimised frame sequence without converting the result into a 32-bit image.

As an example of the difference between the two, here is the same source loaded into the AGS Editor, then exported as individual frames, then merged back into a single GIF file (just to illustrate the problem here and keep both in-sync):

test

The Magick.net version is on the left, the current decoder is on the right. As well as getting the colours confused, the current decoder also leaves a hole in the image where partial frames have been drawn.

This fixes #543, although I was hoping to find a solution that wasn't quite so large (in terms of file size). But since what is currently used doesn't work, and .NET image loading was enough issues that it wouldn't be out of the question to use a library to deal with all image loading, I think this is the best solution to get it working in the short term (perhaps in AGS4 everything will be 32 bit and this will become a non-issue anyway...). The other viable alternatives are probably ffmpeg (but this will be a lot larger) and the WPF GifDecoder (but there is no coalesce function to combine partial frames, it will return the literal frame ignoring the frame disposal methods) 

Note: I've not looked at the installer packaging, but I imagine if this change is accepted it'll need updating to include the Magick.Net files:

<pre>
>dir /b Magick*

Magick.NET-Q8-x86.dll
Magick.NET-Q8-x86.Native.dll
Magick.NET-Q8-x86.xml
</pre>


ivan-mogilko:
> Note: I've not looked at the installer packaging, but I imagine if this change is accepted it'll need updating to include the Magick.Net files:

>dir /b Magick*

Magick.NET-Q8-x86.dll
Magick.NET-Q8-x86.Native.dll
Magick.NET-Q8-x86.xml

I'd like to ask, what is the "xml" file for? I noticed AGS.Types.dll creates one too, but I could not figure out if its really necessary (editor could be run without this file too).

morganwillcock:
I think it is a map between the native binary and .NET, but not sure if it is required at runtime or not. If you can import a GIF without it, I think it would be okay to omit it.

PR #582, opened 2019-01-14
Ags4

fernewelten:
((Note the important "Caveats" below))
Touched up the parser code: Keeping the original logic as far as possible, tried to enhance comprehensibility by:
- putting in functions
- seriously reducing nesting depth
- adding in comments
- renaming some variables and functions to be as apt as I could make them
- converting the code to the AGS formatting/indentation guidelines that I found lying around somewhere

Wrote googletests that cover 75 % of the code lines and that prove that the old parser code generates exactly the same bytecode as the new parser code, byte for byte (cf. the file cc_byte_compat_test.cpp in the solution "Compiler.Lib.Test").

Converted the scanner and tokenizer to C++ classes in the ags namespace.

Converted the arrays that manage code nesting to a proper class in the ags namespace.

Converted some C strings to C++ strings (std::string).

Found an "old" symboltable structure lying around in script/cc_symboldef.h that had been nearly, but not completely superseded by the "new" symboltable structure in script/symboltable.h. As far as I could ascertain, the parser was the last one to use the "old" symboltable. I've rewritten these code places to use the "new" symboltable structure, so the new parser doesn't use it any more, as well.

Added some typedefs in the ags namespace that helped me check my code. Most notably, ags::Symbol_t for bytecode integers. If you don't like it, this is easy to change back (by mechanically changing all "ags::Symbol_t" to "int" and then all "Symbol_t" to int).

Changed some error messages to be more descriptive.

**Caveats:**

- I had to choose my hill to die on, and so I have NOT written any tests yet that check AGS programs with errors; all my tests check correct AGS programs. However, when I started the rewrite I found some pre-existing googletests that check AGS programs with errors. All the googletests that I found still check out ok. 
- My googletests only check bytecode and fixups. See cc_byte_compat_test.cpp in the solution "Compiler.Lib.Test"
- I don't know how to define properties in AGS code, and so I couldn't write googletests for them. This means that all the properties parsing code is untested (knowing my coding capabilities, I am CERTAIN that the properties parsing code WILL CONTAIN BUGS).
- I don't know how stringstructs need to be set up, and so I could NOT check the stringstruct code (again, this code WILL CONTAIN BUGS).
- I could only check old strings (type string). I don't know how to set up new strings (type String). I suspect that macros and startup code are packed to all AGS programs before they are given to the parser. But I haven't been able to determine the specifics, so I didn't know how to reproduce them in googletests.
- I've not been able to determine how macros are handled in AGS.  I don't know where those macros are defined or expanded. So I've only tested macro free code.
For instance, the following is legal AGS code:
	for (function i = 0; i < 10; i++) {
		break;
	}
This probably means that "function" is a macro for "int" in AGS.
- In the pre-existing code, I've found first efforts for varargs functions in AGS code. However, as far as I could determine, they don't compile with the pre-rewrite parser. So they don't compile with the rewritten parser, either. I've tried to port all the code I found to the rewritten version.


ivan-mogilko:
Hello again!

This pr is... overwhelming. I will try to find time to overview it.

But first of all there's a big organizational issue: you were making commits right in the "ags4" branch and had other peoples (e.g. mine) commits brought in later from our repository (not sure if using merge or cherry picking).
For pull request we suggest creating a separate feature branch and keep rebasing it on top of the main one (ags4 in this case) to keep up to date with the upstream repository. This way you will have your own copy of ags4 being practically a clone of ours and your work in a separate branch. It makes it much easier to understand what is going on and was has changed by the pull request.

I'd also like to make some name to this pr rather than "ags4". What about "Script compiler refactoring" (if I understand the intentions correctly?)

AlanDrake:
Oh, it was still here.
Sorry, you have targeted the wrong "ags4" branch, it was an old dead branch we forgot to remove. Try again with "master" which is the correct one.

ivan-mogilko:
@fernewelten , amending to what I said above, I also noticed there's some log file updated in several commits, which is probably not necessary for this change (and seem to contain your personal data, like disk paths). This file should not be a part of pull request.

fernewelten:
So let's try this again, this time in the master branch. :)

I've found the log file; it's from the add-in OpenCppCoverage that tracks the code lines that are executed in tests. I've added a .gitignore line for the file.

At the moment, my main focus is on refactoring the parser code in cs_parser.cpp, to make it easier to understand, to manage, to extend. I'm touching the other files only as much as necessary for that.

PR #608, opened 2019-02-05
Create .travis.yml for building AGS Engine on Linux

ericoporto:
This builds AGS on Linux following the instructions here:

https://github.com/adventuregamestudio/ags/tree/master/debian#building-a-debianubuntu-package-of-ags

Deployed results are upload to Github Releases when a new Release is created.

Requires an user that has TravisCI configured with read and write permissions to this repo and environment variable GITHUB_TOKEN configured for Releases to work.

Build Logs from the PR repo can be seen [on Travis here](https://travis-ci.com/mythsuntold/ags).

It's probably smart to install Allegro [from here](https://github.com/liballeg/allegro5/releases/tag/4.4.3) instead of getting from apt repository - but I didn't do that here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: c
+compiler:
+  - gcc
+addons:
+  apt:
+    packages:
+    - debhelper
+    - build-essential
+    - pkg-config 
+    - cmake 
+    - libfreetype6-dev 
+    - libogg-dev 
+    - libtheora-dev 
+    - libvorbis-dev
+    - liballegro4-dev
+    - libaldmb1-dev
+    - fakeroot
+script:
+- fakeroot debian/rules binary
+deploy:
+  provider: releases
+  skip_cleanup: true
+  api_key: $GITHUB_TOKEN
+  file: 
+  - ../ags_3~git-1_amd64.deb
+  - ../ags-dbg_3~git-1_amd64.deb
+  on:
+    tags: true


morganwillcock:
Should this also be building a tar archive? Otherwise someone will have to pull the package apart if they just want to run the engine without installing the .deb package.

ivan-mogilko:
I have a question, may Travis setup be also used for other kinds of jobs, even if only linux-related?

For instance:

1. Building full pack of native binaries for the editor with this script: https://github.com/adventuregamestudio/ags/blob/master/debian/make_ags%2Blibraries.sh

2. Creating source tarball for those who prefer to build from sources on their systems. I currently use following script:
<pre>#!/bin/bash

cd ags
VERSION=$1
echo Source revision: $VERSION
ORIG_CONTENT="$(git ls-tree --name-only $VERSION | grep -E '(Common|Engine|Plugins|Changes.txt|Copyright.txt|License.txt|OPTIONS.md|README.md)')"
echo Files: $ORIG_CONTENT
echo Running git archive...
git archive --prefix=ags_linux_$VERSION/ --format=tar $VERSION $ORIG_CONTENT | xz > ../ags_linux_$VERSION.tar.xz
echo Done
cd ..</pre>

ericoporto:
Yes, Travis also supports cron jobs for periodic things and a multitude of
different triggers. It only naturally supports MacOS and Linux (Ubuntu
Trusty on default). Travis also supports other distributions and has beta
support for Windows, but I have never used. I find it to have the best
documentation and reliability, but almost no personal support. AppVeyor has
the best support, but not best documentation, but not that bad. Azure has
almost no docs, and almost no support.

ericoporto:
Ahn, is there a command in one of those scripts to build the tar.gz @morganw?

morganwillcock:
It looks like after the Debian package has been built it is unpacked and the binary is copied out:
```cp $BASEPATH/ags+libraries/usr/bin/ags $BASEPATH/ags+libraries/data/ags32```
...and the libraries are also copied by the hook script `B00_copy_libs.sh`.

So actually, the files should already be there, you'd just have to get them treated as build artifacts and/or make an archive from them.

For making the sources tar archive, I guess you could do this anywhere but could also make another build artifact from the git-archive script listed above. I think you can probably do this in a single line though, I'll have a try later on.

morganwillcock:
This should do the same thing in a single command, so you can probably just define it in the script block of the yaml file:

```git -C ags archive --prefix=ags_linux_$VERSION/ --format=tar $VERSION Common Engine Plugins Changes.txt Copyright.txt License.txt OPTIONS.md README.md | xz > ags_linux_$VERSION.tar.xz```

Assuming the repository is cloned to the subfolder 'ags' (this is running from one level above the repository to avoid switching directories) and $VERSION is a release tag. Potentially you might need to define it as using a shell to be able to use a pipe, but I imagine everything in the script block will end up in a shell anyway.

ericoporto:
Just a reminder that someone with access to this repo has to actually activate Travis-CI for this to work, and someone with commit access has to add their token in the Travis-CI environment for the autodeploy on tag to work. :/

ivan-mogilko:
@ericoporto I am sorry, I keep forgetting how this works. Could you open an issue and explain what has to be done for travis activation? This way it will be harder to forget at least.
Do we need to have some dummy user again?

ericoporto:
@ivan-mogilko here is the issue. I think if tzachs and some others are ok this can be moved to a wiki entry, need to figure out if I said any BS there because I wrote from memory on my phone :/

tzachshabtay:
Fine by me, but I'm not running this show :)
(Btw, checked and I don't have permissions to activate travis here).

PR #619, opened 2019-02-11
Don't select layer objects when the room tab isn't the active tab

### LINKED ISSUE #617: Properties panel shows wrong details when setting StartingRoom
Someone sent me a screenshot of the properties panel for a Character showing the properties for an Object.

properties

It seems easy to recreate this problem, With the room editor open and an object defined, just set the starting room for a character. Internal state seems to get broken at this point, so re-opening the Character properties continues to show the wrong information. To recover you have to restart the Editor.

wrong_properties

morganwillcock:
Fixes #617

### DIFF
--- a/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
@@ -679,7 +679,13 @@ namespace AGS.Editor
 
             layerNode.IsVisible = true;
             SelectLayer(layerNode.Layer);
-            layerNode.Layer.SelectItem(node == layerNode ? null : node.RoomItemID);
+
+            // only select the item if the room editor is the active tab; this could
+            // be a refresh from another tab e.g. changing a character's starting room
+            if (Factory.GUIController.ActivePane.Control == this)
+            {
+                layerNode.Layer.SelectItem(node == layerNode ? null : node.RoomItemID);
+            }
         }
 
         private void SelectLayer(IRoomEditorFilter layer)


ivan-mogilko:
I can confirm that this fixes bug, but I'd like to investigate the code more, because it does not seem right to have condition right there.

On one hand, if we need to skip selecting new object on properties pane, then the condition maybe should be checked later, in public void SelectItem where it calls Factory.GUIController.SetPropertyGridObject.
On the other hand, it may be interesting to know why does it call this "editAddressBar_SelectionChange" at all (it does a lot of stuff there, incliding selecting layer and turning item visible).

ivan-mogilko:
I've been investigating the code a little, and seems the two main issues there are:

1) Navigation bar fully recreates all nodes each time an item is added or removed, therefore it has to reselect old item in the end. There's my old comment that notes this which I forgot about.
Ideally it should be subscribing to add/remove/edit item events and change only necessary node, but that's completely different problem ofcourse.

2) The properties pane is a subordinate of actual editors, which may be a design mistake IMHO. This is arguable ofc, but maybe it would be more convenient to make it opposite way, requiring editors to tell their "current selected object" on demand, and property pane subscribing to selection change events. 

Anyway, this fact that it is editors that instead order to display properties at particular moments is bigger cause of trouble, because there's no attempt made to check if the editor is active (and not doing a background task of sorts).

Since there are multiple situations when it's doing that, I'd like to propose to fix this bug differently and instead make a kind of helper function in the editor pane to substitute direct calls to Factory.GUIController.SetPropertyGridObject and tests whether editor pane is active (like in your fix).
Another reason is that layerNode.Layer.SelectItem changes internal state of a pane too which may be allowed in this case.

morganwillcock:
Are checks, to see if the pane is active, always correct?
#620 would suggest that some checks are incorrect outside of a tabbed panel.

ivan-mogilko:
> Are checks, to see if the pane is active, always correct?
#620 would suggest that some checks are incorrect outside of a tabbed panel.

I tried to test it a little just now, and I cannot reproduce this part: _"* In a similar way to #617 it can cause panel properties for the room objects to be shown in place of properties for other Editor components"_
What has to be done to see this issue?

morganwillcock:
I've added a screen capture of an example, but you can break it in a similar way.

1. dock panel on extreme left
2. open character tab and any other tab (just to have something to switch to and from, to demonstrate that the character editor gets broken)
2. make room object
3. delete room object

ivan-mogilko:
> I've added a screen capture of an example, but you can break it in a similar way.

Ok I can reproduce this now (not sure why it did not work before).
The reason this happens is because when object gets deleted and editor tries to set Room as the properties object "_mainForm.ActivePane" is CharacterEditor and not RoomEditor for some reason.
(My blind guess so far is that there's something wrong with pane focus, as with the _selectedObject crash).

In other words, it seems that testing whether ActivePane is our pane before Factory.GUIController.SetPropertyGridObject might fix this problem also (not that it's a fix for ActivePane gets changed to other tab in the middle of the action; this may need separate investigation).

ivan-mogilko:
> In other words, it seems that testing whether ActivePane is our pane before Factory.GUIController.SetPropertyGridObject might fix this problem also

No, sorry, I take my words back.
I did not realize this earlier, but SetPropertyGridObject does two separate actions:  first sets SelectedPropertyGridObject for the ActivePane. and then calls RefreshPropertyGridForDocument, passing ActivePane as an argument.
I guess there could be reasons to do it that way, but in some cases, like when the pane itself changes its selected object, it needs to set its own SelectedPropertyGridObject explicitly regardless whether it is ActivePane or not.

ivan-mogilko:
@morganwillcock , I tried to follow my own suggestions but found that there's much more to that.
Everything related to properties (reacting to target object change, and property value change) rotates around ActivePane and using global singletons in general, making assumptions about their values.

Like, see CharacterComponent.PropertyChanged, it starts as
<pre>
Character itemBeingEdited = ((CharacterEditor)_guiController.ActivePane.Control).ItemToEdit;
</pre>
similar thing is done in multiple other places.

As a hypothetical case, if there's was an async "property changed" event and ActivePane changes between sending event and receiving it, then the logic will break. As now happening with the target object change when the document focus switches for some reason (#620).

I don't know if this can be solved by a quick fix. I'll try some quick hacks for a test, but all this should be redesigned IMO to prevent future issues.

(For instance, why not send a reference to the edited object to PropertyChanged? what could be the reason not to)

ivan-mogilko:
I merged this as a quick fix, but Editor seriously needs a redesign of connection between its components.

morganwillcock:
Sorry for lack of response. I'd only had a chance to look very briefly, but had come to the same conclusion that the proper fix is to rewrite the whole thing, which is probably out-of-scope for the release milestone.

PR #623, opened 2019-02-12
AGS3 - Some warning and compatibility fixes

sonneveld:
This introduces a few fixes and tidy ups from my SDL2 branch.


Engine/ac/dynobj/scriptstring.cpp
```diff
@@ -35,8 +36,10 @@ const char *ScriptString::GetType() {
 }
 
 int ScriptString::Serialize(const char *address, char *buffer, int bufsize) {
-    if (text == NULL)
-        text = "";
+    if (text == nullptr) {
+        text = static_cast<char *>(malloc(1));
```
rofl0r: wouldn't this cause a memory leak ?

sonneveld: Nope, we free it in Dispose. The problem was that we were trying to free a string literal.

ivan-mogilko: Hmm, I guess ScriptString with null pointer inside was never a case...
But frankly, I think this should be fixed differently, changing function logic instead to something like:
<pre>
    size_t len = text == nullptr ? 0 : strlen(text);
    SerializeInt(len);
    if (text)
        strcpy(&serbuffer[bytesSoFar], text);
    bytesSoFar += len + 1;
</pre>

sonneveld: I did something like that. Main difference was that you need to still strcpy or add a null char even if text len is 0.  (to match up with the unserialize implementation)

.gitignore
```diff
@@ -139,3 +139,8 @@ libsrc/*.zip
 # NuGet v3's project.json files produces more ignorable files
 *.nuget.props
 *.nuget.targets
+
+\.DS_Store
```
ivan-mogilko: Could you please add comments to these new entries? Also, there's already "XCode" section above, if DS_Store is relevant please move this item to same group to keep things organized.

sonneveld: Fair point. Added comments. ds_store is an os level file like thumbs.db and desktop.ini, so doesn't quite fall under xcode.

PR #634, opened 2019-02-16
AGS 3: fix reading GUIs in pre-3.4.0 games

ivan-mogilko:
This fixes regression introduced in a2247b8db74b3464718a8a7773d3f76028e0c856 with the removal of GUIMain::ControlCount member variable.


sonneveld:
I tested those changes against my branch and it seems to load correctly now.

PR #641, opened 2019-02-20
AGS3: Remove obsolete (possibly broken) input recording/playback

sonneveld:
This is a refactoring of some of the keyboard/mouse event code to make it easier to rewrite for SDL2.

This pull request is done in a few stages:
1) remove recording/playback
2) rename rec_* functions to ags_*
3) rename record.{cpp,h} to sys_events (since 'events.cpp' was already taken :) )

I deliberately left the functions in sys_events as I will be using that module as a base to slot in the SDL event handling.

I've only tested by building in Linux, I don't believe I've broken anything for Windows. OSX project files will need updating but I suspect they're already out of date from earlier commits :)

### DIFF
--- a/Engine/ac/global_debug.cpp
+++ b/Engine/ac/global_debug.cpp
@@ -119,7 +119,9 @@ void script_debug(int cmdd,int dataa) {
         delete tempw;
         delete view_bmp;
         gfxDriver->DestroyDDB(ddb);
-        wait_until_keypress();
+        // Wait until key press
+        while (!ags_kbhit());
+        ags_getch();
         invalidate_screen();
     }
     else if (cmdd==3) 
@@ -177,7 +179,9 @@ void script_debug(int cmdd,int dataa) {
         delete tempw;
         delete view_bmp;
         gfxDriver->DestroyDDB(ddb);
-        wait_until_keypress();
+        // Wait until key press
+        while (!ags_kbhit());
+        ags_getch();
     }
     else if (cmdd == 99)
         ccSetOption(SCOPT_DEBUGRUN, dataa);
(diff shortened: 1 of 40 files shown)


rofl0r:
imo the commit removing recording could use some more explanations in the commit message about why it's considered obsolete. that's helpful for anybody looking at the commit log or history.

sonneveld:
It's an interesting point, I'll keep in mind for next pull request.

rofl0r:
there's also `git rebase -i HEAD~3` which allows to `reword` commit messages, followed by `git push -f` :)

sonneveld:
Yeah yeah :) I know. I don't want to confuse the pull request.

ivan-mogilko:
> I don't want to confuse the pull request.

Github fixes pull request automatically when the branch is force pushed, I've did that few times already.

sonneveld:
Please just let me be lazy :)

Engine/ac/global_debug.cpp
```diff
@@ -119,7 +119,9 @@ void script_debug(int cmdd,int dataa) {
         delete tempw;
         delete view_bmp;
         gfxDriver->DestroyDDB(ddb);
-        wait_until_keypress();
+        // Wait until key press
+        while (!ags_kbhit());
+        ags_getch();
```
ivan-mogilko: I have only one question, why did you remove "wait_until_keypress"? I used that as a convenience helper function (to avoid putting "while" construct in place every time.

sonneveld: It actually didn't work well with some graphics modes in SDL because we were no longer rendering to screen.  In my SDL version I refactored code that used wait_for_keypress to have a render & check for key event loop.

PR #642, opened 2019-02-23
Plugin save/restore uses file handle.

sonneveld:
Instead of casting a FILE* pointer to an integer, the plugin module will now use an abstract file handle which maps to the currently opened savegame file. 

Plugins are supposed to use the engine->FWrite/FRead functions so this also refactors the inbuilt plugins to use that as well.

Since there's only ever one savegame file open at a time, this reuses the save/restore event code as the file handle.

### DIFF
--- a/Common/util/filestream.h
+++ b/Common/util/filestream.h
@@ -38,14 +38,6 @@ public:
     virtual void    Close();
     virtual bool    Flush();
 
-    // TODO
-    // Temporary solution for cases when the code can't live without
-    // having direct access to FILE pointer
-    inline FILE     *GetHandle() const
-    {
-        return _file;
-    }
-
     // Is stream valid (underlying data initialized properly)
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Right, so this is long due, and I feel like an idiot now because did not guess that you can do this earlier :).

But! You did this even more complicated than necessary. I don't think you even need to cast stream pointer to FileStream, because with these changes plugins won't care about actual stream implementation. So you may just use Stream* everywhere.

sonneveld:
> But I think that you still overcomplicate things casting stream pointer to FileStream, because with these changes plugins won't care about actual stream implementation. So you may just use Stream* everywhere.

Ah I think you're right. I did briefly consider leaving the type as Stream but I thought maybe you had other reasons (other than the FILE* pointer).

PR #648, opened 2019-02-27
AGS3: Modify AGS::Common::String to use std::string instead of custom code

sonneveld:
Controversial pull request!  I tried to arrange this in two parts, the first has some general string based bug fixes, the second has the modification to String to use std::string.

The main benefit is that the c++11 string type does not have COW, which makes String much safer for multithreaded code. 

I also removed the implicit conversion from String to char*. I noticed that the compiler would silently convert to char* if a comparison operator wasn't implemented. That means there was the potential for pointer comparisons instead of content. It also makes it much more obvious which parts of the code are relying on char* strings.

Note: I have only tested on Linux so far, will test in Windows and fix any issues I find.

edit: I've made some fixes to suit Windows. I'm not sure I've fixed everything.

edit2: implicit conversion discussion https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ro-conversion
and https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c46-by-default-declare-single-argument-constructors-explicit

### DIFF
--- a/Common/util/path.cpp
+++ b/Common/util/path.cpp
@@ -24,7 +24,7 @@ bool IsDirectory(const String &filename)
     struct stat_t st;
     // stat() does not like trailing slashes, remove them
     String fixed_path = MakePathNoSlash(filename);
-    if (stat_fn(fixed_path, &st) == 0)
+    if (stat_fn(fixed_path.GetCStr(), &st) == 0)
     {
         return (st.st_mode & S_IFMT) == S_IFDIR;
     }
@@ -34,7 +34,7 @@ bool IsDirectory(const String &filename)
 bool IsFile(const String &filename)
 {
     struct stat_t st;
-    if (stat_fn(filename, &st) == 0)
+    if (stat_fn(filename.GetCStr(), &st) == 0)
     {
         return (st.st_mode & S_IFMT) == S_IFREG;
     }
@@ -77,8 +77,8 @@ bool IsSameOrSubDir(const String &parent, const String &path)
     char can_path[MAX_PATH];
     char relative[MAX_PATH];
     // canonicalize_filename treats "." as "./." (file in working dir)
-    const char *use_parent = parent == "." ? "./" : parent;
-    const char *use_path   = path   == "." ? "./" : path;
+    const char *use_parent = parent == "." ? "./" : parent.GetCStr();
+    const char *use_path   = path   == "." ? "./" : path.GetCStr();
     canonicalize_filename(can_parent, use_parent, MAX_PATH);
     canonicalize_filename(can_path, use_path, MAX_PATH);
     const char *pstr = make_relative_filename(relative, can_parent, can_path, MAX_PATH);
@@ -131,7 +131,7 @@ String MakeAbsolutePath(const String &path)
     return path;
 #endif
     char buf[512];
-    canonicalize_filename(buf, abs_path, 512);
+    canonicalize_filename(buf, abs_path.GetCStr(), 512);
     abs_path = buf;
     FixupPath(abs_path);
     return abs_path;
--- a/Common/ac/common.cpp
+++ b/Common/ac/common.cpp
@@ -26,5 +26,5 @@ void quitprintf(const char *fmt, ...)
     va_start(ap, fmt);
     String text = String::FromFormatV(fmt, ap);
     va_end(ap);
-    quit(text);
+    quit(text.GetCStr());
 }
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -413,7 +413,7 @@ void GameSetupStruct::read_room_names(Stream *in, GameDataVersion data_ver)
             roomNumbers[bb] = in->ReadInt32();
             pexbuf.Read(in, STD_BUFFER_SIZE);
             roomNames[bb] = new char[pexbuf.GetLength() + 1];
-            strcpy(roomNames[bb], pexbuf);
+            strcpy(roomNames[bb], pexbuf.GetCStr());
         }
     }
     else
--- a/Common/core/asset.cpp
+++ b/Common/core/asset.cpp
@@ -27,7 +27,7 @@ AssetInfo::AssetInfo()
 }
 void AssetLibInfo::Unload()
 {
-    BaseFileName = "";
+    BaseFileName.Empty();
     LibFileNames.clear();
     AssetInfos.clear();
 }
--- a/Common/core/assetmanager.cpp
+++ b/Common/core/assetmanager.cpp
@@ -68,7 +68,7 @@ AssetManager::~AssetManager()
 
 /* static */ bool AssetManager::IsDataFile(const String &data_file)
 {
-    Stream *in = ci_fopen(data_file, Common::kFile_Open, Common::kFile_Read);
+    Stream *in = ci_fopen(data_file.GetCStr(), Common::kFile_Open, Common::kFile_Read);
     if (in)
     {
         MFLUtil::MFLError err = MFLUtil::TestIsMFL(in, true);
@@ -80,7 +80,7 @@ AssetManager::~AssetManager()
 
 AssetError AssetManager::ReadDataFileTOC(const String &data_file, AssetLibInfo &lib)
 {
-    Stream *in = ci_fopen(data_file, Common::kFile_Open, Common::kFile_Read);
+    Stream *in = ci_fopen(data_file.GetCStr(), Common::kFile_Open, Common::kFile_Read);
     if (in)
     {
         MFLUtil::MFLError err = MFLUtil::ReadHeader(lib, in);
@@ -288,7 +288,7 @@ AssetError AssetManager::RegisterAssetLib(const String &data_file, const String 
     _basePath = ".";
 
     // open data library
-    Stream *in = ci_fopen(data_file, Common::kFile_Open, Common::kFile_Read);
+    Stream *in = ci_fopen(data_file.GetCStr(), Common::kFile_Open, Common::kFile_Read);
     if (!in)
         return kAssetErrNoLibFile; // can't be opened, return error code
 
@@ -353,7 +353,7 @@ bool AssetManager::GetAssetFromLib(const String &asset_name, AssetLocation &loc,
     if (!asset)
         return false; // asset not found
 
-    String libfile = cbuf_to_string_and_free( ci_find_file(NULL, MakeLibraryFileNameForAsset(asset)) );
+    String libfile = cbuf_to_string_and_free( ci_find_file(NULL, MakeLibraryFileNameForAsset(asset).GetCStr()) );
     if (libfile.IsEmpty())
         return false;
     loc.FileName = libfile;
@@ -364,7 +364,7 @@ bool AssetManager::GetAssetFromLib(const String &asset_name, AssetLocation &loc,
 
 bool AssetManager::GetAssetFromDir(const String &file_name, AssetLocation &loc, FileOpenMode open_mode, FileWorkMode work_mode)
 {
-    String exfile = cbuf_to_string_and_free( ci_find_file(NULL, file_name) );
+    String exfile = cbuf_to_string_and_free( ci_find_file(NULL, file_name.GetCStr()) );
     if (exfile.IsEmpty() || !Path::IsFile(exfile))
         return false;
     loc.FileName = exfile;
--- a/Common/core/assetmanager.cpp
+++ b/Common/core/assetmanager.cpp
@@ -307,8 +307,8 @@ AssetError AssetManager::RegisterAssetLib(const String &data_file, const String 
     String nammwas = data_file;
     String data_file_fixed = data_file;
     // TODO: this algorythm should be in path/string utils
-    data_file_fixed.TruncateToRightSection('\\');
-    data_file_fixed.TruncateToRightSection('/');
+    auto data_file_fixed_sections = data_file_fixed.Split("\\/");
+    data_file_fixed = data_file_fixed_sections[data_file_fixed_sections.size() - 1];
     if (data_file_fixed.Compare(nammwas) != 0)
     {
         // store complete path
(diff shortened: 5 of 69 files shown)


ivan-mogilko:
Perhaps you could include removal of conversion operator to the first commit ("Explicitly convert AGS Strings to char*")? It would help finding places where you need to add .GetStr().

Common/util/path.cpp
```diff
@@ -77,8 +77,8 @@ bool IsSameOrSubDir(const String &parent, const String &path)
     char can_path[MAX_PATH];
     char relative[MAX_PATH];
     // canonicalize_filename treats "." as "./." (file in working dir)
-    const char *use_parent = parent == "." ? "./" : parent;
-    const char *use_path   = path   == "." ? "./" : path;
+    const char *use_parent = (parent == String(".")) ? "./" : parent.GetCStr();
+    const char *use_path   = (path   == String(".")) ? "./" : path.GetCStr();
     canonicalize_filename(can_parent, use_parent, MAX_PATH);
```
ivan-mogilko: Since you remove implicit conversion to const char* perhaps this is no longer necessary?
Both AGS::Common::String and std::string implement comparison operator that accepts const char* as an argument, this is to avoid extra string object allocation.

ivan-mogilko:
TBH I'd rather pick out commits f000788 and 370cd37 , because they are bug fixes and not related to the changes to string implementation (also in case we have to apply them to older version).

sonneveld:
I modified the commit "Explicitly convert AGS Strings to char*" to also remove that conversion from the string.h file. There are a couple more places where we can call GetCStr but that's because I haven't modified the rest of string.h yet to accept String instead of char*

ivan-mogilko:
> I haven't modified the rest of string.h yet to accept String instead of char*

Do you mean you want to not accept arguments of const char* at all?

sonneveld:
I think it makes sense. That way we're only ever implicitly converting to AGS::String. You need to be explicit when converting back to a char*.

sonneveld:
I have pulled out those two fixes into their own pull requests.

ivan-mogilko:
> I have pulled out those two fixes into their own pull requests.
Ah, I was going to just cherry pick and push them right away, because they are small and clear :)

ivan-mogilko:
> I think it makes sense. That way we're only ever implicitly converting to AGS::String. You need to be explicit when converting back to a char*.

Converting back to char* is one thing, but having taking only String arguments? I don't know... that means extra string objects are created for example every time we assign a literal. Like I mentioned, std::string accepts C-strings also as argument to some functions and operators.
Is there really a problem with that?

sonneveld:
Only accepting AGS::String means we can keep the implicit conversion from char* to String. We only have to worry about being explicit in the other direction.

We could overload and implement for both types, but I wonder if we only need to do that in performance hot spots. And in those spots, should we be doing string manipulation at all?

Common/util/path.cpp
sonneveld: Updated this fix.

ivan-mogilko:
I made a decision to pass on merging this to 3.5.0. Perhaps this may be done for later versions.

This does not seem to give any immediate benefit and it completely changes an important utility class, and way some of its operations are performed (like split to sections).
The worst imho is that the engine code currently assumes that String objects are copied very fast (as 1 pointer) and so many functions return and assign them without care. I know that overall perfomance may not be affected much, but still.

Maybe we could copy some changes right now, like making conversion to const char* safer, which you mentioned in the comments above.

sonneveld:
All good, I'll just include it as part of my SDL2 port for now.  Were you able to measure a performance difference though?

The COW implementation copy still needs to perform some checks and balances before assigning that pointer. The c++11 strings can implement something called SSO, which can store small strings (10-22 chars) within the struct without having to allocate on the heap. This means to copy these small strings you only need to copy 3 longints.

rofl0r:
> I'll just include it as part of my SDL2 port for now

how is this change related to the SDL2 port ?

sonneveld:
> how is this change related to the SDL2 port ?

It's just where I'm doing my main work.. I occasionally pull stuff out (like this PR) to put into ags3.

sonneveld:
Another thought to do with performance here. C++11 has move semantics, which we can implement ourselves or rely on the defaults (which works because std::string has move semantics). It ultimately means we wouldn't be copying on return. We could even delete the copy constructor and check all places we're not passing by reference.

ivan-mogilko:
Only for record sake (and maybe for the sake of little rant too); I've been adjusting String class for one 3.5.0 pr a bit, and took a chance to review some of the existing code there, and it's struck me how suboptimal some of the code I wrote there is. Probably I was loaded with work that decided to give up on optimisations at some point and just wrote things the way which was simpliest for me.

Same thoughts bother me in regards to this new string reimplementation. One good example where const char arrays excel is constant string examination. Consider the "split" operation for instance, if we do not need to store results of split but only need to append its result somewhere, with current and yours implementation there will always be at least 1 extra string allocated in the process of copying:
* split out a part of string A, creating a new string B;
* append string B to previously existing string C;
* use/return string C;

This could be optimised if there were a class that would implement a "substring window" stored on stack, basically a pointer to the substring start and its length (it cannot obviously rely on null terminator).
Then the split operation would rather return not the new string, but description of a part of full string.

Of course this may be important or not important depending on which string operations we have in a program and how often they are used. What still annoys me that this proposed change does not aim any thought through string design, but rather to create a dummy std::string wrap.

sonneveld:
> Of course this may be important or not important depending on which string operations we have in a program and how often they are used. What still annoys me that this proposed change does not aim any thought through string design, but rather to create a dummy std::string wrap.

That is exactly what it is. Rely on the behavior of the std::string. with default copy and move semantics. I ran this code with a profiler to see which functions were used a lot. The Split function was not one of them. :)

ivan-mogilko:
There's no point in arguing here is not it? You really want to make a wrap for std::string. Ok.

ivan-mogilko:
My point being: why create a **wrapper** around std::string?

There's no immediate problem with String itself that would break anything or prevent from anything, so no urgent need to have std::string inside.

If you like to have std::string in the end, which is a reasonable wish, _I guess_, then why not work towards replacing String with std::string? I.e. `typedef std::string String;` if you want to keep old name.

This would ofcourse require to move various utility functions out, but this is something that would be beneficial anyway, I guess. I put a lot there, much of them could be replaced with more general algorithms accompanied by helper functions.

Again, my point is, AGS::String is a consistent class, std::string is a consistent class, but AGS::String wrapping std::string is a dumb solution made for some random reason.

sonneveld:
> My point being: why create a **wrapper** around std::string? Why add an ugly code like that? Why such need to replace things with things?

oh I didn't want to fight about this and I think we're in agreement. The wrapper was a step towards using just a plain std::string. I just wanted to start with a wrapper until we iron out any issues, then move to a typedef or rename everything to use std::string directly. I thought your main concern was performance with copying and I think the wrapper around std::string showed it wasn't an issue.

sonneveld:
I'd like to try and get something based on std::string into mainline. It would supply move semantics and doesn't have copy on write so it makes it safer to pass to other threads. I agree some of the implementation isn't optimal but as far as I can see those functions aren't heavily used.

Assuming the wrapper around std::string is a no-go, would you be happier if I implemented `AGS::String` as a typedef to `std::string`, then refactor the current string member functions as individual functions? I imagine that would touch more code though.

An alternative might be to wrap the COW code with mutexes but that might have a negative performance effect.

ivan-mogilko:
> Assuming the wrapper around std::string is a no-go, would you be happier if I implemented AGS::String as a typedef to std::string, then refactor the current string member functions as individual functions? I imagine that would touch more code though.

If compared to wrapper, this seems a better choice to me. (Although the refactor order maybe should be opposite?)

sonneveld:
> (Although the refactor order maybe should be opposite?)

So refactor into individual functions first, using the original string? It's doable but it would mean more refactoring for something that would replaced with the std::string implementation.

ivan-mogilko:
Well, see as it works best.

sonneveld:
Refactoring to use std::string member functions and individual helper functions is going to change a lot of code and require some effort to refactor. I'd like some assurance that the effort isn't going to go to waste.

rofl0r:
could we finally stop making huge factorings just for the sake of some imagined idiomatic conformity improvements? each of those make porting patches back and forth a mess as you'll get merge conflicts all over the place.

btw, funny fact: today i compiled the AGS version from 2013 and it took like 10 seconds whereas current ags3 branch takes several minutes. is that really progress ?
i didn't measure, but i'm pretty sure the old C-style code is a lot faster too.

sonneveld:
I'm trying to get a non-COW string implementation in so I have less issues with multithreaded code. There's already a decent standard library string now so I'd like to move to that. If we move to the wrapper then yes, there will be less merge issues.

The original code has one giant `ac.cpp` file. It might have been fast for the initial compilation but there would be been very little incremental compilation benefits.

rofl0r:
> but again, you're not typically doing clean builds during development.

change a header, and you effectively **do** a "clean build"

> I'm trying to get a non-COW string implementation

then why not simply adapt the existing class ?

what i'm seeing here is that for every tiny improvement you desire, you throw all related existing code over board and march in with some new std::crap with 500+ changes lines in 30+ files.

btw while looking at the super-inefficient VM code, i realized that a lot of the checks could be removed if all pointers to script memory were allocated on a heap under our control.
so one can simlpy check whether the pointer is within the boundaries of that heap.

but if you use std:: stuff, precisely that isn't possible because you lose the control over allocation.

ivan-mogilko:
It is true that old version was compiled almost instantly, but that was at the cost of having all code in 1 file with 30+k lines which were very hard to work with especially if you see it for the first time, so naturally I split it all in smaller files.

ivan-mogilko:
> btw while looking at the super-inefficient VM code, i realized that a lot of the checks could be removed if all pointers to script memory were allocated on a heap under our control.
so one can simlpy check whether the pointer is within the boundaries of that heap.

Which checks do you refer to?

I think I mentioned this before, but the biggest issue we had with VM was that compiled scripts were addressing engine variables using real memory offsets which would prevent us from modifying structs layout and even struct size (latter breaks using arrays of game objects). Since we naturally would need to modify these structs sooner or later, we followed the idea similar to one implement in scummVM port: intercept these memory access operations and remap the address or even call some function (if additional update is necessary after variable change).

I actually admit this whole approach could have been a mistake, and instead we could have separate old game logic with fixed struct layouts which should not be touched ever. Maybe there are other options too. But I think this issue (the need to intercept memory access and check for accessed object type) is one that affected script VMs work most.

rofl0r:
> It is true that old version was compiled almost instantly, but that was at the cost of having all code in 1 file with 30+k lines which were very hard to work with especially if you see it for the first time, so naturally I split it all in smaller files.

that should have made it faster, not slower. because now make -j8 can utilize all your 8 cores, building 8 smaller files in parallel.
what makes C++ compilation so shitslow is templates. the more templates and std:: stuff you use, the slower  your compilation becomes.

sonneveld:
> change a header, and you effectively **do** a "clean build"

That's.. not what incremental builds do.

> then why not simply adapt the existing class ?

That's what I tried to do with the wrapper class. Use the standard library class without affecting too much other code.

> but if you use std:: stuff, precisely that isn't possible because you lose the control over allocation

Not true. Most of the standard libraries let you specify your own allocator if needed.

rofl0r:
> Which checks do you refer to?

the particular checks i looked at was in SCMD_LOADSPOFFS code, which was originally meant *as an optimization* for faster local variable access. what should simply put reg[SP]-X into reg[mar], actually *loops* over every item on the stack and checks lots of stuff. (i looked at it because i was experimenting with how to generate an efficient jump table, as emitted by C switch statements), because the compiled code i'm looking at basically does `if(x() == 1) foo(); if(x() == 2) bar(); if(x() == 3) baz();` which is very inefficient.

rofl0r:
> That's what I tried to do with the wrapper class. Use the standard library class without affecting too much other code.

then why does this PR change 69 files instead of one ?

sonneveld:
> then why does this PR change 69 files instead of one ?

Please read the description. We were also implicitly converting String to char* which I believe would lead to incorrect comparisons (pointers instead of values)

sonneveld:
I'm closing this. Conversation has gone off topic and too toxic for me.

ivan-mogilko:
> the particular checks i looked at was in SCMD_LOADSPOFFS code, which was originally meant as an optimization for faster local variable access. what should simply put reg[SP]-X into reg[mar], actually loops over every item on the stack and checks lots of stuff.

Yes, that's awful, and worst tbh I don't remember why this was done so exactly. I think maybe I was trying to detect when it accesses a member of a struct, because current VM implementation needs to know the type of object.

ivan-mogilko:
@sonneveld I was giving this another thought some day ago. mgambrell once told me (we had a quick discussion about same ags string class) that it may be beneficial to have a custom string class for the specific engine needs.

So, I was thinking about this hypothetically once more, and asked myself a question: what is expected from string resources in ags, based on how strings are most commonly used? and came to following points:
* immutable (string resources are never changed, and those strings which change are temporary local variables);
* persistent (many string resources are stored for whole game life duration);
* shareable (because string resource may be assigned to the number of objects);
* has precalculated hash (because there are dictionaries string resources may be used in as keys, and also mixes well with them being immutable).

Since these strings are persistent and immutable they could be even allocated in a bulk on a large buffer (if that matters), which (theorising) may be supported by setting reference count to +1 more than necessary to prevent pointed memory from deletion.

Hypothetically furthermore engine could use these custom strings to store and assign string resources, while std::string is used for mundane purposes.
The biggest question of course here is how to make convenient internal API since some functions would need to accept only custom strings as arguments.

The "shareable" concept could be implemented in various ways, such as internal/external reference counting, and maybe something else. The fact that string is immutable, and often guaranteed to be persistent may simplify things (and maybe simplify protecting them in multithreaded enviroment, hopefully).

Anyway, these are my recent thoughts on the matter, they are not fully formed, but I thought to just leave them here for consideration.

PR #651, opened 2019-03-01
AGS 3: deprecate sprite resolution tag

ivan-mogilko:
A brief explanation, in AGS the game may be thought as being "low resolution" and "high resolution".
This exists since the times when it did not support any arbitrary game size, and historically 320x240 was the last "low resolution" and 640x400 was the first "high resolution".

Some assets, like sprites in this case, were letting developer to tag them as either "lowres" or "highres", and if this tag did not match the game's then such sprite would be scaled up or down twice in game respectively. I believe these were mostly meant to reuse same sprites in different games, but now with arbitrary game sizes the border between "lowres" and "highres" games became very vague.

This PR deprecates "Low/high resolution" sprite flags and introduces "Real resolution" type to replace them as the new default. This means that sprite is always visually same regardless of the game's native resolution.

Upon loading old games or importing older exported assets AGS "promotes" sprite resolution: if its tag matches the game type then it is changed to "Real". Usually that should promote all or nearly all sprites in game.

This should be an intermediate step between old ags and ags4 where this setting is completely removed (I think).

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -97,6 +97,9 @@ OPT_RENDERATSCREENRES, extended engine caps check, font vertical offset.
 49 : 3.4.1.2
 Font custom line spacing.
 
+50 : 3.5.0.8
+Sprites have "real" resolution.
+
 */
 
 enum GameDataVersion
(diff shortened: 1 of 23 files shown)


Common/ac/spritecache.cpp
```diff
@@ -53,6 +53,28 @@ SpriteInfo::SpriteInfo()
 {
 }
 
+int SpriteInfo::GetAdjustedWidth(bool hires_ctx) const
+{
+    if (!IsVarRes())
+        return Width;
+    if (IsHiRes() && !hires_ctx)
```
sonneveld: Probably don't need the elses if you're just returning

ivan-mogilko: I realized that these functions are wrong for some reason, and I am not using them anyway, so going to remove them.

ivan-mogilko:
Had to fix couple of silly bugs.

PR #652, opened 2019-03-01
AGS3 refactor ManagedObjectPool to address slow lookups

sonneveld:
Uses maps for object handle and address lookups.

### DIFF
--- a/Engine/ac/dynobj/managedobjectpool.cpp
+++ b/Engine/ac/dynobj/managedobjectpool.cpp
@@ -29,22 +29,26 @@ const auto SERIALIZE_BUFFER_SIZE = 10240;
 const auto GARBAGE_COLLECTION_INTERVAL = 1024;
 const auto RESERVED_SIZE = 2048;
 
-int ManagedObjectPool::Remove(int32_t handle, bool force) {
-    auto & o = objects[handle];
+int ManagedObjectPool::Remove(ManagedObject &o, bool force) {
+    if (!o.isUsed()) { return 1; } // already removed
 
     bool canBeRemovedFromPool = o.callback->Dispose(o.addr, force);
     if (!(canBeRemovedFromPool || force)) { return 0; }
 
+    available_ids.push(o.handle);
+
     handleByAddress.erase(o.addr);
-    objects.erase(handle);
-    // <-- obj reference invalid at this point.
+    o = ManagedObject();
+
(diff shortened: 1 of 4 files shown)


Engine/ac/dynobj/managedobjectpool.cpp
```diff
@@ -65,166 +47,155 @@ int32_t ManagedObjectPool::AddRef(int32_t handle) {
...
+
+    objects.insert({handle, {
+        ManagedObject {
+            obj_type: plugin_object ? kScValPluginObject : kScValDynamicObject,
+            handle: handle,
+            addr: address,
+            callback: callback,
+            refCount: 0
+        }
+    }});
```
ivan-mogilko: This does not compile using MSVS 2015 compiler. What version of standard is this kind of initialization?

sonneveld: Interesting. I'm using MSVS 2017.  It's possible I didn't test this under Windows and I've inadvertently used a gcc extension. I was aiming for this: https://docs.microsoft.com/en-us/cpp/cpp/initializing-classes-and-structs-without-constructors-cpp?view=vs-2017

sonneveld: Pushed a fix.

sonneveld:
Rebased and fixed conflicts.

ivan-mogilko:
Although I was planning similar change myself (use hashmap for address -> handle lookup), but after testing the results are not simple. Profiling shows that in regular games AddressToHandle time decreased.
But I also tested this with an experimental game which does a lot of vector/matrix calculations in script, passing them as dynamic arrays with a very short lifespan. This test displays certain downgrade in perfomance. Using the "--fps" arg to max out frames, the game does up to 74 fps with AGS 3.5.0.7 release and only around 65-67 fps with this branch.
It's hard to rely on fps test of course but I have a guess this may be related to insertion and deletion of hashmap elements. In original code elements of array were not deallocated, only reallocated if array is too small.

ivan-mogilko:
I tried increasing the GARBAGE_COLLECTION_INTERVAL constant but it did not do much effect.

But I've got an impression that GC is not working correctly at the moment. The element removal is done at SubRef call. Maybe it should be done differently.

sonneveld:
I've changed how the objects are stored. It's closer to the original with an index to a vector now. Please let me know how it goes with your test game when you have time.

sonneveld:
Just pushed a small fix.

PR #656, opened 2019-03-02
Script API: SkipCutscene and cutscene that cannot be skipped by user input

ivan-mogilko:
Implemented eSkipScriptOnly cutscene mode and SkipCutscene() function to let users little more flexibility in scripting cutscenes.


mgambrell:
this approach worked fine for me

PR #657, opened 2019-03-03
AGS3 fix audio threading

sonneveld:
This wraps all calls to the audio system with mutexes.

Tested in Windows and Linux.

The reason why I didn't tie access to the channels array to the mutex is because the audio state is more than just the channels.  There are some script commands that don't even touch channels but call multiple functions that might, or change other audio state (other variables in audio.cpp or the audio settings in GameState and GameSetup). For example, code that reads volumes, modifies it and saves the volume back needs to be protected.

If we can cover the whole audio state, then we can consider moving more audio system code into the other thread (such as cross fading).


ivan-mogilko:
Ah, you have went the straightforward way and just put locks in every function :) (when I tried this several years ago I planned on a big audiochannels refactoring first).

Ok, this is cool if it works (haven't tested myself yet), but I told this earlier to @mgambrell that I am very wary about desync issues we had with this in the past.

1) There are cases when the game events are bound to music, either by checking Position, or intuitively (by experimentally finding proper game tick perhaps). For example, a character or an object could appear at certain note. Threaded audio may desync these.

2) Some games have video and audio played separately. I've seen at least one case when there was a terrible desync if threaded audio was on.

If these issues are still the case I think an option to disable it is a must have at least until there is a solution (like programmatically synchronizing video and audio).

Engine/media/audio/audio_system.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "media/audio/audio.h"
+
+#include "media/audio/soundcache.h"
+
+#include "media/audio/soundclip.h"
+#include "media/audio/sound.h"
+
+#include "media/audio/queuedaudioitem.h"
+
```
ivan-mogilko: Uh! I was specifically splitting them up to avoid including something that does not need to be in every particular unit. :/

There were couple of more big headers in the project like this which I wanted to remove. Now I will be wanting to remove this. Or refactor whole thing again to prevent the need of such include list.

sonneveld:
> If these issues are still the case I think an option to disable it is a must have at least until there is a solution (like programmatically synchronizing video and audio).

A couple of example games for this would be good to test against.

> **PS.** also in our yesterdays talk mgambrell mentioned something about _std threads_ not being fully portable.

Yes but if that's the case we can wait for his console patches. I'm just using c++11 libraries at the moment.

Engine/media/audio/audio_system.h
sonneveld: Do we need such fine grained include files though? What are we gaining? I'm not suggesting one giant include file, but perhaps one per system would be managable? As it is, I had to add/remove include files just for audio as I added/moved code around.

ivan-mogilko: Maybe not if few headers are always included together into source files.

But I think it helps to understand the ties between program parts, as well as the problems with these ties. Right now  audio code has different layers of abstraction mashed together: from script commands to clip interface and sound cache (which should be completely internal perhaps).

ivan-mogilko: When I am saying "helps to understand", I mean having this (small) inconvenience is a kind of reminder to self that something is wrong with it.

sonneveld: I was thinking maybe we only need to worry about specific headers within the media/audio directory. Source outside of the directory could use the audio_system.h directly. When we tease out higher and lower level apis, we could split the audio directory and have separate headers.

ivan-mogilko: Hmm ok, this may have sense too.

ivan-mogilko: I'd propose to only include this in *.cpp files though.

sonneveld: > I'd propose to only include this in *.cpp files though.
Agreed. In this pull request I deliberately didn't add audio_system.h to any headers. We should still aim to only forward declare types if possible.

ivan-mogilko:
I undestand why this was done like this, but I'd like to mention my worry, since channels are still not hidden behind API in any way you have to add locks everywhere, and this means that if anyone will add a new audio command they have to remember to put lock there too.

I wish I could unbury my old channels refactoring code which main goal was to protect channels from direct access, and apply it again, but idk if I will have spare time for that.

sonneveld:
> If these issues are still the case I think an option to disable it is a must have at least until there is a solution (like programmatically synchronizing video and audio).

Okay, so ideally we try to iron out any desync issues first before we make this an option. If we make multi-thread the default, it would force us to fix issues.

sonneveld:
> since channels are still not hidden behind API in any way you have to add locks everywhere, and this means that if anyone will add a new audio command they have to remember to put lock there too.

At the very least, the new locks highlight exactly what needs to be refactored.  It's not just access to the channels that needs to be serialised but channel operations that require multiple reads/writes. (e.g. get volume, modify value, set volume). 

So protecting channels from direct access wouldn't prevent consistency issues unless the whole operation was atomic.

ivan-mogilko:
This was my approach for protecting the channels in the past: https://github.com/ivan-mogilko/ags-refactoring/commit/171cfd055d436ddf0596d800c34ed302e8f5eaee

I see you chose to use global functions (e.g. is_channel_playing).

sonneveld:
> I see you chose to use global functions (e.g. is_channel_playing).

It meant could check for null channels within the function, and I assumed we didn't need to support multiple audio systems. :)

I think passing around channel ids vs pts is okay. And I think rather than refactoring current code, with the aim to one day add new functionality, maybe we should look at writing new systems and implementing the old system on top? That way we can write new script functions that only use the new system and keep the legacy compatibility system as a separate module.

ivan-mogilko:
Looking back what I did there was mostly gathered data separated into several arrays into one struct, and also cleaned the code to use a reference/pointer to an object instead of `channels[xxx].member` syntax.

Engine/media/audio/audio_system.h
rofl0r: > But I think it helps to understand the ties between program parts, as well as the problems with these ties.

it also helps to speed up the build

mgambrell:
see my alternative above

We already have architecture for systems to provide their own synchronization primitives. std:: stuff is for people with no other alternative. We already have an alternative.

If we're not sure they're recursive in our implementations, then let's decree them to be so. The game works fine with recursive mutexes right now. Recursive mutexes can be implemented if needed on top of non-recursive mutexes.

sonneveld:
> If we're not sure they're recursive in our implementations, then let's decree them to be so. The game works fine with recursive mutexes right now. Recursive mutexes can be implemented if needed on top of non-recursive mutexes.

I can merge this in now and it'll work for our supported platforms. Work can be done in a separate pull request to update the mutexes to ensure recursive is supported and add timeouts.

mgambrell:
Timeouts are rarely needed. If you actually have a use for that (seems unlikely) it should be added as a separate facility.

sonneveld:
> Timeouts are rarely needed. If you actually have a use for that (seems unlikely) it should be added as a separate facility.

I use them here to crash on deadlock just in case.

mgambrell:
If you want to print an 'it crashed' message instead of hang for personal preference reasons in your backend ports, then you can do that through logic in your own mutex implementation

Engine/ac/audiochannel.cpp
```diff
@@ -28,11 +28,15 @@ extern CCAudioClip ccDynamicAudioClip;
 
 int AudioChannel_GetID(ScriptAudioChannel *channel)
 {
+    std::lock_guard<std::recursive_timed_mutex> lock(channels_mutex);
+
     return channel->id;
```
ivan-mogilko: A small note, regarding this, I think the lock is redundant here, because ScriptAudioChannel object itself is never deleted during game play, and it's ID is always fixed.

I think there's a confusion between `ScriptAudioChannel *channel` and "channel" from `channels` array, which is actually SOUNDCLIP* (and may be deleted, and changed).

sonneveld: Good point. I was using the locks to also mark out any audio usage but in this case it's not relevant.

sonneveld:
> If you want to print an 'it crashed' message instead of hang for personal preference reasons in your backend ports, then you can do that through logic in your own mutex implementation

Could we compromise with a macro like "START_CRITICAL_AUDIO_SECTION", so I can use the standard std::mutexes but also allow swapping out with something else ?  Currently it has all the functionality I need and I know it works.

sonneveld:
I added some design notes to the pull request description.

mgambrell:
It is better to control access to the critical data structures than sprinkle critical sections all over and hope you catch everything. Typically you should even serialize update commands into a pipe for processing on the audio thread (this is done so that no locking is required). You can easily ruin performance by locking everything without having thought through what needs locking clearly. It seemed to me all the internal audio state was rightly the business of the main VM thread and it didn't need to lock things to protect it from itself. Only the handoff between threads needs locking. If you can find an instance of a channel using a resource outside the channel, then that should be addressed on a case by case basis.

sonneveld:
> Typically you should even serialize update commands into a pipe for processing on the audio thread (this is done so that no locking is required). You can easily ruin performance by locking everything without having thought through what needs locking clearly.

I hope to one day refactor the audio system into something like this. I believe identifying and marking out the critical sections is a good first step.

mgambrell:
I believe that you have incorrectly identified the critical sections, by marking things which are strictly internal concerns for the scripting logic. The interface to the asynchronous process is the channels. The critical sections are those which touch the data structures shared with the asynchronous process (the channels)

ivan-mogilko:
> I believe that you have incorrectly identified the critical sections, by marking things which are strictly internal concerns for the scripting logic.

BTW, yes, I think this may be incorrect (and at least superflous) to put locks right in the script callbacks.

sonneveld:
The channels is one part, but there will be more if we move cross fading for example. I also wanted to ensure consistency within a script command that calls out to other functions that interacts with critical sections

mgambrell:
Whatever you do to cross fading, if it results in a change to the set of synchronized data structures, then run those data structures through the synchronization. Littering the code with critical sections does not solve the underlying problem of undisciplined communication between threads.

sonneveld:
I'm not saying it solves it. I'm just saying it's a first step in identifying sections that interact with the audio system that might require some more thought. It may be slower (I haven't noticed any problems with that yet) but I think a bit safer.

mgambrell:
You're identifying the wrong sections. Sections that interact internally in the script engine arent important. 

Begin by identifying DATA that's being shared across threads without discipline. Then change the name of that data, so that all the code that references it is now automatically failing to compile. That way the identification of the code is foolproof. Gate access to that data. Make the gate a NOP, and leave your TODO in that gate. That's the first step.

ivan-mogilko:
> but I'm also interested in possible consistency issues where vm code interacts with audio. (as in, what happens if channel[].get_pos_ms changes between calls in a single script command) I essentially did a "flood fill" with the audio locks from the exposed audio.h symbols to cover this.

IMHO the good idea in such case is to acquire the channel's data at the start of the function and use that locked data (e.g. returned by a retrieving/locking function or object) until the end instead of freely accessing channel over and over.

sonneveld:
> Begin by identifying DATA that's being shared across threads without discipline. Then change the name of that data, so that all the code that references it is now automatically failing to compile.

But how does that help with consistency between two separate but gated sections of code? I feel it would be better to start with general critical sections for audio, so we can review and reduce those sections over time.

mgambrell:
Only a gate is going to ensure consistency between two separate but gated sections of code. If they're related, it's because they're modifying mutable global state which should be behind the gate

sonneveld:
I'm breaking up some of the non-multithreading changesets into their own pull requests.

ivan-mogilko:
Merged #689 instead.

PR #659, opened 2019-03-03
AGS3: big refactor to put all audio logic under control of one mutex

mgambrell:
alternative to #657

I don't make the threading mandatory, in deference to ivan's wariness that it could potentially affect a/v synchronization of some games. This commit was the conservative, minimum work required to get the synchronization under control.

My approach was to rename `channels` to `_channels` so that it wasn't possible to use them on accident without revising the code at this time to fix the synchronization. Then access to the channels was kept hidden and provided through a specialized synchronization structure (a mutex) 

That way, nobody can forget to lock the channels even in the future.

I've tested this in my game and it seems bulletproof

In this commit I also made _channels a std::array to help me debug it

### DIFF
--- a/Engine/ac/audiochannel.cpp
+++ b/Engine/ac/audiochannel.cpp
@@ -39,8 +39,11 @@ int AudioChannel_GetIsPlaying(ScriptAudioChannel *channel)
         return 0;
     }
 
-    if ((channels[channel->id] != NULL) &&
-        (channels[channel->id]->done == 0))
+    AudioChannelsLock _lock;
+    auto* ch = _lock.GetChannel(channel->id);
+
+    if ((ch != nullptr) &&
+        (ch->done == 0))
     {
         return 1;
     }
@@ -49,10 +52,13 @@ int AudioChannel_GetIsPlaying(ScriptAudioChannel *channel)
 
 int AudioChannel_GetPanning(ScriptAudioChannel *channel)
 {
-    if ((channels[channel->id] != NULL) &&
-        (channels[channel->id]->done == 0))
+    AudioChannelsLock _lock;
+    auto* ch = _lock.GetChannel(channel->id);
+
+    if ((ch != nullptr) &&
+        (ch->done == 0))
     {
-        return channels[channel->id]->panningAsPercentage;
+        return ch->panningAsPercentage;
     }
     return 0;
 }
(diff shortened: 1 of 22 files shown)


Engine/ac/audiochannel.cpp
```diff
@@ -39,8 +39,11 @@ int AudioChannel_GetIsPlaying(ScriptAudioChannel *channel)
         return 0;
     }
 
-    if ((channels[channel->id] != NULL) &&
-        (channels[channel->id]->done == 0))
+    AudioChannelsLock _lock;
```
rofl0r: wouldn't you need a global object (singleton) for this to work properly ?

mgambrell: AudioChannelsLock internally uses a singleton mutex

rofl0r: seems a bit verbose... maybe sometimes C-style non-OOP functions could be advantageous...

Engine/media/audio/audio.cpp
```diff
@@ -428,39 +386,6 @@ ScriptAudioChannel* play_audio_clip_by_index(int audioClipIndex)
         return NULL;
 }
 
-void stop_and_destroy_channel_ex(int chid, bool resetLegacyMusicSettings) {
```
sonneveld: Could these be moved back so we can compare changes?

Engine/media/audio/audio.h
```diff
@@ -25,15 +27,34 @@
 
 struct SOUNDCLIP;
 
+//controls access to the channels, since that's the main point of synchronization between the streaming thread and the user code
+//this is going to be dependent on the underlying mutexes being recursive
```
sonneveld: I think this is only true for Windows. The pthreads version doesn't configure the mutuex for recursive I think.

Engine/ac/audiochannel.cpp
mgambrell: no, RAII is advantageous

Engine/media/audio/audio.h
mgambrell: Then the pthreads version should configure the mutex to be recursive, and the platform interface headers should document that the mutexes should be recursive.

sonneveld: That should probably be done in this pull request then.

Engine/ac/audiochannel.cpp
sonneveld: Agreed, RAII for locks is extremely helpful.

Engine/ac/dynobj/cc_audiochannel.cpp
```diff
@@ -16,7 +16,7 @@
 #include "media/audio/audiodefines.h"
 #include "ac/dynobj/scriptaudiochannel.h"
 
-extern ScriptAudioChannel scrAudioChannel[MAX_SOUND_CHANNELS + 1];
+extern ScriptAudioChannel scrAudioChannel[MAX_SOUND_CHANNELS];
```
ivan-mogilko: This may break scripts, but earlier probably will crash engine at InitAndRegisterAudioObjects() where it still iterates from 0 to MAX_SOUND_CHANNELS **inclusive**.

I think this comment about `+ 1 just for safety on array iterations` could have tricked you.
MAX_SOUND_CHANNELS is a number of channels the user may reserve for their custom audio types, and +1 is a special crossfade channel. Maybe there's a sense to introduce new constant that would be equal to MAX_SOUND_CHANNELS + 1 and use here.

Engine/game/savegame_internal.h
```diff
@@ -86,7 +86,7 @@ struct RestoredData
         int PanAsPercent;
         int Speed;
     };
-    ChannelInfo             AudioChans[MAX_SOUND_CHANNELS + 1];
+    ChannelInfo             AudioChans[MAX_SOUND_CHANNELS];
     // Ambient sounds
```
ivan-mogilko: Same as in my comment above, this should remain [MAX_SOUND_CHANNELS + 1 because save keeps state of extra 1 crossfade channel too.

Engine/ac/dynobj/cc_audiochannel.cpp
mgambrell: This was an accident. I didn't mean to change any array lengths in this commit. Ignore this, I will revert it

Engine/game/savegame_internal.h
mgambrell: This was an accident. I didn't mean to change any array lengths in this commit. Ignore this, I will revert it

ivan-mogilko:
I want to make a note regarding the **MAX_SOUND_CHANNELS** vs **MAX_SOUND_CHANNELS + 1** array size. This is inconvenient design which caused confusion in the past.

There are MAX_SOUND_CHANNELS of channels that may be set up for user audio types, channel #0 is always reserved for speech voice-over. _Additionally_ to these there's 9th channel for music crossfade (frankly I don't remember if it's only channel that can be used for crossfade, but it's definitely used in legacy audio mode; but I don't want to give too much info and complicate this further).

Hence some arrays that refer to all channels state must be of MAX_SOUND_CHANNELS + 1 size (internal channel count) and some arrays that store extra data for user channels may be of MAX_SOUND_CHANNELS size. I *think* the latter are **AmbientSounds** array and **last_sound_played**. But I won't bet on that without double checking all the loops where engine iterates over them.

mgambrell:
I didn't mean for that to be in this commit. I wish we weren't even talking about it. It's a total mess and doesn't need to be dealt with now.

ivan-mogilko:
Since we have two variants of same change now, I must admit that codewise I like this approach more, because it better encourages to use channels through locks (or maybe even restricts it? I did not examine this thoroughly enough yet). Also here you changed a lot of channel[x].data uses with ch.data for simplicity.

I may be biased here because this looks closer to my [old refactor attempt](https://github.com/ivan-mogilko/ags-refactoring/commit/171cfd055d436ddf0596d800c34ed302e8f5eaee) where I started doing somewhat similar thing, but I guess I had reasons for that back then.

mgambrell:
`_channels` is static in audio.cpp now, but actually I could make it static and private inside the mutex to make it genuinely restricted. perhaps I will do that.

sonneveld:
> Since we have two variants of same change now, I must admit that codewise I like this approach more, because it better encourages to use channels through locks (or maybe even restricts it? I did not examine this thoroughly enough yet).

I intentionally kept the locks separate from accessing the channels because the audio state is more than just the channel data. I was hoping to mark out all the audio sections so we could potentially do more work (such as the crossfading and music queuing) in the audio thread.

mgambrell:
I judged the only point of sensitive synchronization to be the channels. The only thing that happens in another thread is the channel ->Poll() calls, and I assumed those were happening independent of any audio state outside the that channel.
Note, the mixing may also occur in yet another channel in allegro, but that's been dealt with inside allegro adequately, it seems.

Engine/media/audio/audio.cpp
sonneveld: It also makes merging much easier if you don't move code around. :)

mgambrell: I'll file the appropriate paperwork to move it back for purposes of analyzing this PR but frankly I don't care if it makes merging harder. If your workflow means you can't move code, then there's a problem with your workflow. Mixing it with other commits was stupid though. I forgot I'd done that early on before I had a better plan.

sonneveld: Come on, this is a collaborative project, we care if it makes it harder for others to merge.

ivan-mogilko: > there's a problem with your workflow

Yes, there is a problem with workflow: it is that in this branch we have to support multiple legacy versions of ags. Since we not always have an actual specification telling how these versions should work, every time someone finds a regression I have to go back in time and compare code.

ivan-mogilko:
So, after some internal discussion, the plan on this is:
* I will recreate this PR on a new branch, clean it up a little, reduce or get rid moving of big chunks of code and fix unnecessary changes to array sizes and indexing.
* Then I will cherry pick some of the @sonneveld 's additions also which may be useful, first of all is_channel_playing() function that simplifies number of channel access cases.
* Open a new PR allowing for other contributors to push any additional fixes (because e.g something will have to be done with mutexes on Linux, as noted above).

ivan-mogilko:
Merged as a part of #689.

PR #660, opened 2019-03-03
AGS3: Reform do once

mgambrell:
The actual problem here was in StrUtil::ReadString. DoOnce tokens would get deserialized occasionally with garbage at the end of the string. 

While debugging this I was annoyed by the old crustiness of "do once tokens". Then I noticed they mixed new[] and malloc, and that was enough to give me the go-ahead to reform them entirely to be more modern.

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -208,6 +208,7 @@ void StrUtil::ReadString(char *cstr, Stream *in, size_t buf_limit)
     size_t len = in->ReadInt32();
     if (buf_limit > 0)
         len = Math::Min(len, buf_limit - 1);
+    cstr[len] = 0;
     if (len > 0)
         in->Read(cstr, len);
     else
@@ -218,6 +219,7 @@ void StrUtil::ReadString(char **cstr, Stream *in)
 {
     size_t len = in->ReadInt32();
     *cstr = new char[len + 1];
+    (*cstr)[len] = 0;
     if (len > 0)
         in->Read(*cstr, len);
     else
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
I recall once wanting to remake "do once" tokens into the hash-set (because that's what they really should be), but some problem stopped me. I will take a look.

mgambrell:
sounds good to me. but std::unordered_set is what you should use, std::hash_set is some kind of terrible deprecated mistake

ivan-mogilko:
Yes, sorry, std::unordered_set that is, I keep calling them with generic names (hashmap, hashset etc).

Common/util/string_utils.cpp
```diff
@@ -215,6 +217,15 @@ void StrUtil::ReadString(char *cstr, Stream *in, size_t buf_limit)
...
 
+void StrUtil::ReadString(String *s, Stream *in)
+{
+    size_t len = in->ReadInt32();
+    s->FillString('\0',len);
+    char* ptr = (char*)s->GetCStr(); //yes, this is bad.
+    if (len > 0)
+        in->Read(ptr, len);
+}
+
```
ivan-mogilko: I think you do not need this function. There's another variant of ReadString above that returns String object:
<pre>
String StrUtil::ReadString(Stream *in)
{
    size_t len = in->ReadInt32();
    if (len > 0)
        return String::FromStreamCount(in, len);
    return String();
}
</pre>

It works with String::ReadCount which ensures null-terminator (well, supposed to).

mgambrell: OK ill change this to use that, but I should leave the function because there are functions there for reading every other string type in that way and the whole reason I added the function is because it seemed like it should be there, but wasn't. Someone else will add the function on accident again one day otherwise

ivan-mogilko:
I think this is good.
Will fix the ReadString myself to make things simplier.

PR #661, opened 2019-03-04
build system: factor out freetype C/LDFLAGS

rofl0r:
this allows to override them from the make command line like

make FT_CFLAGS=-I ../foo FT_LDFLAGS=/tmp/libfreetype.a

or by defining them in config.mak


Engine/Makefile-defs.linux
```diff
@@ -1,8 +1,13 @@
+FT_CFLAGS = $(shell pkg-config --cflags freetype2)
```
sonneveld: Probably a good idea to use `:=` because they'll only be expanded once.

rofl0r: oh right. oversight on my side

PR #667, opened 2019-03-08
macos: Initial fixes to get macos building with Makefile

sonneveld:
Also add 'debug' target for.. debugging.
(note: colours are off because allegro patch assumed a different pixel format, to be fixed in a future patch) (or the SDL port)

### DIFF
--- a/Engine/Makefile
+++ b/Engine/Makefile
@@ -14,6 +14,10 @@ else
 include Makefile-defs.linux
 endif
 
+ifneq ($(USE_MIDI_PATCH), 1)
+AL_MIDI_PATCH :=
+endif
+
 OBJS_C_CPP_MM = $(ALFONT) $(ALMP3) $(ALOGG) $(APEG) $(AASTR) $(HQ2X3X) $(AL_MIDI_PATCH) $(CDA) $(ALDUMB) $(PLUGINS) $(BASE) $(BASE_PLATFORM)
 OBJS := $(OBJS_C_CPP_MM:.mm=.o)
 OBJS := $(OBJS:.cpp=.o)
@@ -31,7 +35,7 @@ all: printflags ags
 
 ags: $(OBJS) common.a 
 	@echo "Linking engine..."
-	$(CMD_PREFIX) $(CC) -o $@ $^ $(CFLAGS) $(LDFLAGS) $(LIBS)
+	$(CMD_PREFIX) $(CXX) -o $@ $^ $(CXXFLAGS) $(LDFLAGS) $(LIBS)
 
 debug: CXXFLAGS += -UNDEBUG -D_DEBUG -Og -g -pg
 debug: CFLAGS   += -UNDEBUG -D_DEBUG -Og -g -pg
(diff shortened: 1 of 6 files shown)


Engine/Makefile
```diff
@@ -41,13 +49,18 @@ common.a: $(OBJS_COMMON)
 	@echo $@
 	$(CMD_PREFIX) $(CC) $(CFLAGS) -MD -c -o $@ $<
 
+%.o: %.mm
```
rofl0r: do we really want to create dependency information? in my experience this often makes the build almost 2x slower. and the advantage added by this is questionable.

Engine/Makefile-objs
```diff
@@ -70,3 +70,43 @@ APEG = libsrc/apeg-1.2.1/adisplay.c libsrc/apeg-1.2.1/getbits.c libsrc/apeg-1.2.
 AASTR = ../Common/libsrc/aastr-0.1.1/aarot.c ../Common/libsrc/aastr-0.1.1/aastr.c ../Common/libsrc/aastr-0.1.1/aautil.c
 
 AL_MIDI_PATCH = libsrc/allegro-4.2.2-agspatch/midi.c
+
+ALDUMB = libsrc/dumb-0.9.2/allegro/alplay.c \
```
rofl0r: why is this needed now ?

rofl0r: i just checked my build recipe for ags and it was previously possible to use a system-provided libdumb installation. adding this here would force the usage of the sources shipped with ags, afaict.

Engine/Makefile
sonneveld: First off, this is to support objective-c code.  We already build dependency information and while it's not perfect, it does help when I'm doing development and incremental builds.

Engine/Makefile-objs
sonneveld: Ahah, my mistake, will remove.

sonneveld: I made it optional (on by default in macos). Since the Windows build uses the builtin source, I want to do the same to pick up on any issues.

OSX/patches/allegro-4.4.2.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 245be10..105c923 100644
+index 245be10fb..6147a68b3 100644
```
rofl0r: is this patch still required now that allegro 4.4.3 is released ?

Engine/Makefile-defs.linux
```diff
@@ -1,12 +1,18 @@
+USE_BUILT_IN_LIBSRC = 0
+USE_MIDI_PATCH = 1
```
rofl0r: it seems the midi patch is now integrated into allegro as of 4.4.3

OSX/patches/allegro-4.4.2.patch
sonneveld: For macos it is, the patch includes mac specific changes (and 4.4.3)

Engine/Makefile-defs.linux
sonneveld: This behaviour is unchanged from before. I guess it depends on what version of allegro is bundled with your Linux distro

PR #673, opened 2019-03-17
AGS3: toojay's Linux OpenGL support with Glad

sonneveld:
Rebased toojay's branch and included Glad support for Windows and Linux.


rofl0r:
what's glad ?

sonneveld:
> what's glad ?

It's an OpenGL wrapper generator. https://github.com/Dav1dde/glad It'll take the latest opengl specs and generate extension loaders for you.

rofl0r:
> It's an OpenGL wrapper generator. https://github.com/Dav1dde/glad It'll take the latest opengl specs and generate extension loaders for you.

thanks for info. does this really need to be part of this PR though? maybe it would be better to first simply merge the OpenGL support, and then start a new discussion about whether `glad` is wanted or not.

sonneveld:
> ybe it would be better to first simply merge the OpenGL support, and then start a new discussion about whether `glad` is wanted or not.

I guess? This PR already works now, simplifies extension func loading and doesn't require building/linking separate libraries. If anyone has any strong feelings about the tools used, I suggest they submit a PR later that changes it.

Engine/test/test_file.cpp
```diff
@@ -196,10 +196,10 @@ void Test_File()
...
     assert(int32val == 20);
 
-    assert(ptr32_array_in[0] == 0xABCDABCD);
-    assert(ptr32_array_in[1] == 0xFEDCFEDC);
-    assert(ptr32_array_in[2] == 0xFEEDBEEF);
-    assert(ptr32_array_in[3] == 0xBEEFFEED);
+    // assert(ptr32_array_in[0] == 0xABCDABCD);
+    // assert(ptr32_array_in[1] == 0xFEDCFEDC);
+    // assert(ptr32_array_in[2] == 0xFEEDBEEF);
+    // assert(ptr32_array_in[3] == 0xBEEFFEED);
```
ivan-mogilko: Could you explain why these tests were disabled?

sonneveld: At least on 64bit linux and osx, these assertions fail. On linux 64bi, the array has the values 0xffffffffabcdabcd 0xfffffffffedcfedc 0xfffffffffeedbeef 0xffffffffbeeffeed .

I didn't want to change the test because I wasn't sure if I could fix it to be correct for different permutations of operating system and pointer size.

ivan-mogilko: Ok, I was worried because this strange function is used in one place, although I don't remember why, and don't remember why the type of array has to be intptr_t... this could be left in the middle of some old refactoring too.

ivan-mogilko: I think this may be because it uses a buffer of signed integers which it then assigns to intrptr elements, these should be unsigned.

rofl0r:
> simplifies extension func loading and doesn't require building/linking separate libraries

at the cost of 7000 lines of code (which seems to use a jungle of ifdefs with hardcoded platform names, i.e. [the opposite of portable](https://wiki.musl-libc.org/faq.html#Q:-Application-XY-misbehaves-or-crashes-at-runtime-when-linked-against-musl))

the alternative would be to just link -lGL in the makefile (one line)

sonneveld:
> at the cost of 7000 lines of code (which seems to use a jungle of ifdefs with hardcoded platform names, i.e. [the opposite of portable](https://wiki.musl-libc.org/faq.html#Q:-Application-XY-misbehaves-or-crashes-at-runtime-when-linked-against-musl))

Isn't that the reality of a cross platform library for opengl extensions?  Glad lets you generate only the code you need for the extensions you use. That's why it's only 7000 lines vs 75,000 lines in GLEW.

Engine/test/test_file.cpp
ivan-mogilko: I wrote these tests specifically to detect if functions fail under some conditions. Now we know they do, so we need to fix the functions, not disable tests. Sorry, but I don't want to accept this pr with disabled tests. You may keep them disabled locally, just don't push these changes.
I will try running them on 64-bit Linux later today and see if I can fix this.

sonneveld: That's fair! I removed that fix from the PR. I'll try fixing it on my machine now if you like. (in a different pr)

rofl0r:
> Isn't that the reality of a cross platform library for opengl extensions? 

the question is, is it needed ? it seems @toojays changes worked well without.

> only 7000 5500 lines vs 75,000 lines in GLEW

the advantage of glew is that it's not built-in into the source tree here rather used as an external lib. and it's a pretty common library which is for example already installed on my distro.

>> the alternative would be to just link -lGL in the makefile (one line)

> But that's not cross platform either?

why not ?
also, wasn't this PR originally only to add openGL support for linux (meaning the other platforms had it already) ?

sonneveld:
Rebased and fixed merge conflicts.

ivan-mogilko:
> Rebased and fixed merge conflicts.

Sorry, I think I will be cherry-picking commits that fix actual unrelated bugs in code from your pull requests, like the one with headers :).

Regarding above rofl0r's questions, I know a little about GLEW but never heard about glad before, so I am also interested to know if it's necessary or what it improves.
Also, you said that it generates the code necessary for extensions, does it mean that if we need another function we will have to regenerate these files anew?

sonneveld:
> Sorry, I think I will be cherry-picking commits that fix actual unrelated bugs in code from your pull requests, like the one with headers :).

Hah that's totally fine and one of the reasons why I started putting them in their own commit. It's easy enough to rebase on top.

sonneveld:
> I know a little about GLEW but never heard about glad before, so I am also interested to know if it's necessary or what it improves.
> Also, you said that it generates the code necessary for extensions, does it mean that if we need another function we will have to regenerate these files anew?

Glad generates the minimum necessary function pointers to access the OpenGL api at the specified version and also any requested extensions, along with a way to query if those extensions loaded correctly. The headers are much smaller because it only includes the extensions you're interested in. 

Because we support so many different platforms, it's much easier to include the generated wrapper code than to modify our build instructions to include how to build GLEW. (Some Linux distros may distribute it, but Windows, OSX, etc would need to have it built separately)

If you wish to use other extensions, you will need to generate new wrappers. There's a link at the top of the generated files that you can use to generate with the same settings as before.

rofl0r:
> how to build GLEW. (Some Linux distros may distribute it, but Windows, OSX, etc would need to have it built separately)

isn't the MacOS / windows version already using/offering OpenGL ?

sonneveld:
> isn't the MacOS / windows version already using/offering OpenGL ?

Windows had handwritten code to load extension functions (which GLAD handles) and MacOS currently doesn't use the ali3dogl.cpp driver, it hides it in the allegro graphics driver (for now). In the SDL2 port I can rely on SDL to create a window with opengl context in a cross platform way so that allows the same opengl code to be used on mac too.

ivan-mogilko:
So, I overviewed these glad files, and they practically contain the code that I had to add manually for Windows anyway (declaring function pointers and loading adresses from dll), except more verbose and having alternate branches.

> In the SDL2 port I can rely on SDL to create a window with opengl context in a cross platform way so that allows the same opengl code to be used on mac too.

In theory, will all this glad code still used with SDL?

sonneveld:
> So, I overviewed these glad files, and they practically contain the code that I had to add manually for Windows (declaring function pointers and loading adresses from dll), except more verbose and having alternate branches.

That's all any of the opengl wrappers do essentially. Glad just produces fairly straight forward code that isn't too hard to figure out.

> In theory, will all this glad code still used with SDL?

Yes, they use similar files. The only difference is that it can use SDL's library loader.

> PS. I'd like to test compile this branch for Android to see if all those replacements broke anything.

Ideally this shouldn't affect the Android branch because all the changes are behind #ifdefs. I haven't got an android build environment setup at the moment so I would appreciate any suggestions.

ivan-mogilko:
Windows: works, no problems seen so far.

Android: compilation was broken. I had to do a quick patch to be able to run: https://github.com/ivan-mogilko/ags-refactoring/commit/8325723a96c176c527f3eba055aa373436dbeae7
Any thoughts on how to deal with this properly?

Also, for Android I had to fix agspalrender: https://github.com/adventuregamestudio/ags/commit/53d338040e72caf80403c97514be29ba5e69c508

morganwillcock:
> Linux: works, and I can finally play a game without bizzare slowdowns. Still not sure what have caused these, maybe software video mode is incompatible?

Do you mean because you are running it in a virtual machine?

ivan-mogilko:
> Do you mean because you are running it in a virtual machine?

Yes, I guess it happens only on my vm, otherwise every Linux user would report ags is unusable.

sonneveld:
> Yes, I guess it happens only on my vm, otherwise every Linux user would report ags is unusable.
> Maybe I did not set it up correctly too.

I find audio on my linux VM is pretty bad. I believe it to be an allegro issue since it works quite well using openal.

sonneveld:
> Android: compilation was broken. I had to do a quick patch to be able to run: [ivan-mogilko@8325723](https://github.com/ivan-mogilko/ags-refactoring/commit/8325723a96c176c527f3eba055aa373436dbeae7)
> Any thoughts on how to deal with this properly?

Hrmm, It should support at least OpenGL 2 but I don't know if the shaders will compile. It looks like we'd need to use a slightly different extension to access the framebuffers though.  We could ifdef the 'print opengl version' code to remove the need for the opengl version struct. Otherwise, considering the hacks we have in place to get Android working in the first place, your work around doesn't look too bad :)

ivan-mogilko:
>  but I don't know if the shaders will compile

That does not matter atm, because they won't run correctly (I tried before and it refused to pass input values). I have a guess this is related to the fixed pipeline we are still using.

> Otherwise, considering the hacks we have in place to get Android working in the first place, your work around doesn't look too bad :)

The framebuffers extensions should be fixed correctly though, because they are in use on Android. GLVersion may be ignored for now, although I'd wish it printed opengl version somehow for diagnostic reasons.

PR #676, opened 2019-03-18
AGS3 Refactor script code loading

sonneveld:
This removes the use of the `(Read|Write)ArrayOfIntPtr32` and loads code as int32, explicitly mapping to int_ptr when loading into an cc_instance.

This works because we use FIXUPs to replace any use of pointers (which wouldn't work when truncated to 32bit) with new pointers in cc_instance.

In the future, it might be worth editing the Save script method to strip pointers out of the rawCode buffer.

### DIFF
--- a/Common/script/cc_script.cpp
+++ b/Common/script/cc_script.cpp
@@ -59,6 +59,7 @@ ccScript::ccScript()
 {
     globaldata          = NULL;
     globaldatasize      = 0;
+    code                = nullptr;
     rawCode             = nullptr;
     codesize            = 0;
     strings             = NULL;
@@ -80,6 +81,7 @@ ccScript::ccScript()
     capacitySections    = 0;
 }
 
+#if 0
 ccScript::ccScript(const ccScript &src)
 {
     globaldatasize = src.globaldatasize;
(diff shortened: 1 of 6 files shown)


sonneveld:
On mentioning the 'save' method, I realised this might break the compiler. I'll test on a Windows machine when I have a chance.

ivan-mogilko:
Is it possible to avoid renaming the "code" variable? It's used everywhere in compiler.

sonneveld:
> Is it possible to avoid renaming the "code" variable? It's used everywhere in compiler.

I've reintroduced it, but it's only for the compiler now. Turns out the copy constructor was unused so I've disabled it.

ivan-mogilko:
> I've reintroduced it, but it's only for the compiler now. Turns out the copy constructor was unused so I've disabled it.

I don't understand, why is there a need for these complications? In my opinion "raw code" is a misleading name (as, raw compared to what? it's just "code"); it's rather the array in ccInstance that is not exactly code, but also containing substituting pointers/offsets. The only place when the two "codes" meet is when ccInstance is created from ccScript, and I think it may be enough to simply have a comment at the ccInstance member explaining why it has different size.

ivan-mogilko:
Oh, sorry, I see now, the compiler works with intptr_t too for some reason.

This should be investigated, but I am almost certain this has not real effect since the only way this code was used is in 32-bit application, so these values are always 32-bit anyway.

The reason they are intptr_t is probably because all the script related code used "long" in the past and that was at some point auto replaced with intptr_t. Which could be a wrong thing to do.

ivan-mogilko:
For the reference, here's wiki page on byte-code format: https://github.com/adventuregamestudio/ags/wiki/Script-execution#instructions

ivan-mogilko:
I redid this by changing "code" arrays in script compiler from intptr_t to int32_t, as they should be: f6ce25236f70c702e6fe070be94f995d8c8c258a

Then copied your changes to ccInstance initialization and removed Read/WriteIntPtr methods from stream.

PR #679, opened 2019-03-20
AGS3: Make fps display stable and make timing workable when maxed-framerate

sonneveld:
Use a finer time source to calculate fps.

I found some parts of the engine were using the requested framerate instead of the actual framerate. This works fine except when in maxed framerate, which never reaches 1000fps :) This PR fixes that too. (only when in maxed framerate mode, in normal gameplay it uses the requested framerate always)


ivan-mogilko:
Just want to say for the future, I must fix argument naming because calling function arguments _x goes against our own coding convention, and afaik also against C/C++ recommendations because _x names in global namespaces may be reserved by library.

PR #682, opened 2019-03-20
AGS3 audio: tidy up audio polling

sonneveld:
- always update cross fade before rendering graphics (to account for vsync)
- rename to update_audio_system_on_game_loop from update_polled_audio_and_crossfade to make it clearer when to be called.
- move music volume update into update_audio_system_on_game_loop since it relies on timer anyway.


ivan-mogilko:
Used cherry-pick because there were conflicts (after previous pr merges) and fixing them with github tool causes whole main merged into your pr branch).

PR #683, opened 2019-03-20
AGS3 audio: replace channel->done with channel_is_playing()

sonneveld:
Ensures the null check and the done check are done at the same time.

### DIFF
--- a/Engine/ac/audiochannel.cpp
+++ b/Engine/ac/audiochannel.cpp
@@ -39,18 +39,12 @@ int AudioChannel_GetIsPlaying(ScriptAudioChannel *channel)
         return 0;
     }
 
-    if ((channels[channel->id] != NULL) &&
-        (channels[channel->id]->done == 0))
-    {
-        return 1;
-    }
-    return 0;
+    return channel_is_playing(channel->id);
 }
 
 int AudioChannel_GetPanning(ScriptAudioChannel *channel)
 {
-    if ((channels[channel->id] != NULL) &&
-        (channels[channel->id]->done == 0))
+    if (channel_is_playing(channel->id))
     {
         return channels[channel->id]->panningAsPercentage;
     }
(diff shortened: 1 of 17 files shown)


ivan-mogilko:
Same, cherry-picked this commit and resolved conflicts to save time: 0147d091a917adaab191541aba8df5f32b85d1e7

I don't know if I may say this, but I'm starting to hate "auto" keyword. I don't know what benefit does it give when replaces simple "bool" and "int" types, but it makes it harder to understand what kind of variable is created.

ivan-mogilko:
Please take a note: I had to ammend this with a fix: ded7fb034fd21385033717a76f4d90fe4cccefb4

Some of the speech-related checks rely on clip's presence rather than its state. I am not 100% sure about all the cases, but that may be for a reason. There are situations when the speech clip stays in channel a bit longer after it stopped playing, while game does final processing (like waiting for text to be read or animations finished). In these cases it must know that the playback *was* there.

sonneveld:
> Some of the speech-related checks rely on clip's presence rather than its state.

Good point that's why I added the 'wasSpeaking' in display.cpp _display_main(), before the main loop, since it could be interpreted as tested if there was a speech channel open. I also figured with locking, and because we delete clips immediately when done, you wouldn't see a channel that had a value and was done.

sonneveld:
> I don't know if I may say this, but I'm starting to hate "auto" keyword.

I quite like it when it comes to using templated types. I use VSCode on the mac and if I hover over the code it will tell me the inferred type.

There's an argument for not using auto for simple types, but having auto in place means I can change function signatures without worrying about casts.

There's an interesting article about embracing auto here: https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/

ivan-mogilko:
I forgot to reply to this, but this comment was bugging me a lot:

> There's an argument for not using auto for simple types, but having auto in place means I can change function signatures without worrying about casts.

TBH that sounds pretty dangerous because it will silently change types of local variable which in turn may lead to incorrect implicit casts down the code. I'd rather have compiler errors pointing me to these locations.

Personally, when I do changes to function return type or even argument type I often rename the function to some dummy name and use compiler's help to see where program breaks to review every case and make sure there won't be anything implicit.

PR #685, opened 2019-03-22
Timing: use c++ std timers instead of maintaining own timer thread.

sonneveld:
- removes timer thread
- use waitingForTicks and skipMissedTicks for managing loop delays
- absolute time for scheduled music update
- absolute time for ignore_user_input (and correct save/restore behaviour)
- common implementation for platform delay

### DIFF
--- a/Engine/ac/gamestate.cpp
+++ b/Engine/ac/gamestate.cpp
@@ -24,6 +24,7 @@
 #include "util/alignedstream.h"
 #include "util/string_utils.h"
 #include "media/audio/audio_system.h"
+#include "ac/timer.h"
 
 using namespace AGS::Common;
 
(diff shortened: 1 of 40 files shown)


Engine/ac/gamestate.cpp
```diff
@@ -647,7 +648,8 @@ void GameState::WriteForSavegame(Common::Stream *out) const
     }
     out->WriteInt32( text_min_display_time_ms);
     out->WriteInt32( ignore_user_input_after_text_timeout_ms);
-    out->WriteInt32( ignore_user_input_until_time);
+    auto ignore_user_input_until_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(ignore_user_input_until_time - AGS_Clock::now());
+    out->WriteInt32( ignore_user_input_until_time_ms.count() );
```
ivan-mogilko: I'd wish we did not save this at all, does not make much sense to restore this value after loading a game IMO. It's my oversight that I did not remove it from new save format (there were too many variables to investigate so I got lazy and used same function again)

sonneveld: That sounds reasonable. On the other hand, this actually works now, if someone manages to save in the short time input is ignored, it will continue to be ignored. :)

sonneveld:
I updated pull request with a better timing method that will detect if lagging.

ivan-mogilko:
Some conflicts has to be resolved, I was not sure how trivial this is, so will leave them for you this time :).

sonneveld:
> Some conflicts has to be resolved, I was not sure how trivial this is, so will leave them for you this time :).

Heh thank you. I have pushed a new rebased version with floating point fps display.

ivan-mogilko:
I made a mistake not testing this properly before merging, the compilation was broken - in general - because of leftover function declarations, and in Windows in particular because some unix only (?) headers and functions were used in debug configuration. Tried to fix this in c6d1601615b6dd8936a6db50861baa4bda279fc6.

PR #687, opened 2019-03-24
speech: ensure stopping speech always adjust music volume.

sonneveld:
- multithreading audio deletes channels when done, we miss speech events.
- check against 'music_was' to see if we need to update it, not if speech playing.
- move 'is_voice' variables closer to where they're used.


ivan-mogilko:
Sorry, but I have different plans regarding this. Part of it is that "stop speech" becomes callable from script, which means that it may theoretically be called by user when there was no speech at all.
I am not completely sure "music_volume_was" is reliable as it is now. Instead I'd like to introduce a reliable speech state in GameState structure.

sonneveld:
> which means that it may theoretically be called by user when there was no voice-over at all.

No worries. Although I tried to rework stop_speech in such a way that it would still work even if there was no speech playing.

sonneveld:
Closing because of PR https://github.com/adventuregamestudio/ags/pull/696

PR #688, opened 2019-03-24
audio: simplify clip->destroy behaviour (for post-channel lock branch)

sonneveld:
This fixes behaviour where channels were deleted after poll. Now they are only deleted when replaced. This hopefully fixes all issues where speech channels need to hang around when complete.

Previously 'poll would destroy if 'done' && 'multithreaded'. 'play' would destroy/delete on error. This lead to some issues where clips were assigned to channel and played, but then not checked if they succeeded, leading to pointers to freed memory.

Destroy/delete of clips is now up to the main thread. Soundclips will not attempt to delete themselves or destroy during poll. Code calling 'play' will clean up clip if it is not assigned to a channel (where it would get destroyed upon replacement).

Warning: this will break audio multithreading until channel locks are in. A lot of this was work arounds for incorrect locking behaviour, and possibly for not having recursive locks.

### DIFF
--- a/Engine/ac/audiochannel.cpp
+++ b/Engine/ac/audiochannel.cpp
@@ -43,11 +43,10 @@ int AudioChannel_GetIsPlaying(ScriptAudioChannel *channel)
 
 int AudioChannel_GetPanning(ScriptAudioChannel *channel)
 {
-    AudioChannelsLock _lock;
-    auto* ch = _lock.GetChannel(channel->id);
+    AudioChannelsLock lock;
+    auto* ch = lock.GetChannelIfPlaying(channel->id);
 
-    if ((ch != nullptr) &&
-        (ch->done == 0))
+    if (ch)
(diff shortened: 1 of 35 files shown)


sonneveld:
I've pushed a new version with a few misc fixes. Let me know if it's okay to cherry pick from here or you wish me to create a new PR.

ivan-mogilko:
So is it best to merge before or after audio locks?

sonneveld:
Everything but the last commit can be individually cherry picked. But if you like I can rebase after the locks branch goes in.

sonneveld:
I tested this with the new lock changes and it seems to work with no problems. I'll rebase this once the lock story goes in.

sonneveld:
I've rebased on top of your `ags3--audiothreadlocks--chpicked` branch. It should be easy to merge this after now.

PR #689, opened 2019-03-24
AGS 3: cherry picked and tidied mgambrell's AudioChannelsLocks commit

ivan-mogilko:
This pull request does following:

1. Cherry-picked @mgambrell's commit with channel locks from #659, with following edits:
* fixed few cases of broken logic, incorrect changes to array size etc;
* reduced big changes to indentation;
* reverted move of big chunks of code (not related to the commit's purpose);
the last two edits have a purpose to make merge cleaner.

2. Ammended this with little more refactoring, to comply with recent changes to audio code.

I believe that we need two more things here:
* Fix mutexes to work recursively on Linux too
* There are several places in soundclip implementations (OGG, MP3, MIDI) where AudioChannelsLock has to be placed, see TODO comments.
* anything else?

### DIFF
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -274,14 +274,15 @@ void Game_SetAudioTypeVolume(int audioType, int volume, int changeType)
     if ((changeType == VOL_CHANGEEXISTING) ||
         (changeType == VOL_BOTH))
     {
-        AudioChannelsLock _lock;
+        AudioChannelsLock lock;
         for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
         {
             ScriptAudioClip *clip = AudioChannel_GetPlayingClip(&scrAudioChannel[aa]);
             if ((clip != NULL) && (clip->type == audioType))
             {
-                auto* ch = _lock.GetChannel(aa);
-                ch->set_volume_percent(volume);
+                auto* ch = lock.GetChannel(aa);
+                if (ch)
+                    ch->set_volume_percent(volume);
             }
         }
     }
@@ -298,12 +299,11 @@ void Game_SetAudioTypeVolume(int audioType, int volume, int changeType)
 }
 
 int Game_GetMODPattern() {
-    AudioChannelsLock _lock;
-    auto* music_ch = _lock.GetChannel(SCHAN_MUSIC);
-    if (current_music_type == MUS_MOD && music_ch) {
-        return music_ch->get_pos();
-    }
-    return -1;
+    if (current_music_type != MUS_MOD)
+        return -1;
+    AudioChannelsLock lock;
+    auto* music_ch = lock.GetChannelIfPlaying(SCHAN_MUSIC);
+    return music_ch ? music_ch->get_pos() : -1;
 }
 
 //=============================================================================
@@ -1754,13 +1754,13 @@ void stop_fast_forwarding() {
         newmusic(play.end_cutscene_music);
 
     {
-    AudioChannelsLock _lock;
+    AudioChannelsLock lock;
 
     // Restore actual volume of sounds
     for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
     {
-        auto* ch = _lock.GetChannel(aa);
-        if ((ch != nullptr) && (!ch->done))
+        auto* ch = lock.GetChannelIfPlaying(aa);
+        if (ch)
         {
             ch->set_mute(false);
         }
@@ -1876,10 +1876,10 @@ void display_switch_out_suspend()
 
     {
     // stop the sound stuttering
-    AudioChannelsLock _lock;
+    AudioChannelsLock lock;
     for (int i = 0; i <= MAX_SOUND_CHANNELS; i++) {
-        auto* ch = _lock.GetChannel(i);
-        if ((ch != nullptr) && (ch->done == 0)) {
+        auto* ch = lock.GetChannelIfPlaying(i);
+        if (ch) {
             ch->pause();
         }
     }
@@ -1915,10 +1915,10 @@ void display_switch_in_resume()
     display_switch_in();
 
     {
-    AudioChannelsLock _lock;
+    AudioChannelsLock lock;
     for (int i = 0; i <= MAX_SOUND_CHANNELS; i++) {
-        auto* ch = _lock.GetChannel(i);
-        if ((ch != nullptr) && (ch->done == 0)) {
+        auto* ch = lock.GetChannelIfPlaying(i);
+        if (ch) {
             ch->resume();
         }
     }
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -269,18 +269,19 @@ void Game_SetAudioTypeVolume(int audioType, int volume, int changeType)
         quitprintf("!Game.SetAudioTypeVolume: volume %d is not between 0..100", volume);
     if ((audioType < 0) || (audioType >= game.audioClipTypeCount))
         quitprintf("!Game.SetAudioTypeVolume: invalid audio type: %d", audioType);
-    int aa;
 
     Debug::Printf("Game.SetAudioTypeVolume: type: %d, volume: %d, change: %d", audioType, volume, changeType);
     if ((changeType == VOL_CHANGEEXISTING) ||
         (changeType == VOL_BOTH))
     {
-        for (aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
+        AudioChannelsLock _lock;
+        for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
         {
             ScriptAudioClip *clip = AudioChannel_GetPlayingClip(&scrAudioChannel[aa]);
             if ((clip != NULL) && (clip->type == audioType))
             {
-                channels[aa]->set_volume_percent(volume);
+                auto* ch = _lock.GetChannel(aa);
+                ch->set_volume_percent(volume);
             }
         }
     }
@@ -297,8 +298,10 @@ void Game_SetAudioTypeVolume(int audioType, int volume, int changeType)
 }
 
 int Game_GetMODPattern() {
-    if (current_music_type == MUS_MOD && channel_is_playing(SCHAN_MUSIC)) {
-        return channels[SCHAN_MUSIC]->get_pos();
+    AudioChannelsLock _lock;
+    auto* music_ch = _lock.GetChannel(SCHAN_MUSIC);
+    if (current_music_type == MUS_MOD && music_ch) {
+        return music_ch->get_pos();
     }
     return -1;
 }
@@ -1316,14 +1319,12 @@ void restore_game_thisroom(Stream *in, RestoredData &r_data)
 
 void restore_game_ambientsounds(Stream *in, RestoredData &r_data)
 {
-    int bb;
-
     for (int i = 0; i < MAX_SOUND_CHANNELS; ++i)
     {
         ambient[i].ReadFromFile(in);
     }
 
-    for (bb = 1; bb < MAX_SOUND_CHANNELS; bb++) {
+    for (int bb = 1; bb < MAX_SOUND_CHANNELS; bb++) {
         if (ambient[bb].channel == 0)
             r_data.DoAmbient[bb] = 0;
         else {
@@ -1752,14 +1753,19 @@ void stop_fast_forwarding() {
     if (play.end_cutscene_music >= 0)
         newmusic(play.end_cutscene_music);
 
+    {
+    AudioChannelsLock _lock;
+
     // Restore actual volume of sounds
     for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
     {
-        if (channel_is_playing(aa))
+        auto* ch = _lock.GetChannel(aa);
+        if ((ch != nullptr) && (!ch->done))
         {
-            channels[aa]->set_mute(false);
+            ch->set_mute(false);
         }
     }
+    } // -- AudioChannelsLock
 
     update_music_volume();
 }
@@ -1868,12 +1874,16 @@ void display_switch_out_suspend()
     if (set_display_switch_mode(SWITCH_BACKGROUND) == -1)
         set_display_switch_mode(SWITCH_BACKAMNESIA);
 
+    {
     // stop the sound stuttering
+    AudioChannelsLock _lock;
     for (int i = 0; i <= MAX_SOUND_CHANNELS; i++) {
-        if (channel_is_playing(i)) {
-            channels[i]->pause();
+        auto* ch = _lock.GetChannel(i);
+        if ((ch != nullptr) && (ch->done == 0)) {
+            ch->pause();
         }
     }
+    } // -- AudioChannelsLock
 
     platform->Delay(1000);
 
@@ -1904,11 +1914,15 @@ void display_switch_in_resume()
 {
     display_switch_in();
 
+    {
+    AudioChannelsLock _lock;
     for (int i = 0; i <= MAX_SOUND_CHANNELS; i++) {
-        if (channel_is_playing(i)) {
-            channels[i]->resume();
+        auto* ch = _lock.GetChannel(i);
+        if ((ch != nullptr) && (ch->done == 0)) {
+            ch->resume();
         }
     }
+    } // -- AudioChannelsLock
 
     // clear the screen if necessary
     if (gfxDriver && gfxDriver->UsesMemoryBackBuffer())
--- a/Engine/ac/global_audio.cpp
+++ b/Engine/ac/global_audio.cpp
@@ -58,11 +58,8 @@ void PlayAmbientSound (int channel, int sndnum, int vol, int x, int y) {
     if (aclip && !is_audiotype_allowed_to_play((AudioFileType)aclip->fileType))
         return;
 
-    AudioChannelsLock _lock;
-
     // only play the sound if it's not already playing
-    if ((ambient[channel].channel < 1) || (_lock.GetChannel(ambient[channel].channel) == nullptr) ||
-        (_lock.GetChannel(ambient[channel].channel)->done == 1) ||
+    if ((ambient[channel].channel < 1) || (!channel_is_playing(ambient[channel].channel)) ||
         (ambient[channel].num != sndnum)) {
 
             StopAmbientSound(channel);
@@ -79,7 +76,7 @@ void PlayAmbientSound (int channel, int sndnum, int vol, int x, int y) {
             debug_script_log("Playing ambient sound %d on channel %d", sndnum, channel);
             ambient[channel].channel = channel;
             asound->priority = 15;  // ambient sound higher priority than normal sfx
-            _lock.SetChannel(channel, asound);
+            set_clip_to_channel(channel, asound);
     }
     // calculate the maximum distance away the player can be, using X
     // only (since X centred is still more-or-less total Y)
@@ -108,12 +105,10 @@ int IsSoundPlaying() {
     if (play.fast_forward)
         return 0;
 
-    AudioChannelsLock _lock;
-
     // find if there's a sound playing
+    AudioChannelsLock lock;
     for (int i = SCHAN_NORMAL; i < MAX_SOUND_CHANNELS; i++) {
-        auto* ch = _lock.GetChannel(i);
-        if ((ch != nullptr) && (ch->done == 0))
+        if (lock.GetChannelIfPlaying(i))
             return 1;
     }
 
@@ -132,8 +127,6 @@ int PlaySoundEx(int val1, int channel) {
 
     if ((channel < SCHAN_NORMAL) || (channel >= MAX_SOUND_CHANNELS))
         quit("!PlaySoundEx: invalid channel specified, must be 3-7");
-
-    AudioChannelsLock _lock;
 
     // if an ambient sound is playing on this channel, abort it
     StopAmbientSound(channel);
@@ -159,7 +152,7 @@ int PlaySoundEx(int val1, int channel) {
 
     soundfx->priority = 10;
     soundfx->set_volume (play.sound_volume);
-    _lock.SetChannel(channel,soundfx);
+    set_clip_to_channel(channel,soundfx);
     return channel;
 }
 
@@ -201,19 +194,19 @@ int IsMusicPlaying() {
     if ((play.fast_forward) && (play.skip_until_char_stops < 0))
         return 0;
 
-    AudioChannelsLock _lock;
-
     // This only returns positive if there was a music started by old audio API
     if (current_music_type == 0)
         return 0;
 
-    if (_lock.GetChannel(SCHAN_MUSIC) == nullptr)
+    AudioChannelsLock lock;
+    auto *ch = lock.GetChannel(SCHAN_MUSIC);
+    if (ch == nullptr)
     { // This was probably a hacky fix in case it was not reset by game update; TODO: find out if needed
         current_music_type = 0;
         return 0;
     }
 
-    bool result = (_lock.GetChannel(SCHAN_MUSIC)->done == 0) || (crossFading > 0 && (_lock.GetChannel(crossFading) != nullptr));
+    bool result = (ch->is_playing()) || (crossFading > 0 && (lock.GetChannelIfPlaying(crossFading) != nullptr));
     return result ? 1 : 0;
 }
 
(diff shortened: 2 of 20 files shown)


sonneveld:
The two new channel functions will need locks too.

ivan-mogilko:
> The two new channel functions will need locks too.

Which ones?

sonneveld:
> Which ones?

The Channel_is_playing and channel_has_clip ones. I thought this story was meant to prevent accidental use of `channels` without locks?

ivan-mogilko:
I double checked now, and I've put locks in them:
https://github.com/adventuregamestudio/ags/pull/689/commits/57ff5eab5462fd0259043cfbc4a884e660751059#diff-a80a847ddff0dbe95ff3bc30c9390f7dR70
Is this correct?

sonneveld:
Oh sorry, I thought I was looking at the file in github, I must have been looking at the original.

Engine/ac/global_audio.cpp
```diff
@@ -354,24 +370,33 @@ void PlayMP3File (const char *filename) {
     int useChan = prepare_for_new_music ();
     bool doLoop = (play.music_repeat > 0);
 
-    if ((channels[useChan] = my_load_static_ogg(asset_name, 150, doLoop)) != NULL) {
-        channels[useChan]->play();
+    SOUNDCLIP *clip = my_load_static_ogg(asset_name, 150, doLoop);
```
sonneveld: Nice. This was weird before as we'd overwrite the channel with null on failure.

Engine/ac/global_audio.cpp
```diff
@@ -386,21 +411,26 @@ void PlaySilentMIDI (int mnum) {
     play.silent_midi = mnum;
     play.silent_midi_channel = SCHAN_SPEECH;
     stop_and_destroy_channel(play.silent_midi_channel);
-    channels[play.silent_midi_channel] = load_sound_clip_from_old_style_number(true, mnum, false);
-    if (channels[play.silent_midi_channel] == NULL)
+
+    AudioChannelsLock lock;
+    auto* ch = lock.SetChannel(play.silent_midi_channel,load_sound_clip_from_old_style_number(true, mnum, false));
```
sonneveld: Should you do the same thing here? Not overwrite the channel on failure? I suppose we quit() on failure this time.

Engine/game/savegame.cpp
```diff
@@ -555,6 +555,9 @@ HSaveError DoAfterRestore(const PreservedParams &pp, const RestoredData &r_data)
     const int cf_out_chan = play.crossfading_out_channel;
     play.crossfading_in_channel = 0;
     play.crossfading_out_channel = 0;
+    
+    {
+    AudioChannelsLock lock;
     // NOTE: channels are array of MAX_SOUND_CHANNELS+1 size
     for (int i = 0; i <= MAX_SOUND_CHANNELS; ++i)
```
sonneveld: Potential for setting AudioChannelsLock within the lock.
edit: you might be doing a more global lock here because you want a consistent save state.

Engine/main/update.cpp
```diff
@@ -248,14 +248,16 @@ void update_overlay_timers()
 void update_speech_and_messages()
 {
   // we need to know if there is/was voice-over
-  const bool is_voice = channel_has_clip(SCHAN_SPEECH);
+  AudioChannelsLock lock;
+  auto *ch = lock.GetChannel(SCHAN_SPEECH);
+  const bool is_voice = ch != nullptr;
```
sonneveld: `is_voice` here could be the value of `is_playing`, then we won't need the lock for the rest of the function.

Engine/main/update.cpp
```diff
@@ -291,7 +293,9 @@ void update_speech_and_messages()
 void update_sierra_speech()
 {
   // we need to know if there is/was voice-over
-  const bool is_voice = channel_has_clip(SCHAN_SPEECH);
+  AudioChannelsLock lock;
+  auto *ch = lock.GetChannel(SCHAN_SPEECH);
+  const bool is_voice = ch != nullptr;
```
sonneveld: Similarly here, you could grab the speech pos here and free the lock.

Engine/media/audio/ambientsound.cpp
```diff
@@ -12,9 +12,7 @@
 //
 //=============================================================================
 
-#include "media/audio/ambientsound.h"
```
sonneveld: I tried to keep `audio_system.h` use out of `media/audio/*`

Engine/media/audio/audio.cpp
```diff
@@ -39,8 +39,53 @@
 #include "main/game_run.h"
 
 using namespace AGS::Common;
+using namespace AGS::Engine;
+
+//-----------------------
+//sound channel management; all access goes through here, which can't be done without a lock
+
+static std::array<SOUNDCLIP *,MAX_SOUND_CHANNELS+1> _channels;
```
sonneveld: I don't know if this gets null initialised or not for std::array.

Engine/media/audio/audio.cpp
```diff
@@ -39,8 +39,53 @@
...
+
+SOUNDCLIP *AudioChannelsLock::GetChannelIfPlaying(int index)
+{
+    auto *ch = _channels[index];
+    return (ch != nullptr && ch->is_playing()) ? ch : nullptr;
+}
+
+SOUNDCLIP *AudioChannelsLock::SetChannel(int index, SOUNDCLIP* ch)
+{
+    _channels[index] = ch;
```
sonneveld: Is it worth printing a debug warning here if we're overwriting a channel pointer? The channel would not get cleaned up otherwise. 

Another check might be to ensure we're not reusing a channel pointer, which I _think_ we're careful about but we'd have to check the cross fading code (that moves soundclips in/out of the music channel)

I'm fairly sure we're careful about freeing the channel first but it would be a good sanity check.

Engine/media/audio/audio.cpp
```diff
@@ -91,15 +136,18 @@ void start_fading_in_new_track_if_applicable(int fadeInChannel, ScriptAudioClip
...
-void move_track_to_crossfade_channel(int currentChannel, int crossfadeSpeed, int fadeInChannel, ScriptAudioClip *newSound)
+static void move_track_to_crossfade_channel(int currentChannel, int crossfadeSpeed, int fadeInChannel, ScriptAudioClip *newSound)
 {
+    AudioChannelsLock lock;
+
     stop_and_destroy_channel(SPECIAL_CROSSFADE_CHANNEL);
-    channels[SPECIAL_CROSSFADE_CHANNEL] = channels[currentChannel];
-    channels[currentChannel] = NULL;
+
+    auto cfade_clip = lock.SetChannel(SPECIAL_CROSSFADE_CHANNEL, lock.GetChannel(currentChannel));
```
sonneveld: Maybe we need a MoveChannel() function to ensure no duplicate sound clips?

Engine/media/audio/audio.cpp
```diff
@@ -541,22 +602,26 @@ int get_volume_adjusted_for_distance(int volume, int sndX, int sndY, int sndMaxD
 
 void update_directional_sound_vol()
 {
-    for (int chan = 1; chan < MAX_SOUND_CHANNELS; chan++) 
+    AudioChannelsLock lock;
```
sonneveld: Potential for AudioChannelsLock within loop.

Engine/media/audio/audio.cpp
```diff
@@ -541,22 +602,26 @@ int get_volume_adjusted_for_distance(int volume, int sndX, int sndY, int sndMaxD
...
         }
     }
 }
 
-void update_ambient_sound_vol () {
+void update_ambient_sound_vol ()
+{
+    AudioChannelsLock lock;
 
     for (int chan = 1; chan < MAX_SOUND_CHANNELS; chan++) {
```
sonneveld: Potential for AudioChannelsLock within loop.

Engine/media/audio/audio.cpp
```diff
@@ -592,9 +657,9 @@ void update_ambient_sound_vol () {
             wantvol = get_volume_adjusted_for_distance(ambientvol, thisSound->x, thisSound->y, thisSound->maxdist);
         }
 
-        if (channel_is_playing(thisSound->channel)) {
-            channels[thisSound->channel]->set_volume(wantvol);
-        }
+        auto *ch = lock.GetChannel(thisSound->channel);
```
sonneveld: `GetChannelIsPlaying` here?

Engine/media/audio/audio.cpp
```diff
@@ -634,23 +699,26 @@ void shutdown_sound()
...
 // the sound will only be played if there is a free channel or
 // it has a priority >= an existing sound to override
-int play_sound_priority (int val1, int priority) {
+static int play_sound_priority (int val1, int priority) {
     int lowest_pri = 9999, lowest_pri_id = -1;
 
+    AudioChannelsLock lock;
+
     // find a free channel to play it on
     for (int i = SCHAN_NORMAL; i < MAX_SOUND_CHANNELS; i++) {
```
sonneveld: Potential for AudioChannelsLock within loop? We interact with another channel but it's later.

Engine/media/audio/audio.cpp
```diff
@@ -795,15 +867,21 @@ extern volatile char want_exit;
...
+    {
+        auto* ch = lock.GetChannelIfPlaying(i);
+        if (ch)
+            ch->poll();
+    }
 }
 
+//this is called at various points to give streaming logic a chance to update
+//it seems those calls have been littered around and points where it ameliorated skipping
+//a better solution would be to forcibly thread the streaming logic
```
sonneveld: Hah yes, we know.

Engine/media/audio/audio.cpp
```diff
@@ -918,41 +998,46 @@ void update_music_volume() {
                 newvol = targetVol;
                 stop_and_destroy_channel_ex(SCHAN_MUSIC, false);
                 if (crossFading > 0) {
-                    channels[SCHAN_MUSIC] = channels[crossFading];
-                    channels[crossFading] = nullptr;
+                    lock.SetChannel(SCHAN_MUSIC, lock.GetChannel(crossFading));
```
sonneveld: Another potential spot for a MoveChannel function.

Engine/media/audio/audio.cpp
```diff
@@ -963,8 +1048,8 @@ int prepare_for_new_music () {
             if (crossFading > 0) {
                 // It's still crossfading to the previous track
                 stop_and_destroy_channel_ex(SCHAN_MUSIC, false);
-                channels[SCHAN_MUSIC] = channels[crossFading];
-                channels[crossFading] = NULL;
+                lock.SetChannel(SCHAN_MUSIC, lock.GetChannel(crossFading));
```
sonneveld: Another spot for a potential channel move function.

Engine/media/audio/audio.cpp
```diff
@@ -1049,36 +1134,33 @@ void play_new_music(int mnum, SOUNDCLIP *music) {
         return;
     }
 
-    useChannel = prepare_for_new_music ();
-
-    play.cur_music_number=mnum;
+    useChannel = prepare_for_new_music();
+    play.cur_music_number = mnum;
     current_music_type = 0;
-    channels[useChannel] = NULL;
```
sonneveld: Are we assuming prepare_for_new_music sets this to null? should we assert?

mgambrell:
Sorry, I don't really have time to evaluate this. But in general I tried not to think about things too much. And I can tell you, I probably put the locks outside the loops because I didn't want to analyze what happens if the loops get interrupted (probably nothing) but also because sound APIs should be atomic from the user's point of view. If someone says "fade out all the channels" he definitely doesn't mean "Fade out all the channels, but it's OK if some of them start fading a little bit after the others because random timing flukes". It's worth evaluating for loops that may take a long time, but not stuff that's effectively instantaneous anyway. Actually it's best if the entire frame's worth of audio commands are atomic too, but we're not gonna have that any time soon.

ivan-mogilko:
@mgambrell 
> I probably put the locks outside the loops because I didn't want to analyze what happens if the loops get interrupted (probably nothing) but also because sound APIs should be atomic from the user's point of view.

Yes, that makes sense. I am thinking to redo some of my changes related to loops.

sonneveld:
> And I can tell you, I probably put the locks outside the loops because I didn't want to analyze what happens if the loops get interrupted (probably nothing) but also because sound APIs should be atomic from the user's point of view.

Agree about the atomic sound api but as per our discussion a while ago, maybe that's worth considering when we can implement it that way.

I also just realised that it won't help polling that much, the polling loop will just block at the first channel that's locked and will wait before doing the rest anyway.

Engine/media/audio/audio.cpp
```diff
@@ -724,6 +791,8 @@ void clear_music_cache() {
 
 }
 
+void play_new_music(int mnum, SOUNDCLIP *music);
```
sonneveld: Defect: this is actually defined as static below and will fail to compile on some compilers.

Engine/media/audio/audio.cpp
```diff
@@ -634,23 +699,26 @@ void shutdown_sound()
...
+        auto* ch = lock.GetChannel(i);
         if (val1 < 0) {
             // Playing sound -1 means iterate through and stop all sound
             if (channel_is_playing(i))
                 stop_and_destroy_channel (i);
         }
         else if (!channel_is_playing(i)) {
             if (PlaySoundEx(val1, i) >= 0)
-                channels[i]->priority = priority;
+                ch->priority = priority;
```
sonneveld: defect: This code will segfault because `ch` is null but it is expecting channels[i] to have a value.  There may be other examples to watch out for.

Engine/ac/game.cpp
```diff
@@ -1752,14 +1753,19 @@ void stop_fast_forwarding() {
     if (play.end_cutscene_music >= 0)
         newmusic(play.end_cutscene_music);
 
+    {
+    AudioChannelsLock lock;
+
     // Restore actual volume of sounds
     for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
```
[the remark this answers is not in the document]
ivan-mogilko: I have no idea, this code never changed since 2012. This needs separate investigation.

Engine/ac/global_audio.cpp
```diff
@@ -256,18 +256,28 @@ void scr_StopMusic() {
...
     }
 }
 void SeekMP3PosMillis (int posn) {
     if (current_music_type) {
-        debug_script_log("Seek MP3/OGG to %d ms", posn);
-        if (crossFading > 0 && channel_is_playing(crossFading))
-            channels[crossFading]->seek (posn);
-        else if (channel_is_playing(SCHAN_MUSIC))
-            channels[SCHAN_MUSIC]->seek (posn);
+        AudioChannelsLock lock;
```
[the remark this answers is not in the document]
ivan-mogilko: I think the logic remained the same:
* if crossFading > 0 then test crossFading channel first;
* if previous check did not succeed then test music channel.

I may simplify the code to reduce its size though.

Engine/ac/global_audio.cpp
ivan-mogilko: It still overwrites channel with null on failure, in the last else below, just explicitly.

Engine/ac/global_audio.cpp
ivan-mogilko: It seems like channels were usually overwritten with null on failure. I don't know if this is requirement so tried to keep same logic.

As for quitting, this is very inconsistent in AGS. I recieved requests to replace them with warnings to the log on numerous occasions. IMHO this should be configurable, in other words an "assert"-like function that either prints into log or quits depending on setup. Then we may send a message to the IDE if it's connected in debug session too, and even break at this point.

Engine/main/update.cpp
ivan-mogilko: I am not sure, it does different actions depending on whether clip exists and whether it is playing below.
But I may read necessary values and unlock at once.

Engine/main/update.cpp
ivan-mogilko: Right, so I thought about that, and my next pr with nonblocking voice-over adds different way to test for voice mode, so some of these checks become unnecessary and the place when you need to access channel also changes, so I think I'll leave this as it is for now (some code is going to be removed anyway)

Engine/game/savegame.cpp
ivan-mogilko: This is DoAfterRestore function where save data is already loaded and now applied. I am not sure if it's a good idea to make lock more global here because this function does many kinds of stuff, including *loading another room*.

Engine/media/audio/audio.cpp
ivan-mogilko: If I understood correctly from the answer: https://stackoverflow.com/questions/18295302/default-initialization-of-stdarray

We need to add empty initializer list `{}` to initialize all members to default, which is nullptr supposedly in this case.

Engine/media/audio/audio.cpp
ivan-mogilko: I can add warnings, but this situation is complicated, because clips could be saved elsewhere and not deleted.
I do not see any good solution here except for storing SOUNDCLIP in a smart pointer (either unique_ptr or shared_ptr, depending on what kind of manipulations we want to allow).

Engine/media/audio/audio.cpp
ivan-mogilko: Yes, prepare_for_new_music has this in the end:

<pre>
// Just make sure, because it will be overwritten in a sec
    if (_lock.GetChannel(useChannel) != nullptr)
        stop_and_destroy_channel (useChannel);
</pre>

Engine/media/audio/audio.cpp
mgambrell: that topic doesnt address your question, which is affected by the fact that statics are zero-initialized. in this case std::array is probably built with a standard array internally allocated not-on-the-heap, and the standard array is zero-initialized due to the whole thing being static. 

As for whether you ought to initialize static variables, that's a matter of elegance and what type of aesthetics you like and what kind of confidence you want to instill on the reader.  To me, leaving off the initializer means 'danger! must think about what it means that there was no initializer was here'. In this case, the reader is supposed to know that statics are zero-initialized, but I can see why it might be doubtful in this case.

A rule where arrays should always be initialized would hamper performance. And if there's any case where variables can be confidently not initialized, it's global statics.

Engine/media/audio/audio.cpp
ivan-mogilko: And this is why in my old refactored branch I was returning locked *channels* instead of clips!

Engine/media/audio/audio.cpp
ivan-mogilko: Alright, got it.

Engine/ac/game.cpp
```diff
@@ -269,18 +269,19 @@ void Game_SetAudioTypeVolume(int audioType, int volume, int changeType)
...
-        for (aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
+        AudioChannelsLock lock;
+        for (int aa = 0; aa < MAX_SOUND_CHANNELS; aa++)
         {
             ScriptAudioClip *clip = AudioChannel_GetPlayingClip(&scrAudioChannel[aa]);
             if ((clip != NULL) && (clip->type == audioType))
             {
-                channels[aa]->set_volume_percent(volume);
+                auto* ch = lock.GetChannel(aa);
+                ch->set_volume_percent(volume);
```
sonneveld: Needs to check if ch is null

Engine/media/audio/audio.cpp
```diff
@@ -918,41 +998,46 @@ void update_music_volume() {
...
-                    channels[crossFading] = nullptr;
+                    lock.SetChannel(SCHAN_MUSIC, lock.GetChannel(crossFading));
+                    lock.SetChannel(crossFading, nullptr);
                 }
                 crossFading = 0;
             }
             else {
                 if (crossFading > 0)
-                    channels[crossFading]->set_volume((curvol > targetVol) ? targetVol : curvol);
+                    lock.GetChannel(crossFading)->set_volume((curvol > targetVol) ? targetVol : curvol);
```
sonneveld: This grabs the channel without checking for null.   I only just noticed a couple of instance of these.. haven't thoroughly checked for others.

ivan-mogilko: Here, and in other places, engine is relying on other variables like `crossFading`.

Engine/ac/game.cpp
ivan-mogilko: It's relying on returned information about clip playing on this channel. All this is happening under one lock, so probably safe (at least I would think so).

ivan-mogilko: It's relying on scrAudioChannel's information about clip playing on it. There might be more places like this, where engine does not consider that the state may be inconsistent.

mgambrell: Oh yeah this is concerning. The existence of scrAudioChannel seems opposed to the safetiness of using my channel locking approach. It's like, more state, for each channel, stored separately. This really needs to be redesigned so it's kept together under control of the audio channel. could all the scrAudioChannel state be stored attached to the channel?

Engine/ac/viewframe.cpp
```diff
@@ -145,7 +145,11 @@ void CheckViewFrame (int view, int loop, int frame, int sound_volume) {
         }
     }
     if (sound_volume != SCR_NO_VALUE && channel != NULL)
-        channels[channel->id]->set_volume_percent(channels[channel->id]->get_volume() * sound_volume / 100);
+    {
+        AudioChannelsLock lock;
+        auto* ch = lock.GetChannel(channel->id);
+        ch->set_volume_percent(ch->get_volume() * sound_volume / 100);
```
sonneveld: Worth checking if ch is null here since play_audio_clip_by_index can return null.

ivan-mogilko: In that case `if (sound_volume != SCR_NO_VALUE && channel != NULL)` condition would be false

Engine/ac/global_audio.cpp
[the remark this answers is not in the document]
sonneveld: I do agree that the logic is the same, it's just not as obvious.  I tried something like this:

```
void SeekMP3PosMillis (int posn) {
    if (!current_music_type) { return; }

    AudioChannelsLock lock;

    auto *mus_ch = lock.GetChannel(SCHAN_MUSIC);
    auto *cf_ch = (crossFading > 0) ? lock.GetChannel(crossFading) : nullptr;

    if (cf_ch) {
        cf_ch->seek(posn);
    } else if (mus_ch) {
        mus_ch->seek(posn);
    }
}
```

Engine/ac/game.cpp
ivan-mogilko: @mgambrell this is about something I did in my old attempt few years ago, where I tried to gather all channel-related data under one protected struct. I've mentioned it a bit ago in previous threads:
https://github.com/ivan-mogilko/ags-refactoring/commits/refactor_audio_channel

This was somewhat similar to your approach, except I was not returning clips, but *channels*, which was safer imho because clip may be replaced in the middle of function. For convenience IIRC I overloaded -> operator to access clip through this channel struct.

When these pull requests with threading audio was started to open I suggested to restore that branch, because it already contained lots of work, but then let myself get persuaded to not push too much refactoring now.
I keep getting very dissapointed by myself because of this.

Engine/ac/viewframe.cpp
sonneveld: I mean, that's technically correct, but it's a slightly round about way of checking it. Perhaps an assert check for the benefit of future readers?

Engine/ac/game.cpp
ivan-mogilko: The more issues is found in this pr, the more I think that maybe I should restore my old code instead, finish it and add mutexes, using your commit for reference.

Engine/media/audio/audio.cpp
sonneveld: Yes but the crossFading variable is not protected by the channel locks.

Engine/ac/game.cpp
sonneveld: I mean, this PR is mostly ready. I can help add your channel refactoring after this one if you like.

ivan-mogilko:
I've mentioned this briefly in a code comment above, but want to state this more clearly.

Seeing all the issues found with this PR, for example related to potential inconsistency between different objects, such as ScriptAudioChannel and AudioChannel, I begin to think that I should have actually restored my old work from this branch and then ammend it with mutex, using new PRs for the reference:
https://github.com/ivan-mogilko/ags-refactoring/commits/refactor_audio_channel

I haven't check that code thoroughly for some time, but IIRC one of the advantages that it has is that it merged ScriptAudioChannel, AmbientSound and AudioChannel functionality in one object, reducing possible inconsistencies.

Another thing, it actually protected SOUNDCLIP* with smart pointer (we did not use C++11 back then so I wrote a quick emulation of unique_ptr).

Of course it is not necessarily matching the locks logic, but may be adjusted for it.

Engine/ac/game.cpp
sonneveld: Another point is that unless the AudioChannel is an array of staticly allocated objects, you'll still need to lock the channels array when removing/adding channel pointers.

Engine/media/audio/audio.cpp
ivan-mogilko: > Yes but the crossFading variable is not protected by the channel locks.

This is true...

ivan-mogilko:
Or, we may keep it the way it is, except adding more null pointer tests as @sonneveld pointed out above, because we can no longer rely on additional variables and ScriptAudioChannels, which are not protected by locks....

Engine/ac/game.cpp
ivan-mogilko: > Another point is that unless the AudioChannel is an array of staticly allocated objects, you'll still need to lock the channels array when removing/adding channel pointers.

AudioChannel actually is the array of statically allocated objects, they are created at engine startup and never changed.

ivan-mogilko: > I mean, this PR is mostly ready. I can help add your channel refactoring after this one if you like.

Alright, question is: will additional null-pointer tests you suggested be enough to keep things at least thread safe?

sonneveld: I believe so. You might be interested in this commit https://github.com/sonneveld/ags/commit/b2a6fe059cda5dd8be2526766a59f01a96aac8b7 I made some adjustments as I was reviewing this PR.

ivan-mogilko:
Sorry for delay.

I followed the above suggestions, except the suggestion to move locks *inside* loops. On contrary I moved some of the locks I previously missed outside the loop to make whole loop an "atomic" operation (I think these loops usually do some fast processing).

* made sure that clip pointer is always tested after getting it from GetChannel();
* Added MoveChannel function which moves clip pointer from one channel to another (reduces code and maybe safer).
* followed suggestions from https://github.com/sonneveld/ags/commit/b2a6fe059cda5dd8be2526766a59f01a96aac8b7 (more or less)

sonneveld:
> I followed the above suggestions, except the suggestion to move locks _inside_ loops.

I diffed the old version of your branch with the new forced pushed one. All changes look good. It's interesting comparing our different style preferences :)

ivan-mogilko:
Is there anything I should add to this pr, like, fixed locks for Linux? And do we actually need locks in places like this:
https://github.com/adventuregamestudio/ags/pull/689/files#diff-a1c77e7c9d4359583ff8dde23b75e810R32

sonneveld:
> Is there anything I should add to this pr, like, fixed locks for Linux? And do we actually need locks in places like this: ..

I rebased my PR to be on top of this one. It includes the recursive locks and fixes the soundclip behaviour around destroy. https://github.com/adventuregamestudio/ags/pull/688

PR #690, opened 2019-03-25
embed freetype 2.1.3 sources and make linux build use it

rofl0r:
fixes various font issues with lucasFan font.

note that the workaround https://github.com/adventuregamestudio/ags/blob/master/Common/font/ttffontrenderer.cpp#L102 is not removed yet as the freetype sources are not yet built for non-linux systems. adding support for mac etc should be as easy as including the new makefile include file though, but i can't test it.


sonneveld:
Why are we embedding the source now? Can't you just build from the lib-alfont repo?

rofl0r:
1) the build sys in the libalfnt repo is crippled and unusable
2) it's hard to get it built after those issues are fixed by using the original tarball's makefiles
3) it's even harder to get the built .a archive and headers used from the temporary build dir without installing it on the host sys, which would override the system's freetype
4) it must all be done manually, which is intolerable
5) all other required 3rd party sources are in this repo too, including libalfont

sonneveld:
Freetype is quite a big library though.  We do embedd alfont but not its dependencies.

I've played with rewriting the build system of alfont to use cmake if you want to give that a try: https://github.com/sonneveld/lib-alfont/tree/alfont-1.9.1-ags-sdl2 (currently includes SDL2 but you should be able to comment that out) 

The advantages are it would allow us to use the same project for all 3 systems, and you get a proper make file which will let you set a prefix to `~/src/ags/Engine/opt` or something.

rofl0r:
i'm not at all a fan of cmake: https://github.com/libtom/libtomcrypt/issues/391#issuecomment-390434739
and again, alfont is already embedded and built as part of the engine. no need to introduce yet another build system.

freetype's not that big, the tarball is 1MB, and only 29 .c files are built. i could shrink it a bit though by removing the unused Jamfiles and original build system, as well as the documentation.

also, basically ivan asked for it in the referenced issue.

> The advantages are it would allow us to use the same project for all 3 systems

the same is true for my PR. just that it doesn't require a) using an external repo and b) using yet another build system, in other words pulling in yet another dependency.

sonneveld:
My AGS SDL2 port replaces the different Engine build projects (for windows, mac and linux) with a single cmake project. It makes things so much easier from a cross-platform development perspective. I hope to make it standard.

If anything, we should probably remove alfont.c from the ags repo and just use a git submodule to our forked alfont repo.

rofl0r:
> replaces ... with a single cmake project

my impression that you are a hipster grows day by day. you find certain techs like cmake, `auto` and other new C++ features, glad, etc "sexy" or "hip" or whatever, then use your PRs as a backdoor to mix them into this project. you change stuff just for the sake of changing it, which often introduces new issues and dependencies. i find this highly frustrating. why not focus on real issues instead ?

ivan-mogilko:
Ignoring cmake discussion for now...

> also, basically ivan asked for it in the referenced issue.

>> The advantages are it would allow us to use the same project for all 3 systems

Well, my thinking was also that it could be added as a submodule from a our forked repository.

rofl0r:
i had to work with a git submodule recently and it was quite a [pain in the ass](https://github.com/kometbomb/klystrack/pull/263#issuecomment-465337057).
git submodules have their legit usecases, but only when it targets highly volatile codebases.
the alfont repo practically never changes, and neither does the freetype 2.1.3 code.
having to have the hassle to constantly update submodules and checking out 2 different module states inside a single repo to avoid checking in a single set of never-changing sources is hardly justified.
i would recommend anyone in favor of submodules to first try it out with an example project.
it really is annyoing as hell.

rofl0r:
i've traced the files/directory that are needed for the build https://gist.github.com/rofl0r/030b5e90c449930529e2f0d4444ecdc3 , total only 286 (of which 20 are dirs). so about 240, i.e. half of the files could be removed. the files that need to be checked in are totalling 3601964 bytes. that's roughly equivalent the size of the iOS, or the Editor/scintilla dir. PSP dir is even 50% bigger than that.

ivan-mogilko:
Hmm, we would need to also switch Windows version to build from same sources. Not sure about other ports, though, they are getting alfont from patch files iirc (maybe wrong).

rofl0r:
> need to also switch Windows version to build from same sources

right, that would probably simplify the current build differences between the platforms a lot. and it should be equally simple since i've circumvented the build system of freetype itself. where is the windows build procedure documented?

ericoporto:
just a small thing on the submodules discussion. This repo already uses submodules, for googletests and the luaplugin.

rofl0r:
> This repo already uses submodules, for googletests and the luaplugin

interesting, i never had to check them out since apparently they're not part of the Engine build, which i'm interested in. but once a part of the engine source (like freetype) is a submodule, everybody will be forced to deal with it (and its annoyances)

sonneveld:
> interesting, i never had to check them out since apparently they're not part of the Engine build, which i'm interested in. but once a part of the engine source (like freetype) is a submodule, everybody will be forced to deal with it (and its annoyances)

Works for me! If you have the submodule fixed at a particular changeset, it should give you the freedom to update the module without breaking the repos that include it. Just means a two step process if you need to fix something in both repos. Since it's a library, it should still build standalone.

rofl0r:
> Works for me

yes, for me too, because googletest isn't used by the engine build. the directory is empty though. in order to get its contents, i would have to issue a command that populates it

> If you have the submodule fixed at a particular changeset

afaik a submodule needs to always be fixed at a particular changeset

> Since it's a library, it should still build standalone

yeah. however in the case of freetype 2.1.3 we don't need it standalone and always only as part of the engine build, and using my makefile snippet it's trivial to build it not just for linux, but also mac, windows, etc and fixes the font issues once and for all (besides it's statically linked, so it fixes other issues like #693 as a side effect).

ivan-mogilko:
> having to have the hassle to constantly update submodules and checking out 2 different module states inside a single repo to avoid checking in a single set of never-changing sources is hardly justified.
i would recommend anyone in favor of submodules to first try it out with an example project.
it really is annyoing as hell.

I worked with sources that had submodules before, maybe I did not do some operations that you had to do, but I frankly don't remember so much hassle with it *if it does not change*. IIRC main issue with having its source as submodule was that you have to call `init submodules` command when you clone repository, and possibly when you switch back in time to some ancient branch and back.
I agree that this may be annoying if you have to manually do this often, but is this going to be a thing here?

Since submodule is just a single string in a text file, maybe we could first use it as a submodule, and if that proves frustrating go back with your plan?

rofl0r:
> Since submodule is just a single string in a text file, maybe we could first use it as a submodule, and if that proves frustrating go back with your plan?

well, it would force anyone building the engine to pull the submodule in.

so in the end they would need the space on disk anyhow.

as all other 3rd party sources required for the build are copied into the source tree, i don't see what makes this one here special, apart from the fact that we currently have a broken (makefiles are invalid) version as part of the lib-alfont repo, which in turn has 2 things in it, with the second part of it already copied into this repo.

additionally, as i said, i could remove half of the files from the sources, making it a lot slimmer than the version already checked into lib-alfont. so a user would need 4 MB less of checked out files and 4MB less in the .git subfolder. that's not possible with a git submodule, since even if we retroactively delete the unneeded files there, they're still part of .git, unless you force push a new history.

note that i plan to file a PR to lib-alfont repo fixing the broken makefiles regardless of what happens with this one.

ivan-mogilko:
Ah, right, also it will introduce submodule in history at some point too.

rofl0r:
> Ah, right, also it will introduce submodule in history at some point anyway.

ok. shall i go ahead and remove the unneeded files, @ivan-mogilko ?

sonneveld:
Wait so we're not using sub modules then?

rofl0r:
> Wait so we're not using sub modules then?

right, because it offers zero advantages in this situation

rofl0r:
> If we don't use submodules, we should just remove that lib-alfont repo altogether. Or at the very least, the ags specific branch (since the ags repo is effectively that)

we could just stick a disclaimer there that the repo is only for historical purposes. as it stands, it doesn't even have the latest source code in it (ags Engine here uses 2.0.9, whereas the lib-alfont repo has 1.9.1 as default branch, and 2.0.7 as highest available version - issue filed). so yeah, apart from archival purposes it's really not especially useful. also the multi-branch approach is kinda confusing. master could simple have the 1.9.x, followed by the various 2.0.x versions as follow up commits. each of them could use a tag to checkout a specific version if required.

ivan-mogilko:
Alright, so what we do is this: we put **full** alfont/freetype lib sources into Engine/libsrc.
Then, as a separate step switch Windows to use it as well (and other ports that need it).

I'd like to find time soon to try the latter out first just in case to see any possible problems (hopefully there will be none but better test first).

sonneveld:
> Then, as a separate step switch Windows to use it as well (then check other ports).

Yeah I realised that if we delete files it might potentially break builds on other platforms.

rofl0r:
ok. this pr currently has the full source of freetype included. you can easily compare the files of this pr against http://ftp.math.utah.edu/pub/freetype/freetype-2.1.3.tar.gz by using `diff -ru dir1 dir2` to see nothing expect 2 makefiles are different (due to the patches in followup commits). the source of the tarball in turn can be checked against lib-alfont repo by doing `diff -ruw dir1 dir2`, which does skip whitespace differences (the version in libalfont repo has changed unix lineendings `\n` to `\r\n` in some files)

sonneveld:
> the source of the tarball in turn can be checked against lib-alfont repo

Does it have everything from the `alfont-1.9.1-agspatch` branch? I noticed the build/VS* projects weren't there.

rofl0r:
> Does it have everything from the alfont-1.9.1-agspatch branch? I noticed the build/VS* projects weren't there.

those are in the top-level directory, not in the freetype directory.
diffing the freetype directories, this is what i get (all of those files are created during build, i.e. temporary)
```
$ diff -ruw ../lib-alfont/freetype/ Engine/libsrc/freetype-2.1.3 #../lib-alfont/freetype/ | less
Only in Engine/libsrc/freetype-2.1.3: builds/unix/config.log
Only in Engine/libsrc/freetype-2.1.3: builds/unix/config.status
Only in Engine/libsrc/freetype-2.1.3: builds/unix/freetype-config
Only in Engine/libsrc/freetype-2.1.3: builds/unix/ftconfig.h
Only in Engine/libsrc/freetype-2.1.3: builds/unix/libtool
Only in Engine/libsrc/freetype-2.1.3: builds/unix/unix-cc.mk
Only in Engine/libsrc/freetype-2.1.3: builds/unix/unix-def.mk
Only in Engine/libsrc/freetype-2.1.3: config.mk
```

though by checking i noticed something interesting: i previously had branch 2.0.x of lib-alfont checked out, and its included freetype version was 2.1.10, not 2.1.3

so it's possible that 2.1.10 could work too, as we have alfont 2.0.9 (plus agspatch) in the Engine dir.

ivan-mogilko:
> Does it have everything from the alfont-1.9.1-agspatch branch? I noticed the build/VS* projects weren't there.

These are not needed if we link directly, they are solutions that I added for building alfont.lib on Windows.

ivan-mogilko:
Alright, sorry for all the delay. I quickly tested this on Windows, and can confirm it works if I also add alfont.c either from our patched 1.9.1 branch in separate repository, or 2.0.9 one from Engine/libsrc (used on Linux).

One note though, the alfont/freetype sources would likely need to be moved to Common lib (this could be done after the merge too) because Editor's AGS.Native.dll also uses alfont for fonts preview.

sonneveld:
Nice one. I just pushed my fixes to `ags3` to ensure it would build on mac.

PR #694, opened 2019-03-26
Create config.yml for circleci

ericoporto:
As simple as possible `config.yml` for building AGS Engine, on Debian, using [**circleci**](https://circleci.com/), which uses a [docker image](https://cloud.docker.com/repository/docker/ericoporto/ags-primary). Before accepting, this should be pretty similar to the current lost `.travis.yml` in the repo, but it's a different CI system so people can have options and decide which one to pursue. It also won't magically work without configuring things on Github or clicking Circle-CI in the Github Marketplace.

`Dockerfile` is below

```
FROM debian:stretch

# Make no effort to shrink this image as it is only for local development and
# we want to be able to exploit layer caching

RUN apt-get -y update
RUN apt-get -y --no-install-recommends install ssh tar gzip
RUN apt-get -y --no-install-recommends install curl ca-certificates
RUN apt-get -y --no-install-recommends install git
RUN apt-get -y --no-install-recommends install build-essential pkg-config 
RUN apt-get -y --no-install-recommends install debhelper fakeroot
RUN apt-get -y --no-install-recommends install libfreetype6-dev 
RUN apt-get -y --no-install-recommends install libogg-dev libvorbis-dev
RUN apt-get -y --no-install-recommends install libaldmb1-dev libtheora-dev 

RUN curl -O -L https://github.com/Kitware/CMake/releases/download/v3.13.4/cmake-3.13.4-Linux-x86_64.tar.gz
RUN tar -C /opt -xf cmake-3.13.4-Linux-x86_64.tar.gz
ENV PATH="/opt/cmake-3.13.4-Linux-x86_64/bin:${PATH}"

RUN mkdir lib-allegro \
&& cd lib-allegro/ \
&& git clone https://github.com/adventuregamestudio/lib-allegro.git \
&& cd lib-allegro \
&& git checkout allegro-4.4.2-agspatch \
&& sed -i '/INLINE int/c\static INLINE int' /lib-allegro/lib-allegro/addons/jpgalleg/src/decode.c \
&& mkdir build-liballegro \
&& cd build-liballegro \
&& cmake -D SHARED=off -DCMAKE_BUILD_TYPE=Debug .. \
&& make && make install && cd ../../../

CMD /bin/bash
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 1
 jobs:
   build:
     docker:
-      - image: ericoporto/ags-primary:0.0.1
+      - image: ericoporto/ags-primary:0.1.0
 
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,10 @@
+version: 1
+jobs:
+  build:
+    docker:
+      - image: ericoporto/ags-primary:0.0.1
+
+    steps:
+      - checkout
+
+      - run: fakeroot debian/rules binary
--- a/docker/debian/Dockerfile
+++ b/docker/debian/Dockerfile
@@ -1,0 +1,32 @@
+FROM debian:stretch
+
+# Make no effort to shrink this image as it is only for local development and
+# we want to be able to exploit layer caching
+
+RUN apt-get -y update
+RUN apt-get -y --no-install-recommends install ssh tar gzip
+RUN apt-get -y --no-install-recommends install curl ca-certificates
+RUN apt-get -y --no-install-recommends install git
+RUN apt-get -y --no-install-recommends install build-essential pkg-config 
+RUN apt-get -y --no-install-recommends install debhelper fakeroot
+RUN apt-get -y --no-install-recommends install libfreetype6-dev 
+RUN apt-get -y --no-install-recommends install libogg-dev libvorbis-dev
+RUN apt-get -y --no-install-recommends install libaldmb1-dev libtheora-dev 
+
+RUN curl -O -L https://github.com/Kitware/CMake/releases/download/v3.13.4/cmake-3.13.4-Linux-x86_64.tar.gz
+RUN tar -C /opt -xf cmake-3.13.4-Linux-x86_64.tar.gz
+ENV PATH="/opt/cmake-3.13.4-Linux-x86_64/bin:${PATH}"
+
+RUN mkdir lib-allegro \
+&& cd lib-allegro/ \
+&& curl -O -L https://github.com/liballeg/allegro5/releases/download/4.4.3.1/allegro-4.4.3.1.tar.gz \
+&& tar -C . -xf allegro-4.4.3.1.tar.gz \
+&& cd allegro-4.4.3.1 \
+&& mkdir build-liballegro \
+&& cd build-liballegro \
+&& cmake -D SHARED=off -D WANT_ALLEGROGL=off -D WANT_LOADPNG=off -D WANT_LOGG=off \
+         -D WANT_JPGALLEG=off -D WANT_EXAMPLES=off -D WANT_TOOLS=off \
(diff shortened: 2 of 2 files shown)


sonneveld:
Nice! I recommend using these options when running `cmake`, so you don't need to patch those addons (which we don't use as far as I know):

```
              ....
                -D SHARED=off \
                -D WANT_ALLEGROGL=off -D WANT_LOADPNG=off -D WANT_LOGG=off -D WANT_JPGALLEG=off -D WANT_EXAMPLES=off -D WANT_TOOLS=off -D WANT_TESTS=off \
                -D WANT_DOCS=off \
               ....

```

sonneveld:
Also cmake 3.14 is out if you want to update that.

rofl0r:
if you use allegro 4.4.3 you don't need to use cmake at all.

sonneveld:
> if you use allegro 4.4.3 you don't need to use cmake at all.

Sure, until we patch allegro again.

rofl0r:
> Sure, until we patch allegro again.

i don't think that will be necessary

sonneveld:
> i don't think that will be necessary

We will be building custom allegro libraries for later AGS versions (not 3.5 though). And as you say, the distro has to have 4.4.3. I believe stretch still has 4.4.2?

rofl0r:
> We will be building custom allegro libraries for later AGS versions (not 3.5 though).

really ? what features will be added to allegro ?

> I believe stretch still has 4.4.2?

i would think so too. debian is generally a pretty slowly rolling distro. arch, void and others should be on the bleeding edge though.

sonneveld:
> really ? what features will be added to allegro ?

SDL2 support! Most of the SDL usage will be from the AGS engine but we need to maintain a cut down allegro library for legacy software drawing routines.

ericoporto:
I can remake that docker image to get allegro 4.4.3 directly, does it has a tarball release somewhere @rofl0r ?

I can remove the version spec on the `config.yml` to allow updating the docker image without having to update the file in the ags repo. The image is only built once, and then it would be repeatedly gotten from dockerhub. I copied your header @sonneveld on the Dockerfile because I made no effort to optimize the image too. :thinking: 

I don't know if it helps [but allegro here](https://github.com/adventuregamestudio/lib-allegro) is already building itself too.

sonneveld:
> I can remake that docker image to get allegro 4.4.3 directly, does it has a tarball release somewhere

The allegro team publish via github downloads:  https://github.com/liballeg/allegro5/releases/tag/4.4.3.1

sonneveld:
> I can remove the version spec on the `config.yml` to allow updating the docker image without having to update the file in the ags repo

Does circleci support images at the `latest` tag? (as in, it pulls new versions occasionally) I don't think we need to support multiple versions of those images. 

Can you include the Dockerfile in the repo as well? I'm not sure where best to place it since we could have multiple dockerfiles to support different distributions.  `docker/ags-primary/Dockerfile` maybe?

ericoporto:
@sonneveld originally it was in `.circleci/images/primary/Dockerfile`. Looking around it appears circleci indeed supports `latest` tag. I didn't really know allegro was published through GH releases. Just to be sure, ags should build with Allegro 4.4.3.1 ?

sonneveld:
> Just to be sure, ags should build with Allegro 4.4.3.1 ?

That should be fine. The only change was to the tools, which we don't use.

ivan-mogilko:
Sorry, I was not following the discussion, is it okay to merge now?

ericoporto:
Well, it already works. If you can wait, my todo (having hard week, will add soon):

-[ ] add docker folder and Dockerfile in it.
-[ ] update docker image and docker file to use allegro 4.4.3.1

ericoporto:
@sonneveld is this naming scheme ok? A docker folder with os folder and Dockerfile in it? I was also on fence for os-intent folder like debian-agsengine or debian-engine. Also not sure if a comment on circleci config.yml saying the img used is based on that Dockerfile.

sonneveld:
I think that naming scheme looks fine. We can always rearrange later.

PR #695, opened 2019-03-26
Add Math_Random_Range for requesting a random number between a range.

sonneveld:
Since we no longer support recording/playback, assign a random value to `play.randseed` but don't use it for initialising random. Instead we could use it to identify new games?

### DIFF
--- a/Engine/ac/characterinfo_engine.cpp
+++ b/Engine/ac/characterinfo_engine.cpp
@@ -37,8 +37,6 @@ extern int char_speaking;
 extern RoomStruct thisroom;
 extern unsigned int loopcounter;
 
-#define Random __Rand
-
 int CharacterInfo::get_effective_y() {
     return y - z;
 }
@@ -404,7 +402,7 @@ void CharacterInfo::update_character_follower(int &aa, int &numSheep, int *follo
         }
       }
       // wait a bit, so we're not constantly walking
-      else if (Random(100) < (followinfo & 0x00ff)) ;
+      else if (Math_Random_Range(0, 100) < (followinfo & 0x00ff)) ;
       // the followed character has changed room
       else if ((room != game.chars[following].room)
             && (game.chars[following].on == 0))
@@ -451,12 +449,12 @@ void CharacterInfo::update_character_follower(int &aa, int &numSheep, int *follo
         (abs(game.chars[following].y - y) > distaway+30) |
         ((followinfo & 0x00ff) == 0)) {
         // in same room
-        int goxoffs=(Random(50)-25);
+        int goxoffs = Math_Random_Range(-25, 25);
         // make sure he's not standing on top of the other man
         if (goxoffs < 0) goxoffs-=distaway;
         else goxoffs+=distaway;
         walk_character(aa,game.chars[following].x + goxoffs,
-          game.chars[following].y + (Random(50)-25),0, true);
+          game.chars[following].y + Math_Random_Range(-25, 25), 0, true);
         doing_nothing = 0;
       }
     }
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
I don't understand what is the point of identifying the game with a random number? It will be quite diverse but not unique.

Assuming at some point we may want to let user supply random seed (for their script), will it be possible to adjust this system?

sonneveld:
> * I don't understand what is the point of identifying the game with a random number? It will be quite diverse but not unique.
> * Assuming at some point we may want to let user supply random seed (for their script), will it be possible to adjust this system?

I was worried about breaking games if I removed the `play.randseed` altogether, so I decided to fill it with a random value. I'm not sure it's useful to be able to supply seeds as user input introduces another random element we can't control.  (consider the recording/playback functionality that wasn't quite ready). However, if you look at `std::random_device seed_generator; ` this generates strong random numbers, useful for a seed. We can override that with a supplied number if we need to.

> Oh, and, what is the max value this distribution can return? Previous implementation had a limit of 32767 because it used rand(). This need to be noted because some scripts may change behavior (and manual has to be updated).

Oh I didn't realise. This implementation will generate random numbers up to int max.  Would it be better to leave the current behaviour of `Random` to be the same (silently limiting to 32k) and add a new script function for random ranges? It could be useful if you need a random number in the negative range.

ivan-mogilko:
> I was worried about breaking games if I removed the play.randseed altogether

It's not a part of either interface, so it should be safe to remove.

But I was thinking maybe someone would like to have e.g. a world generation based on seed like in Elite or Minecraft for example, then they make like to supply seed to the script VM's Random.
I think I saw someone creating their own random generator in script before, for exactly this reason (to be able to repeat same sequence).

> Oh I didn't realise. This implementation will generate random numbers up to int max. Would it be better to leave the current behaviour of Random to be the same (silently limiting to 32k) and add a new script function for random ranges? It could be useful if you need a random number in the negative range.

I think it's possible to make a switch depending on script API version, like I mentioned earlier. E.g. have a new function exported by default, and backwards-compatible for older games.

sonneveld:
> But I was thinking maybe someone would like to have e.g. a world generation based on seed like in Elite or Minecraft for example, then they have to supply seed to the script VM's Random.
> I think I saw someone creating their own random generator in script before, for exactly this reason (to be able to repeat same sequence).

I think that's a good idea but I wouldn't use the global game seed for that. There's no guarantee that something else would grab a random number while you're using it.  

I think that would be better if you could define your own random number generator, with a specific seed. But that sounds pretty specific and would require more scripting apis.

sonneveld:
> I think it's possible to make a switch depending on script API version, like I mentioned earlier. E.g. have a new function exported by default, and backwards-compatible for older games.

Ok. I'll see about those changes.

ivan-mogilko:
This pr needs a rebase.

What did we decide on changes?

IIRC **play.randseed** may be safely removed as it is not really used anywhere anymore, and not a part of script API. If we'd introduce the unique gameplay id, I guess better make a new variable which is also part of script API (and maybe generated in a more unique way), but that does not seem related to particular change.

Although not critical at all, but if it's possible to easily limit rand number generation to 32767 for backwards compatibility for old games maybe it should be done (in the past I've seen scripts where people tried to pass 1000000+ into script Random, so these script would produce different results now).
I think it's enough to compare script API level with v3507 (the highest number now).

ericoporto:
> I think I saw someone creating their own random generator in script before, for exactly this reason (to be able to repeat same sequence).

@ivan-mogilko yes, it's the Seeded Random module:
https://www.adventuregamestudio.co.uk/forums/index.php?topic=40952.0

sonneveld:
I'll leave this change for a future ags3 release.

PR #696, opened 2019-03-26
Script API: implemented PlayVoiceClip function

ivan-mogilko:
**UPDATE**.

This adds a method to play a voice-over clip in a non-blocking manner and without fixed connection to character state or any text displayed on screen. This might come very useful in scripting custom speech, because right now the only way to play a clip from speech.vox in a custom speech style is to call `Character.Say("&NNNN");`, which is a blocking command and makes it harder to handle any background activity.

The new script function is:
<pre>
/// Play speech voice-over in non-blocking mode, optionally apply music and sound volume reduction
import AudioChannel* PlayVoiceClip(Character*, int cue, bool as_speech = true);
</pre>

PlayVoiceClip plays a voice clip in non-blocking manner, clip is chosen by character's ID and cue ID and returns AudioChannel (atm this is always same channel reserved for speech in AGS).
*as_speech* parameter lets user choose whether they'd like to use default volume reduction rule or not (in the latter case they may apply their own). If this is enabled then voice playback lowers volume of all other sound types by the same rules as regular speech, and restores them back afterwards.

The returned AudioChannel pointer lets you stop the voice clip anytime.

**IMPORTANT:** PlayVoiceClip cannot be used during regular blocking voice speech and will fail (return null pointer) if called in this situation (e.g. from repeatedly_execute_always).

**SIDE EFFECTS:**
1) Acquiring speech's AudioChannel will let you stop even normal blocking speech by cutting of its voice playback but only if it was relying on one (that is - had started one in the first place). But I think you could do this prior anyway by getting System.AudioChannels[0].
2) I think it may be possible to run non-blocking voice-over during regular speech that does not have voice-over for some reason (not tested). However, speech will ignore non-blocking voice-over in this case (won't stop it if done displaying or gets skipped).

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2533,6 +2533,10 @@ builtin managed struct Game {
   /// Accesses the audio clips collection.
   readonly import static attribute AudioClip *AudioClips[];
 #endif
+#ifdef SCRIPT_API_v350
+  /// Play speech voice-over in non-blocking mode, optionally apply music and sound volume reduction
+  import static AudioChannel* PlayVoiceClip(Character*, int cue, bool as_speech = true);
+#endif
 };
 
 builtin struct GameState {
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -571,97 +571,6 @@ builtin managed struct Room {
 #endif
 };
 
-builtin managed struct Game {
-  /// Changes the active translation.
-  import static bool   ChangeTranslation(const string newTranslationFileName);
-  /// Returns true the first time this command is called with this token.
-  import static bool   DoOnceOnly(const string token);
-  /// Gets the AGS Colour Number for the specified RGB colour.
-  import static int    GetColorFromRGB(int red, int green, int blue);
-  /// Gets the number of frames in the specified view loop.
-  import static int    GetFrameCountForLoop(int view, int loop);
-  /// Gets the name of whatever is on the screen at (x,y)
-  import static String GetLocationName(int x, int y);
-  /// Gets the number of loops in the specified view.
-  import static int    GetLoopCountForView(int view);
-  /// Returns the current pattern/track number if the current music is MOD or XM.
-  import static int    GetMODPattern();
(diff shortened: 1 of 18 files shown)


sonneveld:
I'll be able to review properly this in the next day or so.

sonneveld:
>  I have doubts about StopVoiceClip, because it may not have much use as it is. 

Instead of returning an audio clip, could we return a Speech object of some kind? You could still have a property to get the audio clip if you need to adjust volume.

ivan-mogilko:
> Instead of returning an audio clip, could we return a Speech object of some kind? You could still have a property to get the audio clip if you need to adjust volume.

Having a speech object is a good concept and was discussed before (having Character.Say returning or initializing one), but it's not really related to this case, because PlayVoiceClip is intentionally separated from built-in speech.

ivan-mogilko:
One thing I forgotten is that new state flag has to be added to savegame.

ivan-mogilko:
I was thinking over and realized that was doing this slightly wrong.
Earlier I spoke about providing user with "basic blocks" of functionality, but function I did still combined two actions: 1) playing a clip from speech.vox and 2) playing a voice speech which also imposes some rules on game sound (lowers other sound volume).

IMHO ideally there would be a regular function to play any clip from any package without restrictions, and maybe separate function to schedule temporary volume mod for the time certain clip plays, but this may be more complicated to do now when there is no way to choose clip from voice pack, and there's no way to read "volume reduction when speech is playing" property from audio rules in script.

So for now I decided to keep what I have but with couple of adjustments:

1) First of all, I removed StopVoiceClip completely, because you can stop voice clip you've launched using AudioChannel pointer, like you do with rest of sounds, and also because that function would interrupt regular built-in voice speech too which may create inconsistency (we'd rather have Stop/SkipSpeech function).
2) Then I added "as_speech" boolean parameter to PlayVoiceClip that tells it whether to apply volume reduction rule or not. Or, more broadly, whether to consider this a speech voice or clip playback.
The clip is played on the speech channel anyway, which may seem too restrictive, but there's currently no good way in new audio API to select a channel and I did not want to introduce something original here.

PlayVoiceClip now cannot be run during regular blocking voice-over too.

In other words, PlayVoiceClip is meant to be used as a component of custom scripted speech, and lets user choose whether they'd like to use default volume reduction rule or not (in the latter case they may apply their own).

PR #698, opened 2019-03-31
OpenGL: Support screenshots when game bpp is different from opengl bpp of 32.

sonneveld:
Previously we were assuming the opengl driver would only be rendering 32bit, this fails when we use the mode's colour depth to determine the size of the buffer to write pixels to.

Support different colour depths for destination bitmap.

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -1088,12 +1088,46 @@ void OGLGraphicsDriver::ClearRectangle(int x1, int y1, int x2, int y2, RGB *colo
   // NOTE: this function is practically useless at the moment, because OGL redraws whole game frame each time
 }
 
+BITMAP *wrapGlReadPixelsBuffer(int width, int height, unsigned char *pixels) {
+  auto bmp = (BITMAP *)calloc(1, sizeof(BITMAP) + (sizeof(char *) * height));
+  bmp->w = width;
+  bmp->cr = width;
+  bmp->h = height;
+  bmp->cb = height;
+  bmp->clip = true;
+  bmp->cl = 0;
+  bmp->ct = 0;
+  bmp->id = 0;
+  bmp->extra = nullptr;
+  bmp->x_ofs = 0;
+  bmp->y_ofs = 0;
+  bmp->dat = nullptr;
+
+  // crete a fake bitmap to grab the vtable and whatno.
+  auto tmpbitmap = create_bitmap_ex(32, 1, 1);
+  bmp->vtable = tmpbitmap->vtable;
+  bmp->write_bank = tmpbitmap->write_bank;
+  bmp->read_bank = tmpbitmap->read_bank;
+  destroy_bitmap(tmpbitmap);
+
+  // glGetPixels returns image upside down, correct for this:
+  bmp->dat = pixels;
+  auto p = pixels;
+  for (int i=height-1; i>=0; i--) {
+    bmp->line[i] = p;
+    p += width * 4;
+  }
+
+  return bmp;
+}
+
+
 bool OGLGraphicsDriver::GetCopyOfScreenIntoBitmap(Bitmap *destination, bool at_native_res, Size *want_size)
 {
   (void)at_native_res; // TODO: support this at some point
 
   Size need_size = _do_render_to_texture ? _backRenderSize : _dstRect.GetSize();
-  if (destination->GetColorDepth() != _mode.ColorDepth || destination->GetSize() != need_size)
+  if (destination == nullptr || destination->GetSize() != need_size)
   {
     if (want_size)
       *want_size = need_size;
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Could you explain why this became necessary, are you adding display modes other than 32-bit?

ivan-mogilko:
Maybe pr description is confusing but I cannot understand the purpose of this change.
Does this fix a bug? or do you want to support other display modes than 32-bit on opengl?

ivan-mogilko:
The PR description confuses me. I understand the note of supporting different colour depths for bitmap, but what does this mean:

> Previously we were assuming the opengl driver would only be rendering 32bit, this fails when we use the mode's colour depth to determine the size of the buffer to write pixels to.

are you saying there's currently a defect?

sonneveld:
The other problem was that we only check that the destination is the right size, the caller gets the display bpp, which would still be 16bit. I suppose I could change the function to return size _and_ bpp.

The choice of wrapping the buffer means I can use a standard blit function, that will do any colour conversion for me.

sonneveld:
> are you saying there's currently a defect? or you are planning to support other modes?

When I first encountered the issue, I assumed it was because the opengl driver only supports 32bit but it looked like we only check if we're not 8 bit, which allows 16bit modes.

sonneveld:
> [Right now we create temporary buffer bitmap of 32-bits](https://github.com/adventuregamestudio/ags/blob/ags3/Engine/ac/draw.cpp#L373) where we read image from renderer, then blit it to destination of necessary format.

Ah I see now. Why not just return the fake bitmap from the opengl driver? Then we don't have to shuffle multiple buffers.  Is this function called often? I've only seen it used for screenshots in save games.

ivan-mogilko:
> The other problem was that we only check that the destination is the right size, the caller gets the display bpp, which would still be 16bit. I suppose I could change the function to return size and bpp. But I guess we're creating the illusion that the graphics driver is running in 16bit.

No, we were checking if destination and color depth were matching. We don't create any illusions. Display bpp cannot be 16-bit currently, because OpenGL and Direct3D run **only** in 32-bit as of now, and that was deliberate choice.

Perhaps this was not made clear enough, but GetCopyOfScreenIntoBitmap returns false in both cases: if size does not match or color depth does not match.

The reason I return wanted size at all, and did not use display mode's size similar as I relate on display mode's color depth right now, is that renderer may work in "render to texture" mode, in which case backbuffer is different. But we may make it return required color depth too, there's no problem in that. Perhaps I was simply lazy to do that as well.

> When I first encountered the issue, I assumed it was because the opengl driver only supports 32bit but it looked like we only check if we're not 8 bit, which allows 16bit modes.

OpenGL and Direct3D drivers are currently run only in 32-bit mode. What check are you referring to? that may be related to native bitmaps color depth.

> Why not just return the fake bitmap from the opengl driver? Then we don't have to shuffle multiple buffers.

But will that actually optimize much? there will be intermediate buffer anyway. Unless I misunderstand your meaning. Can you explain suggested solution in steps?

I think my intentions were to have as little work as possible inside renderer class implementation, because the process is similar for OpenGL and Direct3D (and partly for software renderer too), so I decided to move repeating code out to the external helper function.

In theory we could reuse same buffer bitmaps in that function.

sonneveld:
> OpenGL and Direct3D drivers are currently run **only** in 32-bit mode. What check are you referring to? that may be related to native bitmaps color depth.

Oh, this check: ```bool OGLGraphicsDriver::SetDisplayMode(const DisplayMode &mode, volatile int *loopTimer)
{
  ReleaseDisplayMode();

  if (mode.ColorDepth < 15)
  {
    set_allegro_error("OpenGL driver does not support 256-color display mode");
    return false;
  }```

and then later:
```  DisplayMode final_mode = mode;
  final_mode.Width = device_screen_physical_width;
  final_mode.Height = device_screen_physical_height;
  OnModeSet(final_mode);```

ivan-mogilko:
Oh right, this check... well it simply means that OpenGL technically cannot run in less than 16-bit, but in practice we are not using any other mode than 32-bit RGBA now. I guess that does not mean that internal functions should not test for possibility that someone demands other modes though... frankly I am not certain how well renderer is prepared for that.

> I know internally the opengl driver is 32bit, but I assumed the intent was for it to mostly present itself as a 16bit device for 16bit games (so getpixel returns expected values etc).

Hmm, okay, I am not fully certain what do you mean by "presenting itself as 16-bit device", so to clarify, the communication between game and renderer goes as this: game asks the renderer to tell which bitmaps it accepts (for creating DDBs from them) - using **GetCompatibleBitmapFormat** method - and then tries to keep its bitmap in this format. It's then up to the renderer to convert them - game does not bother any more than that.
GetCompatibleBitmapFormat may not even be required in practice much. It's left mostly to ensure that bitmaps that renderer receives from the game stay in certain *range* (e.g. strictly 8, 16, and 32 bit), and not some weird format in the middle.

Perhaps to clarify more, engine is separated into game and renderer "domain", game does not bother what and how renders it on screen, it always draws everything in its native format, and the link between these two is when game passes prepared bitmap with DrawSprite.

**GetDisplayDepthForNativeDepth** is simply used for renderer to choose which display mode to run, *in case it wants to choose one to better display the game native format in*, but it does not have to. OpenGL and Direct3D now return always 32 to that.

I'm sorry, but I still don't quite understand what your issue with CopyScreenIntoBitmap is. Did you encounter an error there, or you wanted to prevent some possible issues in the future?

ivan-mogilko:
Oh, right, not DrawSprite, which accepts DDB already, but CreateDDBFromBitmap and UpdateDDBFromBitmap, this is where native bitmaps get converted to renderer's objects.

sonneveld:
> I would like to clarify, what your actual issue with CopyScreenIntoBitmap is? Did you encounter an error there, or you want to prevent possible issues in the future if OpenGL runs in 16-bit display mode for example?

I was running a 16bit game in openGL and `OGLGraphicsDriver::GetCopyOfScreenIntoBitmap` failed because `buffer` was too small. I thought to increase the buffer then I realised the blitting would fail if the destination buffer wasn't 32bit. That's when I introduced the fake bitmap for blitting to allow allegro to do the colour conversion.

ivan-mogilko:
> I was running a 16bit game in openGL and OGLGraphicsDriver::GetCopyOfScreenIntoBitmap failed because buffer was too small.

This is unexpected... I will test this out too.

ivan-mogilko:
@sonneveld I tested most recent ags3 branch with a 16-bit game, and it creates screenshot normally. Renderer was Opengl, in 32-bit mode.

sonneveld:
> This pull request is not longer necessary in its current state in my opinion

Thanks for that. I'll double check with my test games when I have the chance.

PR #699, opened 2019-03-31
Split game lookup into more distinct parts and fixed few mistakes

ivan-mogilko:
This pr seem a bit erratic, because having doubts I started changing code from afar and also added small optimisation along the way.

The main idea was to split game lookup into two parts:
1) explicitly defined game location (by command line or other external data coming from port wrappers);
2) search in common locations around current dir and engine executable.

The first method is enabled always, while the second is only available on certain systems where it makes sense (desktops such as Windows, Linux, Mac).

I also changed it so that if there was an invalid path passed as a startup option then engine simply reports error and bails out instead of trying to search around, because that proved to be confusing users (and developers :) ).

Also:
* Grouped together initialization functions called at startup in a group of functions that find and init game-related data, and functions that setup and init various systems (keyboard, mouse). This part of the engine is still bit messy but unfortunately everytime I have to touch it I cannot spare enough time to rewrite it clean, so have to do it in small steps.
* Optimised opening audio and voice assets by not searching for vox pack everytime and testing whether it's a valid package (instead save paths found at startup). Loading assets from separate packs is still done in suboptimal way though, I will write a ticket about it for the future.
* Fixed a bug when startup option contained relative path: because engine had to switch current working directory there this relative path became invalid, yet still used later causing error. Instead engine will now immediately save it as an absolute path.


ivan-mogilko:
Ammended with couple of fixes.

PR #700, opened 2019-04-03
AGS3 C++ modernisation fixes

sonneveld:
Modernise some of the c++ syntax using clang-tidy. This is only a refactor, it does not change any functionality.

- Replaces all instances of 0 and NULL with `nullptr`. 
- ~in master: Introduces `override` for all overriden methods.~
- Replaces empty `{}` constructors and destructors with `default`
- ~removed for now: Introduces `auto` ONLY for iterators and where the type is already on the right.~

Commands used:

```
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
run-clang-tidy-3.8.py -header-filter='.*'  -checks='-*,modernize-use-nullptr' -fix
run-clang-tidy-3.8.py -header-filter='.*'  -checks='-*,modernize-use-override' -fix
run-clang-tidy-3.8.py -header-filter='.*'  -checks='-*,modernize-use-default' -fix
run-clang-tidy-3.8.py -header-filter='.*'  -checks='-*,modernize-use-auto' -fix
```

### DIFF
--- a/Common/ac/dialogtopic.cpp
+++ b/Common/ac/dialogtopic.cpp
@@ -22,7 +22,7 @@ void DialogTopic::ReadFromFile(Stream *in)
     in->ReadArray(optionnames, 150*sizeof(char), MAXTOPICOPTIONS);
     in->ReadArrayOfInt32(optionflags, MAXTOPICOPTIONS);
     // optionscripts pointer is not used anywhere in the engine
-    optionscripts = NULL;
+    optionscripts = nullptr;
     in->ReadInt32(); // optionscripts 32-bit pointer
     in->ReadArrayOfInt16(entrypoints, MAXTOPICOPTIONS);
     startupentrypoint = in->ReadInt16();
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -24,16 +24,16 @@ using namespace AGS::Common;
 
 GameSetupStruct::GameSetupStruct()
     : filever(0)
-    , intrChar(NULL)
-    , charScripts(NULL)
-    , invScripts(NULL)
+    , intrChar(nullptr)
+    , charScripts(nullptr)
+    , invScripts(nullptr)
     , roomCount(0)
-    , roomNumbers(NULL)
-    , roomNames(NULL)
+    , roomNumbers(nullptr)
+    , roomNames(nullptr)
     , audioClipCount(0)
-    , audioClips(NULL)
+    , audioClips(nullptr)
     , audioClipTypeCount(0)
-    , audioClipTypes(NULL)
+    , audioClipTypes(nullptr)
     , scoreClipID(0)
 {
     memset(invinfo, 0, sizeof(invinfo));
@@ -59,7 +59,7 @@ void GameSetupStruct::Free()
         for (int i = 0; i < numcharacters; ++i)
             delete intrChar[i];
         delete[] intrChar;
-        intrChar = NULL;
+        intrChar = nullptr;
     }
 
     if (charScripts)
@@ -67,7 +67,7 @@ void GameSetupStruct::Free()
         for (int i = 0; i < numcharacters; ++i)
             delete charScripts[i];
         delete[] charScripts;
-        charScripts = NULL;
+        charScripts = nullptr;
     }
     numcharacters = 0;
 
@@ -79,7 +79,7 @@ void GameSetupStruct::Free()
         for (int i = 1; i < numinvitems; i++)
             delete invScripts[i];
         delete invScripts;
-        invScripts = NULL;
+        invScripts = nullptr;
     }
     numinvitems = 0;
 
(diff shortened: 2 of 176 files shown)


sonneveld:
Note: Some of the Windows only files were not changed as I don't have a cmake/clang-tidy setup yet for Windows.

ivan-mogilko:
Okay, I have to honestly say I am against "auto" except when it replaces iterator types in the loop, because this is where I don't care about their real type.

When I read the code I instinctively look to the left side where the variables are declared to know their type. This change forces me to move eyes right to see what kind of actual object was created. This is truly annoying,

ivan-mogilko:
Alright, let's assume I will have to adjust my code reading.

On a side note, regarding auto, there's a different code style where auto is ammended with pointer symbol, like `auto*`, this was used by mgambrell in his pull request, and may be found here for example:
https://github.com/adventuregamestudio/ags/blob/ags3/Engine/ac/audiochannel.cpp#L47

When applying that commit I tried to follow same convention.

Is it a thing? Does it give any benefit?

Common/ac/gamesetupstruct.h
```diff
@@ -100,7 +100,7 @@ struct GameSetupStruct: public GameSetupStructBase {
 
 
     GameSetupStruct();
-    ~GameSetupStruct();
+    ~GameSetupStruct() override;
```
ivan-mogilko: Ha! This actually points to mistake. I think GameSetupStructBase's destructor should not be virtual, because there this inheritance was never supposed to be virtual, its just one struct extending another.

ivan-mogilko:
The "override" commit seem ok, except it broken declaration padding in few places, but I guess that could be fixed manually later when noticed.

sonneveld:
> auto* : Is it a thing? Does it give any benefit?

I think it's a hint that it's a pointer to the reader and the compiler, but it doesn't affect the type. If you're calling `new`, you would expect the return value to be a pointer.

As it turns out, a newer version of clang-tidy can optionally leave those stars in but I didn't have that version with my old Ubuntu.

If you don't like it, I can remove `auto` for the case where types are on the right, and leave for iterators, I just thought it was an instance where `auto` makes sense (since you're duplicating the type name on the one line)

Common/ac/gamesetupstruct.h
sonneveld: I also noticed `UnInit` was not actually overriding in the graphics drivers, but marked as virtual anyway.

Are these things you'd like fixed in this branch?

ivan-mogilko:
> I think it's a hint that it's a pointer to the reader and the compiler, but it doesn't affect the type. If you're calling new, you would expect the return value to be a pointer.

> As it turns out, a newer version of clang-tidy can optionally leave those stars in but I didn't have that version with my old Ubuntu.


I was thinking about this, and I guess I'd rather have these pointers visible as a common rule of thumb, to distinguish raw pointers from smart pointers and other wrappers, as a hint and reminder that may be something that has to be deleted or should be shared with caution.

rofl0r:
wow. 248 changed files for the sake of style. since this touches almost every source file, this will probably create merge issues for every pending PR (even those some people work on currently that are not opened yet), just to make the code more "modern". also as @ivan-mogilko noted, using `auto` makes it hard to figure out what goes on. could we stop being hipsters and invest the effort to fix bugs instead? the usage of nullptr and similar things can be introduced gradually in chunks of codes that need be changed anyhow.

rofl0r:
> could we stop being hipsters and invest the effort to fix bugs instead?

there are 140 open issues, awaiting investigation and bugfixes.

ivan-mogilko:
> there are 140 open issues, awaiting investigation and bugfixes.

Even though I am not that much against auto etc in general sense, I must agree to what rofl0r means here.

My intent was to introduce **override** mainly because it helps when changing parent classes (which we do often during refactoring old code).

The other changes, especially "auto" require time to get used to, personally I don't have much, and afraid I will cling to old style for some time anyway just out of habit.

And yes, meanwhile there are *lots* of more important things that I'd rather have done, ones we have in a tracker is just a top of an iceberg that users and developers were arsed to report and document.

ivan-mogilko:
@sonneveld I think I will just cherry pick your commit with override for now.

sonneveld:
> The other changes, especially "auto" require time to get used to, personally I don't have much, and afraid I will cling to old style for some time anyway just out of habit, so this automatically generated style will remain not for long.

That's fair. I decided I would remove the `auto` commit anyway, since the newer clang-tidy has more options. It'll all change if we decide to start using c++14's `make_unique`

sonneveld:
> could we stop being hipsters and invest the effort to fix bugs instead?

I believe it's more helpful to contribute fixes rather than complain about them.

sonneveld:
These changes do provide a useful purpose however. Using `nullptr` instead of 0 of NULL reduces ambiguity when resolving overloaded methods (int vs ptr). The `default` constructor/destructor usage can potentially allow the compiler to produce slightly more efficient code as it doesn't have to create empty functions.

ivan-mogilko:
Alright.

PR #703, opened 2019-04-06
AGS3: Add CMake as an optional build system.

sonneveld:
This introduces CMake as an optional build system. This helps me develop across all three systems because I don't have to edit 3 different build systems. This PR does not replace the current build systems and they should definitely still work.

The advantages/features of CMake include:

- easier to maintain across multiple platforms
- can define projects and their dependencies
- can generate IDE projects for Visual Studio and XCode (supports multiple Visual Studio versions so we don't need to maintain project files for older IDEs)
- supports toolchain files for cross platform compilation (such as android and ios)
- integrates with tools like clang-tidy. Supported by Visual Studio natively.
- out of source builds! I can share a single repository with VMs and not overwrite .o files
- Integration with google tests
- potential for package management like vcpkg
- will properly rebuild targets if defines and source change.
- different build systems on Linnux. Try out 'ninja' for slightly faster builds
- bundle mac applications
- create installer packages with CPack
- some of our dependencies already use cmake (ogg, vorbis, sdl, allegro)

Brief documentation is in `CMAKE.md`

~now merged in: This includes rofl0r's freetype branch with a `CMakeLists.txt` file that should work on our supported desktop systems.~


sonneveld:
If there's no constructive criticism, I'll merge this in after #690 .

rofl0r:
i heavily object. this is precisely the sort of fanboi-ism i now pointed out repeatedly, and have seen bring other projects to their knees. it is changing stuff solely for the sake of personal preferences, without any respect for the resulting consequences. in the first round this will be made optional in order to not cause too much drama, but in the second round the support for the traditional build system will be phased out, because nobody wants to maintain 2 build systems.
@ivan-mogilko please veto this.

sonneveld:
You realise we're maintaining multiple build systems already right?

Windows MSVC, Makefile (macos/linux), ant (Android), xcode(ios), _another_ xcode(macos)

Of course my plan is to collapse a lot of these into a single build system.  We already have problems where only one build system is updated, to the detriment of the others.

This is solving an actual problem. CMake lets you choose the IDE and build tool you want to use on each system. Just having plain Makefiles does not give you that.

morganwillcock:
I think it is a good idea. People are struggling to build what is there, and we just lost an entire platform because no-one can build it. vcpkg looks useful for Windows builds, because building dependencies from source encounters out-of-date Solutions and automated upgrades are not supported by the build tools (so these will never work with a CI environment unless the environment is an exact match). From a philosophical point of view, perhaps it is desirable to stick with the more traditional tools, but realistically I don't consider the majority of the target platforms to be traditional.

ivan-mogilko:
@sonneveld, does this have to be put into 3.5.0 directly now, will it give any immediate benefit to it?
We were previously discussing your development and I suggested to add additional temp branch in main repository where you could put your work on sdl2 if you wish.

I am asking because this release is a bit late already and yet there is a number of things that would be good to fix or ammend and I'd wish we could focus on these instead. I may even add more tasks to milestone which I kept in my private todo list because was not sure if there will be time for them.
For example also related to automation, @ericoporto was setting automatic building on number of services such as Travis CI, and which are not fully complete yet. Not sure how what he's doing will correlate with added cmake.

@rofl0r, not sure what it has to do with "fanboism". Everyone has their own priorities I guess, but it is also not immediately clear what _majority_ of people using this repository would prefer.

The big problem here is that I cannot make decisions about everything because it's tiresome and I am not an expert in everything. For many periods of time I've been almost sole project maintainer, so that could make an impression I "own" it or have a big plan for its course, but I did not and never intended to do so. Also, I have constant doubts about my future participation here.

If there were actual alternative suggestions on centralized build system we could review them, but no one did in years.

Regarding CMake, what is exactly terrible about it? I recall you mentioned it requires C++ compiler, but so does our engine in the first place. Also that produced makefiles may be only run in presence of CMake. Is this true?

---

BTW, I have my own questions about this. For starters, in the past when I suggested other contributors to build our patched Allegro 4 on Windows using available CMake setup to create MSVS projects, some people protested and asked for actual MSVS solutions, which I added for their convenience. Won't same happen in our case? I guess many people find it preferable to just run existing solution in the IDE.

rofl0r:
> People are struggling to build what is there, and we just lost an entire platform because no-one can build it.

why are they struggling? because they are windows users that expect they can make a double click and everything works automatically. it they need to do anything manually, they struggle. it's because they're lacking skill, not because GNU make is lacking.
if you want **one build system to rule them all** you can extend the existing GNU make files and they can be used with msys and mingw compiler even on windows. the only platform that is special and requires to do everything differently, is M$ visual studio.

> @rofl0r, not sure what this has to do with "fanboism"

well, simply put, @sonneveld is a fanboi of C++ and cmake. he studies the latest features that have been added to the C++ spec and looks where he can try them out. unfortunately, he decided this project is the right place for his experiments.
i don't recall a single productive contribution of him, all he's done so far is pushing changes that use modern C++ features (or using very simple requests to address them, while using them as a backdoor to smuggle in yet more C++ features), and @ivan-mogilko has to clean up behind him to fix the bugs he introduced.

> how this may bring a project to the knees?

you already answered your question yourself with the following quote:

> I have constant doubts about my future participation here.

you're the only person here that actually provides valuable contributions and fixes real issues. but meanwhile, you spend 90% of your effort to review and clean up behind the changes of @sonneveld which provides nothing of value, only changes that introduce dependencies on the latest compilers, on new build systems, and whatever came along his way that he finds sexy and wants to try out. even his SDL2 branch is only a half-assed attempt, as he recently stated that it would require SDL2 + a custom, minimized version of allegro, so in the end the change provides zero benefits but only adds one dependency more.

if this goes on a couple more months like this, you will lose the will to deal with his crap, and @sonneveld will take over the project, and after he will have finished his C++14 experiments, he will lose interest and the project will be dead. besides, he will never add anything useful.

as you can see, lately he got confident started pushing directly to the branch without opening PRs.

people like this are toxic and you better kick this guy out now before its too late.

> Regarding CMake, what is exactly terrible about it? 

it's over a million lines of code which proves it was written by idiots, it takes a very long time to build it from source, it requires learning an ugly and unintuitive new programming language to write build recipes, and the people that wrote it ignored all existing conventions and best practices and reinvented the wheel. its only appeal is that it has a windows gui and can generate a MSVC project, where you can click on, so non-skilled users can use it easily (but only on windows).

it took a lot of effort to come up with the existing build system, it works fine, and there's no reason to throw it over board for something new which will open a whole new can of worms.

ivan-mogilko:
@rofl0r 
> i don't recall a single productive contribution of him

To name a few, in the past he set up automatic build for all the project which made my life much easier when making releases (unfortunately because my temporary avoidance of this project build server licenses were cancelled and it's gone now), worked on OSX and iOS ports which Wadjet Eye use for their games, and also did some script compiler refactoring and fixed few bugs there.

His current particupation may seem messy, but I think that's rather because the timing is bad and there's again a conflict between upcoming version release and people who want to push the engine further with big changes, such as switching away from allegro 4.

> > how this may bring a project to the knees?

> you already answered your question yourself with the following quote:

> > I have constant doubts about my future participation here.

> you're the only person here that actually provides valuable contributions and fixes real issues. but meanwhile, you spend 90% of your effort to review and clean up behind the changes of @sonneveld w

No, this has little to do with issues which happened only recently. The two main reasons always were that a) I felt like working alone most of time and b) I tend to believe that I lead engine development into the wrong direction for last several years.

I lost the will to deal with this project years ago, and not able to explain why I'm still here even to myself. If anyone wants to take over, I honestly would not mind at all (I've been asking if anyone wants on AGS forums for years).

sonneveld:
> If there were actual alternative suggestions on centralized build system we could review them, but no one did in years. 

I settled on CMake mainly because of its popularity and its support in a number of IDEs (including Visual Studio 2017/2019). Almost all of our current dependencies have a bundled CMake file we can link to.

 > @sonneveld, does this have to be put into 3.5.0 directly now, will it give any immediate benefit to it?
 > For example also related to automation, @ericoporto was setting automatic building on number of services such as Travis CI, and which are not fully complete yet. How what he's doing will correlate with added cmake?
 > Even if cmake is optional now, CMakeLists will have to be constantly ammended, aren't they?

I arranged this pull request in such a way that it would not affect 3.5. I kind of wanted to get it in early so people could try it out and iron out any issues.  It might actually make things easier for @ericoporto as he can use standard cmake build commands in the CI services.  Another benefit is that we can create macOS app bundles from cmake, which will help people who want to publish games on that platform.

The CMakeLists will have to be amended but that's business as usual for all our non-Windows build systems. At the very least it is plain text so others can edit it. And on the other hand, you can actually test the cmakefiles in Windows (unlike the current Makefiles and XCode projects)

> in the past when I suggested other contributors to build our patched Allegro 4 on Windows using available CMake setup to create MSVS projects, some people protested and asked for actual MSVS solutions, which I added for their convenience. Won't same happen in our case? I guess many people find it preferable to just run existing solution in the IDE. How are we going to deal with such requests in the future?

Good question. Cmake generates code with absolute paths so its output is not designed to be commited to the repository. The difference now is that current versions of Visual Studio supports cmake natively (I'ved tested it but I've stuck with cmake command line tools for now).  I'm also hoping that when we can build the engine and all the dependencies in one project, it will offset the slight inconvenience of having to run cmake.

Generating the MSVS Solution files via cmake is a once-off thing. After that, the generated project files will update themselves if you edit the CMake files.

 > If anyone wants to take over, I honestly would not mind at all (I've been asking if anyone wants on AGS forums for years).

I would be happy to step up with this. I was mainly worried how other people would feel about it since I'm kind of an outsider.

sonneveld:
> because they are windows users that expect they can make a double click and everything works automatically. it they need to do anything manually, they struggle. it's because they're lacking skill, not because GNU make is lacking.

This is not constructive. The majority of Windows development is done in Visual Studio and we should respect and support that.

 > even his SDL2 branch is only a half-assed attempt, as he recently stated that it would require SDL2 + a custom, minimized version of allegro, so in the end the change provides zero benefits but only adds one dependency more.

This suggests to me you actually haven't reviewed the SDL2 port branch. AGS has a lot of functions that depend on the specific implementation of the Allegro software drawing routines, which is why I decided to keep some of the Allegro library. Everything else is to be provided by C++ libraries or SDL2 (which is maintained and supports the systems we care about without patching).

sonneveld:
I've modified the engine tests to run through cmake's ctest. In Linux, for example, you would run

`make engine-test && make test`

ivan-mogilko:
Is there a way to not have duplicates of library headers (a88c3d15622ad35cdfb744f058fa76cc1fd4c3b4)?

sonneveld:
> Is there a way to not have duplicates of library headers ([a88c3d1](https://github.com/adventuregamestudio/ags/commit/a88c3d15622ad35cdfb744f058fa76cc1fd4c3b4))? What's the purpose of them?

I tried to bundle the headers for each library so I could restrict their scope. That way when we link libraries, they only see the headers that they need.  I didn't want to just delete `Common/libinclude` because it would break all the other build systems.

sonneveld:
So this is working pretty well on Windows,Linux,macOS now.  There was some work I wanted to do with respect to building ogg/vorbis/theora and allegro from the one project but I will leave that for a future `ags3-next` branch (where work for the next version would occur).

rofl0r:
what.da.fuq

ivan-mogilko:
> I tried to bundle the headers for each library so I could restrict their scope. That way when we link libraries, they only see the headers that they need. almp3 also had some fixed paths, which weren't needed when it has its own include dirs correctly set.

> I didn't want to just delete Common/libinclude because it would break all the other build systems.

I don't agree with this approach. If Common library needs library headers for some reasons then we could as well move library sources to Common instead of making duplicates.

For instance, alfont has to be copied to Common anyway (I already noted this in other thread).

But first we need to test which headers are actually required for Common at all (maybe some of them could be removed).

rofl0r:
was there some behind-the-doors agreement to merge this PR, or did i miss something ?

ivan-mogilko:
> was there some behind-the-doors agreement to merge this PR, or did i miss something ?

I have no idea myself.

rofl0r:
> I have no idea myself. I've been very busy today and noticed this only now.

i suppose that asks for a force push, resetting ags3 to before the merge, unless you accept that @sonneveld makes the major decisions on his own now.

sonneveld:
> But first we need to test which headers are actually required for Common at all (maybe some of them could be removed).

The only dependencies that Common has are Allegro, Alfont/Freetype and AAStr. The rest belong to Engine.  The problem I had is that we had a pool of mixed include files in `Common/libinclude` and no easy way to separate them per project.

Issue #701, opened 2019-04-03
Game.DoOnceOnly reset state


Dualnames1:
It would be useful to be able to do a reset on all the Game.DoOnceOnly's used
But also being able to get their state like so

bool HasGetGameDoOnceOnlyBeenUsed(String token)  (possibly a far better name).
But definitely the reset probably. Though to be fair for this custom save game system I'm working on, a replacement function is quite more likely.

ivan-mogilko:
Something to consider maybe, I was wondering other day how easy that would be to implement and expose a key/value map (ordered or not) into ags script.

We already have String, which is a basic container of chars.

What if we make Map or Dictionary in a same way?

Then you could script your own "do once only" and even have multiple instances of these for different purposes.

ericoporto:
I have no idea on how to implement this, but I am interested in using it. So just some random thoughts, that can be wrong, maybe it will be easier to develop it in shipable states, so people can test it:

1. map
2. Game DoOnceOnly using map as a module
3. CustomProperties as a map
4. Plugin interfaces for map

Far future things that should be only thought about in the future: sorted map, custom sorting, allowing("this").or.something.similar() //to ease deep working with deep dictionaries

Being able to test maybe requires landing in 3.5.0, or 3.6.0 (or is ags4 actually working?)

ivan-mogilko:
> I have no idea on how to implement this, but I am interested in using it.

Set or Map script class may actually be relatively easy to do. There is a bunch of small script classes declared in API that may be used for the reference. For example, you may find Camera or Viewport classes added recently. Internally, it should work more like DynamicArray (having a std::unordered_map<String,String> as a storage).

I'd say that on its own could be done in 1-2 days, assuming you know how to add new script API.
Reimplementing DoOnceOnly and custom props as a new map class will take more (and may have some compatibility issues to resolve), but if we will add at least an unordered string set and/or map that may be created in script that alone would cover most of use cases.

I will open a ticket for this soon.

ivan-mogilko:
Set and Dictionary were merged (#714, included with alpha build 3.5.0.9). So on one hand now one could use custom Set instead of Game.DoOnceOnly, on another hand it should be possible to reimplement that function with a Set.

E.g. something like `readonly attribute Set* Game.DoOnce;` (readonly means container object itself cannot be replaced).

[...]

PR #714, opened 2019-04-12
Script API: string Set and Dictionary types

ivan-mogilko:
AGS script currently does not support templates, generics, or any good ways to reuse algorithms or user classes with multiple types, which is probably a reason we don't have any containers rather than dynamic arrays there.

In this situation it may be still a good idea to give users at least most commonly used type of key/value container which works with strings, because even if suboptimal but it is possible to store almost anything in a string, even invent a reference to other objects.

Also, I've been thinking that it may be viable to replace some of the existing API with Set or Dictionary, extending their use and reducing duplicate functionality at the same time. For example, this could be Custom Properties and Game.DoOnceOnly (see discussion in #701).

The suggested API for Set and Dictionary is this:

<pre>
builtin managed struct Dictionary
{
  /// Creates a new empty Dictionary of the given properties.
  import static Dictionary* Create(bool sorted = false, bool caseSensitive = false);

  /// Removes all items from the dictionary.
  import void Clear();
  /// Tells if given key is in the dictionary.
  import bool Contains(String key);
  /// Gets the value by key; returns null if no such key is in dictionary.
  import String Get(String key);
  /// Removes key/value pair from the dictionary, fails if there was no such key.
  import bool Remove(String key);
  /// Assigns a value to the given key, adds this key if it did not exist yet.
  import void Set(String key, String value);

  /// Gets if this dictionary is case-sensitive.
  import readonly attribute bool CaseSensitive;
  /// Gets if this dictionary is sorted by keys in alphabetical order.
  import readonly attribute bool Sorted;
  /// Gets the number of key/value pairs currently in the dictionary.
  import readonly attribute int ItemCount;
  /// Creates a dynamic array filled with keys in same order as they are stored in the Dictionary.
  import String[] GetKeysAsArray();
  /// Creates a dynamic array filled with values in same order as they are stored in the Dictionary.
  import String[] GetValuesAsArray();
};

builtin managed struct Set
{
  /// Creates a new empty Set of the given properties.
  import static Set* Create(bool sorted = false, bool caseSensitive = false);

  /// Adds item to the set, fails if such item was already existing.
  import bool Add(String item);
  /// Removes all items from the set.
  import void Clear();
  /// Tells if given item is in the set.
  import bool Contains(String item);
  /// Removes item from the set, fails if there was no such item.
  import bool Remove(String item);

  /// Gets if this set is case-sensitive.
  import readonly attribute bool CaseSensitive;
  /// Gets if this set is sorted in alphabetical order.
  import readonly attribute bool Sorted;
  /// Gets the number of items currently in the set.
  import readonly attribute int ItemCount;
  /// Creates a dynamic array filled with items in same order as they are stored in the Set.
  import String[] GetItemsAsArray();
};
</pre>

### Notes:

1. Each of these interfaces may have 4 different implementations behind them: sorted, sorted with case-insensitive key comparison, unsorted and unsorted with case-insensitive comparison. This is to compensate lack of generic types and function pointers in AGS script. Of course it's not absolutely universal, but covers most common use cases in my opinion.
At the same time I am still not certain about necessity to have unsorted variant. In a normal programming language the choice of container is defined by perfomance priorities, but in case of AGS  the script's own overhead may be larger than overhead by imperfect container choice, so benefit may be unnoticable, except when having extra large number of elements.

~2. Set and Dictionary provide access to their item arrays, which may normally be useful for sorted variant. I guess for non-sorted implementation it may make sense only for diagnostic purposes.
In any case the random access goes against set/map purpose and using arbitrary indexes may be very slow. I tried to optimise implementation for the case of sequential access, meant for iterating items in a loop.~

2. Both have methods which return Strings from container in dynamic array (the order matches one of container, i.e. - sorted or random). This is not very optimal for keeping in sync with containers that change often, but a trivial and universal method for starters that will remain useful even after something more optimal for fast iteration is added.

3. There's a certain performance problem with lookup operations. std::set and map demand strictly Key type for lookup, and AGS interpreter only passes "const char*" to the script functions (which is a pain and issue on its own). Therefore I have to allocate a String object for each look up (also when removing or setting value).

I've read that C++14 standard supports using "equivalent" types for lookup and insertion ([see docs](https://en.cppreference.com/w/cpp/container/set/find) and [wikipedia paragraph](https://en.wikipedia.org/wiki/C%2B%2B14#Heterogeneous_lookup_in_associative_containers)), which sounds like what I could use, but we don't officially support that in AGS at the moment (and I've never used that before so still no idea how well that would fit).

I did few quick attempts on a wrapper, but it ends up either by storing an ugly type inside containers, or very unsafe subtype of String, so decided to drop this idea until better times.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -457,6 +457,32 @@ internalstring autoptr builtin managed struct String {
   readonly import attribute int Length;
 };
 
+#ifdef SCRIPT_API_v350
+builtin managed struct Set
+{
+  /// Creates a new empty Set of the given properties.
+  import static Set* Create(bool sorted = false, bool caseSensitive = false); // $AUTOCOMPLETESTATICONLY$
+
+  /// Adds item to the set, fails if such item was already existing.
+  import bool Add(String item);
+  /// Removes all items from the set.
+  import void Clear();
+  /// Tells if given item is in the set.
+  import bool Contains(String item);
+  /// Removes item from the set, fails if there was no such item.
+  import bool Remove(String item);
+
+  /// Gets if this set is case-sensitive.
+  import readonly attribute bool CaseSensitive;
+  /// Gets if this set is sorted in alphabetical order.
+  import readonly attribute bool Sorted;
+  /// Gets the number of items currently in the set.
+  import readonly attribute int ItemCount;
+  /// Creates a dynamic array filled with items in same order as they are stored in the Set.
+  import String[] GetItemsAsArray();
+};
+#endif
+
 builtin managed struct AudioClip;
 
 builtin managed struct ViewFrame {
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -458,6 +458,34 @@ internalstring autoptr builtin managed struct String {
 };
 
 #ifdef SCRIPT_API_v350
+builtin managed struct Dictionary
+{
+  /// Creates a new empty Dictionary of the given properties.
+  import static Dictionary* Create(bool sorted = false, bool caseSensitive = false); // $AUTOCOMPLETESTATICONLY$
+
+  /// Removes all items from the dictionary.
+  import void Clear();
+  /// Tells if given key is in the dictionary.
+  import bool Contains(String key);
+  /// Gets value by the key; returns null if such key does not exist.
+  import String Get(String key);
+  /// Removes key/value pair from the dictionary, fails if there was no such key.
+  import bool Remove(String key);
+  /// Assigns a value to the given key, adds this key if it did not exist yet.
+  import bool Set(String key, String value);
+
+  /// Gets if this dictionary is case-sensitive.
+  import readonly attribute bool CaseSensitive;
+  /// Gets if this dictionary is sorted by keys in alphabetical order.
+  import readonly attribute bool Sorted;
+  /// Gets the number of key/value pairs currently in the dictionary.
+  import readonly attribute int ItemCount;
+  /// Creates a dynamic array filled with keys in same order as they are stored in the Dictionary.
+  import String[] GetKeysAsArray();
+  /// Creates a dynamic array filled with values in same order as they are stored in the Dictionary.
+  import String[] GetValuesAsArray();
+};
+
 builtin managed struct Set
 {
   /// Creates a new empty Set of the given properties.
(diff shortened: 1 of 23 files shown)


ivan-mogilko:
Added Dictionary.

ivan-mogilko:
Alright, so I rewrote this without quick attempts to optimise lookups, and so these containers just store String now. The reason is that I'd like to keep potential to use these script types to wrap existing engine dictionaries for the future, and I could not come with a good wrapper type compatible with String at this time.

Editor/AGS.Editor/Resources/agsdefns.sh
```diff
@@ -457,6 +457,60 @@ internalstring autoptr builtin managed struct String {
...
+  import bool Remove(String item);
+
+  /// Gets if this set is case-sensitive.
+  import readonly attribute bool CaseSensitive;
+  /// Gets if this set is sorted in alphabetical order.
+  import readonly attribute bool Sorted;
+  /// Gets the number of items currently in the set.
+  import readonly attribute int ItemCount;
+  /// Accesses the items in the set, sorted or unsorted depending on the set's property.
+  import readonly attribute String Items[];
```
sonneveld: Might be worth nothing if modifying the Set affects this getting or not. I still wonder if it's better to just have a "ToSortedArray" method instead.

ivan-mogilko: I have doubts about returning an array *instead* (as opposed to complementary method): the most common use for iterating over set/dictionary content is to display it somewhere, and if a person would like to have it done repeatedly and keep synced at the same time I'm afraid they will do it in a very suboptimal way (creating new array every game tick). Not every ags user knows these nuances.

But I can agree that this Items[] property seem a bit off for this type of container. Maybe there's a way to make a proper "iterator" which only lets to pass container from start to finish in a direct sequence.

ivan-mogilko:
Regarding awkward Items[] property, one of the options is to replace it with C# style Iterator/Enumerator object.

The simpliest API could be:
<pre>
struct Iterator
{
    String Next(); // returns null if no more items
};
</pre>

More sophisticated one (replicates C#):
<pre>
struct Iterator
{
    bool Next(); // return false if no more items
    void Reset(); // back to first item
    attribute String Current;
};
</pre>

ivan-mogilko:
Alright, another take.
Removed controversial "Items[]" property and added Set,GetItemsAsArray(), Dictionary.GetKeysAsArray(), Dictionary.GetValuesAsArray(). They return dynamic array of managed Strings (by copying). Array lengths may be known by reading .ItemCount property.

PR #717, opened 2019-04-13
Fixes for Debian/Ubuntu build script and packaging

morganwillcock:
* Switch from cowbuilder to pbuilder
* Fix for Debian mirror changes - these are permanent changes so the existing bootstrap for chroots would always fail
* Remove unneccesary package requirements on the host running the chroots
* Ubuntu doesn't have `dh` after installing the documented packages, so I've also added a note that Ubuntu will need to install 'debhelper'
* Ignore CMake in Debian packaging by explicitly specifying the build system - dh_auto_configure was autodetecting CMake files, and then the package would fail to build

The main change is switching from cowbuilder to pbuilder, which shouldn't be functionally different, but will work in more places. I don't think anything here is relying on using cowbuilder, and cowbuilder is more likely to have problems based on trying to hardlink across filesystem boundaries.

I've checked that the build script works on:
* Debian 8
* Debian 9
* Ubuntu 14.04
* Ubuntu 16.04
* Ubuntu 18.04

...just using the base OS install and the packages listed in the script file.

### DIFF
--- a/debian/make_ags+libraries.sh
+++ b/debian/make_ags+libraries.sh
@@ -11,7 +11,7 @@ set -e
 
 # Preliminaries for running this script:
 # This script is intended for use on Debian or Ubuntu
-# Install ubuntu-dev-scripts and pbuilder:
+# Install ubuntu-dev-tools and pbuilder:
 
 # sudo apt-get install ubuntu-dev-tools pbuilder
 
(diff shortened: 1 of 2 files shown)


sonneveld:
I think this looks good. I don't know enough about debian packages to know. :)

PR #718, opened 2019-04-13
Enable Travis-CI builds for the ags3 branch

morganwillcock:
This file needs to be present in the branch in order to build it. For release purposes it probably needs to use the chroot build script or also run a build in a 32 bit container, and there is the Allegro version to consider, but for the moment just checking that builds occur is a good start.

Edit: the current build error should be covered by #717

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: c
+compiler:
+  - gcc
+addons:
+  apt:
+    packages:
+    - debhelper
+    - build-essential
+    - pkg-config 
+    - cmake 
+    - libfreetype6-dev 
+    - libogg-dev 
+    - libtheora-dev 
+    - libvorbis-dev
+    - liballegro4-dev
+    - libaldmb1-dev
+    - fakeroot
+script:
+- fakeroot debian/rules binary
+deploy:
+  provider: releases
+  skip_cleanup: true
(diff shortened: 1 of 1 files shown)


morganwillcock:
Next error:
<pre>Linking engine...
/usr/bin/ld: unrecognized option '--push-state'</pre>
I guess this is related to recent changes for compiler requirements or something has introduced an option; the master branch has previously built without the same issue.
<pre>$ gcc --version
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4</pre>

ivan-mogilko:
Could you find the commit when this error appears? Assuming it was building with -std=c++11 CXXFLAGS then most recent changes suppose to not affect building on gcc at all.

What is "--push-state" related to?

sonneveld:
I had some issues linking static and dynamic libraries. I was using push pop state to temporarily change settings for linking a particular library without affecting later ones (which could include system libs)

morganwillcock:
> Could you find the commit when this error appears?

I believe it is 4166c65

morganwillcock:
It seems the older linker just doesn't support that flag, and the commit adds the flag. Maybe that is worthy of being raised a separate issue, but it might not have any real world impact. Probably something for someone else to decide...

I'll update this PR to use the chroots instead, which should let it build successfully.

Issue #558, opened 2018-12-28
3.5.0 engine's Windows XP support


ivan-mogilko:
As we switched to VS2015 and using v140 toolset when building Window version of the engine (that makes engine depend on Windows 8 SDK), Windows XP can no longer run AGS engine (and compiled games).
This may be fixed by switching to v140_xp toolset, that will make engine use Windows 7.1 SDK which is still compatible with XP.

Apparently modern versions of Windows may still be able to run the engine built with v140_xp. But I do not know if that will rise any compatibility issues in the future.

I think possible options here are:
1. Switch to v140_xp toolset completely and stay there unless something goes wrong;
2. Build a separate configuration with v140_xp toolset and ship it with the AGS as "acwin_xp.exe" like we did with "no mp3" version. We may even implement a separate "Build target" in the Editor for convenience.
3. Drop Windows XP support for **games** made with AGS 3.5.0 and higher.

morganwillcock:
Since there is also v141_xp for the newer build tools, and Win32 programs tend to keep working, probably option 1 unless there is something that actually works differently?

sonneveld:
I suggest we drop Windows XP support. It's no longer a supported operating system.

We have to draw the line somewhere, and it's better for us if we can use the latest Visual Studio tools. (which target Win7 and up)

SuperDre:
I would also suggest dropping Windows XP support. People should really upgrade to (at least) Windows 7 or higher or just move to Linux. Microsoft has ditched XP already for a long LONG time, and you cannot expect developers of new games to keep those few people  in mind who still want to run XP. Has anyone actually tested if the build with v140 doesn't run on XP? Same also applies to other older OSses which has been superseeded by many already (ofcourse if there's a problem). If people really want XP support, let them fork their own version and keep it running themselves. But then again, if using v140_xp makes developing newer versions not even the slightest more painfull, and with it it still runs on XP without any adjustments or needing to keep track of things, then ofcourse it's also an option (but again, you can always leave it up to someone to fork it and create a version for XP).
So IMHO I agree with Sonneveld, drop support for an OS that's not safe anymore and not supported by the creator (and let's not forget, they still have an older version of AGS that works with their OS).

rofl0r:
[...]

PR #720, opened 2019-04-14
AGS3 Commonlib: remove support for older compilers

sonneveld:
As we only support minimum vs2015 and c++11 now.

I wasn't sure if I should remove the TR1INCLUDEs support header entirely.

### DIFF
--- a/Common/core/types.h
+++ b/Common/core/types.h
@@ -18,14 +18,30 @@
 #ifndef __AGS_CN_CORE__TYPES_H
 #define __AGS_CN_CORE__TYPES_H
 
-#include "endianness.h"
-
 #if defined (_WINDOWS) && !defined (WINDOWS_VERSION)
 #define WINDOWS_VERSION
 #endif
 
+#if defined(WINDOWS_VERSION)
+    // MSVC didn't report correct __cplusplus value until 2017
+    #if !defined(_MSC_VER) || (_MSC_VER < 1900)
+    #error Visual Studio 2015 or later required.
+    #endif
+#else
+    #if __cplusplus < 201103L
+    #error C++11 or later required.
+    #endif
+#endif
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h> // for size_t
+#include <limits.h> // for _WORDSIZE
+
+#include "endianness.h"
+
 #ifndef NULL
-#define NULL 0
+#define NULL nullptr
 #endif
 
 #ifndef FORCEINLINE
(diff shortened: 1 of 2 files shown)


rofl0r:
exactly as predicted...

ivan-mogilko:
Right, I was thinking of suggesting same thing recently. Earlier this macro was created as a measure to support *newer* compilers while releases were kept created with an old one (other way around).
The engine code demands quite a bit of C++11 support now.

In fact, I just realized that I was forgetting to use this macro in few places before, so it's use in code is inconsistent.

rofl0r:
> It was never made a secret that we are going to gradually move to newer C++ standard, and we were doing so for few years now.

oh, so was there a pressing need to drop support for older compilers

> As we use more of new standard's features

ah, i see. it's all about features features features

ivan-mogilko:
> oh, so was there a pressing need to drop support for older compilers

No, it's not pressing, it's just that I see little point in *pretending* that we are supporting them while the code is probably already not compilable on older ones as it is now without additional works, like adding more macros, or even writing custom implementations of standard library classes (containers etc).

> ah, i see. it's all about features features features

It's not about "features" on their own, it's about convenience when writing a program. "Override" keyword, for instance, helps when changing virtual interfaces; smart pointers, hash maps from standard library. These are examples of what we currently use.

ivan-mogilko:
Assuming we want to support some older compilers for any reason, we'd need to know which versions exactly, because that would define the set of C++11 we are allowed to use. Even if it's really worth it, some of the C++11 commands may be replaced with macros (like "override") but some cannot be (like with "auto").

All of this had to be discussed and decided long beforehand, and some reasoning given to keep the code compatible with those older compilers. But this time is lost, and speaking most frankly I never had time nor interest for doing that.

ivan-mogilko:
For the reference, right now readme states the minimal GCC compiler for ags is 4.4. Here's the list of supported C++11 set in GCC 4.4:
https://gcc.gnu.org/gcc-4.4/cxx0x_status.html

For comparison, C++11 support in GCC 4.8 ([4.8.1 was released May 31, 2013](https://gcc.gnu.org/releases.html)):
https://gcc.gnu.org/gcc-4.8/cxx0x_status.html

ivan-mogilko:
Regarding TR1INCLUDE macro in particular, it is important for those versions of compilers that have newer C++11 types declared under std::tr1 namespace and/or standard headers put in "tr1" subdirectory. Not sure if I remember correctly but the latter could actually be a case in GCC 4.4 in non-C++11 mode. Note also that we currently have C++11 flag in Makefile for gcc, so I have doubts this is still necessary. Guess this has to be double-checked first.

rofl0r:
> I am noting only GCC compilator, because I doubt anyone on Windows would complain they cannot use MSVS 2008 anymore.

stuff like this can open a can of worms. once you require a newer compiler, suddenly people can no longer build the engine for e.g. windows XP, which you may now say is no longer supported anyway, nonetheless there are still people using it. also there's reactOS which is basically on windows XP API level. so you would exclude them as well.

ivan-mogilko:
> stuff like this can open a can of worms. once you require a newer compiler, suddenly people can no longer build the engine for e.g. windows XP, 

We already require new compiler for 3.5.0 from the beginning. Moving to MSVS 2015 was IIRC the first thing we did in this dev branch. With MSVS situation was worsened by fact that full project could not be built with free version of MSVS 2008, only paid one (aka "professional edition"), and most contributors could not afford to have it just for this project.

MSVS 2015 and following versions have a compilation mode with xp compatibility. We were discussing it here: #558. But as far as I know, XP support depends not on the language version, but on the used "toolset" (linked runtime libraries).

sonneveld:
> nce you require a newer compiler, suddenly people can no longer build the engine for e.g. windows XP, which you may now say is no longer supported anyway, nonetheless there are still people using it. also there's reactOS which is basically on windows XP API level. so you would exclude them as well.

Windows XP hasn't been supported for five years. None of us have reactOS experience, so it would be best if someone did an explicit port to that OS rather than us trying to support an old runtime.

PR #725, opened 2019-04-17
AGS 3: support for multiple cameras and viewports

ivan-mogilko:
This implements support for multiple room Cameras and Viewports.

User is allowed to create any number of Viewports on screen and Cameras in room, or remove them as necessary. There's one "primary" Viewport and "primary" Camera under index 0 that can never be removed (they are also created automatically). Latter also ensures full backwards support for games that do not have this new functionality.

Following was added to script API:

<pre>
builtin struct Room {
  <...>
  /// Gets the Camera by index.
  import static readonly attribute Camera *Cameras[];
  /// Gets the number of cameras.
  import static readonly attribute int CameraCount;
  /// Creates a new Camera.
  import static Camera *CreateCamera();
  /// Removes an existing camera; primary camera will never be removed
  import static void RemoveCamera(int id);
};

builtin managed struct Viewport {
  <...>
  /// Gets/sets the room camera displayed in this viewport.
  import attribute Camera *Camera;
  /// Gets/sets whether the viewport is drawn on screen.
  import attribute bool Visible;
};

builtin struct Screen {
  <...>
  /// Gets a Viewport by index.
  import static readonly attribute Viewport *Viewports[];
  /// Gets the number of viewports.
  import static readonly attribute int ViewportCount;
  /// Creates a new Viewport.
  import static Viewport *CreateViewport();
  /// Removes an existing viewport; primary viewport will never be removed
  import static void RemoveViewport(int id);
};
</pre>

### Missing functionality (TODO)

~1. Support Z-order for Viewports in case they overlap;~
2. Add Clickable property to Viewport?
~3. Write viewport and camera data to game saves.~

### Conceptual problem

Although technically everything works, but there are certain things about viewport and camera behavior that were not thought through beforehand. I am considering opening a separate discussion about that (and it may be decided and changed later too). But in short, the big questions are:

* What does Viewport and Cameras belong to, Game or Room? Viewport may belong to game (screen), while Camera to the room for example, or both belong to game or room.
* Should Cameras save their state and restore it as the room loads and unloads? Should viewports?

Imho it's better to assume hypothetically that multiple rooms may be loaded and displayed at the same time (because that's a future possibility too). Of course this does not have to happen soon, but it's better to devise a proper concept now.

### Demonstration

Demo game: https://www.dropbox.com/s/zp0e4cidkate0rc/test--multiviewport.zip?dl=0

<details>
  <summary>Examples of use and proof of concept:</summary>

  Cutscene with two cameras following 2 characters as they walk around the large room
  

  One camera with regular zoom follows character, another camera displays larger piece of room.
  Note that you can interact with the room clicking on either.
  
</details>

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -622,6 +622,15 @@ builtin managed struct Room {
 #ifdef SCRIPT_API_v3507
   /// Gets the room camera
   import static readonly attribute Camera *Camera;
+
+  /// Gets the Camera by index.
+  import static readonly attribute Camera *Cameras[];
+  /// Gets the number of cameras.
+  import static readonly attribute int CameraCount;
+  /// Creates a new Camera.
+  import static Camera *CreateCamera();
+  /// Removes an existing camera; primary camera will never be removed
+  import static void RemoveCamera(int id);
 #endif
 };
 
@@ -2777,8 +2786,8 @@ builtin managed struct Viewport {
   import attribute int Width;
   /// Gets/sets the viewport's height in screen coordinates.
   import attribute int Height;
-  /// Gets the room camera displayed in this viewport.
-  import readonly attribute Camera *Camera;
+  /// Gets/sets the room camera displayed in this viewport.
+  import attribute Camera *Camera;
 
   /// Returns the viewport at the specified screen location.
   import static Viewport *GetAtScreenXY(int x, int y);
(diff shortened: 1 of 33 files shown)


ivan-mogilko:
Rebased to include linux fixes, and also split one commit into two trying to make changes more clear.

sonneveld:
Small fix needed to compile: 

```
RuntimeScriptValue Sc_Room_RemoveCamera(const RuntimeScriptValue *params, int32_t param_count)
{
    API_SCALL_VOID_PINT(Room_RemoveCamera);
}
```

There was an extra argument passed to the macro.

ivan-mogilko:
Added serialization and fixed remaining bugs I found.

PR #726, opened 2019-04-18
AGS3: GameLoopUntilEvent is slightly more type safe

sonneveld:
Store pointers in void* instead of long (which is 32bit on windows), and introduce some safer methods to call it.

### DIFF
--- a/Engine/main/game_run.cpp
+++ b/Engine/main/game_run.cpp
@@ -839,7 +839,7 @@ int ShouldStayInWaitMode() {
     }
     else if (restrict_until==UNTIL_INTIS0) {
         int*wkptr=(int*)user_disabled_data;
-        if (wkptr[0]<0) retval=0;
+        if (wkptr[0]==0) retval=0;
     }
     else if (restrict_until==UNTIL_SHORTIS0) {
         short*wkptr=(short*)user_disabled_data;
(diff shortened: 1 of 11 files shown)


Engine/main/game_run.cpp
```diff
@@ -839,7 +839,7 @@ int ShouldStayInWaitMode() {
     }
     else if (restrict_until==UNTIL_INTIS0) {
         int*wkptr=(int*)user_disabled_data;
-        if (wkptr[0]<0) retval=0;
+        if (wkptr[0]==0) retval=0;
```
ivan-mogilko: How do you know it's a bug?

Engine/main/game_run.cpp
```diff
@@ -910,30 +920,72 @@ void SetupLoopParameters(int untilwhat,long udata,int mousestuff) {
...
+  our_eip = 78;
+
   restrict_until = cached_restrict_until;
   user_disabled_data = cached_user_disabled_data;
   user_disabled_for = cached_user_disabled_for;
 }
 
+void GameLoopUntilValueIsZero(const char *value) 
+{
+    GameLoopUntilEvent(UNTIL_CHARIS0, value);
```
rofl0r: how does going through one more level of indirection increase type safety ?

ivan-mogilko: The first function demands particular type, and nested one supplies correct enum constant for that type.

PR #728, opened 2019-04-18
AGS3 modify DateTime_Now_Core etc to not call time().. twice.. for no good reason

sonneveld:


Engine/platform/base/agsplatformdriver.cpp
```diff
@@ -65,12 +65,12 @@ const char *AGSPlatformDriver::GetDiskWriteAccessTroubleshootingText()
...
 
 void AGSPlatformDriver::GetSystemTime(ScriptDateTime *sdt) {
-    struct tm *newtime;
-    time_t long_time;
+    time_t t = time(nullptr);
 
-    time( &long_time );
-    newtime = localtime( &long_time );
+    //note: subject to year 2038 problem due to shoving time_t in an integer
+    sdt->rawUnixTime = static_cast<int>(t);
```
ivan-mogilko: Note that we may change the internal representation (need to update save format though), but since AGS script only has 32-bit integers we would have to adjust API somehow to let users get full value.

PR #735, opened 2019-04-23
AGS3 Porting fixes

sonneveld:
A few changes here and there to help with porting to other systems.  The aim is to have less platform specific `#ifdefs` in headers, it means errors due to platform differences are localised to specific .cpp files. (that you can replace or add to)

string refactoring: don't use #defines to replace string functions but instead introduce `ags_*` string functions.

stdio/io: Similarly, support multiple stdio implementation in specific functions instead of using #defines in headers.  Also renames some directory and file functions that were conflicting with Windows names. (we were specifically #undef-ing those functions to work around it)

platform fixups: replaces WIN|LINUX|MAC_VERSION with new `AGS_PLATFORM_OS_*` defines. Including `core/platform.h` is enough to get these definitions. I explicitly test if `AGS_PLATFORM_OS_` is true instead of testing if defined. That way I can get an error if I forget to include `platform.h`.   This should allow a bit less fiddling with compiler definitions and a single source to modify to specify platforms. I also continued to remove psp specific code.

entry_point: Other platforms might have a different `main` method and we might not be able to use or want to use hacks such as redefining `main`.  This renames `main` to `ags_entrypoint` and adds a new simple allegro specific main (to be replaced with an SDL specific one later)

(I've tested by merging this into the new multi cam view port branch and there were no conflicts)


Engine/platform/windows/acplwin.cpp
```diff
@@ -133,6 +125,14 @@ struct AGSWin32 : AGSPlatformDriver {
   virtual bool LockMouseToWindow();
   virtual void UnlockMouse();
 
+#ifndef AGS_NO_VIDEO_PLAYER
+  virtual void PlayVideo(const char* name, int skip, int flags);
+  virtual void PauseApplication();
```
ivan-mogilko: Please don't do this. These functions have very certain meaning, not related to video.

Common/util/directory.h
```diff
@@ -35,11 +28,11 @@ namespace Common
 namespace Directory
 {
     // Creates new directory (if it does not exist)
-    bool   CreateDirectory(const String &path);
+    bool   EnsureDirectoryExists(const String &path);
```
ivan-mogilko: Is this really necessary? We do abide the convention by placing our functions in our own namespace. I'd hate to force inventing different names just because someone else did not abide this convention and declared macros of commonly used names.

Additionally, I believed it may be possible to reduce places where windows headers are included to reduce (if not completely prevent) the conflicts.

ivan-mogilko:
> Also renames some directory and file functions that were conflicting with Windows names. (we were specifically #undef-ing those functions to work around it)

I can honestly tell that I am *very much* against this. I do not want our function names, which are placed in our own namespace, to be enforced by someone elses habit of declaring macros of commonly used names.

sonneveld:
> I do not want our function names, which are placed in our own namespace, to be restricted by someone elses habit of declaring macros of commonly used names.

Wouldn't it be easier to rename rather than always worry about naming conflicts? We already use Windows' `CreateFile` in `namedpipeagsdebugger` but also have to #undef `CreateFile` in other headers. It would reduce confusion if we just renamed the function.

Engine/platform/windows/acplwin.cpp
sonneveld: Fair enough. I checked before hand and the only places these functions were implemented was for video playback. I figured we could break them out of the #ifdef block if we used them for more than just managing video.

ivan-mogilko:
> We already use Windows' CreateFile in namedpipeagsdebugger

IMO this is a mistake to use these macros in modern code, because they were mainly invented for backwards compatibility, their meaning is determined by compilation flags and may not always match our intentions. We should use either CreateFileA or CreateFileW instead, depending on the string argument type.

sonneveld:
> We should use either CreateFileA or CreateFileW instead, depending on the string argument type.

I agree. But we will still have to deal with the effects of those compatibility macros.

ivan-mogilko:
> But we will still have to deal with the effects of those compatibility macros.

This is true, and I find frustrating the idea to live with someone's macros unwantedly and implicitly affecting our code.

Which code is actually affected by these? I'd wish we try to limit where windows headers are included instead first and either undef conflicting macros on per-case basis or make some general header to include.

ivan-mogilko:
On a separate note, I realized that other platform drivers don't have "override" qualifiers yet.

sonneveld:
Ok, I've modified this to remove any rename of the conflicting windows names, instead #udef'ing like we did before.

Issue #734, opened 2019-04-22
Introduce uniform audio driver string ID setting in config


ivan-mogilko:
Don't know how that happened, but historically there are multiple config options that define audio driver ID. Windows version has its own setting that works as an index in hard-coded list of drivers (pretty non-obvious), while other platforms have option that stores actual driver ID but *packed into 32-bit integer* which makes it very inconvenient to set by hand.

The task is following:
1a. Introduce a new config options for digital and MIDI driver ID, set up as a string.
1b. Alternatively, we may use existing options but then we have to test if value is a number or string, and convert number to string (see below).
2. Use these as a primary options on all platforms *including Windows*.
3. For backwards compatibility, read old options and convert to new ones (see: https://github.com/adventuregamestudio/ags/blob/ags3/Engine/main/config.cpp#L284).
4. Adjust winsetup to use new options (I can handle this myself separately, in case someone else will work on this fix and won't know how to change winsetup code).

NOTE: we already have a function that converts number to 4-char string: https://github.com/adventuregamestudio/ags/blob/ags3/Engine/main/engine.cpp#L436
We may need to extract that to common utility file and add a function that does reverse conversion.

ivan-mogilko:
Done: #737.

PR #737, opened 2019-04-25
Uniform audio config option which supports string driver ID

ivan-mogilko:
After #734.

This lets engine use "digiid" and "midiid" config options on all platforms including Windows and deprecates Windows-only "digiwinindx" and "midiwinindx" options.

These two options are now tried as both number and string, letting user to define audio driver ID as a string for convenience, e.g.:
<pre>
[sound]
digiid=ALSA
midiid=AMID
</pre>
of course you have to know driver names from Allegro for this.

Special names: "auto" for getting default driver and "none" to disable sound.

~**NOTE:** I made these names case-sensitive by oversight but they probably should not be, so I'll fix that.~ should be fixed now

Additionally, I adjusted the initialization step; now it does not try to initialize sound blindly by requesting as much voices as engine likes, but instead queries available driver's capability first and adjusts request accordingly. This should prevent confusing initialization errors when driver reports "unsufficient digital voices", and removes the need to retry installing sound with different voice number settings.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1514,9 +1514,6 @@ namespace AGS.Editor
             bool use_default_midi = _game.DefaultSetup.MidiSound == RuntimeAudioDriver.Default;
             NativeProxy.WritePrivateProfileString("sound", "digiid", use_default_digi ? "-1" : "0", configFilePath);
             NativeProxy.WritePrivateProfileString("sound", "midiid", use_default_midi ? "-1" : "0", configFilePath);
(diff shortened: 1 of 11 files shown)


rofl0r:
just so you know, with alsa one can select one of multiple devices with a device string, the default being "default" but there's also an ["interface:card,device" method](https://alsa.opensrc.org/Device).

ivan-mogilko:
Unfortunately it's not the actual device name for ALSA, it's just the internal allegro's driver ID.
In Allegro the device itself can be read from allegro.cfg.
In theory we could also read alsa device from our config separately and pass into allegro before initialization.

PR #744, opened 2019-05-05
Further separating Win-specific code from the general modules

ivan-mogilko:
So, the idea is to further move platform-specific code away from general engine modules, and particularily - avoid to include <windows.h> whenever possible (esp in common headers), because it adds too much unnecessary stuff to global namespace.

Hiding code in cpp files was enough for most parts, but there was one special case with Library class which has winapi type in its declaration. For that case I created a small header with an excrept from windows.h which only declares necessary types. Also, because Library has inlined functions, I declared few winapi functions right in its header rather than including windows.h. TBH I don't think Library class benefits from inlining much. Earlier Windows Mutex and Thread classes had similar problem but they are now implemented using functionality from std:: . Anyway, if either of them, or any new class, will have such issue again, then same solution could be used.


sonneveld:
Fiddly work but appreciated!

PR #748, opened 2019-05-06
Editor: custom replacements for Say and Display in dialog scripts

ivan-mogilko:
This is perhaps the easiest possible solution for using custom speech functions in dialog scripts.

In AGS 3.* dialog scripts are converted to regular ags script by substituting dialog-specific syntax with script commands, for example:
`CHARNAME: Sais something`
gets converted into:
`cCHARNAME.Say("Sais something");`

Obviously this hard-coded behavior was not compatible with custom speech functions many people do, forcing them to use regular functions in dialog scripts as well, which is often inconvenient.

The proposed change introduces two properties in General Settings: DialogScriptSayFunction and DialogScriptNarrateFunction. The values of these properties are simply substituting previously hard-coded "Say" and "Display" commands during dialog script compilation.

Additionally "Say" checkbox is translated to the `player.SayFunction(dDialog.GetOptionText(n));` in script.
GetOptionText is used in case option text could be altered at runtime at some future point.

### DIFF
--- a/Editor/AGS.Editor/DataFileWriter.cs
+++ b/Editor/AGS.Editor/DataFileWriter.cs
@@ -1610,7 +1610,9 @@ namespace AGS.Editor
                 {
                     DialogOption option = curDialog.Options[i];
                     int flags = 0;
-                    if (!option.Say) flags |= NativeConstants.DFLG_NOREPEAT;
+                    // NOTE: we always force "no-say" flag, because "say" checkbox is processed when
+                    // the dialog script is converted into regular script now.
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
Ah, soon after merging I found out that there is Dialog.DisplayOptions function that may demand dialog options to ignore "say" flag. So the way I made "say" flag convert into actual script command won't work.

Need to redo this, and actually implement calling custom function in the engine. Which in turn means that game data format has to contain it.

PR #749, opened 2019-05-06
AGS3 Streams refactor

sonneveld:
Initially this was to add a buffered file stream but I thought maybe I could take the opportunity to also implement a TODO which was to introduce managed pointers for streams.

I refactored the core file and buffered streams into a simple interface `ICoreStream`. That made it easier to override just for buffering. The original Stream and AlignedStream classes still exist, wrapping these core streams to present the classic interface.

It's probably easier to review the first changeset as that has the new Stream code. The second changeset just adjusts everything else to use the shared pointers.

WIP: I haven't tested on Windows yet, and there's some things with error handling to fix.


sonneveld:
Constructive criticism please @rofl0r

ivan-mogilko:
Is there a reason for doing refactor like this in 3.5.0 which is about to be released soon?

I did not checked much about the classes themselves yet, but I can already say that I don't like passing shared pointers into the functions like ReadFromFile. There was a good advice I read somewhere (unfortunately forgot the source) to not pass shared pointers into the functions which are not going to store these as shared pointers themselves.

ivan-mogilko:
Additionally, I very strongly propose to use typedefs instead of std::shared_ptr<AGS::Common::Stream> everywhere.

sonneveld:
> Is there a reason for doing refactor like this for 3.5.0 milestone which is long overdue anyway?

Good point, I'll move into a new milestone like 3.6.0.

> There was a good advice I read somewhere (unfortunately forgot the source) to not pass shared pointers into the functions which are not going to store these as shared pointers themselves, and I tend to agree.

Ideally with managed pointers, we don't expose the raw pointer to anything, otherwise we run the risk of dangling pointers. We should be using unique_ptr/shared_ptr instead of raw pointers in most places to manage ownership.  I was looking into just passing const references of shared_ptr, which would reduce the reference counting and _should_ allow us to assign to objects that have an ownership relationship.  Alternatively, we could use const references of unique_ptr everywhere, except where we know we're storing a second reference to it.

There's a good description here of our options https://docs.microsoft.com/en-us/cpp/cpp/how-to-create-and-use-shared-ptr-instances?view=vs-2019#example-5

> Additionally, I very strongly propose to use typedefs instead of std::shared_ptr < AGS::Common::Stream > everywhere.

I have thoughts on this too. :)  

I would like to propose a flat namespace at some point (like `::ags` and `::agstest`). We can still use classes with static methods if we want to group common functions.  I've been reading that nested namespaces have hidden gotchas, especially when you start to include other libraries with the same namespace as your sub-namespaces. ( http://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0816r0.pdf ) 

I'm not sure how I feel about hiding these types behind typedefs, has it makes it difficult to see where a type is being using globally. (unless we use it everywhere). And it may lead to just using blanket PTypename everywhere instead of considering ownership (shared vs unique). We could use namespace aliases like `namespace agsc = AGS::Common` to reduce it to `std::shared_ptr<agsc::Stream>` but I'm not sure if that's much better if you're looking to reduce length.

ivan-mogilko:
>  I was looking into just passing const references of shared_ptr, which would reduce the reference counting and should allow us to assign to objects that have an ownership relationship. Alternatively, we could use const references of unique_ptr everywhere, except where we know we're storing a second reference to it.

Why not a reference to an object then?

sonneveld:
> Why not a reference to an object then?

Yes that will definitely work, but then you'd have to be careful that the base unique_ptr doesn't change. Consider global unique_ptrs or unique_ptrs stored in the asset manager.  We could dereference the ptr and then pass by reference, but if any of the functions down the call chain reset those unique_ptrs (maybe we load a new asset library), those reference are now effectively pointing to deallocated memory.

If you're passing references, then you'd need to ensure you're the one who owns that pointer, or you use shared pointers and create your own copy with an increased reference count:

```
void blah(T & x) {}
...
auto my_ptr = load_asset()
blah(*my_ptr);
my_ptr->somethingelse();
```

ivan-mogilko:
> We could dereference the ptr and then pass by reference, but if any of the functions down the call chain reset those unique_ptrs (maybe we load a new asset library), those reference are now effectively pointing to deallocated memory.

This suggests some overcomplicated functions that do too much, or bad program design.
I am talking about simple functions that do nothing but read X values from a stream into a struct, or write X values from a struct into a stream. They may (or rather must) be written so that they don't change anything else besides what they are intended to.

ivan-mogilko:
Regarding the stream classes refactor, finally took a closer look and this seems very close to something I wanted to suggest long time ago (divide stream hierarchy into types that do raw reading and writing and wrapping types that do additional processing) so I am all good with that.

Except maybe I'd propose to separate ICoreStream and Stream types in two headers at least, but that's my opinion.

I think only memory stream is missing there now.

PR #755, opened 2019-05-08
Fixes to increase compatibility with GCC 8

onitake:
GCC 8 is much more picky with sprintf and snprintf and fails compilation when the return value remains unchecked.
```c++
script/script.cpp: In function ‘char* make_ts_func_name(const char*, int, int)’:
script/script.cpp:515:40: error: ‘_’ directive output may be truncated writing 1 byte into a region of size between 0 and 60 [-Werror=format-truncation=]
     snprintf(bne,MAX_FUNCTION_NAME_LEN,"%s_%c",bname,subd+'a');
                                        ^~~~~~~
script/script.cpp:515:13: note: ‘snprintf’ output between 3 and 63 bytes into a destination of size 60
     snprintf(bne,MAX_FUNCTION_NAME_LEN,"%s_%c",bname,subd+'a');
     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This used to be a warning only, but with `-Werror`, it causes failure.

Surprisingly, GCC is able to figure out when a static buffer is passed to sprintf - and it range-checks the result automatically. This prompts the aforementioned warning when the return value isn't checked.

I added the necessary checks and replaced sprintf with snprintf. This could cause compatibility issues with Windows, I can change it back to sprintf if requested.

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -27,6 +27,7 @@
 #include "util/alignedstream.h"
 #include "util/path.h"
 #include "util/string_utils.h"
+#include "debug/debug_log.h"
 
 namespace AGS
 {
@@ -403,16 +404,24 @@ void BuildAudioClipArray(const AssetLibInfo &lib, std::vector<ScriptAudioClip> &
             ScriptAudioClip &clip = audioclips.back();
             if (ags_stricmp(temp_name, "music") == 0)
             {
-                sprintf(clip.scriptName, "aMusic%d", temp_number);
-                sprintf(clip.fileName, "music%d.%s", temp_number, temp_extension);
+                int err = snprintf(clip.scriptName, sizeof(clip.scriptName), "aMusic%d", temp_number);
+                if (err >= sizeof(clip.scriptName))
+                    debug_script_warn("Script name length exceeded: %d", err);
+                err = snprintf(clip.fileName, sizeof(clip.fileName), "music%d.%s", temp_number, temp_extension);
+                if (err >= sizeof(clip.fileName))
+                    debug_script_warn("Clip file name length exceeded: %d", err);
                 clip.bundlingType = (ags_stricmp(temp_extension, "mid") == 0) ? AUCL_BUNDLE_EXE : AUCL_BUNDLE_VOX;
                 clip.type = 2;
                 clip.defaultRepeat = 1;
             }
             else if (ags_stricmp(temp_name, "sound") == 0)
             {
-                sprintf(clip.scriptName, "aSound%d", temp_number);
-                sprintf(clip.fileName, "sound%d.%s", temp_number, temp_extension);
+                int err = snprintf(clip.scriptName, sizeof(clip.scriptName), "aSound%d", temp_number);
+                if (err >= sizeof(clip.scriptName))
+                    debug_script_warn("Script name length exceeded: %d", err);
+                err = snprintf(clip.fileName, sizeof(clip.fileName), "sound%d.%s", temp_number, temp_extension);
+                if (err >= sizeof(clip.fileName))
+                    debug_script_warn("Clip file name length exceeded: %d", err);
                 clip.bundlingType = AUCL_BUNDLE_EXE;
                 clip.type = 3;
             }
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -2101,11 +2101,14 @@ void draw_fps()
 
     char fps_buffer[60];
     // Don't display fps if we don't have enough information (because loop count was just reset)
+    int err;
     if (!std::isnan(fps)) {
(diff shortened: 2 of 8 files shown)


ivan-mogilko:
I see few problems with this pr.

First of all in certain cases that would be simplier (and maybe beneficial) to replace char buffer with a String class or ensure that buffer will always have sufficient length when it can be determined, like in case with drawing fps where me may know beforehand max values.

Second, debug_script_warn is meant to report game mistakes, and not all of the cases of buffer exceeding are related to what game is reposible for, like in same case with fps it's more of an internal engine limitation (or mistake).

Finally, the reported warnings do not have much use, since all they say is snprintf result, but they don't explain e.g. which game object this is related to or which name exactly is too long.

onitake:
Thanks for the clarifications about what debug_script_warn does.

I absolutely understand what your point is, and my opinion on this matter is clear: Path concatenation should _never_ be handled by a limited sprintf buffer, especially on w32, where PATH_MAX is only 260 characters. Strings should be formatted using C++ strings or a custom string class, and only converted when calling into legacy C APIs.

However, the focus of this PR, and why I'm targeting against the `ags3` branch is that I wanted to have a quick fix to get AGS 3.5 to compile with GCC 8.

A more permanent fix should replace the sprintfs altogether and target AGS 4. I'm not sure where development stands right now, but I would assume that one of the tasks is to replace all occurrences of sprintf with proper strings.

ivan-mogilko:
@onitake could you explain, what kind of minimal "return value check" gcc require to supress the warning?
What may I replace the logging with if I do not want to log anything in some case?

ivan-mogilko:
I made a mistake merging this PR, because debug_script_warn is not accessible in the Common library.

onitake:
Yes... I was surprised that you merged it as-is.
I'm pretty sure that switching from sprintf to snprintf is a problem on at least msvrt.
I think you should revert the commit, and I'll open a new PR for discussion.

So, to explain the issue a bit further:

Path handling has been relatively poor in AGS, using fixed-length stack buffers to hold composited paths. These buffers were defined as having a length of 260 bytes in some places and PATH_MAX in others. On classic win32, PATH_MAX is 260 bytes, but most POSIX platforms define the value as 1024, 2048 or 4096 bytes or more, and there's no clear definition on how long it should really be. More recently, Microsoft has lifted the restriction, and even older win32 APIs actually support 32K characters long paths. More info here: https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file

So, as you said, the proper way to handle this is to composite a path using regular C++ strings or a custom path type, then convert, limit and warn when an incompatible API is called.

Now, GCC 8 smartly detects static-length buffers and behaves like if snprintf was called with the length of the buffer as the second argument. This causes GCC to do range checking on the format string and all arguments, if possible.
If it detects that the passed buffer may overflow, it kicks off a second check that verifies if the program checks the return value (which is the number of characters that would be written), and bails out if no such check is done in the program.
Since all such instances of sprintf in AGS simply ignore the return value, a warning is triggered.
This warning is then turned into an error, when compiling with more strict GCC arguments - as it happens on the ags3 branch.

ivan-mogilko:
I merged this, having an impression that you do not want to modify the code further:

> However, the focus of this PR, and why I'm targeting against the ags3 branch is that I wanted to have a quick fix to get AGS 3.5 to compile with GCC 8.

After merging I made several commits modifying this further:
7ca179c565dbd8da19d4afeb808d84384bff05ab
79b7c4a7bda0935fa77253f342859a3a95f69b42
8d3ca41adbdd1269fb539697f116c381ad1ff23b
fb4b6af416554009f28ca381d8084f6de4cae302

onitake:
Oh, that's good!

What's the deal with 79b7c4a7bda0935fa77253f342859a3a95f69b42 though?
You just made the `base_buffer` smaller and removed the return value check?
I don't think that's the right solution...
When someone changes `base_buffer` or the message later, the overflow will be back.
Can't you use a `String` or a C++ `std::string` here there, like in 8d3ca41adbdd1269fb539697f116c381ad1ff23b ?

onitake:
As for your point about `debug_script_warn`: I see... I didn't notice that this is engine-specific. My assumption was that I had only touched engine code.

Logging functions are always good, though. Perhaps this is something that the editor should offer as well?

ivan-mogilko:
> What's the deal with 79b7c4a though?
You just made the base_buffer smaller and removed the return value check?
I don't think that's the right solution...
When someone changes base_buffer or the message later, the overflow will be back.
Can't you use a String or a C++ std::string here there, like in 8d3ca41 ?

Yes string may be used there but I thought it may work for now; and since that function is called often we probably should keep the allocated String at all time. Could be I was overthinking again. Whole thing could be refactored into some kind of a "misc info drawing" object later.

PR #756, opened 2019-05-08
AGS3 Buffered streams

sonneveld:
Add a tunable buffered stream to help with systems that have slow seek/low read buffers.

This implementation is based on top of FileStream, so not as reusable. It will also attempt to re-use the buffer after a Seek.

### DIFF
--- a/Common/util/bufferedstream.cpp
+++ b/Common/util/bufferedstream.cpp
@@ -83,7 +83,7 @@ size_t BufferedStream::Read(void *toBuffer, size_t toSize) {
         toSize -= chunkSize;
     }
 
-    return to - toBuffer;
+    return to - (char*)toBuffer;
 }
 
 int32_t BufferedStream::ReadByte() {
--- a/Common/util/bufferedstream.cpp
+++ b/Common/util/bufferedstream.cpp
@@ -66,12 +66,14 @@ size_t BufferedStream::Read(void *toBuffer, size_t toSize) {
     while(toSize > 0) {
         if (_position < _bufferPosition || _position >= _bufferPosition + _buffer.size()) {
             FillBufferFromPosition(_position);
-            if (_buffer.size() <= 0) { break; } // reached EOS
         }
+        if (_buffer.size() <= 0) { break; } // reached EOS
+        assert(_position >= _bufferPosition && _position < _bufferPosition + _buffer.size());  // sanity check only, should be checked by above.
 
-        auto bufferOffset = _position - _bufferPosition;
-        auto bytesLeft = _buffer.size() - bufferOffset;
-        auto chunkSize = std::min<size_t>(bytesLeft, toSize);
+        soff_t bufferOffset = _position - _bufferPosition;
+        assert(bufferOffset >= 0);
+        size_t bytesLeft = _buffer.size() - (size_t)bufferOffset;
+        size_t chunkSize = std::min<size_t>(bytesLeft, toSize);
 
         std::memcpy(to, _buffer.data()+bufferOffset, chunkSize);
 
--- a/Common/util/bufferedstream.cpp
+++ b/Common/util/bufferedstream.cpp
@@ -26,60 +26,57 @@ namespace AGS {
 namespace Common {
 
 BufferedStream::BufferedStream(const String &file_name, FileOpenMode open_mode, FileWorkMode work_mode, DataEndianess stream_endianess)
-        : FileStream(file_name, open_mode, work_mode, stream_endianess), buffer_(BufferStreamSize), bufferPosition_(0), position_(0)
+        : FileStream(file_name, open_mode, work_mode, stream_endianess), _buffer(BufferStreamSize), _bufferPosition(0), _position(0)
 {
-    if (FileStream::Seek(0, kSeekEnd) == false) {
+    if (FileStream::Seek(0, kSeekEnd) == false)
         throw std::runtime_error("Error determining stream end.");
-    }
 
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
I don't know if we should use exceptions there. AGS is not very well suited for exceptions at the moment because there is only 1 common place where it catches them - around the engine start. Maybe it would be better for now to indicate that stream is invalid, using IsValid method?

sonneveld:
> I don't know if we should use exceptions there.

I ensured that we catch those exceptions where the FileStream is created. It will have the same behavior as before.

ivan-mogilko:
But the object of this class could be created anywhere, not necessarily by the function where you guarded it?

ivan-mogilko:
Actually, I am more curious about whether this implementation improves streams in general on some systems, or meant to improve it in particular situations?

sonneveld:
> Actually, I am more curious about whether this implementation improves streams in general on some systems, or meant to improve it in particular situations?

It does. On a particular system with a particular game I get a frame rate increase from 2fps to at least 60. It has an effect on normal desktop systems as well but not as much.

> But the object of this class could be created anywhere, not necessarily by the function where you guarded it?

But not yet, and if necessary we can document the exceptions. We already have to deal with exceptions coming from some of the graphics drivers.

ivan-mogilko:
> But not yet, and if necessary we can document the exceptions. 

Yes, please.

Common/util/bufferedstream.cpp
```diff
@@ -0,0 +1,125 @@
...
+    uint8_t ch;
+    auto bytesRead = Read(&ch, 1);
+    if (bytesRead != 1) { return EOF; }
+    return ch;
+}
+
+size_t BufferedStream::Write(const void *buffer, size_t size) { 
+    FileStream::Seek(position_, kSeekBegin);
+    auto sz = FileStream::Write(buffer, size);
+	position_ += sz;
```
ivan-mogilko: The code style here is contradicting (different identation for instance), and also not compying to our existing code style convention:
https://github.com/adventuregamestudio/ags/wiki/AGS-Coding-Conventions-(Cpp)

Are you planning on moving towards this codestyle from now on?

sonneveld: I started doing this in places because I was getting style warnings. Identifiers that start with `_` are reserved in C++ in certain circumstances. 

But you're right, I'll review the code style again and try to stick with it. I'm not suggesting we run a style formatter tool over the whole project, but it would be useful for new code.

sonneveld:
> > But not yet, and if necessary we can document the exceptions.
> Yes, please.

I was thinking about this. The reason why I prefer exceptions in the constructor is because if the constructor succeeds, you should have a working object. Exceptions seem to be the only way to communicate to `new` and callers that the construction failed. `IsValid` method works but it requires the caller to check that and then delete the object if it failed.

Would it be better to make the constructors private and have a static factory method here instead?

Common/util/bufferedstream.cpp
ivan-mogilko: Regarding underscore, we checked this back in 2012, and we found that having class member variables starting with underscore should be okay.

From https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

> Reserved in any scope, including for use as implementation macros:
identifiers beginning with an underscore followed immediately by an uppercase letter
identifiers containing adjacent underscores (or "double underscore")
Reserved in the global namespace:
identifiers beginning with an underscore

ivan-mogilko: What style warnings were you getting? I've seen quite a few programs that use this naming style.

ivan-mogilko: If that would be a real issue we may change it of course, but may need to work out a convenient naming style for protected class fields.

sonneveld: Hah I just read that same page. But just the fact that prepended underscores are reserved in some situations suggests we should avoid that for consistency.

The warnings were when using clang and turning all warnings on. It specifically complained about our `__` at the start of our header guards for example.

It's all good, I was trying a thing, I can revert in this pull request and future changes.

ivan-mogilko:
> Would it be better to make the constructors private and have a static factory method here instead?

I did something like this to Bitmap class, adding BitmapHelpers namespace with factory methods, but bitmaps are something that are created often.

In case of sreams there's currently only one place where file stream is created so you may be right about that, and this may be sufficient until further refactoring.

I am simply asking to either mention the use of exceptions in comments, or use "throw()" after constructor (but later apparently is deprecated in C++11, so maybe there are other means now).

Common/util/bufferedstream.cpp
ivan-mogilko: > The warnings were when using clang and turning all warnings on. It specifically complained about our __ at the start of our header guards for example.

I did not realize that. We may change that removing leading underscores.

ivan-mogilko:
I keep doing same mistake merging prs without testing all the possible cases. This time I only tested that this buffered stream is reading properly, which it does, but for some reason did not realize it also is used for writing.

Writing was broken, I fixed couple of mistakes (d10a2d3353b9640ea2b3ced45ed4757fbf53e081), but there's still at least 1 inconsistency between buffered stream and filestream logic: when I open existing file for writing in "append" mode the buffered stream positions cursor at the beginning of the stream (file stream supposed to have it at the end). Since this breaks game compilation I had to temporarily switch back to creating simple filestreams instead.

PR #759, opened 2019-05-09
gitignore: Split into directory specific gitignore files where it makes sense

sonneveld:


ivan-mogilko:
*.chm should be added to Manual dir.

Manual/.gitignore
```diff
@@ -1,11 +1,5 @@
...
 *.con
-ags.ref
-ags.chm
-ags.chw
-ags.hhc
-ags.hhk
-ags.hhp
-ags.htx
-ags-help.chm
+ags.*
```
ivan-mogilko: I don't think this is correct, perhaps better filter out by extensions rather than "ags" prefix.
Also ags-help.chm resurfaces because of this.

ivan-mogilko:
Actually, I think it's about the time we remove Manual directory from the repository and use https://github.com/adventuregamestudio/ags-manual instead.
I just need to double check to see if it produces a good chm because I did not find time to earlier.

Issue #697, opened 2019-03-29
Automatically create all nested subdirectories when user opens file for writing in script


ivan-mogilko:
Strangely, but for some reason AGS does not cover this case when user wants to have subdirectories in $APPDATADIR$ and $SAVEGAMEDIR$. I don't see any reasons not to.
This should not be hard to fix too (need to write an utility function that creates all missing subdirectories along the path, starting from base dir).

morganwillcock:
Would this just be for `File.Open` with the mode set to `eFileWrite`, or something more generic?

ivan-mogilko:
> Would this just be for File.Open with the mode set to eFileWrite, or something more generic?

Yes, it should be done when File.Open is called with eFileWrite to ensure it can create a file.

ivan-mogilko:
Done by #764

PR #764, opened 2019-05-12
Create all subdirectories when writing a file by script command

ivan-mogilko:
For #697.

This ensures that when the script commands to open a file for writing in a special game location all subdirectories are created beforehand.
Engine ofcourse must test that these subdirectories are nested inside the allowed location.

Affected API:
* File.Open (write only)
* DynamicSprite.SaveToFile
* Game.SetSaveGameDirectory

### DIFF
--- a/Common/util/directory.cpp
+++ b/Common/util/directory.cpp
@@ -9,6 +9,7 @@
 #endif
 #include "util/directory.h"
 #include "util/path.h"
+#include "stdio_compat.h"
 
 namespace AGS
 {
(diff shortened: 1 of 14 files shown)


i30817:
Could you not allow the `..` relative path? For obvious security reasons.

ivan-mogilko:
@i30817 
> Could you not allow the .. relative path? For obvious security reasons.

This was already restricted before, and still will be of course.

i30817:
ok, thanks.

PR #770, opened 2019-05-14
3.5.0 Android fixes

ivan-mogilko:
Fixing Android build. Most things are trivial, but one I'd like to note is that apparently the older NDK version we suggest building against _does not have std::lround implemented_, even though they support C++11 in general.
Could be related: https://github.com/android-ndk/ndk/issues/82

So I went on to replace all the std::lround in our code to lround from C library.

### DIFF
--- a/Android/library/jni/Android.mk
+++ b/Android/library/jni/Android.mk
@@ -27,7 +27,7 @@ include $(BUILD_SHARED_LIBRARY)
 include $(CLEAR_VARS)
 
 LOCAL_MODULE    := pe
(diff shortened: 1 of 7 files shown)


sonneveld:
TIL about the older NDK's support of std:: lib. What do we lose if we suggest a newer NDK?

Otherwise the changes look good.

ivan-mogilko:
> What do we lose if we suggest a newer NDK?

Older devices afaik. We bumped version of NDK in either 3.4.1 or 3.5.0 to mimimal supporting what we need. This may be bumped again later, after 3.5.0, because right now that does not seem to be necessary.

Issue #201, opened 2014-10-25
Sprite cache works unstable if max cache size is set to a very low value


ivan-mogilko:
If max sprite cache size is set to a very low value (bug first detected at 20 KB), engine sometimes throw unexpected errors during gameplay, like "Attempted to remove sprite N that does not exist".

Normally, if the defined size is too low, the program should either increase it to lowest size that would let it run, or quit with error message that would properly describe the reason of a problem.

On other hand, there may be just some bug in business logic that makes effect only if cache size is not enough.

Known way to reproduce:
1. Download "Ben Jordan 3" game: [link](http://www.grundislavgames.com/benjordan/case3.php)
2. Manually edit acsetup.cfg, under [misc] section put: "cachemax = 20"
3. Run "Ben Jordan 3" and start new game.
4. Wait until the first cutscene is finished, or skip it with Esc.
5. Open inventory, close inventory, wait few seconds or make character walk in various directions (this makes engine load more animation sprites).
6. Error occurs:

<pre>Error: SpriteCache::removeOldest: Attempted to remove sprite 2041 that does not exist</pre>

ivan-mogilko:
Fixed by #776

PR #776, opened 2019-05-20
3.5.0: fix Sprite Cache after faulty refactor

### LINKED ISSUE #773: New game 'Guard Duty' might expose a  memory leak
So this game has the habit of crashing with things simillar to:

`Error: SpriteCache::removeOldest: Attempted to remove sprite 879 that was either remapped or added externally`

After either some complex graphical effects or some anims.

I've seen 877, 878 too. Anyway the fun thing is that if you restart and reload a game save from later than you started the last time you're probably going to get further. I could just keep increasing the memory with 
```
[misc]
cachemax=...
```
until it works from start to end but I'm unsure that this is how it's supposed to work? I'm up to 300mb right now (307200).

[Steam says it's supposed to use 64mb of memory](https://store.steampowered.com/app/872750/Guard_Duty/) and the default on opensource ags is 128mb. Anyway, i found it suspicious, because it looks like a leak.

ivan-mogilko:
Fixes #773

This is to fix broken SpriteCache logic after commit 238a983 and clarify some things in the class.

This all started when I planned to refactor the class but stopped at some point and decided to push one commit which seemed to be pretty simple, yet broke certain logic inside the cache.

To elaborate a bit, there were certain inconsistencies in how sprites are described internally when dynamic sprites are added to the cache at runtime, but old program logic luckily ignored them. With commit 238a983 I wanted to introduce stricter rules in how sprite state is defined and that revealed this old problem. Also I missed couple of cases when a flag should be (re)set but was not.

In the end dynamic sprites got assigned to the MRU (most-recent-use) list along with the cached data, and then engine tried to delete them to free some space, which caused logic error.

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -183,7 +183,7 @@ sprkey_t SpriteCache::EnlargeTo(sprkey_t topmost)
     return topmost;
 }
 
-sprkey_t SpriteCache::AddNewSprite()
+sprkey_t SpriteCache::GetFreeIndex()
 {
     for (size_t i = MIN_SPRITE_INDEX; i < _spriteData.size(); ++i)
     {
@@ -277,8 +277,7 @@ Bitmap *SpriteCache::operator [] (sprkey_t index)
     return _spriteData[index].Image;
 }
 
-// Remove the oldest cache element
-void SpriteCache::RemoveOldest()
+void SpriteCache::DisposeOldest()
 {
     if (_liststart < 0)
         return;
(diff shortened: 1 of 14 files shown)


i30817:
This appeared to fix my save 'test'. I'll be more sure if i replay the whole game in a single sitting thou.

i30817:
Ok, i played through the game with no problems.

ivan-mogilko:
Good to hear!
I found one problem, but it's related to some very old game, investigating now.

ivan-mogilko:
Alright, this PR apparently also fixes the old bug #201.

And at the same time I found there was a hilarous bug in old code which was exploited by some games, like one mentioned in #201. I commented on that: https://github.com/adventuregamestudio/ags/pull/776/commits/f3422175512b551a2e7979c0c5a83f63fed3c7ac

PR #777, opened 2019-05-21
ags3/engine/timing: Introduce a single point of delay after rendering.

sonneveld:
In case we updated and rendered within frame time budget. This should only be useful if vsync is disabled or running in windowed mode (since vsync is always disabled then)


i30817:
This lowers cpu load (on my underclocked cpu) on 'regular' play (not menu, not inventory), windowed and on Wayland (so no vsync - i suspect it doesn't actually exist on Wayland anyway because there is no fullscreen exclusive mode), from 86% to 41%.

ivan-mogilko:
@i30817 could you please compare these results with a build prior to commit f2ab428 ?

i30817:
> @i30817 could you please compare these results with a build prior to commit f2ab428 ?
Ok, numbers on Wayland (so no vsync, fake fullscreen) on the game Guard Duty, my shitty underclocked computer, `top -S` (cumulative mode):

1. this pr main menu: 25% +/- 1%
2. this pr main menu with the 'do you want to really quit' dialog open: 34.8% +/- 0.4%
3. this pr after 'new game' with all the animations and cutscenes done with ego doing nothing on the first playable scene: 27% +/- 0.4%

1. commit f2ab428~1 main menu: 17% +/- 0.5%
2. f2ab428~1 main menu with the 'do you want to really quit' dialog open: 15% +/- 0.4% (!!!)
3. f2ab428~1 after 'new game' with all the animations and cutscenes done with ego doing nothing on the first playable scene: 17% +/- 0.5%

So this is much better than now but still not as good as before that commit.

> Do you mean that menu or inventory still produce high CPU load?
Not exactly, it appears to jump 10% or so if any 'frame' appears on top of the game in any circunstance. For instance Guard's duty main menu with this PR, which has a cloud animation jumps from a average 25% to 35%. This seems bearable and understandable to me, if it's another event loop.

i30817:
BTW, i probably  should redo the measurements of this pr to use make instead of cmake to build ags because the older commit couldn't use cmake so i used make, making it a comparison of a cmake build against make. I'll go and do that.

sonneveld:
Quick question, is this with a Release build? The debug adds a bit of overhead. 

I suspect the extra overhead when an overlay displays could be the software drawing of GUI components?

ivan-mogilko:
> I suspect the extra overhead when an overlay displays could be the software drawing of GUI components?

If I understood correctly from the comment above, additional cpu load was still added after same commit:

> this pr main menu with the 'do you want to really quit' dialog open: 34.8% +/- 0.4%

> f2ab428~1 main menu with the 'do you want to really quit' dialog open: 15% +/- 0.4% (!!!)

i30817:
Hardly, i forced OGL on the config for both.

sonneveld:
Re the software drawing comment. Even with open GL we are still software drawing bits of the GUI to a texture.

i30817:
It can also be xwayland interfering somehow but since both use xwayland as a intermediary with Wayland i doubt it. I do know that Retroarch (for instance) is absolutely miserable to run on XWayland if built without native wayland support, so it's likely both compiles have at least a 'hidden' fixed overhead on top in Wayland.

i30817:
Ok i tried to kcallgrind both executables and there is a major difference in drawing time. Not sure if it's useful but:

old:
old

new:
new

ivan-mogilko:
OGLGraphicsDriver::UpdateTextureRegion changed though:

f2936359958646a20f24316adc5d9c9d87179796
1c0e202bc5db8de35d05c2ab8ae3cb33b0138559

ivan-mogilko:
Alright, I can confirm that merging this PR drops general CPU load back to old values *on Windows*.
For example, Gemini Rue Demo title screen from 25% (full single core) to 6-8%.
I will test on Linux bit later when I have time.

Regarding remaining issues, they may be related to some other changes as well. I think it would be better to open a new ticket and make everyone to test same game and scene for consistency.

rofl0r:
i would still consider undoing the commit with the original regression for now. the graph posted by @i30817 shows that the control flow got a lot more complicated. after that, the original PR can be redone calmly in a way that doesn't add complexity nor performance regressions and with proper code reviews.

Engine/platform/windows/media/video/acwavi3d.cpp
```diff
@@ -129,9 +129,8 @@ int dxmedia_play_video_3d(const char* filename, IDirect3DDevice9 *device, bool u
 
     //device->Present(NULL, NULL, 0, NULL);
 
-		while (waitingForNextTick()) {
       update_polled_stuff_if_runtime();
-		}
+      platform->Delay(10);
```
ivan-mogilko: What's this "delay(10)", is this an arbitrary number of decided somehow?

sonneveld:
> the graph posted by @i30817 shows that the control flow got a lot more complicated.

That graphs seems to be a summarised call graph with heuristics to determine where to place nodes. I wouldn't use them to determine complexity.

> that doesn't add complexity

The timing before was not exactly simple. It was rooted in a thread updating a global variable along with the game thread resetting it in various places.

Engine/platform/windows/media/video/acwavi3d.cpp
sonneveld: That's a weird spot. Originally I used the same timing loops as used everywhere else. But the true drawing is done by dxmedia in, I presume, another thread. So it doesn't really matter how often we check for a key press.

sonneveld: So to answer your question..I just picked 10 :)

ivan-mogilko:
With the removal of **PollUntilNextFrame** and substituting it with **update_polled_stuff_if_runtime** there's now new difference. That function originally did this:
<pre>
void PollUntilNextFrame()
{
    if (play.fast_forward) { return; }
    while (waitingForNextTick()) {
        // make sure we poll, cos a low framerate (eg 5 fps) could stutter mp3 music
        update_polled_stuff_if_runtime();
    }
}
</pre>

Notice the `if (play.fast_forward) { return; }`. In this PR it does not do that.

I am getting a bad feeling that these changes may have implicit consequences that are hard to predict without reviewing whole program flow and all possible events.

sonneveld:
> Notice the `if (play.fast_forward) { return; }`. In this PR it does not do that.

I did notice this. :)  In draw.cpp, in `render_graphics`, there is also a `play.fast_forward` check that will skip the rendering and also the delay. With vsync, there was already a delay during rendering, so this puts it in one place.

ivan-mogilko:
This still means there's difference in how it is skipped.

sonneveld:
> It was not only checked before calling render.

But `PollUntilNextFrame`  was always called in the same loop as `render`, effectively introducing two delay functions with different timers.

ivan-mogilko:
What I mean, if you look at changes to DialogOptions::Run() for example:
1. update_polled_stuff_if_runtime was done in the end of the update, and checking if it's not fast-forward.
2. now it is done in the beginning of the update and not checking if it's fast-forward.

sonneveld:
All `update_polled_stuff_if_runtime` does it check for the quit flag, debug messages and pump the mp3 decoder. It's scattered randomly everywhere.

ivan-mogilko:
Okay, honestly, worst thing here is that I barely understand what these changes do, and don't have enough spare time to review them, being loaded with other things. Now I will have to release a new version, taking responsibility for it, while not knowing how the new code works...

sonneveld:
> taking responsibility for it,

That's not entirely true. I'm still here to fix any issues that come up.

ivan-mogilko:
Well, at this point I agree to anything only to release this version asap and quit.

rofl0r:
> not knowing how the new code works... I cannot spend more time looking at these pull requests, because my brain refuses to comprehend them anyway.

so maybe it'd be better to stick with the old code you (and others) understood for now, and introduce changes gradually ?

i30817:
My suspicion is that the slowdown now is limited to the path that goes down from `draw_gui_and_overlays()` from my kcallgrind messing around. If it is because it uses software now while it didn't before for some reason, seems like something reasonable to investigate; and this optimization could make things even better than before.


Generally, i think someone better than me at debugging and profiling should do the same, profile th the same scene on both PRs and see what's going much different, checking for 'self-time' (which is time spent on the function, not the time spent cumulatively on the function and calls) large differences between the two.

I'll mention i couldn't profile the debug executable (probably because i installed a different libstdc++ than what gcc version i'm using for some unrelated stuff and because i don't have opengl driver debug symbols); and that i couldn't use cg_diff (callgrind diff) directly for some reason, so someone with a better supported profiler, such as on visual studio might get better results.

ivan-mogilko:
> My suspicion is that the slowdown now is limited to the path that goes down from draw_gui_and_overlays() from my kcallgrind messing around. If it is because it uses software there now while it didn't before for some reason

TBH I dont remember any changes to GUI drawing done in past few months.

I am unable to see much difference with opengl without proper profiling, because opengl acts weirdly on Windows, or at least for me, I noticed it always shows 50% cpu for some reason, and this happens even running original 3.4.1 executable.

sonneveld:
> the original PR can be redone calmly in a way that doesn't add complexity nor performance regressions and with proper code reviews.

These changes have had open pull requests available for you to review. There was no additional complexity, you're just misreading a graph. The performance issues are fixed here. I don't understand what more I can do here. Nor do I understand why you are implying we're not calm?

> so maybe it'd be better to stick with the old code you (and others) understood for now, and introduce changes gradually ?

If we really want to do this, it might be better to wait for the 3.5 release branch and partially revert there. There are other changes that fix issues with delays that are tied to frame rate instead of time.

sonneveld:
I've rebased and refactored this PR to make it a bit clearer where the (many) game loops are and where the delays occur.  I'm hoping the comments and splitting it into individual commits help!

Engine/platform/windows/gfx/ali3dd3d.cpp
```diff
@@ -1815,15 +1815,17 @@ void D3DGraphicsDriver::do_fade(bool fadingOut, int speed, int targetColourRed,
...
   for (int a = 1; a < 255; a += speed)
   {
     d3db->SetTransparency(fadingOut ? a : (255 - a));
     this->_renderAndPresent(flipTypeLastTime, false);
 
-    do {
-      if (_pollingCallback)
-        _pollingCallback();
-    } while (waitingForNextTick());
+    // GAME LOOP DELAY has occured by this point.
```
ivan-mogilko: Is this correct? Direct3D _renderAndPresent does not do anything besides actually rendering textures.

Engine/platform/windows/gfx/ali3dd3d.cpp
```diff
@@ -1339,6 +1340,7 @@ void D3DGraphicsDriver::_renderAndPresent(GlobalFlipType flip, bool clearDrawLis
 {
   _render(flip, clearDrawListAfterwards);
   direct3ddevice->Present(NULL, NULL, NULL, NULL);
+  DelayUntilEndOfFrame();
```
ivan-mogilko: I honestly don't like this change, I don't think it's a proper task for this class to do delays.

ivan-mogilko: IMO it was already done wrong that there are fade loops inside the renderer classes, perhaps it would be better to "unwrap" these somehow instead.

sonneveld: I thought putting it there might make sense since there's already a delay for vsync. I can move the `DelayUntilEndofFrame`s outside the render function, there might just be a few more of them :)

I agree with the fade effects needing to be moved out of the graphics driver but I that's possibly for a different PR. It would be tricky to ensure it's correct across the different gfx drivers.

rofl0r:
> Nor do I understand why you are implying we're not calm?

what i meant was doing one small change at a time, with each one being a small PR that can be properly reviewed, rather than stuffing several loosely related changed into a single huge PR.

ivan-mogilko:
So, I guess we decided to open a "release-3.5.0" branch with a purpose of finalizing 3.5.0 there; and then try to revert this timer change there, because imho it has unknown consequences changing the critical engine code.
I don't mean it's necessarily wrong, but I'd like to make this release a bit safer.

sonneveld:
Please don't merge this until we consider the other timing PR.

i30817:
I'm glad you didn't give up. I think it's important work to make the engine fps independent, and would be willing to pay some performance for it so this PR was a step on the right direction, so i await your new solution.

sonneveld:
I cherry-picked a couple of one line changes into `ags3` which should help framerate.

sonneveld:
@i30817  thank you for all your testing btw! After going back and doing similar benchmarks, I know it can get a bit fiddly managing different versions of code.

PR #788, opened 2019-05-26
3.5.0: safety changes for Editor loading and saving sprite file

ivan-mogilko:
This pr attempts to address two potential problems with sprite file after which user may loose an ability to open project and/or loose some data.

1) The current order of actions, and error handling, when saving sprites is such that new data may get lost if there's a trivial file access error, and editor refuses to save it even if it failed to update a backup for example.
2) If the sprite file did not save correctly, next time you load the project it crashes because number of sprites in the sprite file does not match import references in the project itself.

In this pr problem 1) is resolved by saving new sprite file into temporary location provided by system API first, and replacing one in the project folder only after. Even if something fails at the second step, sprite cache *attaches itself to the temporary file* thus giving user ability to continue work inside Editor, try to resave, fix access problem, or just move temp file by hand to a safer place as a backup (it reports the name of the file too).

For the problem 2) editor creates a dummy sprite slot for every entry missing in the sprite file remapping to sprite 0. This should at least prevent the crashing. Ofcourse user will have to reimport missing sprites from sources eventually.

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -803,16 +803,12 @@ HError SpriteCache::InitFile(const char *filnam)
         return HError::None();
     }
 
-    // failed, delete the index file because it's invalid
-    // TODO: refactor loading process and make it NOT delete file running the game!!
-    ::remove(spindexfilename);
-
+    // Failed, index file is invalid; index sprites manually
     return RebuildSpriteIndex(_stream.get(), topmost, vers);
 }
 
 HError SpriteCache::RebuildSpriteIndex(AGS::Common::Stream *in, sprkey_t topmost, SpriteFileVersion vers)
 {
-    // no sprite index file, manually index it
     for (sprkey_t i = 0; i <= topmost; ++i)
     {
         _spriteData[i].Offset = in->GetPosition();
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -43,7 +43,10 @@ extern void get_new_size_for_sprite(int, int, int, int &, int &);
 #define END_OF_LIST   -1
 
 const char *spindexid = "SPRINDEX";
-const char *spindexfilename = "sprindex.dat";
+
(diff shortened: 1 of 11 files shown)


ivan-mogilko:
Changed couple of things:
* when there's missing sprites in spritefile, don't create empty bitmaps, instead assign a "remapped" sprite slot pointing at sprite 0 (like it's done by default). This should prevent from having unknown amount of memory allocated at once. The end result is about the same.
* I noticed that auxiliary sprite index file has hard-coded location and made it use function argument when loading/saving instead. Similarily, index file is also saved to temp file first now before being replaced in the project folder.

ivan-mogilko:
> Would it be possible print the path to both temp files when there is an error, assuming that the two are always in-sync?

Missing index file is not a big issue, because index may be recreated from spritefile data. There's also some kind of time-based tag to test if they match.

morganwillcock:
If the index file is rebuilt if it isn't correct, and the sprite file was always writeable, couldn't it just update the sprite file instead of linking the game to use the temporary copy then?

ivan-mogilko:
> If the index file is rebuilt if it isn't correct, and the sprite file was always writeable, couldn't it just update the sprite file instead of linking the game to use the temporary copy then?

It links to temporary copy only if failed to move new sprite file to project folder.

morganwillcock:
In my tests (just locking the index file) the original sprite file is deleted and there is no lock that should stop the temporary one being successfully moved back into place. I think the error handling isn't differentiating between the two files, but one of these file is apparently optional.

ivan-mogilko:
> In my tests (just locking the index file) the original sprite file is deleted and there is no lock that should stop the temporary one being successfully moved back into place. I think the error handling isn't differentiating between the two files, but one of these file is apparently optional.

Ah okay, I see what you mean now. I will try to adjust it then.

ivan-mogilko:
@morganwillcock I refactored saving a little, now it saves main file and index separately and may react differently on the result.

PR #791, opened 2019-05-27
Revert purpose of AlphaChannel flag

### LINKED ISSUE #772: New sprite import keeps alpha channel when converting 32-bit sprites for non-32-bit games
This was found by a user recently when testing 3.5.0 beta.

If you have a 16-bit game, then importing 32-bit sprites will convert them to 16-bit (at least this is what their import property tells). But if the "Import alpha channel" checkbox is set then the sprite will get AlphaChannel = True property anyway. This breaks their drawing logic in the editor (and maybe engine too) making them invisible when assigned to objects in the room.

Not sure what's the correct course for fixing is here.
Native code may need to remove alpha channel flag if the sprite is not 32-bit because all drawing logic in AGS depends on it, it seems.

morganwillcock:
Fixes #772


morganwillcock:
This restores the original behaviour, so the difference would be where the source file changes to include or remove an alpha channel a bulk re-import currently would use the same setting as the exsting sprite and not a preference from the import UI. I'll probably try and add that back separately to this.

Issue #787, opened 2019-05-26
Finalizing 3.5.0


ivan-mogilko:
I created a new branch "release-3.5.0" for the purpose of finalizing AGS 3.5.0 and free "ags3" branch from its restrictions at the same time. (although I do not know how the development will go from now own, would be there ags3 or ags4, and which branches you'll use)

I guess the idea for this branch is to not do any radical changes nor add major features, and focus on remaining tasks from the 3.5.0 milestone: https://github.com/adventuregamestudio/ags/milestone/3

In the order of importance (more or less):

- [x] (#607) Make final decision on whether we use Allegro 4.4.3 and if yes then do necessary adjustments.
- [x] (#712) Use alfont on Windows; this is not a difficult task but should be done early because CI task will depend on it.
- [x] CI (automatic building): need to make a checklist of tasks that are essential for the proper release and then complete it.
- [x] Switch to [new manual](https://github.com/adventuregamestudio/ags-manual). Update manual, documenting new functionality (guess I'll have to attend this because I added new scripting stuff).
- [x] Fix known problems with existing functionality.
- [x] Optionally implement other minor features that were most wanted for this release, if they do not require changing much code and are possible to do in a small time period.
- [x] Fix any remaining regressions related to legacy game formats.

Regarding minor extra features and non-essential fixes, I guess these could be assigned for a minor update (i.e. 3.5.1) in its own branch if we plan to allow one (since 3.6.0 development would likely be a serious change with introduction of SDL backend).

**Extra references:**
https://github.com/adventuregamestudio/ags/wiki/Public-release%3A-procedure
https://github.com/adventuregamestudio/ags/wiki/Public-release%3A-contents

PR #793, opened 2019-05-27
Disable SSL v3

### LINKED ISSUE #792: Crash reporting and statistic submission has stopped working
As mentioned in #781.

Once the SSL issue is fixed the response is 500 (Internal Server Error) for crash reporting and statistic submission.

morganwillcock:
Partially fixes #792.

I believe in later .NET versions this is already forced off and it shouldn't be being used anyway. Ideally this should just mandate TLS 1.2 or higher, but I don't think this currently works with the TLS negotiation settings between .NET and the server where data is being sent - I tried it and got an SSL error.


ivan-mogilko:
@morganwillcock I have created release-3.5.0 branch for finishing 3.5.0, you need to make pr for that branch if you want this in 3.5.0. (#787)

PR #795, opened 2019-05-28
AGS 3.5 - reintroduce the timer counter for managing frame rate.

sonneveld:
This pull request adds the original timing code used before my earlier timing pull requests. To enable it, define AGS_TIMING_USE_COUNTER. (edit: this is set by default in this PR)

I also moved all compiler definitions into their own visual studio properties file. I was getting sick of changing definitions in a dozen places.

Some benchmarks:

```
Unavowed main menu 60hz game

                            | fps   | cpu   | fps maxed | cpu maxed  
with counter / vsync on     | 60    | 3.5%  | 60        | 3.5%
with counter / vsync off    | 63    | 4%    | 310       | 11.5%
without counter / vsync on  | 60    | 3%    | 60        | 3%
without counter / vsync off | 60    | 4%    | 350       | 12.5%
```

So with the original timing code (i.e. with AGS_TIMING_USE_COUNTER defined), 60hz games are actually running a little fast at 63fps. (edit: to be clear, this was an issue before) Max frame rate is also a little lower but that's only really important for benchmarking. CPU usage seems roughly the same with both methods.

Setup:

```
use Release Builds for AGS and libraries

used vsync by forcing in code:
D3DPRESENT_INTERVAL_ONE
disabled vsync by forcing
D3DPRESENT_INTERVAL_IMMEDIATE
(I changed in code because AGS can sometimes override vsync if in window)

config:
[graphics]
windowed=1
defaultgfxdriver=D3D9
gfxdriver=D3D9
driver=D3D9
filter=StdScale
game_scale_win=1

fps maxed is with --fps and press ctrl-e to go 'unlimited'

I used window mode because it ensures no graphics scaling.
```


sonneveld:
I'm not sure why the Travis build failed. It complains about hash failures.

Engine/platform/base/agsplatformdriver.cpp
```diff
@@ -90,7 +90,10 @@ void AGSPlatformDriver::WriteStdOut(const char *fmt, ...) {
 }
 
 void AGSPlatformDriver::YieldCPU() {
-    std::this_thread::yield();
```
sonneveld: This was interesting. The original timing method relies on this being a delay. If you use a proper yield then the cpu usage jumps up to 100% again.

morganwillcock:
> I'm not sure why the Travis build failed. It complains about hash failures.

I re-ran it and it worked fine on the second try.

sonneveld:
Okay, we decided to go into another direction. This pull request still has things I'd like to merge in but I'll cherry pick and create a new pull request from it.

PR #796, opened 2019-05-28
Revert #685

rofl0r:
since it was decided to revert #685, i went to the effort to actually try to revert it cleanly. since the original commit changed dozens of only loosely related things in over 40 files and 300 lines, it gave a lot of merge conflicts, especially in the audio/voice related code that was heavily refactored since.

i think the only reasonable way to solve the issues with the new code is to go back to a known good version and then **carefully** and selectively apply single, fine-grained changes with a one-change-one-commit strategy; and not by simply hiding some of the new code behind an `#ifdef`, which then causes simply that we have 2 different places to maintain instead of one.

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -80,6 +80,7 @@ extern int said_speech_line;
 extern int numscreenover;
 extern int said_text;
 extern int our_eip;
+extern int cur_mode;
 extern CCCharacter ccDynamicCharacter;
 extern CCInventory ccDynamicInv;
 
@@ -2401,7 +2402,8 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
     play.speech_in_post_state = false;
 
     if (isPause) {
-        postpone_scheduled_music_update_by(std::chrono::milliseconds(play.messagetime * 1000 / frames_per_second));
+        if (update_music_at > 0)
+            update_music_at += play.messagetime;
         GameLoopUntilEvent(UNTIL_INTISNEG,(long)&play.messagetime);
         return;
     }
--- a/Engine/ac/characterinfo_engine.cpp
+++ b/Engine/ac/characterinfo_engine.cpp
@@ -15,6 +15,7 @@
 #include "ac/characterinfo.h"
 #include "ac/common.h"
 #include "ac/gamesetupstruct.h"
+#include "media/audio/soundclip.h"
 #include "ac/character.h"
 #include "ac/characterextras.h"
 #include "ac/gamestate.h"
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -60,6 +60,7 @@ extern ccInstance *dialogScriptsInst;
 extern int in_new_room;
 extern CharacterInfo*playerchar;
 extern SpriteCache spriteset;
+extern volatile int timerloop;
 extern AGSPlatformDriver *platform;
 extern int cur_mode,cur_cursor;
 extern IGraphicsDriver *gfxDriver;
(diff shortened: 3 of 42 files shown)


ivan-mogilko:
I think @sonneveld was going to make another variant without double code under ifdefs. I will wait for a few days.
Also, the revert should go into release-3.5.0 branch now.

sonneveld:
This reverts a lot of stuff unrelated to the cpu usage issue.

sonneveld:
For comparison, this reverts a lot less changes https://github.com/adventuregamestudio/ags/pull/797

sonneveld:
I think you need to consider your attitude here. There have been other major changes done in 3.5 that have needed minor fixes after being merged in. I find it curious you haven't suggested we revert those as well. We should just fix the issue and move on.

ivan-mogilko:
@sonneveld 
> There have been other major changes done in 3.5 that have needed minor fixes after being merged in. I find it curious you haven't suggested we revert those as well. 

what rofl0r probably means is that the commit that caused the cpu load  (https://github.com/adventuregamestudio/ags/commit/f2ab428a596c67d950f2aed5fbd335f6ccc72e3f) had lots of changes mixed in, making it harder to understand what caused the issue and harder to revert only changes that need revert (if one would like to).
I made a mistake by accepting it as-is, usually I was asking people to split commits into one commit per change. This is what I try to do myself when working on big changes. But that time guess I just gave up after multiple pull requests.

rofl0r:
sigh, it seems i forgot to pull latest changes to ags3 branch, thus the additional conflicts... so this needs to be redone anyway.

> what rofl0r probably means is that the commit that caused the cpu load (f2ab428) had lots of changes mixed in, making it harder to understand what caused the issue and harder to revert only changes that need revert (if one would like to).

correct. i think we'd need to revert that commit as good as possible and then re-add one change at a time, e.g. "use float for fps", "use std::chrono for timestamps" etc. that would make it possible to understand precisely what was changed and easily revert a single change that causes issues or simply to find out whether it is the culprit for a specific regression.

> There have been other major changes done in 3.5 that have needed minor fixes after being merged in. I find it curious you haven't suggested we revert those as well

yeah, there have been many big PRs like the one in question that changed much more than necessary, which i saw being applied with a good deal of frustration on my side on where this is all going. however #685 was special in that it introduced a specifically nasty regression. a CPU usage of 100% vs 33% means a difference of battery runtime of 50 mins vs 2.5 hrs, and between CPU fan making noise that's barely standable vs practically unnoticable. also it might preclude from using the engine at all on low-end hardware. that's why i think it's better to completely undo the change in this case to the known good code, and gradually re-do the changes that can easily be verified to be benign.

> I think @sonneveld was going to make another variant without double code under ifdefs (we spoke in different place)

having 2 different approaches with one under ifdef makes it even harder to understand what is going on, and every change to it needs to be done in 2 places. also it is easy to introduce changes that break the code in the ifdef branch that nobody compiles.

sonneveld:
> having 2 different approaches with one under ifdef makes it even harder to understand what is going on, and every change to it needs to be done in 2 places. also it is easy to introduce changes that break the code in the ifdef branch that nobody compiles.

This was because I was concerned that the merge conflicts (when merging 3.5.0 changes back into ags3) would be non-obvious if code was reverted in one branch.

> i think we'd need to revert that commit as good as possible and then re-add one change at a time, 

I don't really agree with you here. But are you volunteering to submit the new PRs to add these changes back?

ivan-mogilko:
Sorry, I lost the track of this issue.

In release-3.5.0 branch CPU load was fixed by reverting last sonneveld's changes to timers.

Were not some fixes made to ags3 branch too?

rofl0r:
i think this is no longer relevant.

PR #797, opened 2019-05-29
Ags3.5  timing use counter only

sonneveld:
This introduces the same changes as https://github.com/adventuregamestudio/ags/pull/795 but without the option of switching back to the new implementation (making comparisons harder).

When merging the 3.5.0 release branch into ags3, please be aware to ignore changeset 5424b49

### DIFF
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -90,7 +90,10 @@ void AGSPlatformDriver::WriteStdOut(const char *fmt, ...) {
 }
 
 void AGSPlatformDriver::YieldCPU() {
-    std::this_thread::yield();
+    // NOTE: this is called yield, but if we actually yield instead of delay,
+    // we get a massive increase in CPU usage.
+    this->Delay(1);
+    //std::this_thread::yield();
 }
 
 void AGSPlatformDriver::InitialiseAbufAtStartup()
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -190,18 +190,20 @@ int cd_player_control(int cmdd, int datt) {
 #endif // AGS_HAS_CD_AUDIO
 
 void AGSPlatformDriver::Delay(int millis) {
-  auto delayUntil = AGS_Clock::now() + std::chrono::milliseconds(millis);
+  auto now = AGS_Clock::now();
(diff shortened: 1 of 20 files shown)


Engine/platform/base/agsplatformdriver.cpp
```diff
@@ -190,18 +190,20 @@ int cd_player_control(int cmdd, int datt) {
...
   for (;;) {
-    if (AGS_Clock::now() >= delayUntil) { break; }
+    if (now >= delayUntil) { break; }
 
-    auto duration = delayUntil - AGS_Clock::now();
-    if (duration > std::chrono::milliseconds(25)) {
-      duration = std::chrono::milliseconds(25);
+    auto duration = delayUntil - now;
+    if (duration > std::chrono::milliseconds(16)) {
+      duration = std::chrono::milliseconds(16);
```
ivan-mogilko: Could you add a comment explaining what does 16 ms mean here?

sonneveld: Sure! Constants works too.

i30817:
So it was a yield instead of a sleep making the scheduler freak out because AGS was continuously the high priority task? Man, i'll never understand threading.

rofl0r:
> I think yield and sleep(0) should probably not be used in game loops, because if the idea was to lower the time spent on the process, it backfires spectacularly.

if you don't use sleep (or select/poll with file descriptors), you will always use 100% cpu. any reasonably designed game or engine will sleep as long as possible, to keep CPU load small (optimally it will sleep whenever there's no need to do calculations/screen refreshes, e.g. if the gamestate is unchanged). also, for anything else than a first person shooter, a framerate of 50 fps is more than sufficient for a smooth gaming experience.

i30817:
vsync changes that right? Now that i'm using wayland in gnome.... everything is vsync i think (there is no fullscreen exclusive too)

Engine/platform/base/agsplatformdriver.cpp
sonneveld: I added the comment. and a constant.

sonneveld:
So please remember this has the same issues we had before the new timing refactor. 60fps games without vsync will run a little bit faster (63fps) and we will again have issues where some gui dialogs use 100% cpu.

ivan-mogilko:
I guess we found how these dialogs may be fixed in this variant, so I'll try doing that for 3.4.x/3.5.0 release branches in a bit.

PR #798, opened 2019-05-29
multifilelib: don't allow head sig search to extend past tail offset

sonneveld:
This fixes issues on systems where it is an error to read at an offset past end. (reading at end should still give you 0 bytes though)

edit: I didn't add a check to Seek itself since it would require grabbing the file size on load. Plus, we can't always rely on that if we're reading sub-resources (in instances where we pass Streams already repositioned to the right resource).

### DIFF
--- a/Common/util/mutifilelib.cpp
+++ b/Common/util/mutifilelib.cpp
@@ -132,15 +132,18 @@ MFLUtil::MFLError MFLUtil::ReadSigsAndVersion(Stream *in, MFLVersion *p_lib_vers
         soff_t abs_offset_32 = in->ReadInt32();
 
         // test for header signature again, with 64-bit and 32-bit offsets if necessary
-        if (abs_offset > 0 && abs_offset < (tail_abs_offset - HeadSig.GetLength())) {
+        if (abs_offset > 0 && abs_offset < (tail_abs_offset - HeadSig.GetLength())) 
+        {
             in->Seek(abs_offset, kSeekBegin);
             sig.ReadCount(in, HeadSig.GetLength());
         }
 
         // try again with 32-bit offset
-        if (HeadSig.Compare(sig) != 0) {
+        if (HeadSig.Compare(sig) != 0) 
+        {
             abs_offset = abs_offset_32;
-            if (abs_offset > 0 && abs_offset < (tail_abs_offset - HeadSig.GetLength())) {
+            if (abs_offset > 0 && abs_offset < (tail_abs_offset - HeadSig.GetLength())) 
+            {
                 in->Seek(abs_offset, kSeekBegin);
                 sig.ReadCount(in, HeadSig.GetLength());
             }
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I find this quite disturbing, there are multiple Seeks in the program, do we need to check them all for this potential problem? Other people may later add/change the code without being aware of this issue too.
I *would* actually consider to control this inside the stream somehow.

Regarding this commit -

Could you, please, fix the coding style?

A bit of nitpicking:
* There's `in->Seek(-(soff_t)TailSig.GetLength(), kSeekEnd);` right above your added code, may be possible to avoid doing seek twice and remember offset after first one.
* Perhaps condition should be `abs_offset < (tail_abs_offset - HeadSig.GetLength())`?

sonneveld:
> Frankly, I find this quite disturbing, there are multiple Seeks in the program, do we need to check them all for this potential problem? What kind of error happens if we read past Seek, does it breaks further reading attempts?

I think it depends. In this instance we're reading unknown data and then doing an absolute seek on it. We should absolutely check this value before passing it to Seek. While we could also check in Seek, it would mean that we're hiding bad behaviour.

On some platforms, reading past the end of the file raises an error and aborts the program. I don't know if this is posix compliant, but that's what happens. Note, reading *at* the end of the file is fine. (you just get 0 bytes back)

> Perhaps condition should be abs_offset < (tail_abs_offset - HeadSig.GetLength())

I figured just checking against `tail_abs_offset` was fine because that's at least as long as HeadSig but I guess no harm adding it in.

sonneveld:
Ok I made those fixes and then attempted to fix the coding style. I split ReadSigsAndVersion into ReadSigs and ReadVersion since they do different things. That allowed me to refactor ReadSigs to be flat via early returns.

ivan-mogilko:
Excuse me, but why is your commit called "fix the coding style" while it does something different?

Also... the code style is not correct in the latest commit again.

sonneveld:
> Excuse me, but why is your commit called "fix the coding style" while it does something different?

I didn't think there was a style violation so I thought maybe you wanted a general code tidy up?

With coding style, are you referring to the brace on a new line? That's the only thing I could see after looking again.

ivan-mogilko:
> With coding style, are you referring to the brace on a new line?

Yes, I meant braces, nothing else. I just that your commits changed the code style in otherwise consistent file.

sonneveld:
> Yes, I meant braces, nothing else. I just that your commits changed the code style in otherwise consistent file.

ahhh ok. fixed. I was comparing variable name styles and thinking I was going crazy.

sonneveld:
> Refactor is ok if you think it makes things more clear.

I believe it does but I'll won't add it to `release-3.5.0`. I thought maybe that's what you were suggesting :)

rofl0r:
> On some platforms, reading past the end of the file raises an error and aborts the program. I don't know if this is posix compliant, but that's what happens.

nope, that's not compliant at all. on which platform did you encounter this broken behaviour?

sonneveld:
> nope, that's not compliant at all. on which platform did you encounter this broken behaviour?

Can't really say. It's more of a posix compatiblity layer though.

ivan-mogilko:
I'll take a guess this is one of the console systems people began porting AGS to recently (but cannot talk about because of legal conditions or something).

PR #804, opened 2019-06-01
3.5.0: fix viewport/camera restoration by reading data into temp structs and applying later

### LINKED ISSUE #775: 3.5.0: when restoring a save camera gets clamped to incorrect room bounds
When a save is restored camera size and position is clamped to the leftover room data from previous session instead of new room, because viewports and cameras are part of game data now and room is loaded later.

This may cause camera to restore in wrong position before "jumping" back to the player.

The solution is either to force-set unclamped data and fixup it later after restoration is complete, or do camera clamping during room update loop (i.e. UpdateViewports + UpdateRoomCameras functions).

### LINKED ISSUE #779: Small 'flashes' on scene startup on ogl driver
Basically (this is A tale of Two Kingdoms Deluxe new game, although it happens on other games too, often when restoring a game, not always though - afaict):
output

On software the screen stays black until the text displays here. Those objects are objects from the prologue scene that will display 'soon' after that text overlay. This is not really a big deal but it may push some game authors to go with the software driver for something that may end being a minor issue in the end.

ivan-mogilko:
This should fix #775 and may also fix #779 (not sure if both games mentioned there have same problem).

Changes:
* When loading a save keep viewport & camera data in a temporary struct and only apply in the end: this allows to clamp camera positions to actual room bounds (and not depend on loading order);
* Make sure load_new_room function only adjusts cameras when called *not* during restoring a save.
(This function is super messy and may benefit from splitting into two or more functions with different purposes, but that's not a task for this PR)

### DIFF
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1144,9 +1144,27 @@ HSaveError restore_game_head_dynamic_values(Stream *in, RestoredData &r_data)
     r_data.CursorID = in->ReadInt32();
     int camx = in->ReadInt32();
     int camy = in->ReadInt32();
-    play.CreatePrimaryViewportAndCamera();
-    play.GetRoomCamera(0)->SetAt(camx, camy);
     set_loop_counter(in->ReadInt32());
+
+    // Precreate viewport and camera and save data in temp structs
+    play.CreateRoomCamera();
+    play.CreateRoomViewport();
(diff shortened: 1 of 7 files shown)


sonneveld:
I had a quick look over the individual changesets. I'm still not 100% familiar with the new viewport/cameras but this looks okay!

i30817:
This only fixes #775 , the other bug appears different.

PR #806, opened 2019-06-01
timing: Refactor timing so all wait loops are done from the one function.

sonneveld:
Refactor timing so all wait loops are done from the one function.

This reverts an earlier change made in 6db4421 that introduced PollUntilNextFrame that was useful but mixed in the concerns of fast forward and polling. Pulling that out that allowed us to use the one wait loop function.

### DIFF
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -1046,7 +1046,13 @@ bool DialogOptions::Run()
             return true; // continue running loop
         }
       }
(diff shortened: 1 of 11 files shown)


sonneveld:
The code 
```    while (timerloop == 0)
    {
        platform->YieldCPU();
    } ```
could be replaced with a function. Like `waitingForNextTick`

sonneveld:
Squashed commits and added details on which commits were reverted.

PR #807, opened 2019-06-01
Fix fullscreen modes on Windows system which use display scaling

### LINKED ISSUE #736: OpenGL window gets displaced on some Windows systems
First reported and discussed here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=56699.0

The latest guess is that Windows treats ags game as "non-DPI-aware" application and tries to adjust its position on screen. This breaks OpenGL fullscreen mode.

Possible problem and solution reference:
https://docs.microsoft.com/en-gb/windows/desktop/hidpi/high-dpi-desktop-application-development-on-windows
https://stackoverflow.com/a/31979996/965638
https://stackoverflow.com/questions/13228185/how-to-configure-an-app-to-run-correctly-on-a-machine-with-a-high-dpi-setting-e
https://stackoverflow.com/questions/47049651/opengl-and-windows-10-per-monitor-aware-dpi-scaling

morganwillcock:
With 'HiDPI' scaling OpenGL and software renderers would only render a corner of the game  and mouse boundaries were mis-calculated. For OpenGL the display resolution was also mistakenly changed. This seems to fix it.

Fixes #736

### DIFF
--- a/Solutions/Engine.App/Engine.App.vcxproj
+++ b/Solutions/Engine.App/Engine.App.vcxproj
@@ -114,6 +114,9 @@
       <SubSystem>Windows</SubSystem>
       <TargetMachine>MachineX86</TargetMachine>
     </Link>
+    <Manifest>
+      <EnableDpiAwareness>true</EnableDpiAwareness>
+    </Manifest>
   </ItemDefinitionGroup>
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Is this setting per-configuration? If so it has to be added to "All Configurations" in MSVS.

morganwillcock:
Sorry, I thought it was for the whole project not per-config. I've re-done it.

PR #813, opened 2019-06-04
Script API: replaced "sorted" and "caseSensitive" boolean params with enums

ivan-mogilko:
By user request, this replaces "sorted" and "caseSensitive" booleans in number of functions to corresponding enum types:

<pre>
enum StringCompareStyle
{
  eCaseInsensitive = 0,
  eCaseSensitive = 1
};

enum SortStyle
{
  eNonSorted = 0,
  eSorted = 1
};
</pre>

This applies to some of the String functions as well as newer Dictionary.Create and Set.Create.

The constant values are chosen to match the meaning of boolean parameters, so that the old scripts still compile and work the same way.

---

As a separate commit I also changed two readonly properties in Dicrionary and Set:
* bool CaseSensitive to CompareStyle;
* bool Sorted to SortStyle;

Was not sure about it, so this is done as a separate commit that could be dropped if others think this is not good.

NOTE: Since this is a very new API I believe it's not necessary to keep support for previous property names.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -415,6 +415,20 @@ enum CharacterDirection {
 };
 #endif
 
+#ifdef SCRIPT_API_v350
+enum StringCompareStyle
+{
+  eCaseInsensitive = 0,
+  eCaseSensitive = 1
+};
+
+enum SortStyle
+{
+  eNonSorted = 0,
+  eSorted = 1
+};
+#endif
+
 internalstring autoptr builtin managed struct String {
   /// Creates a formatted string using the supplied parameters.
   import static String Format(const string format, ...);    // $AUTOCOMPLETESTATICONLY$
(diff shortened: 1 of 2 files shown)


ericoporto:
It's me, the user. My reason for the first set of changes is `Dictionary.Create(true, true)` is less readable than `Dictionary.Create(eSorted, eCaseSensitive)`.

For the second part, the property, `myDict.Sorted == true` vs `myDict.SortStyle == eSorted`, my take is the first is more concise, but both are equally readable, the second (the commit) being more consistent with the creation values. But I like more `myDict.CaseSensitive == true` than `myDict.CompareStyle == eCaseSensitive` (my feeling is the second feels bureaucratic, while the true comparison tells you what you need to know)

ivan-mogilko:
The reason for the properties to be enum too is consistency with the arguments. Enums means that possible variants may be expanded. If they are, then these boolean properties will no longer cover them.

ericoporto:
Ok, this is a good reason I agree with. :)

PR #815, opened 2019-06-05
PushValueToStack: remove validation code for undefined values

rofl0r:
the ags script compiler rarely makes use of the registers "cx" and "dx"
if at all. this means they will contain a scriptvalue of type "undefined"
until first usage.
with the check assuming pushing undefined values to the stack is invalid,
it makes it impossible for a bugfix developer to use those registers
freely, as he needs to push them before usage to store their contents,
and pop the original contents after he is done using them.
if they happen to contain undefined values before the push, the engine
will quit, even though everything happening is perfectly safe:
the registers contained undefined before the push, some valid values
while it's used, and undefined again after the pop.

additionally, push is an operation used very often, and the 2 basically
unnecessary check add quite some overhead to script execution time.
ideally, all virtual cpu operations should be implemented as inline
functions or macros and as short as possible.

### DIFF
--- a/Engine/script/cc_instance.cpp
+++ b/Engine/script/cc_instance.cpp
@@ -1851,16 +1851,6 @@ bool ccInstance::FixupArgument(intptr_t code_value, char fixup_type, RuntimeScri
 
 void ccInstance::PushValueToStack(const RuntimeScriptValue &rval)
 {
-    if (!rval.IsValid())
-    {
-        cc_error("internal error: undefined value pushed to stack");
-        return;
-    }
-    if (registers[SREG_SP].RValue->IsValid())
-    {
-        cc_error("internal error: valid data beyond stack ptr");
-        return;
-    }
     // Write value to the stack tail and advance stack ptr
     registers[SREG_SP].WriteValue(rval);
     registers[SREG_SP].RValue++;


sonneveld:
> push is an operation used very often, and the 2 basically
> unnecessary check add quite some overhead to script execution time

Do you have any benchmark numbers measuring the reduced overhead?

> with the check assuming pushing undefined values to the stack is invalid,
> it makes it impossible for a bugfix developer to use those registers
> freely

Would an alternative be just to initialise all registers to zero first?

Could this PR be written such that we only do the checks if `DEBUG` is defined? And only quit if a register other than CX,DX is used?

ivan-mogilko:
I'd also ask to keep these under DEBUG. If my memory is correct, the "undefined" check helped to detect a hidden regression at least once already.
If this really helps perfomance same could be done for other checks in similar operations.

> with the check assuming pushing undefined values to the stack is invalid,
it makes it impossible for a bugfix developer to use those registers
freely, as he needs to push them before usage to store their contents,
and pop the original contents after he is done using them.
if they happen to contain undefined values before the push, the engine
will quit, even though everything happening is perfectly safe:

I am confused by this description, I will need to refresh my memory on how vm works, but this sounds like either a bug in the engine or incorrect use of registers/stack...

ivan-mogilko:
PS. Maybe replace these checks with `assert`?
Like
<pre>
assert(!rval.IsValid() && "internal error: undefined value pushed to stack");
assert(registers[SREG_SP].RValue->IsValid() && "internal error: valid data beyond stack ptr");
</pre>

rofl0r:
> Could this PR be written such that we only do the checks if DEBUG is defined?

i guess one could leave it as a warning, not error if DEBUG is defined. though imo there's not really an issue with pushing whatever onto the stack when it is later restored.
if a register value can be undefined, then it should be allowed to move them to stack too.

rofl0r:
> I am confused by this description, I will need to refresh my memory on how vm works, but this sounds like either a bug in the engine or incorrect use of registers/stack...

the script compiler almost never emits code using cx and dx (i've only seen one instance where it used cx, and none that use dx).

i wrote some code in my assembler to write a bugfix for quest for infamy, where i pushed cx and dx to use them for some temporary stuff, and restore them later to whatever was in them to be on the safe side, but the registers happen to contain undefined values (i.e. they weren't used yet):

```diff
--- room15.s.org
+++ room15.s
@@ -2210,12 +2210,39 @@
 	li ax, Character::get_ActiveInventory
 	farcall ax
 	pop op
-	push ax
-	li mar, iKnife
-	mr ax, mar
-	pop bx
-	cmpeq bx, ax
-	mr ax, bx
+
+; ---------- patch
+; dx will hold whether we used one of the 3 swords on the head
+push dx
+xor dx, dx
+
+; cx will hold the active inventory item
+push cx
+mr cx, ax
+
+li mar, iKnife
+mr ax, mar
+mr bx, cx
+cmpeq bx, ax
+add dx, bx
+
+li mar, iSwordBastard
+mr ax, mar
+mr bx, cx
+cmpeq bx, ax
+add dx, bx
+
+li mar, iSwordGreat
+mr ax, mar
+mr bx, cx
+cmpeq bx, ax
+add dx, bx
+
+mr ax, dx
+
+pop cx
+pop dx
+
 	jzi label000000005170
 	push ax
 	li mar, oDeadBeast
```
(the bug here is that if you upgraded your sword (iKnife) to a better sword before killing the beast, you won't be able to cut off its head, therefore you can't advance the plot)

rofl0r:
> Do you have any benchmark numbers measuring the reduced overhead?

no, i didn't measure it yet, but having the checks there should add at least 6 additional instructions and 2 branches to every `push`, which i assume makes the functionality 2-3x slower.

sonneveld:
> no, i didn't measure it yet, but having the checks there should add at least 6 additional instructions and 2 branches to every `push`, which i assume makes the functionality 2-3x slower.

A rough check might be to run the engine with `--fps` and use ctrl-e to run at max frame rate.

ivan-mogilko:
> A rough check might be to run the engine with --fps and use ctrl-e to run at max frame rate.

Is this really necessary?
I tend to agree that these checks should be there for debugging reasons only, and the more could be skipped in release build the better. This function is called *very* often.

ivan-mogilko:
> i guess one could leave it as a warning, not error if DEBUG is defined. though imo there's not really an issue with pushing whatever onto the stack when it is later restored.
if a register value can be undefined, then it should be allowed to move them to stack too.

Thing is, what I'd like to have there is something which immediately bring attention or breaks execution if something unexpected was put on stack. In that case I'd wish to know exact moment that happened, otherwise this "undefined" value may resurface later (e.g. as an argument of some operation) and it will be harder to understand where did it come from.

I'd like to think more about this...

rofl0r:
> something which immediately brings attention

well, if it's a warning it will probably cause a lot of spam on the terminal, something that can't be missed... while still allowing correct execution of correct bytecode even in debug mode (correct in the sense that it adheres to the VM spec)

> "undefined" value may resurface later (e.g. as an argument of some operation) and it will be harder to understand where did it come from.

yes, though for this special purpose (hunting down a problem causing crashes in the VM or similar) i think it'd be best to activate the code only with a special define like DEBUG_VM that would only be used in case such a problem is detected. the reason i'd like the default debug mode not to be that restrictive is that it is quite useful for other purposes too, like inspect behaviour with gdb, and the debug command line switches it uses. if the debug mode aborts the program encountering such a push command, it's not possible to debug anything which would be triggered later.
the advantage of a switch like DEBUG_VM is that probably a lot more safety checks which are only needed for very rare in-depth problem analysis could be removed, giving the interpreter a nice speed boost.

> I'd like to think more about this...

sure, take your time

sonneveld:
>  i think it'd be best to activate the code only with a special define like DEBUG_VM that would only be used in case such a problem is detected.

Another options might be to mimics SDLs assertion levels. There are different levels for stringent(always), debug and release.

sonneveld:
If we push and pop an undefined value, will the register still be set as undefined?  I think as long as we check when we go to read the value if it's on stack or the later register, this could work?

ivan-mogilko:
Originally all values in the AGS vm were "undefined", because they carried no type identification, only some value that was treated depending on operation.

IIRC the main reason to add type was to allow to work around direct memory access to engine internals which prevented us from easily modifying game structs.
Another reason was that integers and floats were expected to be 32-bit while stored pointers could be of varying bitness depending on system; therefore we started to store plain numeric values and pointers separately.
They were not stored in a single union because we still wanted to keep offset separate from pointer. This was necessary to distinguish actual object the value is referencing (which would be impossible/hard if pointer also included extra offset).
Then came the need to also store "manager" pointer along the object pointer, because at certain situations we had to call its functions to resolve some legacy mode memory access.

If the above problems could be solved differently, without having to store type and separate numeric value/pointer/manager, then script VM could be brought back to typeless values and also work faster without type checking everywhere. One of the grander options is to have a separate game library for legacy games, because then it may have game structs allocated in a compatible way, thus no need for wrapping/remapping access. But all this is a bigger issue.

---

Regarding this particular problem, my thinking is that since registers may indeed have undefined values (at least at start) we could allow stack have these values too.

Not sure if the place this pr is changing is the only one which tests for undefined values on stack. We might have to search through the vm to see if it tests for that somewhere else.

Still I would like to have some kind of "assert" function there that would trigger a break point automatically under debugger. Mere warning message may not be convenient because when debugging it may be essential to stop at exact moment something unexpected happen to observe the issue.

Perhaps there could be a function like "script_assert" which definition depends on compiler flags. It could be a no-op for release build, a warning output for regular debug build and an actual assert if we have e.g. DEBUG_SCRIPT_VM flag defined.

On a side note, some simplier operations, like SCMD_ADDREG, do not have any type checks at the moment, meaning you could pass undefined values there too, and they will be treated as integers (with value 0 perhaps).

rofl0r:
> Not sure if the place this pr is changing is the only one which tests for undefined values on stack. We might have to search through the vm to see if it tests for that somewhere else.

it did solve the problem, so it seems fine for now. more can be added on an as-needed basis, if required.

> I would like to have some kind of "assert" function there that would trigger a break point automatically under debugger.

the only portable way to do this is to emit SIGTRAP signal, but that's only for POSIX compatible platforms. for windows, if x86 cpu is used, one can use `__asm__("int3");`. both of them will cause the program to abort if there's no debugger attached though. so one could make it dependant on whether a debugger is attached, which on POSIX can be found out via `ptrace` with `PTRACE_TRACEME`, iirc. but i don't know for windows.

> Perhaps there could be a function like "script_assert" which definition depends on compiler flags.

that sounds like a good idea. it's better to have an ifdef in a single header defining the macro, than spread all over the code.

ivan-mogilko:
Alright, I guess this may be merged because it does not seem to break any actual logic, and since you say it is necessary for you to apply these script fixes.

Guess the assertions could be added later in a more consistent way.

Note that "valid data beyond stack ptr" check would not be useful anymore if we allow undefined data pushed around.

PR #819, opened 2019-06-07
Use embedded alfont source in Windows version

ivan-mogilko:
Since we now have alfont + freetype lib sources directly in the repository for linux version, this makes MSVS solution also use these for windows version.

Please note that I had to move these sources from Engine/libsrc to Common/libsrc and include them in Common.lib. This is because Editor also needs them for its native component.

I was not sure if there is any difference what they are compiled with when using Makefile: as a part of building **common.a** or as main program; I made them a part of common.a there too for consistency, but I can revert this.

Please check that the Linux port will still build correctly with this change.

PS. not sure if CMakeLists need to be adjusted and how.


rofl0r:
in alfont.h and alfontdll.h you seem to have introduced line ending changes from CRLF to LF or vice versa. this can be fixed by using dos2unix or unix2dos tools on linux and amending the changes to the commit that introduced it.

morganwillcock:
I think this means that the property sheet I added can be removed, and references deleted.

The file is:
```
Solutions/alfont.props
```

referenced in:
```
Solutions/Engine.App/Engine.App.vcxproj:    <Import Project="..\alfont.props" />
Solutions/Engine.App/Engine.App.vcxproj:    <Import Project="..\alfont.props" />
Solutions/Engine.App/Engine.App.vcxproj:    <Import Project="..\alfont.props" />
Editor/AGS.Native/NativeDLL.vcxproj:    <Import Project="..\..\Solutions\alfont.props" />
Editor/AGS.Native/NativeDLL.vcxproj:    <Import Project="..\..\Solutions\alfont.props" />
```

ivan-mogilko:
@rofl0r Ok, I just used `git mv` from console which seem to have better results
@morganwillcock also ammended.

ivan-mogilko:
@sonneveld although using `git mv` fixed this, but before that I had a nasty issue when freetype-2.1.3/docs/BUILD text file got ignored because our .gitignore has this:
<pre>## Xcode Build generated
**/build
</pre>

Looks like on Windows gitignore's entries are sometimes applied case-insensitive (maybe a bug).
Is this entry still relevant, and is there a way to make it more accurate?

ericoporto:
Since `BUILD` is a file, a quick solution would be having the code below

> ```
> ## Xcode Build generated
> **/build
> ```

changed to

```
## Xcode Build generated
**/build/
```

> A leading "**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo". "**/foo/bar" matches file or directory "bar" anywhere that is directly under directory "foo".

https://git-scm.com/docs/gitignore

Also reading on Xcode, apparently it only generates a `build/` folder on the root of the directory. So my guess is that Xcode projects are generated nested in some directories, in lower levels.

> Patterns read from a .gitignore file in the same directory as the path, or in any parent directory, with patterns in the higher level files (up to the toplevel of the work tree) being overridden by those in lower level files down to the directory containing the file. These patterns match relative to the location of the .gitignore file. A project normally includes such .gitignore files in its repository, containing patterns for files generated as part of the project build.

My best guess would be having `.gitignore` files on those directories - just note that it would override rules from the root `.gitignore`. I don't know which directories those are, just giving possible ideas. For this approach is better to wait @sonneveld to answer.

sonneveld:
> Looks like on Windows gitignore's entries are sometimes applied case-insensitive (maybe a bug).
> Is this entry still relevant, and is there a way to make it more accurate?

The entry is relavent like @ericoporto  says, but we can just move it into a separate .gitignore file for iOS and OSX I believe.

sonneveld:
> PS. not sure if CMakeLists need to be adjusted and how.

As long as you copied alfont's `CMakeLists.txt` file over, you only need to modify the path in the root `CMakeLists.txt` file:

```
add_subdirectory(Common/libsrc/aastr-0.1.1      EXCLUDE_FROM_ALL)
+ add_subdirectory(Common/libsrc/alfont-2.0.9     EXCLUDE_FROM_ALL)
add_subdirectory(Common                         EXCLUDE_FROM_ALL)

- add_subdirectory(Engine/libsrc/alfont-2.0.9     EXCLUDE_FROM_ALL)
add_subdirectory(Engine/libsrc/almp3-2.0.5      EXCLUDE_FROM_ALL)
```

ivan-mogilko:
@sonneveld please check if it's correct.

sonneveld:
The change looks good but I won't be free to build it until tomorrow morning

sonneveld:
This builds fine on my machine once I merged in the changes from https://github.com/adventuregamestudio/ags/pull/825 (needed because the macbook I'm borrowing is running macos 10.10 and I can't get recent headers)

PR #824, opened 2019-06-08
AGS3 revisit `_rgb_[argb]_shift_\d+`

sonneveld:
This is just a refactor to make it clearer what we're doing when we're resetting the shift variables. 

The `convert_16bit_bgr` check wasn't actually useful because we were checking shift variables _after_ we'd already set them. So I moved it to the top.

Other than that it should have the exact same behaviour as before.

The order of setting variables was sometimes `r,g,b`, sometimes `b,g,r`, which made it hard to visually scan and has probably led to issues.

I think OpenGL ES is the only reason we need to set these shifts differently now.  We could set the texture format to `GL_BGRA, GL_UNSIGNED_INT_8_8_8_8_REV`, which is equivalent to direct3d's `D3DFMT_A8R8G8B8` format but unfortunately OpenGL ES doesn't support that pixel format.

I'm posting as a draft request because I wanted people to have a look and tell me what I'm missing in regards to simplifying that final switch statement.  Why do we change shifts if ! Windows, did we mean Ios and Android?

Should we just have the standard pixel formats here and require the graphics drivers to do conversion if needed?

### DIFF
--- a/Engine/main/engine_setup.cpp
+++ b/Engine/main/engine_setup.cpp
@@ -185,6 +185,19 @@ void engine_pre_gfxsystem_screen_destroy()
 // Setup color conversion parameters
 void engine_setup_color_conversions(int coldepth)
 {
+    // Most cards do 5-6-5 RGB, which is the format the files are saved in
+    // Some do 5-6-5 BGR, or 6-5-5 RGB, in which case convert the gfx
+    if ((coldepth == 16) && ((_rgb_b_shift_16 != 0) || (_rgb_r_shift_16 != 11)))
+    {
+        convert_16bit_bgr = 1;
+        if (_rgb_r_shift_16 == 10) {
+            // some very old graphics cards lie about being 16-bit when they
+            // are in fact 15-bit ... get around this
+            _places_r = 3;
+            _places_g = 3;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


sonneveld:
This is broken up into lots of commits to show my working, but we can squash later if it goes in.

ivan-mogilko:
Should not this target release-3.5.0 instead?

sonneveld:
> Should not this target release-3.5.0 instead?

This PR doesn't really fix things, it was more of a refactor with a question mark at the end. If we know for sure that we only need those shift changes for opengl, I can make that more explicit in the commit and change this for 3.5.0 if you like.

ivan-mogilko:
> I can make that more explicit in the commit and change this for 3.5.0 if you like.

Well, the reported bug is in 3.5.0 so I thought the fix should be applied to 3.5.0 too. I don't know about overall osx port state in there though.

ivan-mogilko:
> This PR doesn't really fix things, it was more of a refactor with a question mark at the end. If we know for sure that we only need those shift changes for opengl, I can make that more explicit in the commit and change this for 3.5.0 if you like.

Oh alright, I was not sure what is it for. Only bug fix is necessary for 3.5.0 at this time.

PR #831, opened 2019-06-11
Add alpha channel import preference for Sprites

### LINKED ISSUE #772: New sprite import keeps alpha channel when converting 32-bit sprites for non-32-bit games

morganwillcock:
Adds a Sprite property to store the request to use an image's alpha channel. When reloading sprites from source the use of the alpha channel is checked against this property (so if you manually modify the source you can also modify this to suit). When replacing a sprite using the import UI this is used as a the default option for the 'use alpha channel' checkbox.

This should re-add the intended behaviour for bulk sprite reloading that had to be removed in order to fix #772

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -81,7 +81,7 @@ namespace AGS.Editor
          *                  Real sprite resolution; Individual font scaling; Default room mask resolution
          * 19: 3.5.0.11   - Custom Say and Narrate functions for dialog scripts. GameFileName.
         */
-        public const int    LATEST_XML_VERSION_INDEX = 19;
+        public const int    LATEST_XML_VERSION_INDEX = 20;
         /*
          * LATEST_USER_DATA_VERSION is the last version of the user data file that used a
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
I just have 1 small request: could you add a short comment about what has changed to the list above LATEST_XML_VERSION_INDEX?

morganwillcock:
Regarding the XML version number, I didn't know the definitive place to get the release number to list alongside it, or whether I could have actually stuck with the current version number.

ivan-mogilko:
These comments are just for reference, I can ammend it myself when releasing next build.

PR #832, opened 2019-06-12
Android: embed alfont and freetype lib sources

ivan-mogilko:
Same as for Windows earlier, this should make Android port reuse embedded alfont and freetype lib sources.

TBH it took me several iterations of hacking Makefiles to make it work, and I still not sure how it does.
Android's makefile's location is different and it has different level of nesting compared to the rest, and since our makefiles use a lot of relative paths I could not figure out how Android.mk actually gets them all fixed.
But there's definitely some nasty magic involved, because during my experiments I was getting errors with paths like:

`../jni/../jni/../../../Engine/../Common/real path follows`

Because of that I was not able to use FT_CFLAGS and had to put explicit path to freetype's include directory.

Maybe this could be reviewed and improved later.


rofl0r:
LGTM

seems the only "issue" is that you manually add the include dir. it's kinda unexpected that android uses a different file system layout, thus making the relative paths pointing to nirvana. maybe this should be revised at some point, though the fix you did is simple enough for the time being.

PR #846, opened 2019-06-19
AGS3-next: reworking of previous streams pull request

sonneveld:
This is a reworking of the pull request https://github.com/adventuregamestudio/ags/pull/749

The main difference is that I kept the raw pointers to Stream and AlignedStream. This still has the unique_ptrs for managing the core stream classes but it is hidden behind the "classic" Streams.

This is intended for the next version of AGS (which I think `ags` is now), not 3.5


ericoporto:
This is on milestone 3.6.0 but it's closed. Is there any reason why this is closed? Either it should be open or removed from the milestone.

Issue #821, opened 2019-06-07
benchmark mode


rofl0r:
it's currently pretty hard to properly benchmark codechanges.

@sonneveld suggested the following in #815 : 

> A rough check might be to run the engine with --fps and use ctrl-e to run at max frame rate.

i tried this out, but the FPS jump quite a bit, so it is pretty unreliable.

i think a good way to measure performance would be a special mode (could be triggered with --benchmark), which, upon receipt of a certain signal (could be ctrl-e as well) runs at max framerate for a specified timeframe, e.g. 60 seconds (this could be an option to --benchmark switch), and then terminates the program and prints the average framerate for that time (the user-based triggering is needed so one can load a specific savegame and prepare things)
optimally, the mode could even suppress video and sound output, so only the raw speed of the engine is measured, without artifial caps due to hardware driver limitations.
instead of average fps, the total number of loop iterations which is currently shown in FPS display could be a good indicator for performance too.

morganwillcock:
I think that potentially you may also need a forced 'max speed and full screen' update mode to prevent any kind of overlay that is updating more quickly than the game from creating a smearing effect on the screen e.g. Steam Acheivement pop-ups leaving a trail on the game display or on the black masking bars.

So perhaps this feature is for benchmarking but could also be relevant for scripts and plug-ins.

rofl0r:
turns out there's a way the fps display can be used to reason about performance improvements. this is when a game is used that's extremely script-heavy, such as "Operation Forklift". 

for example, when using my new assembly optimizer pass in [agsutils](https://github.com/rofl0r/agsutils), i get almost 20% higher FPS (without the optimization pass, the game doesn't even reach the targeted 40 FPS).

therefore, if we had some sort of specialized benchmark testgame, which does a lot of script stuff and very rarely updates the screen, this could be a halfways reliable way to benchmark.
basically that testgame only needs a single loop that's repeated very often for every frame, and uses all possibilities of the VM. e.g. it needs to do function calls to external scripts, to the same script, read and write stuff to the memory (script objects), array indices, all arithmetic / boolean/ logical ops, etc. optimally, that game achieves max 40 fps on very fast desktop machines, so even small performance improvements in the order of 0.5% can be reliably measured.

sonneveld:
> therefore, if we had some sort of specialized benchmark testgame, which does a lot of script stuff and very rarely updates the screen, this could be a halfways reliable way to benchmark.

[...]

PR #847, opened 2019-06-20
DRAFT: Readvalue() optimization

rofl0r:
this should speed up Readvalue() which is one of the most used functions during script executions (taking up as much as 3% cpu in my tests). inlining it *should* speed it up.
"should" because it's hard to measure any improvements to script interpreter code with the --fps method which is only adequate to measure gfx performance.

### DIFF
--- a/Engine/script/runtimescriptvalue.h
+++ b/Engine/script/runtimescriptvalue.h
@@ -19,10 +19,11 @@
 #define __AGS_EE_SCRIPT__RUNTIMESCRIPTVALUE_H
 
 #include "script/script_api.h"
+#include "util/memory.h"
+#include "ac/dynobj/cc_dynamicobject.h"
 
 struct ICCStaticObject;
 struct StaticArray;
-struct ICCDynamicObject;
 
 enum ScriptValueType
 {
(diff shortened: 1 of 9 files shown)


Engine/script/runtimescriptvalue.h
```diff
@@ -309,13 +310,63 @@ struct RuntimeScriptValue
         return !(*this == rval);
     }
 
+    // FIXME: find out all certain cases when we are reading a pointer and store it
+    // as 32-bit value here. There should be a solution to distinct these cases and
+    // store value differently, otherwise it won't work for 64-bit build.
```
ivan-mogilko: Frankly, I don't remember what this comment means and whether it is still applicable.

ivan-mogilko:
Sorry. the title sais "draft", but there were no ammendments for 20 days; this probably may be merged as it is then?

TBH I don't like having extra header for just 1 function, but the code is constantly reorganized as it is refactored and rewritten, so that may not be a big deal.

rofl0r:
it means it's a proposed change, waiting for discussion. but yeah, you can go ahead and merge it, at the very least it untangles the recursive dependencies of headers. it's a bit hard to measure performance tweaks without a way to benchmark, unfortunately.

sonneveld:
> it's a bit hard to measure performance tweaks without a way to benchmark, unfortunately

Do you use the opengl drivers? It should be fairly low overhead, especially with vsync disabled. You could create a game with a blank room and write test benchmark test scripts.

rofl0r:
> Do you use the opengl drivers?

no, those are really slow on my device. as stated in #821, the only way to get reliable script interpreter performance measurements is by turning rendering off completely.

sonneveld:
> as stated in #821, the only way to get reliable script interpreter performance measurements is by turning rendering off completely.

Sure, but until something like that is implemented, this is the best we've got.

PR #855, opened 2019-06-27
Add CMake project files for 3.4.4 release

sonneveld:
Similar CMake files to what we have in the 3.5 branch. I decided to not move include files around, which led to some funny paths to `../../../Common/libinclude` for different dependencies.

Currently in draft for testing. So far I've tested on macOS, Linux and Windows. If others could test on other systems such as Linux and Windows, I would appreciate it. :)


rofl0r:
usually, stable branches are strictly for bug fixes. there's no point in adding a new build system to it.

sonneveld:
I suggested adding it since we have people comparing builds between 3.4 and 3.5. I don't think there's any harm in adding it.

rofl0r:
> besides we're adding features like OpenGL for other platforms.

yeah, i was kinda baffled by that decision

ivan-mogilko:
I decided to add OpenGL because some people were already producing custom 3,4,* builds for linux with OpenGL. It was my wish to add it too back when I was preparing 3.4.3, but I've been waiting for anyone knowing this stuff to make a pull request and no one had.

sonneveld:
> In theory it should not interfere with the rest of the engine.

Yeah sorry, I wasn't intending to throw you under the bus. I just meant that yes we're backporting features but we're not breaking the API, which is usually what people care about with minor releases.

sonneveld:
I've tested this on macOS, Linux and Windows now so it should be ready.

PR #859, opened 2019-07-02
3.5.0: made String::Split also return first and last empty segment

ivan-mogilko:


Engine/test/test_string.cpp
```diff
@@ -462,11 +462,19 @@ void Test_String()
...
+        assert(result.size() == 5);
         assert(strcmp(result[0], "C:") == 0);
         assert(strcmp(result[1], "Games") == 0);
         assert(strcmp(result[2], "AGS") == 0);
         assert(strcmp(result[3], "MyNewGame") == 0);
+        assert(strcmp(result[4], "") == 0);
+        String str2 = "test,,,test";
+        result = str2.Split(',');
+        assert(result.size() == 4);
+        assert(strcmp(result[0], "test") == 0);
```
sonneveld: Would it be worth adding another test for an empty first item? like ",,test,,"

PR #870, opened 2019-07-14
Use NuGet packaged DockPanelSuite

morganwillcock:
Since package restore will be happening anyway, DockPanelSuite can just come from the packaged version instead of commiting the binary.

(Upgrading DockPanelSuite is needed at some point, to avoid a crash, if running on Mono)


ivan-mogilko:
Hmm, ok, not sure how did I skipped this for 2 months...
@morganwillcock is this PR still relevant? we should merge it then.

morganwillcock:
Yes, still relevant. It just moves the dependency from a commited binary to the package spec, to match the others.

PR #873, opened 2019-07-20
Engine: allow to quit from max framerate mode

rofl0r:
for some reason, the code here didn't allow one to use CTRL-E a second
time to turn off maxed framerate again. since the fps setting is even
saved into savegames, this left no other choice than to the quit the engine
and start from scratch.

### DIFF
--- a/Engine/ac/global_game.cpp
+++ b/Engine/ac/global_game.cpp
@@ -373,9 +373,10 @@ void SetRestartPoint() {
 
 
 void SetGameSpeed(int newspd) {
-    // if Ctrl+E has been used to max out frame rate, lock it there
-    auto maxed_framerate = (frames_per_second >= 1000) && (display_fps == 2);
-    if (maxed_framerate) { return; }
+    // if Ctrl+E has been used to max out frame rate, switch back to default
+    // when it is passed again.
+    if (frames_per_second >= 1000 && display_fps == 2 && newspd == 1000)
+        newspd = 40;
 
     newspd += play.game_speed_modifier;
     if (newspd>1000) newspd=1000;


sonneveld:
Does this still work for 60fps games?

ivan-mogilko:
I guess it should remember and then restore previous speed.

ivan-mogilko:
I'd rather suggest to not do such hack in SetGameSpeed, because it's exported to script API and we don't want an unexpected behavior there.

Perhaps this should be done where Ctrl+E command is handled, since it's a part of debug feature. Also previous speed may be saved and restored instead of passing hard-coded 40.

sonneveld:
> Perhaps this should be done where Ctrl+E command is handled, since it's a part of debug feature. Also previous speed may be saved and restored instead of passing hard-coded 40.

Ah true. There's the possibility of scripts using it for debugging but I don't think we should remain backwards compatible for that.

Could we store the framerate as a negative number to signify max speed? That way we can easily revert back.  

The save game format does not have to be changed. We could just include the absolute value in the save games. 

For loading older saves with max speed set, we could ignore that value and reuse the current game value. I think that would work for most things unless the game changes the speed throughout the game, or the game is started from the save game (do we still support that?)

ivan-mogilko:
> Ah true. There's the possibility of scripts using it for debugging but I don't think we should remain backwards compatible for that.

Hmm no, that's not exactly what I meant.

I was looking into history of changes in this function, and maybe it should've been done differently from the beginning.

To think of it, "game speed" is a game's property set by script, while maxed framerate is a testing feature of an engine. If we are going to make it toggleable then we should also save requested game speed separately and apply it when exiting "maxed" mode.

Also, for the same reason I am not sure if we should write "maxed mode" in a savegame.

sonneveld:
> Also, for the same reason I am not sure if we should write "maxed mode" in a savegame.

I think we're in agreement, I might not have expressed myself clearly.

ivan-mogilko:
Remade as #1028.

Issue #875, opened 2019-07-20
Installer build issues


morganwillcock:
**Build issues**
(Mainly for #800)

- [x] Wrapper script doesn't work with all Python versions

Since this also required the win32 API module and calls a Windows binary, there doesn't seem much point in requiring Python here. I've already made a replacement script in Powershell that does the same thing.

- [x] Wildcarded source paths means that the installer will successfully build with some ciritical files missing

I've restructured source directories to there is a clearer separation between what is copied in vs what is already there. The only wildcard paths are now for URL links, Templates, and sub-directories of Linux build components (these are not optional though).

- [x] Requires vcredist_x86-9.0.30729.6161.exe to be present, which is now the wrong version and I'm not sure 100% sure whether these are allowed to be distributed

I've re-implemented this as a registry check and also added a checksum verification to the build script to be 100% sure that the version being included is the correct one.

- [x] .NET 4.5 version check and installation

I've updated the registry check and updated the download URL to go to the current  .NET Framework downloads page

- [x] Filename collision for different versions

Filename now uses the longer version (sp text was already included)

**Compiler warnings**
(tested with Inno Setup 6.0.2)

- [x] Warning: Constant "pf" has been renamed. Use "commonpf" instead or consider using its "auto" form.

Renamed to use the 32 bit equivalent (in-case the registry checks ever need the installer to run in 32 bit mode)

- [x] Warning: The [Setup] section directive "PrivilegesRequired" is set to "admin" but per-user areas (userappdata) are used by the script. Regardless of the version of Windows, if the installation is running in administrative install mode then you should be careful about making any per-user area changes: such changes may not achieve what you are intending. See the "UsedUserAreasWarning" topic in help file for more information.

I've removed all the quick launch stuff.

- [x] Warning: Line 145, Column 37: [Hint] Support function "IsComponentSelected" has been renamed. Use "WizardIsComponentSelected" instead.

Renamed.

sonneveld:
Powerscript is probably the better option here. I just have zero experience with it so I stuck with what I knew!

Unless something has changed you should still be able to bundle the redist package. I think Microsoft only explicitly forbids bundling the debug version.

ivan-mogilko:
I may look into InnoSetup warnings; existing script was written for InnoSetup 5 though.

morganwillcock:
I'll probaby go through it today. I was most worried about the user appdata warning, as this implies the logic is broken. I think it is actually just this:
```
[...]

PR #876, opened 2019-07-22
Update installer for 3.5 release

morganwillcock:
As well as the changes described in #875 this also:

- Adds Git ignores for Installer directory
- Switches all website URLs to use HTTPS
- Updates Inno Seup details in the license file
- Stops the web browser being opened as an elevated process if the installer self-elevated
- Adds a status message when installed Visual C++ runtime
- Stops the web browser opening when running in silent mode

Any testing would be appreciated, I've only tried Windows 7 (in various states) and Windows 10 (but was trying not to break it).


morganwillcock:
I think checksum verification may depend on the working directory, so I'll re-check this.

morganwillcock:
Any objections to me merging this now?

sonneveld:
This looks good to me. Did you say that you found someone to test XP?

morganwillcock:
This version of .NET isn't available for XP. Unless you mean, did I test how it fails on XP? I could add a check for the Windows version and tell the installer to exit if the version isn't Vista or newer. I'm not sure what the defaults for WINE are though, it probably identifies itself as XP by default.

morganwillcock:
It looks like this file was intially commited with Unix line endings, but the Inno Setup editor writes Windows lines endings for any changes that are made. I've rebased the changes and switched the line endings, so they are at least consistent.

The other change was to pop-up a message box if the Windows version is too old, but the installer will still carry on running rather than exit.

PR #883, opened 2019-08-03
script interpreter: use modern assembly mnemonics

rofl0r:
since the mnemonics aren't used anywhere except debugging, no particular
care has been exerted when designing them (and basically adding them on
an ad-hoc basis), and many of them are ambiguous, outright misleading,
or simply ugly.

for example, there are different instructions for adding a literal to a
register, and adding to 2 registers, yet they have the same mnemonic, making
it harder to reason what happens.
in the new syntax, these now have distinct mnemonics:
addi reg, literal works on an immediate value (thus the i suffix), whereas
add reg1, reg2 works on 2 registers.
the new syntax uses the i suffix consistently for all operations that work
with immediate values.

then there are e.g. "memread", which works on a 4-byte value, whereas
the equivalent mnemonic to work on a 2-byte word is called "memread.w" and
the one for 1-byte access is called "memread.b".
the new syntax calls them memread4, memread2, and memread1, respectively.

the misleading mnemonic "strcmp" was renamed to "streq", since in C strcmp
means a negated value, e.g. if strcmp returns 1, the strings are *not* equal.
other misleading mnemonics such as "checkbounds" and "memwrite.ptr.0" have
also been renamed to something that accurately hints at what the mnemonic
does.

this is the same syntax as used in agsutils assembler and disassembler, making
the 2 finally compatible (the only thing that differs to agsutils now is
"mov", which was left untouched, since it has the awkward semantics of putting
the destination register as second argument, unlike *all* other mnemonics.
agsutils therefore uses a mnemonic called "mr" (move register) instead, which
swaps the dest and source register in the textual form (but it emits the
bytecode in the original order). support for a non-swapping "mov" instruction
can easily be added to agsutils, to have 100% compatibility to AGS.


fernewelten:
In my opinion, the information in `sccmd_info[]` should also be _exposed_ in some way . It is very awkward that the only way to work with AGS Bytecode instructions is by having to copy code from `Engine/script/cc_instance.cpp` into other code entities. (Knowing which instructions take which parameters, how the instructions are "officially" called, etc. etc.) This forced code duplication is prone to get out of sync.

So perhaps that array could be moved into a dedicated source file, for instance a file `script_common.cpp`. The Engine would reference that dedicated entity through the header file instead of having its own copy, and any other code that wants to work with Bytecode would bind that entity into their executable, too.

What do you think?

ivan-mogilko:
> In my opinion, the information in sccmd_info[] should also be exposed in some way .
> So perhaps that array could be moved into a dedicated source file, for instance a file script_common.cpp, and made available in some way in the source's header. The Engine would reference that dedicated entity through the header file instead of having its own information, and any other code that wants to work with Bytecode would bind that entity (and header file) into their executable, too.

This array may be moved to a more "common" location, of course, if that helps, but I do not think it's correct to rely on this anyway. This is not engine's API; the way data is stored may be engine specific and not suitable for other utilities (especially if they are written in other languages).
IMHO the syncing should be done by use of official specification rather than this code.

Engine/script/cc_instance.cpp
```diff
@@ -125,33 +125,33 @@ const ScriptCommandInfo sccmd_info[CC_NUM_SCCMDS] =
...
+    ScriptCommandInfo( SCMD_FLTE            , "flte"              , 2, kScOpTwoArgsAreReg ),
     ScriptCommandInfo( SCMD_ZEROMEMORY      , "zeromem"           , 1, kScOpNoArgIsReg ),
-    ScriptCommandInfo( SCMD_CREATESTRING    , "newstring"         , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_STRINGSEQUAL    , "strcmp"            , 2, kScOpTwoArgsAreReg ),
-    ScriptCommandInfo( SCMD_STRINGSNOTEQ    , "strnotcmp"         , 2, kScOpTwoArgsAreReg ),
-    ScriptCommandInfo( SCMD_CHECKNULLREG    , "checknull"         , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_CREATESTRING    , "newstr"            , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_STRINGSEQUAL    , "streq"             , 2, kScOpTwoArgsAreReg ),
+    ScriptCommandInfo( SCMD_STRINGSNOTEQ    , "strne"             , 2, kScOpTwoArgsAreReg ),
+    ScriptCommandInfo( SCMD_CHECKNULLREG    , "assert"            , 1, kScOpOneArgIsReg ),
```
sonneveld: This instruction specifically checks for null registers, but `assert` sounds too general. At least `checknull` told us what it did.

Engine/script/cc_instance.cpp
```diff
@@ -80,38 +80,38 @@ struct ScriptCommandInfo
...
-    ScriptCommandInfo( SCMD_MEMREADB        , "memread.b"         , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMREADW        , "memread.w"         , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMWRITEB       , "memwrite.b"        , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMWRITEW       , "memwrite.w"        , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_JZ              , "jz"                , 1, kScOpNoArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMREADB        , "memread1"          , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMREADW        , "memread2"          , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMWRITEB       , "memwrite1"         , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMWRITEW       , "memwrite2"         , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_JZ              , "jzi"               , 1, kScOpNoArgIsReg ),
```
sonneveld: This should stay `jz`, it doesn't really accept a literal but an offset relative to the instruction pointer.

Engine/script/cc_instance.cpp
```diff
@@ -80,38 +80,38 @@ struct ScriptCommandInfo
...
+    ScriptCommandInfo( SCMD_OR              , "lor"               , 2, kScOpTwoArgsAreReg ),
     ScriptCommandInfo( SCMD_CALL            , "call"              , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMREADB        , "memread.b"         , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMREADW        , "memread.w"         , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMWRITEB       , "memwrite.b"        , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_MEMWRITEW       , "memwrite.w"        , 1, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_JZ              , "jz"                , 1, kScOpNoArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMREADB        , "memread1"          , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMREADW        , "memread2"          , 1, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_MEMWRITEB       , "memwrite1"         , 1, kScOpOneArgIsReg ),
```
sonneveld: If we're going to rename the mnemonics, why not just to `memwrite` ? In assembly we can infer the instruction from the argument types. This is typically how the intel assembler works with overloaded mnemonics.

sonneveld:
I think in general the instruction mnemonics do not have to be so specific. The assembler should be able to infer the correct instruction op code from the combination of mnemonic and argument types.  I guess I'm arguing for an intel style syntax vs at&t.

sonneveld:
> IMHO the syncing should be done by the means of official specification rather than this code.

Yes, I think I agree. It would help with writing alternative assemblers if we can agree on the syntax first.

Engine/script/cc_instance.cpp
rofl0r: no, it takes an absolute code address in the current script

Engine/script/cc_instance.cpp
rofl0r: that's precisely one of the main reason i changed the mnemonics: in order to have a single mnemonic for each distinct opcode (as mentioned in the commit message).

rofl0r:
> I think in general the instruction mnemonics do not have to be so specific. The assembler should be able to infer the correct instruction op code from the combination of mnemonic and argument types

that's requiring more code to parse the mnemonics (which nobody apart from agsutils does anyhow), plus makes the disassembly harder to read. explicit is always better than implicit.

Engine/script/cc_instance.cpp
rofl0r: assert precisely documents what the instruction does. if the value is 0, is raises an error, just like the assert everybody knows.
in C you'd write `assert(REG);` and `assert reg` in asm comes closest.

Engine/script/cc_instance.cpp
fernewelten: Are you sure? Then it must have changed _quite_ recently. I'm working with release 3.5.2, and here's the code for the JZ command:
```
case SCMD_JZ:
          if (registers[SREG_AX].IsNull())
              pc += arg1.IValue;
          break;
```

rofl0r: oh ? maybe i'm mistaken here, or maybe the engine changes the relocation from absolute to relative on fixup; can't recall.
it doesn't really matter anyhow, the value is an immediate whether absolute or relative, as opposed to a register number.

Engine/script/cc_instance.cpp
sonneveld: Not quite, it checks if the pointer value is null. I'm not sure it's meant to be used for registers with non-pointer values.

ivan-mogilko: Technically it checks if the any value is 0. But engine is currently hardcoded to report a "string is null error", maybe because this was only case it was used so far.

sonneveld:
> explicit is always better than implicit.

You are being explicit, with the types of the arguments.

> that's requiring more code to parse the mnemonics (which nobody apart from agsutils does anyhow), plus makes the disassembly harder to read.

I would argue `mov ax, [mar]` is clearer and easier to read than `ptrget ax`. You allow the mnemonic and the types of the arguments to determine how to encode the instruction.

Engine/script/cc_instance.cpp
ivan-mogilko: I guess there's a conflict between what the operation is actually doing and what it was used for. E.g. compare with SCMD_CHECKBOUNDS, which is practically testing if integer is between 0 and X, but was used specifically to test array index.

Engine/script/cc_instance.cpp
sonneveld: > maybe i'm mistaken here, or maybe the engine changes the relocation from absolute to relative on fixup; can't recall.

It's stored as relative in the byte code as well. The problem I see here is if we decide to add indirect jumps, we have to define more jmp mnemonic variations. The reader should be able to determine what type of jump it is from the arguments.

Engine/script/cc_instance.cpp
rofl0r: > things may become complicated if these operations will be used for other purposes though

yes, possibly. though i don't see anybody writing asm directly (except for agsutils users, or maybe a future script compiler), so this shouldn't be an issue.

Engine/script/cc_instance.cpp
rofl0r: > problem I see here is if we decide to add indirect jumps, we have to define more jmp mnemonic variations

yes of course. if you add new instructions, you logically also have to add a new mnemonic - at least if you want to have one mnemonic per instruction, which eases readibility and codegen.

> The reader should be able to determine what type of jump it is from the arguments.

well first of all, there aren't really that many readers of asm (i suspect @fernewelten and i are so far the only ones that spent more than an hour reading asm). second, if the mnemonic adds a hint about what the instruction does, it certainly helps readibility, even if theoretically one can guess from the argument what happens.
thirdly, i can't see a good reason why one would like to add more jump types. i can imagine a lot of instructions that (when used by the compiler) could make scripts a good bit faster, but none of them involve jumps.
for example, if there was an instruction `cmpeqi reg, value`, the usual tuple of `li ax, value; cmpeq ax, bx`, often accompanied by register swaps or push/pops could be replaced with a single operation. i have a number of opcodes in mind that could give a good speed boost, but that is a matter for another PR/discussion.

rofl0r:
> I would argue mov ax, [mar] is clearer and easier to read than ptrget ax. You allow the mnemonic and the types of the arguments to determine how to encode the instruction.

i'm not trying to introduce a new asm syntax just for the sake of changing things, but mainly because i'm one of few users of asm, and because the existing syntax is so awkward i needed to redesign it for agsutils, with a lot more thought put into it - and the purpose of this PR is to make ags and agsutils compatible in their syntax, not to split hairs about a single mnemonic or invent yet another new, third syntax just for the sake of changing things, which in the end nobody uses anyway.

rofl0r:
> In my opinion, the information in sccmd_info[] should also be exposed in some way

like here ?
https://github.com/rofl0r/agsutils/blob/master/ags_cpu.h#L5

sonneveld:
> which in the end nobody uses anyway.

Then maybe this pull request isn't required?

> like here ? https://github.com/rofl0r/agsutils/blob/master/ags_cpu.h#L5

My main concern with using any code from agsutils is the license that restricts C++ usage. An open spec would at least allow people to implement with whatever license they choose.

rofl0r:
> Then maybe this pull request isn't required?

it is, because it looks like the new script compiler @fernewelten works on makes some use of assembly (if only for debugging), so it's now or never before the nonsensical default asm syntax creeps into more products.

sonneveld:
> so it's now or never before the nonsensical default asm syntax creeps into more products.

Then lets have a discussion on what sort of syntax we would like.

rofl0r:
> My main concern with using any code from agsutils is the license that restricts C++ usage.

there's nothing copyrightable in that header anyway, but if you prefer so i hereby explicitly release it into the public domain.
the license of the AGSutils code was chosen precisely so C++ zealots like you can't turn it into a C++ clusterfuck with 5 levels of class inheritance and full of boilerplate which is a pain in the ass to read and modify, just like the code in this repo; but instead stays clean, simple, readable and compiles in an instant.
note that there's a linking exception so you can use it in any C++ project, as long as agsutils code itself stays C.

Engine/script/cc_instance.cpp
fernewelten: As far as I can see, the compiler has generated loops that needed to be relocated afterwards, so that was probably the reason to decide that relative jumps are easier to work with. 

Proper assemblers used to have jump labels and jumps to labels, and if the jump instruction was relative, it was the assembler's job to figure out the relative distance. At least that's what I recollect from the z80 / TRS-80 era, ages ago.

fernewelten:
The compiler I am writing generates Bytecode directly, just as the current compiler does. However, I need to be able to understand what the Bytecode does or doesn't do, and for that I disassemble chunks of Bytecode. Separately. There's no part of the compiler that generates assembly instructions, and there's no assembler in the compiler that inputs assembly code and outputs Bytecode.

This means that the compiler won't be impacted, no matter what format of assembly code we decide on.

fernewelten:
As far as I can see, the fixup process happens at runtime. It is a part of the Bytecode machine. So the fixups are really an integral part of the Bytecode. 

In order to reason about assemblies, you need to know that _that_ number has a "global" fixup and _that_ number a "function" fixup, etc. So it would help if that could be indicated as part of the assembly statements. Something like `move AX, 15 fixup datadata`.

fernewelten:
I won't comment on the elegance or gunkiness of C++ right here right now, but the decision to switch programming languages shouldn't be taken lightly. You lose megatons of productivity that is buried in the existing code for all its gunkiness. 

As a case in point, SAP decided about a decade ago to switch from ABAP, a very wordy COBOL-style language, to Java, which was sleek and trendy at the time. They still haven't recovered from that ill-fated move: The pre-existing ABAP code is still their workhorse that pulls nearly the whole weight, so they don't manage to cut that off. And under the hood, the Java code has little to show after tons of man years of programming have been poured into it. 

And then it turned out that Java wasn't so universal, hip and trendy as it was once hyped, and now they are saddled with **two** ugly niche languages and a split world that doesn't mingle well. If SAP hadn't been “too big to fail”, it would have gone bust on that decision.

sonneveld:
> So it would help if that could be indicated as part of the assembly statements. Something like `move AX, 15 fixup datadata`.

I was thinking symbols could be defined at the head of the assembly file (like `import function <scriptfunction>`, `import data <name>` or `data byte <name> <value>`). If the symbols are referred to later in the assembly source, the assembler would know the symbol is a function or data, and to add fix ups. For save compatibility it _might_ be important for the order of function/data imports since they have a given slot index but I'm not certain. The order data is defined in the global data section is definitely important for save compatibility :)

> Let's leave the code base in C++, as it is now.

Agreed. I think the question is how much of modern C++ do we embrace? I'm happy to use some of the more useful library functionality like unique_ptr and the new threading lib.

> Joel Spolsky on rewriting from scratch 

I keep this in mind a lot but there are some costs. They ended up maintaining their own language for years to avoid a rewrite. :)

rofl0r:
> So it would help if that could be indicated as part of the assembly statements. Something like move AX, 15 fixup datadata.

agsutils assembler does it like `li ax, GUIControl::GetAtScreenXY^2`. the fixups are automatically done right according to the value.

> I won't comment on the elegance or gunkiness of C++ right here right now, but the decision to switch programming languages shouldn't be taken lightly

i don't see anyone mentioning switching programming languages ?
if you refer to agsutils, they are standalone programs and can be used without linking to them, or putting them into the source tree here. but yeah, i have this idea that for writing a new ASC frontend, the simplest solution would be to emit asm, which is then processed by the assembler to create the compiled object files, just like gcc creates .s asm files which are then passed to `gas`, the GNU assembler, by invoking the program directly.

here's how an asm file generated by agsdisas looks like:

```asm
.data
char var000000 = 0 ; unreferenced variable, assuming char
char var000001 = 0 ; unreferenced variable, assuming char
char var000002 = 0 ; unreferenced variable, assuming char
char var000003 = 0 ; unreferenced variable, assuming char
string var000004 = 0
int var000204 = .data + 4
...
export int mouse_clicked = 0
int var000881 = 0
int var000885 = 0  
.text   
pseudo_isbuttondown$1: ; 1 args
        sourceline 13
        thisaddr 0
        sourceline 14
        li ax, 1
        ret 
        sourceline 15
        ret 

check_mouse_clicked$0: ; 0 args
        sourceline 17
        thisaddr 13
        sourceline 18
        li mar, @mouse_clicked
        memread4 ax
        mr mar, sp
        memwrite4 ax
        addi sp, 4
        sourceline 19
        li ax, 0
        push ax
        li ax, pseudo_isbuttondown$1
        call ax
        subi sp, 4
        li mar, @mouse_clicked
        memwrite4 ax
        sourceline 21
        mr mar, sp
        subi mar, 4
        memread4 ax
        push ax
        li mar, @mouse_clicked
        memread4 ax
        mr mar, sp
        memwrite4 ax
        addi sp, 4
        sourceline 19
        li ax, 0
        push ax
        li ax, pseudo_isbuttondown$1
        call ax
        subi sp, 4
        li mar, @mouse_clicked
        memwrite4 ax
        sourceline 21
        mr mar, sp
        subi mar, 4
        memread4 ax
        push ax
        li mar, @mouse_clicked
        memread4 ax
        pop bx
        cmpne bx, ax
        mr ax, bx
        subi sp, 4
        ret 
        sourceline 22
        subi sp, 4
        ret 
func$0: ; 0 args
        sourceline 39
        thisaddr 276
        sourceline 40
        li ax, mpos_changed$0
        call ax
        jzi label000000000342
        sourceline 41
        li ax, 1
        push ax
        li ax, 0
        li cx, 0
        muli ax, 1
        add cx, ax
        pop ax
        li mar, @var000413
        add mar, cx
        memwrite1 ax
        sourceline 42
        li ax, "foo"
        farpush ax
        li mar, @var000413
        mr ax, mar
        farpush ax
        setfuncargs 2
        li ax, StrCat
        farcall ax
        farsubsp 2
label000000000342: ; inside func$0, ; referenced by 1 spots
        sourceline 45
        ret 

variadic1$101: ; variadic, 1 fixed args
        sourceline 48
        thisaddr 345
        sourceline 49
        ret 

```
basically, only data and text section need to be emitted, the assembler then creates string, export, import. fixups etc all from analysis of the asm (and the result is much slimmer than the current script compiler's - for example it emits tons of strings that aren't even used).

sonneveld:
Perhaps it would be easily to remove the assembly like mnemonics from the engine debugger? That way the engine does not state an opinion on the syntax.

Engine/script/cc_instance.cpp
```diff
@@ -80,38 +80,38 @@ struct ScriptCommandInfo
 const ScriptCommandInfo sccmd_info[CC_NUM_SCCMDS] =
 {
     ScriptCommandInfo( 0                    , "NULL"              , 0, kScOpNoArgIsReg ),
-    ScriptCommandInfo( SCMD_ADD             , "add"               , 2, kScOpOneArgIsReg ),
-    ScriptCommandInfo( SCMD_SUB             , "sub"               , 2, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_ADD             , "addi"              , 2, kScOpOneArgIsReg ),
+    ScriptCommandInfo( SCMD_SUB             , "subi"              , 2, kScOpOneArgIsReg ),
     ScriptCommandInfo( SCMD_REGTOREG        , "mov"               , 2, kScOpTwoArgsAreReg ),
-    ScriptCommandInfo( SCMD_WRITELIT        , "memwritelit"       , 2, kScOpNoArgIsReg ),
+    ScriptCommandInfo( SCMD_WRITELIT        , "memcpy"            , 2, kScOpNoArgIsReg ),
```
sonneveld: I'm not sure this instruction was intended as a memcpy. It really is used for writing a literal value to memory and will fail if outside the size values of (1,2,4).

The original implementation did use a memcpy but it will read out of bounds memory if > 4 (and I think it's a little buggy as it doesn't take into account big endian systems like the other memwrite instructions do).

It looks like the compiler currently only uses this instruction for clearing pointers, so it would be interesting to see if earlier versions used other literal values.

ivan-mogilko:
Did not have an opportunity to comment on this in the past.

I do believe that having 1 mnemonic per operation is more beneficial than grouping few operation under one mnemonic (only different in types of args), because -
* it's honest, and also seem formally proper, since each operation is represented by a distinct code;
* it's much easier for human to read, since you do not have to remember the potential multiple meanings of same command and how to find out the actual meaning;
* tools probably won't care in either way.

Also, we are speaking of renaming existing operations without changing what they do here, so comparing things like `ptrget ax` and `mov ax, [mar]` [like in this comment above](https://github.com/adventuregamestudio/ags/pull/883#issuecomment-518046081) is not very relevant imho, since these two operations can replace each other only if compiler's logic changes, and one of them becomes deprecated (if i understood implications correctly).

Can't say anything about how well the new names match right now (I'd have to take another look), but, frankly speaking, it seems like arguing about code style at this point, because the practical use of agsscript assembly is very limited, and, guess, whoever decides to work with it will have to learn commands anyway. Perhaps it would be convenient if basic operations had a resemblance to asm commands, and unusual, non-standard ones had distinct names - this would let one tell them apart.

Finally, IMHO it's correct if it is not engine that defines these, because engine as a program does not care about assembly and what these are called (as it works with numeric opcodes).
But otoh I would not remove these from the engine completely, as at the moment this is the only way to log *executed* assembly. While in most case it's easier to just run under debugger, but that works if you know what you are looking for (at least line number, or a script function). In case the culprit is not very clear and there's a complex script, it may be sometimes beneficial to make engine run certain section and then see executed commands printed to a file. I used this several times in the past.

sonneveld:
Personally I'm inclined to leave the mnemonics as is. Having working within the interpreter, I would find it a little confusing to rename instructions. Sure, the names are not perfect but they're fairly short and distinct. Plus we're not likely to add many new instructions in the future. There's also the issue where I don't necessarily agree with all the renames (like `memwritelit` to `memcpy`). I don't really see a large benefit to changing them.

ivan-mogilko:
> Sure, the names are not perfect but they're fairly short and distinct. 

The point of this proposed change was that they are *not* distinct, but often ambiguous.

ivan-mogilko:
The only two problems that I see from above are:
* you may have got used to some names, and don't like this to change
* some of the proposed names don't seem good.

I can understand the first thing, but honestly don't think this will be critical so long as new mnemonics are well matching what they mean. After all, at this point this change would affect 2-3 people max, and only require little exercise to learn new variants.

Regarding second, we may try to suggest better names. Personally there are changes that I like, like "memread.w" to "memread.2" (no need to remember what "w" means), and ones that I don't quite like - such as "land" and "lor", since they look confusing to me, and I'd rather prefered a postfix than prefix.

sonneveld:
> The point of this proposed change was that they are _not_ distinct, but often ambiguous.

I edited my comment to include 'argument types' but I meant to say that in general they're fairly distinct. The ambiguity is consistent in that it's same operation which can involve a literal or a register. I don't find it unusual as it's something that intel assemblers have dealt with.

ivan-mogilko:
> I edited my comment to include 'argument types' but I meant to say that in general they're fairly distinct. The ambiguity is consistent in that it's same operation which can involve a literal or a register. I don't find it unusual as it's something that intel assemblers have dealt with.

Point is, you are looking at this from some perspective of a person who already knows what he's dealing with. But there may be people who are not. I honestly can't understand, what could be the problem with having a distinct mnemonic? It makes it more clear what's going on, but does it add any disadvantages?

sonneveld:
> Point is, you are looking at this from some perspective of a person who already knows what he's dealing with. But there may be people who are not. I honestly can't understand, what could be the problem with having a distinct mnemonic? It makes it more clear what's going on, but does it add any disadvantages?

Fair point, but then why not just rename the couple of ambiguous mnemonic, appending ".r" and ".lit" for example?

ivan-mogilko:
> Fair point, but then why not just rename the couple of ambiguous mnemonic, appending ".r" and ".lit" for example?

Maybe? Personally, I am open for any proposal here, and, like I said, I don't have actual opinion on *particular* name changes yet. I found this PR interesting mainly for idea of making these names more distinct.

sonneveld:
In regards to a minimal name change, the "ambiguous" names are

    2 mov
    2 add
    2 mul
    2 sub
    2 f.add
    2 f.sub

`mov` is intended to accept literal values of int and float. The other instructions require an integer, so to make these distinct you could rename to

    mov
    movl
    add
    addi
    mul
    muli
    sub
    subi
    f.add
    f.addi
    f.sub
    f.subi

This wouldn't change the "feel" of the syntax too much.

ivan-mogilko:
Frankly, I still don't know why is it a priority here to do a minimum of changes. These mnemonics are not a part of any data format (or did I miss something?). They (ones from AGS engine) were probably not used as-is by more than couple of people and few times in the past years. As such, these names are not solidifed in any way.

But then, it's quite possible that along with decouple building tools from the editor there will appear a use for a tool that builds bytecode from ags asm, or someone, like fernewelten, may make a cleaned up version of a compiler, and may want to have these mnemonics. After all, rofl0r's agsutils already had a bunch of users too.

Therefore, in my opinion, the priority should be in making these names clear, consistent and easily reflecting what the operation does.

---

Looking into porposed naming changes, if I understand correctly, rofl0r's suggestion was based on following ideas:
1. Distinct names per operation (I guess we may put a check here now);
2. Changing names that incorrectly reflect what operation does;
3. Make these names closer to standard asm, where applicable; which in theory makes sense IMHO, because AGS vm pretty much simulates asm, and people may compare them.


Besides duplicates, as in your above list, I grouped proposed changes for convenience.
I also did some search on certain instructions to see if their equivalents exists in ASM, because I don't know all of them (for instance, this resource: https://en.wikibooks.org/wiki/X86_Assembly/X86_Instructions)
I tried to list those below in the order from a clear case to ones I have more doubts about.

---

**1.** The dots and underscores in the names (e.g. `f.add` vs `fadd`). Also the fact that in some cases it uses dots to separate type, and in other it uses underscores (`bit_add`).

**2.** `bit_and` and `bit_or` are not consistent with anything else in this list.
Proposed alternative: `and` and `or` for bitwise operations, 'land' (logical-and) and `lor` for logical operations.

**3.** Writing and reading simple values in reg[MAR]: `memread.b`, `memread.w` -- I honestly agree that prefix with number of bytes would be easier to read, like `memread1`,`memread2`,`memread4` and same for `memwrite`.

**4.** Use if "cmp" for (non)equality test operation. The argument for this was that in C "cmp" is associated with non-boolean result, e.g. strcmp is expected to return any negative, positive or 0.
IMHO this makes sense. Proposed `cmpeq` also seem to match existing asm mnemonic.
Operations/mnemonics in question are:
```
SCMD_ISEQUAL -- cmp
SCMD_NOTEQUAL -- ncmp
SCMD_STRINGSEQUAL -- strcmp
SCMD_STRINGSNOTEQ -- strnotcmp
```
Proposed alternative:
```
cmpeq
cmpne
streq
strne
```

**5.** JMP renames, by adding `i` postfix. Here the argument was that they all accept literal ("immediate" in ASM) as a parameter, as opposed to getting the value from the register (EIP in x86).
This seem to be the case in AGS, so probably makes sense too, if we aim consistency with other instructions that accept literals and therefore have `i` postfix?
Operations/mnemonics in question are:
```
SCMD_JZ -- jz
SCMD_JMP -- jmp
SCMD_JNZ -- jnz
```


**6.** CHECKNULL renames to "assert". These may be more controversial, because the suggested renames are not limited to adding postfixes / removing dots.
The proposed changes are these:
```
SCMD_CHECKNULLREG -- checknull --> assert
SCMD_CHECKNULL -- checknull.ptr --> ptrassert
SCMD_CHECKBOUNDS -- checkbounds --> assertlte
```
Here `checknull` vs `assert` seem pretty logical, since assert is more broadly used term. The performed operation is also checking whether the value of register is 0, disregarding its type. In the engine failed `checknull` reports "null string", which is a bit strange, because above "checknull.ptr" reports "null pointer".
`checknull.ptr` is testing value from reg[MAR]. `ptr` postfix is there probably because MAR is expected to store a handle in this case, and so to keep consistency with other instructions meant to work with handles (see below). But, strictly speaking, this instruction is still testing for any 0, the only difference is that it is meant exclusively for MAR.

OTOH `checkbounds` is different. It also "asserts", but tests that `0 >= x < literal`. Frankly, I don't get why is `lte` (<=) postfix is suggested in a rename.
Also, `checkbounds` currently is consistent with `dynamicbounds`, which is a similar check for dynamic array. Perhaps `assertbounds` could work if someone does not like "check"? Idk really.


**7.** Writing and reading managed object handles in reg[MAR]: `memwrite.ptr`, `memread.ptr`, `memwrite.ptr.0` etc.
Ok, these are uncommon, very AGS-specific, and don't have any equivalents in real ASM.
These operations basically convert between managed object pointer and a handle; handle is stored in reg[MAR] as a plain integer.
The proposed renames are `ptrset`, `ptrget`, etc. The argument behind these is probably that these operations are not exactly `memwrite` and `memread`, but they additionally involve a conversion between a handle and actual object pointer.
But, to be honest, I am not particularily convinced that this is necessary, so long as existing names are consistent and give a hint of what they do.


**8.** Separate case is `load.sp.offs` --> `ptrstack`. The instruction writes an address from a stack's element into reg[MAR].
I can agree that `load.sp.offs` is a weird name, and currently you won't know what it does without refering to the engine code. Perhaps, it could be made more consistent to other MAR instructions, like ones mentioned above?


**9.** Shorter names.
```
newstring --> newstr
newarray --> newarr
```
Well... matter of convenience mostly, IMO. Also, this was not suggested, but if we accept these, we could as well rename `newuserobject` into `newobj`, for example.


**10.** Other renames.
`SCMD_LITTOREG -- mov --> li` -- this is sort of consistent to real ASM, I think, where `li` means "Load Immediate", and this is what AGS does there. OTOH I found that some variants of ASM also use `mov` for this too.
`SCMD_REGTOREG -- mov --> mr` -- probably meaning "move reg"? Not sure why this was necessary. Also, I found that in some ASM syntax MR instruction stands for "Multiply Registers".
`SCMD_WRITELIT -- memwritelit --> memcpy` -- I tend to agree that this is not exactly memcpy, but rather a special variant of `memwrite`, where both value and it's size (1, 2 or 4) come from literals. So old name may be ok too.

fernewelten:
As I said: neither the old nor the rewritten compiler generate assembly code or use mnemonics -- they generate Bytecode directly, using the `SCMD_…` macros. 

AFAICS, the current "official" mnemonics are fairly hidden: They reside in an array that isn't declared in any header file. So changing those won't have any impact, at least not on the current compiler or the rewritten one. 

On the other hand, changing the `SCMD_…` macro names *would* impact the compiler as well as the Engine. This would need to be done in sync.

ivan-mogilko:
> As I said: neither the old nor the rewritten compiler generate assembly code or use mnemonics

Of course not, and I did not claim they do. My meaning was, that mnemonics could be used for debugging/logging purposes, e.g. to output or report something in human-readable way.

> On the other hand, changing the SCMD_… macro names would impact the compiler as well as the Engine

But how, they are numbers inside? It's numbers that are important is not it?

ivan-mogilko:
I will elaborate again, the reason I keep addressing this issue is this: these mnemonics were mostly unused within AGS project in practice, but at the same time they were used by the assembler/disassembler tools in agsutils, and some people were already using these tools, as well as assemblies for their needs. So the way I see, this is something that may not concern us much (yet), but concerns _them_.

rofl0r:
> checknull.ptr is testing value from reg[MAR]. ptr postfix is there probably because MAR is expected to store a handle in this case, and so to keep consistency with other instructions meant to work with handles

right, all *ptr mnemonics seem to work on garbage collected/reference-counted objects

> OTOH checkbounds is different. It also "asserts", but tests that 0 >= x < literal. Frankly, I don't get why is lte (<=) postfix is suggested in a rename.

hmm, you seem to have a point there. engine does indeed check x < literal, so the proper mnemonic should be "assertlt", or something else that implies the value is positive additionally (i suppose i misinterpreted comment `#define SCMD_CHECKBOUNDS  46    // check reg1 is between 0 and arg2`)

> 8. Separate case is load.sp.offs --> ptrstack. The instruction writes an address from a stack's element into reg[MAR].
> I can agree that load.sp.offs is a weird name, and currently you won't know what it does without refering to the engine code. Perhaps, it could be made more consistent to other MAR instructions, like ones mentioned above?

since "ptr" is usually used to refer to managed objects, this may be a misnomer indeed, though i found "ptrstack" to be way more to the point than "load.sp.offs"

(btw main reason i removed dots from mnemonics was to simplify parser code)

> SCMD_LITTOREG -- mov --> li -- this is sort of consistent to real ASM, I think, where li means "Load Immediate", and this is what AGS does there. OTOH I found that some variants of ASM also use mov for this too.
> SCMD_REGTOREG -- mov --> mr -- probably meaning "move reg"? Not sure why this was necessary. Also, I found that in some ASM syntax MR instruction stands for "Multiply Registers".

both li (load immediate) and mr (move register) are mnemonics used on powerpc. i used them because original engine uses the same "mov" mnemonic for both operations, even though they have different bytecode (the only such case), and subsequently not to collide with mnemonic meaning as defined in ags code.

> SCMD_WRITELIT -- memwritelit --> memcpy -- I tend to agree that this is not exactly memcpy, but rather a special variant of memwrite, where both value and it's size (1, 2 or 4) come from literals. So old name may be ok too.

right. i'm not too happy with the name either, and from what i've seen engine uses it only to "zeromem", in which case "memset" could be a better name, though not really suiting either because engine code hardcodes it to only work on 1,2,4byte values, and the value, if not <256, would behave differently than memset.
"memwritelit" seems to be a bit clunky though.

ivan-mogilko:
@sonneveld 

What about following change?
As you may notice, below are mostly changes involving adding prefix/postfix, the only ones that actually change the word are equal/not-equal instructions.
[my above comment](https://github.com/adventuregamestudio/ags/pull/883#issuecomment-573361004) could be also used as a reference on reasoning.


| enum | old mnemonic | new mnemonic |
|-------|-----------------|--------------------|
| SCMD_ADD | add | addi |
| SCMD_SUB | sub | subi |
| SCMD_MUL | mul | muli |
|-------|-----------------|--------------------|
| SCMD_FADD | f.add | faddi |
| SCMD_FSUB | f.sub | fsubi |
| SCMD_FMULREG | f.mul | fmul |
| SCMD_FDIVREG | f.div | fdiv |
| SCMD_FADDREG | f.add | fadd |
| SCMD_FSUBREG | f.sub | fsub |
| SCMD_FGREATER | f.gt | fgt |
| SCMD_FLESSTHAN | f.lt | flt |
| SCMD_FGTE | f.gte | fgte |
| SCMD_FLTE | f.lte | flte |
|-------|-----------------|--------------------|
| SCMD_BITAND | bit_and | and |
| SCMD_BITOR | bit_or | or |
| SCMD_AND | and | land |
| SCMD_OR | or | lor |
|-------|-----------------|--------------------|
| SCMD_LITTOREG | mov | movl |
|-------|-----------------|--------------------|
| SCMD_MEMREADB | memread.b | memread1 |
| SCMD_MEMREADW | memread.w | memread2 |
| SCMD_MEMREAD | memread | memread4 |
| SCMD_MEMWRITEB | memwrite.b | memwrite1 |
| SCMD_MEMWRITEW | memwrite.w | memwrite2 |
| SCMD_MEMWRITE | memwrite | memwrite4 |
|-------|-----------------|--------------------|
| SCMD_ISEQUAL | cmp | cmpeq |
| SCMD_NOTEQUAL | ncmp | cmpne |
| SCMD_STRINGSEQUAL | strcmp | streq |
| SCMD_STRINGSNOTEQ | strnotcmp | strne |
|-------|-----------------|--------------------|
| SCMD_JZ | jz | jzi |
| SCMD_JMP | jmp | jmpi |
| SCMD_JNZ | jnz | jnzi |
|-------|-----------------|--------------------|

ivan-mogilko:
Alright, I commited changes as in the above table (c464c74d884c7bb03b85273a5f30eda4e42af61f).

I will close this PR now, as the rest of the changes seem to be too controversial at the moment, but maybe this could be discussed again later, depends on whether these will be used withing our project at all, I guess.

@fernewelten , we may also move these definitions to common library to share with compiler project, if you think it's necessary.

PR #889, opened 2019-08-14
Sort AudioClip select list using the ScriptName property

morganwillcock:
This is a workflow improvement, for selecting AudioClips in game settings and the ViewFrame editor. If you have more than a trivial amount of AudioClips to select from, it is easier to find them if they are sorted alphabetically.

User quote:
> Is it possible to organize all my sound files in this window?
I will most likely add sound to most of my animations in the View editor
however, it's kind of hard to find my sounds when they are all thrown into one window here... not even sorted by alphabet:sweat_smile:
if not possible ... adding sounds to the game will become a nightmare lol

- Easiest and clearest way to do this was using Linq
- I don't know why the AudioClip was being referred to as 'type', I think that was probably just a mistake, so I've also changed that for clarity

### DIFF
--- a/Editor/AGS.Types/PropertyGridExtras/AudioClipTypeConverter.cs
+++ b/Editor/AGS.Types/PropertyGridExtras/AudioClipTypeConverter.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Drawing;
 using System.Drawing.Design;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Have you tested if it actually assigns correct ID when name order is different from ID order?

morganwillcock:
Yes, I checked the 'score audio' from the game settings. The value saved is the AudioClip ID and not the sorted position.

Issue #878, opened 2019-07-24
Automatic font outlines are too thin in higher resolution (800x600 and up)


fernewelten:
This is an issue I reported over a year ago, in the AGS forums bug tracker (not here in github). As far as I can see, it's still open:

Automatic font outlines in LucasArts style have a fixed width of 1 pixel for ttf fonts and 0.6 pixel (I think) for SCI fonts.This is fine for 320x200 games or 640x400 games, but at higher resolutions the texts aren't offset enough from the background and so become **very hard** to read. 

This is so off-putting that I've read of programmers that refuse to use LucasArts mode for higher resolutions just because of this one issue, on the AGS boards.

So, when the option "automatic outline" is selected for some font, it should be possible to also select, in pixels, **how wide** the outline should be. This could be implemented in a fully backward-compatible way if the outline width "0" would select the old calculation and be the default unless changed.

To implement this, the following needs to be done:
1. Logic must be provided that calculates outlines of variable thickness at runtime. One straight-forward way to do it is to only change EIGHT code lines (where `outlineDist` is the outline thickness, already declared in the surrounding code), namely:
In file ags/Engine/ac/display.cpp, around line ~450, from: 
```
wouttextxy(ds, xxp - outlineDist, yyp, usingfont, outline_color, texx);
wouttextxy(ds, xxp + outlineDist, yyp, usingfont, outline_color, texx);
wouttextxy(ds, xxp, yyp + outlineDist, usingfont, outline_color, texx);
wouttextxy(ds, xxp, yyp - outlineDist, usingfont, outline_color, texx);
wouttextxy(ds, xxp - outlineDist, yyp - outlineDist, usingfont, outline_color, texx);
wouttextxy(ds, xxp - outlineDist, yyp + outlineDist, usingfont, outline_color, texx);
wouttextxy(ds, xxp + outlineDist, yyp + outlineDist, usingfont, outline_color, texx);
wouttextxy(ds, xxp + outlineDist, yyp - outlineDist, usingfont, outline_color, texx);
```
to:
```
int const max_sq_euclid_dist =
            outlineDist * outlineDist + outlineDist ;
for (int xdiff = -outlineDist; xdiff <= outlineDist; xdiff++)
    for (int ydiff = -outlineDist; ydiff <= outlineDist; ydiff++)
    {
        if (xdiff * xdiff + ydiff * ydiff > max_sq_euclid_dist)
            continue;
        if (0 == xdiff && 0 == ydiff)
            continue;
[...]

PR #891, opened 2019-08-16
Automatic font outlines

fernewelten:
Have a look: This is proposed code for issue #878 (automatic outlines).
Added two additional fields to fonts:
- `AutoOutlineStyle` (`Rounded`, `Squared`).
- `AutoOutlineThickness` (positive integer, or 0 for default) 
(I couldn't use "1" as the default for `AutoOutlineThickness` because currently, scaled bitmap fonts get an outline that is thicker than 1). 

The fields will only show in the editor if `OutlineStyle==Automatic`.
The field `OutlineFont` will now only show if `OutlineStyle==UseOutlineFont`.

The algo for generating the outline is based on messengerbag's. The text is only rendered _once_ per outline into a local bitmap; from then on, the bitmap is blitted repeatedly. Another local bitmap accumulates the vertical blits. In all, the number of blits is proportional to the thickness of the outline. 

For anti-aliased ttf fonts, a thickness of 2 to 3 seems to work nicely.

What do you think?


rofl0r:
why are there commits of ivan in the commit list ?

ivan-mogilko:
Oh, it seems this is because the PR is based on release-3.5.0 branch, but aimed for ags3 which does not have latest 3.5.0 merged in yet.

fernewelten:
> rebase your feature branch on top of "ags3"?

Can do.

fernewelten:
> Can do.

Cannot do. And I'm to exhausted to fight git right now. It's easier to open a new pull request, this time with no more than the correct commits, under the same short name. And cleaner, too. See #892.

Sorry.

PR #892, opened 2019-08-16
Automatic font outlines

fernewelten:
So, let's try that again, this time without spurious commits (see https://github.com/adventuregamestudio/ags/pull/891#issuecomment-522176713).

Have a look: This is proposed code for issue #878 (automatic outlines).
Added two additional fields to fonts:
- `AutoOutlineStyle` (`Rounded`, `Squared`).
- `AutoOutlineThickness` (positive integer, or 0 for default) 
(I couldn't use "1" as the default for `AutoOutlineThickness` because currently, scaled bitmap fonts get an outline that is thicker than 1). 

The fields will only show in the editor if `OutlineStyle==Automatic`.
The field `OutlineFont` will now only show if `OutlineStyle==UseOutlineFont`.

The algo for generating the outline is based on messengerbag's. The text is only rendered _once_ per outline into a local bitmap; from then on, the bitmap is blitted repeatedly. Another local bitmap accumulates the vertical blits. In all, the number of blits is proportional to the thickness of the outline. 

For anti-aliased ttf fonts, a thickness of 2 to 3 seems to work nicely.

What do you think?

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -137,8 +137,8 @@ enum GameDataVersion
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
-    kGameVersion_350_1          = 51,
-    kGameVersion_Current        = kGameVersion_350_1
+    kGameVersion_351            = 51,
+    kGameVersion_Current        = kGameVersion_351
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -101,6 +101,9 @@ Font custom line spacing.
 Sprites have "real" resolution. Expanded FontInfo data format.
 Option to allow legacy relative asset resolutions.
 
+51 :
+Fonts have adjustable outline
+
 */
 
 enum GameDataVersion
@@ -134,7 +137,8 @@ enum GameDataVersion
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
-    kGameVersion_Current        = kGameVersion_350
+    kGameVersion_350_1          = 51,
+    kGameVersion_Current        = kGameVersion_350_1
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -225,7 +225,7 @@ struct SpriteInfo
 // multiple lines, and similar cases.
 struct FontInfo
 {
-    enum AutoOutlineStyleT : int
+    enum AutoOutlineStyle : int
     {
         kRounded = 0,
         kSquared = 1,
@@ -246,7 +246,7 @@ struct FontInfo
     // When automatic outlining, thickness of the outline (0 = use legacy thickness)
     int           AutoOutlineThickness;
     // When automatic outlining, style of the outline
-    AutoOutlineStyleT AutoOutlineStyle;
+    AutoOutlineStyle AutoOutlineStyle;
 
     FontInfo();
 };
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -225,6 +225,12 @@ struct SpriteInfo
 // multiple lines, and similar cases.
 struct FontInfo
 {
+    enum AutoOutlineStyleT : int
+    {
+        kRounded = 0,
+        kSquared = 1,
+    };
+
     // General font's loading and rendering flags
     uint32_t      Flags;
     // Font size, in points (basically means pixels in AGS)
@@ -237,6 +243,10 @@ struct FontInfo
     int           YOffset;
     // custom line spacing between two lines of text (0 = use font height)
     int           LineSpacing;
+    // When automatic outlining, thickness of the outline (0 = use legacy thickness)
+    int           AutoOutlineThickness;
+    // When automatic outlining, style of the outline
+    AutoOutlineStyleT AutoOutlineStyle;
 
     FontInfo();
 };
--- a/Editor/AGS.Types/Font.cs
+++ b/Editor/AGS.Types/Font.cs
@@ -7,8 +7,8 @@ using System.Xml;
 namespace AGS.Types
 {
     [DefaultProperty("OutlineStyle")]
-    public class Font
-    {
+    public class Font : ICustomTypeDescriptor
+	{
         private int _id;
         private string _name;
         private int _pointSize;
@@ -19,14 +19,16 @@ namespace AGS.Types
         private int _sizeMultiplier = 1;
         private int _verticalOffset;
         private int _lineSpacing;
+		private int _autoOutlineThickness = 0;
+		private FontAutoOutlineStyle _autoOutlineStyle = FontAutoOutlineStyle.Rounded;
 
         public Font()
         {
-            _outlineStyle = FontOutlineStyle.None;
-            _outlineFont = 0;
             _name = string.Empty;
             _pointSize = 0;
-            _fontHeight = 0;
+			_outlineFont = 0;
+			_outlineStyle = FontOutlineStyle.None;
+			_fontHeight = 0;
             _lineSpacing = 0;
         }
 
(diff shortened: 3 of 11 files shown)


Editor/AGS.Types/Font.cs
```diff
@@ -19,14 +19,16 @@ public class Font
         private int _sizeMultiplier = 1;
         private int _verticalOffset;
         private int _lineSpacing;
+		private int _autoOutlineThickness = 0;
```
rofl0r: we seem to have gratuitous whitespace changes here

Editor/AGS.Types/Font.cs
```diff
@@ -19,14 +19,16 @@ public class Font
...
+		private FontAutoOutlineStyle _autoOutlineStyle = FontAutoOutlineStyle.Rounded;
 
         public Font()
         {
-            _outlineStyle = FontOutlineStyle.None;
-            _outlineFont = 0;
             _name = string.Empty;
             _pointSize = 0;
-            _fontHeight = 0;
+			_outlineFont = 0;
```
rofl0r: likewise

Common/ac/game_version.h
```diff
@@ -134,7 +137,8 @@ enum GameDataVersion
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
-    kGameVersion_Current        = kGameVersion_350
+    kGameVersion_350_1          = 51,
```
ivan-mogilko: Version constant should be kGameVersion_351, because this feature will be added after 3.5.0 already. XXX_N naming  was used previously when there were additional format changes after first public (alpha or beta) releases.

Common/ac/gamestructdefines.h
```diff
@@ -225,6 +225,12 @@ struct SpriteInfo
 // multiple lines, and similar cases.
 struct FontInfo
 {
+    enum AutoOutlineStyleT : int
```
ivan-mogilko: Please, don't add suffixes like 'T' to the enum names, we never had this naming convention before.

Engine/ac/display.cpp
```diff
@@ -446,39 +445,112 @@ bool ShouldAntiAliasText() {
...
+            get_fixed_pixel_size(1) : 1;
+        style = FontInfo::kSquared;
+    }
+    if (thickness <= 0)
+        return; 
+
+    size_t const t_width = wgettextwidth(texx, font);
+    size_t const t_height = wgettextheight(texx, font);
+    AGS::Common::Bitmap *outline_stencil =
+        AGS::Common::BitmapHelper::CreateTransparentBitmap(t_width, t_height + 2 * thickness, ds->GetColorDepth());
```
ivan-mogilko: There's no need to use these long notations like AGS::Common::BitmapHelper, and similar. Instead you could add
`using namespace AGS::Common;` at the beginning of the cpp file and omit that everywhere.

Engine/ac/display.cpp
```diff
@@ -446,39 +445,112 @@ bool ShouldAntiAliasText() {
...
-            // if it's a scaled up bitmap font, move the outline out more
-            outlineDist = get_fixed_pixel_size(1);
+// Draw outline that is calculated from the text font, not derived from an outline font
+void wouttextxy_AutoOutline(AGS::Common::Bitmap *ds, size_t font, int32_t color, const char *texx, int &xxp, int &yyp)
+{
+    int thickness = game.fonts.at(font).AutoOutlineThickness;
+    auto style = game.fonts.at(font).AutoOutlineStyle;
+    if (0 == thickness)
+    {
+        // Thickness that corresponds to 1 game pixel
```
ivan-mogilko: Instead of doing this conversion every time when the drawing function is called, I suggest doing it once in the UpgradeFonts instead, see:
https://github.com/adventuregamestudio/ags/blob/ags3/Common/game/main_game_file.cpp#L470

This way you may also remove this special case when thickness is 0.

ivan-mogilko:
I left few comments regarding code and style, will also do proper test later.

Editor/AGS.Types/Font.cs
fernewelten: This file partially has leading tabs, each counting for 4 spaces, and partially leading spaces. This is all mixed throughout the file. And this accounts for the gratuitous whitespace because Github seems to count tabs as 8 spaces.

Speaking of which, I only know the `C++` conventions; what are our current `C#` coding conventions concerning leading whitespace? I plan on doing a change that does nothing but changing **all** of the leading whitespace to our standard, just for this specific file `Font.cs`.

Editor/AGS.Types/Font.cs
fernewelten: See above, https://github.com/adventuregamestudio/ags/pull/892#discussion_r314921723

Editor/AGS.Types/Font.cs
ivan-mogilko: In the new C# code we were using 4 space indents, just like in the engine. The most tabs are leftovers of the old CJ's code that has mixed style.

fernewelten: Okay, pushing that dedicated whitespace change.

rofl0r: > I plan on doing a change that does nothing but changing all of the leading whitespace to our standard

that sounds like a good idea in the first moment, but experience has shown that such huge whitespace-changing commits are very intrusive, it prevents from easily backporting or forward-porting or reversing *any* change that happened before and after such a commit. in my projects i therefore only fix whitespace when it happens to be in an area that needs to be changed during an unrelated commit.

morganwillcock: I think I would probably agree that large whitespace changes are a bad idea, for the reasons above, but also because a lot of the other Editor source files have mixed whitespace characters and mixed line endings so this is isn't an isolated problem. It would probably be better to find some future event (like the final merge of SDL2 changes or some other milestone release) where everyone is aware that any branches they have might be affected.

ivan-mogilko:
I posted this in a wrong place previously, so reposting here:

@fernewelten , I did some tests, and this is what I found in regards to behavior:

1. Project conversion from old format (editor).

* Something I forgot to mention earlier, you also need to increase project format index, after adding new properties.
  It's defined by constant `LATEST_XML_VERSION_INDEX`.
* Is there a reason to have "OutlineThickness" default value = 0 instead of 1? Will it have special meaning rather than being just a extra case when it has to be converted to 1 later?
  Note that right now engine assumes that thickness = 0 indicates old font format and not only changes it to 1, but also changes style to Squared (there's another problem that I mention below).
* There has to be an explicit conversion from old format done in AGS.Editor.Tasks.SetDefaultValuesForNewFeatures. For previous format, all fonts with Auto outline should be set to style = Square thickness = 1.

2. Game data conversion (engine).

* Data conversion in UpgradeFonts currently has incorrect condition. It is applied only if game format is *3.5.0 or higher*, while logically it should be applied if format is *3.5.0 or lower*. Also, in this case there is no need to test for Thickness = 0, because in these old formats thickness was never a thing.
OTOH, if you mean to correct even new format, adjusting thickness = 0 to 1, then this should be better done separately; also in such case it should not force style to Squared like it does now.
I guess the question here is if thickness = 0 should be a valid value in a new format at all (this is also mentioned above).

3. New functionality (engine).

* I found that if I set large thickness, then the engine cannot correctly calculate size of bitmap necessary to store the text line.
In my test this happened when I took font with height = 9 and assigned outline thickness = 9.

ivan-mogilko:
@fernewelten , could you clarify, are you still planning to work on this pr?

Also, your last 3 commits in this branch are confusing, because you are reverting changes that were actually proper ones.

ericoporto:
@fernewelten are you still going to work on this PR?

ivan-mogilko:
The 3.5.0 will be complete very soon (few days, if nothing goes wrong), after that I was thinking that I could help with doing final fixes for this (the ones mentioned in my latest notes above) and merge this. Besides one confusing problem (incorrect bitmap size) it should all be pretty trivial.

ivan-mogilko:
I opened a "remake" (or rather revisit) of this PR here, using most of original fernewelten's commits, and adding suggested fixes: #1018

ivan-mogilko:
Implemented as #1018

Issue #520, opened 2018-10-14
acsetup.cfg contains no game name, settings write to "Saved Games/<game name>/"


morganwillcock:
I was making a replacement launcher, but it seems this isn't possible using just the data from the config file.
The config file doesn't include any identifier for the game, but then writes config changes into a path which includes the game name. This makes it impossible to use another tool to edit the files.

As a side issue, but related to the above, winsetup is writing config data into the "Saved Games" directory of the user profile. Surely this should be going into roaming or local appdata? This isn't save game data and isn't affected by configuring a custom save game path.

ivan-mogilko:
I was not sure where to store these settings and decided to place them in save folder to avoid creating any new paths (custom paths did not exist at that time). The location of the savegame folder is currently determined by the engine based on the game data, but it has to pre-read that folder setting from the game package first before proceeding with the game or running setup dialog.

This of course may be changed if there is a better alternative.

Speaking of config file, I was quite reluctant to rely on its contents in general, because it may be easily edited, this is why I made such decision to read most important properties from the game data instead. Maybe that was another case of overthinking the problem, but I could not figure out what should engine/setup do if this field is missing in config.

Since we are at this, in the past we were discussing how could alternate (non-embedded) setup program get some data from the game or engine (for example, engine's version etc), and we thought that passing some data by standard output may be not a bad idea. For example, if engine is called with something like "--tell gfxdrivers" it will return the list of supported renderers, and so forth. I just thought I'd mention this, since this is also another option to get paths or similar data from the game using engine as a tool to extract it.

morganwillcock:
> Speaking of config file, I was quite reluctant to rely on its contents in general, because it may be easily edited, this is why I made such decision to read most important properties from the game data instead. Maybe that was another case of overthinking the problem, but I could not figure out what should engine/setup do if any of those fields are missing in config.

I would say that, since the config file shipped with the game has some options chosen by the game author, plus the defaults to get the engine to work, it should just be treated as part of the installation (i.e. it is a static component). If someone manually edits it and then the game doesn't start, that would be okay, they can just re-install it. Equally, if someone deletes the vox file and then says they have no sound, they can just re-install it. I'm not actually sure that a .cfg extension would default to opening in a text editor, so I think anyone savvy enough to be fiddling in there would also be savvy enough to fix what they break.

[...]

PR #893, opened 2019-08-17
Implemented basic functionality for --tell command

ivan-mogilko:
For #520

* Engine detects command line arguments starting with `--tell-`, the suffixes are stored in a set as keys.
* If there was at least one such argument, engine will print out requested info to std out and stop. Because some of this info requires data from the game header, printing is done after game files were found and pre-read.
* Information is printed out in JSON format and may contain multiple key/value pairs and groups of those.

TODO and potential issues:
* **[DONE]** Engine prints some other messages to std out, such as its name and version, game location and version, maybe some warnings and errors. Should these be suppressed to simplify parsing for external process?
* **[DONE]** Need to detect early whether it is required to read game at all, because some of the info may not need that (like, if we need to print only list of supported gfx drivers);
* Need to figure out good names for the keys. I called implemented one "config-meta" because it contains settings that help to find and work with config, but maybe there are better alternatives.
* Implement a minimal necessary set of info keys, as noted in #520.

### DIFF
--- a/Common/ac/common_defines.h
+++ b/Common/ac/common_defines.h
@@ -17,8 +17,9 @@
 
 #include "core/platform.h"
 
-#define EXIT_NORMAL 91
-#define EXIT_CRASH  92
+#define EXIT_NORMAL 0
+#define EXIT_ERROR  1
+#define EXIT_CRASH  2
 
 
 #if defined (OBSOLETE)
--- a/Common/debug/out.h
+++ b/Common/debug/out.h
@@ -100,7 +100,8 @@ enum MessageType
     // Convenient aliases
     kDbgMsg_Default             = kDbgMsg_Debug,
     kDbgMsgSet_NoDebug          = 0x001D,
-    kDbgMsgSet_Errors           = 0x0019,
+    kDbgMsgSet_Errors           = 0x0018,
+    kDbgMsgSet_InitAndErrors    = 0x0019,
     // Output everything
     kDbgMsgSet_All              = 0xFFFF
 };
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -86,6 +86,26 @@ StrUtil::ConversionError StrUtil::StringToInt(const String &s, int &val, int def
     return StrUtil::kNoError;
 }
 
+String StrUtil::JsonEscape(const char *cstr)
+{
+    String s;
+    for (; *cstr; ++cstr)
+    {
+        switch (*cstr)
+        {
+        case '\b': s.Append("\\b"); break;
+        case '\f': s.Append("\\f"); break;
+        case '\n': s.Append("\\n"); break;
+        case '\r': s.Append("\\r"); break;
+        case '\t': s.Append("\\t"); break;
+        case '"': s.Append("\\\""); break;
+        case '\\': s.Append("\\\\"); break;
+        default: s.AppendChar(*cstr);
+        }
+    }
+    return s;
+}
+
 String StrUtil::ReadString(Stream *in)
 {
     size_t len = in->ReadInt32();
--- a/Common/util/string_utils.h
+++ b/Common/util/string_utils.h
@@ -55,6 +55,10 @@ namespace StrUtil
     // def_val on failure
     ConversionError StringToInt(const String &s, int &val, int def_val);
 
+    // Creates a new string compliant to JSON key/value format;
+    // assumes that **EVERY** special symbol in the source string must be escaped
+    String          JsonEscape(const char *cstr);
+
     // Serialize and unserialize unterminated string prefixed with 32-bit length;
     // length is presented as 32-bit integer integer
     String          ReadString(Stream *in);
--- a/Engine/debug/debug.cpp
+++ b/Engine/debug/debug.cpp
@@ -97,13 +97,20 @@ const String WarningFileID = "warnfile";
 const String OutputSystemID = "stderr";
 const String OutputGameConsoleID = "console";
 
-void init_debug()
+void init_debug(bool stderr_only)
 {
     // Register outputs
+    if (stderr_only)
+    {
+        PDebugOutput std_out = DbgMgr.RegisterOutput(OutputSystemID, AGSPlatformDriver::GetDriver(), kDbgMsg_None);
+        std_out->SetGroupFilter(kDbgGroup_Main, kDbgMsgSet_Errors);
+        AGSPlatformDriver::SetOutputToErr(true);
+        return;
+    }
     DebugMsgBuff.reset(new MessageBuffer());
     DbgMgr.RegisterOutput(OutputMsgBufID, DebugMsgBuff.get(), kDbgMsgSet_All);
     PDebugOutput std_out = DbgMgr.RegisterOutput(OutputSystemID, AGSPlatformDriver::GetDriver(), kDbgMsg_None);
-    std_out->SetGroupFilter(kDbgGroup_Main, kDbgMsgSet_Errors);
+    std_out->SetGroupFilter(kDbgGroup_Main, kDbgMsgSet_InitAndErrors);
 }
 
 void apply_debug_config(const ConfigTree &cfg)
@@ -115,13 +122,13 @@ void apply_debug_config(const ConfigTree &cfg)
 #ifdef DEBUG_SPRITECACHE
         file_out->SetGroupFilter(kDbgGroup_SprCache, kDbgMsgSet_All);
 #else
-        file_out->SetGroupFilter(kDbgGroup_SprCache, kDbgMsgSet_Errors);
+        file_out->SetGroupFilter(kDbgGroup_SprCache, kDbgMsgSet_InitAndErrors);
 #endif
-        file_out->SetGroupFilter(kDbgGroup_Script, kDbgMsgSet_Errors);
+        file_out->SetGroupFilter(kDbgGroup_Script, kDbgMsgSet_InitAndErrors);
 #ifdef DEBUG_MANAGED_OBJECTS
         file_out->SetGroupFilter(kDbgGroup_ManObj, kDbgMsgSet_All);
 #else
-        file_out->SetGroupFilter(kDbgGroup_ManObj, kDbgMsgSet_Errors);
+        file_out->SetGroupFilter(kDbgGroup_ManObj, kDbgMsgSet_InitAndErrors);
 #endif
         String logfile_path = platform->GetAppOutputDirectory();
         logfile_path.Append("/ags.log");
@@ -173,7 +180,7 @@ void debug_set_console(bool enable)
     if (enable && DebugConsole.get() == nullptr)
     {
         DebugConsole.reset(new ConsoleOutputTarget());
-        PDebugOutput gmcs_out = DbgMgr.RegisterOutput(OutputGameConsoleID, DebugConsole.get(), kDbgMsgSet_Errors);
+        PDebugOutput gmcs_out = DbgMgr.RegisterOutput(OutputGameConsoleID, DebugConsole.get(), kDbgMsgSet_InitAndErrors);
         gmcs_out->SetGroupFilter(kDbgGroup_Main, kDbgMsgSet_All);
         gmcs_out->SetGroupFilter(kDbgGroup_Script, kDbgMsgSet_All);
         if (DebugMsgBuff.get())
--- a/Engine/debug/debug_log.h
+++ b/Engine/debug/debug_log.h
@@ -21,7 +21,7 @@
 #include "platform/base/agsplatformdriver.h"
 #include "util/ini_util.h"
 
-void init_debug();
+void init_debug(bool stderr_only);
 void apply_debug_config(const AGS::Common::ConfigTree &cfg);
 void shutdown_debug();
 
--- a/Engine/main/engine.cpp
+++ b/Engine/main/engine.cpp
@@ -1355,23 +1355,25 @@ bool engine_do_config(const String &exe_path, const ConfigTree &startup_opts)
 extern std::set<String> tellInfoKeys;
 inline static void json_make_block(String &full, const char *name, const char *block, const char *indent = "")
 {
-    if (full.GetLength() > 0 && full.GetLast() == '}')
-        full.AppendChar(',');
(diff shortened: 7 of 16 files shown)


ericoporto:
About suppression of name and version, I think it would be best. This would make the output pure json, easier to parse. Other possiblity is, in such case of suppression, add also these information as elements in this json.

morganwillcock:
OK, I built this and tried it out:
<pre>./ags --tell data
Adventure Game Studio v3.5 Interpreter
Copyright (c) 1999-2011 Chris Jones and 2011-2019 others
ACI version 3.5.0.15

Initializing allegro
Initializing game data
Located game data file: /Users/mwillcock/Downloads/In-Bloom/data/In-Bloom.ags
Opened game data file: game28.dta
Game data version: 49
Compiled with: 3.4.1.14
"config-meta":{"gamename":"In Bloom","defaultconfig":"/Users/mwillcock/Downloads/In-Bloom/data/acsetup.cfg","globalconfig":"./acsetup.cfg","userconfig":"/Users/mwillcock/Library/Application Support/In Bloom/acsetup.cfg"}</pre>

The first issue is, a lot of the benefit of outputting in json is lost when you have to actually locate the json in the the output. So I think it would be best to just output *only* the json, but in a style that could also be processed line-by-line:

<pre>{
	"gamename": "In Bloom",
	"defaultconfig": "/Users/mwillcock/Downloads/In-Bloom/data/acsetup.cfg",
	"globalconfig": "./acsetup.cfg",
	"userconfig": "/Users/mwillcock/Library/Application Support/In Bloom/acsetup.cfg"
}</pre>

i.e. if doing something where there is no JSON decoder you've at least got a chance of reading the data by searching for the line that starts with the key you are interested in. I'm not saying this is promoted way to use the output, just that `grep '"gamename"'` might be good enough when you have nothing better.

Associated with this, I don't think there should be any output going to stdout other than the JSON, so all the preceding lines shouldn't be present (they should be part of the JSON output if they are relevant), and if there is an error it needs to go to stderr. It should also set the appropriate error code (not just for this --tell command, but whenever there is a query or launch failure). The usage message can still go to stdout, because that should only be present on an error or when requested.

As a psuedo example of what I mean, when the errors go to correct place and you know whether an error occurred or not, there should be no additional work needed to process the JSON string:
<pre>process = new Process("ags --tell data");
process.Run();

if (process.exit_code == 0) {
    # stdout is guaranteed to be JSON only
    json.Decode(process.stdout);
} else {
    Print("Something went wrong");
    # print the error but not the usage message
    Print(process.stderr);
}</pre>

morganwillcock:
Here is a working (probably Linux and macOS) script that could be used to test the changes that I suggest. This tests the error codes, splitting the output between stdout and stderr, and reading the JSON without a JSON decoder.

<pre>#!/bin/sh

./ags --tell data >stdout 2>stderr

if [ $? -eq 0 ]; then
	printf "Game name: "
	sed -E '/^[[:blank:]]"gamename":/ { s/.*"(.*)",?$/\1/; q; }; d' stdout
else
	echo "An error occured!"
	cat stderr
fi</pre>

**Where success is equivalent to:**
<pre>#!/bin/sh

echo '{
	"gamename": "In Bloom",
	"defaultconfig": "/Users/mwillcock/Downloads/In-Bloom/data/acsetup.cfg",
	"globalconfig": "./acsetup.cfg",
	"userconfig": "/Users/mwillcock/Library/Application Support/In Bloom/acsetup.cfg"
}'</pre>

**And an error is equivalent to:**
<pre>#!/bin/sh

echo Error reading game data > /dev/stderr
echo Usage:
echo This is how you use it
exit 1</pre>

**Output on success:**
<pre>Game name: In Bloom</pre>

**Output on failure:**
<pre>An error occured!
Error reading game data</pre>

ivan-mogilko:
Ammended, please try now.

morganwillcock:
> PS. Also, does it have to add \t to indent json, or linebreaks is enough?

Well, it depends whether you want it to be easier to read by eye, vs slightly more complex to parse. I guess that since my argument is to parse with as a basic a tool as possible, lack of whitespace is probably a good thing as it means you don't have to check for optional characters.

Technically, this isn't [valid JSON](https://jsonlint.com/) though, because of the first line:

<pre>$ ./ags --tell data
"config-meta":
{
"gamename":"In Bloom",
"defaultconfig":"/Users/mwillcock/Downloads/In-Bloom/data/acsetup.cfg",
"globalconfig":"./acsetup.cfg",
"userconfig":"/Users/mwillcock/Library/Application Support/In Bloom/acsetup.cfg"
}</pre>

And on failure the return code is still 0:

<pre>$ ./ags --tell data2
ERROR: Unable to determine game data.
Defined game location is not a valid path.
Path: '/Users/mwillcock/Downloads/In-Bloom/data2'
Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]

Options:
  --fps                        Display fps counter
  --fullscreen                 Force display mode to fullscreen
  --gfxdriver <id>             Request graphics driver. Available options:
                                 ogl, software
  --gfxfilter <filter> [<scaling>]
                               Request graphics filter. Available options:
                                 hqx, linear, none, stdscale
                                 (support differs between graphic drivers);
                                 scaling is specified by integer number
  --help                       Print this help message and stop
  --hicolor                    Downmix 32bit colors to 16bit
  --log                        Enable program output to the log file
  --no-log                     Disable program output to the log file,
                                 overriding configuration file setting
  --version                    Print engine's version and stop
  --windowed                   Force display mode to windowed

Gamefile options:
  /dir/path/game/              Launch the game in specified directory
  /dir/path/game/penguin.exe   Launch penguin.exe
  [nothing]                    Launch the game in the current directory

$ echo $?
0
</pre>

Also the error "ERROR: Unable to determine game data." isn't going to stderr:

<pre># redirect stdout but don't touch stderr
./ags --tell data2 > /dev/null</pre>

Also, I wasn't sure why the defined error codes were negative (unless I misunderstand how they are used):
<pre>#define EXIT_NORMAL 0
#define EXIT_ERROR  -1
#define EXIT_CRASH  -2</pre>

0, 1, and 2 would be fine.

ivan-mogilko:
I did some additional fixes. TBH don't know what was wrong with return code earlier, it correctly displays non-0 value for me when I test this out.

There's currently one issue that if failing to find a game it will still print "help" into stdout. I will think more, maybe redo previous commits in different way to guarantee such does not happen by mistake.

morganwillcock:
Ah, I was assuming the result would be a flat name space rather than having a hierarchy, so to contradict myself slightly, if you have named sections like "config-meta" or "graphics-driver" the indent would be beneficial to indicate how deep in the structure you are. So for example, given a single line from here:
<pre>{
	"config-meta": {
		"gamename": "In Bloom",
		"defaultconfig": "/Users/mwillcock/Downloads/In-Bloom/data/acsetup.cfg",
		"globalconfig": "./acsetup.cfg",
		"userconfig": "/Users/mwillcock/Library/Application Support/In Bloom/acsetup.cfg"
	}
}</pre>
...the tab count also gives you the depth within the structure.

Printing errors to stderr seems to work now:
<pre>$ ./ags --tell data2 > /dev/null
ERROR: Unable to determine game data.
Defined game location is not a valid path.
Path: '/Users/mwillcock/Downloads/In-Bloom/data2'</pre>

The return code on failure is still 0 for me:
<pre>./ags --tell data2
ERROR: Unable to determine game data.
Defined game location is not a valid path.
Path: '/Users/mwillcock/Downloads/In-Bloom/data2'
Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]

Options:
  --fps                        Display fps counter
  --fullscreen                 Force display mode to fullscreen
  --gfxdriver <id>             Request graphics driver. Available options:
                                 ogl, software
  --gfxfilter <filter> [<scaling>]
                               Request graphics filter. Available options:
                                 hqx, linear, none, stdscale
                                 (support differs between graphic drivers);
                                 scaling is specified by integer number
  --help                       Print this help message and stop
  --hicolor                    Downmix 32bit colors to 16bit
  --log                        Enable program output to the log file
  --no-log                     Disable program output to the log file,
                                 overriding configuration file setting
  --version                    Print engine's version and stop
  --windowed                   Force display mode to windowed

Gamefile options:
  /dir/path/game/              Launch the game in specified directory
  /dir/path/game/penguin.exe   Launch penguin.exe
  [nothing]                    Launch the game in the current directory

$ echo $?
0</pre>

> There's currently one issue that if failing to find a game it will still print "help" into stdout. I will think more, maybe redo previous commits in different way to guarantee such does not happen by mistake.

I think as long as the error code is always correct, and the usage message always go to stdout (if this is the 'help' you are talking about), this is probably the most straight forward way to separate the help from the errors. So I think usage messages should only ever go to stdout.

ivan-mogilko:
> I was assuming the result would be a flat name space rather than having a hierarchy, so to contradict myself slightly, if you have named sections like "config-meta" or "graphics-driver" the indent would be beneficial to indicate how deep in the structure you are. 

Probably I missed something earlier. Do you mean the problem is that it writes the output as a named group? or that it does not have tabs?

Currently it supports multiple tell commands in row, and if called just like --tell it prints all known ones. Since there may be multiple groups of keys I thought having them grouped would be logical to prevent occasional naming conflicts.

> The return code on failure is still 0 for me:

Ok, I am confused. I ran practically same commands as you and get
`
$ ./ags --tell data2
ERROR: Unable to determine game data.
Defined game location is not a valid path.
<...>

$ echo $?
1
`

morganwillcock:
> Currently it supports multiple tell commands in row, and if called just like --tell it prints all known ones. Since there may be multiple groups of keys I thought having them grouped would be logical to prevent occasional naming conflicts.

Yes, I think if grouping things use an indent so you can measure the depth when given any line:
<pre>{
	"config-meta": {
		"gamename": "In Bloom",
		"globalconfig": "./acsetup.cfg"
	},
	"graphics": {
		"driver": "software",
		"screen_width": "1920",
		"screen_height": "1080"
	}
}</pre>

If you just have a flat structure, then don't indent:
<pre>{
"gamename": "In Bloom",
"globalconfig": "./acsetup.cfg",
"graphics_driver": "software",
"screen_width": "1920",
"screen_height": "1080"
}</pre>

I guess there should also be some version number of the actual JSON data too, in-case the format ever needs to change or a value is changed (like graphics driver may return different values or the game gains a GUID).

> Ok, I am confused. I ran practically same commands as you and get

It must be a platform specific problem, if I try on Linux it returns 1, it I try on macOS it returns 0.

ivan-mogilko:
> It must be a platform specific problem, if I try on Linux it returns 1, it I try on macOS it returns 0.

This is out macOS port, maybe its wrapper intercepts the return value?

morganwillcock:
It is just a binary named 'ags', I don't think there is any wrapper around it?

ivan-mogilko:
I mean, in the patched allegro code, it wraps "main" function from ags engine.
Unless there's another wrapper with objective-C code, or is it only used in iOS port?

morganwillcock:
I just cloned https://github.com/ivan-mogilko/ags-refactoring/tree/release-3.5.0--command-tell and built that.

ivan-mogilko:
Yes, but where does allegro library come from?

morganwillcock:
The AGS repo as far as I know:
https://github.com/ivan-mogilko/ags-refactoring/blob/release-3.5.0--command-tell/CMake/FetchAllegro.cmake
I built it with cmake, which downloads and builds the dependencies.

ivan-mogilko:
Perhaps @sonneveld would know more about this, but at a quick glance the return value of Allegro's main is completely ignored when it's called from objective-C wrapper:
https://github.com/adventuregamestudio/lib-allegro/blob/allegro-4.4.3.1-agspatch/src/macosx/main.m#L143

ivan-mogilko:
@morganwillcock 
I remade the branch squashing commits and added json "beautification" (tabs).

morganwillcock:
Reading a little, the call to `NSApp terminate` looks for a delegate `applicationShouldTerminate`. There are examples of this just calling exit:

https://github.com/Shufflepuck/MacAdminQuitApp/blob/e6094de337a88b358a32d6def9d2d04f85a40a7d/MacAdminQuitApp/AppDelegate.swift#L121-L130

Similar response is here:
https://stackoverflow.com/a/2745585
(standard terminate isn't critical and a difference of behaviour for a 'force quit' is deemed problematic)

ivan-mogilko:
I guess macos port should be separate issue, unfortunately I don't have time right now to look into this myself (and I don't own mac to test anything).

> I guess there should also be some version number of the actual JSON data too, in-case the format ever needs to change or a value is changed (like graphics driver may return different values or the game gains a GUID).

Will it be sufficient to output engine version, which may serve as a format reference?

Also, I noticed that in all json examples that you post you keep opening bracket on same line e.g.:
<pre>
        "config-meta": {
		"gamename": "In Bloom",
		"globalconfig": "./acsetup.cfg"
	},
</pre>
Is this a desired format, or not significant?

ericoporto:
Traditionally JSON has the opening bracket on the same line. You can use a [JSON linter](https://jsonlint.com/) to check validity of json outputs. 

```
{
"config-meta": {
		"gamename": "In Bloom",
		"globalconfig": "./acsetup.cfg"
	}
}
```

ivan-mogilko:
I just realize that what I do here is a dirty implementation of a very raw idea.
For instance, there's already a ConfigTree object where engine gathers data for config. It's read and written in INI format, therefore could be reused to write into JSON as well.
Same, there are already names for config options, that could be reused in json of course to match with game config.

morganwillcock:
Outputting in INI format would also be fine and is easier to parse. If it was easier to just dump the ConfigTree object using the INI writer and then also print some extra sections, I think that would be fine too.

ivan-mogilko:
INI has only 1 level of nesting. Can that be a problem?

For example, what if at some point you'd want to have --tell-x command which returns full user config, which also consists of groups and values in them. How do you distinct config from other data then?

morganwillcock:
I think normally you would prefix or postfix the sections, or use a section delimiter to indicate a parent object. This sounds a bad idea but in practice it can work quite well. Here is a good example of how it might be done:

http://orx-project.org/wiki/en/orx/config/syntax

ivan-mogilko:
New pull request: #897

PR #895, opened 2019-08-23
Add Editor preference for automatic script reload

### LINKED ISSUE #867: Can't edit a file outside of AGS simultaneously
Screenshot from 2019-07-09 18-46-09

When I load a file both on the AGS Editor and on another program, and save it on the other program, I get a black window on AGS and the Editor freezes.

I am though using AGS Editor through Wine on Linux, so I don't know if this is the reason I am getting this. AGS Editor used is 3.5.0.14.

morganwillcock:
This is mostly for #867, but also cleans up the spacing and alignment on the preferences form. If a script is modified externally, one of the following preferences is applied:

- `Always ask what to do`    <-- this is the default
- `Always reload the file`
- `Never reload the file`

So hopefully, if this is set to `Always reload the file` it should fix #867.

### DIFF
--- a/Editor/AGS.Editor/Entities/EditorPreferences.cs
+++ b/Editor/AGS.Editor/Entities/EditorPreferences.cs
@@ -25,6 +25,14 @@ namespace AGS.Editor.Preferences
     }
 
     [Flags]
+    public enum ReloadScriptOnExternalChange
+    {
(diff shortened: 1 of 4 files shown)


ericoporto:
Thank you so much! :heart:

PR #897, opened 2019-08-25
Implemented basic functionality for --tell command (take 2, INI version)

ivan-mogilko:
For #520 (attempt 2)

* Engine detects command line arguments starting with `--tell-`, the suffixes are stored in a set as keys.
* If there was at least one such argument, engine will print out requested info to std out and stop. Because some of this info requires data from the game header, printing is done after game files were found and pre-read.
* Information is printed out in INI format, just like the game config, and may contain multiple sections of key/value pairs.
* This is not very common for INI, but currently the output also has "nested" sections which are indicated by '@' character(s) in their names. This is currently used to group number of sections together and give an extra leeway in avoiding possible naming conflicts in the future. The main reason for this addition was that one of the supported commands is "--tell-config" that prints all the parsed user config, which contains multiple sections on its own.

TODO and potential issues:
* **[DONE]** Engine prints some other messages to std out, such as its name and version, game location and version, maybe some warnings and errors. Should these be suppressed to simplify parsing for external process?
* **[DONE]** Need to detect early whether it is required to read game at all, because some of the info may not need that (like, if we need to print only list of supported gfx drivers);
* **[sort of]** Need to figure out good names for the keys.
* **[DONE]** Implement a minimal necessary set of info keys, as noted in #520.

Currently supported keys:
* --tell or --tell-all: prints everything;
* --tell-config: prints config (must locate game data and parse user config)
* --tell-configpath: prints paths to all 3 configs (default, global and user) and other info necessary to locate these (like game name - to find user data dirs).
* --tell-graphicdrivers: list of supported gfx drivers.

### DIFF
--- a/Common/ac/common_defines.h
+++ b/Common/ac/common_defines.h
@@ -17,8 +17,9 @@
 
 #include "core/platform.h"
 
-#define EXIT_NORMAL 91
-#define EXIT_CRASH  92
+#define EXIT_NORMAL 0
+#define EXIT_ERROR  1
+#define EXIT_CRASH  2
 
 
 #if defined (OBSOLETE)
--- a/Common/ac/gamesetupstruct.h
+++ b/Common/ac/gamesetupstruct.h
@@ -61,6 +61,7 @@ struct GameSetupStruct: public GameSetupStructBase {
     // TODO: why we do not use this in the engine instead of
     // loaded_game_file_version?
     int               filever;  // just used by editor
+    Common::String    compiled_with; // version of AGS this data was created by
     char              lipSyncFrameLetters[MAXLIPSYNCFRAMES][50];
     AGS::Common::PropertySchema propSchema;
     std::vector<AGS::Common::StringIMap> charProps;
--- a/Common/debug/out.h
+++ b/Common/debug/out.h
@@ -100,7 +100,8 @@ enum MessageType
     // Convenient aliases
     kDbgMsg_Default             = kDbgMsg_Debug,
     kDbgMsgSet_NoDebug          = 0x001D,
-    kDbgMsgSet_Errors           = 0x0019,
+    kDbgMsgSet_Errors           = 0x0018,
+    kDbgMsgSet_InitAndErrors    = 0x0019,
     // Output everything
     kDbgMsgSet_All              = 0xFFFF
 };
--- a/Common/util/ini_util.cpp
+++ b/Common/util/ini_util.cpp
@@ -96,6 +96,23 @@ void IniUtil::Write(const String &file, const ConfigTree &tree)
     writer.ReleaseStream();
 }
 
+void IniUtil::WriteToString(String &s, const ConfigTree &tree)
+{
+    for (ConfigNode it_sec = tree.begin(); it_sec != tree.end(); ++it_sec)
+    {
+        const String &sec_key = it_sec->first;
+        const StringOrderMap &sec_tree = it_sec->second;
+        if (!sec_tree.size())
+            continue; // skip empty sections
+        // write section name
+        if (!sec_key.IsEmpty())
+            s.Append(String::FromFormat("[%s]\n", sec_key.GetCStr()));
+        // write all items
+        for (StrStrOIter keyval = sec_tree.begin(); keyval != sec_tree.end(); ++keyval)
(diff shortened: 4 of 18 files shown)


ericoporto:
Screenshot_20190825-213022_bVNC Free

Pretty cool, it needs to be added to `ags --help`

morganwillcock:
Would it be possible to get this section as `--tell-engine`:
<pre>[engine]
name = Adventure Game Studio run-time engine
version = 3.5.0.15</pre>
...instead of outputting it with every other section?

And would it be possible to remove the count key from the `--tell-graphicdrivers` output, since it means you would have to start testing the keys based on some criteria to establish whether they are keys or key metadata (test if the key is an integer and the value is a string to verify that the value is actually a graphics driver, in this case).

ivan-mogilko:
> Would it be possible to get this section as --tell-engine:
> ...instead of outputting it with every other section?

Initially I added this because you asked for some kind of format version, so I thought having engine version may serve as a hint.

morganwillcock:
It is good to have, but potentially you would just want this item by itself to know what else you want to query, but after that you probably don't need it again. Or you just ask for `--tell-all`.

But if you print as its own section, then I think it follows the pattern that `--tell-<section>` will print `[<section>]` or a delimited variant `[<section>@<subsection>]`.

ivan-mogilko:
I did the changes.

ivan-mogilko:
Come on, you can post the plain output instead of screenshots.

Yes I forgot to adjust the help this time.

morganwillcock:
A few things I noticed, perhaps one is problematic.

<hr>

The first is just whether to make the section name and argument name plural or not. Right now there is `--tell-configpath` and `tell-graphicdrivers`, where both can return multiple values (either paths of driver names). So it is probably best to standardise on whether these arguments and sections names end in an 's' or not.

<hr>

The second is that the game name is listed under `tell-configpath`, but the game name isn't a config path. There was also some other game related information that doesn't get printed. Would it be possible to have `--tell-data` (or maybe `--tell-game` ? ) and an separate section `[data]` that is something like:

<pre>[data]
gamename = In Bloom
version = 49
compiledwith = 3.4.1.14</pre>

...or is this extra data not parsed at this point?

<hr>

Third problem, and this is probably not straightforward, on platforms that don't support the global config is it effectively disabled by using '.' as the location in the base class? On Linux you won't see any issue, as the values comes out to this:

<pre>default = /home/mwillcock/Downloads/In-Bloom/data/acsetup.cfg
global = /home/mwillcock/.local/share/ags/acsetup.cfg
user = /home/mwillcock/.local/share/ags/In Bloom/acsetup.cfg</pre>

... so given these values, it is easy to see whether these files exist or not. But if you try the same thing on macOS you get a global config path of:

<pre>global = ./acsetup.cfg</pre>

...and this is impossible to test, because where '.' is will change depending on the current working directory. So under some circumstances ./acsetup may exist, but it could be for a completely different game, or maybe it is the one in the game directory or the user one, and you mistakenly process it as the global one. It is feasible that any kind of wrapper process will be running from somewhere else, as demonstrated by the Linux launch wrapper being called the itch.io client.

I guess I don't know what to suggest to fix this, if all valid config paths are guaranteed to be absolute, perhaps it is best to omit any config paths which are relative?

morganwillcock:
Sorry, one bit I forgot, it doesn't say the path to the game file either. So if the engine locates the path by looking inside 'data' it would be good to know which file it is using as the game.

e.g. `/home/mwillcock/Downloads/In-Bloom/data/In-Bloom.ags`

<pre>[data]
gamename = In Bloom
filename = /home/mwillcock/Downloads/In-Bloom/data/In-Bloom.ags
version = 49
compiledwith = 3.4.1.14</pre>

ivan-mogilko:
Ammended.

morganwillcock:
The `--help` message is out of sync again (`--tell-graphicdrivers` vs `--tell-graphicdriver`), but other than that I think everything is covered.

Regarding the macOS error code, a solution was more obvious when I actually ran the program, rather than tried to think of a solution. As well as trying to look at what went to stdin, If stderr isn't empty then there was an error. So I think that is probably good enough to tell the difference, when you know the exit code isn't going to be set.

PR #901, opened 2019-08-27
Updates to script file remapping and path token rule

ivan-mogilko:
NOTE: path remapping is only done in backwards-compatibility mode - when running old games where script was opening files for writing in the game directory.

1. Introduced a config option that tells where to remap such file operations:
<pre>
[override]
remap_installdir = appdatadir; remaps to $APPDATADIR$
remap_installdir = savegamedir (or anything invalid); remaps to $SAVEGAMEDIR$
</pre>
Remapping is done to $SAVEGAMEDIR$ by default now.

2. Completely deny using paths without tokens since 3.5.0. In the past I let still use these, and for writing they were remapping to $APPDATADIR$ with a warning, but this behavior was not well understood and causing confusion, so was a mistake on my part.
So, considering this, and that remapping rules are now changing, I see no other good option than to require game developers explicitly point where they want to read and write files every time ($INSTALLDIR$, $SAVEGAMEDIR$ or $APPDATADIR$).

### DIFF
--- a/Editor/AGS.Editor/DataFileWriter.cs
+++ b/Editor/AGS.Editor/DataFileWriter.cs
@@ -533,7 +533,7 @@ namespace AGS.Editor
             options[NativeConstants.GameOptions.OPT_NATIVECOORDINATES] = (game.Settings.UseLowResCoordinatesInScript ? 0 : 1);
             options[NativeConstants.GameOptions.OPT_WALKONLOOK] = (game.Settings.WalkInLookMode ? 1 : 0);
             options[NativeConstants.GameOptions.OPT_DISABLEOFF] = (int)game.Settings.WhenInterfaceDisabled;
-            options[NativeConstants.GameOptions.OPT_SAFEFILEPATHS] = 1; // always use safe file paths in new games
+            options[NativeConstants.GameOptions.OPT_SAFEFILEPATHS] = 2; // always use safe file paths and require dir token
             options[NativeConstants.GameOptions.OPT_DIALOGOPTIONSAPI] = (game.Settings.UseOldCustomDialogOptionsAPI ? -1 : 1);
             options[NativeConstants.GameOptions.OPT_BASESCRIPTAPI] = (int)game.Settings.ScriptAPIVersionReal;
             options[NativeConstants.GameOptions.OPT_SCRIPTCOMPATLEV] = (int)game.Settings.ScriptCompatLevelReal;
--- a/Engine/ac/file.cpp
+++ b/Engine/ac/file.cpp
@@ -323,15 +323,18 @@ bool ResolveScriptPath(const String &orig_sc_path, bool read_only, ResolvedPath 
 
         // For games which were made without having safe paths in mind,
         // provide two paths: a path to the local directory and a path to
-        // AppData directory.
+        // safe writeable directory (aka user path).
         // This is done in case game writes a file by local path, and would
-        // like to read it back later. Since AppData path has higher priority,
-        // game will first check the AppData location and find a previously
+        // like to read it back later. Since user path has higher priority,
+        // game will first check the user location and find a previously
         // written file.
         // If no file was written yet, but game is trying to read a pre-created
         // file in the installation directory, then such file will be found
         // following the 'alt_path'.
-        parent_dir = MakeAppDataPath();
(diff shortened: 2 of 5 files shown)


i30817:
I just have a question, the 'read_only' argument on that function is derived from the version of the editor that created the game right? It's not something the dev can set or something related to the 'actual' read only property; in short, all new games never are supposed to write to the game dir, even specifying `GameInstallRootToken` right?

I'm asking this because if the fallback is going to be early exited by the `else if(game.options[OPT_SAFEFILEPATHS] > 1)` section, which i assume is set to all new games from this version of the editor on, if they try to use `GameInstallRootToken` and `read_only` is false, the engine might try to write there and fail on a **actual** read only dir.

ivan-mogilko:
So, to put it shortly, you suggest to write into $SAVEGAMEDIR$ by default if there's no token, for old and new games?

i30817:
Yes. It doesn't matter because the ability to be 'shared' is not fundamental to games working. People don't use the same writable datafiles at the same time, that idea is nonsense in this engine. Moreover, the common thing is to *not* share because otherwise it's a logical bug _very often_.

So at most, changing it that way will make some computers with multiple users playing the same game not share some editor maps (because the dev forgot to use the token, or it was before the token was available). If it was before the token was available... well, it's as designed. If he forgot about it, that's too bad but not as bad as the multitude of games doing 'let's share achievements and checkpoints' by accident.

messengerbag:
I think it's better to make $APPDATADIR$ not point to a shared-user location by default. If game devs or players want that for some reason, they should (on Windows) change it in the setup.

i30817:
Imo that's just a question of the readme saying explicitly that it's supposed to be for *writable, shared* locations, in that order and that the second may not be possible in some platforms, but that the default no token write will always try to write to a writable location (and gloss over that the first read before a write may be to another location entirely since they can't tell, hopefully).

ivan-mogilko:
Well, this is going further and further... There are already games made with 3.5.0, idk how people react if we just begin changing things and rules like this. And AGS manual was stating for years that `$APPDATADIR$, which allows you to write/read files to a folder on the system which is accessible by and shared by all users`.

ivan-mogilko:
OTOH, as a temporary solution, I could just add the remap config option and leave everything else as it is. Since there's a global config, you could add this remap to savegamedir there and that will work for all the games.

For the next major version the path rules could be revised completely.

i30817:
Remaps are useless for filesystem portability if they're not resolved at runtime.

ie: i have a game on  a pen, and i move the pen to another user on another computer, i do not want the remap to fail because the original user no longer exists.

So a option `mapappdir2save=1` is better imo. I'd use it on all games.

But i honestly think you're exaggerating the amount devs care. They might complain about losing settings and achievements ... once, until you tell them in the forum to copy over the dirs over the save dir and forget about it.

ivan-mogilko:
> So a option mapappdir2save=1, or a symbolic option to place on the RHS of shared_data_dir that means 'the engine savegame dir, whatever it is at this time' is a better 'hack'. 

Alright, to clarify again, you mean, instead of all the above, let user set custom paths in config pointing not to literal paths but to tokens, like for example:
<pre>
[misc]
shared_data_dir = $user_data_dir
</pre>
?

i30817:
Now i'm vacillating on the idea for a simple reason. If you merge the other PR, that alters the linux `GetAllUsersDataDirectory` to point to something other than the (readonly) game dir on squash, my problems with that get solved.

It might be better to close this PR until you decide if you want to get rid of $APPDATADIR$, since that other PR is enough for my use case and i only noticed the mess of a idea that is a writable, shared dir while investigating it. If you feel it's too radical to change that in this release, if the other PR is merged, that's enough for me (though it would crash on android, i don't use android ags, yet).

i30817:
Yeah, i'm pretty convinced. If you feel you can do a better design when you're not forced to rush for release, close this and just merge the other.

morganwillcock:
I think I would also suggest to close this in favour of just merging the change which gives Linux the faked shared data folder, since getting consistent and obvious behaviour across all platforms probably needs a re-think of token names/functions as well as what is the lowest common denominator.

ivan-mogilko:
Alright then.

PR #904, opened 2019-08-30
AGS3.5: Legacy pathfinding

sonneveld:
This introduces the old pathfinding behaviour for older (< v3.5) games. 

I tried to implement in such a way that we could easily remove the legacy pathfinder (so no shared code) but also without changing the header of `route_finder.h` too much.  The implementation modules are essentially unchanged but wrapped with namespaces.

I'm creating the pull request early to check if people are happy with implementation. ~It needs a few more checks to make sure I haven't broken the save game system, and modifications to visual studio files.~


Engine/ac/route_finder.cpp
```diff
@@ -17,47 +17,42 @@
...
 #include "ac/common_defines.h"
-#include <string.h>
-#include <math.h>
 #include "gfx/bitmap.h"
 
 #include "route_finder_jps.inl"
 
+extern MoveList *mls;
+
+#ifdef DEBUG_PATHFINDER
```
ivan-mogilko: Please, use Debug:: functionality to output logs, not Display command.

ivan-mogilko:
Is it possible to have pathfinder code moved to another file as a separate commit before implementing these interfaces? This way (I hope) the code history will be cleaner.

Engine/ac/route_finder.cpp
```diff
@@ -232,12 +227,16 @@ int find_route(short srcx, short srcy, short xx, short yy, Bitmap *onscreen, int
...
+  Display("Route from %d,%d to %d,%d - %d stages", srcx,srcy,xx,yy,num_navpoints);
+#endif
 
   int mlist = movlst;
   mls[mlist].numstage = num_navpoints;
   memcpy(&mls[mlist].pos[0], &navpoints[0], sizeof(int) * num_navpoints);
-//    fprintf(stderr,"stages: %d\n",num_navpoints);
+#ifdef DEBUG_PATHFINDER
+  fprintf(stderr,"stages: %d\n",num_navpoints);
```
ivan-mogilko: Also, since you are uncommenting these, please use same debug output method everywhere (preferrably Debug::Printf).

Engine/ac/route_finder.cpp
```diff
@@ -11,242 +11,91 @@
...
-void init_pathfinder()
+#include "ac/route_finder_impl.h"
+#include "ac/route_finder_impl_legacy.h"
+
+#include "debug/out.h"
+
+
+class IRouteFinder {
+public:
+virtual void init_pathfinder() = 0;
```
ivan-mogilko: I will use an opportunity to note again, that this is not the code style we are tend to use in the project. According to the code convention,
* the curve brackets are on their own line
* there should be indentation for members of class.
* member function names are in CamelCase.

Engine/ac/route_finder_impl.h
```diff
@@ -0,0 +1,41 @@
...
+
+int can_see_from(int x1, int y1, int x2, int y2);
+void get_lastcpos(int &lastcx, int &lastcy);
+
+void set_route_move_speed(int speed_x, int speed_y);
+
+int find_route(short srcx, short srcy, short xx, short yy, AGS::Common::Bitmap *onscreen, int movlst, int nocross = 0, int ignore_walls = 0);
+void calculate_move_stage(MoveList * mlsp, int aaa);
+
+}
```
ivan-mogilko: Another general request, could you add // Namespace at these closing brackets? this helps to understand what is being closed (especially in larger files).

Engine/ac/route_finder_impl.cpp
```diff
@@ -0,0 +1,270 @@
...
+namespace RouteFinder {
+
+#define MAKE_INTCOORD(x,y) (((unsigned short)x << 16) | ((unsigned short)y))
+
+static const int MAXNAVPOINTS = MAXNEEDSTAGES;
+static int navpoints[MAXNAVPOINTS];
+static int num_navpoints;
+static fixed move_speed_x, move_speed_y;
+static Navigation nav;
+static AGS::Common::Bitmap *wallscreen;
```
ivan-mogilko: Please note that you have `using AGS::Common::Bitmap;` above, so you don't have to repeat namespaces here and below. Perhaps same could be done for other ***.cpp** files too.

ivan-mogilko:
Other than that I have mostly style critisisms (or nitpicks, depending on how you see this).
I will do actual tests bit later.

ivan-mogilko:
Oh, I was actually wondering if it may be worth to add a config option instead of switching strictly by game version. Or would that let too many potential errors through? What others will think about it?

BTW I wonder what happens if engine loads a save made when different pathfinder was used. Presumably pathfinder does not save anything on its own (only resulting paths will be saved)...

ericoporto:
If it's an option people can keep using it, if there's value in keeping it then do as option, if it's going to be removed in next version (3.6.0 ?) then by detection will avoid people depending on the resulting paths for games that are being made with 3.5.0. :smiley_cat:

ivan-mogilko:
> If it's an option people can keep using it, if there's value in keeping it then do as option, if it's going to be removed in next version (3.6.0 ?) then by detection will avoid people depending on the resulting paths for games that are being made with 3.5.0. smiley_cat

I should elaborate, I was talking about config option, not project option. Something that you may change at will as a player when you are playing old games. It will also be something that is easy to remove, because it does not depend on game data but on engine capability (it's like with graphics driver, if we don't support one we just drop it and ignore it in config).

Engine/ac/route_finder.cpp
sonneveld: I'll tidy the style. It was an attempt to keep the boilerplate code small and tidy.

I would like to keep the member functions the same as the function names, to make it obvious during a text search that they are the same thing. The interface is hidden within the the cpp file, so it's not exposed to the rest of the project. If we ever do create a proper class to encapsulate the path finding then I would definitely prefer CamelCase for function names.

sonneveld:
> Is it possible to have pathfinder code moved to another file as a separate commit before implementing these interfaces? This way (I hope) the code history will be cleaner.

Good point. I've now done this explicitly.

> Oh, I was actually thinking if it may be worth to add a config option instead of switching strictly by game version. Or would that let too many potential errors through? What others will think about it?

I think if a game was developed and tested with a particular pathfinder, we should only provide that one. 

> BTW I wonder what happens if engine loads a save made when different pathfinder was used. Presumably pathfinder does not save anything on its own (only resulting paths will be saved)...

I haven't noticed any issues so far but might require a little testing.

sonneveld:
I've force pushed the latest changes. It should address the majority of the style issues you raised. If you do feel strongly about the class member names I can change that.

Engine/ac/route_finder.cpp
ivan-mogilko: > It was an attempt to keep the boilerplate code small and tidy.

The code style may be adjusted to this if you think it is beneficial, but having separate styles in chosen places is confusing.

ivan-mogilko:
I have only one extra request: could you add "uninitialize" function to pair with initialize, and call it from some place inside "quit" engine function? Note that it should be called not after allegro_exit in case it uses library resources (bitmaps etc).

sonneveld:
Okay, this should hopefully be ready to be merged.

Issue #624, opened 2019-02-13
Changing character's room don't add it to the item list on properties pane


ivan-mogilko:
* Open one of the rooms (let's call it room X);
* Have Characters layer selected in the room editor;
* Create new character and open its editing pane;
* Set this new character's StartingRoom to X;
* Return back to the room editor.

What you will notice is that new character was added to the navigation bar, but was not added to the item list on the properties pane.

**UPDATE:** opposite is also true: if you change StartingRoom to different room while Character preview pane is active, switching back to room editor will keep old list of characters in the properties pane list.

Known workaround: switch to different layer and back to Characters - that will reinit the items list on properties pane.

ivan-mogilko:
Should be fixed with #906.

PR #906, opened 2019-08-31
Editor: explicitly update room document's property list on character add/removal

ivan-mogilko:
For #624.

There's a lot to say about how editor handles property grid, for a quick recap see sequential comments to this thread: https://github.com/adventuregamestudio/ags/pull/619#issuecomment-464797516

In short the editor's event system was built with the assumption that only 1 working pane will be open at same time and that this pane will have exclusive ownership on objects it edits. As development progressed which ceased being true, especially after AGS 3.2.2 where docking UI was added and since when you may have several working panes at the same time. And this conflicts with how property grid gets updated, which presumes that it needs to collect objects and properties from the *active pane*, while some actions may have effect on other panes at the background.

Anyhow, this PR contains a rather dirty hack where Characters Filter explicitly tells Room Editor to update its own property list, around the default callback that would instead do this in the active pane.

This solves the bug that occurs when you have character filter active in the room editor, then switch to character editor and change StartingRoom, adding or removing that char to/from the edited room. In such case the room's property list does not get updated and upon switching to room editor user will see wrong list of characters in the property grid.

**NOTE:** The alternative to this seem to be forced property list update (recollecting objects, etc) whenever user changes active pane (on focus). May this be a better solution? After all, property grid only reflects the active pane. Don't know how easy that would be to add such update into current system yet.

### DIFF
--- a/Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs
@@ -22,6 +22,7 @@ namespace AGS.Editor
         private Game _game = null;
         private Room _room;
         private Panel _panel;
+        RoomSettingsEditor _editor;
         private Character _selectedCharacter = null;
         private bool _movingCharacterWithMouse = false;
         private int _menuClickX = 0;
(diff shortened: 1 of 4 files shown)


morganwillcock:
> NOTE: The alternative to this seem to be forced property list update (recollecting objects, etc) whenever user changes active pane (on focus). May this be a better solution? After all, property grid only reflects the active pane. Don't know how easy that would be to add such update into current system yet.

I probably don't know enough to say for sure, but I wouldn't have thought that using the focus change as a trigger is 100% reliable where the order of the instructions was important. I think I've have had issues in the past where just running with the Visual Studio debugger attached was enough to get different outcomes.

ivan-mogilko:
I should probably also note that at this very moment Characters are only objects that are shared among two editor panes (room editor and character view). So all similar problems related to them, and if temporary hotfixes like this are okay we could stick to them for now.

Issue #909, opened 2019-09-02
Update available types in Global Variables panel


ivan-mogilko:
The type list on Global Variables pane has not been updated for a while.
Missing new script types should be added there.

Also, I have no idea if that is feasible, but ideally this list should be generated from the built-in script header to match API.

ivan-mogilko:
Done by #910.

PR #910, opened 2019-09-03
In Global Variables pane gather list of types from script header

ivan-mogilko:
For #909.

Removes hard-coded types from the variable dialog (which were not updated in ages anyway) and instead use precreated list of types gathered by parsing script API header with regex.

### DIFF
--- a/Editor/AGS.Editor/GUI/GlobalVariableDialog.cs
+++ b/Editor/AGS.Editor/GUI/GlobalVariableDialog.cs
@@ -17,12 +17,12 @@ namespace AGS.Editor
         private GlobalVariable _variable;
         private Game _game;
 
-        public GlobalVariableDialog(GlobalVariable variable, Game game)
+        public GlobalVariableDialog(GlobalVariable variable, Game game, IEnumerable<APITypeDef> apiTypes)
         {
             InitializeComponent();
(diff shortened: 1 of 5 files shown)


Editor/AGS.Editor/GUI/GlobalVariableDialog.cs
```diff
@@ -38,25 +38,30 @@ public GlobalVariableDialog(GlobalVariable variable, Game game)
...
+            {
+                foreach (var def in apiTypes)
+                {
+                    if (!def.Managed || def.String) continue;
+                    if (def.Autoptr)
+                        cmbType.Items.Add(new GlobalVariableType(def.Name, null));
+                    else
+                        cmbType.Items.Add(new GlobalVariableType(string.Format("{0}*", def.Name), null));
+                }
+                cmbType.SelectedIndex = 0;
```
morganwillcock: Since there will definitely be 4 values added to the control, I would imagine that if `apiTypes` was null for some reason, you would still want to set the SelectedIndex property to 0?

ivan-mogilko: Probably misplaced bracket, fixed now.

PR #912, opened 2019-09-05
Editor: send mouse events only to active room layer

ivan-mogilko:
Existing behavior of the room editor, where mouse events may be handled by multiple layers at once until one of them claims it, came along with the new navigation bar code. IMO it is not very well thought through for two reasons:
* this mixes "visible" and "interactable" property in one (Visible), which is not necessarily wanted, and contradicts to how layers usually work in editing programs (e.g. paint software).
* event is handled in the order layers are stored internally in the room editor, and this order is arbitrary and hard-coded.

I suggest to disable this behavior for the time being and instead send events only to layers explicitly activated by user.

### DIFF
--- a/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
@@ -43,6 +43,7 @@ namespace AGS.Editor
         protected Room _room;
         protected Panel _panel;        
 		protected ToolTip _tooltip;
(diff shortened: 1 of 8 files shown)


morganwillcock:
I think this is probably the best solution, lock interaction per layer and maybe we can try to provide better layer switching (or layer picker tool) at a later date. The only problem I can see is that the cursor state is still switched back to the arrow, so the visual clue that that the picker is still active is lost:

picker

ivan-mogilko:
Um, maybe there's something I missed, I will take a look.

ivan-mogilko:
Remade some commits, and added more fixes.

PR #916, opened 2019-09-10
Editor: disable sending data to ags server

ivan-mogilko:
For #792.

This disables user ability to send crash reports, because it's no longer working and is not particulary useful without proper accessible database (which we don't have anymore).

For now I simply hid the "Send report" button and changed the crash window text, telling user to post on forums instead.


ivan-mogilko:
Ammended, now disabled sending user statistics and hid the related option in the preferences.

PR #921, opened 2019-09-18
Set the tab index on sprite import controls

morganwillcock:
* 0 is the 'Import' button, so if you just want to import with no changes you can press enter (this was a user request)
* 1 is the 'Import all' button, so pressing tab and then enter is a shortcut to import everything where there are multiple files
* Past 1, the controls are treated as vertical columns, and you are already at the bottom of the first column

I looked at setting keyboard handlers to accept enter on other controls, but I worry this would lead to accidental form submission when people press enter after entering a value, which would then lead to implementing custom keyboard handlers for all controls.


ericoporto:
Downloaded from the [build here](https://cirrus-ci.com/task/6493741633765376) and verified it does work.

PR #923, opened 2019-09-22
Add Android CI build and release

morganwillcock:
This adds the missing Android build that I couldn't get to work before.

Container build:
- download NDK version r16b to /opt and build toolchains for Arm, MIPS, and x86
- download SDK tools build to /opt and install required SDK packages using sdkmanager
- merge (no overwrite) SDK tools 25.2.5 over the top to regain Ant support
- download OpenJDK8 (via API query to get the latest LTS release from [AdoptOpenJDK](https://adoptopenjdk.net/)) to /opt
- download Ant to /opt
- build nativelibs and move them to /opt (this is the part that downloads from AWS so this is cached in the container to avoid being repeated, one of the Docker build arguments lets you specify which branch is used for this action)

Note: I still find that merging SDK tools to get Ant support is breaking things like the SDK manager, although it isn't used after this point

Build task:
- update git submodules
- symlink /opt/nativelibs as Android/nativelibs
- run ndk-build for 'library'
- run Ant in debug and release mode for 'launcher_list'
- rename APKs to insert the build version

Release task:
 - upload `AGS-<VERSION>-debug.apk`

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -113,6 +113,32 @@ build_macos_task:
   binaries_artifacts:
     path: build_*/ags
 
+build_android_task:
+  container:
+    dockerfile: ci/android/Dockerfile
+    docker_arguments:
+      FROM_DEBIAN: debian:buster
+      AGS_BRANCH: release-3.5.0
+  git_submodules_script: git submodule update --init --recursive
+  symlink_nativelibs_script: ln -s /opt/nativelibs Android/nativelibs
+  build_library_script: |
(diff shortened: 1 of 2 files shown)


ericoporto:
Your cirrusci.yml file has the previous script to get the manual in it, the one on AGS repo works by using that shortuct getting latest manual released, so it works.

morganwillcock:
It should be OK once merged, I can't rebase the branch without losing my authentication tokens.

Issue #560, opened 2018-12-30
add the option to start animation at any frame of the loop


selmiak:
at the moment the Animate command resets the animation. SetView lets you set a frame from a loop but then starting the animation with that loop leads to stuttering when the loop jumps to frame 0.

selmiak:
expanding on that, I mean adding a "frame" property to Animate, but then when a specific frame that is not 0 is set within SetView then Animate without that property set should start from that frame set in setview.

ivan-mogilko:
I guess this means adding 1 argument to Object.Animate and Character.Animate functions, e.g.:
Animate (loop, frame, ...other params...).

Besides that, maybe investigate if it's worth to bring these functions to the similar argument list with Button.Animate, which lets you set view as well. (Not sure about this tbh, since characters have special rules about Views).

selmiak:
okay, thinking about it, it would be better to set the starting frame with SetView and maybe a final argument for Animate that is eReset or eContinue to say start from current from of the object or reset to frame 0 of the loop.

ivan-mogilko:
> after some thinking about it, I think it would be better to set the starting frame with SetView and maybe a final argument for Animate would be needed that is eReset or eContinue to say start from the current frame of the object or reset to frame 0 of the loop.

My thinking is that "frame" argument in Animate() would be more universal:
* if you want to reset you pass 0,
* if you want to continue from current one you pass Character.Frame,
* if you want to begin from any random one, you pass that number.

ivan-mogilko:
> Hard to say what is best. That's why I suggested adding this with the setview. But this tears it apart some more, also not good.

The thing about SetView's loop and frame parameters is that they were not really meant as a starting point for Animate. It is rather a coincidence that SetView function is used for 2 separate purposes: to set a view before animation and to set certain static frame as a new object's sprite.
If you compare with character's functions, there you have to call LockView which only set "view" but not "loop" or "frame" (these may be locked too but with different functions like LockViewFrame).

If we make Animate follow SetView's frame, then more questions arise: what's the point of having "loop" parameter in Animate (we could also use one from SetView), and more importantly why we have to call 2 functions every time to start animation from particular frame instead of just 1?

The way to place a new parameter in Animate is non-trivial problem of course; currently I don't see other option than to create a new function where frame is not optional. The previous function is either left (then the new one needs to be called differently, like AnimateEx etc), or called deprecated and only available in backwards compatibility mode.

ivan-mogilko:
[...]

PR #929, opened 2019-09-26
Script API: added "frame" parameter to Character.Animate and Object.Animate

ivan-mogilko:
For #560

Added optional "frame" parameter to Character.Animate and Object.Animate functions that tells which frame to begin animating with.

Trying to comply with existing behavior, reverse animations begin with the *previous frame*, so e.g. if you pass frame 0 they will begin with the last frame instead, if you pass frame 1 they will begin with 0th and so on.

Note: I was also thinking about adding same for Button.Animate, but I noticed that its parameter list does not have several other options, such as Direction and BlockingStyle, so maybe later there should be a separate task of adding them all at once. Besides, buttons are much easier to animate in script on your own.

### DIFF
--- a/Engine/ac/global_object.cpp
+++ b/Engine/ac/global_object.cpp
@@ -204,7 +204,7 @@ int GetObjectBaseline(int obn) {
     return objs[obn].baseline;
 }
 
-void AnimateObjectEx(int obn,int loopn,int spdd,int rept, int direction, int blocking) {
+void AnimateObjectImpl(int obn, int loopn, int spdd, int rept, int direction, int blocking, int sframe) {
     if (obn>=MANOBJNUM) {
         scAnimateCharacter(obn - 100,loopn,spdd,rept);
         return;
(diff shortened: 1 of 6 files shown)


Engine/ac/global_object.cpp
```diff
@@ -213,24 +213,27 @@ void AnimateObjectEx(int obn,int loopn,int spdd,int rept, int direction, int blo
...
 
     objs[obn].cycling = rept+1 + (direction * 10);
     objs[obn].loop=loopn;
-    if (direction == 0)
-        objs[obn].frame = 0;
-    else {
-        objs[obn].frame = views[objs[obn].view].loops[loopn].numFrames - 1;
+    // reverse animation starts at the *previous frame*
+    if (direction) {
+        if (--sframe < 0)
```
sonneveld: I recommend putting the decrement on another line to reduce confusion here.

ivan-mogilko: Hmm ok, I wanted to reduce number of lines :)

Issue #600, opened 2019-02-01
Get Android port ready for 64-bits requirements


ericoporto:
Google sent me an e-mail that contains the text here:

https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html

I think it's just adding the **`arm64-v8a`** and **`x86_64`** targets will work but I haven't tested. Note armeabi and mips are dropped for recent ndk.

https://developer.android.com/ndk/guides/abis

Testing here, on `makestandalones.sh` . Platform has to be updated to 21 (from 14) and arch x86_64 and arm64 is added

```
...
PLATFORM=android-21

# standalone toolchains cannot share same directory
export NDK_STANDALONE=$NDK_HOME/platforms/$PLATFORM

# if you don't pass force, the directory is not rewritten
# verbose is needed to see why making the toolchain fails
for arch in arm x86 x86_64 arm64 mips
...

```

Then copied `x86` folder and renamed to `x86_64`, fixed ndkenv, and copied `armeabi-v7a` and renamed to `arm64-v8a`, and also updated it's ndkenv.

x86_64 ndkenv changes
```
export NDK_PLATFORM_ROOT=$NDK_STANDALONE/x86_64
export NDK_PLATFORM_NAME=x86_64
export NDK_HOST_NAME=x86_64-linux-android
```

arm64 ndkenv changes
```
export NDK_PLATFORM_ROOT=$NDK_STANDALONE/arm64
export NDK_PLATFORM_NAME=arm64-v8a
export NDK_HOST_NAME=aarch64-linux-android
```

and both
```
export CC=$NDK_PLATFORM_ROOT/bin/${NDK_HOST_NAME}-gcc
export CXX=$NDK_PLATFORM_ROOT/bin/${NDK_HOST_NAME}-g++ 
```

Then on buildall.sh did:

```
# android-21 is the minimum we can go with current Android SDK and 64-bits
PLATFORM=android-21

# standalone toolchains cannot share same directory
export NDK_STANDALONE=$NDK_HOME/platforms/$PLATFORM

for arch in armeabi armeabi-v7a x86 x86_64 arm64-v8a mips 
```

Lua will fail for platform-21 because bsd_signal is removed. 

`lua.c:function laction: error: undefined reference to 'bsd_signal'`

[There is a fix here](https://stackoverflow.com/questions/36746904/android-linker-undefined-reference-to-bsd-signal)

[...]

PR #932, opened 2019-09-28
changes for android arm64 build

ericoporto:
Please do not merge without testing this.

As stated on #600 , Google Play doesn't accept app builds that don't contain 64-bit Arm binaries. This PR adds Arm64 builds.

### DIFF
--- a/Android/buildlibs/arm64-v8a/allegro.sh
+++ b/Android/buildlibs/arm64-v8a/allegro.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -e 
+
+source ./ndkenv
+
+SRC_DIR=allegro-4.4.2
+rm -rf $SRC_DIR
+mkdir $SRC_DIR
+tar xf ../../../libsrc/allegro-4.4.2.tar.gz --strip-components=1 -C $SRC_DIR
+
+pushd $SRC_DIR
+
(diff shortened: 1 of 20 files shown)


ivan-mogilko:
So, this requires changing platform to android-21?

ericoporto:
[android-libs.zip](https://github.com/adventuregamestudio/ags/files/3666056/android-libs.zip)

From: https://developer.android.com/ndk/guides/standalone_toolchain

> The --arch argument is required, but the API level will default to the minimum supported level for the given architecture (currently 16 for 32-bit architectures and 21 for 64-bit architectures).

I got the info from above. I also added the builds from my computer above.

The previous error was because the CI environment uses a Dockerfile with

```bash
# Make NDK toolchains
RUN export NDK_HOME=/opt/android-ndk-${NDK_VERSION} && \
  for arch in arm mips x86; do \
    echo Building toolchain for $arch; \
    $NDK_HOME/build/tools/make_standalone_toolchain.py --arch $arch --api 14 --install-dir $NDK_HOME/platforms/android-14/$arch; \
  done
```

Instead of the [`makestandalones.sh`](https://github.com/adventuregamestudio/ags/blob/release-3.5.0/Android/buildlibs/makestandalones.sh). 

I now updated it.

Issue #237, opened 2015-03-13
Function in Engine/ac contains unreachable code


MoreMoschops:
The function  "int IsKeyPressed (int keycode)" in file "engine/ac/global_game.cpp" contains unreachable code at lines 623, 625, 627 and 629.

Clearly at some point someone wanted those lines to be reachable and that code to execute, but instead each of them has a forerunner that does something slightly different. I don't know enough about those lines to have the confidence to either delete them or change them, although I'm happy to poke at it and see what I can deduce if nobody stops me...

MoreMoschops:
Having had a poke round, though, I can't actually see anywhere that the function is called from, unless it's done magically through function pointers or that huge RegisterGlobalAPI() function or some such. Sure are a lot of globals hanging around!

monkey0506:
This function is only called by game scripts, it is never called directly by the engine. Still, best not to have unreachable code.

monkey0506:
It seems like the entire IsKeyPressed function really seems rather naive in its implementation. It reuses the keycode argument to ultimately be forwarded to the rec_iskeypressed function, but yet it is intermingled with other calls to that same function. This makes the function far less readable than if it simply called rec_iskeypressed with the appropriate parameters as needed.

In particular, if keycode is 380, then what would make sense would be to check rec_iskeypressed(85) (actual down arrow key) or that Num Lock is not activated and rec_iskeypressed(39) (Numpad Key 2). The way the function is right now I would expect it to yield incorrect results for Numpad Key 2 as a down arrow -- unless Allegro 4 makes an exception for this. From what I can see though, it really looks like Allegro 4's keyboard handling was kind of a mess in the first place...

My recommendation would really be to rewrite the entire method.

MoreMoschops:
I freely confess I just happened to notice the unreachable code in passing; exactly what this function is up to is a bit of a mystery to me. Looks like it really should be kicking back a bool rather than an int (given that the possible return values are 1 or 0, and the function name is a question), but for such a function it seems to have a lot of side effects going on; is it not just checking is a key is pressed, but also updating global state in the event that a key is being pressed? 

Plus, of course, it's heaving with magic numbers! I did stick a breakpoint in here, run up a game and see what I had to do to provoke this function call, but nothing doing. Presumably the test game I used simply doesn't call this? 

I don't know nearly enough about this; thanks for your patience. I just saw something odd and started digging.

monkey0506:
[...]

PR #934, opened 2019-09-29
AGS3.5 Key mapping refactor and some macos key bugs

sonneveld:
`IsKeyPressed` and `ags_getch` used some complicated arithmetic to map between allegro4 keycodes and ags keycodes (which are mostly based on the older allegro3 keycodes). This led to potentially unreached code, as noted here: https://github.com/adventuregamestudio/ags/issues/237

I ran the functions through some loops, to generate all values, and created these more straightforward switch statements. This is essentially how the SDL2 port maps keys as well.

I also found some issues with macos key maps which should be fixed now. (essentially any alt-key combination did not work, nor did '?')

Note: the switch statements are "squashed" to one line per mapping to make it clearer but this does go against the style guide.

### DIFF
--- a/Engine/ac/global_game.cpp
+++ b/Engine/ac/global_game.cpp
@@ -604,134 +604,132 @@ int IsKeyPressed (int keycode) {
 #ifdef ALLEGRO_KEYBOARD_HANDLER
     if (keyboard_needs_poll())
         poll_keyboard();
-    if (keycode >= AGS_EXT_KEY_SHIFT) {
-        // function keys are 12 lower in allegro 4
-        if ((keycode>=359) & (keycode<=368)) keycode-=12;
-        // F11-F12
-        else if ((keycode==433) || (keycode==434)) keycode-=76;
-        // left arrow
-        else if (keycode==375) keycode=382;
-        // right arrow
-        else if (keycode==377) keycode=383;
-        // up arrow
-        else if (keycode==372) keycode=384;
-        // down arrow
-        else if (keycode==380) keycode=385;
-        // numeric keypad
-        else if (keycode==379) keycode=338;
-        else if (keycode==380) keycode=339;
-        else if (keycode==381) keycode=340;
-        else if (keycode==375) keycode=341;
-        else if (keycode==376) keycode=342;
-        else if (keycode==377) keycode=343;
-        else if (keycode==371) keycode=344;
-        else if (keycode==372) keycode=345;
-        else if (keycode==373) keycode=346;
-        // insert
-        else if (keycode == AGS_KEYCODE_INSERT) keycode = KEY_INSERT + AGS_EXT_KEY_SHIFT;
-        // delete
-        else if (keycode == AGS_KEYCODE_DELETE) keycode = KEY_DEL + AGS_EXT_KEY_SHIFT;
 
-        // deal with shift/ctrl/alt
-        if (keycode == 403) keycode = KEY_LSHIFT;
-        else if (keycode == 404) keycode = KEY_RSHIFT;
-        else if (keycode == 405) keycode = KEY_LCONTROL;
-        else if (keycode == 406) keycode = KEY_RCONTROL;
-        else if (keycode == 407) keycode = KEY_ALT;
-        else keycode -= AGS_EXT_KEY_SHIFT;
+    switch(keycode) {
+        case eAGSKeyCodeBackspace: return ags_iskeypressed(__allegro_KEY_BACKSPACE); break;
+        case eAGSKeyCodeTab: return ags_iskeypressed(__allegro_KEY_TAB); break;
+        case eAGSKeyCodeReturn: return ags_iskeypressed(__allegro_KEY_ENTER) || ags_iskeypressed(__allegro_KEY_ENTER_PAD); break;
+        case eAGSKeyCodeEscape: return ags_iskeypressed(__allegro_KEY_ESC); break;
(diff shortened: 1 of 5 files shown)


messengerbag:
> I also found some issues with macos key maps which should be fixed now. (essentially any alt-key combination did not work, nor did '?')

That's great! A possible test case could be the 2018 AGS Awards client, which tries to map the full set of keycodes to their normal text entry functionality. The Mac port was missing a bunch of these (as well as copy/paste, which is a different issue).

Perhaps a related problem: Several users have [reported problems](https://www.adventuregamestudio.co.uk/forums/index.php?topic=56805.msg636604350#msg636604350) in the Awards client with certain keys not working, including Right Shift for Stupot and Right Shift and apostrophe for CassieBsg (who was using a Danish keyboard). Testing has shown that this is because the keys are not generating on_key_press() events or returning true for IsKeyPressed(). I believe at least in some cases they work correctly for regular TextBox input, so on some level AGS is aware they are being pressed. It would be nice if these things could be fixed at some point.

There are also some issues with the whole keycode system in general (particularly how you have to rely on IsKeyPressed() to figure out control-key combinations, which is unreliable because you don't get the state when the button was actually pressed), but that's probably a matter for another time.

sonneveld:
>  A possible test case could be the 2018 AGS Awards client, which tries to map the full set of keycodes to their normal text entry functionality. 

This might not fix this. While it does fix a couple of mac issues, it mostly has the same behaviour as before. The danish keyboard and shift issues might be worth investigating though.

The ags keycodes don't include the full set of ascii characters, it's mostly mapped to scan codes. The exception to this are text fields, which do get proper ascii character events. I wonder if it's worth having an almost invisible or off to the side text box which you can use to capture proper ascii characters, using IsKeyPressed for the other key combinations.

ivan-mogilko:
@sonneveld , you very first refactor breaks IsKeyPressed, because now special keys like Alt are not detected. They were the reason original function had `keycode -= AGS_EXT_KEY_SHIFT;` operation.

In general I think this pr is too serious for 3.5.0, which is practically at release candidate, because it does not only tidies the code but changes some logic. That would require more research to find out why some of the old code existed.

I would accept only actual fixes (like ones you did for mac keys, if you think they are necessary).

ivan-mogilko:
@sonneveld , your version of IsKeyPressed does not detect special keys like Alt now.
This is how they were dealt in original code:

<pre>
        // deal with shift/ctrl/alt
        if (keycode == 403) keycode = KEY_LSHIFT;
        else if (keycode == 404) keycode = KEY_RSHIFT;
        else if (keycode == 405) keycode = KEY_LCONTROL;
        else if (keycode == 406) keycode = KEY_RCONTROL;
        else if (keycode == 407) keycode = KEY_ALT;
        else keycode -= AGS_EXT_KEY_SHIFT;
</pre>

Also note that `keycode -= AGS_EXT_KEY_SHIFT`, is it known which key codes were affected by this?

I am very concerned that this pr could break some logic or miss more keys, and is too serious for 3.5.0, which is practically at release candidate.

I'd rather accept only actual fixes (like ones you did for mac keys, if you think they are necessary).

The rest could be left for next development version.

ivan-mogilko:
Regarding 
<pre>
if (keycode == 0)
     keycode = ags_getch() + AGS_EXT_KEY_SHIFT;
</pre>

I have vague memories of testing this long ago, but trying various key combinations right now does not hit this condition at all.

ivan-mogilko:
Regarding 
<pre>
int keycode = ags_getch();
if (keycode == 0)
     keycode = ags_getch() + AGS_EXT_KEY_SHIFT;
</pre>

It appears that ags_getch() will not normally return 0 (at least not on Windows and Linux), because it has this code: https://github.com/adventuregamestudio/ags/blob/release-3.5.0/Engine/ac/sys_events.cpp#L121

<pre>
if ((gott & 0x00ff) == EXTENDED_KEY_CODE) {
        gott = scancode + AGS_EXT_KEY_SHIFT;
</pre>

and EXTENDED_KEY_CODE is 0 except for MacOs and iOS.

sonneveld:
> your version of IsKeyPressed does not detect special keys like Alt now.
> This is how they were dealt in original code:

Ah, thankyou. I missed that because they don't have enums defined, only raw numbers described in the manual. I have now included them, and any other allegro keys that were solely available in that range.

> Also note that keycode -= AGS_EXT_KEY_SHIFT, is it known which key codes were affected by this?

I've attached the full mapping of keys if you'd like to see: [iskeypressed_output.txt](https://github.com/adventuregamestudio/ags/files/3668669/iskeypressed_output.txt)

> Another thing, why commit ada5d15 does not replace the code but puts old one under #if 0 ? Does this mean that you are not sure it's the correct change?

I am very sure it is correct. I literally looped through all possibilities. The only reason I missed the ALT and SHIFT keys earlier is because there wasn't an enum defined. I left it for documentation but you're right, I've removed it now.

> I am very concerned that this pr could break some logic or miss more keys, and is not well suited for 3.5.0, which is practically at release candidate. Are there any serious reasons to do such change now?

No serious reason for the whole thing to go in right now. I mentioned in chat that I was completely happy for this PR to sit until after the 3.5 release.  It was only during the refactor that I noticed the issues with macos keys.

> It appears that ags_getch() will not normally return 0 (at least not on Windows and Linux),

I think it would only happen on macos if an alt/option key combination was used, because that set ascii value to '\0' instead of '\?' which it normally did for scancode only keys. So before this PR, if you press option-anything, the game will freeze until you press another key.

PR #941, opened 2019-10-06
Editor: remember last selected item in the room

ivan-mogilko:
By user request, this complements room editor saving the "design state" of a room by adding last selected item.

This is a trivial change from data saving perspective, but required additional modifications in the room editor to allow selecting item in "background mode" , that is - while the room pane is not active, because this will be the case when room just getting loaded.

For this I had to write function ~(per each room layer, unfortunately, because of how things are currently abstracted, or rather not abstracted enough)~ that checks whether room editor is an active pane or not. If it is, it then proceeds by a usual route and calls Factory.GUIController.SetPropertyGrid* method which will both update current property list and display one in the grid. If room editor is at the background then it will just assign new property list to the document; that list will be displayed in the grid as soon as room editor receives focus.

Note that I also removed a previos fix made by #619, because new function seem to counter same bug in its own way.

### DIFF
--- a/Editor/AGS.Editor/Components/RoomDesignData.cs
+++ b/Editor/AGS.Editor/Components/RoomDesignData.cs
@@ -29,13 +29,15 @@ namespace AGS.Editor.Components
     {
         private const string XML_DATA_ROOT_NODE = "AGSRoomUserData";
         private const string XML_ROOM_NODE = "Room";
+        private const string XML_ROOM_SELECTION = "SelectedItem";
+        private const string XML_SELITEM_ATTRIB_PATH = "BreadcrumbPath";
         private const string XML_ROOM_LAYERS_NODE = "Layers";
         private const string XML_LAYER_ELEM_VISIBLE = "Visible";
         private const string XML_LAYER_ELEM_LOCKED = "Locked";
         private const string XML_LAYER_ITEMS_NODE = "Items";
         private const string XML_LAYER_ITEM_NODE = "Item";
         private const string XML_ITEM_ATTRIB_ID = "ID";
-        private const int CURRENT_DATA_VERSION_INDEX = 1;
+        private const int CURRENT_DATA_VERSION_INDEX = 2;
 
         /// <summary>
         /// Load a previously saved design-time user data for the current room.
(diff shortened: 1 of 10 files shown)


ivan-mogilko:
Hmm, actually I just got an idea how to reduce amount of code by picking it out, will push ammended commits shortly.

morganwillcock:
Is there any other identifier that could be used for the identification of objects, other than the node paths for the breadcrumbs/URL controls? I'm assuming not, but workflow for this control is not that great so I'm not sure it would always exist (thinking longer term).

ivan-mogilko:
> Is there any other identifier that could be used for the identification of objects, other than the node paths for the breadcrumbs/URL controls

This was the problem I first met when was implementing saving the design state of a room: in AGS objects do not have *obligatory* unique identifier. Script names are common, but not required for game to compile, because items may be accessed through arrays, and, for example, generally you do not have to access room objects in scripts at all.

So the only way to identify them currently is their numeric IDs in project, which is how they are referenced now, both in this file and in the internal data of the breadcrumb control.

In other words these IDs are not the part of breadcrumb control specification, but vice-versa the control reuses them as they seem to be only project data which is more or less reliable.

> I'm assuming not, but workflow for this control is not that great so I'm not sure it would always exist (thinking longer term).

This file (*.crm.user) is not required for the project to work, and I think format may be adjusted for the new editing methods if necessary. Worst that can happen is that user will loose this temporary design-time data.

ivan-mogilko:
BTW, as an option, I could rename this new XML element to indicate that it belongs to particular control (breadcrumb navigation), so that it could be ignored by editor with different UI.

morganwillcock:
> BTW, as an option, I could rename this new XML element to indicate that it belongs to particular control (breadcrumb navigation), so that it could be ignored by editor with different UI.

That might be useful, or maybe add a property for 'SelectionPathSource' or something similar. I guess it depends if the value is always a string or could be more generic. Using a path-type string seems a logical choice though.

> (although data form like "Room/Objects/xxx" may be pretty universal imho)

You are probably right that there is nothing better to use right now, until there is better resource management. The root node is always the same though, so it is equal in information to "Object/xxx", but I have similar complaints with the control itself so may as well map it 1-1 until something changes.

ivan-mogilko:
Well, I renamed the inner attribute to "BreadcrumbPath".

Issue #940, opened 2019-10-06
Script API: consider change meaning of undefined Object.SetView's parameters


ivan-mogilko:
Apparently, **Object.SetView** function has a pretty counter-intuitive behavior if you skip optional arguments:

From the manual:
> Object.SetView(int view, optional int loop, optional int frame)

> Sets the object's view to VIEW, and changes the object's graphic to FRAME of LOOP in VIEW. If you do not supply the loop or frame, **they will be left unchanged**.

TBH I think this was a very bad decision to make this a default behavior. Maybe it was made by mistake and then left for compatibility reasons?
Something tells me user would rather expect loop and frame reset to 0 if they are not specified (and looking at existing scripts that is the case).

This may lead to weird issues, for example if the last frame in the previous view matches a frame with linked sound in a new view. Then a sound may be played unexpectedly, and users will have hard time realizing what is going on. (This happened just today, and made me write this ticket)

My suggestion is following:
* By default (if optional arguments are not specified) set loop and frame to 0.
* Old behavior may be kept for backwards compatibility, under a condition checking game format index, or maybe script API (OPT_BASESCRIPTAPI) (?).

User may still ask to keep previous loop & frame by passing Object.Loop and Object.Frame respectively, e.g.:
`oObject.SetView(10, oObject.Loop, oObject.Frame);`

**UPDATE**: 
I'd like to elaborate more on my opinion about this.

Characters are quite different from objects, because they are complex systems which have a concept of "direction"/"facing", which in turn requires having animation frames in particular loops; but object does not have that explicitly. IMHO Room Object is supposed to be a very simple struct that does not have any special built-in behavior, but allows to program itself in script.
This is why I tend to disagree with the argument about consistency between characters and objects in this regard.

My reasoning behind the change is that it works better for "default" case when you do not expect any special behavior. At the same time function still may be used to sustain or update running animation in any way you prefer (passing corresponding values).

messengerbag:
It's probably done this way for consistency with `Character`, right? Where you don't want changing the view to affect the loop (direction) of the character.

ivan-mogilko:
But Object does not support loops for direction, unlike character loops are not changed when you move object around.

ivan-mogilko:
Hmm, I made a small research regarding original (legacy) functions. Earlier there were two functions later merged into Object.SetView:
* SetObjectView
* SetObjectFrame

Quoting their manual entries from AGS 2.55:

> SetObjectView
[...]

PR #942, opened 2019-10-09
Script API: changed Object.SetView default loop & frame values to 0

ivan-mogilko:
See #940
This may be somewhat controversial, so I'd like to hear more opinions on this.

The original Object.SetView function had a somewhat unexpected behavior: if no explicit values were provided for loop and frame then it kept the last loop and frame object had from the previously set view.
Although documented, but such behavior seem to cause confusion among users who expect it to reset loop and/or frame instead.

This pr changes default values to 0, which resets loop and frame if they are not provided.
Original behavior is kept for games made with previous script API version.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2143,8 +2143,14 @@ builtin managed struct Object {
   import function RunInteraction(CursorMode);
   /// Instantly moves the object to have its bottom-left at the new co-ordinates.
   import function SetPosition(int x, int y);
+#ifdef SCRIPT_API_v351
+  /// Sets the object to use the specified view, ahead of doing an animation.
+  import function SetView(int view, int loop=0, int frame=0);
+#endif
+#ifndef SCRIPT_API_v351
   /// Sets the object to use the specified view, ahead of doing an animation.
   import function SetView(int view, int loop=-1, int frame=-1);
+#endif
   /// Stops any currently running animation on the object.
   import function StopAnimating();
   /// Stops any currently running move on the object.
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -162,7 +162,8 @@ enum ScriptAPIVersion
     kScriptAPI_v341 = 5,
     kScriptAPI_v350 = 6,
     kScriptAPI_v3507= 7,
-    kScriptAPI_Current = kScriptAPI_v3507
+    kScriptAPI_v351 = 8,
+    kScriptAPI_Current = kScriptAPI_v351
 };
 
 // Determines whether the graphics renderer should scale sprites at the final
--- a/Engine/ac/global_object.cpp
+++ b/Engine/ac/global_object.cpp
@@ -151,20 +151,9 @@ void SetObjectView(int obn,int vii) {
 void SetObjectFrame(int obn,int viw,int lop,int fra) {
     if (!is_valid_object(obn)) quit("!SetObjectFrame: invalid object number specified");
     viw--;
-    if (viw>=game.numviews) quit("!SetObjectFrame: invalid view number used");
-    if (views[viw].numLoops == 0) quit("!SetObjectFrame: specified view has no loops");
-    if (lop>=views[viw].numLoops) quit("!SetObjectFrame: invalid loop number used");
-    objs[obn].view=viw;
-    if (fra >= 0)
-        objs[obn].frame=fra;
-    if (lop >= 0)
-        objs[obn].loop=lop;
-
-    if (objs[obn].loop >= views[viw].numLoops)
-        objs[obn].loop = 0;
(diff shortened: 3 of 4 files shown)


Editor/AGS.Editor/Resources/agsdefns.sh
```diff
@@ -2143,8 +2143,14 @@ builtin managed struct Object {
   import function RunInteraction(CursorMode);
   /// Instantly moves the object to have its bottom-left at the new co-ordinates.
   import function SetPosition(int x, int y);
+#ifdef SCRIPT_API_v3507
+  /// Sets the object to use the specified view, ahead of doing an animation.
+  import function SetView(int view, int loop=0, int frame=0);
```
sonneveld: Wouldn't changing the defaults here be enough for newer code while keeping backward compatibility with older games?

sonneveld:
I guess I don't know enough about the use cases, but I guess the previous behaviour was useful for configuring objects that were already animating, you had the option of changing the view, or the view and the loop.  And there's probably some value in keeping it consistent with the api of changing character views.

I know you've mentioned in the other ticket that you can just be explicit with Object.loop and Object.frame, but could you use that same argument for some of the character api? (assuming changing character views doesn't change frame/loop)

Editor/AGS.Editor/Resources/agsdefns.sh
ivan-mogilko: Strictly speaking it won't be enough if you import existing scripts that rely on old behavior.

ivan-mogilko:
> And there's probably some value in keeping it consistent with the api of changing character views.

Characters are quite different from objects, because they are complex systems which have a concept of "direction"/"facing", which in turn requires having animation frames in particular loops; but object does not have that. IMHO Room Object is supposed to be a very simple struct that does not have any special built-in behavior, but allows to program itself in script.

That's true that changing character's view keeps the loop if possible; though I should note that it also has a special processing for the loop selection which skips loops with missing frames (which objects do not do). But it never kept old frame, frame was always reset to 0.

> I know you've mentioned in the other ticket that you can just be explicit with Object.loop and Object.frame, but could you use that same argument for some of the character api? (assuming changing character views doesn't change frame/loop)

I frankly do not see how same argument can be used for character's api, because character's SetView does not have extra loop and frame arguments, exactly because it has more complex mechanics and their rules. Closest it has function LockViewFrame, but there loop and frame are not optional and are required to be always set.
This is also why I can't agree that there's any consistency between characters and objects in this regard.

PS. On a side note, I was not going to make this change at first (at least for 3.5.0), but then I saw Dave Gilbert on Discord channel also saying about Object.SetView _"I always thought that behavior was odd as well. I just accepted it as being "by design" for... reasons. Never thought to question it."_ ; and that hinted me that old users also fail to use this function as intended.

ivan-mogilko:
As an additional note, which may be important, current animation is stopped for all view changing functions - Character.ChangeView, Character.LockView* and Object.SetView.

Character.ChangeView is not corresponding to Object.SetView much, because it sets "normal view", and character animates on its own when walking. So we'd have to compare Character.LockView and Object.SetView, which are followed by a call to Animate.

But Animate requires you to pass animating loop anyway, which sort of defeats the purpose of keeping previous loop when SetView is called without loop parameter.

ivan-mogilko:
Actually, I myself don't feel very good putting this into 3.5.0, so will remake this pr against "ags3" branch instead, for later consideration.

ivan-mogilko:
Alright, I switched this PR to target "ags3" branch instead of "release-3.5.0". Something got into me after a user had game bugs with this function, and I thought to do this "quick hack" in 3.5.0.

I will leave this for future consideration. I stated my own opinion and reasoning behind this change in comments above, and don't think I have anything else to add.

Having a survey/discussion about this with end-users on AGS forums may also be a good idea.

ericoporto:
Before, if the specified view and loop had no frame, it would get a null pointer error, but if it had at least one frame, and you passed an invalid frame, it would switch to zero, and now, it will error on an invalid frame - which is fair, but this is the part that could break backwards compatibility.

Other than this I also thought that object setview would default to frame and loop 0!

ivan-mogilko:
> Before, if the specified view and loop had no frame, it would get a null pointer error, but if it had at least one frame, and you passed an invalid frame, it would switch to zero, and now, it will error on an invalid frame - which is fair, but this is the part that could break backwards compatibility.

That is true, but I don't know if this matters. Above fixup could've been made to accomodate the behavior I am changing here - that previous loop and frame values are carried over and applied with the the new view, which may have less loops and less frames. It's unclear if that was meant to fix incorrect explicitly passed values.

In fact, if you get very old AGS and try using SetObjectFrame, you will notice that not only frame but also loop and view may be passed incorrectly, in which case engine does not report any errors but does not change view at all.

So, I'd rather leave this unless it's found there was actual game that depended on such fixup.

ivan-mogilko:
Made a quick survey among older AGS users, and they all support this change, so I'll give it a go.

PR #951, opened 2019-10-15
release-3.5.0: libsrc/libcda: strip non-zlib code, and unused platforms.

sonneveld:


ivan-mogilko:
I have only one question, are we allowed to change COPYING file withing the library, and whether this is necessary to do?
Inside it states that although part of the source is under the different license, one can omit these files.

ivan-mogilko:
@sonneveld I _actually_ do think we should not be changing library's license file.

sonneveld:
> I have only one question, are we allowed to change COPYING file within the library, and whether this is necessary to do?

I assumed this was okay as I left the original license text intact. I removed the text for the removed files to avoid confusion.

Issue #971, opened 2019-10-23
D3D driver should clear whole screen before rendering 


AlanDrake:
At the moment the black borders aren't redrawn so if there's a steam overlay or an external shader it will leave a smear over the undrawn regions.
Replicable at fullscreen using the pixelperfect rendering, max round multiplier/keep aspect ratio and reshade with a crt filter that leaks outside the game area.

ivan-mogilko:
Fixed by #974.

PR #974, opened 2019-10-24
Engine: clear D3D9 borders

AlanDrake:
ColorFill is used on pBackBuffer to redraw the black background and ensure that steam overlays and shader wrappers won't cause hall of mirror effects.
Related to issue #971.

### DIFF
--- a/Engine/platform/windows/gfx/ali3dd3d.cpp
+++ b/Engine/platform/windows/gfx/ali3dd3d.cpp
@@ -734,7 +734,7 @@ int D3DGraphicsDriver::_initDLLCallback(const DisplayMode &mode)
 
 void D3DGraphicsDriver::InitializeD3DState()
 {
-  direct3ddevice->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_RGBA(30, 0, 0, 255), 0.5f, 0);
+  direct3ddevice->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_RGBA(0, 0, 0, 255), 0.5f, 0);
 
   // set the render flags.
   direct3ddevice->SetRenderState(D3DRS_CULLMODE, D3DCULL_NONE);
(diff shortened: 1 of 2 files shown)


Engine/platform/windows/gfx/ali3dd3d.cpp
```diff
@@ -734,7 +734,7 @@ int D3DGraphicsDriver::_initDLLCallback(const DisplayMode &mode)
 
 void D3DGraphicsDriver::InitializeD3DState()
 {
-  direct3ddevice->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_RGBA(30, 0, 0, 255), 0.5f, 0);
```
sonneveld: Should we set the viewport here as well? I had a thought that maybe in debug mode we should set this to a non-black colour to make it obvious of rendering issues.

sonneveld:
I tested on my Windows machine with the Steam overlay, seems to correct the problem nicely. 

I am a bit perplexed that direct3d9 lets you clear the viewport outside of the single render target's backbuffer, where is it writing to?

Engine/platform/windows/gfx/ali3dd3d.cpp
AlanDrake: I thought the "30" was unintentional. Want me to put it back, or change the color?

sonneveld: It's probably fine the way you have it. I'm not actually sure this line has any effect! 

As an experiment, I changed the clear colour to bright red, initialised the screen viewport earlier, disabled the Clear() call in `_render` and moved the Clear to later in the initialisation. I expected to see a left over red borders but it was just default black (with a flash of red where we render the game).

I'm not sure what I'm missing here (not a direct3d expert), but I could not find a way to clear the screen to a colour during Initialisation before we do Render calls. 

Although I do see some funky stuff if I toggle between full screen and windowed multiple times. (with the changes for the experiment)

ivan-mogilko:
@I am a bit perplexed that direct3d9 lets you clear the viewport outside of the single render target's backbuffer, where is it writing to?

What do you mean? There are two targets currently supported, one refers to the buffer of a size of display resolution, and another is a temp surface of a game's resolution (for pixel perfect mode).

sonneveld:
> There are two targets currently supported, one refers to the buffer of a size of display resolution, and another is a temp surface of a game's resolution (for pixel perfect mode).

Ah, I was looking at CreateRenderTarget in CreateVirtualScreen and thinking that was the only render target.

AlanDrake:
Pushed the version with just one viewport class member and ColorFill.
If the need arises to change border colors we'll just have to work on the ColorFill.

sonneveld:
These new changes look good. Can someone else test by toggling steam overlay?

ivan-mogilko:
Yes, steam overlay is fixed too.

PR #976, opened 2019-10-24
D3D: Hack for pixel perfect mode

AlanDrake:
The StretchRect upscale makes the last row and column always one single pixel in size.
This workaround stretches the main area, right/bottom edges and bottom-right corner separately in order to avoid the distortion.

This is a hack. It would be better if we found the actual reason for which StretchRect causes this distortion, or render to texture and draw it on the scene.
Until then this will make pixel perfect rendering actually pixel perfect.

### DIFF
--- a/Engine/platform/windows/gfx/ali3dd3d.cpp
+++ b/Engine/platform/windows/gfx/ali3dd3d.cpp
@@ -734,7 +734,7 @@ int D3DGraphicsDriver::_initDLLCallback(const DisplayMode &mode)
 
 void D3DGraphicsDriver::InitializeD3DState()
 {
-  direct3ddevice->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_RGBA(30, 0, 0, 255), 0.5f, 0);
+  direct3ddevice->Clear(0, NULL, D3DCLEAR_TARGET, D3DCOLOR_RGBA(0, 0, 0, 255), 0.5f, 0);
 
   // set the render flags.
   direct3ddevice->SetRenderState(D3DRS_CULLMODE, D3DCULL_NONE);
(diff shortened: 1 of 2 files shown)


sonneveld:
I'm wondering if the better solution involves accommodating the half texel offsets.

Engine/platform/windows/gfx/ali3dd3d.cpp
```diff
@@ -1335,10 +1335,53 @@ void D3DGraphicsDriver::_render(bool clearDrawListAfterwards)
...
     {
-      throw Ali3DException("IDirect3DSurface9::StretchRect failed");
+        // Bug: StretchRect makes the last right and bottom pixel one single pixel, always, no matter how big you stretch (reason uknown)
+        // HACK: we draw the surface in four parts to ensure the last row and column have the proper size
+        const int hackOffset = pixelSize - 1;
+        const int src_width = _srcRect.GetWidth();
+        const int src_height = _srcRect.GetHeight();
+
+        const RECT rectsHack[4][2] = {
+            {   // main area
```
sonneveld: Is the main area still drawing a 1 pixel slice on the margins?

Engine/platform/windows/gfx/ali3dd3d.cpp
```diff
@@ -1335,10 +1335,53 @@ void D3DGraphicsDriver::_render(bool clearDrawListAfterwards)
...
+        // since the pixel size is 1 the distortion will not happen
+        if (direct3ddevice->StretchRect(pNativeSurface, NULL, pBackBuffer, &viewport_rect, filterType) != D3D_OK)
+        {
+            throw Ali3DException("IDirect3DSurface9::StretchRect failed");
+        }
+    } 
+    else // TODO find a way to make this workaround unnecessary
     {
-      throw Ali3DException("IDirect3DSurface9::StretchRect failed");
+        // Bug: StretchRect makes the last right and bottom pixel one single pixel, always, no matter how big you stretch (reason uknown)
```
sonneveld: I wonder if we're working around a driver issue here. I read here ( http://sjbrown.co.uk/2003/05/01/fix-directx-rasterisation/ ) that some drivers do not take into account half texels for StretchRect, so maybe this is a similar issue?

Engine/platform/windows/gfx/ali3dd3d.cpp
AlanDrake: Yes, that's why I'm accounting for one extra pixel

Engine/platform/windows/gfx/ali3dd3d.cpp
sonneveld: I experimented with using the direct3d reference device, and you do indeed get doubled up rows with this patch:
image

AlanDrake: I don't know, many other places seem to use shaders to perform the scaling or textures drawn with quads.
There are no mentions of this issue where StretchRect is used, so I wonder is it just us or is it overlooked?
I think I'd rather implement a D3D10 driver rather than trying again to fix something I haven't been able to fix since the first time we found out about the issue.

sonneveld:
For my testing, I was using this simple game: 
[smallresgame.zip](https://github.com/adventuregamestudio/ags/files/3770896/smallresgame.zip)

Engine/platform/windows/gfx/ali3dd3d.cpp
sonneveld: I think drawing with a textured quad would certainly give us more control over the problem. If you don't mind waiting, this might be better suited for after the SDL port goes in, as we rely on SDL to do the window management.

sonneveld:
I think that we should hold off on this pull request or rework it to use a textured quad instead of StretchRect. Depending on the driver, the work around will introduce other corruption issues.

ivan-mogilko:
> I think that we should hold off on this pull request or rework it to use a textured quad instead of StretchRect.

TBH I agree with this, I don't think the issue is actually critical, and iirc in all this time only 1 or 2 users talked about it.

AlanDrake:
I've upgraded this workaround with a bug detection routine.

AlanDrake:
I'm closing this PR because I managed to get the render to texture method to work.
If we happen to find incompatibility for some users we might implement this hack as an alternative.

PR #977, opened 2019-10-24
Add editable sprite properties for import width and height

### LINKED ISSUE #973: Replace Sprites from Source doesn't work correctly for Sprite 1 of a Tiled Sprite
If you import a sprite with the Tiled Sprite Import option, if you later use the Replace Sprites from Source context menu on the first sprite (offset 0, 0), it is replaced by the whole image instead of the first tile.

morganwillcock:
Two new properites specify import width and height, and re-importing a sprite now uses these properties. This fixes the case where a re-import of a sprite from offset 0,0 would reload the frame as the entire image, and also lets the user reconfigure the tile size.

Fixes #973

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -84,7 +84,7 @@ namespace AGS.Editor
          * 21: 3.5.0.15   - AudioClip ID.
          * 22: 3.5.0.18   - Settings.ScaleMovementSpeedWithMaskResolution.
         */
-        public const int    LATEST_XML_VERSION_INDEX = 22;
+        public const int    LATEST_XML_VERSION_INDEX = 23;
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
There's a serious issue I noticed with sprite data loading. Currently any format exception will cause rest of the fields skipped. Because you added new fields in the middle of reading/writing functions loading previously saved projects will result in loss of several last fields (Transparency, Frame etc).

I guess you need either always have new fields in the end or explicitly test project's format version.

morganwillcock:
I thought it was already patched forwards based on the index check, by the time the values are read? Or is this done afterwards?

morganwillcock:
Updated so that the original import options are left alone, and the two new items (width and height of the import) get reset to the sprite size if either is missing.

PR #982, opened 2019-11-03
Engine: D3D9 render to texture

AlanDrake:
This solves the StretchRect distortion in a clean way for pixel perfect mode by using a texture as a render target.

### DIFF
--- a/Engine/platform/windows/gfx/ali3dd3d.cpp
+++ b/Engine/platform/windows/gfx/ali3dd3d.cpp
@@ -268,6 +268,7 @@ D3DGraphicsDriver::D3DGraphicsDriver(IDirect3D9 *d3d)
   _legacyPixelShader = false;
   set_up_default_vertices();
   pNativeSurface = NULL;
+  pNativeTexture = NULL;
   availableVideoMemory = 0;
   _smoothScaling = false;
   _pixelRenderXOffset = 0;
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
I think there's an oversight in GetCopyOfScreenIntoBitmap now: you are creating an offscreen surface but it won't get released, because there's this condition remaining from previous code:
<code>
    surface->UnlockRect();
    if (!at_native_res) // native surface is released elsewhere
        surface->Release();
</code>
Guess now the surface should be released always, unconditionally?

AlanDrake:
You're right, I overlooked that one. I'll fix it at once.

ivan-mogilko:
I have another question. Function _renderToTexture accepts filter as an argument, but you get that filter value like this:
<pre>
    // use correct sampling method when stretching buffer to the final rect
    _filter->SetSamplerStateForStandardSprite(direct3ddevice);
    D3DTEXTUREFILTERTYPE filterType;
    direct3ddevice->GetSamplerState(0, D3DSAMP_MAGFILTER, (DWORD*)&filterType);
    <...>
    _renderFromTexture(filterType);
</pre>

then apply it in the function like this:

<pre>
    direct3ddevice->SetSamplerState(0, D3DSAMP_MINFILTER, filterType);
    direct3ddevice->SetSamplerState(0, D3DSAMP_MAGFILTER, filterType);</pre>

Note that this is exactly what line `_filter->SetSamplerStateForStandardSprite(direct3ddevice);` does above before the function call.

Is not this a redundant action, since you practically reset a sampler state that was just queried from the device?

AlanDrake:
I wanted to make it more generic because I was thinking about cases where it could be reused.
It's not really necessary at the moment, if you want I'll just move everything inside the function.

ivan-mogilko:
Yes, I think that would be better for this time. Because the renderer may be refactored more later, in some different way.

Another thing is that this somehow defeats the purpose of the filter object. In theory it may set other options, not only these two. Its just that it has not very flexible design (gfx filter interface is way too abstract imho). Maybe in the future there will be a filter object for each variant, and then you could just pass them around instead.

AlanDrake:
Done.

Issue #985, opened 2019-11-07
Game file name is not set after setting on new game


ericoporto:
**Describe the bug**
After creating a new game, and setting the game filename field, after the project loads, the game file name is `Empty Game`.

**AGS Version**
AGS 3.5.0.19

**To Reproduce**
1. Click to create a new game
2. At new game details, the second field should be the game filename field
3. This will correctly set the initial directory name
4. Unfortunately, the corresponding Game file name field will have the initial value set to `Empty Game`.

**Expected behavior**
I understand the idea is to be able to set the Game file name later, without having to change the directory name, but the message at the create new game dialog says it's going to be the name of the game. Also it says you "can't change later on", which is no longer true.

**Screenshots**
Here is a video

https://streamable.com/gen6c

![image](https://user-images.githubusercontent.com/2244442/68437036-75bcf600-019e-11ea-829d-0c77eb501eeb.png)

![image](https://user-images.githubusercontent.com/2244442/68437024-68077080-019e-11ea-9c06-c66c751a9783.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu Linux + Wine
 - Version 18.04

morganwillcock:
Possibly this is just some race condition or problem with WINE. I tried a recent build on Windows and it set 'Game file name' and 'Game title' to the values specified, when I used the old 'Empty game' template.

ericoporto:
Well, then just removing the "You can't change later on" sentence is fine.

morganwillcock:
Hmm, maybe not. I tried the latest CI build and the template that it comes with, and now I have the same problem. So it is probably something changed between 3.5.0.18 and 3.5.0.19, or something related to the template.

ericoporto:
The new setting [was introduced here](https://github.com/adventuregamestudio/ags/commit/4ca8f675ca2d0c8191997de577d53954c586b33e), at PR #766 , which dealt with issue #629.

ivan-mogilko:
Oh, I see, so I forgot to actually assign that input field value to the property.

ivan-mogilko:
Actually, the solution could be to not set GameFileName at all at project import, and remove it in templates. In such case directory name will be used, unless user changes the property. This sounds more logical too.

ericoporto:
Fixed by #986

PR #986, opened 2019-11-08
Update new game wizard for new filename

morganwillcock:
This just fixes what is already here, by retreiving the specified filename and setting it in the game properties. Plus fixes the label.

For #985
Edit: and separate issue that having no preference for the game directory did not fallback to using the documents folder


ericoporto:
This fixes for me the issue reported.

PR #993, opened 2019-11-22
Add sprite property 'ImportAsTile'

morganwillcock:
This allows you to re-import a sprite of a different size without needing to adjust the tile size, which fixes the issue of re-importing images which are not from a spritesheet and have been resized. There isn't enough information to automatically set the flag automatic for existing projects where there is already tile size information for the sprite, but I think the upgrade mechansim should cover every case except where the tile selection begins at 0,0 and the size is larger than any other tile from the same source.

Secondary changes were:
- internal import functions are modified to be more consistent in terms of the parameters that are passed
- import size and offset values are hidden from the property grid whilst ImportAsTile is set to false

I'm not entirely sure on the name ImportAsTile, but it does mean the optional properties are listed directly below this in the property grid, just based on the alphabetical order.

Any testing would be appreciated, I've been using it and it seems OK, but there might be something I've not clicked.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -84,7 +84,7 @@ namespace AGS.Editor
          * 21: 3.5.0.15   - AudioClip ID.
          * 22: 3.5.0.18   - Settings.ScaleMovementSpeedWithMaskResolution.
         */
-        public const int    LATEST_XML_VERSION_INDEX = 23;
+        public const int    LATEST_XML_VERSION_INDEX = 24;
         /*
          * LATEST_USER_DATA_VERSION is the last version of the user data file that used a
          * 4-point-4-number string to identify the version of AGS that saved the file.
--- a/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
+++ b/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
@@ -405,7 +405,7 @@ namespace AGS.Editor
             try
             {
                 SpriteTools.ReplaceSprite(replace, image, UseAlphaChannel, RemapToGamePalette,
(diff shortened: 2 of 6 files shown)


ivan-mogilko:
Well, it looks like working. Tested upgrading 3.4.3 project, previous 3.5.0 project (tile and non-tile sprites), and reimporting resized images in this branch too.

ivan-mogilko:
Something that I forgot to ask, would not it be more convenient to simply have sprite properties read in the order of format extension? Since it's XML the reading order could be any, but in that case there won't be a need to add extra try/catch around them to keep reading.

morganwillcock:
You mean read in the order that the version number was increased? That is probably a good idea. Shall I re-order them and add a comment for the corresponding version number?

ivan-mogilko:
> You mean read in the order that the version number was increased? That is probably a good idea. Shall I re-order them and add a comment for the corresponding version number?

Yes, certainly, if you think that will work (unless I missed anything).

ivan-mogilko:
Regarding above new commit ^, perhaps there was a misunderstanding. My proposal was not in simply changing the order of reading, but also removing the try/catch wrapping individual properties, while keeping a global one wrapping all of them. If they will be read in the order of format extension, then in case of old format exception will occur after all necessary fields are already read.

morganwillcock:
I think that increases the 'damage' if one property was missing. Perhaps the question is, does the deserialise assume that what is loaded is always 100% correct and it should reset the parent node to ensure that any values cannot be contradictory (although in this specific case that doesn't really apply), or should it limit any resets to just restoring defaults for missing values?

ivan-mogilko:
Alright, this is not relevant for current pr anyway; maybe one could introduce some helper function to reduce code repetition later?

PR #1000, opened 2019-12-07
[DO NOT MERGE] Switches palrender plugin to use script functions

ericoporto:
**DO NOT MERGE**
Because ags4 has the plugin interface for Characters, Objects and ViewFrames reduced, this switches palrender to use Script functions in place. Unfortunately, I don't have a game to test it.

```C++
int IAGSEngine::CallGameScriptFunction(const char *name, int32 globalScript, int32 numArgs, long arg1, long arg2, long arg3) {
    if (inside_script)
        return -300;

    ccInstance *toRun = GetScriptInstanceByType(globalScript ? kScInstGame : kScInstRoom);

    RuntimeScriptValue params[3];
    params[0].SetPluginArgument(arg1);
    params[1].SetPluginArgument(arg2);
    params[2].SetPluginArgument(arg3);
    int toret = RunScriptFunctionIfExists(toRun, (char*)name, numArgs, params);
    return toret;
}
```

My preoccupation is the above check for **`inside_script`**, because as far as I can tell, this would be true, so this would not work, but I don't understand enough to know if this is true, so I decided to do a PR as a question.

### DIFF
--- a/Plugins/agspalrender/ags_palrender.cpp
+++ b/Plugins/agspalrender/ags_palrender.cpp
@@ -29,6 +29,7 @@
 
 
 #include "plugin/agsplugin.h"
+#include "script/executingscript.h"
 #include "palrender.h"
 #include "raycast.h"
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
inside_script means that there's another script function running at this time.
Plugins may only call functions during event callbacks. So if you call CallFunction from event callback, and not, for example, from some thread that calls engine when a timer runs out, then you should be safe.

ivan-mogilko:
I think [this game](https://www.adventuregamestudio.co.uk/site/games/game/2170/) uses the palrender plugin, but it's made with AGS 3.21, which no longer run in this branch. Maybe you could try your changes in ags3 branch first to see if they work. Or ask Scavenger to give some test game that works in 3.4.1+.

ericoporto:
ok, so this won't work since the function will run inside a script and not an event callback, so I will close the PR.

ivan-mogilko:
@ericoporto I did not pay attention first, but you were using wrong function. CallGameScriptFunction is used to call a **user script function**, not an engine API function.

You need to use GetScriptFunctionAddress to get these.

See commit 70983bbeb9896477ea9029b849cd0f780bc0942a

ericoporto:
@ivan-mogilko I see, thanks, I can try that!

ericoporto:
OK, I submitted #1001

PR #1011, opened 2019-12-18
Editor: remove transparency color from all sprite previews

morganwillcock:
This removes magenta from all sprite previews in the Editor, where the presence of the magenta is meant to indicate transparency. Whether this is correct or not, I'm not sure, as it isn't obivous why some parts of the Editor show it and others don't. I thought submitting as a PR to let people try the build is the best way to describe the change that it provides. The main difference is a slightly slower redraw within the View editor (although it was already slow) since all previews are now going through the same function that provides the previews for the main sprite viewer.

So, anyone's opinion either way would be appreciated...

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -1666,18 +1666,9 @@ namespace AGS.Editor
         }
 
         void IGUIController.DrawSprite(Graphics g, int spriteNumber, int x, int y, int width, int height, bool centreHorizontally)
-		{
-            SpriteInfo info = Factory.NativeProxy.GetSpriteInfo(spriteNumber);
-
-            int spriteWidth = info.Width;
-			int spriteHeight = info.Height;
-            // Draw low-res sprites larger (TODO: find out why, perhaps just for the better looks?)
-            if (info.Resolution == SpriteImportResolution.LowRes)
-            {
-                spriteWidth *= 2;
-                spriteHeight *= 2;
-            }
-			Bitmap bmp = Utilities.GetBitmapForSpriteResizedKeepingAspectRatio(new Sprite(spriteNumber, spriteWidth, spriteHeight), width, height, centreHorizontally, false, SystemColors.Control);
+        {
+            Sprite sprite = Factory.AGSEditor.CurrentGame.RootSpriteFolder.FindSpriteByID(spriteNumber, true);
+            Bitmap bmp = Utilities.GetBitmapForSpriteResizedKeepingAspectRatio(sprite, width, height, centreHorizontally, false);
 			g.DrawImage(bmp, x, y);
 			bmp.Dispose();
 		}
--- a/Editor/AGS.Editor/Panes/CursorEditor.cs
+++ b/Editor/AGS.Editor/Panes/CursorEditor.cs
@@ -52,11 +52,15 @@ namespace AGS.Editor
                 IMAGE_SCALE_FACTOR = Factory.AGSEditor.CurrentGame.GUIScaleFactor;
 
                 if (_item.Image > 0)
-				{
-					IntPtr hdc = e.Graphics.GetHdc();
-					Factory.NativeProxy.DrawSprite(hdc, 0, 0, _item.Image);
-					e.Graphics.ReleaseHdc();
-				}
+                {
+                    Sprite sprite = Factory.AGSEditor.CurrentGame.RootSpriteFolder.FindSpriteByID(_item.Image, true);
+
+                    using (Bitmap bmp = Utilities.GetBitmapForSpriteResizedKeepingAspectRatio(sprite, sprite.Width * IMAGE_SCALE_FACTOR, sprite.Height * IMAGE_SCALE_FACTOR, false, false))
+                    {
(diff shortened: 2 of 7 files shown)


ericoporto:
Removed:
Screenshot from 2019-12-18 21-46-41
Original:
Screenshot from 2019-12-18 21-55-04

I still have some magenta in the preview that animates and the GUI editor.

I see many programs using a checkered board pattern to signify transparency, but I think I prefer the magenta, it's usually a strong color not very used in large areas and it doesn't make things busy like checkered patterns. The same color meant transparency in older versions of RPG Maker, so I am kinda used to it too.

The confusing bit is the sprite viewer used white background for alpha transparency and magenta for import with corner color as transparency, so for the sprite viewer, it looks more consistent with the magenta removed.

ivan-mogilko:
From what I found, in the sprite manager the sprites with "AlphaChannel = false" property are displayed with magenta transparency, and ones with "AlphaChannel = true" are displayed with white background (which probably means alpha value = 0, or maybe < 255)

ivan-mogilko:
I am honestly not sure what to think about this.
On one hand, the previews are definitely inconsistent: white & magenta in sprite manager (depending on alpha channel flag) and only magenta in animation preview and other places.
On the other hand, users may have get used to magenta for all the past years and changing it to plain white may cause confusion (maybe?). Also, as you say, this will make editor work slower.

I guess the questions here are:
* do we need an indication of alpha channel in transparency color? otherwise all of the transparent parts should be displayed similarily.
* what background should editor display behind transparent parts?

Also, do you think there's a chance to have a white outline/sprite, and if it's is more or less common than magenta one?

morganwillcock:
> On the other hand, users may have get used to magenta for all the past years and changing it to plain white may cause confusion (maybe?).

It is very tricky, just from the screenshots above you can see that the magenta is present on some of those sprites used in the views, but when they are displayed in the view editor they are *all* magenta. So I think there is at least an argument to be made to make the display more consistent, and make the magenta always represent non-alpha-channel transparency.

> Also, do you think there's a chance to have a white outline/sprite, and if it's is more or less common than magenta one? E.g. how would a white mouse cursor look in the new preview?

Presumably this is why the sprite preview also draws a box around the original import, so that when something is white on a white background you can still see that something is there (although you would only be able to tell the aspect ratio of it, not what it was or how big it was).

For this branch, perhaps the only workable solution is to have a variable background color for each location where the sprites are displayed, although I'm not sure that is the best investment of anyone's time without a clear plan of future changes (that might undo these).

tzachshabtay:
The best way I know of to show transparency is with alternating squares, for example: https://images.app.goo.gl/3vtcZtvzvgSFQPKZ6

messengerbag:
I agree with alternating squares as being pretty standard and probably the best option. Of course, no single background will work for every possible sprite (what if your sprite is an alternating chessboard pattern?). To get around that, you could let the user define the background, or you could have an animated one. (But that's probably overkill.)

persn:
1+ for alternating squares.

This is off topic but worth mentioning, I noticed this PR makes code changes that gets rid of several calls to `NativeProxy`, no matter what we end up doing it would be nice to get those changes merged into `master`, then we could maybe kill off some editor C++ code

ivan-mogilko:
@morganwillcock, do you have any interest in this pr?

Judging by above comments, the proposed implementation was to:
1) Have every sprite preview have a consistent but variable background (magenta for no alpha channel, white for ones having alpha);
2) As a bonus, have checkered background for sprites with alpha channel. (Or maybe have it everywhere?)

How complicated this sounds? p 1 probably requires to put a condition where you replace background colour. p 2 could maybe be done by trans-blitting a sprite over a checkered board.

If you don't want to work on this, maybe someone else could look into adjusting your code?

If there's no interest in this change at all anymore, then we should close the pr.

morganwillcock:
> @morganwillcock, do you have any interest in this pr?

It is inconsistent as it is, but I think people are familiar with the inconsistencies. So, probably not, since the only correct thing to do seems to be to re-implement all sprite displays across all form controls.

rofl0r:
although not an editor user, i'd vote for leaving magenta/color1 for sprites that do not use alpha channel, and checkerboard-style for those that do, so the user has a visual feedback which kind of image he's using.

ivan-mogilko:
So, I've been considering updating/reimplementing this, but feel like changing my mind.

* Converting all the bitmaps from native to .NET objects is an extra step which means things will be getting slower especially in the View preview for the high-resolution games (they probably already are in the Sprite Manager).
* Hence keeping drawing in the native will be benefical for perfomance, but there's no method for proper alpha blending to .NET surface (HDC) out of the box in Allegro (only plain blitting and transparent draw with magic pink transparency). This means we won't have a correct "alpha" sprites preview anyway unless we write our own again in native code...

So, what we may have is to a) have consistent background, b) may add checkered background for "alpha" sprites, while keeping drawing in the native code (maybe factoring out few functions). Idk if that's really worth spending time on.

The *only* reason we have to use native code to draw these in the preview panels is that Editor reads sprites from the native SpriteCache array. If we really follow the plan and have editor work with the sprites directly, and/or having them directly on disk in project source, then any changes made now will become redundant anyway.

Issue #1010, opened 2019-12-16
3.5.0: IsKeyPressed() compatibility broken


ivan-mogilko:
**Describe the bug**
Historically AGS script functions returned integers, and before "bool" type was introduced in the script syntax users did compare logical return value to 1 or 0.

There was [some refactoring in ags key handling](https://github.com/adventuregamestudio/ags/pull/934) recently, and as an oversight IsKeyPressed returns not strictly 1 or 0, but a value stored inside Allegro implementation of keys array. It was just recently found that certain keys actually have -1 stored there instead of 1 (maybe there are other variants, and in theory that may be anything that `char` can store).

This breaks existing scripts (was reported by a user), and supposedly may break existing games where developers treated return value as an integer instead of a boolean and compared to '1'.

**IMPORTANT**: This is *not only for backwards compatibility*. Unfortunately, in AGS script `bool`, `true` and `false` are not real type and keywords, but macros defined as `int`, `1` and `0`. Therefore following won't work either:
<pre>
if (IsKeyPressed(eKeySpace) == true)
</pre>

**AGS Version**
3.5.0.22.

**To Reproduce**
* Run AGS 3.4.3 (previous stable editor), create a game from a non-Empty template (for simplicity) and put following in GlobalScript.asc:
<pre>
function repeatedly_execute()
{
    if (IsKeyPressed(eKeySpace) == 1) Display("SPACE WAS PRESSED");
}
</pre>
* Run the game, press and hold Space button. Observe a blinking message.
* Now copy the script to a game made in 3.5.0. If you run it, pressing Space will not have this result.
* Change script to
<pre>
function repeatedly_execute()
{
    if (IsKeyPressed(eKeySpace)) Display("SPACE WAS PRESSED");
}
</pre>
* Run game and observe that now it's displaying the message as expected.

**Expected behavior**
Every script function declared with 'bool' return value, but implemented as returning `int` (for historical reasons or backwards compatibility) must explicitly return only 0 or 1.

**Known workarounds**
Test IsKeyPressed() as a condition directly - that will convert its return value to proper boolean range.

ivan-mogilko:
Fixed by #1012

PR #1012, opened 2019-12-19
Ensure IsKeyPressed only returns 1 or 0

sonneveld:
To fix script compatibility. Because allegro4's `key` array can hold values other than 0 or 1, this breaks scripts as the values are converted to script bool (which is really just an int define)

Fixes https://github.com/adventuregamestudio/ags/issues/1010

### DIFF
--- a/Engine/ac/sys_events.cpp
+++ b/Engine/ac/sys_events.cpp
@@ -53,7 +53,7 @@ int ags_kbhit () {
 int ags_iskeypressed (int keycode) {
     if (keycode >= 0 && keycode < __allegro_KEY_MAX)
     {
-        return key[keycode];
+        return key[keycode] != 0;
     }
     return 0;
 }


ivan-mogilko:
@sonneveld Please, rebase the pr and target "release-3.5.0" branch , as this is supposed to be a hotfix without adding anything from "ags3" branch.

sonneveld:
Apologies. Done!

ericoporto:
Just to take the opportunity to learn something new, I thought this was going to be implemented with:

```return key[keycode] != 0 ? 1 : 0 ;```

But the implementation in the commit actually works. Why does it works?

ivan-mogilko:
> Just to take the opportunity to learn something new, I thought this was going to be implemented with:
> 
> `return key[keycode] != 0 ? 1 : 0 ;`
> 

I would also like it to be implemented like that, but decided not to nitpick. I suspect we will have to change to to that later anyway, because some compilers will post a warning.


> But the implementation in the commit actually works. Why does it works?

It works because of implicit type cast, when two types may be converted between each other easily.

sonneveld:
Ivan is correct. Bool is implicitly converted to int. However in this case it is an expression and expressions in C return 1 or 0.

ericoporto:
I read a little online now and I think I understand. I don't like that the comment on commit isn't expressed on the code -> if I read a ternary expression I would know why the expression is there in the code in the future. Because there's not, now I need to use git blame to remind me why there's a different than 0 comparison and reason with the returned type be expected as int. This may be only me though since I am not that familiar with the language and still learning.

ivan-mogilko:
>  Because there's not, now I need to use git blame to remind me why there's a different than 0 comparison and reason with the returned type be expected as int.

I actually agree with this. The return type of a function is "int". The key array is filled with ints. So it is pretty unclear why it has to do the implicit int->bool->int conversion.

PR #1018, opened 2019-12-22
Automatic font outlines (2)

ivan-mogilko:
This is a remake of @fernewelten's #892 PR. Old one still had several mistakes, but unfortunately I didn't know if and when fernewelten is planning to complete it, so decided to speed things up and fix it on my own.

Following is the original description:

---

_This is proposed code for issue #878 (automatic outlines)._
_Added two additional fields to fonts:_
- _`AutoOutlineStyle` (`Rounded`, `Squared`)._
- _`AutoOutlineThickness` (positive integer, or 0 for default)_ 
_(I couldn't use "1" as the default for `AutoOutlineThickness` because currently, scaled bitmap fonts get an outline that is thicker than 1)._ 

_The fields will only show in the editor if `OutlineStyle==Automatic`._
_The field `OutlineFont` will now only show if `OutlineStyle==UseOutlineFont`._

_The algo for generating the outline is based on messengerbag's. The text is only rendered _once_ per outline into a local bitmap; from then on, the bitmap is blitted repeatedly. Another local bitmap accumulates the vertical blits. In all, the number of blits is proportional to the thickness of the outline._ 

_For anti-aliased ttf fonts, a thickness of 2 to 3 seems to work nicely._

---

This PR consists of two parts: first are original fernewelten's commits (some minor later tweaks are squashed into bigger ones), and second part is my fixes following this list of issues: https://github.com/adventuregamestudio/ags/pull/892#issuecomment-522987666

**UPDATE**
Also, fixed following problem noted in fernewelten's original post:

> I couldn't use "1" as the default for `AutoOutlineThickness` because currently, scaled bitmap fonts get an outline that is thicker than 1

According to my research, this means the font outline is doubled (becomes = 2) under following condition:
* it is a hi-res game (according to legacy resolution rules);
* it is a bitmap font;
* it is a scaled up font (general settings have the deprecated "fonts are made for hi-res games" turned off, or newer format has individual scaling multiplier set to > 1).

It's possible to detect this condition during the format upgrade, in both editor and engine, so this will no longer be a concern.

### DIFF
--- a/Editor/AGS.Types/Font.cs
+++ b/Editor/AGS.Types/Font.cs
@@ -120,7 +120,7 @@ namespace AGS.Types
             set { _outlineStyle = value; }
         }
 
-        [Description("Thickness of the automatic outline")]
+        [Description("Thickness of the automatic outline. WARNING: automatic outlines are generated at runtime and large sizes may negatively affect your game performance.")]
         [Category("Appearance")]
         [DefaultValue(1)]
 		public int AutoOutlineThickness
--- a/Editor/AGS.Types/Font.cs
+++ b/Editor/AGS.Types/Font.cs
@@ -7,8 +7,8 @@ using System.Xml;
 namespace AGS.Types
 {
     [DefaultProperty("OutlineStyle")]
-    public class Font
-    {
+    public class Font : ICustomTypeDescriptor
+	{
         private int _id;
         private string _name;
         private int _pointSize;
@@ -19,14 +19,16 @@ namespace AGS.Types
         private int _sizeMultiplier = 1;
         private int _verticalOffset;
         private int _lineSpacing;
+		private int _autoOutlineThickness = 0;
+		private FontAutoOutlineStyle _autoOutlineStyle = FontAutoOutlineStyle.Rounded;
 
         public Font()
         {
-            _outlineStyle = FontOutlineStyle.None;
-            _outlineFont = 0;
             _name = string.Empty;
             _pointSize = 0;
-            _fontHeight = 0;
+			_outlineFont = 0;
+			_outlineStyle = FontOutlineStyle.None;
+			_fontHeight = 0;
             _lineSpacing = 0;
         }
 
@@ -101,7 +103,7 @@ namespace AGS.Types
             get { return "Font: " + this.Name; }
         }
 
-        [Description("Font to use as an outline for this one (only if FontOutlineStyle is OutlineFont)")]
+        [Description("Font to use as an outline for this one")]
         [Category("Appearance")]
         public int OutlineFont
         {
@@ -111,11 +113,28 @@ namespace AGS.Types
 
         [Description("Whether this font should be drawn with an outline")]
         [Category("Appearance")]
-        public FontOutlineStyle OutlineStyle
+		[RefreshProperties(RefreshProperties.All)]
+		public FontOutlineStyle OutlineStyle
         {
             get { return _outlineStyle; }
             set { _outlineStyle = value; }
         }
+
+		[Description("Thickness of the automatic outline (0 = default)")]
+		[Category("Appearance")]
+		public int AutoOutlineThickness
+		{
+			get { return _autoOutlineThickness; }
+			set { _autoOutlineThickness = value; }
+		}
+
+		[Description("Style of the automatic outline")]
+		[Category("Appearance")]
+		public FontAutoOutlineStyle AutoOutlineStyle
+		{
+			get { return _autoOutlineStyle; }
+			set { _autoOutlineStyle = value; }
+		}
 
 		[Description("The file path that this font was imported from")]
 		[Category("Design")]
@@ -178,5 +197,84 @@ namespace AGS.Types
             SerializeUtils.SerializeToXML(this, writer);
         }
 
-    }
+		#region ICustomTypeDescriptor Members
+		public AttributeCollection GetAttributes()
+		{
+			return TypeDescriptor.GetAttributes(this, true);
+		}
+
+		public string GetClassName()
+		{
+			return TypeDescriptor.GetClassName(this, true);
+		}
+
+		public string GetComponentName()
+		{
+			return TypeDescriptor.GetComponentName(this, true);
+		}
+
+		public TypeConverter GetConverter()
+		{
+			return TypeDescriptor.GetConverter(this, true);
+		}
+
+		public EventDescriptor GetDefaultEvent()
+		{
+			return TypeDescriptor.GetDefaultEvent(this, true);
+		}
+
+		public PropertyDescriptor GetDefaultProperty()
+		{
+			return TypeDescriptor.GetDefaultProperty(this, true);
+		}
+
+		public object GetEditor(Type editorBaseType)
+		{
+			return TypeDescriptor.GetEditor(this, editorBaseType, true);
+		}
+
+		public EventDescriptorCollection GetEvents()
+		{
+			return TypeDescriptor.GetEvents(this, true);
+		}
+
+		public EventDescriptorCollection GetEvents(Attribute[] attributes)
+		{
+			return TypeDescriptor.GetEvents(this, attributes, true);
+		}
+
+		public PropertyDescriptorCollection GetProperties()
+		{
+			return TypeDescriptor.GetProperties(this, true);
+		}
+
+		public PropertyDescriptorCollection GetProperties(Attribute[] attributes)
+		{
+			PropertyDescriptorCollection properties = TypeDescriptor.GetProperties(this, attributes, true);
+			List<PropertyDescriptor> wantedProperties = new List<PropertyDescriptor>();
+			foreach (PropertyDescriptor property in properties)
+			{
+				if (property.Name == "AutoOutlineStyle" ||
+					property.Name == "AutoOutlineThickness")
+				{
+					if (_outlineStyle != FontOutlineStyle.Automatic)
+						continue;
+				}
+				else if (property.Name == "OutlineFont")
+				{
+					if (_outlineStyle != FontOutlineStyle.UseOutlineFont)
+						continue;
+				}
+
+				wantedProperties.Add(property);
+			}
+			return new PropertyDescriptorCollection(wantedProperties.ToArray());
+		}
+
+		public object GetPropertyOwner(PropertyDescriptor pd)
+		{
+			return this;
+		}
+		#endregion
+	}
 }
--- a/Editor/AGS.Types/Font.cs
+++ b/Editor/AGS.Types/Font.cs
@@ -19,7 +19,7 @@ namespace AGS.Types
         private int _sizeMultiplier = 1;
         private int _verticalOffset;
         private int _lineSpacing;
-		private int _autoOutlineThickness = 0;
+        private int _autoOutlineThickness = 1;
 		private FontAutoOutlineStyle _autoOutlineStyle = FontAutoOutlineStyle.Rounded;
 
         public Font()
@@ -120,17 +120,24 @@ namespace AGS.Types
             set { _outlineStyle = value; }
         }
 
-		[Description("Thickness of the automatic outline (0 = default)")]
-		[Category("Appearance")]
+        [Description("Thickness of the automatic outline")]
+        [Category("Appearance")]
+        [DefaultValue(1)]
 		public int AutoOutlineThickness
 		{
 			get { return _autoOutlineThickness; }
-			set { _autoOutlineThickness = value; }
+            set
+            {
+                if (value < 1)
+                    throw new ArgumentException("AutoOutlineThickness must be 1 or greater.");
+                _autoOutlineThickness = value;
+            }
 		}
 
 		[Description("Style of the automatic outline")]
 		[Category("Appearance")]
-		public FontAutoOutlineStyle AutoOutlineStyle
+        [DefaultValue(FontAutoOutlineStyle.Rounded)]
+        public FontAutoOutlineStyle AutoOutlineStyle
 		{
 			get { return _autoOutlineStyle; }
 			set { _autoOutlineStyle = value; }
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -101,6 +101,9 @@ Font custom line spacing.
 Sprites have "real" resolution. Expanded FontInfo data format.
 Option to allow legacy relative asset resolutions.
 
+51 :
+Fonts have adjustable outline
+
 */
 
 enum GameDataVersion
@@ -134,7 +137,8 @@ enum GameDataVersion
     kGameVersion_341            = 48,
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
-    kGameVersion_Current        = kGameVersion_350
+    kGameVersion_351            = 51,
+    kGameVersion_Current        = kGameVersion_351
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -143,6 +143,12 @@ void GameSetupStruct::read_font_infos(Common::Stream *in, GameDataVersion data_v
             fonts[i].YOffset = in->ReadInt32();
             fonts[i].LineSpacing = Math::Max(0, in->ReadInt32());
             AdjustFontInfoUsingFlags(fonts[i], flags);
+            if (data_ver >= kGameVersion_351)
+            {
+                fonts[i].AutoOutlineThickness = in->ReadInt32();
+                fonts[i].AutoOutlineStyle = 
+                    static_cast<enum FontInfo::AutoOutlineStyle>(in->ReadInt32());
+            }
         }
     }
 }
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -226,6 +226,12 @@ struct SpriteInfo
 // multiple lines, and similar cases.
 struct FontInfo
 {
+    enum AutoOutlineStyle : int
+    {
+        kRounded = 0,
+        kSquared = 1,
+    };
+
     // General font's loading and rendering flags
     uint32_t      Flags;
     // Font size, in points (basically means pixels in AGS)
@@ -238,6 +244,10 @@ struct FontInfo
     int           YOffset;
     // custom line spacing between two lines of text (0 = use font height)
     int           LineSpacing;
+    // When automatic outlining, thickness of the outline (0 = use legacy thickness)
+    int           AutoOutlineThickness;
+    // When automatic outlining, style of the outline
+    AutoOutlineStyle AutoOutlineStyle;
 
     FontInfo();
 };
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -62,7 +62,7 @@ FontInfo::FontInfo()
     , YOffset(0)
     , LineSpacing(0)
     , AutoOutlineStyle(kRounded)
-    , AutoOutlineThickness(0)
+    , AutoOutlineThickness(1)
 {}
 
 
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -61,6 +61,8 @@ FontInfo::FontInfo()
     , Outline(FONT_OUTLINE_NONE)
     , YOffset(0)
     , LineSpacing(0)
+    , AutoOutlineStyle(kRounded)
+    , AutoOutlineThickness(0)
 {}
 
 
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -154,6 +154,13 @@ int get_font_outline(size_t font_number)
     return fonts[font_number].Info.Outline;
 }
 
+int get_font_outline_thickness(size_t font_number)
+{
+    if (font_number >= fonts.size())
+        return 0;
+    return fonts[font_number].Info.AutoOutlineThickness;
+}
+
 void set_font_outline(size_t font_number, int outline_type)
 {
     if (font_number >= fonts.size())
--- a/Common/font/fonts.h
+++ b/Common/font/fonts.h
@@ -44,6 +44,7 @@ bool font_supports_extended_characters(size_t fontNumber);
 // at random times (usually - drawing routines).
 // Need to check whether it is safe to completely remove it.
 void ensure_text_valid_for_font(char *text, size_t fontnum);
+// Get font's scaling multiplier
 int get_font_scaling_mul(size_t fontNumber);
 // Calculate actual width of a line of text
 int wgettextwidth(const char *texx, size_t fontNumber);
@@ -55,7 +56,11 @@ int getfontheight(size_t fontNumber);
 int getfontlinespacing(size_t fontNumber);
 // Get is font is meant to use default line spacing
 bool use_default_linespacing(size_t fontNumber);
+// Get font's outline type
 int  get_font_outline(size_t font_number);
+// Get font's automatic outline thickness (if set)
+int  get_font_outline_thickness(size_t font_number);
+// Set font's outline type
 void set_font_outline(size_t font_number, int outline_type);
 // Outputs a single line of text on the defined position on bitmap, using defined font, color and parameters
 int getfontlinespacing(size_t fontNumber);
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -28,6 +28,7 @@
 #include "util/path.h"
 #include "util/string_compat.h"
 #include "util/string_utils.h"
+#include "font/fonts.h"
 
 namespace AGS
 {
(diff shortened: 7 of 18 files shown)


ivan-mogilko:
Note, I think I found a way to get rid of special value "0" for outline thickness and instead set thickness to either 1 or 2 depending on game and font settings during game import process (see UPDATE in the end of first post).

ivan-mogilko:
How to test the old game import/run.

1. All fonts from old games should have outline = Squared.
2. Any fonts from low-res game (like 320x200) should have thickness = 1.
3. In hi-res game (like 640x400 and above) WFN fonts that either
a) made by prior to 3.5.0 and have "Fonts designed for hi-res games" setting = false, or
b) made in 3.5.0 and have SizeMultiplier > 1,
should have thickness = 2

ivan-mogilko:
UPDATE: fixed following problem:

> New functionality (engine).
>* I found that if I set large thickness, then the engine cannot correctly calculate size of bitmap necessary to store the text line.
In my test this happened when I took font with height = 9 and assigned outline thickness = 9.

I believe it's important to note thought, that with larger outlines text lines are spread pretty far, which may be fixed by setting manual LineSpacing. When trying that I also found another potential future problem: ags does not know how "merge" outlines from different lines of text, and if they are too close the outline from the lower line overlap upper line. Guess the possible solution would demand a different drawing algorithm: where outline is drawn first for all the lines, and actual text is drawn only after (also for all the lines).

ericoporto:
[test_outlines.zip](https://github.com/adventuregamestudio/ags/files/3998485/test_outlines.zip)

Screenshot from 2019-12-24 11-18-09

Hey, I did some tests, and things appear to work. On anti aliased fonts though, I get a weird artifact on the font at some specific points - on the white body at right you can see the font without outlines has no lines on top. Also on Roboto font, you can see some lines on outline on the back. Also dismiss the wrongly placed square there at the S letter, it's just a start from the background! xD

ivan-mogilko:
> On anti aliased fonts though, I get a weird artifact on the font at some specific points - on the white body at right you can see the font without outlines has no lines on top.

Maybe it's alpha channel issue? and these vertical lines are where certain letter pixels became half-transparent and you can see black outline behind them? just a thought.
But did same happen before or it appeared only after these changes?

ericoporto:
Screenshot from 2019-12-24 14-01-28

Above is the previous result - they didn't had artifacts. Click on the images to load on a new tab and expand them.

Some comments on the c++ code, bear in mind I had difficulties trying to follow it and understand why it's as is:
- wouttextxy_AutoOutline arguments are reversed when comparing to other functions.
- `Bitmap *outline_stencil = CreateTransparentBitmap(t_width, t_height + 2 * thickness, ds->GetColorDepth());` : Why it doesn't have the same 2 times thickness on the width too?
-   why this below is done in two lines
```
wouttextxy(texx_stencil, 0, 0, font, color, texx);
wouttextxy_AutoOutline_Semitransparent2Opaque(texx_stencil);
```
- shouldn't the blit to ds only happen once?

ivan-mogilko:
Oh, I did not look into the drawing itself yet, I will check that out.

ericoporto:
Uhm, other thing, because on AGS we can't dynamically change font properties like font size, or in this case the thickness of the outline, if you feel you need to, it's alright to pre-generate the outline on the Editor - at build time, if it's not updated.

ivan-mogilko:
> Uhm, other thing, because on AGS we can't dynamically change font properties like font size, or in this case the thickness of the outline, if you feel you need to, it's alright to pre-generate the outline on the Editor - at build time, if it's not updated.

I don't think this kind of optimization should be done in the current PR.
Also, doing this on Editor's side would mean extra compilation step and creating a font cache, with all the consequences of maintaining it, so it's a serious change (and Idk how to do that for TTF fonts).

ericoporto:
image

No worries. Just to emphasize the artifacts, above are weird horizontal lines in the outline itself, and on the text at front, the vertical lines.

ivan-mogilko:
@fernewelten , if you see this, could you explain the purpose of using wouttextxy_AutoOutline_Semitransparent2Opaque()?

ericoporto:
What does this blit thing does? I feel the bits are being laid on top of other without being actually blended.

ivan-mogilko:
> What does this blit thing does? I feel the bits are being laid on top of other without being actually blended.

I am pretty much sure this is the case. For blending there are other functions.

ericoporto:
OK, this means the resulting alpha is always the last one thrown at that pixel.

ivan-mogilko:
@ericoporto 

> shouldn't the blit to ds only happen once?

if you are referring to this:
<pre> // stamp the outline stencil to the left and right of the text
        ds->Blit(outline_stencil, xxp - x_diff, outline_y, kBitmap_Transparency);
        if (x_diff > 0)
            ds->Blit(outline_stencil, xxp + x_diff, outline_y, kBitmap_Transparency);</pre>
This code draws dark font with left and right shift so that outline appears on both sides horizontally.

ivan-mogilko:
Alright, I was probably confused at first. What `Semitransparent2Opaque` is used for - is to increase opaqueness of the _outline shadow_. How may come it is translucent in the first place? Is it what may happen if you outline piece with a TTF font (or anti-aliased TTF font)?

But in this case, should not it be simply made 100% opaque by stripping alpha channel without this extra algorithm?

ericoporto:
I found something online -> https://d1cxvcw9gjxu2x.cloudfront.net/attachments/606810

It was here: ttpss://www.allegro.cc/forums/thread/611384

Not sure if relevant or if it works.

ivan-mogilko:
I begin to doubt that this PR should be adding a special algorithm for TTF outlines at all. This seem to be superflous for now, and maybe better left until real problem is stated. Perhaps it will be found more optimal to actually use outlining abilities of TTF library.

ivan-mogilko:
NOTE: I adjusted your test game by reducing resolution and adding another non-transparent GUI underneath the labels to make errors more visible:
[test_outlines_640x480.zip](https://github.com/adventuregamestudio/ags/files/3999118/test_outlines_640x480.zip)

messengerbag:
> Alright, I was probably confused at first. What `Semitransparent2Opaque` is used for - is to increase opaqueness of the _outline shadow_. How may come it is translucent in the first place? Is it what may happen if you draw outline piece with a TTF font (or anti-aliased TTF font)?

Yes, exactly. The naive way of drawing outlines for anti-aliased fonts tends to create semi-opaque sections (I've experienced this with the SpeechBubble module). The function looks like an attempt to turn the outline version opaque -- compare this comment:

https://github.com/adventuregamestudio/ags/issues/878#issuecomment-515603670

I would also assume there is a better way. (I would also add that for the best results, there should be two versions of the outline stamp: the fully opaque one used to fill the outline, and the anti-aliased one used along the edge of the outline so it will also be anti-aliased.)

But if there is a way to do this outlining in the font library I would think that would be better.

ericoporto:
A second step after this PR is done is probably text caching, so if you draw something, next time you just get it from the cache if it's there draw it instead of constructing it again. I don't remember how text drawing works in AGS, if it's figured on every frame or only on the first call.

ivan-mogilko:
> I don't remember how text drawing works in AGS, if it's figured on every frame or only on the first call.

Can't tell for 100% cases, but in majority of them the text and other elements are only prepared once, and then sent into renderer each frame while the message is same and image does not have to be updated for any reason.

ericoporto:
Gave a try, but it didn't work :grimacing: 

```C++
// Draw outline that is calculated from the text font, not derived from an outline font
void wouttextxy_AutoOutline(Bitmap *ds, size_t font, int32_t color, const char *texx, int &xxp, int &yyp)
{
    using AGS::Common::kBitmap_Transparency;

    int thickness = game.fonts.at(font).AutoOutlineThickness;
    auto style = game.fonts.at(font).AutoOutlineStyle;
    if (thickness <= 0)
        return; 

    size_t const t_width = wgettextwidth(texx, font);
    size_t const t_height = wgettextheight(texx, font);

    
    // move start of text so that the outline doesn't drop off the bitmap
    xxp += thickness;
    int const outline_y = yyp;
    yyp += thickness;
    
    int largest_y_diff_reached_so_far = -1; 
    for (int x_diff = thickness; x_diff >= 0; x_diff--)
    {
        // Integer arithmetics: In the following, we use terms k*(k + 1) to account for rounding.
        //     (k + 0.5)^2 == k*k + 2*k*0.5 + 0.5^2 == k*k + k + 0.25 ==approx. k*(k + 1)
        int y_term_limit = thickness * (thickness + 1);
        if (FontInfo::kRounded == style)
            y_term_limit -= x_diff * x_diff;

        // extend the outline stencil to the top and bottom
        for (int y_diff = largest_y_diff_reached_so_far + 1; 
            y_diff <= thickness && y_diff * y_diff  <= y_term_limit; 
            y_diff++)
        {
            printf("x_diff=%d , y_diff=%d, text=%s \n", x_diff,y_diff, texx);

            wouttextxy(ds, xxp - x_diff, outline_y + thickness - y_diff, font, color, texx);
            wouttextxy(ds, xxp, outline_y + thickness - y_diff, font, color, texx);
            if (x_diff > 0)
                wouttextxy(ds, xxp + x_diff, outline_y + thickness - y_diff, font, color, texx);


            wouttextxy(ds, xxp - x_diff, outline_y + thickness , font, color, texx);
            wouttextxy(ds, xxp, outline_y + thickness , font, color, texx);
            if (x_diff > 0)
                wouttextxy(ds, xxp + x_diff, outline_y + thickness , font, color, texx);

            if (y_diff > 0) {
                wouttextxy(ds, xxp - x_diff, outline_y + thickness + y_diff, font, color, texx);
                wouttextxy(ds, xxp, outline_y+ thickness + y_diff, font, color, texx);
                if (x_diff > 0)
                    wouttextxy(ds, xxp + x_diff, outline_y + thickness + y_diff, font, color, texx);
            }
            largest_y_diff_reached_so_far = y_diff;
        }
    }
}
```

Console log
```bash
x_diff=5 , y_diff=0, text=test_outlines 
x_diff=5 , y_diff=1, text=test_outlines 
x_diff=5 , y_diff=2, text=test_outlines 
x_diff=4 , y_diff=3, text=test_outlines 
x_diff=3 , y_diff=4, text=test_outlines 
x_diff=2 , y_diff=5, text=test_outlines 
x_diff=5 , y_diff=0, text=Score: 0 
x_diff=5 , y_diff=1, text=Score: 0 
x_diff=5 , y_diff=2, text=Score: 0 
x_diff=4 , y_diff=3, text=Score: 0 
x_diff=3 , y_diff=4, text=Score: 0 
x_diff=2 , y_diff=5, text=Score: 0 
x_diff=6 , y_diff=0, text=Test Roboto Font 
x_diff=6 , y_diff=1, text=Test Roboto Font 
x_diff=6 , y_diff=2, text=Test Roboto Font 
x_diff=6 , y_diff=3, text=Test Roboto Font 
x_diff=6 , y_diff=4, text=Test Roboto Font 
x_diff=6 , y_diff=5, text=Test Roboto Font 
x_diff=6 , y_diff=6, text=Test Roboto Font 
x_diff=4 , y_diff=0, text=Girassol Font Test 
x_diff=4 , y_diff=1, text=Girassol Font Test 
x_diff=4 , y_diff=2, text=Girassol Font Test 
x_diff=3 , y_diff=3, text=Girassol Font Test 
x_diff=2 , y_diff=4, text=Girassol Font Test 
```

image

messengerbag:
Well, that picture at least shows pretty clearly what the problem is (or one of them, anyway): The horizontal displacement only draws the leftmost, rightmost and center (undisplaced) copies, rather than looping all the way from left to right.

ericoporto:
I tried removing the "ifs" and it didn't work too... Found a different algorithm, but think it's not applicable here: https://stackoverflow.com/questions/1109536/an-algorithm-for-inflating-deflating-offsetting-buffering-polygons

ivan-mogilko:
@ericoporto what are you trying or suggesting to do exactly?

I am planning to do speed test with `wouttextxy` vs current implementation or current implementation with Blits replaced with proper blending (did not yet test that it actually works). But I have not had time today and will probably try tomorrow.

ivan-mogilko:
@ericoporto only for the reference, here's the proper blending function I mentioned earlier. It's definitely going to be somewhat slower than the Blit one though.

Only changes compared to original are:
* included "gfx/blender.h"
* `set_blender_mode(nullptr, nullptr, _argb2argb_blender, 0, 0, 0, 0xff);` call before the drawing
* replaced all Blit with TransBlend.

<details><summary>CLICK ME</summary>
<p>

```C++
#include "gfx/blender.h"

void wouttextxy_AutoOutline_alphablend(Bitmap *ds, size_t font, int32_t color, const char *texx, int &xxp, int &yyp)
{
    int thickness = game.fonts.at(font).AutoOutlineThickness;
    auto style = game.fonts.at(font).AutoOutlineStyle;
    if (thickness <= 0)
        return;

    size_t const t_width = wgettextwidth(texx, font);
    size_t const t_height = wgettextheight(texx, font);
    Bitmap *outline_stencil =
        CreateTransparentBitmap(t_width, t_height + 2 * thickness, ds->GetColorDepth());
    Bitmap *texx_stencil =
        CreateTransparentBitmap(t_width, t_height, ds->GetColorDepth());
    if (!outline_stencil || !texx_stencil)
        return;
    wouttextxy(texx_stencil, 0, 0, font, color, texx);

    // move start of text so that the outline doesn't drop off the bitmap
    xxp += thickness;
    int const outline_y = yyp;
    yyp += thickness;

    set_blender_mode(nullptr, nullptr, _argb2argb_blender, 0, 0, 0, 0xff);

    int largest_y_diff_reached_so_far = -1;
    for (int x_diff = thickness; x_diff >= 0; x_diff--)
    {
        // Integer arithmetics: In the following, we use terms k*(k + 1) to account for rounding.
        //     (k + 0.5)^2 == k*k + 2*k*0.5 + 0.5^2 == k*k + k + 0.25 ==approx. k*(k + 1)
        int y_term_limit = thickness * (thickness + 1);
        if (FontInfo::kRounded == style)
            y_term_limit -= x_diff * x_diff;

        // extend the outline stencil to the top and bottom
        for (int y_diff = largest_y_diff_reached_so_far + 1;
            y_diff <= thickness && y_diff * y_diff <= y_term_limit;
            y_diff++)
        {
            outline_stencil->TransBlendBlt(texx_stencil, 0, thickness - y_diff);
            if (y_diff > 0)
                outline_stencil->TransBlendBlt(texx_stencil, 0, thickness + y_diff);
            largest_y_diff_reached_so_far = y_diff;
        }

        // stamp the outline stencil to the left and right of the text
        ds->TransBlendBlt(outline_stencil, xxp - x_diff, outline_y);
        if (x_diff > 0)
            ds->TransBlendBlt(outline_stencil, xxp + x_diff, outline_y);
    }

    delete texx_stencil;
    delete outline_stencil;
}
```
</p>
</details>

ericoporto:
Hey, xml version 26 is kept for some patching on 3.5.0.23+ ?

ivan-mogilko:
> Hey, xml version 26 is kept for some patching on 3.5.0.23+ ?

Yes, 26 is already taken by 3.5.0 patch. I will rebase this PR when patch is released and merged to master.

Again for the reference, here's probably the correct variant that only uses direct text output:

<details><summary>CLICK ME</summary>
<p>

```C++

void wouttextxy_AutoOutline_dumb(Bitmap *ds, size_t font, int32_t color, const char *texx, int &xxp, int &yyp)
{
    int thickness = game.fonts.at(font).AutoOutlineThickness;
    auto style = game.fonts.at(font).AutoOutlineStyle;
    if (thickness <= 0)
        return;

    size_t const t_width = wgettextwidth(texx, font);
    size_t const t_height = wgettextheight(texx, font);

    // move start of text so that the outline doesn't drop off the bitmap
    xxp += thickness;
    yyp += thickness;

    for (int x_diff = thickness; x_diff >= 0; x_diff--)
    {
        // Integer arithmetics: In the following, we use terms k*(k + 1) to account for rounding.
        //     (k + 0.5)^2 == k*k + 2*k*0.5 + 0.5^2 == k*k + k + 0.25 ==approx. k*(k + 1)
        int y_term_limit = thickness * (thickness + 1) + 1;
        if (FontInfo::kRounded == style)
            y_term_limit -= x_diff * x_diff;

        for (int y_diff = 0; y_diff <= thickness
            && y_diff * y_diff <= y_term_limit;
            y_diff++)
        {
            wouttextxy(ds, xxp - x_diff, yyp - y_diff, font, color, texx);
            wouttextxy(ds, xxp - x_diff, yyp + y_diff, font, color, texx);
            if (x_diff > 0)
            {
                wouttextxy(ds, xxp + x_diff, yyp - y_diff, font, color, texx);
                if (y_diff > 0)
                    wouttextxy(ds, xxp + x_diff, yyp + y_diff, font, color, texx);
            }
        }
    }
}
```
</p>
</details>

Compared to the Blit or TransBlend versions this gives more jagged edges, my guess that is because of how alpha composition works inside alfont renderer.

IMHO for visual test it's best to arrange pairs of lines with same font but different outline types.

ivan-mogilko:
I did  some speed tests in a separate branch: https://github.com/ivan-mogilko/ags-refactoring/commit/1616d4a0effc063151fc61604f4cf5b7ecdfb75b

Was drawing same text line with an outline 1000 times each, trying 5 different functions:
- stencils - original fernewelten's function using intermediate stencil bitmaps and simple blits (not suitable for anti-aliased TTFs)
- stencils alphablend - variant of above meant specially for antia-aliased TTFs in 32-bit games.
- singlestencil - variant of above but using only 1 stencil with text on it (does more drawing on destination bitmap, but not creates another intermediate bitmap).
- singlestencil alphablend - "singlestencil" but for AA TTFs.
- wouttextxy only (dumb) - no intermediate bitmaps, font renderer fills all the outline to the destination (works rather well for all fonts, but with TTFs it may produce jagged edges).

WFN (bitmap) font (roughly same height), auto outline thickness = 4

* stencils (1000 times): 979 (ms)
* stencils alphablend (1000 times): 937 (ms)
* singlestencil (1000 times): 2049 (ms)
* singlestencil alphablend (1000 times): 25050 (ms) (!!!)
* wouttextxy only (dumb) (1000 times): 3608 (ms)

TTF font, NO anti-aliasing, auto outline thickness = 4

* stencils (1000 times): 1215 (ms)
* stencils alphablend (1000 times): 1094 (ms)
* singlestencil (1000 times): 2180 (ms)
* singlestencil alphablend (1000 times): 32718 (ms)
* wouttextxy only (dumb) (1000 times): 9776 (ms)

Same TTF font, with anti-aliasing, auto outline thickness = 4

* stencils (1000 times): 1382 (ms)
* stencils alphablend (1000 times): 3390 (ms)
* singlestencil (1000 times): 2361 (ms)
* singlestencil alphablend (1000 times): 49916 (ms)
* wouttextxy only (dumb) (1000 times): 21785 (ms)

AlanDrake:
> One amazing thing from above is that alpha-blending is actually slightly faster (?!) 
Could it be compiler optimizations?
I remember my old software filters were much faster with bad code rathen than the version where I had tried to optimize, in the end  it was only thanks to compiler optimizations.

messengerbag:
Hmm, the measurements for the "singlestencil" method seem to have disappeared, but I'm a bit surprised by the values given initially. It seems to take a little more than twice as long, while by my estimate it should perform (4+1+4)^2=81 draw operations to create the outline, vs. only 2*(4+1+4)=18 for the "stencils" method: 4.5 times as many. I suppose writing to the temp bitmap might add some overhead, but not that much, surely?

ivan-mogilko:
Yes, it was actually weird, and I was suspecting doing something wrong there, but unfortunately was out of spare time to look again. Idk if it's worth now though, because original method seem to outperform everything else.

My only concern was (an is) that it continiously allocates temp bitmaps. Perhaps later we could consider a cache for these, assuming that many texts on screen may stay unchanged for a while and therefore have same sizes (such as menu texts, static gui labels etc).

ivan-mogilko:
@messengerbag Ah, thanks for noticing, I found that I used a simple Blit there instead of Blit with transparency.

New results look like following

<details><summary>CLICK ME</summary>

WFN

* stencils (1000 times): 882 (ms)
* stencils alphablend (1000 times): 911 (ms)
* singlestencil (1000 times): 15679 (ms)
* singlestencil alphablend (1000 times): 24351 (ms)
* wouttextxy only (dumb) (1000 times): 3602 (ms)

TTF font, NO anti-aliasing

* stencils (1000 times): 1181 (ms)
* stencils alphablend (1000 times): 1134 (ms)
* singlestencil (1000 times): 21261 (ms)
* singlestencil alphablend (1000 times): 31048 (ms)
* wouttextxy only (dumb) (1000 times): 9804 (ms)

Same TTF font, with anti-aliasing, auto outline thickness = 4

* stencils (1000 times): 1319 (ms)
* stencils alphablend (1000 times): 3385 (ms)
* singlestencil (1000 times): 22194 (ms)
* singlestencil alphablend (1000 times): 50117 (ms)
* wouttextxy only (dumb) (1000 times): 22157 (ms)

</details>

ivan-mogilko:
@messengerbag actually, that "singlestencil" function was stupid, I made an alternate one that removes different bitmap (https://github.com/ivan-mogilko/ags-refactoring/commit/7d5a052c51aac81d49c1cd4502d47d269d9d73d2). According to tests it's much closer to best results, but still not as fast:

<details><summary>CLICK ME</summary>

WFN

* singlestencil2 (1000 times): 900 (ms)
* singlestencil2 alphablend (1000 times): 1295 (ms)

TTF font, NO anti-aliasing

* singlestencil2 (1000 times): 1700 (ms)
* singlestencil2 alphablend (1000 times): 2169 (ms)

Same TTF font, with anti-aliasing

* singlestencil2 (1000 times): 3099 (ms)
* singlestencil2 alphablend (1000 times): 3873 (ms)

</details>

ericoporto:
Hey, I think the results for the Engine are good, could select one that is visually correct and is reasonably fast and commit. 

On the Editor I noticed only minor things, I will comment, but they may be dismissed.
- some inconsistency with identation on the file `Editor/AGS.Types/Font.cs`. 
- I think the content of the  `PropertyDescriptorCollection GetProperties(Attribute[] attributes)` call is there to hide properties depending of the condition? Maybe a comment there just to clarify this intent.

Actual functionality appears correct for me.

ivan-mogilko:
For 32-bit it seems the "alphablend" variant makes good results, both in terms of speed and looks. Speed is close to blit version in average for non-AA fonts, and it's the variant that draws correctly for AA-fonts, so it should be okay to use for all of them. AA outline is also created similar, if not identical, to the one drawn by rendering a font directly. So all good here.

For 16-bit variant atm I was not able to achieve "proper" looks for AA TTF fonts, because in 16-bit you cannot have alpha channel, and text has to be drawn directly on final surface to blend with the background colors, but that is not happening if we are using a transparent bitmap ("stencil") in between. So the text itself is anti-aliased, but outline is not.

Did not test 8-bit version yet. Anti-aliasing is disabled for 8-bit games, so only need to check non-AA font rendering.

ivan-mogilko:
> For 32-bit it seems the "alphablend" variant makes good results, both in terms of speed and looks. Speed is close to blit version in average for non-AA fonts

So, when trying to make new outlines work in 16-bit, I found that I did a stupid mistake earlier, and was passing wrong "color" value without alpha component. Upon fixing, "alphablend" function became noticeably slower.
I was planning on writing a universal function that switches between blending and blitting anyway because of 16- and 8-bit games, which I did (latest commits here: https://github.com/ivan-mogilko/ags-refactoring/commits/ags3--AFO-tests).
Now the blending is only done for anti-aliased TTF fonts (in 16->32 bit games), and masked blit is done in all other cases.

Latest speed test results:
[test outlines 8bit -- with universal funs.txt](https://github.com/adventuregamestudio/ags/files/4014187/test.outlines.8bit.--.with.universal.funs.txt)
[test outlines 16bit -- with universal funs.txt](https://github.com/adventuregamestudio/ags/files/4014188/test.outlines.16bit.--.with.universal.funs.txt)
[test outlines 32bit -- with universal funs.txt](https://github.com/adventuregamestudio/ags/files/4014189/test.outlines.32bit.--.with.universal.funs.txt)

Notes:
- single stencil variant is actually faster with WFN fonts, but not much faster and the relative difference decreases with the thickness.

ivan-mogilko:
Pushed universal function that should support anti-aliased outlines in 32-bit and 16-bit games, but also falls back to regular blits if font is not anti-aliased.

Editor/AGS.Types/Font.cs
```diff
@@ -120,7 +120,7 @@ public FontOutlineStyle OutlineStyle
             set { _outlineStyle = value; }
         }
 
-        [Description("Thickness of the automatic outline")]
+        [Description("Thickness of the automatic outline. WARNING: automatic outlines are generated at runtime and large sizes may negatively affect your game perfomance.")]
```
morganwillcock: There is a typo here as 'performance' is missing the 'r'.

PR #1028, opened 2020-01-03
Engine: allow to quit from max framerate mode (2)

ivan-mogilko:
Remake of #873.

* No longer store special "maxed" value in "frames_per_second" variable, because it's a game property that may be changed anytime by a script.
* Allow to toggle "maxed fps" mode on and off with the same key combo (Ctrl+E); note: as before, works only if you run game with `--fps`. When turning off should return to the latest fps value set by game script.
* Added enum defining fps display mode (no need to wonder what "display_fps == 2" means).

### DIFF
--- a/Engine/debug/debugger.h
+++ b/Engine/debug/debugger.h
@@ -46,8 +46,15 @@ void check_debug_keys();
 #define DBG_REGONLY   0x200
 #define DBG_NOVIDEO   0x400
 
+enum FPSDisplayMode
+{
+    kFPS_Hide = 0,    // hid by the script/user command
+    kFPS_Display = 1, // shown by the script/user command
+    kFPS_Forced = 2   // forced shown by the engine arg
+};
+
 extern float fps;
-extern int display_fps;
+extern FPSDisplayMode display_fps;
 extern int debug_flags;
 
 #endif // __AC_DEBUGGER_H
--- a/Engine/main/game_run.h
+++ b/Engine/main/game_run.h
@@ -35,7 +35,8 @@ void GameLoopUntilNoOverlay();
 void RunGameUntilAborted();
 // Update everything game related
 void UpdateGameOnce(bool checkControls = false, IDriverDependantBitmap *extraBitmap = nullptr, int extraX = 0, int extraY = 0);
-// Gets current wanted game FPS, this is usually a fixed number set in script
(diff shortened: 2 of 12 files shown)


Engine/debug/debugger.h
```diff
@@ -46,8 +46,15 @@ void check_debug_keys();
 #define DBG_REGONLY   0x200
 #define DBG_NOVIDEO   0x400
 
+enum FPSDisplayMode
+{
+    kFPS_Hide = 0,    // hid by the script/user command
+    kFPS_Display = 1, // shown by the script/user command
```
sonneveld: To avoid confusion, I suggest commenting out kFPS_Display until it is actually supported by a script command.

Engine/main/game_run.h
```diff
@@ -35,7 +35,7 @@ void GameLoopUntilNoOverlay();
 void RunGameUntilAborted();
 // Update everything game related
 void UpdateGameOnce(bool checkControls = false, IDriverDependantBitmap *extraBitmap = nullptr, int extraX = 0, int extraY = 0);
-
+// Gets current wanted game FPS, this is usually a fixed number set in script
```
sonneveld: If it's maxed, then it is the measured fps. (some animations were running slowly because it was assuming 1000fps I think)

rofl0r:
unfortunately, this doesn't seem to work anymore. using latest mater, once i hit CTRL-e the framerate will stay maxed and can't be returned.

ivan-mogilko:
> hmm, actually it works once the intro is complete. i tested abducted - 10 minutes, and while the intro plays you can hit CTRL-e as often as you like but it will stay maxed. once the normal game starts it can be toggled back and forth as expected.

Could you give a link to the game? Is it called simply "Abducted"?

rofl0r:
https://www.adventuregamestudio.co.uk/site/games/game/455-abducted-10-minutes-/

ivan-mogilko:
Ah, I think the issue is that the Ctrl+E check is done after some of the game-related checks, like dialog skipping, which consumes this keys. It should be moved out to the earlier stage.

rofl0r:
interesting... when you have unlimited framerate during the intro in that game, it somehow messes up the timer. you should have exactly 9:59 left to act when intro is over, but having the high framerate (300+ in native game resolution) causes the timer in the top right to jump up and down quite randomly looking...

Issue #480, opened 2018-07-16
API for drawing room masks


ivan-mogilko:
Ability to edit room masks at runtime was asked for number of times in the past, and [mentioned again recently](https://github.com/adventuregamestudio/ags/issues/473#issuecomment-405184815).

Code-wise this should not be too hard to implement. We already have DrawingSurface API, and same could be used to draw on room mask, including painting a ready sprite.

Thinking of potential problems, I guess most important is to provide a function for getting proper drawing color corresponding to a mask. And restrict drawing to these colors regardless of game's color depth.
Worst case we'd have to check for every pixel of a painted sprite to make sure it stays in the range (or maybe we should not bother? wrong pixels will simply get ignored by the engine anyway).

Also, question is whether these modifications should be written into savegame. Right now modification of the room background is not saved, and script must redo it every time player enters the room. Maybe same could be done for masks, in the sake of consistency.
(Then again, it is possible to have these changes in savegame too, as a bonus feature)

Probably this won't require any critical change to the code, and thus is possible to do in AGS 3 branch as well.

messengerbag:
Brilliant!

Yeah, a function to get a DrawingSurface for each mask would be sufficient, along with – as you say – a way to find the correct drawing color. I sort of assumed that it would simply be the AGS color number.

I don't think including it in the savegame (or even when you leave a room) is necessary, if it would complicate the task. It's easy enough to handle in script.

ivan-mogilko:
@messengerbag First experiment was successful.

https://github.com/ivan-mogilko/ags-refactoring/commits/dev--drawroommasks

Compiled test game: https://www.dropbox.com/s/tktylmrqri4a6ez/RoomMaskDrawing.zip?dl=0

ivan-mogilko:
Done, #1031

Issue #724, opened 2019-04-16
Redesign runtime debug overlay


ghost:
Engine is capable to display some debug overlay for game, such as showing walkable area masks and character walking paths. Unfortunately this is all done in a hard way, by overriding game update and render. During display game is unconditionally paused and debug information is drawn on black screen, which makes it inconvenient to compare with actual room background (not certain if this is not a bug though).

Anyway, my suggestion is to
1) redesign the debug overlay in a non-blocking way, with debug info layers being inserted at wanted order in the render stack as half-translucent sprites.
2) redesign debugging API to avoid using "Debug(a,b)" command.

after that it would be easier to add more debugging options.

**UPDATE**
p 1) may be done in either ags3 or ags4.
p 2) is changing script api, so meant for ags4.

selmiak:
sounds like a great addition.

also people tend to forget to turn off debug mode when releasing games. Maybe some subtle message when you start or quit the game to remind the dev that debug mode is still on would be helpful. But could also turn out to be annoying as there is one more click when testing over and over again...

PR #1031, opened 2020-01-11
Script API: get DrawingSurface for the room masks

ivan-mogilko:
For #480.

This is a minor addition, but may have numerous practical uses.
There's already a DrawingSurface interface, and room areas - walkable areas, hotspots, regions and walk-behinds - are represented as 8-bit bitmap masks, so it's trivial to connect these.

The 4 new script functions allow to get corresponding room mask as a DrawingSurface, and do with it whatever a game developer wants.
Surfaces will be 8-bit, and each color index corresponds to an area ID. Color index = 0 is "no area", or eraser.
There are necessary safety tests in the engine to make sure that indexes beyond supported area limit are simply ignored (treated as "no area").

For example, following script will expand walkable area 1:
<pre>
DrawingSurface* ds = GetDrawingSurfaceForWalkableArea();
ds.DrawingColor = 1; // drawing color = area ID
ds.DrawRectangle(100, 100, 200, 200);
ds.Release();
</pre>

Changes made this way **are not saved between room visit**, which is in line with drawing upon a room background (this was done to prevent save data bloat). If user wants these changes to persist, the usual approach is to save custom data, and use it as an instruction, performing drawing on room load event.

### DIFF
--- a/Common/game/roomstruct.cpp
+++ b/Common/game/roomstruct.cpp
@@ -205,6 +205,31 @@ void RoomStruct::SetResolution(RoomResolutionType type)
     _resolution = type;
 }
 
+Bitmap *RoomStruct::GetMask(RoomAreaMask mask) const
+{
+    switch (mask)
+    {
+    case kRoomAreaHotspot: return HotspotMask.get();
+    case kRoomAreaWalkBehind: return WalkBehindMask.get();
+    case kRoomAreaWalkable: return WalkAreaMask.get();
+    case kRoomAreaRegion: return RegionMask.get();
+    }
+    return nullptr;
+}
+
+float RoomStruct::GetMaskScale(RoomAreaMask mask) const
+{
+    switch (mask)
+    {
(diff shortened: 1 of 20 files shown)


ericoporto:
OK, I had some fun with this. Click the gif below for full video.
[clicking_rts](https://streamable.com/fqqjx)

AGS Project: [maskDrawing.zip](https://github.com/adventuregamestudio/ags/files/4074728/maskDrawing.zip)

I think it works, the only thing I wanted more when drawing was being able to see the other masks like we have the shortcut with Ctrl+A for walkbehinds, which I had to improvise with an overlay, but other than this, it appears to work.

On the video above, the houses are drawn in the background and the walkbehinds are updated, along with the walkable areas. The army of rogers will keep walking randomly, using pathfinder to avoid the non-walkable areas.

ivan-mogilko:
> , the only thing I wanted more when drawing was being able to see the other masks like we have the shortcut with Ctrl+A for walkbehinds

There's a related ticket: #724

PR #1032, opened 2020-01-13
iOS Scale Fix

jonreyno:
These changes were necessary to get games to scale properly on iOS with minimal artifacts in 3.5. One of the files also received some formatting updates to conform to AGS standards. I'll be updating the other iOS files in other commits, but wanted to get this fix in sooner since there was already someone asking about it.


ivan-mogilko:
Well, I don't think there's anyone around who would be able to review this in depth, so I will trust it works for you (and most of changes are on iOS part)

My only big request is though, could you split this into separate commits with 1 fix per commit?
I am asking to keep only related changes in commit, because sometimes we need to compare code behaviors in different versions, and it may be inconvenient if one won't be able to tell quickly which changes are actually relevant to which fix. (And especially if we want to revert particular change)

And, please remove the seemingly unnecessary "test" commit that changes the json file.

ivan-mogilko:
Well, I don't think there's anyone around who would be able to review this in depth, so I will trust it works for you (and most of changes are on iOS part too).

My only big request for the future, if you'd want to make more patches: please split unrelated changes into separate commits, with ideally only thematically linked fixed per commit. For instance, it's better to have the code style changes as a separate commit.

I have to ask this because sometimes we need to compare code behaviors in different versions, and it may be inconvenient if one won't be able to tell quickly which changes are actually relevant to which fix (And especially if we want to revert only particular change).

jonreyno:
Noted. I’ll do this in the future. Thanks!

Issue #954, opened 2019-10-15
Object.Scaling available from Script


ericoporto:
Currently it's not possible to manually scale an object from Script. I would like it was possible similar to how it's possible for Characters, through Object.Scaling. This may require and additional property to enable disable manual scaling, like Character.ManualScaling, but instead Object.ManualScaling.

ericoporto:
I don't know if there's a reason why scaling is limited to 5 to 200 (is stored as a short for instance), but if there isn't, I would really like it to be able to go bigger than 200. A 5 to 1000 limit would be good enough for my usage.

ericoporto:
Just for register, the probable reason for the 5 to 200 limit for character scaling is due to the walkspeed linked to scaling on : https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/character.cpp#L2038

ivan-mogilko:
Done #1033

PR #1033, opened 2020-01-15
Script API: implemented Object.ManualScaling and Object.Scaling

ivan-mogilko:
For #954, and for consistency with Character class, this implements Object.ManualScaling and Object.Scaling, and deprecates Object.IgnoreRoomScaling.

No internal data was changed (room object already has got a suitable variable for storing current scale).

Unlike current Character's scaling, Object's don't have an arbitrary limit and instead limits value to 1 -> INT16_MAX percents range ((object's scale variable is int16), or, speaking multipliers, - from x0.01 to x327.67;

For backward compatibility, using old IgnoreRoomScaling property will reset scaling to 100% (also, for consistency with character).

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2173,8 +2173,10 @@ builtin managed struct Object {
   import attribute int  Graphic;
   /// Gets the object's ID number.
   readonly import attribute int ID;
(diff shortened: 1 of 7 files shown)


ericoporto:
[object_scaling](https://streamable.com/w70gq)

AGS Project: [ObjectScaling.zip](https://github.com/adventuregamestudio/ags/files/4074136/ObjectScaling.zip)

I did some tests using the Scaling and it works.

PR #1036, opened 2020-01-21
Tab switching performance improvements

AlanDrake:
On windows 10 the performances of switching between script tabs was horrible compared to windows 7 due to the function list updates to the combobox
I've wrapped the functions combobox changes between BeginUpdate-EndUpdate which halved the impact, then I also made so that only the currently focused script pane performs the update, because the update is triggered at any change of focus anyway so there's little point to update every single combobox from every open script.

~~To improve further I'm also updating the DockPanelSuite to the latest 3.0.6 version which brings back the performances I see on a W7 VM.~~

~~There's just one problem, the newer DockPanelSuite versions use a different api for theming so the theme loading is not working correctly and causes a NullPointerException at some point with the existing VisualStudioDark theme.~~ Moved to another branch

### DIFF
--- a/Editor/AGS.Editor/ColorThemes/Controls/ComboBoxCustom.cs
+++ b/Editor/AGS.Editor/ColorThemes/Controls/ComboBoxCustom.cs
@@ -25,10 +25,12 @@ namespace AGS.Editor
             Size = original.Size;
             TabIndex = original.TabIndex;
 
+            BeginUpdate();
             foreach (var i in original.Items)
             {
                 Items.Add(i);
             }
+            EndUpdate();
         }
 
         protected override void OnCreateControl()
--- a/Editor/AGS.Editor/Panes/ScriptEditor.cs
+++ b/Editor/AGS.Editor/Panes/ScriptEditor.cs
@@ -260,6 +260,7 @@ namespace AGS.Editor
 
         private void UpdateFunctionList()
         {
+            if (!this.ContainsFocus) return; // only update for the active pane to avoid expensive combo Add operations
             List<string> functions = new List<string>();
             foreach (ScriptFunction func in _script.AutoCompleteData.Functions)
             {
@@ -278,13 +279,12 @@ namespace AGS.Editor
                     }
                 }
             }
+            cmbFunctions.BeginUpdate();
             cmbFunctions.Items.Clear();
             cmbFunctions.Items.Add("(general definitions)");
             functions.Sort();
-            foreach (string func in functions)
-            {
-                cmbFunctions.Items.Add(func);
-            }
+            cmbFunctions.Items.AddRange(functions.ToArray());
+            cmbFunctions.EndUpdate();
             SelectFunctionInListForCurrentPosition();
         }
 


ivan-mogilko:
I'd strongly suggest to separate updating major version of a library from a speed fix. Even if it helps in its regard, this may be a significant change, especially since you already noticed that it breaks skin compatibility. This way we may test and fix both changes separately (and maybe merge them separately if one is not yet stable).

AlanDrake:
Okay, I'll leave this one with just the changes on the update logic and split the DockPanel upgrade in another branch

persn:
Is the performance acceptable with the current changes? I notice the code adds the functions to the list and sorts them after. Would it be more efficient to sort while adding using a `SortedList` or `SortedSet`?

AlanDrake:
According to the profiler what really slows things down on W10 is the Add/AddRange.
Combining this fix and the updated DockPanelSuite makes it almost as fast as I see on the W7 VM.

persn:
You say almost as fast so I assume it's desirable to squeeze more out of this. How about running `UpdateFunctionList()` async?

AlanDrake:
I had tried, both async and threading, they didn't seem to matter before.
Perhaps it's different now, but it's fast enough that you can only see the slowness by keeping pressed ctrl+tab. (Instant fast on W7, slower on W10 but acceptable)

persn:
Is the slowness just visible? It doesn't make the editor stutter or hang or anything like that?

AlanDrake:
This is what happens on my machine without fixes:
https://www.dropbox.com/s/ddbhuub0bjnd7u5/2020-01-23%2021-59-34.mkv?dl=0

It hangs until it finishes the `cmbFunctions.Items.Add` operations.
Upgrading DockPanelSuite makes it a little faster, while this fix makes so only the current pane's combobox is updated (with Begin/EndUpdate which halve the time spent) so the slowness does not pile up the more open script tabs you have.

persn:
Looks pretty nasty. In theory the performance should scale with the number of functions and the size of the file so it could possibly hang again when switching to a big file. But if you vouch for the performance in its present state I'll trust you 👍

AlanDrake:
Well, look at that... there was also a malicious ComboBoxCustom doing bare Items.Add in a foreach from the original combobox items.
I'll need to prepare a new commit.

ivan-mogilko:
> The custom combobox (only used for themes) already inherits the list from the original element, there's no need to add anything, everything is updated correctly.

Are you sure about this? I don't see where it inherits the original element. I see it inherits the type, but does it really inherit the data?

AlanDrake:
I tried it and the list updates as the original.

persn:
I find that a little hard to swallow. Like @ivan-mogilko says we can't see no code that updates the data of the new combobox, it has to inherit the data from the original somehow. Unfortunately I won't be able to do any testing since I'm at a seminar for the weekend. I implore you to triple test and verify that all the items of the original ComboBox is inherited correctly. If it really did then logically speaking the old code should contain duplicates of every item in the combo boxes where this  is used

AlanDrake:
Hmmm, maybe I get it.
Since the props panel and script comboboxes are populated later it doesn't matter.
You're right, I should to put the Add back, I thought there was some clever casting somewhere but no.

AlanDrake:
Done. I just wrapped the Add between BeginUpdate, EndUpdate for ComboBoxCustom, which should be the best practice.

ivan-mogilko:
@AlanDrake you could also use AddRange in the second commit as well?

AlanDrake:
@ivan-mogilko I would, but I couldn't find a way to turn the "Items" back to something palatable to AddRange().
I could try replacing with a Insert(0, xxx), it didn't throw errors that time I tried but without static comboboxes to test I don't know if it's actually working.
Therefore I just wrapped Add() withing Beghin/End, which already boosts performance.

AlanDrake:
Anyway, just a note.
If we ever use DataSource on comboboxes, we'll probably need to add some checks or it will throw errors,

morganwillcock:
> I would, but I couldn't find a way to turn the "Items" back to something palatable to AddRange().

You can do it inline using LINQ and the call `ToArray()` on the result of that, I believe.

PR #1037, opened 2020-01-21
Editor: Updated DockPanelSuite to 3.0.6

AlanDrake:
Better performances for Windows 10 and the tabs are no longer inserted at wrong positions.


AlanDrake:
This is it.
I was expecting more work to get it up and working but it looks like sometimes good things do happen.

ericoporto:
I tested quickly on Wine moving the panels around and scripting, and it seems to work fine.

One thing I noticed is in the About window, the copyright for the DockPanel still has 2007 date.

morganwillcock:
> One thing I noticed is in the About window, the copyright for the DockPanel still has 2007 date.

There is also a licence file that gets included for distribution, if the date or licence need to change.

Windows/Installer/Source/Licenses/Weifen Luo DockPanel Suite.txt

AlanDrake:
Looks like the license and date are still the same.
But they moved the "MIT License" line and used wrap-around lines instead of limiting width.
I'm gonna update it just in case it's a clever loophole to open for lawsuits.

PR #1038, opened 2020-01-22
Scintilla: Hotfix for cursor flickering and cpu usage on Windows 10

AlanDrake:
Turns out that my suggested fix was more legit than I thought.
Newer Scintilla.NET versions leave the `base.WndProc(ref m);`, which is causing the cursor flickering and cpu usage spikes on Windows 10, as the very last case to call.
Therefore this hotfix should be good enough until the day we get to upgrade Scintilla to the latest version.

### DIFF
--- a/Editor/AGS.Controls/ScintillaNET/ScintillaControl.cs
+++ b/Editor/AGS.Controls/ScintillaNET/ScintillaControl.cs
@@ -657,6 +657,12 @@ namespace Scintilla
                     return;
                 }
             }
+            // [AVD] moving this above WM_NOTIFY fixes cursor flickering and cpu usage on W10
+            else if (m.Msg == WinAPI.WM_SETCURSOR)
+            {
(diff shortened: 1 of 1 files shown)


AlanDrake:
> 
> 
> @AlanDrake I think you could commit a wrong file?

Yes, sorry.

AlanDrake:
> 
> 
> This is what I did 5 years ago so it should be good. 👍 #293

Fixing what we had already fixed... how did we not notice up till now?

persn:
> Fixing what we had already fixed... how did we not notice up till now?

Well the original fixer wasn't paying attention so.. It was just random that I went into the discord channel today and saw the issue being discussed

persn:
Related forum thread https://www.adventuregamestudio.co.uk/forums/index.php?topic=52805.0

ivan-mogilko:
Oh, I see now, this somehow got reverted here: 487cd9c7a619213ae0a9c20042718a46b18e5496

Issue #1035, opened 2020-01-17
Editor Testing Frameworks


persn:
I think we should add projects that allow writing tests for the editor.

- [x] Decide on a test framework
- [x] Decide on a mock framework
- [x] Add test projects for the editor projects that use the previously mentioned frameworks
- [x] Add some few tests to begin with (We can use these tests to discuss guidelines for writing tests if we want)
- [x] Add test runs to the CI

persn:
I propose [NUnit](https://nunit.org/) and [NSubstitute](https://nsubstitute.github.io/), simply because these are the tools I use, and I know they are very good. If no one makes objections in a reasonable amount of time (a few days?) I'll make a PR for it

rofl0r:
> I think we should add projects that allow writing tests for the editor.

what kind of tests ?

persn:
Unit Tests to begin with

ivan-mogilko:
I think the question is rather, what functionality do you want to test.
This is important, because there's an interest in changing the editor workflow, such as the way it works with the source files, and decoupling build functions into standalone tools.

persn:
The idea for starting out is just to make the infrastructure for testing available, I don't have plans to write 90% plus test coverage or even 5%. This is important because I've fixed bugs in the past where having tests could have made my workflow smoother instead of manually testing every case for every change for every feedback. So no specific functionality is what I'm trying to say I guess

My motivation for having test frameworks available is because I'm researching if I'm capable of solving the issue of "All the room management should be reimplemented in .NET." as stated in #469, and I think it might help to have tests that checks that room handling works as it should before I start messing with it.

rofl0r:
> what kind of tests ?

> I think the question is rather, what functionality do you want to test.

right. i'm asking because "writing tests for the editor" actually sounds as if it would require some click automation software like autoit in order to assert that clicking certain stuff in GUI produces the desired result.

morganwillcock:
I've twice started something similar, using XUnit. I think it would be OK to add it tests for specific parts which are being modified (and write new code to be testable), but I'm not sure how much of the existing C# code will exist long term (so aiming for total test coverage could end up being a waste) as there isn't a long term plan at the moment.

It may be beneficial to continually track the latest Visual Studio project version and increase the .NET Framework version to the highest compatible level, if builds are then depending on tests that run in .NET. I *think* that the XUnit console runner needed 4.5.2, otherwise it forces the environment to have a mass of extra Nuget packages.

[...]

PR #1040, opened 2020-01-26
Added testing framework to the AGS Editor

persn:
Reference #1035 

In short I
- Added a project for writing tests to the editor. With NUnit and NSubstitute
- I refactored some code to not make use of Singletons because any impure static call makes testing complicated
- Wrote a few tests

My goal here is to make the testing framework available so people can start using it when they write new code, it's not my plan to start adding coverage to already existing code

What I haven't done is added test execution to the CI system, since I'm not familiar with Cirrus CI and have only tried Dockerfiles a little bit, maybe someone else could do that. The easiest way of running the tests from the command line is by using the dotnet CLI, I used the following command `dotnet test AGS.Editor.NoNative.sln --no-build`. The caveat with using dotnet CLI I think is that it works poorly with the AGS.Native project, but I'm assuming that most people will not be writing tests directly to that project

I used the new csproj format for creating a new project, because that's what VS2019 do, I may need someone to test and verify that it opens fine in VS2015, if not I'll have to change the format

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -178,6 +178,16 @@ build_editor_task:
     cd Solutions &&
     cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\8.0\Lib\Win8\um\x86;!LIB!" &&
     msbuild AGS.Editor.Full.sln /p:PlatformToolset=v140 /p:Configuration=%BUILD_CONFIG% /p:Platform="Mixed Platforms" /maxcpucount /nologo"
+  test_script: |
+    curl -fLOJ https://github.com/nunit/nunit-transforms/raw/master/nunit3-junit/nunit3-junit.xslt
+    pushd Solutions\packages\NUnit.ConsoleRunner.3.*
+    set RUNNER="%CD%\tools\nunit3-console.exe"
+    popd
+    cmd /v:on /c "!RUNNER! Editor\AGS.Editor.Tests\bin\%BUILD_CONFIG%\AGS.Editor.Tests.dll --x86 --result=TestResult-junit.xml;transform=nunit3-junit.xslt"
+  test_result_artifacts:
+    path: TestResult-junit.xml
+    type: text/xml
+    format: junit
   ags_editor_pdb_artifacts:
     path: Solutions/.build/*/*.pdb
   ags_native_pdb_artifacts:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -184,10 +184,11 @@ build_editor_task:
     set RUNNER="%CD%\tools\nunit3-console.exe"
     popd
     cmd /v:on /c "!RUNNER! Editor\AGS.Editor.Tests\bin\%BUILD_CONFIG%\AGS.Editor.Tests.dll --x86 --result=TestResult-junit.xml;transform=nunit3-junit.xslt"
-  test_result_artifacts:
-    path: TestResult-junit.xml
-    type: text/xml
-    format: junit
+  always:
+    test_result_artifacts:
+      path: TestResult-junit.xml
+      type: text/xml
+      format: junit
   ags_editor_pdb_artifacts:
     path: Solutions/.build/*/*.pdb
   ags_native_pdb_artifacts:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -165,7 +165,9 @@ build_editor_task:
     AGS_DIRECTX_LIB: C:\Lib\DirectX
   nuget_packages_cache:
     folder: Solutions\packages
-    fingerprint_script: type Editor\AGS.Editor\packages.config
+    fingerprint_script:
+      - type Editor\AGS.Editor\packages.config
+      - type Editor\AGS.Editor.Tests\packages.config
     populate_script: nuget restore Solutions\AGS.Editor.Full.sln
   build_script: >
     "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -183,7 +183,7 @@ build_editor_task:
     pushd Solutions\packages\NUnit.ConsoleRunner.3.*
     set RUNNER="%CD%\tools\nunit3-console.exe"
     popd
-    cmd /v:on /c "!RUNNER! Editor\AGS.Editor.Tests\bin\%BUILD_CONFIG%\AGS.Editor.Tests.dll --x86 --result=TestResult-junit.xml;transform=nunit3-junit.xslt"
+    cmd /v:on /c "!RUNNER! Editor\AGS.Editor.Tests\bin\%BUILD_CONFIG%\AGS.Editor.Tests.dll --result=TestResult-junit.xml;transform=nunit3-junit.xslt"
   always:
     test_result_artifacts:
       path: TestResult-junit.xml
--- a/Editor/AGS.Editor.Tests/AGS.Editor.Tests.csproj
+++ b/Editor/AGS.Editor.Tests/AGS.Editor.Tests.csproj
@@ -1,9 +1,8 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\Solutions\packages\NUnit3TestAdapter.3.16.1\build\net35\NUnit3TestAdapter.props" Condition="Exists('..\..\Solutions\packages\NUnit3TestAdapter.3.16.1\build\net35\NUnit3TestAdapter.props')" />
-  <Import Project="..\..\Solutions\packages\JunitXml.TestLogger.2.1.15\build\net45\JUnitXml.TestLogger.props" Condition="Exists('..\..\Solutions\packages\JunitXml.TestLogger.2.1.15\build\net45\JUnitXml.TestLogger.props')" />
   <Import Project="..\..\Solutions\packages\Microsoft.NET.Test.Sdk.16.4.0\build\net40\Microsoft.NET.Test.Sdk.props" Condition="Exists('..\..\Solutions\packages\Microsoft.NET.Test.Sdk.16.4.0\build\net40\Microsoft.NET.Test.Sdk.props')" />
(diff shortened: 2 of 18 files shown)


persn:
I just uploaded it and I found a trouble with the test execution, I'll look at it later in the week so this is not ready to review yet

persn:
The CI builds definitely did not like the new csproj format so I guess I'll have to rewrite unless we fancy upgrading our VS versions

morganwillcock:
New error is:
> The missing file is ..\..\Solutions\packages\NUnit.3.12.0\build\NUnit.props

...maybe it is this?
https://stackoverflow.com/questions/52582330/this-project-references-nuget-packages-that-are-missing-on-this-computer-the

persn:
No, the problem is that the new projects packages is not being restored my guess is. If they were then that file should have existed. I'm guessing what happens is that nuget packages are cached to cirrus using the Editor/AGS.Editor/packages.config as a hash, which is not going to work because we now have multiple packages.config. Since no change is detected in the original file the command is skipped and the cached packages are restored.

Someone who knows Cirrus better will have to verify my theory, and maybe suggest a fix

morganwillcock:
I don't think `type` supports wildcards, but you can just call it twice:
```
type Editor\AGS.Editor\packages.config && type Editor\AGS.Editor.Tests\packages.config 
```

persn:
I wanted to try the wildcard first because if it works it looks nicer

morganwillcock:
Do you know if it is possible to get the test results in JUnit format? I think this is the only format that the CI environment can process (to display as test results in GitHub).

It looks like there is a plugin to convert the format, but I don't know if this comes from Nuget or NUnit has to request this somehow:

https://github.com/nunit/nunit-transforms/tree/master/nunit3-junit

persn:
JUnit of all things, not exactly what comes to mind when I think of .NET compatible tech. I'll have to do some research and get back to you. In the past I've used coverlet, but I can't see they support JUnit

persn:
@morganwillcock check my latest commit

morganwillcock:
Thanks. I think I will need to build the container locally to check what the environment settings should be, hopefully I can test it out over the next few days.

persn:
I fixed every issue, now we just have to integrate it into the CI

rofl0r:
> I refactored some code to not make use of Singletons because any impure static call makes testing complicated

in fact, this seems to be the biggest part of this PR. there's a lot of changes around color theme handling, with the apparent intention of adding support for more than one theme, shifting use of singletons from one place to the other.

if using this testing framework, that's introduced for no specific purpose other than "it's cool to have it" requires that many codechanges in existing classes in order to make testing "less complicated", maybe this all should be postponed/revised at a date when somebody actually has a specific use case.

ericoporto:
@AlanDrake I think the changes here are relevant to you.

AlanDrake:
Does ColorThemes really need an interface with just one implementation?
Make sure a broken json won't nag the user, especially when it's not the active theme.
I hope loading all the themes doesn't instance multiple ComboBoxCustom simultaneously, that class was just a hack to fully recolor the control, there should be just one at all times, I think.

I don't know yet for sure what I'm going to do with the themes, but I'll probably have to store the DockPanel Theme object and ColorPalette when I get around to implementing things properly for the new VS2015 style theme.

persn:
> in fact, this seems to be the biggest part of this PR. there's a lot of changes around color theme handling, with the apparent intention of adding support for more than one theme, shifting use of singletons from one place to the other.

Actually the biggest part of the PR is extracting interfaces, the other changes are minor, shifting use of singletons to constructor inject said interfaces. The interfaces are also needed to write mocks.

> if using this testing framework, that's introduced for no specific purpose other than "it's cool to have it" requires that many codechanges in existing classes in order to make testing "less complicated

It has nothing to do about being cool, it has to do with facilitating us to write better quality code, fix bugs, prevent regressions, give us safety that re-write of old code won't break anything. Writing tests can give us all of those things and more.

> maybe this all should be postponed/revised at a date when somebody actually has a specific use case.

Postponing introducing tests I'm not going to accept easily. That's just another way of saying we're not going to do this at all, what would a specific use case for introduce tests even be? You don't wait to introduce tests, you just introduce it so people can start writing them. Just going back to last week I can think of a couple of examples where writing tests could have prevented some bad situations, but it wasn't most probably because the framework to do so wasn't available

1. In #1036 @AlanDrake tried to remove code because he misunderstood how some of the code worked, his assumption was that the copy contructor of the custom themed combo box copied over the items twice, he even stated that he verified his "fix" by manual testing, but he was mistaken. A test would have proved for himself and others whether or not that was truly the case. Luckily the error was avoided in a peer review.

2. In #1038 we fixed an issue that was actually fixed 4-5 years ago, but due to a wrong git merge it was undone and no one noticed. If a test was in place they would have caught this regression.

persn:
> Does ColorThemes really need an interface with just one implementation?

Extracting interfaces in this manner is a necessity for writing mocks/fakes, mostly due to the nature of C# and it being type strong and impure The mocking framework needs to able to control a type directly https://nsubstitute.github.io/help/getting-started/ This wouldn't be a problem in JavaScript for instance where this behavior is built into the language, or Haskell where a function always returns the same result no matter what you put into it. But since this is C# we need to live with some caveats.
  Alternatively I could undo the interface extraction of `ColorThemes` and do it on `ColorTheme` instead, it would probably make more sense, although if we ever wanted to write unit tests on a component where `ColorThemes` is used we'd have the dilemma again. It may not be a problem we ever have to address though.. What do you think?

> Make sure a broken json won't nag the user, especially when it's not the active theme.

I was kind of hoping to avoid adding large amounts of test coverage here, the tests so far are just some I wrote to demo the frameworks, what it would take to implement them, and I also I wrote some tests just to debug a small bug I encountered

> I hope loading all the themes doesn't instance multiple ComboBoxCustom simultaneously, that class was just a hack to fully recolor the control, there should be just one at all times, I think.

The themes have factory functions for generating a `ComboBoxCustom`, but the GUI components themselves are in control of the creation of UI elements, this is done by intercepting and replacing the original `ComboBox` in each GUI components constructor so it shouldn't be any multiple anythings. I wish there was a nicer way to make that stuff but WinForms is hard to work with for coloring.

morganwillcock:
One issue with these new files, they aren't using Windows line endings. Whether that is correct or not is arguable, but the majority of the rest of the source is using Windows line endings so probably best to keep it all consistent.

But also, `dotnet` seems to be a .NET Core command so that isn't in the build environment (and I wouldn't want to mandate that someone must install it on their own computer just to run the tests). So I'll first try to run it through the console runner instead, unless anyone knows of an alternative that is better.

persn:
> One issue with these new files, they aren't using Windows line endings. Whether that is correct or not is arguable, but the majority of the rest of the source is using Windows line endings so probably best to keep it all consistent.

This hasn't been an issue in the past, I think even @ivan-mogilko is commting linux style endings, I seem to remember him saying that sometime. Is this really important? I can try to fix it if we think so but I'm a little skeptical that this is worth any time.

> But also, `dotnet seems to be a .NET Core command so that isn't in the build environment

It can be installed pretty easily

```
ARG DOTNETCORESDK_VERSION=3.1.101
RUN cinst dotnetcore-sdk --version %DOTNETCORESDK_VERSION% -y
```

> (and I wouldn't want to mandate that someone must install it on their own computer just to run the tests). So I'll first try to run it through the console runner instead, unless anyone knows of an alternative that is better.

I've been using Visual Studio, is it fair to assume everyone else writing editor code will be too?

ericoporto:
Line endings can be fixed on git by using `--whitespace=fix` on [rebase](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---whitespaceltoptiongt). It follows what's [configured on your repo](https://git-scm.com/docs/git-apply#Documentation/git-apply.txt---whitespaceltactiongt). I think it will pick the line ending from there and things will work.

persn:
If we are enforcing a specific rule on newlines, shouldn't they be set in `.gitattributes` to avoid mishaps? Otherwise we're just begging for this issue to happen and again for any contributor

ivan-mogilko:
I was commiting with Unix endings from the beginning of this project (the settings on my client side are: "checkout Win / commit Unix")

morganwillcock:
Right, well that's the opposite to how the source was imported.
`git show d6516494837a7ac8ed2bb0d9c862545376e4c4c3`
But OK, Unix line endings it is.

Regarding installing .NET Core, at the moment it is pretty loose as to the build tools and .NET versions that are needed to build everything, so I don't want to jump into installing a completely different SDK that might be versioned differently if the tests can run with what is already there. Personally speaking, I use the same computer for work, so I have to keep the additional dependencies to a minimum. For CI, the container execution starts in an environment that doesn't have any of the required variables set, so it isn't trivial to just install something and then build with it.

persn:
Okay I'll look into using the console runner instead

ivan-mogilko:
> Right, well that's the opposite to how the source was imported.
> `git show d6516494837a7ac8ed2bb0d9c862545376e4c4c3`

That was probably before things were set up.

morganwillcock:
> Okay I'll look into using the console runner instead

I am testing it. It looks like it might need a different (standalone) transform file to convert the output to JUnit format, but the tests are running and passing.

persn:
> If you want to work on it fine, but I can revert my commit of importing the nuget package with the filter because I think it only works for the dotnet CLI

persn:
@morganwillcock Thank you

Ready for review

morganwillcock:
Ah sorry, I think the block to upload the result needs to be modified to always run.
https://github.com/morganwillcock/ags/commit/7a6883746bfe05e0685690b337235768dbcdc16c

I'll try to look more carefully at the interface changes over the weekend.

persn:
Whenever you're ready

morganwillcock:
So, along with change to always upload the result, I think the architecture of the test libraries also needs to be changed to match the configuration of the Editor. Currently it is flagged as a mismatch, and this results in the 64 bit test runner failing, which is why it is currently forced to run in 32 bit mode.

So I think this:
https://github.com/morganwillcock/ags/commit/2d4c726d181a3f1f763be751bc81090253cef80e
...is the correct thing to do.

For the actual code changes, it seems that testing interfaces is pretty standard and also a good way to actually avoid code changes just for testing. I'm not an expert in it, but I can't see anything very objectionable here. When I've written similar tests before I've typically used `InternalsVisibleToAttribute` instead of modifying class access; `ComboBoxCustom` is switched to public but I'm not aware of any reasoning to decide what should be public in the Editor's assembly (nor of anything that would look to load it and use it). So with the two changes linked above, I don't have any objections to merging this.

persn:
> `ComboBoxCustom` is switched to public

I'm pretty sure I didn't have a good reason for making it `internal` in the first place

persn:
I'll cherry pick your commit when I get home from the gym

persn:
Fixes #1040

PR #1046, opened 2020-02-13
adds mouse pan with click+Ctrl on Room panel

ericoporto:
for #613

Mouse middle click is currently busy with copy coordinates functionality. Space was also taken by rolling the zoom controls of the room. Here I am adding Ctrl+Mouse Left Click (and hold) to pan. The ctrl doesn't have to be kept pressed.


AlanDrake:
We could invert middle-click with ctr-click for the coordinates, I keep middle clicking to scroll around like an idiot.

ivan-mogilko:
@ericoporto could you fix the tiny indentation difference on the opening "if" lines?

ericoporto:
Yes that could work.

Other interaction I had in mind (but I couldn't figure out how to set the cursors) was turning the cursor to have open hand to be in "pan mode" and then regular clicking turns the cursor into "closed hand" while the click is hold and you can drag-pan like this. Releasing click makes the cursor pan mode again.

@ivan-mogilko I can fix identation, yes. I think I have something wrong, can someone tell me how many spaces a tab has? I wrote the code directly in GitHub.

ericoporto:
I think I fixed the identation.

ericoporto:
@AlanDrake I switched the shortcut so now either ctrl+shift+left click pans or simply **mouse middle click hold**. For coordinates any modifier key you use with middle click will bring the _copy coordinates to clipboard_ context menu. This is an uglier set of keys but I feel works better.

ericoporto:
Without Composite
https://streamable.com/2rlef

With Composite
https://streamable.com/r0cn9

ericoporto:
I think I am done now. Ideally I would add one thing, which would be zoom on mouse position instead of just zoom, so that the scroll was correct. I even found an answer on SO on this: https://stackoverflow.com/questions/10694397/how-to-zoom-in-using-mouse-position-on-that-image .

But I couldn't make it work correctly with the existing `_state` .

ivan-mogilko:
This seem to work, but I have few notes on how the code is organized.

First of all, if you look at how controls were handled beforehand in the room editor, there was a certain order: first it passed an event into the active layer, and layer told whether it handled current action or not, if not, then the main editor class itself handles it.
Similarily, the cursor pic was set in a dedicated function SelectCursor.

I propose to keep this order for consistency. If it seems inconvenient for some reason, then I'd suggest to explicitly define the new rule and create an extra input-handling function that would be called before layers (and similarily, check whether it handled the input). That would make handling code more clear.

Secondly, less important, I notice that there are _menuClickX, Y variables that store last click coordinates, and this PR adds _panGrabPoint variable. Would it be a good idea to merge these in one point (their meaning is somewhat similar)?

ericoporto:
the _menuClickX, Y are for the copy coordinates menu, not last click coordinates at all times. I can merge them, but I don't like the idea of something being repurposed for two different usages, makes changing each of them harder. `_menuClickX, Y` could just be a point though, instead of two variables. And _panGrabPoint is for latest click or last frame with mouse hold, I can add a comment to make this obvious.

The order/extra input-handling function I am taking a look; I am leaning at adding extra input handling functions. Select Cursor has a per layer cursor, and this was supposed to be an extra cursor that exists independently of the layer - so I am not sure of what to do.

ivan-mogilko:
> The order/extra input-handling function I am taking a look; I am leaning at adding extra input handling functions.

This is a question of a UI design rule: should editor always check for the active layer first, or should there be global actions that override everything, then active layer, and then global actions that do not override layer.

ericoporto:
I have a problem that for this to be seamless, I need to some changes to how zoom is handled so that the position the mouse is gets accounted for instead of the wheel just changing the zoom level.

I did some changes to reduce the amount of modified lines, and removed one extra variable I did not need. I can rebase these actually become:
- one commit for the "enable WS_EX_COMPOSITED";
- one commit for the pan commands;
- one commit for the not yet implemented zoom fix.

ivan-mogilko:
Sorry, I was planning to look again into this pr after dealing with other things because I already forgot how this works.

One thing though, could you rebase your pr branch on top of master instead of merging master into it?

ericoporto:
I can, yes! But I think you were right that the code wasn't great and I think the implementation was cleaner in a separate class - just really need a good name on it.

ericoporto:
I am closing this one in favor of #1089 . I think it's a better place for the code than the one presented here.

PR #1050, opened 2020-03-01
Supports wrapping a const buffer for the String class

ivan-mogilko:
This pr modifies a String, moving char buffer pointer and length parameter out of the allocated buffer header, and have them as a plain class members. This has two consequences:
* the simple reading operations no longer require extra pointer dereference;
* this allows to reference an external buffer, as an option.
(Note: I used [ScummVM's string as an example too](https://github.com/scummvm/scummvm/blob/master/common/str.h), although it's a bit different)

Latter lets support wrapping an external constant buffer without owning it. Which allows to even wrap string literals. This should be safe, because any writing operation in String class first ensures that owned string buffer exists and has only 1 reference (String::BecomeUnique does that).
Of course, when I say "safe", I assume it is used correctly, as in - this String should not be used after the the original const char buffer cease to exist.

As an addition, I modified script Dictionary and Set implementations to wrap over lookup keys instead of cloning these. This will be a proof of concept for the above changes.
Having to clone lookup keys there was bugging me for a while now. The perfomance gain may be not particulary big, but OTOH the cost for this is negligible, in my opinion, as most of the String class works the same as before, as it does not care where the string pointer is pointing to.

Same wrapping could be later done in other cases where it may be beneficial.

### DIFF
--- a/Common/util/string.cpp
+++ b/Common/util/string.cpp
@@ -24,39 +24,41 @@ namespace AGS
 namespace Common
 {
 
-String::Header::Header()
-    : RefCount(0)
-    , Capacity(0)
-    , Length(0)
-    , CStr(nullptr)
-{
-}
-
 String::String()
-    : _data(nullptr)
+    : _cstr(nullptr)
+    , _len(0)
+    , _buf(nullptr)
 {
 }
 
 String::String(const String &str)
-    : _data(nullptr)
+    : _cstr(nullptr)
+    , _len(0)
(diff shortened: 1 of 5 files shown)


morganwillcock:
I've not looked too closely, but it looks like this causes 64 bit builds to fail.

ivan-mogilko:
Oh right, this might be because of conflict between uint32 and size_t which is uint64 or more on 64-bit systems. I will fix this.

PR #1052, opened 2020-03-22
Script API: implemented Game.IgnoreUserInputForNextTextMs

ivan-mogilko:
By user request, this implements Game.IgnoreUserInputForNextTextMs, as a complement to existing Game.IgnoreUserInputAfterTextMs.

( Definitely, I am aware that these names are ugly :) )

The point of the existing property (Game.IgnoreUserInputAfterTextMs) was to help with misclicks when skipping character speech. When speech is also skipped automatically (by a timer), player may click too late, before realizing that speech was already changed to the next line, or final line ended. In this case player's click would skip next line before player is able to read it, or perform some other action. To counter this, the feature would block user's input for a limited amount of time after auto-skipping the text.

A new user request was to be able to restrict input blocking to only the case when there's next line of text. Because it's impossible to know if there's one at the time when the previous ends, following function is implemented:
* when the text is skipped automatically, it schedules an input block and starts a timeout, but does not disable input just yet.
* if a new line of text will be displayed during this timeout, then the scheduled input block is activated for the remainer of this timeout.
* if user input was already disabled by the first property (Game.IgnoreUserInputAfterTextMs), then the timeout that lasts longer is chosen.

This way a game dev may set up separate timeouts for disabling user input after any text line, and disabling it only if there's a new text line right afterwards.

Additionally, this PR fixes inconsistent handling of few "service" input commands, and cutscene skipping (because input is still handled separately in various game states in the engine code). By the rule, the cutscene should be always skippable, even if player's input is otherwise blocked.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2650,6 +2650,10 @@ builtin struct Game {
   /// Gets the number of cameras.
   import static readonly attribute int CameraCount;
 #endif
+#ifdef SCRIPT_API_v351
+  /// Gets/sets the time for which user input is ignored for the new text, after previous one was automatically removed
+  import static attribute int IgnoreUserInputForNextTextMs;
+#endif
 };
 
 builtin struct GameState {
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -954,9 +954,9 @@ bool DialogOptions::Run()
           parserActivated = 1;
       }
 
-      int mouseButtonPressed = ags_mgetbutton();
-
-      if (mouseButtonPressed != NONE)
+      int mouseButtonPressed = NONE;
+      int mouseWheelTurn = 0;
+      if (run_service_mb_controls(mouseButtonPressed, mouseWheelTurn) && mouseButtonPressed >= 0)
       {
         if (mouseison < 0 && !new_custom_render)
         {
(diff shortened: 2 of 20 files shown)


messengerbag:
I'm still sort of confused about why this approach is necessary.

Instead of all this, wouldn't it make more sense to just alter the logic that allows a user to skip a line (i.e., takes the user input and causes the dialog to advance under certain conditions), so that it simply checks whether the timeout is active before triggering the skip? That should automatically ensure that there aren't any side effects (blocking players from skipping cutscenes, etc.), right?

I would also cancel the timeout when it blocks a click/keypress skip, so that it only ever blocks the first one.

ivan-mogilko:
> Instead of all this, wouldn't it make more sense to just alter the logic that allows a user to skip a line (i.e., takes the user input and causes the dialog to advance under certain conditions), so that it simply checks whether the timeout is active before triggering the skip? That should automatically ensure that there aren't any side effects (blocking players from skipping cutscenes, etc.), right?

This is what the new property is doing, unless I misunderstand what are you refering to.

> I would also cancel the timeout when it blocks a click/keypress skip, so that it only ever blocks the first one.

It cancels the timeout if there's a click while there's no new text on screen. I could make it cancel it always.

ivan-mogilko:
I am currently investigating whether it's possible to implement wanted behavior in script using BlockingTextIndex property (#1053).

ivan-mogilko:
Cancelled, since #1053 seem to be enough for now.

Issue #769, opened 2019-05-14
Define a project goal


morganwillcock:
To resolve #762, to then resolve #757, the primary goal of AGS needs to be defined. At the moment the position of any future development is not clear, and any change is potentially deemed to a move in an incorrect direction.

I'll leave some relevant text from the website here, which is probably the closest thing to a goal that exists for the moment:

> Adventure Game Studio (AGS) provides the tools to make your own adventure, for free! Bring your story and artwork and slot it in, and let AGS do the rest. AGS provides everything you need from within one easy-to-use application. Create, test and debug your game, all in one place.

gurok:
I suggest creating a sub-task for assembling a goal-defining committee.

AlanDrake:
At the risk of sounding obvious, I think the goal of a tool for making adventure games is making adventure games.

ivan-mogilko:
Every tool has to meet certain demands. What is important for it, what is not, what is priority, what is not, what functionality should development aim and what should not. This is what I meant when I suggested discussing the goal. "Making adventure games" does not define anything. You can make adventure games on paper - problem solved, you don't need AGS?

Anyhow, it seems no one is really interested in doing this, as little were interested in defining these goals years ago. So apparently program will continue to be developed as it was: everyone will just add what they want without consistency.

AlanDrake:
I only wrote that to remind what the most important purpose is. 
As for development, to me the primary goal is making the codebase simpler and easier to handle, because that makes further change easier.
Other goals are implementing new features, coming up with ideas to improve the scripting (like a better audio api as you also suggested), give the possibility to overwrite hardcoded behaviours and so on.
Ideally, we should work toward decreasing the mess because we wanted to rebuild a better foundation.

ivan-mogilko:
> Other goals are implementing new features, coming up with ideas to improve the scripting (like a better audio api as you also suggested), give the possibility to overwrite hardcoded behaviours and so on.
Ideally, we should work toward decreasing the mess because we wanted to rebuild a better foundation.

I guess we are talking about different "levels" of goals. The "new features" is not what I had in mind.
The "better foundation" is. But my belief is that this foundation should be defined on its own, as an ideal we are going towards. After the goal is set we may decide how to reach it from where we are, which is a separate problem.

[...]

PR #1059, opened 2020-04-23
Add project specification document

morganwillcock:
I've been working for a while try to create a document to establish direction of the project and try to gather my thoughts regarding how I feel AGS should be updated. After much note-taking and rewriting I came to realise that by following by own logic it was impossible to make any meaningful and valid decision without first establishing what AGS is meant to do. In fact, my only goal here is to close a single issue (#769).

Within the document I've started the issues which I feel are relevant to make a decision, but also I have purposefully omitted any sort of conclusion before submitting this as a change request.

I would like to request feedback on any missing relevant information or errors, to get the document to the point that it is deemed a good representation of the current situation. Perhaps after some time has elapsed (maybe 3-4 weeks?) active committers will agree to merge the document to record a decision, and then the project goal will be set.

Note: the document may contain diagrams so a CI task is configured to export everything to HTML

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -360,3 +360,14 @@ make_release_task:
         --header "Content-Type: application/octet-stream" \
         "https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_RELEASE/assets?name=${fpath##*/}"
     done
+
+project_specification_task:
+  container:
+    image: alpine:3.10
+  install_packages_script: apk add --no-cache emacs-nox graphviz ttf-freefont
+  tangle_readme_script: >
+    cd project &&
+    emacs --batch --quick --load=org --eval "(org-babel-tangle-file \"README.org\")"
+  build_documents_script: project/build.sh
+  documents_artifacts:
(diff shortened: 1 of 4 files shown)


AlanDrake:
I stand by my former decisions:
AGS primary goal is to make games
The current incarnation, which deviated from the original goals, can keep existing in the AGS3 branch with appropriate bugfixes and support but no new fetures.

morganwillcock:
Yes. I think I would agree that it has accidentally deviated from its original course, so I am interested to hear from anyone with a counter argument or different view.

morganwillcock:
I was expecting that someone would challenge the position, but it has been two months. No-one has directly argued the opposite and the [forum poll](https://www.adventuregamestudio.co.uk/forums/index.php?topic=57990.0) shows minimal objections (as well as numerous people questioning why this is even an issue).

So I think I suggest to close this PR, submit a new PR to the main readme file which states the purpose, without being too specific but making it clear that this is a game development tool, and then this closes #769.

ericoporto:
Agreed. Let's move forward! :D

PR #1060, opened 2020-04-25
adds alleg-sdl2digi.so to Linux build target

ericoporto:
`alleg-sdl2digi.so` is currently shipped with the Editor/Linux package but it's not part of the built target of the Editor, so it's missing when building for Linux using the Editor.

### DIFF
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetLinux.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetLinux.cs
@@ -22,6 +22,7 @@ namespace AGS.Editor
             string[] libs =
             {
                 "alleg-alsadigi.so",
+                "alleg-sdl2digi.so",
                 "alleg-alsamidi.so",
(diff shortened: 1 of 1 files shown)


morganwillcock:
I don't think there is reason not to add this, it was likely only left out by accident and should probably go into the release as a fix. Unless anyone knows any different?

ericoporto:
I believe so. 

I was curious if it should additionally be added here actually
https://github.com/adventuregamestudio/ags/blob/master/debian/ags%2Blibraries/hooks/B00_copy_libs.sh

Because the CI resulting package had the library already, I was unsure if this Script was still used.

morganwillcock:
Technically that script was just for the chroot build and is copying the packaged Debian libraries rather than the patched+SDL2 Allegro build from the Allegro CI. The AGS CI build is also using the script (with small modifications) but expects that the SDL2 parts are not there. So I think this script is correct as it is, for now at least.

PR #1063, opened 2020-05-06
Fix OSX build

dploeger:
This PR fixes and correctly documents the OSX build to build on macOS up to and including Catalina.

It also removes the obsolete Xcode project. Currently, signing informations are missing, but I'll add them afterwards when I know how to do it. 😉 

I don't currently know the AGS build infrastructure, but would it be possible to add an automated mac engine build as well? If you're using travis, they support a osx build agent.


morganwillcock:
It is actually already building for macOS with CMake (if you look at 'checks' which ran on your PR then you will see a debug and release build was created). But since the old build method is still in the repository it is probably best to either update or remove it. There might be an issue with the downloads being continually pulled from someone's AWS account (which increases their bill), which is why I was reluctant to add the old automated build in the past, although in theory we can cache some downloads with the CI provider. If you give me a few days I will try to test the changes out and try adding the tasks to the automated builds.

dploeger:
Oh, okay. Then it perhaps really was just missing documentation. My goal would be to have an automated build for macOS that is added to the assets of each release, so non tech-savvy users can quickly enable their game for macOS.

morganwillcock:
> My goal would be to have an automated build for macOS that is added to the assets of each release, so non tech-savvy users can quickly enable their game for macOS.

The problem isn't so much with not having the binaries (since they already exist), it is primarily the build process within the editor that doesn't support it - it has no concept of release packaging that can preserve filesystem metadata. As a secondary problem there are the signing and notarization requirements, and as far as I know the tooling for this is macOS only.

dploeger:
@morganwillcock true. That would be even better. For now, I'd settle with: Copy game data into an unsigned package and ship that out. :)

morganwillcock:
It would be likely the game would not be executable if you just copied the files. The build process needs to write within an archive and set flags for how permissions should be reapplied, instead of to whichever filesystem Windows is using. The Linux build target has the same problem; non tech-savvy users distribute a Linux version of their game and it doesn't run.

But at the very least, having an outdated Xcode project sticking around is confusing, so if the changes here just help to clean things up I think that is beneficial.

dploeger:
No, it actually works. Just copy the compiled engine binary and game resources to the .app-folder I added in this PR and that makes an (unsigned) app. It's really easy as that.

morganwillcock:
The executable flag isn't part of the binary, it is set on the filesystem. Whether it stays executable when moved to a different filesystem isn't under its control.

dploeger:
Maybe that's no problem on macOS. I have converted some games now and never had to care about that.

dploeger:
On macOS, the info.plist in the .app/Content is the most important path for running an app. So maybe macOS takes care of the executable flag itself.

morganwillcock:
> 
It seems that this build is using the wrong Allegro version for AGS 3.5, it needs to use an Allegro release from here: https://github.com/adventuregamestudio/lib-allegro/releases

...I think this is why someone had reported that the colors were incorrect by someone else who tested the process.
 
> On macOS, the info.plist in the .app/Content is the most important path for running an app. So maybe macOS takes care of the executable flag itself.

I don't think it does, if you remove the executable flag on the engine binary I wouldn't expect the .app bundle to work when you double click it.

dploeger:
> It seems that this build is using the wrong Allegro version for AGS 3.5, it needs to use an Allegro release from here: https://github.com/adventuregamestudio/lib-allegro/releases

I could change that in the download.sh script in the libsrc directory. But wouldn't that break the other builds as well? Or isn't another build using that anymore?

> I don't think it does, if you remove the executable flag on the engine binary I wouldn't expect the .app bundle to work when you double click it.

Yeah, sorry. You're right. I just tested that on a release. I've seen hints about macOS automatically setting the flag in an application even if it is packaged on a filesystem that has no concept of an executable bit.

We should test this.

morganwillcock:
> I could change that in the download.sh script in the libsrc directory. But wouldn't that break the other builds as well? Or isn't another build using that anymore?

As far as I know, the CMake build deals with the libraries separately. So if you can modify the script to get the newer Allegro source I think that would likely be fine.

One thing I wanted to check: when you are setting `-mmacosx-version-min=10.14` is that because 10.14 is the actual minimum, or is that because 10.14 is the version you are testing on?

> Yeah, sorry. You're right. I just tested that on a release. I've seen hints about macOS automatically setting the flag in an application even if it is packaged on a filesystem that has no concept of an executable bit.

I think that could only be true if the application was signed, for unsigned programs you will likely find the opposite is true and you will get quarantine flags applied to anything of potentially questionable origin. A program downloaded through a web browser in a zip file is likely to have every file and directory 'quarantined' when extracted, whereas the same program downloaded in a .dmg image might not.

dploeger:
> As far as I know, the CMake build deals with the libraries separately. So if you can modify the script to get the newer Allegro source I think that would likely be fine.

Can surely do, but let me reiterate on the cmake build. I wasn't aware of that one and am a bit confused. Can it be used to build on macos as well? Because if it's the standard now, we could simply ditch the old make build and fix the docs.

> One thing I wanted to check: when you are setting -mmacosx-version-min=10.14 is that because 10.14 is the actual minimum, or is that because 10.14 is the version you are testing on?

That thing was something I didn't change. I just moved from 32bit to 64bit there. I'm testing this on 10.15 so maybe I should change that line as well.

> I think that could only be true if the application was signed, for unsigned programs you will likely find the opposite is true and you will get quarantine flags applied to anything of potentially questionable origin. A program downloaded through a web browser in a zip file is likely to have every file and directory 'quarantined' when extracted, whereas the same program downloaded in a .dmg image might not.

No. Regardless wether an app is distributed as a zip or dmg (this doesn't really matter, dmg is just another package format native to macOS), the application will always have to be confirmed on first run like this:

image

(The text reads in english: "The Crimson Diamond" was downloaded from the internet. Do you really want to open it" with more details about when the file was downloaded and from what domain)

When opening unsigned apps for the first time, this dialog comes up:

image

("Shism Tracker" can't be opened, because it was not signed. With the only options to delete it or cancel)

To work around this dialog, the app has to be right clicked and then "open" from the context menu has to be right clicked as well. Then this dialog comes up:

image

(macOS can't verify the developer of "Shism Tracker". Do you really want to open the app? With the additional option to open the app)

The only way to get around any dialog is by distributing the (signed) app over the Mac App Store by submitting it to the Apple gatekeepers, but that's probably mostly out of the question for most developers anyway.

morganwillcock:
> Can surely do, but let me reiterate on the cmake build. I wasn't aware of that one and am a bit confused. Can it be used to build on macos as well? Because if it's the standard now, we could simply ditch the old make build and fix the docs.

Yes, building on macOS was the original purpose for it being added.

<pre>export BUILD_TYPE=release
mkdir build_$BUILD_TYPE
cd build_$BUILD_TYPE
cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
make</pre>

> That thing was something I didn't change. I just moved from 32bit to 64bit there. I'm testing this on 10.15 so maybe I should change that line as well.

It is part of your commit:
<pre>- PLAT_FLAGS:=-mmacosx-version-min=10.7 -arch i386 -arch x86_64
+ PLAT_FLAGS:=-mmacosx-version-min=10.14 -arch x86_64
</pre>

Maybe it is set based on the XCode version which is installed. I mention it because, the CMake build works on 10.13 and is 64 bit.

> No. Regardless wether an app is distributed as a zip or dmg (this doesn't really matter, dmg is just another package format native to macOS), the application will always have to be confirmed on first run like this...

The quarantine flag is part of the trigger for Gatekeeper. dmg files are formatted volumes where all metadata is preserved, an archive contains requests to apply metadata to files as they are extracted to requested paths. These are fundamentally two different things and they are not treated as equal by macOS security. If you would iike an example of it making a difference here is one:

https://github.com/renpy/renpy/issues?q=is%3Aissue+macos+author%3Amorganwillcock
(I won't link directly the issue to avoid generating noise on it)

dploeger:
> Yes, building on macOS was the original purpose for it being added.

Oh, okay. This project still confuses me. 😄 Then I will try compiling with cmake on 10.15 and if it works ditch the old make build and fix the docs. I'll get back to you.

> It is part of your commit:

Hmm, whoops. I guess, then it *was* me changing that, though I don't remember it.

> The quarantine flag is part of the trigger for Gatekeeper. dmg files are formatted volumes where all metadata is preserved, an archive contains requests to apply metadata to files as they are extracted to requested paths. These are fundamentally two different things and they are not treated as equal by macOS security. If you would iike an example of it making a difference here is one:

Awesome. Thanks for clarifying that. I wasn't aware of that as I had no such problems before. I guess, it was just luck for me.

dploeger:
Cmake build works like a charm on 10.15. So, I guess, I should close this PR and recreate a new PR where I remove everything from the old build, right?

morganwillcock:
Maybe just adding the build instructions is OK for the short term. In theory CMake can generate an Xcode project, but I've not had a chance to take a proper look at how CMake works yet.

dploeger:
Do we actually need an xcode project? It seems it should be okay to just [have the .app](https://www.digicert.com/kb/code-signing/mac-os-codesign-tool.htm)

morganwillcock:
It would still build an app bundle, but I think the old build method combines the engine and data and signs it. If signing is just applied to a standalone engine that would be rejected for store submission and risks the developer certificate being revoked, because technically what was signed is a script interpreter that anyone could use it to run their own script/game/virus.

dploeger:
The old build method with make? No, it just creates a binary of the engine and I haven't seen it sign anything. The data is taken from the resources folder in the app bundle. I've always compiled the binary, moved it to my template-app-bundle and added the game resources.

morganwillcock:
> The old build method with make?

No, building within Xcode using an Xcode project where each game is defined inside it.
https://github.com/JanetGilbert/ags/tree/osx-nick/OSX/xcode

dploeger:
Oh that. Yes, I trashed the whole directory, because the references to the sources were pretty old and unstable. 😄

ivan-mogilko:
IDK about OSX, but I think all other platforms are now using same freetype src from within the repository: https://github.com/adventuregamestudio/ags/tree/master/Common/libsrc/freetype-2.1.3

to keep font looks consistent between systems.

ivan-mogilko:
I also think it's better to leave old xcode files for now. I am no expert about xcode and don't know what most of these are, but I recall for example OSX/xcode/ags/plugin_registration.cpp was made only few years ago for a purpose of embedding plugins in osx application.

ericoporto:
I tried to follow the instructions in [dploeger/ags/blob/osx64build/OSX/README.md](https://github.com/dploeger/ags/blob/osx64build/OSX/README.md). It's a clean Mojave system with the latest Xcode available for it installed. Unfortunately I hit [an error after getting in the `make install`](https://gist.github.com/ericoporto/1a7c55f68364a6869287ef9688dd10c3) part. Here is the kicker. After I copy pasted and run the commands in [dploeger/ags/blob/osx64build/OSX/**buildlibs**/README.md](https://github.com/dploeger/ags/blob/osx64build/OSX/buildlibs/README.md) it did work!

CMake build also worked. I also tried to generate an Xcode project using CMake:

```bash
mkdir buildXcode
cd buildXcode
cmake -G Xcode ..
```

This created an AGS.xcodeproject under the buildXcode directory. I loaded and built it on Xcode and it also worked! (all comments here were using the branch involved in this PR)

morganwillcock:
The Xcode project needs to also incorporate the game data if it is to be signed. If you look at that previous ones that were built by hand there is an abstraction within them called something like `AGSKit` which (I'm assuming) would be missing from the one that CMake generates.

But just to reiterate: At the moment everything except a CMake based build will be using the wrong Allegro version and won't operate correctly. It is probably best to just document the CMake build as the one to use and come back to the Xcode part later on.

ericoporto:
I did a test using the Xcode project generated from CMake from the master branch.
There, there is a directory called `OSX/Resources`.
image
It appears this directory is part of the Xcode project generated on the CMake there. The game has to have the name in the CMakeLists.txt, so rename it to acgame.dat or change this after in the generated Xcode project. Double clicking the generated AGS.app directory in finder makes the game run.

ivan-mogilko:
> The game has to have the name in the CMakeLists.txt, so rename it to acgame.dat or change this after in the generated Xcode project.

BTW, "acgame.dat" is an incorrect name which was used in original ports as a result of misconception. It is unfortunate to still have it mentioned there.

The official game file name is <game>.ags.

ericoporto:
https://github.com/adventuregamestudio/ags/blob/3e65dd364e81103d9e87083a55953f67aa22ef34/Engine/platform/osx/alplmac.mm#L7

https://github.com/adventuregamestudio/ags/blob/3e65dd364e81103d9e87083a55953f67aa22ef34/Engine/CMakeLists.txt#L654

I think these are the two places to fix in the OSX code to change to `game.ags`. Also need to rename the empty `ac2game.dat` file that is there so that the cmake build doesn't fail. I did a test on my MacOS environment and it worked. I can do a PR for the mentioned changes.

dploeger:
@ericoporto if you could correct the build documentation in your pr as well, we could use that for the final merge. I would close this pr then.

ericoporto:
These would be the mentioned changes: https://github.com/adventuregamestudio/ags/compare/master...ericoporto:osx-instructions

It would be good if someone validated it.

ivan-mogilko:
Better to make a new PR and discuss it there.
Other than this, I always suggest to keep unrelated changes in separate commits (e.g. one for filename change, another for instruction change).

ericoporto:
OK, done, opened #1083 .

morganwillcock:
@dploeger, just to check... are you happy for this PR to be closed, now that the instructions have been updated?
https://github.com/adventuregamestudio/ags/blob/master/OSX/README.md

dploeger:
Yes, absolutely. I can add some minor tweaks to the readme in another PR. Thanks.

PR #1065, opened 2020-05-12
Compiler: Default parameters and improvements

### LINKED ISSUE #188: Compiler Inconsistency: Default value for float 0.0 needs to be integer 0
Better to explain with an example:

In a global script/module header, let's say you declare a function with a default float value:

``` C++
import function FadeOut(float delaySeconds = 0.0);
```

The compiler will error out completely with the message:

>  Parameter default value must be literal

It turns out that the default value must be expressed as integer:

``` C++
import function FadeOut(float delaySeconds = 0);
```

Which is fairly inconsistent with the rest of the complier who is really strict about using decimals for the rest of the variables.

This is a really old error, perhaps there since the introduction of modules to the scripting system. I just forgot about it until I ran into it recently. :)

AlanDrake:
Allows for float values to be used and shows errors when trying to use non-zero integers for float, String, or pointer types.
Fixes #188

### DIFF
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -603,13 +603,26 @@ int check_not_eof(ccInternalList &targ) {
   return 0;
 }
 
-int check_for_default_value(ccInternalList &targ, int funcsym, int numparams) {
+// return the float as an int32 (but not actually converted to int)
+int float_to_int_raw(float toconv) {
+    union
+    {
+        float   f;
+        int32_t i32;
+    } conv;
+    conv.f = toconv;
+    return conv.i32;
+}
 
+int check_for_default_value(ccInternalList &targ, int funcsym, int numparams) {
     if (sym.get_type(targ.peeknext()) == SYM_ASSIGN) {
+        const char *vartype = sym.get_name(targ.script[targ.pos - 2]);
         // parameter has default value
         targ.getnext();
         int defValSym = targ.getnext();
         bool negateIt = false;
+        bool isFloat = false;
+        bool isPointer = strchr(vartype, '*') != NULL;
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Could you clarify what are rules for assignment now?

In the comment [here](https://github.com/adventuregamestudio/ags/issues/188#issuecomment-627603685) you say:
> 0 (null) is always allowed (even for floats, so nothing breaks)

I tried typing:
`float yyy = 0;`
in script (both inside and outside of function), and it shows error message:
`GlobalScript.asc(5): Error (line 5): Expected floating point value after '='`

ivan-mogilko:
I think this is working. I had an irk when I saw `targ.script[targ.pos - 2]`, but guess the way this function is called it's guaranteed there's at least 2 words before.

One request though, could you add a test for this new rule in parser tests:
<https://github.com/adventuregamestudio/ags/blob/master/Compiler/test/cs_parser_test.cpp>
(the tests could be run out of Compiler.Lib solution).

AlanDrake:
> So, there's now a strange inconsistency, you can do
> `function f( float = 0 );`
> but you cannot do
> `float f = 0;`
> 
> to confirm, is this the expected behavior?

Yes, this change is only for defaults, and I kept the `function f( float = 0 );` for compatibility because some modules would break otherwise, also it was always a standard that 0 meant either "null" or "0". I just wanted to add float support and throw errors misusing this feature.

I think it could be done better (eg: using keyword "null" for pointers), but perhaps I'd leave that for AGS4.
Sorry for the "-2", unfortunately there was no convenient way to discern the variable name or type in this part of the code, but it should be safe the way it's called. It should only be either the name of the parameter or "*".

I'll see to add the tests.

AlanDrake:
Added the new test rules inside cs_parser_test.cpp

PR #1070, opened 2020-05-16
Fixed that cursor drawn in editor didn't scale

### LINKED ISSUE #1068: Editor, mouse cursor editing pane, setting cursor hotspot via click produces wrong coordinates
In the editor's mouse cursor pane, clicking the cursor image to set the hotspot produces wrong HotspotX / HotspotY coordinates

**AGS Version**
3.5.0_22

**To Reproduce**
1. Edit any cursor mode
2. Click the image to set a hotspot
3. HotspotX / HotspotY are set to wrong values (half of actual offset)

**Expected behavior**
Coordinates corresponding to click

persn:
Fixes #1068

### DIFF
--- a/Editor/AGS.Editor/Panes/ViewPreview.cs
+++ b/Editor/AGS.Editor/Panes/ViewPreview.cs
@@ -113,20 +113,17 @@ namespace AGS.Editor
                     spriteWidth *= 2;
                     spriteHeight *= 2;
                 }
-                int x = 0, y;
-                y = previewPanel.ClientSize.Height - spriteHeight;
-                if (chkCentrePivot.Checked)
+                if (spriteWidth <= previewPanel.ClientSize.Width && spriteHeight <= previewPanel.ClientSize.Height)
                 {
-                    x = previewPanel.ClientSize.Width / 2 - spriteWidth / 2;
+                    int scale = Factory.AGSEditor.CurrentGame.GUIScaleFactor;
+                    int x = chkCentrePivot.Checked ? previewPanel.ClientSize.Width / 2 - (spriteWidth * scale) / 2 : 0;
+                    int y = previewPanel.ClientSize.Height - (spriteHeight * scale);
+                    IntPtr hdc = e.Graphics.GetHdc();
+                    Size spriteSize = Utilities.GetSizeSpriteWillBeRenderedInGame(spriteNum);
+                    Factory.NativeProxy.DrawSprite(hdc, x, y, spriteSize.Width * scale, spriteSize.Height * scale, spriteNum, thisFrame.Flipped);
+                    e.Graphics.ReleaseHdc();
                 }
-				if ((spriteWidth <= previewPanel.ClientSize.Width) &&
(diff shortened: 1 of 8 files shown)


persn:
I might be able to write a test for this, so don't review it for now

ivan-mogilko:
This is not the only place that requires correcting. Previously editor relied on the native part doing scaling, and it's still done there although under wrong condition. I am looking into this right now.

Ideally, I think, in current version the native part should provide an image scaled according to the "relative resolution" rule, which editor then could zoom in or out for user convenience.

ivan-mogilko:
Please also note, that GetSizeSpriteWillBeRenderedInGame and GUIScaleFactor are conceptually different. First determines the final look of sprite in game at no scaling (this function's main goal is to account for backwards compatibility), second is strictly for user convenience (draw low-res sprites larger), and should probably be replaced with a proper zoom in later versions.

ivan-mogilko:
I *think* that it would be enough if you also remove lines 1179-80 here:
https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Native/agsnative.cpp#L1179
and use "1" in place of "scale" below.

Originally the scale there was chosen simply as 1 for hi-res sprites and 2 for low-res, but I misunderstood their meaning and rewrote into more complicated condition.

I am still bothered which other parts of the editor could've been broken by this. E.g. if there was this extra zoom for low-res sprites in other panes it's also now works incorrectly. From a quick research, GUI has its own sprite drawing routine, so what's left to check is:
- Inventory pane
- View pane

And I don't remember how it's done in rooms.

ivan-mogilko:
Oh, completely forgot, this fix ideally should go into "release-3.5.0" branch.

persn:
Maybe I misunderstood the suggestions, please check.

The View pane looked fine.

Cannot write a test here because testing framework isn't in the branch, also the code relies on the native proxy which I don't feel like writing mocks for.

ivan-mogilko:
Yes, the native change is correct.

But there are still problems in two panes that I noticed:
* Character pane, where you preview character sprite,
* View pane, on the side where you preview animation.

The easiest way to see this is to compare same template opened in AGS 3.4.3 and 3.5.0. In AGS 3.4.3 low-res sprites on above panels were doubled (similar to cursor, inventory item etc).

persn:
Both panes mentioned was actually the same pane used in 2 places, which is why I fixed only one pane since last feedback. Ready to review

Editor/AGS.Editor/Panes/ViewPreview.cs
```diff
@@ -113,20 +113,17 @@ private void previewPanel_Paint(object sender, PaintEventArgs e)
                     spriteWidth *= 2;
```
ivan-mogilko, line 113: A sprite size is doubled right above your new code under same condition, it should be possible replace hardcoded x2 with GUIScaleFactor and perhaps use same variables when drawing too, instead of having (n * scale) everywhere?

ivan-mogilko:
It seems fine now!

PR #1071, opened 2020-05-16
Updates for room editor controls

### LINKED ISSUE #1054: Crash when address bar overflows at both ends
Possible bug [reported by forum user](https://www.adventuregamestudio.co.uk/forums/index.php?topic=57814.msg636617834#msg636617834).

> Whenever I put a walk-behind area in a room, save the room, and later edit it again, it says:
>
> "Unable to read the room design-time preferences. This should not affect the actual game data, but the design time-state of all room items will be reset."
>
>Also, there is this error information, which it says to post on the AGS forum:

<pre>
"Error: Der Index lag außerhalb des Bereichs. Er darf nicht negativ und kleiner als die Sammlung sein.
Parametername: index
Version: AGS 3.5.0.24

System.ArgumentOutOfRangeException: Der Index lag außerhalb des Bereichs. Er darf nicht negativ und kleiner als die Sammlung sein.
Parametername: index
   bei System.Collections.ArrayList.get_Item(Int32 index)
   bei System.Windows.Forms.ToolStripItemColle ction.get_Item(Int32 index)
   bei AddressBarExt.Controls.AddressBarExt.Ad dOrUpdateOverflow()
   bei AddressBarExt.Controls.AddressBarExt.Up dateBar()
   bei AddressBarExt.Controls.AddressBarExt.se t_CurrentNode(IAddressNode value)
   bei AGS.Editor.RoomSettingsEditor.SelectOld Node(IAddressNode currentNode)
   bei AGS.Editor.RoomSettingsEditor.RefreshLa yersTree()
   bei AGS.Editor.Components.RoomsComponent.Cr eateRoomSettings(DockData previousDockData)
   bei AGS.Editor.Components.RoomsComponent.Lo adDifferentRoom(UnloadedRoom newRoom)
   bei AGS.Editor.Components.RoomsComponent.Lo adRoom(String controlID)
   bei AGS.Editor.Components.RoomsComponent.It emCommandClick(String controlID)
   bei AGS.Editor.Components.BaseComponentWith Folders`2.CommandClick(String controlID)
   bei AGS.Editor.ProjectTree.ProcessClickOnNo de(String nodeID, MouseButtons button)
   bei AGS.Editor.ProjectTree.projectTree_Node MouseDoubleClick(Object sender, TreeNodeMouseClickEventArgs e)
   bei System.Windows.Forms.TreeView.OnNodeMou seDoubleClick(TreeNodeMouseClickEventArgs e)
   bei System.Windows.Forms.TreeView.WndProc(Message& m)
   bei System.Windows.Forms.Control.ControlNat iveWindow.OnMessage(Message& m)
   bei System.Windows.Forms.Control.ControlNat iveWindow.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callb ack(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)"
</pre>

> In the room, I then can't select anything from the "Edit this room's" window. When I test the room, it still works fine though.

[room10.crm.user](https://github.com/adventuregamestudio/ags/files/4397198/room10.crm.user.txt)

morganwillcock:
- Fixes that the current address bar node is not always saved into design time data
- Stop the address bar from implementing its own overflow to prefer the built-in ToolStrip overflow. Indirectly fixes #1054 
- Remove unused checkbox for "Display Coordinates as Low Res"
- Rearrange room editing controls to give more space to the address bar

To maintain vertical alignment of the dropdown buttons within the ToolStrip overflow, it seems to be a requirement that text be present in the item. So to fix this, empty dropdown items with an arrow are now replaced by the literal text "···" and no arrow. This has a side-effect of the buttons becoming wider and easier to click, which I think is an improvement over what is there now.

updated_address_bar

### DIFF
--- a/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
+++ b/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
@@ -348,6 +348,9 @@ namespace AddressBarExt.Controls
             else
                 tsButton = new ToolStripButton(node.DisplayName, null, NodeButtonClicked);
 
+            //AGS: match the padding on the dropdown buttons
+            tsButton.Padding = new Padding(1, 0, 1, 0);
+
             //attach the node as the tag
             tsButton.Tag = node;
 
@@ -372,7 +375,12 @@ namespace AddressBarExt.Controls
                 if (node.Children.Length > 0)
                 {
                     //create the drop down button
-                    tsddButton = new ToolStripDropDownButton("");
+                    //tsddButton = new ToolStripDropDownButton("");
+
+                    //AGS: use some text to pickup text layout styling
+                    //and provide a bigger target to click on
+                    tsddButton = new ToolStripDropDownButton("···");
+                    tsddButton.ShowDropDownArrow = false;
 
                     //check if we have any tag data (we cache already built drop down items in the node TAG data.
                     if (node.Tag == null)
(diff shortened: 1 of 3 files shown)


ericoporto:
Appears to work here! Any reason to comment code instead of deleting? They make harder to understand the comments that are about the intention of the code.

image

imgae

This is just in case if it's possible, when the list of objects or characters is presented, if it's big, it would be useful the scroll arrows to be a little bigger in height or the list accept mouse wheel scroll.

morganwillcock:
> Any reason to comment code instead of deleting? They make harder to understand the comments that are about the intention of the code.

This is more of a hack job on the original control code. I did think about deleting the unused code instead but:

- there is already commented out code, and since this control was imported with no traceable origin I can't tell whether this is the original author or something AGS specific
- if I delete the unused parts it actually makes all the original author's comments incorrect, since most of the functionality is now bypassed

...so overall, I felt it was more beneficial for someone trying to work out what had happened to leave the original code in-place. But to be clear, I also think this control was unsuitable for use in the first place, and should be replaced.

> This is just in case if it's possible, when the list of objects or characters is presented, if it's big, it would be useful the scroll arrows to be a little bigger in height or the list accept mouse wheel scroll.

I hadn't tried a large list of objects, but I will double check. There are various styles for the overflow layout, so it might be possible to modify the button size or work with the mouse wheel.

morganwillcock:
It looks like the vertical scrolling buttons are fixed height and not configurable, so changing the size needs the whole thing replaced by a custom control. The mouse scrolling would also need a custom control or faking keyboards events based on redirecting mouse wheel events (but I doubt this will work in this particular case because the visibility and lock buttons have broken the keyboard based scrolling).

So in summary, whilst using this particular control I think this is the best it can be.

PR #1073, opened 2020-05-16
Editor: show aspect ratio on resolution select

### LINKED ISSUE #1057: Game resolution settings should also display aspect ratio
This is a user suggestion, that the resolution selection should also display the associated aspect ratio. It seems a good idea and is probably fairly easy to add.

morganwillcock:
Adds extra labels to display the aspect ratio of the selected resolution.

game_res

Fixes #1057

### DIFF
--- a/Editor/AGS.Editor/GUI/CustomResolutionDialog.Designer.cs
+++ b/Editor/AGS.Editor/GUI/CustomResolutionDialog.Designer.cs
@@ -36,6 +36,8 @@
             this.btnOK = new System.Windows.Forms.Button();
             this.label1 = new System.Windows.Forms.Label();
             this.cbResolutionPreset = new System.Windows.Forms.ComboBox();
+            this.lblAspectRatioCalc = new System.Windows.Forms.Label();
+            this.lblAspectRatio = new System.Windows.Forms.Label();
             ((System.ComponentModel.ISupportInitialize)(this.udWidth)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.udHeight)).BeginInit();
             this.SuspendLayout();
@@ -59,7 +61,7 @@
             // label2
             // 
             this.label2.AutoSize = true;
-            this.label2.Location = new System.Drawing.Point(36, 91);
+            this.label2.Location = new System.Drawing.Point(39, 91);
             this.label2.Name = "label2";
             this.label2.Size = new System.Drawing.Size(38, 13);
             this.label2.TabIndex = 1;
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Perhaps also add aspect ratio to the items in combo box too?

morganwillcock:
I had thought about it. Showing it just in the dropdown list would require the custom entry in the list to be updated and there already seems to be repeated calls to the same line of code just for a UI update. So to avoid spending time trying to reverse engineer the update logic as well as the original intent, I've instead done exactly what the original pattern does and not introduced more dependencies on the update function chain/loop being called. i.e. the label update doesn't trigger any additional update event and so was the only thing I deemed a safe change.

ivan-mogilko:
I just thought it would ease selection for the user if they can see ratios along with resolutions in the preset list.

It was me who wrote this dialog originally, which logic is unclear? I will check it out to refresh my memories.

morganwillcock:
> It was me who wrote this dialog originally, which logic is unclear? I will check it out to refresh my memories.

There is a boolean value _updatingControls which is used to opt out of control changes, I imagine because responding to a change event is causing additional changes events, which is also why it looks like the same lines of code are running more than once for a single change and the 'set' code is duplicated.

ivan-mogilko:
I think the controls on this dialog are interconnected: numeric up-down control changes result in combo box switching to "Custom resolution" item, and setting combo box results in up-down control value change. Therefore I protected this with the `_updatingControls` flag.

But all this may be irrelevant as it's possible to adjust displayed text modifying Utilities.ResolutionToUserString function, which is used as a conversion delegate for this dialog.

I tested with following code:
<pre>
        public static Size UserStringToResolution(string s)
        {
            int aspect_c = s.IndexOf("[");
            if (aspect_c >= 0) s = s.Substring(0, aspect_c);
            String[] parts = s.Split('x');
            return new Size(Int32.Parse(parts[0]), Int32.Parse(parts[1]));
        }

        public static string ResolutionToUserString(Size size)
        {
            return String.Format("{0} x {1} [{2}]", size.Width, size.Height, ResolutionToAspectRatio(size));
        }
</pre>

Note that UserStringToResolution does not seem to be called in common use scenarious, so I had to test it's work by calling directly from another function.

ivan-mogilko:
BTW, I noticed that you retargeted this PR from master to release-3.5.0, was this intentional or by mistake?

morganwillcock:
It seems that the two utility functions are called independently of type conversion, so I think it would be a mistake to make the change there since:

- it would place a reliance on the two functions having similar string formats in order for the type converter to reliably use them

- use of the functions outside of the type converter would suddenly get the aspect ration appended to the resolution, and I think that should be probably be an opt-in result rather force all callers to deal with the extra text

So in the end, since preset dropdown items are only constructed once I've just appened the aspect ratio there. I've also changed the PR target to master.

ivan-mogilko:
> it would place a reliance on the two functions having similar string formats in order for the type converter to reliably use them

but it's devs who decide what type converter does, so if a utility function has a specification that defines what it does, cannot type converter rely on it?

> use of the functions outside of the type converter would suddenly get the aspect ration appended to the resolution, and I think that should be probably be an opt-in result rather force all callers to deal with the extra text

perhaps, then we may have another pair of functions?

morganwillcock:
> but it's devs who decide what type converter does, so if a utility function has a specification that defines its behavior, cannot type converter rely on it?

Just going on the function names and parameters, how it works at the moment can be argued to correct, even though the dependencies are a little fragile. i.e. all use cases are expecting the same result and there is no question about whether a textual description of a resolution should also include the ratio.

> perhaps then we could add an option to that function, or another pair of functions?

I had an intermediate revision which made the aspect ration output optional, but later felt that this fundamentally breaks the idea of type conversion. At the very least it wouldn't make sense to have the converter return a different result than function named to do the same thing, and we are then saying that the sole purpose of the converter is only to work with this particular dropdown list and nothing else.

> PS. Alternatively we could perhaps put these functions right in the converter class, and make them static , thus callable by external user too. This way both property display and this dialog will rely on same converter.

To be honest this feels like a separate issue. I don't think that what I've changed that makes the situation worse and refactoring the form and type conversion wasn't really my goal here.

ivan-mogilko:
Converter is not working with dropdown list, it's working with the property display in the property grid (General Settings). I though it will be simple to display aspect ratio there too while it's changed in the dialog.
But okay, if you wanted to only display it in the dialog, then this PR is done.

Issue #1015, opened 2019-12-22
String.AsFloat and String.Format locale dependant when print float number


ericoporto:
**Describe the bug**
[`String.AsFloat`](https://adventuregamestudio.github.io/ags-manual/String.html#stringasfloat) is Locale dependant, so as `String.Format` when using `%f`, this makes the example on the manual for AsFloat fail on a locale that uses comma instead of dot for floating point.

**AGS Version**
AGS 3.5.0.22

**To Reproduce**
Run the code below on a locale that uses comma instead of dot as decimal separator.

```AGS Script
String text1 = "57.362";
float number1 = text1.AsFloat;
Display("%f",number1);
```
You should get `57,0000` printed on screen, because `number1` has output 57.0 and the formatting `%f` uses comma instead of dot, so you get `57,0000` and not `57.0000`.

**Expected behavior**
number1 should receive 57.362 as value, and Display should use dot instead of comma, displaying `57.362`.

**Desktop (please complete the following information):**
 - OS: Ubuntu Linux 
 - Version 18.04

ericoporto:
I had no success using `setlocale(LC_ALL, "C");` on  `ags/Engine/game/game_init.cpp` and a grep for setlocale finds it in some libs.

But once I placed in  [ags/Engine/main/game_start.cpp#L86-L87)](https://github.com/adventuregamestudio/ags/blob/master/Engine/main/game_start.cpp#L86-L87) like this:

    ...
    void start_game() {
        setlocale(LC_ALL, "C");
        set_cursor_mode(MODE_WALK);
    ...

It worked!

ericoporto:
I opened a PR for this on #1075

ericoporto:
Just for reference, this turned out to be a bug in how allegro set locale for keyboard in the x allegro backend used in the Linux port. Because of this, the bug only affected the Linux port.

I am closing this issue since the fix has been pushed to master.

PR #1075, opened 2020-05-18
Fixes float locale dependant in #1015

ericoporto:
need to not use `LC_ALL`on MacOS

### DIFF
--- a/Engine/main/engine.cpp
+++ b/Engine/main/engine.cpp
@@ -426,6 +426,9 @@ void engine_init_keyboard()
 
     install_keyboard();
 #endif
+#if AGS_PLATFORM_OS_LINUX
+    setlocale(LC_NUMERIC, "C"); // needed in X platform because install keyboard affects locale of printfs
+#endif
 }
 
 void engine_init_timer()


AlanDrake:
So it causes probems on MacOS? What about just `LC_NUMERIC` instead of `LC_ALL`?

ericoporto:
I added the ifdef for it just for Linux since this wasn't reported in other platforms.

ivan-mogilko:
Sorry for the delay.

My only concern is the place where you put this; `start_game` function may be called more than once and is related to start running with already loaded data and configured options.
Also, it's best to avoid placing system #ifdefs in the game logic code.

Would it still work if placed in `ags_entry_point` function? There's already some system-specific code that looks like this:
`#if AGS_PLATFORM_OS_WINDOWS
    setup_malloc_handling();
#endif`

ericoporto:
Hey, I to tried the suggestion here, but for reasons I don't understand it doesn't work with `ags_entry_point`. I placed right after the mentioned `#if AGS_PLATFORM_OS_WINDOWS`.

ericoporto:
Found where the locale is being changed!!! It's in the `install_keyboard();` line on `void engine_init_keyboard()` in `Engine/main/engine.cpp`. I can't trace the error more than this, I think it's somewhere in allegro from this [forum messa here](). I can switch to use `LC_NUMERIC` and do it like this:

```cpp
void engine_init_keyboard()
{
#ifdef ALLEGRO_KEYBOARD_HANDLER
    Debug::Printf(kDbgMsg_Init, "Initializing keyboard");

    install_keyboard();
#endif
#if AGS_PLATFORM_OS_LINUX
    setlocale(LC_NUMERIC, "C");
#endif
}
```

ivan-mogilko:
This makes sense: first initialize library, then adjust settings.

Also, apparently xkeyboard.c is the only place where Allegro calls `setlocale`, which explains why no other port has this issue.

ivan-mogilko:
Actually, I think this may be cherry picked to release-3.5.0 too, since it's a simple fix to an obvious bug.

ericoporto:
Cool! I can finally run my ags tests without it failing in my PC because my float conversion checks xD

PR #1079, opened 2020-05-24
3.5.0: Display character speech relative to nearest viewport

ivan-mogilko:
This is not exactly a bugfix but rather unfinished business, it had to be in 3.5.0 new cameras but I forgot to add.

Originally, and still even after addition of new camera system, speech coordinates are set based on where character is located relative to **primary viewport**.

This change makes it find either _first viewport_ in the array this _character is seen in_, or one _which camera is nearest to the character_.

**NOTE:** this pr is based on #1076 code so should be merged after that.

A test game that illustrates the problem and allows to test the solution.
[test_camera_speech.zip](https://github.com/adventuregamestudio/ags/files/4674063/test_camera_speech.zip)
Controls:
* space - make character talk
* 1-4 - turn viewports on and off
* click around to make character walk, and see how speech position changes.

### DIFF
--- a/Common/util/geometry.cpp
+++ b/Common/util/geometry.cpp
@@ -12,6 +12,8 @@
 //
 //=============================================================================
 #include "util/geometry.h"
+#include <algorithm>
+#include <cmath>
 
 //namespace AGS
 //{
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Updated with a better demo game.

PR #1084, opened 2020-06-05
Script API: expand Wait functions set for more versatility

### LINKED ISSUE #647: WaitMouse(), allow indefinite Wait, more info about Wait-interrupt
AGS currently has `Wait()`, `WaitKey()` and `WaitMouseKey()` functions to block for a fixed time (or rather, number of game loops), with option to interrupt either with a keypress, or with a mouseclick _or_ keypress.

There is no way to set a block that can only be interrupted by a mouseclick. I propose a function `WaitMouse()` that blocks until there's a mouseclick.

There is also no way to block _indefinitely_ until there's a mouseclick or keypress. I propose that `WaitKey()`, `WaitMouseKey()` and the new `WaitMouse()` (but not plain `Wait()`) should be modified to accept negative `time` arguments, which would be interpreted as "block forever (until interrupted)".

Finally, it would be useful in some circumstance to know exactly how a block was interrupted. I propose that instead of returning just 0 if timed out and 1 if interrupted, `WaitKey()` and `WaitMouse()` would return the `keycode` or `MouseButton` that were pressed, and 0 if timed out. `WaitMouseKey()` could return the `keycode` value if interrupted by a keypress, `-MouseButton` (the negative of the `MouseButton`) value if interrupted by a mouseclick, and 0 if timed out.

ivan-mogilko:
Resolves #647

* Added WaitMouse(timeout) to complement existing functions (WaitKey, WaitMouseKey).
* All Wait functions return a more precise reason they were skipped:
  - `0` = ended by timeout;
  - `> 0` = keycode
  - `< 0` = negated mouse button code minus one (because of how mouse codes start with 0). In order to get mouse code do this: `MouseButton btn = -(result + 1);`
* All Wait* functions (*including* Wait) may have infinite timeout if negative argument is passed; in all functions except Wait() timeout argument is now optional with default value -1.
* Added SkipWait() function that immediately cancels any active Wait*.

NOTE: SkipWait will be only useful if called from `repeatedly_execute_always` or `late_repeatedly_execute_always`.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1397,6 +1397,10 @@ import void UnPauseGame();
 import void Wait(int waitLoops);
 /// Blocks the script for the specified number of game loops, unless a key is pressed.
 import int  WaitKey(int waitLoops);
+#ifdef SCRIPT_API_v351
+/// Blocks the script for the specified number of game loops, unless the mouse is clicked.
+import int  WaitMouse(int waitLoops);
+#endif
 /// Blocks the script for the specified number of game loops, unless a key is pressed or the mouse is clicked.
 import int  WaitMouseKey(int waitLoops);
 /// Checks whether the specified key is currently held down.
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1396,13 +1396,13 @@ import void UnPauseGame();
 /// Blocks the script for the specified number of game loops.
 import void Wait(int waitLoops);
 /// Blocks the script for the specified number of game loops, unless a key is pressed.
-import int  WaitKey(int waitLoops);
+import int  WaitKey(int waitLoops = -1);
 #ifdef SCRIPT_API_v351
 /// Blocks the script for the specified number of game loops, unless the mouse is clicked.
-import int  WaitMouse(int waitLoops);
+import int  WaitMouse(int waitLoops = -1);
 #endif
 /// Blocks the script for the specified number of game loops, unless a key is pressed or the mouse is clicked.
-import int  WaitMouseKey(int waitLoops);
+import int  WaitMouseKey(int waitLoops = -1);
 /// Checks whether the specified key is currently held down.
 import bool IsKeyPressed(eKeyCode);
 import void SetGlobalInt(int globalInt, int value);
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1397,12 +1397,14 @@ import void UnPauseGame();
 import void Wait(int waitLoops);
 /// Blocks the script for the specified number of game loops, unless a key is pressed.
 import int  WaitKey(int waitLoops = -1);
+/// Blocks the script for the specified number of game loops, unless a key is pressed or the mouse is clicked.
+import int  WaitMouseKey(int waitLoops = -1);
 #ifdef SCRIPT_API_v351
 /// Blocks the script for the specified number of game loops, unless the mouse is clicked.
 import int  WaitMouse(int waitLoops = -1);
+/// Cancels current Wait function, regardless of its type, if one was active at the moment.
+import void SkipWait();
 #endif
-/// Blocks the script for the specified number of game loops, unless a key is pressed or the mouse is clicked.
-import int  WaitMouseKey(int waitLoops = -1);
 /// Checks whether the specified key is currently held down.
 import bool IsKeyPressed(eKeyCode);
 import void SetGlobalInt(int globalInt, int value);
--- a/Engine/ac/gamestate.h
+++ b/Engine/ac/gamestate.h
@@ -147,6 +147,8 @@ struct GameState {
     int   bg_frame,bg_anim_delay;  // for animating backgrounds
     int   music_vol_was;  // before the volume drop
     short wait_counter;
+    char  wait_skipped_by; // tells how last wait was skipped [not serialized]
+    int   wait_skipped_by_data; // extended data telling how last wait was skipped [not serialized]
     short mboundx1,mboundx2,mboundy1,mboundy2;
     int   fade_effect;
     int   bg_frame_locked;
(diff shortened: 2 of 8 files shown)


ivan-mogilko:
@messengerbag, here's a temp build if you'd like to test this too:
https://cirrus-ci.com/task/5888965205032960

messengerbag:
@ivan-mogilko 
This is great! I will see about testing it as soon as I can.

ivan-mogilko:
I found and fixed few mistakes, so the download link above is updated.

But the problem remaining is that Left Mouse button has code of 0, which makes it undistinguishable from "timeout" return code.

ericoporto:
Did a minor investigation on Mouse Buttons:

- `Engine/device/mousew32.cpp` 
  - *defines the mouse buttons*
  
- `Engine/ac/mouse.h` 
  - *defines the mouse buttons again*
  
- `Engine/ac/mouse.cpp` 
  - *relies on mouse buttons numeric values, and also uses values directly on an array! Zero being first element and -1 being out of the array at least shines a light on button number decision.*
  
- `Common/gui/guibutton.h` 
  - *defines the mouse buttons again!*
  
- `Engine/main/game_run.cpp`
  - *mathematically manipulate mouse number*
  
My conclusion is the engine relies too much on Defines where the actual Mathematical value of constants matter. If it used more Enums in this place, C++ type checking would at least help track these values.

On the API, the problem is that information has to be compressed into an int so there is no need to create and maintain a new struct type to hold this information. One solution I could think was to provide an additional function to convert this returned value to enum types of AGS (one for Mouse and other to Keyboard buttons).

ivan-mogilko:
> My conclusion is the engine relies too much on Defines where the actual Mathematical value of constants matter. If it used more Enums in this place, C++ type checking would at least help track these values.

Are you suggesting to change that in this pr?

ericoporto:
> Are you suggesting to change that in this pr?
Was more trying to figure out how hard it would be to change. My conclusion is that for this PR would be too much work, but would not be against changing in a future version.

> In theory, we could return a pointer to a struct from Wait, but... I would not like to do that because such things require designing something more universal, and preferrably using it in other similar places. Not something for 3.5.1 update.

Agreed.

Other idea I had now was `int SkipWaitKeyPressed()` (or a better name), which would be a new function to get the reason previous wait was skipped that you could use in case it was not timeout, if such behavior was desired. This would keep the wait with the current behavior of 0 for timeout and 1 for player interruption.

Alternatively, don't see a problem with going with the current suggested implementation too.

messengerbag:
> the problem remaining is that Left Mouse button has code of 0, which makes it undistinguishable from "timeout" return code.

Hmm, this does seem like a pretty significant problem. I did not expect that. I suppose if we're already negating the MouseButton values, we could do it as `-(MouseButton + 1)`, but it's ugly.

ivan-mogilko:
There's an option to modify mouse button values as a part of API update for 3.5.1, which would mean that the constants will be one thing if script is compiled as 3.5.0 and another if it's compiled as 3.5.1.
But frankly that does not seem to be worth, at least in this minor update.

ivan-mogilko:
Hmm, another alternative is to pack values in this integer differently, e.g. use first short (2 bytes) as a code of interrupt type, and second short as key/mouse code.
Then getting mouse code would be done like MouseButton btn = (result >> 16);

ivan-mogilko:
Also, @ericoporto's suggestion above does not seem terrible: return only a type of interrupt, and require to either call a function, a property, or read a game.* variable if you need to know actual mouse/key code.
Latter may be similar to `game.inv_activated`, which tells index of last clicked inventory item, or `game.used_mode`, which tells a cursor mode of the last click.

morganwillcock:
I would worry that while 0 is a valid wait input there would be a risk of script module ordering resetting the value by the time that it was read. Not that I think that 0 should be an invalid timeout value, just that there is an increased risk of accidentally reading the wrong value back where user input is not the only factor in what may be returned (which contrasts the examples of `game.inv_activated` and `game.used_mode`).

ivan-mogilko:
@morganwillcock iirc another Wait cannot be called during Wait, and if these extra variables are declared readonly then only engine will be able to set them, which in theory means as long as you check them immediately after Wait call they should be consistent.

---

@messengerbag updating one of the above proposals, Wait could return 0 for timeout and >0 for key/mouse codes packed together as 2 shorts. Because exact mouse/key code is not needed too often, for most cases user will simply compare return value as bool (==0 or !=0). If actual codes are necessary, following value extraction would apply:
`mousecode = result >> 16;
if (mousecode == 0)
     keycode = result;
`
or
`keycode = result & 65535; // because AGS iirc does not support hex`
If this was a real language I would not bother much, but have to admit that with AGS scripting this is uncommon.

ivan-mogilko:
@messengerbag Maybe reading mouse value as `mousecode = -(result + 1)` is not that bad, considering other options require either more uncommon (in ags scripting) conversions or extra variables/functions/structs.

messengerbag:
Yes, I think that would usually be the case. Occasionally, someone might want to only skip with the left or the right mouse button, but I would expect that it would be rare.

As an API-user, I have no strong opinion either way on how it should be done: both the options presented (or even redefining the values of the `MouseButton` enum) would serve all needs I can think of, and would be approximately equally easy to work with.

ivan-mogilko:
Ok, I changed how the mouse code is returned. It will be ready for testing here:
https://cirrus-ci.com/task/6051200439156736

ivan-mogilko:
Perhaps someone else could test this a little? In case I missed something.

ivan-mogilko:
@messengerbag out of the sudden, while reading engine code, I realized that did a mistake with the mouse button return value.
If you remember, that was an annoying surprise that mouse button code starts with left=0, and that's why I had to add this ugly remapping `-(MouseButton + 1)`.
Now I found out that this was only **internal** value in the engine. If you check the script definitions here: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Resources/agsdefns.sh#L87
the mouse buttons actually start with 1. I found that they are transformed into internal codes for some obscure reason (e.g.: https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/mouse.cpp#L310).
But because `Wait*` are script functions, they must also return script constants, ones that start with 1.

This is pretty silly, wonder how I tested this. Good thing that this was not yet released as an official public version.

Issue #1019, opened 2019-12-23
Feature Suggestion: Engine writes Alert messages to stdout on Windows


ericoporto:
On the Linux Engine, whenever something goes wrong, Abort is called, or sound is not working, the Engine prints messages to the command line.

On Windows, instead we get message boxes. I would like to be possible to have the Engine output to stdout on Windows like it does on Linux, when I launch the Engine from command line on Windows (like `cmd.exe` or powershell). This could be done with an additional flag like `--no-message-box`.

This is because I would like to run windows built games on a Continuous Integration system, and right now, if something is wrong, they spawn a message box, hanging the pipeline.

morganwillcock:
It is probably easiest to get the engine to attempt to attach to the console that it is launched from (if it exists), and set a flag to not use message boxes if this succeeds, rather than add command line arguments.

ericoporto:
I thought about that, but searching on Google, I couldn't find any hint on how to do this on Windows. `Attach` keyword on search engine seems to bring up topics on how to debug things with an IDE.

morganwillcock:
...although something like a CI build probably captures the output directly, rather than launch via a console/shell. So perhaps you do need an argument like `--no-message-box` regardless of whether a console gets attached or not.

morganwillcock:
> I thought about that, but searching on Google, I couldn't find any hint on how to do this on Windows. Attach keyword on search engine seems to bring up topics on how to debug things with an IDE.

It would be this one, I think:
https://docs.microsoft.com/en-us/windows/console/attachconsole

jduo:
I have a fix for this problem in this PR: https://github.com/adventuregamestudio/ags/pull/1088

PR #1088, opened 2020-06-15
Windows: Add option to disable message boxes

jduo:
Add the --no-message-box option in Windows for showing messages on stdout instead of
a message box.

This required changing the SubSystem from Windows to Console, and writing a simple
main() function that calls WinMain(). By making the application a "Console" app,
stdout/stderr handlers properly get initialized now.

This also makes the Windows port behavior more consistent with other platforms:

Startup information now gets shown.
The help menu now appears if there's an error.
The --tell parameter works.

### DIFF
--- a/Engine/main/main.cpp
+++ b/Engine/main/main.cpp
@@ -84,6 +84,8 @@ bool justRunSetup = false;
 bool justRegisterGame = false;
 bool justUnRegisterGame = false;
 bool justTellInfo = false;
+bool attachToParentConsole = false;
+bool hideMessageBoxes = false;
 std::set<String> tellInfoKeys;
 const char *loadSaveGameOnStartup = nullptr;
 
@@ -172,8 +174,11 @@ extern char return_to_room[150];
 
 void main_print_help() {
     platform->WriteStdOut(
-           "Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]\n\n"
+        "Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]\n\n"
            "Options:\n"
+#if AGS_PLATFORM_OS_WINDOWS
+           "  --console-attach             Write output to the parent process's console\n"
+#endif
            "  --fps                        Display fps counter\n"
            "  --fullscreen                 Force display mode to fullscreen\n"
            "  --gfxdriver <id>             Request graphics driver. Available options:\n"
@@ -192,6 +197,7 @@ void main_print_help() {
            "  --no-log                     Disable program output to the log file,\n"
            "                                 overriding configuration file setting\n"
 #if AGS_PLATFORM_OS_WINDOWS
+           "  --no-message-box              Disable reporting of alerts to message boxes\n"
            "  --setup                      Run setup application\n"
 #endif
            "  --tell                       Print various information concerning engine\n"
@@ -328,6 +334,8 @@ static int main_process_cmdline(ConfigTree &cfg, int argc, char *argv[])
         // Special case: data file location
         //
         else if (arg[0]!='-') datafile_argv=ee;
+        else if (ags_stricmp(arg, "--console-attach") == 0) attachToParentConsole = true;
+        else if (ags_stricmp(arg, "--no-message-box") == 0) hideMessageBoxes = true;
     }
 
     if (datafile_argv > 0)
@@ -414,6 +422,9 @@ int ags_entry_point(int argc, char *argv[]) {
     if (res != 0)
         return res;
 
+    if (attachToParentConsole)
+        platform->AttachToParentConsole();
+
     if (justDisplayVersion)
     {
         platform->WriteStdOut(get_engine_string());
@@ -426,8 +437,9 @@ int ags_entry_point(int argc, char *argv[]) {
         return EXIT_NORMAL;
     }
 
-    if (!justTellInfo)
+    if (!justTellInfo && !hideMessageBoxes)
         platform->SetGUIMode(true);
+
     init_debug(justTellInfo);
     Debug::Printf(kDbgMsg_Init, get_engine_string());
 
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -49,6 +49,7 @@ AGSPlatformDriver *platform = nullptr;
 
 void AGSPlatformDriver::AboutToQuitGame() { }
 void AGSPlatformDriver::PostAllegroInit(bool windowed) { }
+void AGSPlatformDriver::AttachToParentConsole() { }
 void AGSPlatformDriver::DisplaySwitchOut() { }
 void AGSPlatformDriver::DisplaySwitchIn() { }
 void AGSPlatformDriver::PauseApplication() { }
--- a/Engine/platform/base/agsplatformdriver.h
+++ b/Engine/platform/base/agsplatformdriver.h
@@ -57,6 +57,7 @@ struct AGSPlatformDriver
     virtual void AboutToQuitGame();
     virtual void Delay(int millis);
     virtual void DisplayAlert(const char*, ...) = 0;
+    virtual void AttachToParentConsole();
     virtual int  GetLastSystemError() { return errno; }
     // Get root directory for storing per-game shared data
     virtual const char *GetAllUsersDataDirectory() { return "."; }
--- a/Engine/platform/windows/acplwin.cpp
+++ b/Engine/platform/windows/acplwin.cpp
@@ -95,9 +95,11 @@ extern SetupReturnValue acwsetup(const ConfigTree &cfg_in, ConfigTree &cfg_out, 
 
 struct AGSWin32 : AGSPlatformDriver {
   AGSWin32();
+  ~AGSWin32();
 
   virtual void AboutToQuitGame();
   virtual int  CDPlayerCommand(int cmdd, int datt);
+  virtual void AttachToParentConsole();
   virtual void DisplayAlert(const char*, ...);
   virtual int  GetLastSystemError();
   virtual const char *GetAllUsersDataDirectory();
@@ -155,10 +157,21 @@ private:
 
   bool _isDebuggerPresent; // indicates if the win app is running in the context of a debugger
   DisplayMode _preFullscreenMode; // saved display mode before switching system to fullscreen
+  bool _isAttachedToParentConsole; // indicates if the win app is attached to the parent console
 };
 
-AGSWin32::AGSWin32() {
-  _isDebuggerPresent = ::IsDebuggerPresent() != FALSE;
+AGSWin32::AGSWin32() :
+    _isDebuggerPresent(::IsDebuggerPresent() != FALSE),
+    _isAttachedToParentConsole(false)
+{
+    // Do nothing.
+}
+
+AGSWin32::~AGSWin32() {
+    if (_isAttachedToParentConsole)
+    {
+        ::FreeConsole();
+    }
 }
 
 void check_parental_controls() {
@@ -821,6 +834,33 @@ int AGSWin32::CDPlayerCommand(int cmdd, int datt) {
 #endif
 }
 
+void AGSWin32::AttachToParentConsole() {
+    if (_isAttachedToParentConsole)
+        return;
+
+    _isAttachedToParentConsole = ::AttachConsole(ATTACH_PARENT_PROCESS) != FALSE;
+    if (_isAttachedToParentConsole)
+    {
+        // Require that both STDOUT and STDERR are valid handles from the parent process.
+        if (::GetStdHandle(STD_OUTPUT_HANDLE) != INVALID_HANDLE_VALUE &&
+            ::GetStdHandle(STD_ERROR_HANDLE) != INVALID_HANDLE_VALUE)
+        {
+            // Re-open STDOUT and STDERR to the parent's.
+            FILE* fp = NULL;
+            freopen_s(&fp, "CONOUT$", "w", stdout);
+            setvbuf(stdout, NULL, _IONBF, 0);
+
+            freopen_s(&fp, "CONOUT$", "w", stderr);
+            setvbuf(stderr, NULL, _IONBF, 0);
+        }
+        else
+        {
+            ::FreeConsole();
+            _isAttachedToParentConsole = false;
+        }
+    }
+}
+
 void AGSWin32::DisplayAlert(const char *text, ...) {
   char displbuf[2500];
   va_list ap;
@@ -829,7 +869,9 @@ void AGSWin32::DisplayAlert(const char *text, ...) {
   va_end(ap);
   if (_guiMode)
     MessageBox(win_get_window(), displbuf, "Adventure Game Studio", MB_OK | MB_ICONEXCLAMATION);
-  else if (_logToStdErr)
+
+  // Always write to either stderr or stdout, even if message boxes are enabled.
+  if (_logToStdErr)
     AGSWin32::WriteStdErr("%s", displbuf);
   else
     AGSWin32::WriteStdOut("%s", displbuf);
@@ -1103,6 +1145,5 @@ LPDIRECTINPUTDEVICE IAGSEngine::GetDirectInputKeyboard() {
 LPDIRECTINPUTDEVICE IAGSEngine::GetDirectInputMouse() {
   return mouse_dinput_device;
 }
-
 
 #endif


ivan-mogilko:
Changing application type to "Console" makes a console window open along with the main one regardless of whether user asked for it or not. I don't think that's an optimal solution, and can already predict game developers complaining about this.

If I understand correctly, in #1019 it was suggested to try to programmatically create and attach console to the running process instead.

ericoporto:
When I ran the resulting `acwin.exe`, the thing that caught my attention is how the mouse position in the game window became wrong - can't quite figure out how to move the mouse.

jduo:
I initially went the route of using AttachConsole with attach_to_parent and got error -6 - invalid handle. I also tried using AllocConsole which appeared to work, but no output ever appeared. I think I know what's needed to fix that approach though.

So I think attaching to the parent doesn't work, which necessitates the parameter. Ideally we could create/attach to the console when setGuiEnabled is true as a callback, but it's static.

Can you guys point me to some sample input to acwin? I'll take another stab

ericoporto:
Any game built with ags 3.4.0 or above here-> https://www.adventuregamestudio.co.uk/forums/index.php?board=5.0
Here's one I made -> https://eri0o.itch.io/dungeon-hands
Just place `acwin.exe` in the same directory of the gameName.exe and run `acwin.exe` instead of the gameName.exe.

morganwillcock:
I thought that the attach would be the easy part, perhaps the Windows version is set too low as apparently this function only works on XP or newer.

```#define _WIN32_WINNT 0x0501```

**Comment by jduo** [comment]
I've updated the PR to use AttachConsole. It turns out that was succeeding when I ran from cmd.exe, but when I had launched from VS's debugger it would fail. There's a bit more work involved in actually getting the console to display the output which I've added to the constructor..

**Comment by ericoporto** [comment]
This appears to work for me. 
- the default behavior available when double clicking the exe doesn't appear to have changed, which is good;
- I noticed that after it's run, I need to press enter (return) to get back to the prompt line, not sure if it's expected behavior;
- It appears I can't redirect the output to a file (tried `.\acwin.exe > log.txt`)

Looking the changes [here (ags/pull/1088/files)](https://github.com/adventuregamestudio/ags/pull/1088/files), there's an unneeded new line committed to `Engine/main/main.cpp`. 

If possible, also squash the five commits here into one. One way to do this is the following
```
git rebase -i HEAD~5
```
- picking the commit at top and squashing the ones below, in your checked branch for this PR and then force pushing so it updates here in this PR.

**Comment by jduo** [comment]
@ericoporto , I've addressed the unneeded newline in main.cpp and squashed the commits. I haven't figured out a solution to redirecting output and am not sure if this is possible.

The extra enter keypress is actually not necessary. You can actually start running commands without hitting enter, just the prompt isn't visible yet.

**Comment by ericoporto** [comment]
I think I found a particular solution in this answer of stackoverflow:
https://stackoverflow.com/questions/11523595/attachconsole-shows-data-on-pipe-but-the-operator-doesnt-correctly-redirect

There's this interesting function created to help solve the particular case:

```
private static bool IsRedirected(IntPtr handle) { 
  FileType fileType = GetFileType(handle); 
  return (fileType == FileType.Disk) || (fileType == FileType.Pipe); 
}
```

Apparently it would be possible to figure out the handle type and only go ahead with the code to put text stuff in stdout and stderr if handle is not file or pipe.

Here are Windows docs on the C API to get the file type
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype?redirectedfrom=MSDN

jduo:
> I think I found a particular solution in this answer of stackoverflow:
> https://stackoverflow.com/questions/11523595/attachconsole-shows-data-on-pipe-but-the-operator-doesnt-correctly-redirect
> 
> There's this interesting function created to help solve the particular case:
> 
> ```
> private static bool IsRedirected(IntPtr handle) { 
>   FileType fileType = GetFileType(handle); 
>   return (fileType == FileType.Disk) || (fileType == FileType.Pipe); 
> }
> ```
> 
> This function in the SO answer is used once in stderr handle and other in the stdout handle, after attaching to the console. Apparently it would be possible to figure out the handle type and only go ahead with the code to put text stuff in stdout / stderr if it's handle is not file or pipe.
> 
> Here are Windows docs on the C API to get the file type
> https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype?redirectedfrom=MSDN
> 
> It appears to provide `FILE_TYPE_DISK` and `FILE_TYPE_PIPE` for comparison.

This suggestion worked. I've updated the PR to support redirection of STDOUT and 
STDERR to files. Basically we don't try to re-open these handles under the parent console's 
handles if the user is already redirecting them.

ericoporto:
Sorry, haven't had time to check the actual code or build yet.

jduo:
Hey @ericoporto , do you have an idea on when you'd be able to check this out? I would be interested in taking on other BountySource tickets on this project but would want to resolve them one-at-a-time.

morganwillcock:
I've only tried it very quickly, but it looks like what should be going to stderr is going to stdout.
<pre>
acwin.exe > %TEMP%\stdout 2> %TEMP%\stderr
</pre>
Where there is no game data the stderr output is empty and stdout shows the search path twice. I would guess that the value of `_logToStdErr` is set or reset as part of the platform driver, but for Windows it is set to false.

ericoporto:
It's better to not rebase and squash things into a single commit until it's certain that the pr is going to be accepted. The previous commit was more concise, I am not sure what changed now. I know it was I who instructed wrong, it was just because I was unsure if the problem of showing/redirection was fixable at the time - Windows is a mystery to me.

ivan-mogilko:
Did not check or test this all, only wanted to note that I have doubts about this condition in DisplayAlert:

<pre>
if (_guiMode && !_isRunFromConsole)
    MessageBox(win_get_window(), displbuf, "Adventure Game Studio", MB_OK | MB_ICONEXCLAMATION);
</pre>

As running from console does not necessarily mean "don't display dialog messages".

morganwillcock:
> As running from console does not necessarily mean "don't display dialog messages".

I think that `_isRunFromConsole` is actually meaning 'are existing handles already redirected or I've succesfully attached to my parent', so not neccesairly about where the process was started, but this is was gives the automatic switching and means there is no reliance on command line options. Is it not OK for the same messages to just appear on the console or get redirected to where the user has requested to capture it?

ivan-mogilko:
Perhaps the variable should be called differently to indicate its real meaning.
Also I'd propose to have single variable controlling that condition. If `_isRunFromConsole` efficiently cancels `_guiMode`, then you could simply set `_guiMode = false` at startup in case output is redirected.

morganwillcock:
The main thing I am not sure of here is the check that existing handles are already redirected to opt out of attaching to the parent. I think that before this PR the redirection for stdout was working OK without any issue, it was just missing the console attach to get output that wasn't redirected. With this PR there are now situations where explicit redirections are now ignored (I guess there is no match against FILE_TYPE_DISK or FILE_TYPE_PIPE).

As an example, the PR breaks:
<pre>acwin.exe > nul</pre>

So my train of thought at the moment is to skip the checks on existing handles and always try to attach (and if attachment succeeds that both stdout and stderr handles are not INVALID_HANDLE_VALUE), as I think this is the only way to not accidentally override an explicit redirection.

ivan-mogilko:
I am yet planning to investigate this myself, but it also seem strange to me that program itself has to test for redirected output. I'd imagine this is what system that runs program should be doing, and not program itself.

jduo:
> Perhaps the variable should be called differently to indicate its real meaning.
> Also I'd propose to have single variable controlling that condition. If `_isRunFromConsole` effectively cancels `_guiMode`, then you could simply set `_guiMode = false` at startup in case output is redirected.

The reason I did not do this is that there is no callback on the driver invoked _after_ command-line arguments are evaluated. This is compounded by the fact that SetGUIMode is static. We don't know when someone has changed the state of GUI mode.

The constructor was a reliable place to put the code to platform-specific code, such as auto-detection of the console.

jduo:
> > As running from console does not necessarily mean "don't display dialog messages".
> 
> I think that `_isRunFromConsole` is actually meaning 'are existing handles already redirected or I've successfully attached to my parent', so not neccesairly about where the process was started, but this is what gives the automatic switching and means there is no reliance on command line options. Is it not OK for the same messages to just appear on the console or get redirected to where the user has requested to capture it?

I don't think you want this though. One of the use cases for this was to plug into a CI system, but having pop-ups would break this.

ivan-mogilko:
> The reason I did not do this is that there is no callback on the driver invoked after command-line arguments are evaluated. 

Yes, it looks like the callbacks might need to be enhanced in the future.

> I don't think you want this though. One of the use cases for this was to plug into a CI system, but having pop-ups would break this.

Imho there are two different things here: directing output somewhere and disabling popups. In theory these options may be used in various combinations.

ivan-mogilko:
> This is compounded by the fact that SetGUIMode is static. We don't know when someone has changed the state of GUI mode.

Errrmmmmmm..... yes, this is weird, and I don't remember why it was made so...
Alright, guess this is a separate problem that could be dealt with later.

jduo:
> > The reason I did not do this is that there is no callback on the driver invoked after command-line arguments are evaluated.
> 
> Yes, it looks like the callbacks might need to be enhanced in the future.
> 
> > I don't think you want this though. One of the use cases for this was to plug into a CI system, but having pop-ups would break this.
> 
> Imho there are two different things here: directing output somewhere and disabling popups. In theory these options may be used in various combinations.

In the first PR we added the --no-message-box option, which would have been for disabling pop-ups and writing to console, but we wanted to remove that option. Are you saying we want something like this now, but change DisplayAlert() to always write to console and suppress the message box based on the flag?

ivan-mogilko:
Hm, ok, sorry, it looks like everyone has different view on this. I will re-read all the thread from start to see what were the people's arguments.

jduo:
> The main thing I am not sure of here is the check that existing handles are already redirected to opt out of attaching to the parent. I think that before this PR the redirection for stdout was working OK without any issue, it was just missing the console attach to get output that wasn't redirected. With this PR there are now situations where explicit redirections are now ignored (I guess there is no match against FILE_TYPE_DISK or FILE_TYPE_PIPE).
> 
> As an example, the PR breaks:
> 
> acwin.exe > nul
> 
> So my train of thought at the moment is to skip the checks on existing handles and always try to attach (and if attachment succeeds, check that both stdout and stderr handles are not INVALID_HANDLE_VALUE), as I think this is the only way to not accidentally override an explicit redirection.

I don't know if the suggested logic works. When you call AttachConsole, you get the parent's stdout/stderr handles. If your parent process isn't redirecting these, your process won't be redirecting these.

To clarify, when you launch acwin.exe > foo.txt, you are, from your parent console which possibly isn't redirecting anything, launching a new process with its stdout configured to point to foo.txt. When you AttachConsole, you're basically discarding what the handles that were automatically configured for your app.

I'd recommend keeping the logic the same, but just figuring out what type NUL is. I'm guessing it's FILE_TYPE_UNKNOWN.

jduo:
> I've only tried it very quickly, but it looks like what should be going to stderr is going to stdout.
> 
> acwin.exe > %TEMP%\stdout 2> %TEMP%\stderr
> 
> Where there is no game data the stderr output is empty and stdout shows the search path twice. I would guess that the value of `_logToStdErr` is set or reset as part of the platform driver, but for Windows it is set to false.

The handling of what goes into stderr vs. what goes into stdout on Windows seems orthogonal to the problem of letting you suppress message boxes. Would it make more sense to make this a separate issue?

If not, can we document some more examples of what should go into stderr vs. stdout to avoid back-and-forth?

ivan-mogilko:
Currently AGS only writes to stderr when justTellInfo is true. That is - when it is required to print certain game etc information to stdout (hence errors are written to stderr in order to not mix up with this printout).

ivan-mogilko:
So to summarize, the tasks were:
* attaching console to a program that was ran from one; if I understand correctly, this is decided to be done unconditionally?
* ensure that when AGS writes to stdout or stderr the data actually gets where expected;
* ensure that when output is redirected by user command then AGS's output actually ends up there;

A separate issue, but also needed to cover for the ticket to be closed is: decide how popups are disabled. Alternatives here are:
* implicitly disable if ran from console;
* implicitly disable if ran from console and output is redirected by user;
* explicitly disabled by command line argument.

Everything else, including logic of when ags should write to stdout and when to stderr, would be separate issues.

jduo:
> So to summarize, the tasks were:
> 
>     * attaching console to a program that was ran from one; if I understand correctly, this is decided to be done unconditionally?
> 
>     * ensure that when AGS writes to stdout or stderr the data actually gets where expected;
> 
>     * ensure that when output is redirected by user command then AGS's output actually ends up there;
> 
> 
> A distinct issue, but also needed to cover for the ticket to be closed is: decide how popups are disabled. Alternatives here are:
> 
>     * implicitly disable if ran from console;
> 
>     * implicitly disable if ran from console and output is redirected by user;
> 
>     * explicitly disabled by command line argument.
> 
> 
> TBH, personally I'd propose the 3rd variant.
> 
> Everything else, including logic of when ags should write to stdout and when to stderr, are separate matter.

Thanks, I was in the middle of writing a long response that essentially asked the same thing.

FYI, No more code changes until everyone's in agreement. If we can't come to a consensus by Tuesday at the latest on this then I'd be concerned that we have too many cooks here and would consider dropping the issue.

My responses:
- Let's always attach to the parent console if there is one.
- We'll change DisplayAlert() such that it writes its output to stdout/stderr always. It will conditionally show a pop-up based on if message boxes are suppressed (--no-message-box)

PS: The amount of code changes to do this is about inline for what I expected from the bounty when I signed up. However I've burned much more time on this than anticipated on PR responses. Would greatly appreciate if the bounty on this could be raised).

ivan-mogilko:
> Let's always attach to the parent console if there is one.
> We'll change DisplayAlert() such that it writes its output to stdout/stderr always. It will conditionally show a pop-up based on if message boxes are suppressed (--no-message-box)

I agree to this.

ivan-mogilko:
> The amount of code changes to do this is about inline for what I expected from the bounty when I signed up. However I've burned much more time on this than anticipated on PR responses. Would greatly appreciate if the bounty on this could be raised).

I must admit, this is first time I hear about this "bounty". Is this some new github thing?

morganwillcock:
> Let's always attach to the parent console if there is one.

Based on the problem which was described above, doesn't this mean that explicit redirection such as `acwin > file` will never work?

ivan-mogilko:
I thought above means "attach to console unconditionally"? Maybe I was wrong.

jduo:
> > Let's always attach to the parent console if there is one.
> 
> Based on the problem which was described above, doesn't this mean that explicit redirection such as `acwin > file` will never work?

We can, before attach, identify what handles are on stdout/stderr, then after attach, reset those if they are pointing to a file/pipe/NUL.

jduo:
> > The amount of code changes to do this is about inline for what I expected from the bounty when I signed up. However I've burned much more time on this than anticipated on PR responses. Would greatly appreciate if the bounty on this could be raised).
> 
> I must admit, this is first time I hear about this "bounty". Is this a github feature, and who sets them up?
> 
> PS. Is it possible to see the full list of the tickets that have this "bounty"? I'd like to check them out and ensure that they are explained properly.

There's a site, www.bountysource.com that lets users post, fund, and commit to bounties.

@ericoporto made the request for this one here. https://www.bountysource.com/issues/85921247-feature-suggestion-engine-writes-alert-messages-to-stdout-on-windows.

The bounties are directly related to Issues in the repository.

This is the first bounty I've taken on from that site. There are a few listed for AGS and I'd look into taking others on if we can close this in a timely manner.

jduo:
> > > Let's always attach to the parent console if there is one.
> > 
> > Based on the problem which was described above, doesn't this mean that explicit redirection such as `acwin > file` will never work?
> 
> We can, before attach, identify what handles are on stdout/stderr, then after attach, reset those if they are pointing to a file/pipe/NUL.

Just to clarify your one a little more:
in the existing PR, we already have the logic to do what's described above. Right now there's an optimization to skip attaching to console if all handles are detected to be redirections but what I am suggesting here is to skip this optimization and attach anyways for simplicity's sake.

What's missing is detecting redirection to NUL, which I wouldn't expect to be a complicated fix.

ivan-mogilko:
> This is the first bounty I've taken on from that site. There are a few listed for AGS and I'd look into taking others on if we can close this in a timely manner.

Then I just want to mention, in order to be honest, this project may not be the best to take on by a newcomer, because it has a lot of old code and hidden issues due to its history, and not all things are organized well. It was mostly developed as a hobby project for a while with little contribution from outside, and many tickets are written in sort of "notes to self" way with a thought to come back to them later.

It was very ... unthoughtfull to put bounty on the issue that has no explicit task description.

It's important to make certain that task is clearly stated before doing anything. I will try to find which other issues here have bounties, and clarify them if necessary.

jduo:
> > This is the first bounty I've taken on from that site. There are a few listed for AGS and I'd look into taking others on if we can close this in a timely manner.
> 
> Then I just want to mention, in order to be honest, this project may not be the best to take on by a newcomer, because it has a lot of old code and hidden issues due to its history, and not all things are organized well. It was mostly developed as a hobby project for a while with little contribution from outside, and many tickets are written in sort of "notes to self" way with a thought to come back to them later.
> 
> It's important to make certain that task is clearly stated before doing anything. I will try to find which other issues here have bounties, and clarify them if necessary.

Thanks. I'm a newcomer to doing bounties, but have a lot of experience dealing with legacy codebases, including  the Win32 API. If we can clean up the issues to have more clear requirements, such as the planned inputs/use cases, and examples of how to validate the result I think we can close these off much more efficiently.

@morganwillcock , @ericoporto , do you agree with my/@ivan-mogilko 's suggestions about how this issue should be solved?

ivan-mogilko:
@jduo we discussed this a little with @morganwillcock and he said he will try to answer soon.

Meanwhile, I did a small investigation and found what *might* be a better test for redirected output:
http://archives.miloush.net/michkap/archive/2010/05/07/10008232.html
If you look for `IsConsoleRedirected` function there. It works other way: uses condition which combines FILE_TYPE_CHAR (which is true for consoles) and GetConsoleMode to make sure that's indeed a console.

jduo:
> @jduo we discussed this a little with @morganwillcock and he said he will try to answer soon.
> 
> Meanwhile, I did a small investigation and found what *might* be a better test for redirected output:
> http://archives.miloush.net/michkap/archive/2010/05/07/10008232.html
> If you look for `IsConsoleRedirected` function there. It works other way: uses condition which combines FILE_TYPE_CHAR (which is true for consoles) and [GetConsoleMode](https://docs.microsoft.com/en-us/windows/console/getconsolemode) to make sure that's indeed a console.

Let's see if this helps with redirecting to null.

Fyi I plan on validating these use cases:
- show in console
- show in console + alerts
- redirect to file
- show in vs debugger
- possibly pipe to another command

Let's bring up any other specific use cases before we start implementing.

ivan-mogilko:
I know that @morganwillcock was planning on posting a comment today, but idk of what his timezone is, so I decided to comment on our recent discussion we had elsewhere. Apparently some of his tests have shown that under certain conditions both known variants of "IsRedirected" still display ambigious results. IIRC the variant I posted above failed when he was running program under msys2 console.

I also better understood his position after this. Initially I hoped there is an unambigious way to detect redirection, but judging by these tests results it may not be the case. Testing individual types of handles does not seem an optimal method to me either, as we may simply not know or miss some cases, and may have to update the list of conditions as more are found by users. Maybe even comply to some specific shell/terminal rules.

On another hand, having an explicit switch (e.g. suggested `--console-attach`) gives an unambiguity and guarantees that program follows user's request rather than its own judgement which could be wrong. So in the end I am leaning to agree with @morganwillcock on this.

There's another hypothetical reason why having such cmd line option may be beneficial: if in the future we decide to be able to create a console if one's not present, then we will already have a switch that makes this happen (only need to extend its function).

jduo:
I'm ok with this. @ericoporto please confirm.

morganwillcock:
@jduo firstly, apologies for the conflicting information which appeared early on. The initial ticket didn't have enough information and it was only when testing things and doing more research that new issues came to light. I've tried various combinations for various use cases and have still concluded that checking the redirected targets is likely to cause future issues. Increasing I think we will have to run the engine in environments that aren't a full traditional copy of Windows, so relying on specifics of built-in console would probably cause some issues. Similarly it would seem a mistake to hard code switching logic directly against one specific environment. As an example, attaching to the console actually worked fine in the msys2 terminal, but results of testing handles or trying to identify a console gave different results. So I feel the best option is to always follow the requested (or inferred by omission) targets for output, unless the user specifically requests to attempt to attach to the console and the result is usable. This also ensures that most predictable behaviour is the default. For fallback behaviour where, for whatever reason, the attach process has succeeded but the parent's stdout or sdterr handles are returned as invalid, just detaching at that point doesn't seem to intefere with any existing redirection.
<hr>
Without request for console attach maintain existing redirection regardless of target type

i.e. for the following command:

```acwin > target1 2> target2```

- stdout always goes to target1
- stderr always goes to target2
<hr>
When console attach is requested yield redirection to parent if possible
  
i.e. for the following command:

```acwin --console-attach > target1 2> target2```

- stdout and stderr are redirected to the parent's handles if console attach is successful and the parent's stdout and stderr handles are not invalid. Both handles are to be attached if all three conditions are true, regardless of whether the child handles had been redirected or not. Otherwise if any of these three conditions are not met return to previous behaviour and use the specified targets
<hr>

Then, completely independently, another option such as `--no-message-boxes` will prevent messages boxes appearing. I am told that what appears in the message boxes is already going to stdout (and sometimes stderr) so hopefully this is just a simple flag to skip the message box function.

jduo:
> @jduo firstly, apologies for the conflicting information which appeared early on. The initial ticket didn't have enough information and it was only when testing things and doing more research that new issues came to light. I've tried various combinations for various use cases and have still concluded that checking the redirected targets is likely to cause future issues. Increasing I think we will have to run the engine in environments that aren't a full traditional copy of Windows, so relying on specifics of built-in console would probably cause some issues. Similarly it would seem a mistake to hard code switching logic directly against one specific environment. As an example, attaching to the console actually worked fine in the msys2 terminal, but results of testing handles or trying to identify a console gave different results. So I feel the best option is to always follow the requested (or inferred by omission) targets for output, unless the user specifically requests to attempt to attach to the console and the result is usable. This also ensures that most predictable behaviour is the default. For fallback behaviour where, for whatever reason, the attach process has succeeded but the parent's stdout or sdterr handles are returned as invalid, just detaching at that point doesn't seem to intefere with any existing redirection.
> Without request for console attach maintain existing redirection regardless of target type
> 
> i.e. for the following command:
> 
> `acwin > target1 2> target2`
> 
>     * stdout always goes to target1
> 
>     * stderr always goes to target2
> 
> 
> When console attach is requested yield redirection to parent if possible
> 
> i.e. for the following command:
> 
> `acwin --console-attach > target1 2> target2`
> 
>     * stdout and stderr are redirected to the parent's handles if console attach is successful and the parent's stdout and stderr handles are not invalid. Both handles are to be attached if all three conditions are true, regardless of whether the child handles had been redirected or not. Otherwise if any of these three conditions are not met return to previous behaviour and use the specified targets
> 
> 
> Then, completely independently, another option such as `--no-message-boxes` will prevent messages boxes appearing. I am told that what appears in the message boxes is already going to stdout (and sometimes stderr) so hopefully this is just a simple flag to skip the message box function.

Sure, the reasoning you have is sound. I think we're in alignment here.
Note that @ivan-mogilko mentioned the message box option should be "--no-message-box" and you've mentioned it should be "--no-message-boxes". I'm going to assume it should be "--no-message-box" since other command line arguments don't use plurals.

ericoporto:
@jduo I am ok with the set plan!

jduo:
I believe I've implemented what was suggested now. I had to add a new method to AGSPlatformDriver that is a no-op for everything but Windows which seems less than ideal.

ivan-mogilko:
@jduo please rebase/remake your pr branch, right now it includes all the commits with the previous and reverted changes, making it difficult to understand what is going to be merged.

jduo:
Whoops, forgot to push the final commit.
I rebased/squashed the commits to one now.

ivan-mogilko:
Thank you.
I think the code looks fine overall.

`isRedirected` function is still there but unused.
I think for correctness AttachToParentConsole could test if console is already attached and exit. But calling twice does not seem to break anything.

I tested few common uses so far and they work:
- running game from explorer (no changes noticed)
- running game from console without `--console-attach` (no changes)
- running game from console with `--console-attach` (log outputs to console)
- running with `--console-attach --tell` (information outputs to console)
- running from console with --console-attach and pressing Ctrl+C will abort the program :)
- running from MSVS: log outputs to IDE's console regardless of options.

There's slightly annoying problem that console does not show input line after program exit (@ericoporto mentioned it already), but idk if it's something you can fix in program.

jduo:
> Thank you.
> I think the code looks fine overall.
> 
> `isRedirected` function is still there but unused.
> I think for correctness AttachToParentConsole could test if console is already attached and return if it is. But calling twice does not seem to break anything.
> 
> I tested few common uses so far and they work:
> 
>     * running game from explorer (no changes noticed)
> 
>     * running game from console without `--console-attach` (no changes)
> 
>     * running game from console with `--console-attach` (log outputs to console)
> 
>     * running with `--console-attach --tell` (information outputs to console)
> 
>     * running from console with --console-attach and pressing Ctrl+C will abort the program :)
> 
>     * running from MSVS: log outputs to IDE's console regardless of options.
> 
> 
> There's slightly annoying problem that console does not show input line after program exit (@ericoporto mentioned it already), but idk if it's something you can fix in program.

Yeah I don't think the missing input line after exit is fixable. I'll make the two other changes (deleting isRedirected and making AttachToParentConsole idempodent).

ivan-mogilko:
There's an error in the code now, AttachToParentConsole is void but returns false.

morganwillcock:
Also the usage message for `--no-message-box` is misaligned:

<pre>
  --no-log                     Disable program output to the log file,
                                 overriding configuration file setting
  --no-message-box              Disable reporting of alerts to message boxes
  --setup                      Run setup application
</pre>

jduo:
> Also the usage message for `--no-message-box` is misaligned:
> 
>   --no-log                     Disable program output to the log file,
>                                  overriding configuration file setting
>   --no-message-box              Disable reporting of alerts to message boxes
>   --setup                      Run setup application

Corrected, along with the build issue.

ivan-mogilko:
I think this can be merged, unless someone found another error?

ericoporto:
I did some tests with `--tell` and aborting in the middle of the game and using the `--console-attach` and `--no-message-box` flags and it appears to work on my end.

PR #1089, opened 2020-06-15
adds mouse pan with middle click on Room panel

ericoporto:
Creates FreePanControl.cs
- replaces BufferedPanel in Room Editor with FreePanControl
- adds pan with middle mouse click hold to ags editor
- also accepts left click+shift+ctrl 
- adds simple isolated tests for FreePanControl in `AGS.Editor.Tests` to document expected behavior.

**[CirrusCi builds of this PR](https://cirrus-ci.com/github/adventuregamestudio/ags/pull/1089)**
This is a rewrite of #1046
Attempt to solve #613 , [agsforums-57490](https://www.adventuregamestudio.co.uk/forums/index.php?topic=57490.0)

### DIFF
--- a/Editor/AGS.Editor.Tests/Panes/FreePanControlTests.cs
+++ b/Editor/AGS.Editor.Tests/Panes/FreePanControlTests.cs
@@ -1,0 +1,71 @@
+﻿using System.Windows.Forms;
+using NUnit.Framework;
+
+namespace AGS.Editor
+{
+    [TestFixture]
+    public class FreePanControlTests
+    {
+        private FreePanControlTestableWrapper _pane;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _pane = new FreePanControlTestableWrapper();
+        }
+
+        [Test]
+        public void IsPanningDefaultsToFalse()
+        {
+            Assert.That(_pane.IsPanning, Is.False);
+        }
+
+        [Test]
+        public void IsPanningIsTrueOnMiddleMouseClick()
+        {
+            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
+            Assert.That(_pane.IsPanning, Is.True);
+        }
+
+        [Test]
+        public void PanGrabbedInvokesOnMiddleMouseClick()
+        {
+            bool panGrabbedInvoked = false;
+            _pane.PanGrabbed += (s, e) => panGrabbedInvoked = true;
+            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
+            Assert.That(panGrabbedInvoked, Is.True);
+        }
+
+        [Test]
+        public void IsPanningReturnsToFalseOnPan()
+        {
+            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
+            Assert.That(_pane.IsPanning, Is.True);
+            _pane.OnMouseMove(new MouseEventArgs(MouseButtons.Middle, 1, 4, 4, 0));
+            Assert.That(_pane.IsPanning, Is.True);
+            _pane.OnMouseUp(new MouseEventArgs(MouseButtons.Middle, 1, 3, 3, 0));
+            Assert.That(_pane.IsPanning, Is.False);
+        }
+
+        [Test]
+        public void CursorIsModifiedAndReturned()
+        {
+            _pane.Cursor = Cursors.IBeam;
+            Assert.That(_pane.Cursor, Is.EqualTo(Cursors.IBeam));
+            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
+            Assert.That(_pane.Cursor, Is.EqualTo(_pane.PanCursor));
+            _pane.OnMouseUp(new MouseEventArgs(MouseButtons.Middle, 1, 3, 3, 0));
+            Assert.That(_pane.Cursor, Is.EqualTo(Cursors.Default));
+            // this pane always go back to default cursor, if you want to change it to other cursor
+            // use OnPanRelease to change the cursor to the state desired.
+            // this prevents storing a "state" WaitCursor and restoring to it when it's uneeded.
+        }
+
+        private class FreePanControlTestableWrapper : FreePanControl
+        {
+            public new void OnMouseDown(MouseEventArgs e) => base.OnMouseDown(e);
+            public new void OnMouseUp(MouseEventArgs e) => base.OnMouseUp(e);
+            public new void OnMouseMove(MouseEventArgs e) => base.OnMouseMove(e);
+        }
+    }
+}
--- a/Editor/AGS.Editor/Panes/FreePanControl.cs
+++ b/Editor/AGS.Editor/Panes/FreePanControl.cs
@@ -1,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Text;
+using System.Windows.Forms;
+
+namespace AGS.Editor
+{
+    public class FreePanControl : BufferedPanel
+    {
+        private const int WS_EX_COMPOSITED = 0x02000000;
+        private int _lastX, _lastY; //used to prevent useless mouse move events
+        private Point _panGrabPoint = Point.Empty;
+        public event EventHandler PanGrabbed;
+        public event EventHandler PanReleased;
+        public bool IsPanning { get; private set; }
+        public Cursor PanCursor { get; private set; } = Cursors.Hand; // We need a better hand cursor, using the "link" hand for now
+
+        // Enabling Composite makes scrolling smoother
+        protected override CreateParams CreateParams
+        {
+            get
+            {
+                CreateParams cp = base.CreateParams;
+                cp.ExStyle |= WS_EX_COMPOSITED;
+                return cp;
+            }
+        }
+
+        protected virtual void OnPanGrab(EventArgs e)
+        {
+            PanGrabbed?.Invoke(this, EventArgs.Empty);
+        }
+
+        protected virtual void OnPanRelease(EventArgs e)
+        {
+            PanReleased?.Invoke(this, EventArgs.Empty);
+        }
+
+        protected override void OnMouseDown(MouseEventArgs e)
+        {
+            // defines the shortcut for click pan Control+Shift+Left Click Hold or
+            // Middle Click Hold without any modifier.
+            if ( (e.Button == MouseButtons.Left   && ModifierKeys == (Keys.Control|Keys.Shift)) ||
+                 (e.Button == MouseButtons.Middle && ModifierKeys == Keys.None) )
+            {
+                OnPanGrab(e);
+                _panGrabPoint = e.Location;
+                Cursor = PanCursor;
+                IsPanning = true;
+            }
+
+            base.OnMouseDown(e);
+        }
+
+        protected override void OnMouseUp(MouseEventArgs e)
+        {
+            if (_panGrabPoint != Point.Empty)
+            {
+                _panGrabPoint = Point.Empty;
+                Cursor = Cursors.Default;
+                IsPanning = false;
+                OnPanRelease(e);
+            }
+
+            base.OnMouseUp(e);
+        }
+
+        protected override void OnMouseMove(MouseEventArgs e)
+        {
+            if ((e.X == _lastX) && (e.Y == _lastY))
+            {
+                return;
+            }
+
+            _lastX = e.X;
+            _lastY = e.Y;
+
+            if (_panGrabPoint != Point.Empty)
+            {
+                Point scrollPosition = AutoScrollPosition;
+                scrollPosition.X = _panGrabPoint.X - e.X - scrollPosition.X;
+                scrollPosition.Y = _panGrabPoint.Y - e.Y - scrollPosition.Y;
+                AutoScrollPosition = scrollPosition;
+                Refresh(); // this prevents the room image to look garbled when panning
+                _panGrabPoint = e.Location; // <- without this the pan doesn't work
+            }
+
+            base.OnMouseMove(e);
+        }
+
+    }
+}
+
+
--- a/Editor/AGS.Editor.Tests/AGS.Editor.Tests.csproj
+++ b/Editor/AGS.Editor.Tests/AGS.Editor.Tests.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\..\Solutions\packages\Microsoft.NET.Test.Sdk.16.4.0\build\net40\Microsoft.NET.Test.Sdk.props" Condition="Exists('..\..\Solutions\packages\Microsoft.NET.Test.Sdk.16.4.0\build\net40\Microsoft.NET.Test.Sdk.props')" />
   <Import Project="..\..\Solutions\packages\Microsoft.CodeCoverage.16.4.0\build\netstandard1.0\Microsoft.CodeCoverage.props" Condition="Exists('..\..\Solutions\packages\Microsoft.CodeCoverage.16.4.0\build\netstandard1.0\Microsoft.CodeCoverage.props')" />
@@ -68,6 +68,7 @@
   <ItemGroup>
     <Compile Include="ColorThemes\ColorThemesTests.cs" />
     <Compile Include="ColorThemes\Controls\ComboBoxCustomTests.cs" />
+    <Compile Include="Panes\FreePanControlTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -198,6 +198,9 @@
     </Compile>
     <Compile Include="Panes\Room\RoomEditFilters\CharactersEditorFilter.cs" />
     <Compile Include="Panes\Room\RoomEditFilters\DesignTimeProperties.cs" />
+    <Compile Include="Panes\FreePanControl.cs">
+      <SubType>Component</SubType>
+    </Compile>
     <Compile Include="Plugins\IAGSEditor.cs" />
     <Compile Include="Plugins\NativePlugin.cs" />
     <Compile Include="Panes\BufferedPanel.cs">
--- a/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.Designer.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.Designer.cs
@@ -44,7 +44,7 @@ namespace AGS.Editor
             this.btnChangeImage = new System.Windows.Forms.Button();
             this.label1 = new System.Windows.Forms.Label();
             this.cmbBackgrounds = new System.Windows.Forms.ComboBox();
-            this.bufferedPanel1 = new AGS.Editor.BufferedPanel();
+            this.bufferedPanel1 = new AGS.Editor.FreePanControl();
             this.mainFrame.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)(this.sldTransparency)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.sldZoomLevel)).BeginInit();
@@ -275,7 +275,7 @@ namespace AGS.Editor
         private System.Windows.Forms.GroupBox mainFrame;
         private System.Windows.Forms.Label label1;
         private System.Windows.Forms.ComboBox cmbBackgrounds;
-        private BufferedPanel bufferedPanel1;
+        private FreePanControl bufferedPanel1;
         private System.Windows.Forms.Button btnExport;
         private System.Windows.Forms.Button btnDelete;
         private System.Windows.Forms.Button btnChangeImage;
--- a/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
@@ -33,7 +33,6 @@ namespace AGS.Editor
         private List<IRoomEditorFilter> _layers = new List<IRoomEditorFilter>();
         private CharactersEditorFilter _characterLayer; // need to store the reference for special processing
         private bool _editorConstructed = false;
-        private int _lastX, _lastY;
         private bool _mouseIsDown = false;
         private int _menuClickX;
         private int _menuClickY;
@@ -605,6 +604,7 @@ namespace AGS.Editor
 
         private void bufferedPanel1_MouseDown(object sender, MouseEventArgs e)
         {
+            if (bufferedPanel1.IsPanning) return;
             if (_layer != null && !IsLocked(_layer))
                 _layer.MouseDown(e, _state);
             _mouseIsDown = true;
@@ -648,14 +648,7 @@ namespace AGS.Editor
 
 		private void bufferedPanel1_MouseMove(object sender, MouseEventArgs e)
         {
-            if ((e.X == _lastX) && (e.Y == _lastY))
-            {
-                return;
-            }
-
-            _lastX = e.X;
-            _lastY = e.Y;            
-
+            if (bufferedPanel1.IsPanning) return;
             int mouseXPos = _state.WindowXToRoom(e.X);
             int mouseYPos = _state.WindowYToRoom(e.Y);
             string xPosText = mouseXPos.ToString();


Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Text;
+using System.Windows.Forms;
+
+namespace AGS.Editor
+{
+    public class FreePanControl : BufferedPanel
```
persn, line 9: Your ordering in the class seems to be random. Please organize your class so that things appear in the following order

1. Fields like `private int field;`
2. Properties `public int Field { get; }`
3. Constructor `public FreePanControl() {}`
4. Functions `public void TheFunction() {}`

ericoporto: I pushed a new commit below, please check if it fixes the ordering. 

Also, some winforms method appear to not have explicitly mentioned public/private in the docs. Should I add the public keywords in these?
https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.createparams?view=netframework-4.5

persn: The one you linked to in the doc is `protected` instead of private/public. It means private, but also accessible for classes that inherits from the class that declares the function.

persn: Anyway the ordering looks good now. I haven't checked anything else in this PR

ericoporto: Thanks, I will click this resolve conversation button and squash these two commits into one.

morganwillcock:
The only issue I've found is that when you have the drawing tools active the shift+ctrl+click will place the cursor in the wrong visual state until the next time that the cursor moves (but the selected tool will still be active). I don't have a middle click to test the mouse equivalent.

CTRL+SHIFT+CLICK

CTRL+SHIFT+CLICK_2

ericoporto:
Good catch. I pushed a fix.

ericoporto:
There's a theorical edge case that I could not reproduce that is in my head: you are in the room editor, and for some reason the cursor turns into a wait cursor and then you decide to pan, on the pan release, the curso is a wait cursor forever.

Now to solve this case what I could think of is adding custom events for OnPanGrab and OnPanRelease and on the room editor side, figure out the right cursor. I could keep the previous behavior then to pair this.

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        public FreePanControl() : base()
+        {
+
+        }
+
+        protected virtual void OnPanGrab(EventArgs e)
+        {
+            EventHandler handler = PanGrabbed;
```
persn, line 35: C# trick. Squeeze 3 lines into 1.

```suggestionL35-37
            PanGrabbed?.Invoke(this, EventArgs.Empty);
```

ericoporto: I pushed a fix doing this squeeze!

ericoporto:
I prefer to get the pan working and later in a not PR figure out how to unify scaling from RoomEditorSettings and GUIEditor, moving it to FreePanControl.

persn:
I think you can write unit tests for this class, that way you can reproduce theoretical edge cases that may be difficult to reproduce when flinging the mouse around in the editor. I'll have to test around a little first to see how the unit tests react on WinForms component and then I can coach you in writing the test if you need it.

ericoporto:
@persn I would like that! I don't have much experience writing tests, specially with GUI apps, so it would be interesting to learn to. You can poke me on Discord at any time.

persn:
Writing unit tests for Win Forms panels

<details>
  <summary>Background</summary>

I don't know how much you know exactly about testing so I'll give you a theory crash course straight away because I don't want to get confused about terminology and such. Tests comes in 3 different main categories unit tests, integration tests, and E2E (End to End, typically UI tests in a UI driven app). 

These 3 main categories are usually expressed on a [test pyramid](https://medium.com/better-programming/the-test-pyramid-80d77535573), where unit tests are at the bottom because as you can see in the diagram the base of the pyramid is the widest, this expresses the quantity, there's ideally more unit tests than any kind of test in a software. The reason for this, which you can also see by the arrow on the left of the pyramid is that unit tests are run the fastest and they are also the easiest, quickest, and cheapest to write. In the same way as you follow the pyramid up you'll see integration test we should have less of than unit tests (because the pyramid is narrower there), they're slower and more expensive, and of course UI tests are the slowest and most expensive to implement.

We'll focus on unit test, the other types requires their own framework and infrastructure which we don't have. Unit test are test that test only units of code, what this means is when we test a class we should only test that class, we don't rely on outside GUI, databases, or even other classes that communicate with the class we want to test. We usually achieve this designing our systems so that these different systems are loosely coupled, and writing mocks for the few dependencies there are. Unfortunately for us the AGS editor has a lot of tight coupling, especially in the GUI bits, and Win Forms isn't designed with unit testing in mind. It's likely we'll have to resort to some hacks to achieve testable code.

I'm going to assume your familiar with Inversion of Control (IoC) and such principles so I'll skip that, as long as you don't rely on any external dependencies in your class we don't need to get into it right now.

We use [NUnit](https://github.com/nunit/docs/wiki/NUnit-Documentation) for testing
</details>

<details>
  <summary>Execution</summary>

We're going to avoid witchcraft as much as possible (except for the mocking framework if we have to use it), after all unit tests are supposed to be easy to write.
1. You'll have to convert the fields you want to test to publicly accessible properties so that the tests can access the data you want.
`public Point PanGrabPoint { get; private set; } = Point.Empty;`
2. Now we're going to hack a little, all your code takes place inside WinForms events which are `protected`, raising events outside of the class is not valid so we can't do this from tests. Usually we'd solve this by designing a solution that separates WinForms UI logic with our own custom logic. Instead I'm going to propose a very simple little change on the OnMousexxx functions you have

Instead of `protected override void OnMouseMove(MouseEventArgs e)` 

Try `public new void OnMouseMove(MouseEventArgs e)`

AFAIK it's not considered good practice to use the `new` keyword this way, but I think it's acceptable given that Winforms isn't designed for unit testing, also I peeked into the WinForms souce code and they use this technique to write tests also 🙄 , I'm hoping we can avoid making any fancier solutions
3. Now we can finally write tests. I wrote a quick setup, you can copy paste that and start trying it out. Make a folder `Panes` in the project `AGS.Editor.Tests` and put it there, remove the comments I wrote
```
using System.Drawing;
using System.Windows.Forms;
using NUnit.Framework;

namespace AGS.Editor.Panes
{
    [TestFixture]
    public class FreePanControlTests
    {
        // Generic naming of the field here means we don't have to refactor if the class ever changes
        // names, unit tests only test on thing at the time so we don't have to worry about any other
        // panes being introduces to the same class
        private FreePanControl _pane;

        [SetUp]
        public void SetUp()
        {
            // Unit tests should never be affected by other running tests so create new instance every time
            _pane = new FreePanControl();
        }

        [Test]
        public void PanGrabPointDefaultsToEmpty()
        {
            // Unit tests should tests micro units of things at once
            // In this test we test the default value of _pane.PanGrabPoint
            // If you want to test if another property has received a default value, make a NEW test for it
            Assert.That(_pane.PanGrabPoint, Is.EqualTo(Point.Empty));
        }

        [TestCase(1)]
        [TestCase(5)]
        [TestCase(int.MaxValue)]
        public void PanGrabPointSetsXOnMiddleMouseDown(int expected)
        {
            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, clicks: 1, expected, y: 0, delta: 0));
            Assert.That(_pane.PanGrabPoint.X, Is.EqualTo(expected));
        }
    }
}
```
</details>

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Drawing;
+using System.Text;
+using System.Windows.Forms;
+
+namespace AGS.Editor
+{
+    public class FreePanControl : BufferedPanel
+    {
+        private const int WS_EX_COMPOSITED = 0x02000000;
+        private int _lastX, _lastY; //used to prevent useless mouse move events
```
persn, line 12: Use a `Point` instead, then you can also change the code in `OnMouseMouse` to `e.Location == lastLocation` and `lastLocation = e.Location`.

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        protected virtual void OnPanRelease(EventArgs e)
+        {
+            PanReleased?.Invoke(this, EventArgs.Empty);
+        }
+
+        protected override void OnMouseDown(MouseEventArgs e)
+        {
+            // defines the shortcut for click pan Control+Shift+Left Click Hold or Middle Click Hold without any modifier.
+            if ( (e.Button == MouseButtons.Left   && Control.ModifierKeys == (Keys.Control|Keys.Shift)) ||
```
persn, line 47:
```suggestion
            if ( (e.Button == MouseButtons.Left   && ModifierKeys == (Keys.Control|Keys.Shift)) ||
```

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,98 @@
...
+                 (e.Button == MouseButtons.Middle && Control.ModifierKeys == Keys.None) )
+            {
+                OnPanGrab(e);
+                _panGrabPoint = e.Location;
+                Cursor = _panCursor;
+                return;
```
persn, line 53: Is the `return` intentional? It's normally anti-pattern to not continue to `base` when overriding `virtual` functions, in this case someone who might have subscribed to the `MouseDown` event will be expecting their event to trigger, but it won't in certain cases because you circumvented it.

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            if (_panGrabPoint != Point.Empty)
+            {
+                Point scrollPosition = AutoScrollPosition;
+                scrollPosition.X = _panGrabPoint.X - e.X - scrollPosition.X;
+                scrollPosition.Y = _panGrabPoint.Y - e.Y - scrollPosition.Y;
+                AutoScrollPosition = scrollPosition;
+                Refresh(); // this prevents the room image to look garbled when panning
+                _panGrabPoint = e.Location; // <- without this the pan doesn't work
+                return;
```
persn, line 89: Same as my comment in `OnMouseDown`. Do we really want to `return` here?

Editor/AGS.Editor/Panes/FreePanControl.cs
ericoporto: the idea is to prevent firing the events like editing the walkable areas. The alternative way is adding a property like `IsGrabbing` and only process the Events in the Room Editor if `IsGrabbing` is false. My mentality came from using `ClaimEvent` in AGS Script.

Editor/AGS.Editor/Panes/FreePanControl.cs
ericoporto: updated to apply this simplification. Bonus Control doesn't confuse with Keys.Control, so a good improvement!

ericoporto:
image

https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.mouseeventargs?view=netframework-4.5

Why the docs make no sense when comparing with the thing I see in Visual Studio?

Anyway, the Point in `e.Location` doesn't appear to be compatible with the `Point` used elsewhere and simpler operations (like subtraction) results in error. 

screenshot_error_grab

Both points are from System.Drawing.Point! I will keep operating with X and Y individually for now.

---

I don't like the idea of making protected/private methods/properties public for testability. I thought about to make a partial class to and just write the tests there, but I can't figure out how to fit this solution into NUnit "decorators" (the text between `[` and `]`, not sure if called decorators).

persn:
> Anyway, the Point in e.Location doesn't appear to be compatible with the Point used elsewhere and simpler operations (like subtraction) results in error.

You're just misunderstanding what the error message is saying. It's trying to tell you that the `-` operator isn't implemented for `Point`, so this is invalid `var test = new Point(5, 0) - new Point(3, 0)`. According the docs it's implemented for `Size` however (don't ask me why not also `Point`), so you could do `AutoScrollPosition = PanGrabPoint - new Size(e.Location) - new Size(AutoScrollPosition);` https://docs.microsoft.com/en-us/dotnet/api/system.drawing.point.op_subtraction?view=netcore-3.1#System_Drawing_Point_op_Subtraction_System_Drawing_Point_System_Drawing_Size_

> I don't like the idea of making protected/private methods/properties public for testability.

I'm going to insist you get over this to a certain extent, you cannot test data if you don't have data to test. The added robustness tests give you makes up for any read-only properties that has to be made public. I will try to meet you halfway here though, I'll spend some more time and try to architect a solution that will let us solve this.

> I thought about to make a partial class

Don't mix tests in with partial classes. Production code and test code should be separated, there's a lot of reasons why. Just read this stackoverflow page for instance https://stackoverflow.com/questions/9022547/should-test-code-be-separate-from-source-production-code

> I can't figure out how to fit this solution into NUnit "decorators"

Off track but you would just add it over your partial class, but you can't reference it from the AGS Editor project because the Nuget isn't configured to add that package. Tests are separated into a separate project and that's how they should be. That's another reason the partial trick isn't going to work because you can't make partial between files that exist in 2 different projects, 1 file will already be finished compiled when the other project references it so it can't be added for that reason.

> not sure if called decorators

[Attributes](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/), but I think anyone will know what you mean when you say decorator.

morganwillcock:
Will this sort of test run (correctly or at all) where there is no graphical context? Maybe these would have to be marked as optional to be able to be skipped on the CI system; I have my doubts that WinForm components would initialise. I guess there is only one way to find out...

persn:
Great question, it was worth pointing out so I don't chase dead ends. I plan to come up with a design that separates the Winforms execution from our own code, so it should be fine.

ivan-mogilko:
@persn does C# have any means to change accessibility depending on something (compilation flags, and so on)?

One quick idea came to my mind, is to wrap private fields in public properties that are only available  if a macro is declared.

persn:
Yeah that's doable. .NET has something called shared projects, which are just holder projects that contain a set of source code, these projects can be referenced by other regular .csproj projects, each .csproj can define their own macros. On compile the source code in the shared project will be compiled for each .csproj, taking the macros in consideration.

There's a lot of overhead with this solution though. Not sure if it's worth it, or maybe it is? **Theoretically** if we wanted to something like change UI frameworks and continue to use C# in the future we could just define macros that filter out the WinForms parts and make a new project that uses a new UI framework and reuses a lot of the already existing code.

I need to think about this.

ivan-mogilko:
@persn sorry, maybe I was not clear about what I am refering to. I did that comment in relation to unit tests.

persn:
Yes I understand, I was just going off topic slightly because I was trying to justify the change needed by adding more potential use cases, which is why I mentioned they UI stuff.

Anyway the code would look something like this

```
#if TEST
    public Point PanGrabPoint => _panGrabPoint;
#endif
```

Editor/AGS.Editor/Panes/FreePanControl.cs
ericoporto: I moved to using IsGrabbing and placed the Return in the code. There's only one return now in Event handling portion but it's actually solving an unexpected behavior from WinForms, so I will leave there.

Editor/AGS.Editor/Panes/FreePanControl.cs
ericoporto: I removed and did a different approach.

persn:
I figured it out. We can access private members by reflection. This is probably the most elegant solution since it doesn't require us to muddle with the production code in any way (although we don't know what'll happen when the tests run on the CI server yet). I'm going to write a small framework that allows us to access hidden values from the test project and then we'll see

ericoporto:
I understand the need for unit tests. I don't want to do it for this particular PR, UI testing is hard, and for legacy Windows Forms, harder. 

This PR just highlighted that I want to move away from anything Windows exclusive as soon as possible - like [MAUI](https://github.com/dotnet/maui) appears to be a cool thing to move to.

ericoporto:
I am hours staring at the screen trying to figure out what to do.

persn:
I had a Eureka moment today. this code sample lets you test a class with hidden accessors without messing with any of the accessors in the production code. Try adding the following contents to a file Panes/FreePanControlTests.cs in the AGS.Editor.Tests project.

```
using System.Windows.Forms;
using NUnit.Framework;

namespace AGS.Editor
{
    [TestFixture]
    public class FreePanControlTests
    {
        // Generic naming of the field here means we don't have to refactor if the class ever changes
        // names, unit tests only test on thing at the time so we don't have to worry about any other
        // panes being introduces to the same class
        private FreePanControlTestableWrapper _pane;

        [SetUp]
        public void SetUp()
        {
            _pane = new FreePanControlTestableWrapper();
        }

        [Test]
        public void IsPanningDefaultsToFalse()
        {
            // Unit tests should tests micro units of things at once
            // In this test we test the default value of _pane.IsPanning
            // If you want to test if another property has received a default value, make a NEW test for it
            Assert.That(_pane.IsPanning, Is.False);
        }

        [Test]
        public void IsPanningIsTrueOnMiddleMouseClick()
        {
            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
            Assert.That(_pane.IsPanning, Is.True);
        }

        [Test]
        public void PanGrabbedInvokesOnMiddleMouseClick()
        {
            bool panGrabbedInvoked = false;
            _pane.PanGrabbed += (s, e) => panGrabbedInvoked = true;
            _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
            Assert.That(panGrabbedInvoked, Is.True);
        }

        private class FreePanControlTestableWrapper : FreePanControl
        {
            public new void OnMouseDown(MouseEventArgs e) => base.OnMouseDown(e);
            public new void OnMouseUp(MouseEventArgs e) => base.OnMouseUp(e);
            public new void OnMouseMove(MouseEventArgs e) => base.OnMouseMove(e);
        }
    }
}
```

ericoporto:
Hey @persn I added the mentioned test but I can't figure out if it's really running.

persn:
I'll ping you on discord

ericoporto:
I could do something like this, for status, and it worked.

```C#
[Test]
public void IsPanningReturnsToFalseOnPan()
{
    _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
    Assert.That(_pane.IsPanning, Is.True);
    _pane.OnMouseMove(new MouseEventArgs(MouseButtons.Middle, 1, 4, 4, 0));
    Assert.That(_pane.IsPanning, Is.True);
    _pane.OnMouseUp(new MouseEventArgs(MouseButtons.Middle, 1, 3, 3, 0));
    Assert.That(_pane.IsPanning, Is.False);
}
``` 

But forcing the setup of the sizes in WinForms doesn't appear to be trivial. I am not sure I understand how to force this. The idea is to make a control of 80x80 pixels of scrollable area restricted in a viewable 20x20 pixels area. So this apparently doesn't work:

```C#
[SetUp]
public void SetUp()
{
    _pane = new FreePanControlTestableWrapper();
    _pane.MinimumSize = new Size(80, 80);
    _pane.AutoScrollMinSize = new Size(20, 20);
    _pane.ClientSize = new Size(80, 80);
}
```

The actual fake pan would then be something like this, I believe this is not wrong:

```C#
[Test]
public void IsScrollingGrabbedDistance()
{
    Point initialOffset = new Point(0, 0);
    Point finalOffset = new Point(20, 20);
    Assert.That(_pane.AutoScrollPosition.X == initialOffset.X, Is.True);
    Assert.That(_pane.AutoScrollPosition.Y == initialOffset.Y, Is.True);
    _pane.OnMouseDown(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
    _pane.OnMouseMove(new MouseEventArgs(MouseButtons.Middle, 1, 5, 5, 0));
    _pane.OnMouseMove(new MouseEventArgs(MouseButtons.Middle, 1, 15, 15, 0));
    _pane.OnMouseUp(new MouseEventArgs(MouseButtons.Middle, 1, 15, 15, 0));
    Console.WriteLine(_pane.AutoScrollPosition.X.ToString());
    Assert.That(_pane.AutoScrollPosition.X == finalOffset.X, Is.True);
    Assert.That(_pane.AutoScrollPosition.Y == finalOffset.Y, Is.True);
}
```

I know the tests were meant to be succinct, but this is the actual panning.

ericoporto:
I think I have done all in this PR, if there's any change regarding the commits done, please inform.

Editor/AGS.Editor.Tests/Panes/FreePanControlTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            Assert.That(_pane.IsPanning, Is.True);
+            _pane.OnMouseUp(new MouseEventArgs(MouseButtons.Middle, 1, 3, 3, 0));
+            Assert.That(_pane.IsPanning, Is.False);
+        }
+
+        [Test]
+        public void CursorIsModifiedAndReturned()
+        {
+            _pane.Cursor = Cursors.IBeam;
+            Assert.That(_pane.Cursor != _pane.GetPanCursor(), Is.True);
```
persn, line 54: When comparing 2 values do this instead. Same for every other test
```suggestion
            Assert.That(_pane.Cursor Is.EqualTo(_pane.GetPanCursor());
```

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            get
+            {
+                CreateParams cp = base.CreateParams;
+                cp.ExStyle |= WS_EX_COMPOSITED;
+                return cp;
+            }
+        }
+
+        public FreePanControl() : base()
```
persn, line 30: You don't need to a empty constructor

ericoporto: Can you elaborate? You mean I can omit all these 4 lines of code of the constructor and it will still work in C# ?

ivan-mogilko: Empty constructors are generated automatically unless there's a constructor with parameters.

ivan-mogilko: IIRC default constructor is generated automatically unless there are constructors with parameters too.

ericoporto: Há, Very cool, didn't knew about it. I will fix this in a previous commit and push force

persn: When programming classes the compiler will make a default constructor for you if you haven't explicitly made any. The difference when programming structs is that it will always make a default empty constructor even you make a constructor with parameters, otherwise the line `DateTime dateTime` would be illegal, which it isn't

ericoporto: I removed the constructor! :]

Editor/AGS.Editor.Tests/Panes/FreePanControlTests.cs
ericoporto: I fixed this in the last commit.

ericoporto:
I think some people may not like that the middle click button now requires a modifier (ctrl or shift or alt) to be able to get the coordinate. This could be configurable, in some way, to revert to old behavior, maybe - but I don't know how, so I won't be adding it here.

In my opinion, the majority of use cases of copying points from the Editor is to paste them in the Script Editor when scripting the things that happen in the room, where the player should go and stuff like that. For these I think the best approach would be a points layer to store points of interest so one could do `pTargetUnderFallingPiano` or some other description for Point objects that would be read only and instantiated in the room.

Editor/AGS.Editor/Panes/FreePanControl.cs
```diff
@@ -0,0 +1,96 @@
...
+                scrollPosition.X = _panGrabPoint.X - e.X - scrollPosition.X;
+                scrollPosition.Y = _panGrabPoint.Y - e.Y - scrollPosition.Y;
+                AutoScrollPosition = scrollPosition;
+                Refresh(); // this prevents the room image to look garbled when panning
+                _panGrabPoint = e.Location; // <- without this the pan doesn't work
+            }
+
+            base.OnMouseMove(e);
+        }
+        public Cursor GetPanCursor() { return _panCursor; }
```
persn, line 91: You can just get rid of the field `_panCursor` and instead do this

```
public Cursor PanCursor { get; private set; }
```

Rule of thumb; Try to use fields/properties to read values of the class, and save methods for doing calculations on the class.

Put the property with the rest of the properties in the class

ericoporto: I did it! :grinning:

ericoporto:
If it's alright I can squash the first 6 commits so they are one to add FreePanControl and other for FreePanControlTests.

ivan-mogilko:
@persn is your pr #1094 needed for this to actually run tests?

persn:
No, not anymore. This is ready to merge as far as I'm concerned. There's still stuff in #1094 I'd like to get merged though but I'll look into that after this gets merged so I can clean it up without conflicting changes

morganwillcock:
I'm still not sure where the annotations are meant to show in GitHub, but if you look at the test result file directly you can see that they have already run and passed:
```
<testsuite time="0.273749" failures="0" tests="5" name="AGS.Editor.Tests.dll.AGS.Editor.FreePanControlTests." timestamp="2020-06-30 16:05:45Z" skipped="0" errors="0">
  <testcase time="0.260785" name="CursorIsModifiedAndReturned" classname="AGS.Editor.FreePanControlTests" status="Passed" assertions="3"/>
  <testcase time="0.001178" name="IsPanningDefaultsToFalse" classname="AGS.Editor.FreePanControlTests" status="Passed" assertions="1"/>
  <testcase time="0.000589" name="IsPanningIsTrueOnMiddleMouseClick" classname="AGS.Editor.FreePanControlTests" status="Passed" assertions="1"/>
  <testcase time="0.006266" name="IsPanningReturnsToFalseOnPan" classname="AGS.Editor.FreePanControlTests" status="Passed" assertions="3"/>
  <testcase time="0.001248" name="PanGrabbedInvokesOnMiddleMouseClick" classname="AGS.Editor.FreePanControlTests" status="Passed" assertions="1"/>
</testsuite>
```

ericoporto:
I had a brief email exchange with a person from Cirrus Ci about the agent not coming up and he gave me the following answer

> Such errors most of the time related to OOM and Windows containers unfortunately don't have memory charts so can't confirm it. If you see it frequently then I'd suggest trying to increase memory.

For now the task has been rerun. I had force pushed the commits with some squashed.

---

Anyway, I will stop touching this PR now (unless someone say something more about it, which is alright too).

morganwillcock:
> I had a brief email exchange with a person from Cirrus Ci about the agent not coming up

I think the issues we see are because when tasks get rescheduled there appears to be potential for the archives that store the artifacts in each task to become corrupted. Possibly two tasks are uploading to the same place, or maybe the archive is streamed back from a failed task, or merges a failed with a successful one, but the usual problem is that the AGS Controls library has just disappeared from the Editor build artifacts and re-running the Editor build will fix it.

Similarly I need to speak to someone there about the incorrect expansion of variables in the batch file processing, so if you can point me in someone's direction then maybe I can try to explain the weirdness that I tend to see.

ericoporto:
@morganwillcock I sent you a message through the forums. :)

PR #1091, opened 2020-06-17
removes not used using reference to AGS.Types

ericoporto:
I was looking if there was any dependency to AGS.Types in the AGS.CScript.Compiler. Apparently there isn't but I found a spurious AGS.Types mention. This PR removes it.


ericoporto:
It appears the compiler picks AGS Version from the AGS.Types...

PR #1094, opened 2020-06-18
More setup for the editor tests

persn:
Did a couple of things, the most important is that I added a way to get and set private/protected members of a class and also invoke private/protected methods as related to the discussion of testing in #1089


persn:
Don't review this yet. I came up with an alternate proposal to the reflection code.

ivan-mogilko:
> Don't review this yet. I came up with an alternate proposal to the reflection code.

@persn are you still working on this?

persn:
The alternative proposal was implemented and merged. There's still some minor work here I think should be merged, but to avoid confusion I'll open a new PR for them

Issue #258, opened 2015-08-07
Multiple monitors results in very large screen sizes


Mailaender:
On 3.4.0.3 (Alpha) the game detects very large (combined) screen sizes when multiple monitors are used. I tried tweaking `acsetup.cfg` but my settings (like no full screen, letterbox yes/no) seem to get ignored. Running on openSUSE 13.2 (Linux).

```
AGS: Resonance
AGS: Checking for disk space
AGS: Initializing resolution settings
AGS: Device display resolution: 3286 x 1080
AGS: Game native resolution: 320 x 240 (32 bit) letterbox-by-design
AGS: Game settings: windowed = no, screen def: max, screen size: 0 x 0, match device ratio: yes, frame placement: center
AGS: Init gfx driver
AGS: Created graphics driver: Allegro/DX5
AGS: Supported gfx modes (32-bit): 
        640x480;800x600;800x600;1024x768;1366x768;
AGS: Supported gfx modes (24-bit): 
        640x480;800x600;800x600;1024x768;1366x768;
AGS: Requested gfx filter: StdScale, filter scaling: max uniform
AGS: Applying scaling filter: StdScale
AGS: Switching to graphics mode
AGS: Attempting to find nearest supported resolution for screen size 3286 x 1080 (32-bit) fullscreen
AGS: Could not find compatible graphics mode
AGS: Attempting to find nearest supported resolution for screen size 3286 x 1080 (24-bit) fullscreen
AGS: Could not find compatible graphics mode
AGS: Attempting to find nearest supported resolution for screen size 3286 x 1048 (32-bit) windowed
AGS: Attempt to switch gfx mode to 3286 x 1048 (32-bit) windowed; game frame: 1280 x 960, frame placement: center
AGS: Succeeded. Using gfx mode 3286 x 1048 (32-bit) windowed
        filter dest (1003, 44, 2282, 1003 : 1280 x 960), render dest (1003, 44, 2282, 1003 : 1280 x 960)
AGS: Preparing graphics mode screen
AGS: Initializing colour conversion
AGS: Check for preload image
AGS: Initialize sprites
AGS: Set up screen
AGS: Initialize game settings
AGS: Prepare to start game
```

This doesn't happen on stable version 3.3.4.1.

ivan-mogilko:
This is a bug; AGS does not detect multiple monitors.

[...]

PR #1095, opened 2020-06-23
fixes for multimonitor of AGS on Linux

ericoporto:
- adds each monitor current resolution as an extra mode in the listing;
- adds new `Size AGSPlatformDriver::GetDesktopSize()` method;
- adds a logic to try to figure out the right resolution for GetDesktopSize on Linux (X11 desktop);
- if things fail, fallback to base GetDesktopSize implementation, which relies on previously used Allegro;

My desktop only has 1 monitor and 1 videocard, so need someone to test too. 

Possible fix to #258 ;

### DIFF
--- a/Engine/main/graphics_mode.cpp
+++ b/Engine/main/graphics_mode.cpp
@@ -91,9 +91,7 @@ DisplayModeSetup::DisplayModeSetup()
 
 Size get_desktop_size()
 {
-    Size sz;
-    get_desktop_resolution(&sz.Width, &sz.Height);
-    return sz;
+    return platform->GetDesktopSize();
 }
 
 Size get_max_display_size(bool windowed)
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -8,6 +8,8 @@ set_target_properties(engine PROPERTIES
     CXX_EXTENSIONS NO
 )
 
+option(AGS_X11_WITH_XINERAMA "Enable X11 Xinerama fixes" ON)
+
 target_include_directories(engine PUBLIC .)
 
 target_sources(engine
@@ -591,6 +593,16 @@ if (WIN32)
     target_link_libraries(engine PUBLIC DirectX::Direct3D9 shlwapi)
 elseif (LINUX)
     target_link_libraries(engine PUBLIC X11::X11)
+
+    if(AGS_X11_WITH_XINERAMA)
+        check_include_file("X11/extensions/Xinerama.h" HAVE_XINERAMA_H)
+        check_library_exists(Xinerama XineramaQueryExtension "" CAN_XINERAMA)
+        if(CAN_XINERAMA AND HAVE_XINERAMA_H)
+            target_compile_definitions(engine PRIVATE AGS_XWINDOWS_WITH_XINERAMA)
+            find_library(XINERAMA_LIB "Xinerama")
+            target_link_libraries(engine PRIVATE "${XINERAMA_LIB}")
+        endif()
+    endif(AGS_X11_WITH_XINERAMA)
 endif()
 
 get_target_property(ENGINE_SOURCES engine SOURCES)
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -62,6 +62,13 @@ void AGSPlatformDriver::RegisterGameWithGameExplorer() { }
 void AGSPlatformDriver::UnRegisterGameWithGameExplorer() { }
 void AGSPlatformDriver::PlayVideo(const char* name, int skip, int flags) {}
 
+Size AGSPlatformDriver::GetDesktopSize()
+{
+    Size sz;
+    get_desktop_resolution(&sz.Width, &sz.Height);
+    return sz;
+}
+
 const char* AGSPlatformDriver::GetAllegroFailUserHint()
 {
     return "Make sure you have latest version of Allegro 4 libraries installed, and your system is running in graphical mode.";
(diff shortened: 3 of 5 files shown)


Engine/main/graphics_mode.cpp
```diff
@@ -91,9 +91,7 @@ DisplayModeSetup::DisplayModeSetup()
 
 Size get_desktop_size()
```
ivan-mogilko, line 92: It's worth trying if you can remove this function and use platform->GetDesktopSize everywhere directly.

ericoporto: Ah, good point, should be cleaner this way. I will do this in a separate commit later.

ericoporto:
Note, this commit should not be ported to ags4, hopefully by the time we work on it, something better will be in place.

ivan-mogilko:
Who may confirm that this is working? It's complicated for me to reliably reproduce this issue on VM.

ivan-mogilko:
There's something suspiciously looking in GetSystemDisplayModes, I will check that out later today.

ivan-mogilko:
@ericoporto so, I have a question, why GetSystemDisplayModes combines modes from `get_gfx_mode_list` and `XineramaQueryScreens`?

Do you have some log running this patch to see which modes it detects and how does engine choses resolution?

ericoporto:
> ericoporto so, I have a question, why GetSystemDisplayModes combines modes from get_gfx_mode_list and XineramaQueryScreens, what does that achieve?

So, I fully believe we should not do that and just use the desktop resolution (the monitor resolution being used) for full screen, because this also prevents some case of a crash that the monitor state is kept at a different resolution - no this has never happened for me on ags, but has happened for me with other non ags games, this is why I think it may be possible. So in my head the mode list would be only that single value, the desktop resolution. And I could not find from where to figure out the monitor colors - I know passing 32 makes it work.

So in sum, I think this is wrong, but it retains old behavior, please keep reading.

---

> Do you have some log running this patch to see which modes it detects and how does engine choses resolution?

With that said, here's what I did.  First I passed a long hdmi cable between my PC (2560x1440p monitor) and my TV (1920x1080). Then, ran the debug version of the engine with `--log` for DungeonHands. [ags.log.zip](https://github.com/adventuregamestudio/ags/files/4861488/ags.log.zip)

Here are the modes this branch brings in this case: 
```
Supported gfx modes (32-bit): 
	640x480;640x480;640x480;720x480;720x576;800x600;800x600;1024x768;
	1024x768;1152x864;1280x720;1280x720;1280x800;1280x1024;1280x1024;1366x768;
	1440x900;1600x1200;1680x1050;1920x1080;1920x1080;1920x1080;1920x1080;1920x1200;
	2048x1080;2048x1080;4480x1440;2560x1440;1920x1080;
Attempting to find nearest supported resolution for screen size 2560 x 1440 (32-bit) windowed
```

If you look at the list it skips my main monitor size `2560x1440` and instead writes `4480x1440`, which is 2560+1920, the sum of my monitors width. Last two modes are the ones I picked using Xinerama (`2560x1440;1920x1080;`).

My general idea is that at this point I don't know where the game is so I just want the monitors resolution so when I launch later I can use this information to validate there is a valid fullscreen mode.

As far as I can tell, the other modes don't really matter unless you want to do the scaling in the monitor - I imagine people don't want that because the scaling in the monitor is VERY ugly. But they are there if someone wants to force this. 

I think modes resolution different than the monitor are not relevant for the game launching windowed - from experimenting, I found the config code logic hard to understand.

ivan-mogilko:
I have another question, this new code is built under AGS_X11_WITH_XINERAMA, 
1) does this mean that one would have to know to explicitly set this flag when compiling on their system?
2) what will happen if a version built with AGS_X11_WITH_XINERAMA is run on machine where it is not supported? (This is more a question of how the pre-built linux binaries are supposed to be made, since they go to many different machines)

ericoporto:
1. Yes, but it defaults to ON. If the person has set it on and has the library at build time, it defines `AGS_XWINDOWS_WITH_XINERAMA`. For now this only affects CMAKE builds - I simply don't know how to perform this verification in the Makefile.
2. I expect XineramaQueryScreens to return null in this case, so the code should perform the regular Allegro query.

i30817:
Does this work on wayland already? Ubuntu 20 should have the wayland login avaliable to test if you want to and have multiple monitors (i don't). Remember to logout and change back to x11 because it remembers it.

ericoporto:
No and I don't intend too. The only sane way is to use something not Allegro to figure the surface/Windows for drawing things. Also I have Nvidia, so Wayland naturally doesn't work for me.

i30817:
Closed source problems i guess. Glad i always had a open driver, even if my gpu&cpu is crap.

ivan-mogilko:
Why was this closed, is it decided to be not applicable? I was planning to check this out soon.

ericoporto:
If there's any changes required to the code I won't have time to give attention to it soon - like providing a Makefile build, which I don't have desire to provide.

In my mind it's still something that the proper fix should be patching allegro and upstreaming the patch above if desired, by having it on AGS means the responsibility of AGS includes figuring out how monitors work on each system. We have done that for mobile, but the work is done by patching Allegro.

I will leave open then, but these are my thoughts. It could be in a 3.5.1 release branch but I don't think having on ags3-dev (our master) is appropriate.

ivan-mogilko:
>  It could be in a 3.5.1 release branch but I don't think having on ags3-dev (our master) is appropriate. It's a "patch" but not a "fix".

Please elaborate, our master **is** for 3.5.1 currently.

ericoporto:
You probably won't need this fix if we don't use Allegro4 and instead use some other maintained library to poll current monitor resolution.

rofl0r:
> poll current monitor resolution

is monitor resolution subject to change during ags execution?

ivan-mogilko:
Alright, after some thought, since you were insisting this is not an optimal solution, I'll close this again.

PR #1099, opened 2020-06-27
3.5.0 Patch: fix "Change ID" does not work correctly + related oversights

ivan-mogilko:
This fixes underlying character etc arrays not updated properly when user changes game object ID. This mistake caused these objects to remain on previous positions, and, for example, retain previous indexes at runtime (until user reopens project in the editor).

Additional fixes:
* fixed value constraints when changing View's ID;
* update character's preview when View's ID is changed;
* fixed property grid was switching to object whose ID you are changing even if it has different active pane;
* at the same time fixed property grid of inactive panel did not have ID updated in the object title for the second object when two IDs swapped (i.e. if you have Char1 and Char2 panes opened, and swap their IDs, then only one of them got their property grid title updated).
* fixed Character Room Filter not updating itself after character's ID was changed (resulted in a wrong character displayed in room).

### DIFF
--- a/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
+++ b/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
@@ -154,17 +154,6 @@ namespace AGS.Editor.Components
             string newNodeID = AddTreeNodeForItem(item);
             return newNodeID;
         }
-
-        // Swaps two items in the flat item list only, keeping their folder location unchanged.
-        protected void SwapItemsInFlatList(int index1, int index2)
-        {
-            if (index1 == index2)
-                return;
-            var list = GetFlatList();
-            ItemType item = list[index1];
-            list[index1] = list[index2];
-            list[index2] = item;
-        }
                        
         protected void CreateSubFolder(string parentNodeID, FolderType parentFolder)
         {
--- a/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
+++ b/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
@@ -14,6 +14,7 @@ namespace AGS.Editor.Components
         where FolderType : BaseFolderCollection<ItemType,FolderType>
     {
         protected abstract FolderType GetRootFolder();
+        protected abstract IList<ItemType> GetFlatList();
         protected abstract ProjectTreeItem CreateTreeItemForItem(ItemType item);
         protected abstract void AddNewItemCommandsToFolderContextMenu(string controlID, IList<MenuCommand> menu);
         protected abstract void AddExtraCommandsToFolderContextMenu(string controlID, IList<MenuCommand> menu);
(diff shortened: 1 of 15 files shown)


Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
```diff
@@ -153,6 +154,17 @@ protected string AddSingleItem(ItemType item)
             string newNodeID = AddTreeNodeForItem(item);
             return newNodeID;
         }
+
+        // Swaps two items in the flat item list only, keeping their folder location unchanged.
+        protected void SwapItemsInFlatList(int index1, int index2)
```
persn, line 159: Swapping items in a array is a very useful feature. Consider making this more generic so it's reusable for us for any array. Like an extensions method that could be added to AGS.Types/Utilities.cs for example.

```
public static void Swap<T>(this T[] array, int index1, int index2)
{
    if (index1 == index2)
        return;

    T temp = array[index1];
    array[index1] = array[index2];
    array[index2] = temp;
}
```

Editor/AGS.Editor/Components/ViewsComponent.cs
```diff
@@ -61,7 +61,13 @@ protected override void ItemCommandClick(string controlID)
...
                 View viewClicked = _items[_rightClickedID];
                 int oldNumber = viewClicked.ID;
-                int newNumber = Factory.GUIController.ShowChangeObjectIDDialog("View", oldNumber, 1, _items.Count);
+                // Note that the views are not sequential, there may be gaps in IDs
+                int maxNumber = 1;
+                foreach (var obj in _items)
+                {
+                    maxNumber = Math.Max(maxNumber, obj.Value.ID);
+                }
+                int newNumber = Factory.GUIController.ShowChangeObjectIDDialog("View", oldNumber, 1, maxNumber);
```
persn, line 70: This can be be reduced with LINQ

```suggestion
                int newNumber = Factory.GUIController.ShowChangeObjectIDDialog("View", oldNumber, 1, _items.Max(i => i.Value.ID));
```

Editor/AGS.Editor/Components/ViewsComponent.cs
```diff
@@ -186,6 +192,11 @@ private void OnItemIDChanged(View item)
         {
             RePopulateTreeView(GetNodeIDForView(item));
             UpdateOpenWindowTitles();
+            Components.CharactersComponent cmp = ComponentController.Instance.FindComponent<Components.CharactersComponent>();
```
persn, line 195: Qualifying `Components` namespace isn't necessary and null check can be done with the `?.` operator

```suggestion
            ComponentController.Instance.FindComponent<CharactersComponent>()?.UpdateCharacterViews;
```

ivan-mogilko:
^ Is this correct?

persn:
Yeah, exactly what I had in mind.

PR #1101, opened 2020-07-05
AGS compiler rewritten in C++

fernewelten:
**Summary of additions to the language:**

Structs can contain variables that are other structs  In this case, managed structs can only be included via pointers; non-managed structs may not be included via pointers.
```
struct Pair { 
    int X, Y; 
};
managed struct Trio {
    int X, Y, Z; 
};

struct TalkingPositions { 
    Pair PosOfChar1, PosOfChar2; // these components can't be Pair * because Pair isn't managed
}; 
struct FloatingHead { 
    Trio *  HeadPos;     // HeadPos must be pointer because Trio is managed;
}; 
``` 

Can now have dynamic arrays of elements that contain dynamic arrays. 

Functions can now  be defined in any sequence as long as they are in the same compilation unit. So you can reference a function first and define it afterwards. This makes recursive algorithms much easier to code.
```
int  Func1(int a)
{
    return 2 * Func2(a);    // uses Func2()
}

int Func2(int a)     // defines Func2() and is below its first use -- this used to be illegal.
{ … }
```

Forward declarations of functions don't need the "import" keyword if you don't need the functions to be importable elsewhere. Basically, say what you mean: If a function is meant to be importable, use "import"; if it isn't, don't use it. 

```
import int foo (int a, int b = 5, int c = 7); // okay, declares a function that can be imported elsewhere. Header files have lots of those.

struct s
{
    int a, b;
    function foo(int a); // declares the function s::foo() but doesn't make it importable -- so no "import" keyword
};
```

Defaults for function parameters can be defined in any function header no matter whether the function body follows; however, for all the function headers that name defaults at all, they must name _all_ the defaults and these must be exactly the same. 
```
import int foo (int a, int b = 5, int c = 7); // okay, declares some parameter defaults for that function

int foo (int a, int b = 5, int c = 7) // okay
{ 
   ... 
}
int foo (int a, int b, int c); // okay, doesn't name ANY defaults -- can have any number of those lines
int foo (int a, int b = 5, int c = 7 ); // okay, names ALL the defaults that exist  -- can have any number of those lines
int foo (int a, int b, int c = 7) // not okay, names the default for c but not for b
int foo (int a, int b = 6, int c = 7) // not okay, b has a different default than in other declarations of foo.
int foo (int a, int b, int c) // would be okay if there wasn't another function foo with a body already 
{
   ...
}
```

Ternaries: `int foo = (x > 5) ? 1 : 2;` -- foo is set to `1` if `(x > 5)`, to `2` otherwise.

Can leave out the second part of a ternary so that "?" and ":" go right next to each other. For instance,
```int foo = bar ?: 15; ``` 
In this case, if `bar` is  0 or `null` then the result is `15` else the result is `bar`. Handy for specifying a default in case a variable turns out to be `null`.  (This is the same functionality as the `??` operator in C#.)

As in C or C++, can define a struct and global variables in the same statement . For instance, the following will define a vartype `Pair` and a variable `Pair1` that has this new vartype. 
```
struct Pair 
{ 
    int X, Y; 
} Pair1;
``` 

When a variable is declared as a `managed` type, then the compiler _knows_ that it must be a pointer variable because non-pointers to managed types are illegal. (See below for one big exception to this rule.) If you forget the "*" in these cases, then the compiler will happily imply it and carry on compiling. For instance, in the example below, `Var1`, `Var2`, and `W` will be implied to have the type `Weapon *` because `Weapon`  is a `managed` struct.
```
managed struct Weapon
{
        int A;
} Var1;
     
Weapon Var2;
     
function Foo(Weapon W)
{ … }
```
In practice, this means that you can get away with omitting nearly all the pointer symbols in your AGS code.

The big exception is `import` declarations. In import declarations, it _is_ legal to declare `managed struct`s  that aren't pointers, and automatically generated AGS code will actually do that, so the compiler does _not_ imply any `*` for things that are declared `import`. Any pointers must be explicitly written out there.

Multi-dimensional non-dynamic arrays: `int a[5, 3];` or equivalently, `int a[5][3];`. You can have any positive number of dimensions `int a[2][3][5][7][11];`.

String and character literals accept any `\` codes that C or C++ accept, e.g. `\0` or `\n`. (Didn't change the customary handling of `[`.)

Sequences such as `a.b.c[17 * d].e[f]` may now be arbitrarily long. 

Attributes / properties: The getter and setter functions of attributes can now be defined in the same compilation unit as the respective attributes.

Implied `this`: If you are in a struct function, then you can access all components of `this` directly.

```
struct Armor
{
    protected int _damage;

    function Repair();
}
function Armor::Repair()
{
    _damage = 0; // same as this._damage = 0;
}
```

Bitwise negation `~`.

**Summary of compiler code:**

Googletests that cover > 80 % of the code lines

Scanner and parser are C++ classes in the   `AGS` namespace.

Revamped variable access code.

SymbolTable and SymbolTableEntry are classes in the AGS namespace and contain inline functions that characterize symbols.

`#define`s for symbol types converted into `enum` for better debugging. 

Changed some error messages to be more descriptive. 

Many error messages that point out a discrepancy of current use to former code will now name the exact location of the former code that clashes with the current use. (`This function has 3 parameters. This clashes with another definition. See globalscript.ash, line 332`);

The scanner checks that all opening symbols have matching closing symbols. Code like `int a[9)` or `for(i = 0; i < 5;` _(end of code)_ won't even pass the scanner. 

String literal handling is completely within the scanner.

In the original compiler, functions that defined long static arrays of dynamic pointers would yield very long Bytecode sequences. For instance, 
```
int foo()
{
    Object *bar[10000];
}
```
would generate 10000 individual Bytecode statements to release each of the array element pointers. Found a way to release the pointers in a loop instead. So in this case the compiler only generates about a half-dozen of Bytecode statements.

**Caveats:**

- string buffers (`string foo`) are barely tested.

### CODE REVIEW

**Comment by fernewelten** [comment]
I'm still running into the linker issues.
ivan:
> In both of the two error messages you posted, it mentions Common_d.lib:
> But Common_d.lib should not built with AGS.Native, only Common_md_d.lib should.

You seem to be on the right track -- about 80 linker errors say that common_d has some symbol that common_md has already defined. So where is that configured, and why would that fail for Compiler2.Lib when it works for Compiler.Lib? I don't see any different -- or identical -- lines in those two configuration files that would deal with that.

**Comment by fernewelten** [comment]
I'm running into linker issues when compiling the solution `Compiler2.Lib`, as well. There are only 5 of them, and they all seem to be related to `string.cpp` calling functions that the linker can't find. Again, Compiler.Lib compiles and links without problems but Compiler2.Lib doesn't..
```
LNK2019	unresolved external symbol _ags_strlwr referenced in function "public: void __thiscall AGS::Common::String::MakeLower(void)" (?MakeLower@String@Common@AGS@@QAEXXZ)	Compiler2.Lib.Test	ags\Solutions\Compiler.Lib\Compiler_d.lib(string.obj)	1
```
The others pertain to `_ags_strupr`, `_ags_stricmp`, `_ags_strnicmp`, and `_ags_strdup`.

**Comment by ericoporto** [comment]
Wait, I thought this was supposed to have the solution with only one of the compilers in it. I don't know how to have both there.

**Comment by ivan-mogilko** [comment]
> Wait, I thought this was supposed to have the solution with only one of the compilers in it. I don't know how to have both there.

There's only Compiler2 in editor's solution from what I see.

**Comment by ivan-mogilko** [comment]
@fernewelten I am afraid, I will be able to check this out only on the weekends. Have you joined AGS Discord? Perhaps it would be faster and more convenient to discuss your problems there. It seems that you have something set up incorrectly locally, as we can successfully build your changes only with few fixes (mentioned here: https://github.com/adventuregamestudio/ags/pull/585#issuecomment-653942833)

**Comment by ivan-mogilko** [comment]
@fernewelten it compiles correctly now!

Would you be able to squash/remake latest commits to make the redundant commit 8fadae13ab7fd898751d0f415c9acbf78a6fc983 and "revert" commit dissapear?

**Comment by ivan-mogilko** [comment]
Alright, so, I began testing new compiler again, with my game (updated for 3.5.0 script API): https://github.com/ivan-mogilko/ags-lastfurious

Found issues so far:

1. Any error reported for the script header is pointing to asc file instead (which is being compiled).
For example: go to GlobalScript.ash and type any random garbage. When you compile, error will point e.g. to first room's asc (line number seem to match the line in header where error is found though).

2. Unary minus operator cannot be applied to a float.
Code:
<pre>
static float Maths::AbsF(float value)
{
  if (value >= 0.0)
    return value;
  return -value;
}
</pre>
Error: `MathUtil.asc(46): Error (line 46): The operator cannot be applied to a float and a non-float value`
It probably thinks that this is a binary minus operator between float and non-float?
(temporarily fixed by typing `0.0-value`)

3. A function declaration (RotatedView.ash):
<pre>
import static DynamicSprite *[] CreateLoop(int view, int loop, int base_loop = eDirectionDown);
</pre>
A function definition (RotatedView.asc):
<pre>
static DynamicSprite *[] RotatedView::CreateLoop(int view, int loop, int base_loop)
</pre>
Compiler throws error: `RotatedView.asc(40): Error (line 40): Return type is declared as DynamicSprite *[] * here, as DynamicSprite *[] elsewhere. See RotatedView.ash line 7`
Notice it treats return type as having a second `*` for some reason.

4. There is a difference between new and old compiler. Old compiler required function definitions to have same modifiers as imports, e.g.
<pre>
protected import void ResetBase(VectorF *pos, VectorF *dir);
</pre>
<pre>
protected void VehicleBase::ResetBase(VectorF *pos, VectorF *dir)
{
}
</pre>
New compiler throws error: `VehicleBase.asc(7): Error (line 7): Cannot use '[sentinel]' together with 'protected'` (???)
It does not throw any errors if I remove `protected` in function definition.
Note that this does not happen with `static` modifier, only `protected` (maybe few others not met in this script).
What is supposed to be a correct behavior in the new compiler, and what does the error message mean?





I will begin investigating new compiler features next. :)

**Comment by fernewelten** [comment]
1. So now errors in `Globalscript.ash` should be reported in `Globalscript.ash`; I ran a short test in the Editor to check that.
2. Unary minus should now work on floats, too, e.g. `float f = -7.0;` or `return -7.0 + f;`.

**Comment by fernewelten** [comment]
Looking into 4. next; that shouldn't happen.

**Comment by ivan-mogilko** [comment]
After the most recent changes there's a new error:
`_BuiltInScriptHeader.ash(466): Error (line 466): 'SortStyle' is already in use as a type name`

This refers to following enum in a built-in header:
<pre>
enum SortStyle
{
  eNonSorted = 0,
  eSorted = 1
};
</pre>
https://github.com/adventuregamestudio/ags/blob/ags4/Editor/AGS.Editor/Resources/agsdefns.sh#L385

**Comment by fernewelten** [comment]
Very annoyingly, a bug had crept in that invalidated the results of some googletests: Those tests would “pass” no matter what happens. After I fixed that bug, I found that some tests had indeed failed in the meantime, I only hadn't noticed that because of that bug.  So fixed the bugs that those googletests pointed out.

The builtin headers should now compile correctly.

**Comment by fernewelten** [comment]
Very annoyingly, a bug had crept in that invalidated the results of some googletests: Those tests would “pass” no matter what happens. After I fixed that bug, I found that some tests had indeed failed in the meantime, I only hadn't noticed that because of that bug.  So fixed the bugs that those googletests pointed out.

The builtin headers should now compile correctly.

**Comment by ivan-mogilko** [comment]
Trying [the game](https://github.com/ivan-mogilko/ags-lastfurious) with the latest commit 3ab7b90, there are new regressions:

1. If there's a `readonly` attribute NNN, then compiler fails at get_NNN function, e.g.:
`Timer.asc(70): Error (line 70): The qualifiers of 'Timer::get_IsActive' are different here than elsewhere. See Timer.ash line 17`.
The declaration at Timer.ash line 17 is `import readonly attribute bool IsActive;`
Of course, I cannot add readonly to get_NNN as it cannot be applied to a function.
It's interesting that `static` attribute works.

2. Creating a dynamic array of pointers to built-in types fail:
<pre>this.Glyphs = new DynamicSprite[total];</pre>
with error
`SpriteFont.asc(30): Error (line 30): Cannot use 'new' with the built-in type 'DynamicSprite'`
looks like it does not realize I am creating array of pointers there.
(I tried `new DynamicSprite*[total]` for a test, but it did not work either)

**Comment by ivan-mogilko** [comment]
My game fully compiles now, but there's still an issue I found when compiling some of the classic templates. Not sure if it's result of most recent commit, or maybe I did not notice this earlier. But now if there's `static attribute` in a struct, compiler demands that get/set functions were also defined with static keyword (and it works, since normally static may be applied to function definition), which AGS did not require before. Is it a mistake or a new rule?

Besides that there's still a weird error which may be only noticed at runtime. I was able only to narrow it down to a piece of code in my game, but was not able to create a minimal working example, because in a blank game similar case does not occur.

Here's the interesting lines: https://github.com/ivan-mogilko/ags-lastfurious/blob/master/Last'n'Furious/SpriteFont.asc#L37
What happens there is that SpriteFont struct has 2 dynamic arrays: Offs and Widths. By default Offs are initialized with 0s and Widths initialized with gl_width variable.

If you test Offs values right after they were initialized, they all are correctly equal to 0. But if you test them again after initializing Widths, then Offs appear to have gl_width value. Widths are equal to gl_width too.
If you do this:
<pre>
this.Widths[0] = 123;
Display("Offs[0] = %d", this.Offs[0]);
</pre>
Then AGS will display "123".
At the moment I have no idea what is happening, maybe compiler did not write correct memory handles/addresses when dealing with these arrays?

PS. Also I'll be now testing new features mentioned in the first post.

**Comment by fernewelten** [comment]
> But now if there's static attribute in a struct, compiler demands that get/set functions were also defined with static keyword […], which AGS did not require before. 

I wasn't aware that AGS had treated this differently before.

I'm basically treating attributes as a shorthand, the way C# does it: 
- In the struct declaration, I'm converting the attribute into exactly the same as  `import QUALIFIERS VARTYPE get_ATTRIBUTE();` (and a similar declaration for the setter, if the attribute isn't `readonly`);
- in expressions, I'm converting every non-`static` attribute into exactly the same as `OBJECT.get_ATTRIBUTE()`. and every `static` attribute into exactly the same as `STRUCT.get_ATTRIBUTE()`. 

For the declaration in the `struct`, I'm basically copying the qualifiers of the attribute over to the getter and/or setter -- except for stripping `readonly` (if present) and adding `import` (if missing). 

After the struct declaration, the function `get_ATTRIBUTE()` is basically indistinguishable from any other function that is declared in the struct "by hand". They share the same code and the same symbol entries. So that means that after a `static` attribute declaration, the code is expecting a `static` getter and/or setter.

By the way, since attribute getters/setters have the same symbol entries as "conventional" functions, the programmer is free to declare the getter or setter code with a conventional function header: `static int Timer::IsActive() { ... }`. No need for `int IsActive(static Timer)`, although that is allowed and possible, of course.

**Comment by fernewelten** [comment]
Looking into that weird error now.

**Comment by ivan-mogilko** [comment]
Sorry for delay, I think I found minimal necessary code to reproduce this problem:
header:
<pre>
struct TestStruct
{
  int            a[];
  int            b[];
  
  import void    Test(int arg);
};
</pre>
script
<pre>
void TestStruct::Test(int arg)
{
  this.a = new int[1];
  this.b = new int[1];
  this.b[0] = 123;
  Display("TEST %d, %d", this.a[0], this.b[0]);
}
</pre>

So:
* there has to be a struct with 2 dynamic array members and a non-static method.
* method must take at least 1 argument of any type (this is important).

**Comment by ivan-mogilko** [comment]
There are also several issues with error reporting I found while testing:

1. There is no error when struct member's implementation has less arguments than declaration!
E.g. you may have a declaration:
<pre>struct Test
{
    import void Func(int a, int b, int c, int d);
};</pre>
and in script:
<pre>void Test::Func(int a)
{
    Display("whoops");
}
</pre>
Nothing is reported, and the function even gets called!

2. If there's only function prototype found in script and no "import" or function body, then compiler reports error, but the location name of such error is a string full of garbage, e.g. copied output text to clipboard may look like this:
`ЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭЭyсЭЭ(189): Error (line 189): Function 'TestFunc()' has been called but not defined with body nor imported`

3. A very strange error. If you have a struct declared in another lower module *it looks like* new compiler can somehow detect it when compiling an upper module. I am not completely sure about this, as the first post does not mention such feature. In any case, if you try to call that struct's static member function from upper script, then it does not report usual "unexpected token" as in cases of undeclared symbol, but instead it outputs following:
`TestCompiler.asc(25): Error (line 25): Must specify a specific struct for non-static component __dummy_sym_0__`

**Comment by fernewelten** [comment]
Thanks a lot for the narrowed case; this gives me a good idea where the bug might be -- it seems that I have been barking up a completely wrong tree so far. 

You're also right about the parameters; it seems that the compiler will complain about too many of them but not of too few of them. I'll issue a fix shortly.

Will issue fixes shortly.

**Comment by fernewelten** [comment]
> If there's only function prototype found in script and no "import" or function body, then compiler reports error, but the location name of such error is a string full of garbage

My bad. That's due to calling a `printf` type function that expects a C string, but passing a C++ `std::string`  to it instead. Shouldn't happen, but is easily fixed (add `.c_str()` to that parameter)

**Comment by ivan-mogilko** [comment]
Unfortunately after latest commits the compiled script seem completely broken. Most of the time vm fails reporting accessing stack in wrong offset (like "trying to access stack data inside stack entry" which means it points in the middle of some variable larger than 1 byte), or reads from wrong place causing various problems (like array size reported to be 0).

**Comment by fernewelten** [comment]
> Unfortunately after latest commits the compiled script seem completely broken. 

Bummer.  My own small examples didn't trigger the bug. I __did__ check that all the files of your game would compile.

I'm currently looking into it. This time, I'm fairly confident that the bug should be easy to find; The changes to the bytecode haven't been that many.

**Comment by ivan-mogilko** [comment]
After latest commits "Last & Furious" game compiles and runs, and I haven't noticed anything bad so far.

I tried compiling few other random projects, and got a strange error in one of these. Here's what it sais:
`room400.asc(231): Error (line 231): Internal error: Fixup #125 references non-existent import #6`

The source project is available here, in branch "ags4-test-new-compiler": https://github.com/ivan-mogilko/ags-camdemo/tree/ags4-test-new-compiler

---

I'd also like to note something about the code, something I randomly noticed in one of the latest commits [here](https://github.com/adventuregamestudio/ags/pull/1101/commits/59ece2e827c3bd722793b4a0ccaa9caee8956adf#diff-38af0bd9d51e2f2be4fc3c9c71e83f24R217) you are casting a value to have matching types on both sides of > or >= operator., and you cast unsigned type to signed.

IMHO in similar case it's better to do opposite and cast signed parameter to unsigned type:
* if you cast large enough unsigned to signed then it wraps to negative number, and the condition will return wrong result (will be always true),
* casting left signed value to unsigned is safer as you've already tested whether it's high enough (at least > 0) in the first part of the condition.
E.g.
`if (signed_index < 0 || (size_t)signed_index >= array.size())` - the second part won't be evaluated if `signed_index` is lower than zero.
`if (signed_index >= 0 && (size_t)signed_index < array.size())` - the second part will only be evaluated if `signed_index` is greater or eq to zero.

**Comment by fernewelten** [comment]
> room400.asc(231): Error (line 231): Internal error: Fixup #125 references non-existent import #6

Some time ago, I put in the following sanity check for `import` fixups: After the compiler is completely done, every fixup of type `import` must relate to a code cell that points into the `imports` array, and the slot of that `imports` array must contain a non-empty string: the import.  So this check was now triggered for the very first time. 🙂

Turns out that the compiler (both the rewritten and the old one) sometimes generates some bytecode, then rips out that bytecode, then re-inserts it at a different place. 

In particular, this happens in `switch` statements: The switch condition is compiled, then ripped out. All the `case`s are compiled, then the compiler re-inserts the bytecodefor the switch condition to build up a so-called “jump table”. 

Problem is, when the compiler rips out or re-inserts bytecode, it does not take the fixups into account. (AFAIK, the old compiler doesn't either, or at least didn't when I started the rewriting project). So whenever the moved bytecode contains cells that have a fixup , the fixups will get corrupted. 

And look what I found when I traced back the corruption to the program code lines:
```
switch (inventory[game.inv_activated]) {
    case i400Mirror:
        ToggleMirror(!gMagicMirror.Visible);
        break;
    …
}
```
Well, `inventory` is imported. 

So, well, my watchdog code seems to have uncovered a real bug. I'll have to amend the ripping/re-inserting code in a way that the fixups move together with the bytecode.

**Comment by fernewelten** [comment]
> I believe that in similar case it's better to do opposite and cast signed parameter to unsigned type

Thanks for the hint. I believe that I've eliminated or rewritten all the instances of that now.

I'll upload the new fixes in a day or two, when I've incorporated the changes for ripping out/re-inserting bytecode.

**Comment by rofl0r** [comment]
>Turns out that the compiler (both the rewritten and the old one) sometimes generates some bytecode, then rips out that bytecode, then re-inserts it at a different place.
> In particular, this happens in switch statements

oh? when reading through the disassembly of compiled games i had the impression the old compiler never generates jump tables from switch statements, but instead generates a huge ineffienct block of if else conditions.

**Comment by fernewelten** [comment]
> the old compiler never generates jump tables from switch statements, but instead generates a huge ineffienct block of if else conditions.

You're right. It might do better whenever all the cases are discrete constants that are known at compile time, but this isn't implemented. Instead, the generated bytecode always simply checks each case in turn. That isn't a “jump table” approach.

It wouldn't help in this particular case. The cases in this switch statements are inventory objects so they don't have a value that is known at compile time.

**Comment by ericoporto** [comment]
What do these commits change?

**Comment by ivan-mogilko** [comment]
@ericoporto if you scroll up you will see I've been reporting various errors. I am still doing testing of this branch.

**Comment by ivan-mogilko** [comment]
@fernewelten After recent commits, the previously reported errors seem to be fixed, but I found one more testing another script.

It looks like compiler fails handling "this" keyword when it is used in comparisons.
For example:
<pre>
void TestCharacter(this Character*)
{
    if (this == player)
        Display("Test");
}
</pre>
The compiler reports following error on line 3 (with comparison):
`Error (line 3): Expected struct component after '.' but did not find it`

Also, there's a somewhat confusing error message if you do not use array correctly:
<pre>
struct MyStruct
{
    int i;
};

void TestArray()
{
    MyStruct arr[100];
    arr.i = 0;
}
</pre>
The reported error is: `Expected a component of 'MyStruct[100]', found 'i' instead`. I feel like users will be confused by what the "component" means in this context.

**Comment by fernewelten** [comment]
> The reported error is: `Expected a component of 'MyStruct[100]', found 'i' instead`. I feel like users will be confused by what the "component" means in this context.

You're right. That's misleading. 

The important thing to point out instead IMO is that the user put a `.` after something that isn't a struct. That's the main point. Of course, when there isn't any struct then there isn't any component of it, but that's a bit of a side issue.

So I've changed the message to `Expected a struct in front of '.' but found an expression of type 'MyStruct[100]' instead`. Would that be clearer?

**Comment by ericoporto** [comment]
What information is available at that time? Is there information enough to specify with words that `arr` is an array of any type and not struct? In the end I believe the user missed `[]` to actually point an element of the array.

**Comment by fernewelten** [comment]
> Is there information enough to specify with words that `arr` is an array of any type and not struct?

I've implemented this idea now.
```
int i; MyStruct arr[100];
arr.i = 0;
```
--> `Expected a struct in front of '.' but found an array instead`
```
int i, j;
Display(i.j);
```
--> `Expected a struct in front of '.' but found an expression of type 'int' instead`

**Comment by ericoporto** [comment]
Looks great @fernewelten 😄

**Comment by ivan-mogilko** [comment]
@fernewelten  I apologize for delay, there are couple of more notes I was planning to make, hopefully will be able to post them on this week.

**Comment by ivan-mogilko** [comment]
@fernewelten So, I ended up testing the new compiler features that you mention in your first post, and they seem to generally work. But there are three things that are wanted to ask/mention in regards to these. I'll do this in the order of increasing importance.

1. 
> Functions can now be defined in any sequence as long as they are in the same compilation unit. So you can reference a function first and define it afterwards. This makes recursive algorithms much easier to code.
> Forward declarations of functions don't need the "import" keyword if you don't need the functions to be importable elsewhere. Basically, say what you mean: If a function is meant to be importable, use "import"; if it isn't, don't use it.

My question in regards to this is, what is the use case for "forward declaration" given that function order inside a unit is not necessary, and referencing function from a different unit requires an import? Or is this just a side-effect of "import" keyword not required to be present near function declaration?

2. 
> The big exception is import declarations. In import declarations, it is legal to declare managed structs that aren't pointers, and **automatically generated AGS code will actually do that**, so the compiler does not imply any * for things that are declared import. Any pointers must be explicitly written out there.

I am quite concerned about you mentioning generated code having managed structs without pointers in imports, because I do not think that's a right thing. Please clarify where did you see this and what were these imports?

3. 
> Can now have dynamic arrays of elements that contain dynamic arrays.

So, this alerted me, and I also confirmed that managed structs also are allowed have managed pointers in them (and basically any combination of nested normal structs and managed structs).

While syntactically there's nothing wrong in that (that would be quite great to have really), the biggest problem, and also the reason why this was *explicitly disabled* in the old compiler, is that engine is currently *unable to dispose managed objects that are stored in other managed objects*.
The only exception is dynamic array of managed objects, because when such array is created it has a special flag telling that it contains pointers, and so engine can read this flag and decrease reference count for each of array's elements when array itself gets deleted (which in turn leads to these objects deletion if their ref count reaches 0).
As for the regular pointer variables, they have special handling by compiler, where SCMD_MEMZEROPTR is called for a pointer leaving its life scope, which signals engine to reduce reference count.
Unfortunately, compiler cannot do this for pointers stored inside managed objects, as their disposal is performed in time which is unknown at compilation (this is why engine is supposed to handle these).

For example, if you have:
<pre>
managed struct A
{
    int a;
};

managed struct B
{
    A* p;
    int b;
};

B* obj = new B;
obj.p = new A;
</pre>

When the object of type B gets disposed, it cannot unreference managed object A, simply because *information about its contents is not generated during compile time*, and is not present anywhere at runtime. This leads to managed object A staying in the managed memory pool forever.

The easiest way to test this is to create managed objects of large size by key press, and then in game keep pressing it and watch for process memory stats.
For example:
<pre>
managed struct MyStruct
{
    int bigarray[];
};

function on_key_press(eKeyCode key)
{
    MyStruct* s = new MyStruct;
    s.bigarray = new int[1000000];
    s = null;
}
</pre>

In theory `s = null;`, or just exiting this function, should lead to managed memory cleanup. In reality, it only cleans up MyStruct (compiler knows that its pointer leaves life scope), but never disposes bigarray.


We've discussed possible solutions in the past, and the only viable one seems to be to generate type information during compilation and save it along with scripts (either extend script format, or as a separate data item). This information must contain at minimum offsets of pointer variables inside the struct.
Then anytime a user managed object is created with `new` command, the compiler will pass some kind of a type's ID as an extra argument. Whenever disposing a managed struct at runtime, engine will refer to this type information to know if there are any pointers to decrease their reference counts.
Maybe there are easier ways, but I could not figure anything.


So, the question now is, whether we are going to have this in nearby future or not.
If yes, then we may keep this syntax enabled in the new compiler, in anticipation for it fully supported by the engine later.
But if not, then it would be desirable to disable this syntax, at least temporarily.

**Comment by fernewelten** [comment]
Concerning #2, `Object` is `managed`, but `objects[…]` is declared as an array of `Object`, not as an array of `Object *`.

For instance, in your game, the following is exactly what arrives as input to the compiler after pre-processing (empty lines edited out):
```
builtin managed struct Object {

import int Animate(int loop, int delay, RepeatStyle=eOnce, BlockingStyle=eBlock, Direction=eForwards, int frame=0);

import static Object* GetAtScreenXY(int x, int y);
import int GetProperty(const string property);
import String   GetTextProperty(const string property);
import bool IsCollidingWithObject(Object*);
import int MergeIntoBackground();
import int Move(int x, int y, int speed, BlockingStyle=eNoBlock, WalkWhere=eWalkableAreas);
import int RemoveTint();
import int RunInteraction(int);
import int SetPosition(int x, int y);

import int SetView(int view, int loop=0, int frame=0);



import int StopAnimating();
import int StopMoving();
import int Tint(int red, int green, int blue, int saturation, int luminance);
readonly import attribute bool Animating;
import attribute int  Baseline;
import attribute int  BlockingHeight;
import attribute int  BlockingWidth;
import attribute bool Clickable;
readonly import attribute int  Frame;
import attribute int  Graphic;
readonly import attribute int ID;




readonly import attribute int  Loop;
readonly import attribute bool Moving;
readonly import attribute String Name;
import attribute bool Solid;
import attribute int  Transparency;
readonly import attribute int View;
import attribute bool Visible;
import attribute int  X;
import attribute int  Y;
import bool     IsInteractionAvailable(int);
import int SetLightLevel(int light_level);
import bool SetProperty(const string property, int value);
import bool SetTextProperty(const string property, const string value);
readonly import attribute bool HasExplicitLight;
readonly import attribute bool HasExplicitTint;
readonly import attribute int  LightLevel;
readonly import attribute int  TintBlue;
readonly import attribute int  TintGreen;
readonly import attribute int  TintRed;
readonly import attribute int  TintSaturation;
readonly import attribute int  TintLuminance;

import static Object* GetAtRoomXY(int x, int y);

import attribute bool ManualScaling;
import attribute int  Scaling;

readonly int reserved[2];
};
[…]
```

fernewelten:
Concerning #2, for example  `Object` is autodeclared as `import builtin managed struct`, but the individual Objects of the room are autodeclared as `import Object oObject0;` (not `import Object *oObject0;`). 

In the same vein, autogenerated code does `import builtin managed struct Hotspot {…}` and  `import Hotspot hHotspot0;`  (not `import Hotspot *hHotspot0;`).

ivan-mogilko:
> Concerning #2, for example `Object` is autodeclared as `import builtin managed struct`, but the individual Objects of the room are autodeclared as `import Object oObject0;` (not `import Object *oObject0;`).
> 
> In the same vein, autogenerated code does `import builtin managed struct Hotspot {…}` and `import Hotspot hHotspot0;` (not `import Hotspot *hHotspot0;`).

This is very strange, I never noticed that. But these things are indeed pointers, and may be assigned to pointer variables, although I don't know how it works in old and new compiler (does it automatically treat them as pointers or allows to cast to pointers?)

fernewelten:
You can go `Object *MyObject; MyObject = oObject0;` In this case the (old and new) compiler assigns a __pointer to__ `oObject0` to `MyObject`. 

I tripped over that one early in the project. I couldn't explain why the generated code for expressions involving `MyObj`  (as in `Object *MyObj`) was different from code for expressions involving `oObject1` -- until I noticed that `MyObj` is pointered but `oObject1` is not.

loquimur:
As concerns #3, that's what I've just discovered myself -- I've already put a temporarily lock into the compiler that prevents dynamic objects pointing dynamic objects. (The lock will become active with the next push.) 

There's a problem with serialization, too; seemingly the engine can only serialize vectors of pointers. I'm not going to try and change that right now in any way.

In general, I'm not touching the engine in any way; I am treating it strictly as a black box with given behaviour.

fernewelten:
As concerns #3, that's what I've just discovered myself -- I've already put a temporarily lock into the compiler that prevents dynamic objects pointing dynamic objects. (The lock will become active with the next push.)

There's a problem with serialization, too; seemingly the engine can only serialize vectors that consist of pointers only. I'm not going to try and change that right now in any way.

In general, I'm not touching the engine in any way; I am treating it strictly as a black box with given behaviour.

ivan-mogilko:
> There's a problem with serialization, too; seemingly the engine can only serialize vectors that consist of pointers only. I'm not going to try and change that right now in any way.

Could you elaborate on this, by vectors do you mean dynamic arrays or something else? (dynamic arrays may also have basic types, such as ints and floats)

fernewelten:
> what is the use case for "forward declaration"

It's a side effect of having function declarations with/without a body. Internally, the compiler processes the function header (stub) first.  If a body follows it then processes the body in a separate part of the code.

There are some details that make this a bit convoluted:
- A module typically has a header file, say `MODULE.ash`. In this header file, you typically define some function stub, typically with `import`. This `import` is important so that the function can be called from other modules.
- Then there is the module script proper, say `MODULE.asc`. In this script proper, the function is defined with body.
    - So when the module itself is compiled, the compiler gets the header `MODULE.ash` concatenated to the script proper `MODULE.asc`. The compiler will first see an `import` stub and then, later on, "Oops, that `import` is a false alarm, it turns out to be a **local** function really. Do **not** import this!" Moreover, "Other modules do want to import the function, so this `import` turns out to be an `export`".
    - On the other hand, when anything other than that module is compiled, then **only** `MODULE.ash` is concatenated to it and the import stub really **does** mean that it's an import. 

- When you define a `struct`, then it can contain member functions. In contrast to ` C#`, you mustn't define such a member function then and there. You only have a stub in the struct, and the function must be defined with body outside the `struct`.
    - EITHER this struct is defined in `MODULE.asc`. In this case, member function is local and will always stay local. It can't be referenced from other modules. However, the old syntax mandated that the stub in the struct begins with `import`. This doesn't really make sense.
    - OR the struct might be defined in `MODULE.ash`. In this case, the member function is visible from outside and **should** be import and **should** begin with `import`. 

The best I could come up with to reconcile all that was: So you can have stubs with body and stubs without body, and you can have stubs with `import` and stubs without `import`, and the sequence doesn't matter, but they must at least match as best as they can.

ivan-mogilko:
> * A module typically has a header file, say `MODULE.ash`. In this header file, you typically define some function stub, typically with `import`. This `import` is important so that the function can be called from other modules.
>   * So when the module itself is compiled, the compiler gets the header `MODULE.ash` concatenated to the script proper `MODULE.asc`. The compiler will first see an `import` stub and then, later on, "Oops, that `import` is a false alarm, it turns out to be a **local** function really. Do **not** import this! If it is called, use the **local** calling protocol instead of the import calling protocol!" Moreover, "Other modules do want to import the function, so this `import` turns out to be an `export`".

I don't know if this matters, but afaik in ags script there's no such thing as "function export", all functions are automatically exported. For example, you may not have function imports declared in MODULE.ash, but you may instead declare an import right in another script where you call them, and this still works. (This still work in the new compiler too, so everything seem fine here)

fernewelten:
I've made an attempt to put both compilers into the editor side-by-side. I'm nearly there but not quite. Here's what I've got so far.

1. We had split off the second compiler from the first by copying the solution and modifying some names. I remembered that solutions have GUIDs, and of course GUIDs are supposed to be unique. So I reasoned that something would probably get out of whack when I put **both** `Compiler.Lib` and `Compiler2.Lib` into the **same** Editor: 

   So I made a complete list of all the different GUIDs in`Compiler2.Lib`, `Compiler\Compiler2.Lib.vcxproj`, and `Compiler\Compiler2.Lib.Test.vcxproj`. For each such GUID, I generated a new GUID and replaced each occurrence of the  old GUID with the new one in all three files. Thankfully, Notepad++ has a feature where you can find/replace in "all open files" so I could make sure that this was done consistently.

2. The old and the new compiler mostly have the same file names, and so compiling them will yield intermediate files with the same names. I feared that compiling the one compiler would wipe out the files of the other compiler, but when both are in the Editor then the Editor or the linker or the building process might need both:

   So to be on the safe side, I copied the folder `Compiler.Lib` to `Compiler2.Lib`. I modified the `Compiler2` solution so that that its name is now `Compiler2.Lib` and all its files go to folder `Compiler2.Lib`.

3. In the Editor, I removed `Compiler.Lib` from the Native Libraries; then I added both `Compiler.Lib` and `Compiler2.Lib` to Native Libraries anew.

4. By now the new compiler has hardly any global variables left; it's all in objects and classes. So there are hardly any name collisions when the old and the new compilers are linked together. But I had to rename `char ccCopyright[]` to `char ccCopyright2[]`. That variable is unused; seemingly, it's  only in the compiler so that this string will be put into the respective library or executable.

5. The new compiler will need to be called in C++  as an alternative to the old compiler: 

   So I renamed `ccCompileText()` to `ccCompileText2()` in the new compiler. The `#include` file of the new compiler will need to be included in parallel to the `#include` file of the old compiler, so I changed its guarding define from `__CS_COMPILER_H` to `__CS_COMPILER2_H`.

6. Users will need a way to switch between compilers: 

   So in the Editor, I added `AGS.Types/Settings.cs` to add a `bool` to the `Compilers` section of the `General Settings` tab.  

    **That's probably not enough by far:** The `General settings` tab is bound to be read/written somewhere and that code doesn't know about the new `bool` yet. Also, I'm not convinced that I've found the ideal name for the `bool` in the first place (`ExperimentalCompiler`).

7. As far as I can see, there's just **one** place of the Editor from which the compiler is called, namely `Editor\AGS.Native\ScriptCompiler.cpp`. I added the new `#include` at the top of the file. I left the call to the old compiler completely untouched, but I put an `if()` around it in order to alternatively call the new compiler. Well, I _tried hard_ to keep the call to the old compiler *completely* untouched, but whenever I put in the closing `}` of the enclosing `if()`, Visual Studio absolutely insisted on formatting it completely anew.

The Editor compiles. (If there are old build files, it might need a `rebuild` to make it work.) 

I wrote a rudimentary game with AGS code that is legal in the new compiler but illegal in the old one; by that I verified that either the old or the new compiler is called depending how the switch is set in General Settings.

ivan-mogilko:
Yes, it's best to find some kind of generic name for the compiler selection, perhaps calling it "script compiler" of type String and provide a list of compiler names (need to decide how to call them, like their "official" names).

Also, I think that new compiler should be selected by default, otherwise 99% of users will never use it.

> The General settings tab is bound to be read/written somewhere and that code doesn't know about the new bool yet

All these properties are read and written by generic serialization code that uses reflection, so any new property is hooked up automatically.

fernewelten:
> All these properties are read and written by generic serialization code that uses reflection, so any new property is hooked up automatically.

Cool!

ivan-mogilko:
But I feel like this property may be beyond the scope of this PR. Me or someone else might be able to help with it later, for now internal switch might be enough (defaulted to the new compiler).

fernewelten:
So it now defaults to the new compiler.

ivan-mogilko:
@fernewelten I am testing this now.

Would it be possible to make new compiler define some macro to test in script, like real compilers let you test for compiler version?

ivan-mogilko:
Regarding the solution, some things are broken there, which I am afraid may be complicated to explain, but I'll try. In the worst case I could fix that myself after merging pr, as it's still possible to build things by manually starting each project build in an order.

1. It looks like solution is saved in a different version of MSVS (2017 maybe?), we still require 2015 to be used at the moment. Thankfully MSVS 2015 was able to open them.
2. There are two new Solution Configurations: "Debug MD" and "Release MD" that are not needed. Note that there's a difference between "Solution Configuration" and individual "Project Configuration". Common.lib, Compiler and Compiler2.lib in the editor are built in _MD configurations.
3. On contrary "Debug" and "Release" solution configs had Common, Compiler and Compiler2 libs excluded from build for some reason. They need Build checkbox set in configuration manager, and their project config should be Debug_MD and Release_MD correspondingly.
4. AGSNative project should have Compiler and Compiler2 in dependencies so that they are built automatically when the AGSNative is built.
5. Compiler2 has incorrect "Target Name" setting in Release and Release_MD configs, should be "Compiler2" and "Compiler2_md" respectively.

ivan-mogilko:
Last thing, you did not really set default value for compiler to true, only [DefaultValue] attribute which is only a cosmetic change (it highlights non-default values in the editor). `private bool _experimentalCompiler` is still initialized as false.

ericoporto:
> Would it be possible to make new compiler define some macro to test in script
I think macros are already preprocessed before the code "arrives" at the compiler, so this would be in the Editor code, wouldn't?

ivan-mogilko:
> I think macros are already preprocessed before the code "arrives" at the compiler, so this would be in the Editor code, wouldn't?

Oh right. But then maybe it's compiler library that should provide the macro itself, as it's not Editor's task, and it won't even know about what compiler is when it becomes a separate tool.

ericoporto:
> I'd make a separate function nearby, to not mix these up.

Nearby the compiler you mean? I think that is cool, if it's nearby the other function it works too. I kinda wanted to make a PR to move build to it's own menu (in ags4), so target selection (datafile, windows, linux) moves to there, and then I think I will move this setting too after this is merged. :) Anyway, it's just an idea for now, I am kinda in wait mode for the moment until the new compiler gets in.

ivan-mogilko:
>> I'd make a separate function nearby, to not mix these up.
> Nearby the compiler you mean? I think that is cool, if it's nearby the other function it works too

I am not sure if we are talking about same thing here...

ivan-mogilko:
@fernewelten I hope it's okay with you, I decided to speed things up a bit in case you are busy, and apply the necessary solution fixes myself. Compiler itself seem working (I tested it with all the previously used projects).

If you have any more additions, fixes, and so on, please open new PRs.

Thank you for this work, now it's part of ags4 branch :).

fernewelten:
@ivan-mogilko 

Thanks for all the work, that's fantastic!

Here's what my Visual Studio Editor shows for the Compiler projects:
image

So as far as my Visual Studio is concerned, **both** Compiler.lib and Compiler2.lib are shown as Visual Studio **2015** projects, not 2017 projects. They seem to have slight differences in the data format inside but still both seem to be recognized as Visual Studio 2015. 

So that must be the explanation that Visual Studio 2015 can open both sets of project files.

ivan-mogilko:
Ah, I think I broke CI with this, I suppose this is because it uses CMake scripts? Will try to fix these using preexisting compiler lib as an example...

PR #1102, opened 2020-07-08
Be explicit when converting AGS String to char*

ericoporto:
From past conversations I think we want to be explicit when converting AGS Strings to `const char*`, this PR picks up a previous commit in a different branch where Sonneveld did this, and additionally put the explicit conversion in more places.

### DIFF
--- a/Common/ac/common.cpp
+++ b/Common/ac/common.cpp
@@ -25,5 +25,5 @@ void quitprintf(const char *fmt, ...)
     va_start(ap, fmt);
     String text = String::FromFormatV(fmt, ap);
     va_end(ap);
-    quit(text);
+    quit(text.GetCStr());
 }
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -376,7 +376,7 @@ void GameSetupStruct::read_room_names(Stream *in, GameDataVersion data_ver)
             roomNumbers[bb] = in->ReadInt32();
             pexbuf.Read(in, STD_BUFFER_SIZE);
             roomNames[bb] = new char[pexbuf.GetLength() + 1];
-            strcpy(roomNames[bb], pexbuf);
+            strcpy(roomNames[bb], pexbuf.GetCStr());
         }
     }
     else
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -96,7 +96,7 @@ ScriptAudioClip* GetAudioClipForOldStyleNumber(GameSetupStruct &game, bool is_mu
 
     for (size_t i = 0; i < game.audioClips.size(); ++i)
     {
-        if (clip_name.Compare(game.audioClips[i].scriptName) == 0)
+        if (clip_name.Compare(game.audioClips[i].scriptName.GetCStr()) == 0)
             return &game.audioClips[i];
     }
     return nullptr;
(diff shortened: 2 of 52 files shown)


ivan-mogilko:
> From past conversations I think we want to be explicit when converting AGS Strings to const char*

Could you remind what was the reasoning behind this and post it in the PR's description?

Regarding the code itself, it misses the most important part: you have to remove `operator const char *()` from String class, otherwise nothing would prevent programmers from automatic conversion in the future. After you do that all cases will be identified and reported by compiler.

ericoporto:
I think it's mostly about how it should be done, either always implicit or always explicit. Currently, both ways are done in our code and it's not clear what approach should be used when writing new code.

ivan-mogilko:
> I think it's mostly about how it should be done, either always implicit or always explicit. Currently, both ways are done in our code and it's not clear what approach should be used when writing new code.

IIRC making it explicit was the point of original sonneveld's changes. These changes make no sense otherwise, as if not enforced there will be a mixture of styles in the future code.

ivan-mogilko:
Sorry, I think I misunderstood your last comment. I will make a better reply in a short while.

ivan-mogilko:
What I meant to say,...

> I think it's mostly about how it should be done, either always implicit or always explicit. Currently, both ways are done in our code and it's not clear what approach should be used when writing new code.

I'd rather double check that this was exactly the reason behind sonneveld's proposal, as it sounds more like something that belongs to code-style.
The approach used in current code IMHO is pretty clear and logical on its own: use implicit conversion whenever possible, and explicit when implicit does not work. For example, in printf-like variadic functions you cannot pass String as one of the arguments, because compiler cannot guess to convert it to const char* on its own.

Furthermore, I have to underline again, that if you want to *enforce* explicit conversion, then you *must* remove or disable the conversion operator, which is not done in this PR. Without that this change is meaningless, as nothing will prevent user from ignoring this convention.

> Alternatively, we could just make the operator explicit.

The conversion operator was added primarily for implicit casting. For explicit casting there's GetCStr(), and having both operator and function for explicit cast seem redundant.
According to the above link, explicit convesion operator requires coder to put an explicit cast. Compare following ways to convert String to const char*:

`mystring.GetCStr();` - using existing method
`static_cast<const char*>(mystring)` - proper C++ cast
`(const char*)(mystring)` - C-style cast (probably will work too)

ericoporto:
Is it reasonable to consider `const char *` is used when there's no intend of modifying the original string and `String*` is used when it's desired to modify the string? I am trying to understand why the String own functions operate with `const char *` instead of `String*`. Sonneveld was pursuing threading so I guess he was trying to figure where it could cause trouble.

I will close this for now and try to grasp the String situation of AGS.

ivan-mogilko:
> Is it reasonable to consider const char * is used when there's no intend of modifying the original string and String* is used when it's desired to modify the string? I am trying to understand why the String own functions operate with const char * instead of String*.

Not quite, if it was about read-only operation you could pass `const String&` or `const String*` too.

The purpose of having `const char*` as an argument is to not allocate a redundant String object if you are passing a raw char array of string literal.
E.g.:
<pre>
mystring.Compare( "some literal text" );
</pre>
If `Compare` accepts `const String&`, then a new String object will be allocated with a copy of that text, only for the purpose of passing it into function.

In some circumstances it is viable to have two functions, one accepting `const char*` and another accepting `const String&`, if that allows to do some optimization (but we don't have many examples of that here).

Issue #640, opened 2019-02-19
Add script function for writing into a console log


alkhimey:
The script debugging features of AGS are very basic. It is possible to break the script on certain lines watch the call stack and and watch room information, but there is not way to see the values of script variables.

Implementing a watch table might be hard task. Alternative easier suggestion is to add a debug console output command. This will allow script developers to debug by "printing" the values of the variables which is a very reasonable technique.

ivan-mogilko:
That is an interesting suggestion. Engine already has a logging system which may configure the "message type" and where it logs to (file, in-game console, etc).
There's also a hook to AGS Editor output panel if game was run from IDE. I believe it should be possible  write such messages there too.

mgambrell:
we had to add this ourselves while doing console ports. it's essential. implemented as a private plugin though.

messengerbag:
I'm sure most big AGS projects have some sort of logging system implemented. Here's the one for the AGS Awards client:

https://bitbucket.org/agsa/ags-awards-source/src/master/AGS%20Project/Log.ash
https://bitbucket.org/agsa/ags-awards-source/src/master/AGS%20Project/Log.asc

It would certainly be convenient to output log messages to the IDE console or built-in debug console.

ericoporto:
A simple print string to stdout would be pretty useful, specially when calling the Engine from command line on Windows. I can currently print strings through plug-in on Linux to the command line, but I wasn't able to figure out how it's done on Windows.

ivan-mogilko:
>  I wasn't able to figure out how it's done on Windows.

It's done about the same way, except engine is run as a non-console application, which means it is not attached to console by default, this is something that has to be done (can't tell if it's possible to do in plugin though, but maybe).


Since we'are at this, I think in general we need to decide on API and what it expected to do. For example, I doubt that having an explicit "print to stdout" function is ideal, because on some systems stdout may not be easily accessible, so there has to be something more abstract that translates to actual output depending on a system.

BTW, if I understand correctly, the original proposal was to make a proper in-game *console* where you could print to from script. AGS currently has one, but it has limited use, and also serious convenience issue because it has to rely on provided game fonts and is rendered in game's native resolution (could be pretty low).

ericoporto:
[...]

PR #1103, opened 2020-07-10
3.5.1: Configurable log options

ivan-mogilko:
This is done as a preliminary step for #640.

**As a preface**, engine logging system in a nutshell works like this:
* each message has a "verbosity level" ranging from "debug" to "fatal error".
* there may be a number of message groups, which allow to separate messages related to a component or engine's function (main, script, spritecache, managed pool);
* there may be a number of outputs (stdout, log file, in-game console); each output has its own filter settings, that define which groups & level of messages to let through.

When a message is sent to logging system it tries each registered output implementation, and if its filter allows, message is being printed.

---

**What this PR does**: it actually allows to configure these filters by user (player).

A new section called "[log]" is introduced to the user config. It may have a number of entries - one for each log output - which define a list of filters for that output. These entries have a common syntax:

`output_id = group1:level, group2:level, etc`

where `output_id` is output's name, and entry value is a list of "group name" : "message level" pairs.

Group names may be full names or short 1-letter IDs. **NOTE:** in the latter case these IDs may be put in groups but must be preceded by '+' symbol (e.g. `+msco` - prints groups `m`,`s`,`c` and `o`).
Message level (verbosity) may be level name, or corresponding number (see below).

Possible outputs at this moment are:
* `stdout` - standard output (system console, etc)
* `file` - regular log file
* `console` - in-game console

Possible log groups:
* `main` (m) - default group for most of messages
* `script` (s) - script API messages
* `sprcache` (c) - verbose sprite cache messages (currently not available unless engine is built with DEBUG_SPRITECACHE)
* `manobj` (o) - managed objects / pool log, very verbose! (currently not available unless engine is built with DEBUG_MANAGED_OBJECTS)

Possible message levels:
* `alert` (1) - most important messages (like engine version info and startup/shutdown notification)
* `fatal` (2) - fatal errors
* `error` (3) - any errors
* `warn` (4) - warnings
* `info` (5) - general notifications
* `debug` (6) - verbose debugging messages
* `all` - all messages (which equals to `debug` level actually)

---

**Examples of config:**

<pre>
[log]
stdout = main:info, script:error
file = main:all, script:all
</pre>
Print only important notifications and script API errors to stdout (system console), print everything from main and script groups into file, and use default settings for in-game console.

### DIFF
--- a/Common/debug/debugmanager.cpp
+++ b/Common/debug/debugmanager.cpp
@@ -58,6 +58,21 @@ void DebugOutput::SetGroupFilter(DebugGroupID id, MessageType verbosity)
         _unresolvedGroups.insert(std::make_pair(id.SID, verbosity));
 }
 
+void DebugOutput::SetAllGroupFilters(MessageType verbosity)
+{
+    for (auto &group : _groupFilter)
+        group = verbosity;
+    for (auto &group : _unresolvedGroups)
+        group.second = verbosity;
+}
+
+void DebugOutput::ClearGroupFilters()
+{
+    for (auto &gf : _groupFilter)
+        gf = kDbgMsg_None;
+    _unresolvedGroups.clear();
+}
+
 void DebugOutput::ResolveGroupID(DebugGroupID id)
 {
     if (!id.IsValid())
--- a/Common/debug/debugmanager.cpp
+++ b/Common/debug/debugmanager.cpp
@@ -82,7 +82,7 @@ bool DebugOutput::TestGroup(DebugGroupID id,  MessageType mt) const
     DebugGroupID real_id = DbgMgr.GetGroup(id).UID;
     if (real_id.ID == kDbgGroup_None || real_id.ID >= _groupFilter.size())
         return false;
-    return (_groupFilter[real_id.ID] & mt) != 0;
+    return (_groupFilter[real_id.ID] >= mt) != 0;
 }
 
 DebugManager::DebugManager()
--- a/Common/debug/debugmanager.h
+++ b/Common/debug/debugmanager.h
@@ -64,7 +64,7 @@ struct DebugGroup
 class DebugOutput
 {
 public:
-    DebugOutput(const String &id, IOutputHandler *handler, MessageType def_verbosity = kDbgMsgSet_All, bool enabled = true);
+    DebugOutput(const String &id, IOutputHandler *handler, MessageType def_verbosity = kDbgMsg_All, bool enabled = true);
 
     String          GetID() const;
     IOutputHandler *GetHandler() const;
@@ -110,7 +110,7 @@ public:
     DebugGroup RegisterGroup(const String &id, const String &out_name);
     // Registers output delegate for passing debug messages to;
     // if the output with such id already exists, replaces the old one
-    PDebugOutput RegisterOutput(const String &id, IOutputHandler *handler, MessageType def_verbosity = kDbgMsgSet_All, bool enabled = true);
+    PDebugOutput RegisterOutput(const String &id, IOutputHandler *handler, MessageType def_verbosity = kDbgMsg_All, bool enabled = true);
     // Unregisters all groups and all targets
     void UnregisterAll();
     // Unregisters debugging group with the given ID
--- a/Common/debug/debugmanager.h
+++ b/Common/debug/debugmanager.h
@@ -73,6 +73,10 @@ public:
     void            SetEnabled(bool enable);
     // Setup group filter: either allow or disallow a group with the given ID
     void            SetGroupFilter(DebugGroupID id, MessageType verbosity);
+    // Assign same verbosity level to all known groups
+    void            SetAllGroupFilters(MessageType verbosity);
+    // Clear all group filters; this efficiently disables everything
+    void            ClearGroupFilters();
     // Try to resolve group filter unknown IDs
     void            ResolveGroupID(DebugGroupID id);
     // Test if given group id is permitted
--- a/Common/debug/out.h
+++ b/Common/debug/out.h
@@ -80,30 +80,27 @@ namespace Common
 enum MessageType
 {
     kDbgMsg_None                = 0,
-    // Initialization messages, notify about certain engine components
-    // being created and important modes initialized.
-    kDbgMsg_Init                = 0x0001,
-    // Debug reason is for arbitrary information about events and current
-    // game state
-    kDbgMsg_Debug               = 0x0002,
-    // Warnings are made when unexpected or non-standart behavior
-    // is detected in program, which is not immediately critical,
-    // but may be a symptom of a bigger problem.
-    kDbgMsg_Warn                = 0x0004,
+    // Fatal errors are ones that make program abort immediately
+    kDbgMsg_Fatal               ,
     // Error messages are about engine not being able to perform requested
     // operation in a situation when that will affect game playability and
     // further execution.
-    kDbgMsg_Error               = 0x0008,
-    // Fatal errors are ones that make program abort immediately
-    kDbgMsg_Fatal               = 0x0010,
+    kDbgMsg_Error               ,
+    // Warnings are made when unexpected or non-standart behavior
+    // is detected in program, which is not immediately critical,
+    // but may be a symptom of a bigger problem.
+    kDbgMsg_Warn                ,
+    // Initialization messages, notify about certain engine components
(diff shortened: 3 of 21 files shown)


ericoporto:
This will be really useful to help debugging AGS games! I have a minor nitpick for future proof, in case there's ever an idea to pass a parameter like `--log-stdout=script:error`,  the colon `:` character has special meaning in both `bash` and `cmd.exe`, so I would use something like `_` .

ivan-mogilko:
> I have a minor nitpick for future proof, in case there's ever an idea to pass a parameter like `--log-stdout=script:error`, the colon `:` character has special meaning in both `bash` and `cmd.exe`, so I would use something like `_` .

Underscore may be a part of name. I think it's possible to replace with another '=', because only first one separates key and value in ini format.

ericoporto:
Sorry, I was wrong. I just wrote a [mini program](https://github.com/ericoporto/quicktest/blob/master/main.cpp) and the `:` is fine. Let's keep it, it looks nicer too!

ericoporto:
Hey, in the context of #640 , user generated messages would be a new group, a new level (between info and warn) or neither?

ivan-mogilko:
> user generated messages would be a new group (eg: dev or user), a new level (between info and warn) or neither?

I was thinking about renaming current "script" group to "game" (because it's related to game logic), and giving "script" group to messages from user script.

rofl0r:
is there something like an ultra-verbose debug level, i.e. "give me all log messages" ?

ivan-mogilko:
@rofl0r 
> is there something like an ultra-verbose debug level, i.e. "give me all log messages" ?

Do you mean all message levels from all groups? I forgot to add an quick option for that, but I will do.

@ericoporto actually, since you mentioned that, I realized that command-line arguments would also be needed.

ericoporto:
> Do you mean all message levels from all groups? I forgot to add an quick option for that, but I will do.
Maybe if it's only the message level, without specifying the log group it applies to all?

> @ericoporto actually, since you mentioned that, I realized that command-line arguments would also be needed.

I think (not sure) there's already a command line argument there, so you do add one with these capabilities, then that one may be removed

rofl0r:
> Do you mean all message levels from all groups? I forgot to add an quick option for that, but I will do.

yes

> Maybe if it's only the message level, without specifying the log group it applies to all? 

sounds like a good idea


btw i'd prefer if this was a commandline switch rather than a config file option

morganwillcock:
> btw i'd prefer if this was a commandline switch rather than a config file option

I had the same thought.

ivan-mogilko:
Do you mean, having a command line option to pair the config, or have it instead of config (and not have a config setting at all)?

ericoporto:
I imagine the log can config is more comfortable on Windows when you are running your game by double clicking the executable and the command line is more comfortable when invoking the game in Linux system where you run the engine from the command line (double click execution of binaries from file navigator is not a thing in Linux since this belongs to the launcher).

rofl0r:
> Do you mean, having a command line option to pair the config, or have it instead of config (and not have a config setting at all)?

yeah, imo command-line only is preferable. the user interested in log output can run `./game.exe --help` and read the usage output, if it is a config file option he's got to study (and find) the online manual for config files.

ericoporto:
@rofl0r do you have suggestions for how to write the log help concisely? I was thinking on it and could only figure out text that was more appropriate to a man entry.

rofl0r:
@ericoporto something like:
```
--log-level=value
  1: fatal errors only
  2: medium
  3: debug
--log-subsys=[mscp]
  specify one or more log subsystems to be logged (default: all)
  m: main
  s: script
  ...
--log-output=value
  specify where log output is displayed (default: stdout)
  stdout: log to stdout
  stderr: log to stderr
  console: log to game console
  file:name.log: log to name.log
```

ericoporto:
Engine currently log to `~/.local/share/ags.log` , I would not change this on this PR, if the log goes to stdout the person can redirect it to a file. Also current default I think is info on main and error on script.

ivan-mogilko:
IMO there is at least one reason to keep config options as alternative: there are ports that are not normally run from command line (mobile).

ericoporto:
Yes that too. On Android the stdout log should show up in logcat.

ivan-mogilko:
Looking at rofl0r's cmdline proposal above, it defines same subsystem/level setup for any output, as opposed to each output having its own message filter, which exists now. Initially I was basing this on some of the existing logging systems that allow to pass different message groups into different targets.

Idk if it's generally necessary for AGS, but there are special cases, like in-game console that normally was used to print all kinds of verbose stuff, like room object state changes, and a "warnings.log" file that is used to output any kind of script warnings when game is run from the editor (bringing user attention to potential mistakes in game).

ivan-mogilko:
Okay, I will try to make something up soon (there has to be bit more changes to the code), but one thing that concerns me about above suggestion is that "log-level" default is "nothing", which means that if a user simply runs a game from terminal without parameters, they won't get any messages whatsoever.

rofl0r:
> above suggestion is that "log-level" default is "nothing", which means that if a user simply runs a game from terminal without parameters, they won't get any messages whatsoever?

well, that's probably not a wise default, at least for non-console-like platforms like android, i suppose it should be 1 for platforms that have a terminal.

ericoporto:
`--log-stdout=main:info script:error` or `--log-stdout=m:i s:e` should work as alternative command line parameters to specify a level per loggroup per log sink (stdout, file, console).

ivan-mogilko:
> `--log-stdout=main:info script:error` or `--log-stdout=m:i s:e` should work as alternative command line parameters to specify a level per loggroup per log sink (stdout, file, console).

In the above, how do you know where the command arguments end?

ericoporto:
Yes, you can enclose in double quotes. You can also accept multiple ([item 9](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)) `--log-stdout=main:info` `--log-stdout=script:error` passed at same time. This [so details both](https://stackoverflow.com/questions/43195553/how-to-pass-multiple-parameters-for-the-same-commandline-argument).  Maybe is best to not accept multiple arguments in the parameter unless it's enclosed by double quotes. If it's desired for some reason, than an option is to require comma as separator.

ivan-mogilko:
The problem is that engine still does not have a proper generic parsing of command line; for that reason I'd rather stick to one simple implementation for this moment.

ericoporto:
I had a pleasant experience with `cxxopts` here https://github.com/ericoporto/ags/blob/ags4-fernewelten-agscc/Compiler/agscc.cpp

rofl0r:
> I had a pleasant experience with cxxopts

is that an external library ? if so it's best to avoid it for simplicity

> I'd rather stick to one simple implementation

the simple and generic implementation is getopt(3p) or [getopt_long](https://linux.die.net/man/3/getopt_long) . the latter is a bit less portable but used by almost all command line tools in the unix world.

as far as parsing is concerned, in my proposal you'd parse one or more message groups using a single character representing a group, no need for separator characters, in which case a simple `switch(optarg[x]) { case 'm': message_group_m = true; break; ...` would be sufficient (iterating over each char in the optarg) . the only option that requires a separate parsing step is the suggested --log-output option if used with "file", in which case one would check whether the option argument starts with "file:" and then start parsing the actual filename from position 5.

ivan-mogilko:
AFAIK getopt is not supported on Windows.

ivan-mogilko:
So, here's an experimental variant done in a kind of smart-arse way. It has *only one* command line option which is parsed similarily to the config, except the option's name is formed slightly differently:

`--log-OUTPUT=group:verbosity,group,verbosity,etc`

Groups may be defined by both full names or letters or groups of letters. Nuance is that to make parsing easier letter groups must be preceded by special letter '+' (I can change it to something else).
Verbosity may now also be defined by either name or number.

Examples of command line arguments:
`--log-file=main:error` - prints into file
`--log-stdout=all:3` - prints error level to stdout
`--log-stdout=all` or `--log-stdout=all:all` - prints everything possible to stdout.

Additionally, log file path may be now specified both in config and command line:
`--log-file-path=./1.txt`
or
<pre>
[log]
file-path = ./1.txt
</pre>

ivan-mogilko:
Added help, if it's fine I can merge this.

ericoporto:
I verified on Linux and for me it appears fine.

ivan-mogilko:
Ammended with examples.

PR #1104, opened 2020-07-15
Engine: addresses positioning bug with custom text window GUI

### LINKED ISSUE #1098: Textual Overlays change position if run in Sierra-speech game
**Describe the bug**
If the game uses Sierra-with-background speech style and TextWindow gui for text background, changing Overlay's text in rep-exec function results in it changing it's position unexpectedly.

This is easiest to see if using one of the "Typewriter" modules that imitate text typed letter by letter on overlays etc.

**AGS Version**
Seems like a fairly old bug, reproduced from AGS 3.5.0 to 3.2.1 (did not test in older).

**Game**
Following test game uses TypedText module to print text on overlay.
Compiled windows version: [test_sierrasp_overlay--win.zip](https://github.com/adventuregamestudio/ags/files/4834021/test_sierrasp_overlay--win.zip)
Project source: [test_sierrasp_overlay.zip](https://github.com/adventuregamestudio/ags/files/4834023/test_sierrasp_overlay.zip)

**To Reproduce**
Unfortunately I was not able to reproduce this with minimal set of commands. I know that it happens when you have a previously created Overlay and change it's text using SetText method, but I was not able to achieve stable reproduction when doing this in a blank script for some reason (maybe I did something wrong).
It's 100% reproducing if you will use one of the two existing Typewriter modules.

1. Create a new game; it's best to use Sierra template as it already has got a TextWindow set up.
2. In General Settings change speech style to "Sierra with background".
3. In General Settings set Custom text window GUI (it should be 7 if you use Sierra template).
4. Import Typewriter module,
* either Phemar's one here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=48093.0
* or mine here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=54450.0
5. Run a simple typewriter on overlay upon some event (key press etc). E.g. if using my module:
<pre>
Overlay.Typewriter(60, 60, 0, eFontNormal, "This is a test of a typewriter text", eBlock);
</pre>
6. Observe Overlay changing position every tick, as if it's X and Y were decrementing and looping in some range.

**Expected behavior**
Overlay must retain it's position on screen unless script explicitly sets new one.

AlanDrake:
Introduces an internal x,y offset so the original coordinates are not lost due to adjustments for image borders.

Fixes #1098

### DIFF
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -172,6 +172,10 @@ int _display_main(int xx,int yy,int wii,const char*text,int blocking,int usingfo
     // inform draw_text_window to free the old bitmap
     const bool wantFreeScreenop = true;
 
+    // may later change if usingGUI, needed to avoid changing original coordinates
+    int adjustedXX = xx;
+    int adjustedYY = yy;
+
     if ((strlen (todis) < 1) || (strcmp (todis, "  ") == 0) || (wii == 0)) ;
     // if it's an empty speech line, don't draw anything
     else if (asspch) { //text_color = ds->GetCompatibleColor(12);
@@ -190,7 +194,7 @@ int _display_main(int xx,int yy,int wii,const char*text,int blocking,int usingfo
 
         if (drawBackground)
         {
-            draw_text_window_and_bar(&text_window_ds, wantFreeScreenop, &ttxleft, &ttxtop, &xx, &yy, &wii, &text_color, 0, usingGui);
+            draw_text_window_and_bar(&text_window_ds, wantFreeScreenop, &ttxleft, &ttxtop, &adjustedXX, &adjustedYY, &wii, &text_color, 0, usingGui);
             if (usingGui > 0)
             {
                 alphaChannel = guis[usingGui].HasAlphaChannel();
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
My suggestion would be to pass offsets as parameters into `add_screen_overlay` instead of setting them externally, because that seem to be a part of common case (define how bitmap is drawn relative to user coords), and just generally a better style.

criezy:
I am not sure if this is the best place to comment on this, but the changes in this pull request seem to introduce a regression causing the text boxes to be shifted to the right in some games when they use a custom GUI and `asspch` is 0.

In `_display_main` you only changed one of the two calls to `draw_text_window_and_bar()` to use `adjustedXX` and `adjustedYY` instead of `xx` and `yy` (it is changed for the `asspch != 0` case but not for the `asspch == 0` case). As a result when `asspch is 0, the xx and yy get adjusted for the additional border, but then it is drawn at the original position that was stored in `adjustedXX` and `adjustedYY`.

I found this while debugging the issue in the new AGS ScummVM engine, so I can't say for sure the same issue occur here, but I don't see why it would not. In ScummVM I fixed the issue by also passing `adjustedXX` and `adjustedYY` instead of `xx` and `yy` to `draw_text_window_and_bar()` for the `asspch == 0` case.

ivan-mogilko:
@criezy what you say makes sense, feel free to open pr with bug fix in this repo, or if you dont want I will apply your patch later today.

PR #1106, opened 2020-07-18
Script should serialize back empty imports as empty strings.

ericoporto:
- When engine resolves script imports, empty string imports are expected to be `nullptr`.
  [`Engine/script/cc_instance.cpp#L1591`](https://github.com/adventuregamestudio/ags/blob/c52217180969576717173777d8ac195dd4347047/Engine/script/cc_instance.cpp#L1591)
- This works because when reading script objects, empty strings (a single `\0` character) are replaced by a `nullptr`.
  [`Common/script/cc_script.cpp#L39`](https://github.com/adventuregamestudio/ags/blob/c52217180969576717173777d8ac195dd4347047/Common/script/cc_script.cpp#L39)
- When writing a script object, we don't treat the `nullptr` case, so if you read a script object, and want to write it back to a file, it breaks on empty strings.
  [`Common/script/cc_script.cpp#L211`](https://github.com/adventuregamestudio/ags/blob/c52217180969576717173777d8ac195dd4347047/Common/script/cc_script.cpp#L211)
- Currently Compiler only produces empty string for each unused import.

The fix proposed here is to simply treat this case of `nullptr` when writing the script object :

```C++
if(imports[n]== nullptr){
    out->WriteArray("\0", 1, 1);
} else {
    out->WriteArray(imports[n], strlen(imports[n]) + 1, 1);
}
```

This should not change nothing currently but is needed for a compiled script that is serialized, and then de-serialized, can be serialized back, which happens when updating the script object of a `room.crm` file. 

If this change is not applied, the original line `out->WriteArray(imports[n], strlen(imports[n]) + 1, 1)` will crash because `imports[n] == nullptr` when `n` leads to an de-serialized empty string.

This or other fix so a ccScript can be written back once read is necessary for the future compiler tools.

### DIFF
--- a/Common/script/cc_script.cpp
+++ b/Common/script/cc_script.cpp
@@ -45,6 +45,15 @@ void freadstring(char **strptr, Stream *in)
     strcpy(strptr[0], ibuffer);
 }
 
+void fwritestring(const char *strptr, Stream *out)
+{
+    if(strptr == nullptr){
+        out->WriteByte(0);
+    } else {
+        out->Write(strptr, strlen(strptr) + 1);
+    }
+}
+
 ccScript *ccScript::CreateFromStream(Stream *in)
 {
     ccScript *scri = new ccScript();
@@ -208,15 +217,15 @@ void ccScript::Write(Stream *out) {
     }
     out->WriteInt32(numimports);
     for (n=0;n<numimports;n++)
-        out->WriteArray(imports[n],strlen(imports[n])+1,1);
+        fwritestring(imports[n], out);
     out->WriteInt32(numexports);
     for (n=0;n<numexports;n++) {
-        out->WriteArray(exports[n],strlen(exports[n])+1,1);
+        fwritestring(exports[n], out);
         out->WriteInt32(export_addr[n]);
     }
     out->WriteInt32(numSections);
     for (n = 0; n < numSections; n++) {
-        out->WriteArray(sectionNames[n], strlen(sectionNames[n]) + 1, 1);
+        fwritestring(sectionNames[n], out);
         out->WriteInt32(sectionOffsets[n]);
     }
     out->WriteInt32(ENDFILESIG);


ivan-mogilko:
This sounds reasonable, but could you give a step by step example of how to reproduce the error?

ivan-mogilko:
This sounds like a reasonable safety fix, but could you give an example of how this error may be reproduced?

ericoporto:
> This sounds like a reasonable safety fix, but could you give an example of how this error may be reproduced?

Sure! Here is a minimum way to reproduce this is shown below:
```C++
#include "script/cc_script.h"
#include "util/filestream.h"
#include "util/file.h"
#include "util/stream.h"

int main()
{
    ccScript scriptObj1 = ccScript::ccScript();
    char arrayOfchar[1];
    arrayOfchar[0] = '\0';
    char* buffer[1];
    buffer[0] = arrayOfchar;
    scriptObj1.imports = buffer;
    scriptObj1.numimports = 1;
    AGS::Common::String scriptFileName = AGS::Common::String("script.o");
    AGS::Common::FileStream* script_file_stream = (AGS::Common::FileStream*) AGS::Common::File::OpenFileWrite(scriptFileName);
    scriptObj1.Write(script_file_stream);
    script_file_stream->Close();
    script_file_stream = (AGS::Common::FileStream*) AGS::Common::File::OpenFileRead(scriptFileName);
    ccScript* scriptObj2 = ccScript::CreateFromStream(script_file_stream);
    script_file_stream->Close();
    script_file_stream = (AGS::Common::FileStream*) AGS::Common::File::OpenFileWrite(scriptFileName);
    scriptObj2->Write(script_file_stream); // <<-- should crash on this line!
    script_file_stream->Close();
    return 0;
}
```
This can be easily tested by overriding ags main, like this: https://github.com/adventuregamestudio/ags/compare/master...ericoporto:test-crash-write

> PS. BTW, same freadstring is used to read exports. Do these also need similar fix for writing?

Theoretically yes! This didn't happened in my experimental test, so I didn't bother, maybe it needs at least one unused export in the header to trigger this?

---

I built a command line interface to the compiler and for a roomtool on top of release-3.5.0 here: https://github.com/ericoporto/ags/tree/release-3.5.0-min-ags

Building is really simple, [provided we have cmake installed in Visual Studio, we just load the folder and it should build everything in `out/` directory once we hit build all](https://streamable.com/pk498x). With this we can grab this [empty_game.zip](https://github.com/ericoporto/ags/files/4935333/empty_game.zip), use `agscc.exe` to build an ags script file (that has no macros or comments) passing the provided header file, and then use the `roomtool.exe` to update the script in the `room1.crm`, and finally we load with either `ags.exe` or `acwin.exe` (it's 3.5.0.25) the engine, the game is empty and by placing `room1.crm` in the same directory. This is how I faced the above bug.

ivan-mogilko:
> Theoretically yes! This didn't happened in my experimental test (see below), so I didn't bother, maybe it needs at least one unused export in the header to trigger this?

Well, since it's theoretically possible, it's better to add a fix there too.

ivan-mogilko:
> Sure! Here is a minimum way to reproduce this:

Hm, do I understand correctly that this is only abstract example where you initialize a script object by hand? I was wondering if an actual proper script can lead to this situation.

ericoporto:
I changed approach and created a new `fwritestring` function and used it on imports, exports and sectionNames, I think this makes all that's read with `freadstring` safer and looks better on code.

PR #1107, opened 2020-07-19
[WIP] minor fixes for future Engine in 64-bit windows builds

ericoporto:
Gave a little try at 64 bit Windows builds and picked these. The replacing functions are compatible with both 32-bit and 64-bit builds. 

https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlonga
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongptra
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclasslonga
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclasslongptra

Note for anyone attempting 64-bit Windows builds: after these, and changing the Directx9 paths to use 64bit, it will build. These other changes are not interesting at this point in time, so I didn't want to add in this PR.

### DIFF
--- a/Engine/platform/windows/media/video/acwavi3d.cpp
+++ b/Engine/platform/windows/media/video/acwavi3d.cpp
@@ -184,7 +184,7 @@ CVMR9Graph::CVMR9Graph(HWND MediaWindow, IDirect3DDevice9 *device, int NumberOfS
 	}
 
   //m_pD3DDevice = device;
-  m_oldWndProc = GetWindowLong(m_hMediaWindow, GWL_WNDPROC);
+  m_oldWndProc = GetWindowLongPtr(m_hMediaWindow, GWLP_WNDPROC);
 }
 
 // Function name	: CVMR9Graph::~CVMR9Graph
@@ -193,7 +193,7 @@ CVMR9Graph::CVMR9Graph(HWND MediaWindow, IDirect3DDevice9 *device, int NumberOfS
 CVMR9Graph::~CVMR9Graph()
 {
 	ReleaseAllInterfaces();
-  long newProc = GetWindowLong(m_hMediaWindow, GWL_WNDPROC);
+  long newProc = GetWindowLongPtr(m_hMediaWindow, GWLP_WNDPROC);
 }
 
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
> Note for anyone attempting 64-bit Windows builds: after these, and changing the Directx9 paths to use 64bit, it will build. These other changes are not interesting at this point in time, so I didn't want to add in this PR.

Do you mean that Windows version still won't build with these fixes until more fixes are added?

ivan-mogilko:
Checking above quickly, `GetWindowLongPtr` returns `LONG_PTR`, so the variable type should be changed too for consistency.

ericoporto:
> Do you mean that Windows 64-bit version still won't build with these fixes until more fixes are added?
Why these other changes are "not interesting"?

I didn't want to disturb the master much for the 3.5.1 release, but basically after changing the above, you need to fix the libraries so they are 64-bit, for DirectX If you don't change it just fails on the linker, but once this is done it does build. Note: I am ignoring the Editor here.

> Checking above quickly, GetWindowLongPtr returns LONG_PTR, so the variable type should be changed too for consistency.

Uhm, I didn't picked this, will change it!

> Also the line in CVMR9Graph::~CVMR9Graph() does not make any sense :D, it probably got there after a bad merge.

Right, I didn't noticed. I will check it.

I will put a WIP in the title for now but can also close and reopen later. 

**Ignore this PR until the WIP is removed**

ivan-mogilko:
My point is, that moving to 64-bit Windows build should be done systematically, and small fixes like this don't really make much sense.

First of all, we need to find out if it's at all possible to have libraries work in 64-bit program (maybe it's possible to keep some of the same libs though?)

Then, probably, a new build configuration(s) has to be made in solution to distinct x32 and x64 for the first time, so that we have a fallback option.

Then we need to scan code locations which potentially address WinAPI functions. From the top of my head these are:

* Windows PlatformDriver implementation;
* Direct3D renderer and anything else using DirectX, like the CVMR9Graph in your PR.
* Utility functions, like Common/path.cpp and similar.

ivan-mogilko:
@ericoporto I'd like to close this PR, because it does not make much sense as it is.

BTW, after I checked the changes in commit more closely, these lines in acwavi3d.cpp should be just removed, because they initialize variables that are never used anywhere...

ericoporto:
It's alright :). I dropped of the face of GitHub for now but I am actually using AGS literally everyday, decided to work on my game. You did lots of nice things on the latest versions and it's been fun to use it. Thanks.

PR #1108, opened 2020-07-19
Script API: added System.Log()

### LINKED ISSUE #640: Add script function for writing into a console log

ivan-mogilko:
Resolves #640 (probably).

**IMPORTANT:** depends on #1103, so should not be merged before that.

This adds `System.Log(LogLevel, const string fmt, ...)` script function, as well as LogLevel enumeration that matches engine's log levels:

<pre>
enum LogLevel
{
	eLogAlert = 1,
	eLogFatal = 2,
	eLogError = 3,
	eLogWarn = 4,
	eLogInfo = 5,
	eLogDebug = 6
};
</pre>

Internally, renamed previously existing "Script" log group to "Game", and new "Script" group is exclusive for posting System.Log messages.


ericoporto:
The only thing I am not sure here is `game` is a good description of the AGS Script API, Engine 
API or AC API.

ivan-mogilko:
> The only thing I am not sure here is game is a good description of the AGS Script API, Engine
API or AC API. In my mind the reverse names are better - game for logs produced by the game and script for the log of script API.

My meaning was - "game" = "game logic" component. Assuming the program may be seen as a set of components - "main" for engine's "mainframe", "game" for game logic, and so on. If logging from the game script is "game", then we need another name for internal logging, as current "script" is way too narrow.

ericoporto:
> My meaning was - "game" = "game logic" component. 

I thought about this, I think this actually aligns well with `Game` struct and `game` state variables. I agree with selected names now.

ivan-mogilko:
Well, it is not strictly bound to script API, it's just the coincidence that these are messages that notify of failures in API calls.

We could have separate groups for script API reaction and general game logic, but it's just seem bit overcomplicated this way, and there's not much logging there at the moment.

rofl0r:
will this be added to ags3 branch too ?

ericoporto:
I think master is ags3, and ags3 is old ags3.

rofl0r:
> master is ags3

oh? cool. i was thinking master is ags4... sorry for the noise

ivan-mogilko:
Yes it is still 3.5.* branch, for now.

rofl0r:
could someone upload a tiny compiled testgame that calls the new function with some arguments ? i need to figure out how the bytecode looks like so it can be used with agsutils' assembler to insert debug code into existing games

ericoporto:
@rofl0r if you update the header in your compiler you should be able to build a script that calls it I think.

ericoporto:
[Logtest.zip](https://github.com/adventuregamestudio/ags/files/4989629/Logtest.zip)

@rofl0r try this, the compiled versions are inside too. It should produce:
```
Script : something
```

when running with ./ags --log-stdout=script=all

rofl0r:
@ericoporto: thanks!

Editor/AGS.Editor/Resources/agsdefns.sh
```diff
@@ -406,6 +406,19 @@ enum SortStyle
...
+enum LogLevel
+{
+	eLogAlert = 1,
+	eLogFatal = 2,
+	eLogError = 3,
+	eLogWarn = 4,
+	eLogInfo = 5,
+	eLogDebug = 6
+};
+#endif
```
rofl0r, line 419: i just tried to use this on master but it doesn't seem to be there ??

ericoporto: It's on ags4 branch

rofl0r: well, it's using api level 3.5.1, and 3.6.0 supercedes 3.5.1. it being in ags4 only is clearly a mistake.

ivan-mogilko: It's using api 3.5.1 in this pr. Currently in ags4 it refers to another api level: https://github.com/adventuregamestudio/ags/blob/ags4/Editor/AGS.Editor/Resources/agsdefns.sh#L400

Originally there was a branch for 3.5.1 that was discontinued.

ivan-mogilko: This is all happening because the plans for ags3/ags4 development are changing all the time.

Currently I'm actually considering to copy some of these back into 3.5.* or 3.6 because ags4 has some long-term goals, and particular API requires trivial changes to the engine, and may be useful in ags3 too.

ericoporto: if we are going to add something back then we better have this well defined to tell the Scumm VM people about it, since so far ags3 is the branch that is safe to run in scummvm.

rofl0r: 50% of the comments above were about whether this makes it into ags3 or ags4 and the answer was always "ags3"

if this is ags4 only, it is useless for me

ivan-mogilko: > if we are going to add something back then we better have this well defined to tell the Scumm VM people about it, since so far ags3 is the branch that is safe to run in scummvm.

It's not like I want to add this to already released version. If the scummvm team's goal is to sync with our 3.* branch, they are likely keep a close eye on our releases and changes.

ivan-mogilko: > if this is ags4 only, it is useless for me

Very well, I will see to port it to current master (3.6.0). Probably will take me few days to do everything carefully.
There was a number of functions, including this log, that were relatively simple and do not require big changes to the engine.

rofl0r: don't worry about scummvm, we discussed in #640 that we need a script function that can log to stdout for *debugging* purposes so its unlikely this is going to be used by game authors.

> Edit: btw System.Log is not a printf if that's what you were thinking...

what you mean by that? it's the debug printf function we've all been waiting for 1.5 years since #640 was opened, until it was finally merged into ags3 and then magically disappeared from ags3 just to re-appear in ags4 branch, and hurray, today when i finally got around to hacking on ags script and required this function for debugging, i noticed that it's no longer there.

ivan-mogilko: The System.Log prints to whatever engine is printing to. If engine prints to stdout, then the System.Log will print to stdout.

ericoporto: I needed it for my tap module, but if I am not mistaken it doesn't print the text purely, it adds a prefix, which breaks for me, I needed a clean print so that the CI could pickup the test results.

ivan-mogilko: Well, I did not know that. I think there has been a separate idea of something like System.PrintStdOut().

ivan-mogilko: @rofl0r I ported System.Log back to master: 0c226592c679106e15f4ec3847ee0d57c2afae19

It is supposed to work as described in this PR (logging level is controlled by engine's log config).

sorry for the inconvenience.

rofl0r: great, thanks! this is probably one of the most useful additions to ags in last 2 years.

PR #1124, opened 2020-09-28
Engine: BlendMode Features

AlanDrake:
Provides the BlendMode property for Character/Object/GUI/ScreenOverlay.
Known issues: Linear Burn and Dodge are approximations on D3D9/OGL, to work properly they'd require a writing a shader since I can't find any blend operation that would result in the correct formulas.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -104,6 +104,10 @@ Option to allow legacy relative asset resolutions.
 51 :
 Fonts have adjustable outline
 
+60 :
+Room.Exists
+BlendModes
+
 */
 
 enum GameDataVersion
@@ -138,7 +142,8 @@ enum GameDataVersion
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
     kGameVersion_351            = 51,
-    kGameVersion_Current        = kGameVersion_351
+    kGameVersion_399            = 60,
+    kGameVersion_Current        = kGameVersion_399
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -142,8 +142,7 @@ enum GameDataVersion
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
     kGameVersion_351            = 51,
-    kGameVersion_399            = 60,
-    kGameVersion_Current        = kGameVersion_399
+    kGameVersion_Current        = kGameVersion_351
 };
 
 extern GameDataVersion loaded_game_file_version;
--- a/Common/gui/guidefines.h
+++ b/Common/gui/guidefines.h
@@ -72,7 +72,8 @@ enum GuiVersion
     kGuiVersion_331         = 117,
     kGuiVersion_340         = 118,
     kGuiVersion_350         = 119,
-    kGuiVersion_Current     = kGuiVersion_350,
+    kGuiVersion_399         = 120,
+    kGuiVersion_Current     = kGuiVersion_399,
 };
 
 namespace AGS
@@ -174,7 +175,8 @@ enum GUITextBoxFlags
 enum GuiSvgVersion
 {
     kGuiSvgVersion_Initial  = 0,
-    kGuiSvgVersion_350      = 1
+    kGuiSvgVersion_350      = 1,
+    kGuiSvgVersion_399      = 2
 };
 
 } // namespace Common
--- a/Common/gui/guidefines.h
+++ b/Common/gui/guidefines.h
@@ -72,8 +72,7 @@ enum GuiVersion
     kGuiVersion_331         = 117,
     kGuiVersion_340         = 118,
     kGuiVersion_350         = 119,
-    kGuiVersion_399         = 120,
-    kGuiVersion_Current     = kGuiVersion_399,
+    kGuiVersion_Current     = kGuiVersion_350,
 };
 
 namespace AGS
@@ -176,7 +175,7 @@ enum GuiSvgVersion
 {
     kGuiSvgVersion_Initial  = 0,
     kGuiSvgVersion_350      = 1,
-    kGuiSvgVersion_399      = 2
+    kGuiSvgVersion_399      = 100
 };
 
 } // namespace Common
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -1558,6 +1558,20 @@ void Character_SetTransparency(CharacterInfo *chaa, int trans) {
     chaa->transparency = GfxDef::Trans100ToLegacyTrans255(trans);
 }
 
+int Character_GetBlendMode(CharacterInfo *chaa) {
+
+    return charextra[chaa->index_id].blend_mode;
+}
+
+void Character_SetBlendMode(CharacterInfo *chaa, int blendMode) {
+    /*
+    // TODO validation when blendmodes finalized
+    if ((blendMode < 0) || (blendMode > 4))
+    quit("!SetBlendMode: invalid blend mode");
+    */
+    charextra[chaa->index_id].blend_mode = blendMode;
+}
+
 int Character_GetTurnBeforeWalking(CharacterInfo *chaa) {
 
     if (chaa->flags & CHF_NOTURNING)
@@ -3760,6 +3774,18 @@ RuntimeScriptValue Sc_Character_SetZ(void *self, const RuntimeScriptValue *param
     API_OBJCALL_VOID_PINT(CharacterInfo, Character_SetZ);
 }
 
+// int (CharacterInfo *chaa)
+RuntimeScriptValue Sc_Character_GetBlendMode(void *self, const RuntimeScriptValue *params, int32_t param_count)
+{
+    API_OBJCALL_INT(CharacterInfo, Character_GetBlendMode);
+}
+
+// void (CharacterInfo *chaa, int blend_mode)
+RuntimeScriptValue Sc_Character_SetBlendMode(void *self, const RuntimeScriptValue *params, int32_t param_count)
+{
+    API_OBJCALL_VOID_PINT(CharacterInfo, Character_SetBlendMode);
+}
+
 //=============================================================================
 //
 // Exclusive API for Plugins
@@ -3942,6 +3968,9 @@ void RegisterCharacterAPI(ScriptAPIVersion base_api, ScriptAPIVersion compat_api
     ccAddExternalObjectFunction("Character::get_TintRed",               Sc_Character_GetTintRed);
     ccAddExternalObjectFunction("Character::get_TintSaturation",        Sc_Character_GetTintSaturation);
     ccAddExternalObjectFunction("Character::get_TintLuminance",         Sc_Character_GetTintLuminance);
+    
+    ccAddExternalObjectFunction("Character::get_BlendMode",             Sc_Character_GetBlendMode);
+    ccAddExternalObjectFunction("Character::set_BlendMode",             Sc_Character_SetBlendMode);
 
     /* ----------------------- Registering unsafe exports for plugins -----------------------*/
 
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -1564,11 +1564,9 @@ int Character_GetBlendMode(CharacterInfo *chaa) {
 }
 
 void Character_SetBlendMode(CharacterInfo *chaa, int blendMode) {
-    /*
-    // TODO validation when blendmodes finalized
-    if ((blendMode < 0) || (blendMode > 4))
-    quit("!SetBlendMode: invalid blend mode");
-    */
+    if (blendMode > 0) blendMode += 1; // Script API only uses from "Add" and later, 0 is leave-as-is
+    if ((blendMode < 0) /*|| (blendMode > kNumBlendModes)*/)
+        quit("!Character.BlendMode: invalid blend mode");
     charextra[chaa->index_id].blend_mode = blendMode;
 }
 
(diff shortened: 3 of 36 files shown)


AlanDrake:
Current issues:
- (Android) Darken and Lighten don't work because GLES2 lacks GL_MIN/GL_MAX, but maybe GLES3 supports them
- (D3D, OGL) Dodge is an approximation, Burn doesn't work properly

The unsupported blend modes could be supported through pixel shaders.
I'll keep this link [here](http://blog.onebyonedesign.com/unity/photoshop-blend-modes-for-unity3d/>) for potential reference.

Common/ac/game_version.h
```diff
@@ -138,7 +142,8 @@ enum GameDataVersion
     kGameVersion_341_2          = 49,
     kGameVersion_350            = 50,
     kGameVersion_351            = 51,
-    kGameVersion_Current        = kGameVersion_351
+    kGameVersion_399            = 60,
+    kGameVersion_Current        = kGameVersion_399
```
ivan-mogilko, line 146: The changes to game_version.h are unnecessary, because game data format itself was not modified.
That would be necessary if you had add these properties to the game data as well, so that they could be set in the editor and applied at startup.

Common/gui/guidefines.h
```diff
@@ -72,7 +72,8 @@ enum GuiVersion
     kGuiVersion_331         = 117,
     kGuiVersion_340         = 118,
     kGuiVersion_350         = 119,
-    kGuiVersion_Current     = kGuiVersion_350,
+    kGuiVersion_399         = 120,
+    kGuiVersion_Current     = kGuiVersion_399,
```
ivan-mogilko, line 76: Same goes for GuiVersion, as GUI game data format is not modified, only runtime save format.

Engine/ac/character.cpp
```diff
@@ -1558,6 +1558,20 @@ void Character_SetTransparency(CharacterInfo *chaa, int trans) {
...
+
+    return charextra[chaa->index_id].blend_mode;
+}
+
+void Character_SetBlendMode(CharacterInfo *chaa, int blendMode) {
+    /*
+    // TODO validation when blendmodes finalized
+    if ((blendMode < 0) || (blendMode > 4))
+    quit("!SetBlendMode: invalid blend mode");
+    */
```
ivan-mogilko, line 1571: I guess this commented code can be restored now?
Also, what's the meaning of "finalize" here, cannot blend mode be extended in the future by adding more constants to enum, or you think that will break something?

Engine/ac/draw.cpp
```diff
@@ -914,14 +914,15 @@ void sort_out_char_sprite_walk_behind(int actspsIndex, int xx, int yy, int basel
 void clear_draw_list() {
     thingsToDrawList.clear();
 }
-void add_thing_to_draw(IDriverDependantBitmap* bmp, int x, int y, int trans, bool alphaChannel) {
+void add_thing_to_draw(IDriverDependantBitmap* bmp, int x, int y, int trans, bool alphaChannel, int blendMode = 0) {
```
ivan-mogilko, line 917: Since BlendMode is a enum, could you use that enum here instead of type "int"?

Engine/ac/characterextras.h
```diff
@@ -43,8 +44,9 @@ struct CharacterExtras {
     char  process_idle_this_time;
     char  slow_move_counter;
     short animwait;
+    int blend_mode;
```
ivan-mogilko, line 47: Since BlendMode is a enum, could you use that enum here instead of type "int"? And everywhere where it's passed as a function parameter.

Engine/ac/draw.cpp
ivan-mogilko, line 917: Please use actual blend constant instead of "0", since blend mode is enum and supposed to stay in defined range.

Engine/game/savegame.h
```diff
@@ -52,7 +52,8 @@ enum SavegameVersion
     kSvgVersion_Cmp_64bit = 10,
     kSvgVersion_350_final = 11,
     kSvgVersion_350_final2= 12,
-    kSvgVersion_Current   = kSvgVersion_350_final2,
+    kSvgVersion_399       = 13,
```
ivan-mogilko, line 55: Please give new save format a much higher number in case there will be change in 3.* branch

Common/gui/guidefines.h
```diff
@@ -174,7 +175,8 @@ enum GUITextBoxFlags
 enum GuiSvgVersion
 {
     kGuiSvgVersion_Initial  = 0,
-    kGuiSvgVersion_350      = 1
+    kGuiSvgVersion_350      = 1,
+    kGuiSvgVersion_399      = 2
```
ivan-mogilko, line 179: Please give new save format a much higher number in case there will be change in 3.* branch

Engine/ac/characterextras.cpp
```diff
@@ -34,6 +34,22 @@ void CharacterExtras::ReadFromFile(Stream *in)
     process_idle_this_time = in->ReadInt8();
     slow_move_counter = in->ReadInt8();
     animwait = in->ReadInt16();
+    if (cmp_ver >= 2)
```
ivan-mogilko, line 37: Please give new save format a much higher number in case there will be change in 3.* branch

Engine/ac/roomobject.cpp
```diff
@@ -149,12 +150,28 @@ void RoomObject::update_cycle_view_backwards()
...
 }
 
-void RoomObject::ReadFromFile(Stream *in)
+void RoomObject::ReadFromFile(Stream *in, int32_t cmp_ver)
 {
     in->ReadArrayOfInt32(&x, 3);
     in->ReadArrayOfInt16(&tint_r, 15);
     in->ReadArrayOfInt8((int8_t*)&cycling, 4);
     in->ReadArrayOfInt16(&blocking_width, 2);
+    if (cmp_ver >= 1)
```
ivan-mogilko, line 159: Please give new save format a much higher number in case there will be change in 3.* branch

Engine/ac/screenoverlay.cpp
```diff
@@ -38,6 +38,22 @@ void ScreenOverlay::ReadFromFile(Stream *in, int32_t cmp_ver)
         _offsetX = in->ReadInt32();
         _offsetY = in->ReadInt32();
     }
+    if (cmp_ver >= 2)
```
ivan-mogilko, line 41: Please give new save format a much higher number in case there will be change in 3.* branch

Engine/game/savegame_components.cpp
```diff
@@ -1092,7 +1092,7 @@ ComponentHandler ComponentHandlers[] =
     },
     {
         "Characters",
-        1,
+        2,
```
ivan-mogilko, line 1095: Please give new save formats much higher numbers in case there will be change in 3.* branch

Engine/gfx/blender.h
```diff
@@ -62,4 +62,109 @@ void set_opaque_alpha_blender();
...
+
+#define BLEND(bpp, r, g, b)   _blender_trans##bpp(makecol##bpp(r, g, b), y, n)
+// some formulas from AGSBlend
+#define _BLENDOP_BURN(B,L) ((B + L < 255) ? 0:(B + L - 255)) // note: burn was called subtract
+#define _BLENDOP_LIGHTEN(B,L) ((L > B) ? L:B)
+#define _BLENDOP_DARKEN(B,L) ((L > B) ? B:L)
+#define _BLENDOP_EXCLUSION(B,L) (B + L - 2 * B * L / 255)
+#define _BLENDOP_SUBTRACT(B,L) (L - B < 0) ? 0:(L - B)
+// not very pretty but forces the original alpha of "x" to "blender_result"
+inline unsigned long _blender_mask_alpha24(unsigned long blender_result, unsigned long x, unsigned long y, unsigned long n)
```
ivan-mogilko, line 79: This is not critical, but I'd like to note that defining these functions inlined in the header makes little sense, because they are passed as pointers to sprite drawing routine, and code inlining will not happen anyway.

Engine/gfx/blender.h
```diff
@@ -62,4 +62,109 @@ void set_opaque_alpha_blender();
...
+}
+inline unsigned long _my_blender_subtract24(unsigned long x, unsigned long y, unsigned long n)
+{
+    const unsigned long h = _BLENDOP_SUBTRACT(getr24(x), getr24(y));
+    const unsigned long j = _BLENDOP_SUBTRACT(getg24(x), getg24(y));
+    const unsigned long k = _BLENDOP_SUBTRACT(getb24(x), getb24(y));
+    return BLEND(24, h, j, k);
+}
+
+inline unsigned long _blender_masked_add32(unsigned long x, unsigned long y, unsigned long n)
```
ivan-mogilko, line 132: This is not critical, but I'd like to note that defining these functions (_blender_masked_add32 and so on) inlined in header makes little sense, as they are passed into sprite drawing routine as pointers, and code inlining will not happen anyway.

ivan-mogilko:
I added comments to the code, some things are serious others are not so much. There are things in actual blending implementations which I cannot comment on since never tried these myself.

Something I want to ask, do you plan on keeping DrawingSurface (raw drawing) in ags4, and if so, do you like to have a DrawImage variant with software blending (e.g. like BlendImage, to not increase parameter list of simple DrawImage further)?

ivan-mogilko:
Regarding the API, what supposed to be a difference between eBlendModeDefault and eBlendModeAlpha?
Currently eBlendModeDefault by value corresponds with BlendMode_NoAlpha in the engine, which (in engine) means the sprite is drawn without alpha blending: as opaque, and treating magic pink as transparency.

In practice setting an object with alpha sprite to eBlendModeDefault does not make any noticeable change compared to eBlendModeAlpha.

Common/ac/game_version.h
AlanDrake: It was in my plans to add the property to property grid, what do you think?
If we do add it, do we also try to implement the blending modes in the editor preview?

Engine/game/savegame_components.cpp
AlanDrake: You're right, I somehow missed the Overlays component upgrade...
Do we really need to keep aligned to ags3 savegames? How much higher, I was kinda counting on nobody ever daring increase formats in the 3 branch ever again.

Common/ac/game_version.h
ivan-mogilko: Well, my personal opinion is that properties should be same everywhere, as people may want to set up constant mode for object too. Affecting preview is an optional thing here imho, because editor is not perfectly synced with the engine anyway.

AlanDrake:
Thanks for the pointers, I'll see to fix the glaring issues as soon as possible.

I don't have plan for DrawingSurfaces, but they're probably necessary in certain scenarios so I wouldn't remove them outright.

Right, the blendmodes enums, my conundrum was that 0 is the sane default value, but the original enum was not under this logic, also I may have removed all cases where kBlendMode_NoAlpha was used (the wrong style legacy guy blending).
To be honest I forgot if it was safe to remove, so I ended up with that mismatch.
I didn't implement NoAlpha in the blending modes because I doubt anyone would have use of it.

Engine/game/savegame_components.cpp
ivan-mogilko: This is int32, so it could be anywhere, like 100. It's an arbitrary thing.

Even if you dont support loading old saves, and dont want to know their format numbers, I still believe it's a good thing to have distinction, because it's easy for us, and will keep any emergency cases covered.
3.5.1 for example is still not released, and it's not known when ags4 branch will be having proper releases (and how they will look like).

ivan-mogilko:
I have to reinvestigate how is kBlendMode_NoAlpha used now, because it was a while...

ivan-mogilko:
Is there a reason why you dont want to check the upper limit of the blend modes in these script functions? Imo that's the only significant thing remaining.

Other couple of nitpicks,
* I dont think there should be any comments about what engine does in the script header. Script API is a specification, and implementation is engine's own business, there's no need for script users to know that. On contrary, engine could have a comment about this to justify conversion between user argument and final value.
* IMO it's better to validate bitmap->_blendMode when it's set (once), rather than when it's used (every time game is redrawn).

ivan-mogilko:
Is there a reason why you dont want to check the upper limit of the blend modes in these script functions? Imo that's the only significant thing remaining.

Other couple of nitpicks,
* I dont think there should be any comments about what engine does in the script header. Script API is a specification, and implementation is engine's own business, there's no need for script users to know that. On contrary, engine could have a comment about this to justify conversion between user argument and final value.
* IMO it's better to validate and fixup bitmap->_blendMode when it's set (once), rather than when it's used (every time game is redrawn). I also wonder why _bitmap does not have blendmode as Common::BlendMode right away... That object is already a part of software rendering implementation, so it's perfectly fine to have it use these blend codes directly.

AlanDrake:
Yes, there are several reasons I'm not checking yet. I'm trying to implement new ones, and then the actual reason: I don't know if to put the MAX amount as a Game constant or into the Script API enum, and if the latter how to name it (max, count,last?), none of the enums in agsdefns.sh do this so far so it would be rude of me to decide on a standard, making these trivial decision is literally impossible for me.

The comment in the Script API was just for us developers, users can't see it (can they?), but if you think it shouldn't be there I'll remove it.

About the bitmap->_blendMode thing, I'm looking into it.

ivan-mogilko:
> Yes, there are several reasons I'm not checking yet. I'm trying to implement new ones, and then the actual reason: I don't know if to put the MAX amount as a Game constant or into the Script API enum, and if the latter how to name it (max, count,last?),

It does not matter whether the enum will be expanded in the future or not. Any enum can be expanded after all.
Every given version of engine knows which modes it supports, and therefore should be checking for that range. That's also the point of "NumModes" constant - if you use it in the code you only have to update its value in declaration later if more modes are added.
Putting MAX into script enum is not necessary for this either, but could be done if wanted. Again, it does not matter whether enum will be expanded, as this constant will stay valid and truthful for every given version of the script API.

ivan-mogilko:
Opened an updated pr which contains almost all of these commits, and some additions: #1191.

PR #1125, opened 2020-10-03
Sprite Manager: adds context menu View option.

### LINKED ISSUE #922: Suggestion on import sprite of sprite manager tab
- Add a small row of buttons inside the sprite manager tab, above the sprite list area, with the first button being "import new sprite from files...". Other functionalities may or may not be added as buttons, but this is an important one I miss when I am on my laptop, because I rarely bring my mouse and my laptop right click is really awful.
- Alternatively, support file drop for the supported images on the sprite manager tab.

ericoporto:
sprite_preview_wheel

List of Features:

- Allows changing preview sizes for 1x, 2x, 3x, and 4x on right click;
- Slider gives 0.5 steps from 0.5x to 4x;
- Ctrl+Mouse Wheel also modifies the slider;
- Default size is 1x;
- Ctrl+0 resets to default size;
- Added a button for importing new sprite files;
- Image File Drop on Sprite Manager is supported, just click hold the desired image files on Windows Explorer and drag them in the Sprite Manager for import. Fix #922 .

Now I noticed View means something different in AGS and I just stole the naming from Windows Explorer, if a different name is better, please chime in.

### DIFF
--- a/Editor/AGS.Editor/Panes/SpriteSelector.Designer.cs
+++ b/Editor/AGS.Editor/Panes/SpriteSelector.Designer.cs
@@ -31,9 +31,10 @@ namespace AGS.Editor
             this.splitWindow = new System.Windows.Forms.SplitContainer();
             this.folderList = new System.Windows.Forms.TreeView();
             this.splitContainer1 = new System.Windows.Forms.SplitContainer();
-            this.spriteList = new System.Windows.Forms.ListView();
             this.panel1 = new System.Windows.Forms.Panel();
+            this.button_importNew = new System.Windows.Forms.Button();
             this.sliderPreviewSize = new System.Windows.Forms.TrackBar();
+            this.spriteList = new System.Windows.Forms.ListView();
             ((System.ComponentModel.ISupportInitialize)(this.splitWindow)).BeginInit();
             this.splitWindow.Panel1.SuspendLayout();
             this.splitWindow.Panel2.SuspendLayout();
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
It's interesting that the branch's name mentions "slider" :). Have you considered making an actual slider?

TBH I am not certain what would be best way to handle these scales. We already have got a zoom slider on room and gui panel, perhaps it would be desirable to be consistent?

ericoporto:
Oh, yeah, it was supposed to be a slider but I was fighting a bit with positioning (it just looked wrong when resizing it) it in the sprite selector pane so I decided to see how windows explorer did it and found the right click implementation. 

I actually like the right click implementation too, but I could not figure out how to indicate the selection with a bullet like in Windows Explorer, I could only do checkmark, but it felt wrong having a check mark with radio button behavior so I scrapped.

I also wondered if I the zoom was something that needed saving in Game.agf.user but I thought it was not needed. I may be wrong on that.

ericoporto:
I did a commit that adds back the slider I had previously, I was just unsure if things look good... I feel like I need to add icons for at least some sprite import actions for it to look good.

image

ivan-mogilko:
> I also wondered if the zoom was something that needed saving in Game.agf.user but I thought it was not needed. I may be wrong on that.

That's a good question, idk what's more convenient, to have on in preferences, or on per-project basis.

ericoporto:
slider_test

I hear ya on the idea of the slider, it can be modified.

I have a practical problem with resizing, I think when something needs to get bigger I want to have the resizing be Nearest Neighbor and want resizing down, some aliasing gives best results. 

I used a simple math to judge the right size of the sprite in the preview, which I am not sure yet is the best but appears to give a reasonable preview.

ericoporto:
I found a way to make the resize look good! Last results are better, I will add a gif to demonstrate here in this comment after I eat.

rofl0r:
> The slider

in many gui programs scale/zoom can be modified using CTRL+scroll wheel

ivan-mogilko:
Maybe keep the context menu for options like 100, 150, 200% and so on, and also have a ctrl + mouse wheel for more gradual zoom?

ericoporto:
mouse_wheel_support

- Ctrl+Mouse Wheel support is in! If you don't use ctrl, it just scrolls.
- slider goes from 0.5x to 4x of the current sizes.
- default value is 1x!
- Erh, kinda unrelated, but I took the opportunity to support dropping an image file and having it be imported.

rofl0r:
does this issue closing spree mean there was some ugly discussion in AGS forums?

PR #1127, opened 2020-10-08
Script API: add Game.BlockingTextOverlay and Game.BlockingWaitSkipped

ivan-mogilko:
Done by user's request, these two properties give game dev little more control over built-in blocking speech in AGS.
These do not require any data format change, and only expose data which already exists internally in the engine.

### Game.BlockingTextOverlay
`readonly attribute Overlay *Game.BlockingTextOverlay` returns special overlay created for the blocking speech, or null if no such text is on screen.

Because of how blocking speech works, this overlay will be only accessible in `repeatedly_execute_always` and `late_repeatedly_execute_always` script callbacks.<br>
But this is a valid overlay in every aspect: you can even change it's coordinates and text.<br>
Calling Remove() function is also legal, and will correctly stop the current blocking speech (the engine was already programmed to rely strictly on overlay's presence).

**IMPORTANT:** in theory, this will return both overlay for blocking Say command and Display command. But the caveat here is that no script callbacks are running during Display in current engine, it's a kind of a "super-blocking" function, but if there will be such function then this property will work for Display as well.

When the speech is removed, script object is detached from actual overlay and invalidated: its Overlay.Valid property will be returning false, and attempt to call any other property or function will result in script error. This is essential to know if you store Overlay pointer in a variable for later use.

For every speech line there's a new overlay generated by the engine. This gives an opportunity to detect the exact game frame when the speech line changed, by comparing previously saved pointer with Game.BlockingTextOverlay's value.

### eSkipNone

`eSkipNone` option added for SkipSpeechStyle. When set, speech may only be skipped by a script command (i.e. - calling Game.BlockingTextOverlay.Remove()).<br>
This complements similar mode added for Wait and cutscenes lately.

### Game.BlockingWaitSkipped
`readonly attribute int Game.BlockingWaitSkipped` returns a code telling how the last blocking wait action was ended. This code is equivalent to the return value of the Wait* functions, introduced previously (#1084). At the moment this property is updated by three sorts of skippable blocking actions: Say*, Display* and Wait*.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2812,6 +2812,7 @@ builtin struct GameState {
   
 #ifdef SCRIPT_API_v330
 enum SkipSpeechStyle {
+  eSkipNone         = -1,
   eSkipKeyMouseTime = 0,
   eSkipKeyTime      = 1,
   eSkipTime         = 2,
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2687,6 +2687,10 @@ builtin struct Game {
   /// Gets the number of cameras.
   import static readonly attribute int CameraCount;
 #endif
+#ifdef SCRIPT_API_v351
+  /// Gets the overlay representing displayed blocking text, or null if no such text none is displayed at the moment.
+  import static readonly attribute Overlay* BlockingTextOverlay;
+#endif
 };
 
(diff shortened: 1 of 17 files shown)


ericoporto:
Is there a reason why `Game.BlockingWaitSkipped` does not return an enum? I haven't looked at the code yet.

ivan-mogilko:
> why is this on Game struct and not on Speech struct?

Because this overlay is created not only for speech.

rofl0r:
alright, i see reason for closing this was a message posted to engine_development in AGS discord. *sigh*

PR #1128, opened 2020-10-19
Engine: let D3D9 decide refresh rate on its own

AlanDrake:
GetMode() may return the wrong refresh rate in the case the monitor is not at 60hz, moreover the config refresh setting never makes it far enough to be used.
Leaving it as D3DPRESENT_RATE_DEFAULT will ensure switching from windowed to fullscreen is quick and reliable, as long as it's the same resolution.


ivan-mogilko:
> moreover the config refresh setting never makes it far enough to be used

Could you clarify this? Is there a bug in the engine where config setting is lost? Maybe there's more to fix then.

Is this setting is useful at all?

AlanDrake:
The config file override is lost somewhere along the way, maybe even before GetMode() but I'm not sure.

I don't know how useful the `refresh=` override is, at least on Windows, maybe it only works for software mode? I don't know about the other platforms.

Issue #469, opened 2018-07-06
Editor: change game project's Rooms to open format


ghost:
Current situation is this:
* in the game project rooms are stored as compiled binary blobs, which include room description, background images, area masks and compiled room scripts.
* some of the operations (need to reinvestigate to see which ones) in the Editor rely on C++ code in AGS.Native dll, where rooms are accessed in their runtime format.

This causes number of issues, including inability for multiple users to work on same room in parallel (because changes are impossible to merge).

### The suggestion

1) All the room management should be reimplemented in .NET.

2) Have rooms in the game project serialized as a group of files:
* XML for the room description (we have a ready generic serializer for that);
* image files for backgrounds and masks;
* room script is already available in text source, so probably nothing should be changed here (perhaps, that may have sense to cache compiled room scripts, but that's another problem).

### Possible steps

1) We already have a Room class in .NET application. We need to research how Editor works with rooms, and make sure that only that and accompanying classes are used in all room editing operations. In the end, Editor should not rely on AGS.Native.dll when working with rooms (except for compilation, for now).
Editor should be fully able to load up a Room class from the runtime format.

2) Editor has a class-to-XML serialization mechanism based on reflection, which is currently used to save main game settings. Same could be used to store rooms. Implement Room class serialization using that serializer.
**NOTE:** if it would be desired to use other format than XML, then we'd only need to modify serializer later. Then all game data will save in a new format at once. Of course in such case we'd additionally have to implement conversion from the old game projects.

3) Since room source will probably consist of several files, I'd assume we need to move them into their own folder (e.g. "Rooms") and maybe even have subfolders per each room there. So, the next task is to invent a good file/folder structure and complete room serialization in the new way.

4) Necessary changes to game compilation: currently it relies on updating binary room files. This again need to be researched, what would be better - to keep cached binary rooms (or parts of these) and update them as user saves one in the editor, or fully rebuild a room every time when game binaries are generated.

~~5) Bonus step: reimplement room compilation in .NET app.~~
**UPDATE:** it appears more desirable to have stand-alone compilation tools, so the room files -> CRM compiler could as well be a stand-alone tool run by the editor. In this case it may as well be written in C++ for max platform compatibility.

rofl0r:
XML is an abomination and finally the whole industry agrees. it shouldn't be used for any new code.

my 2 shekels.

ivan-mogilko:
Editor already uses XML. If XML is not to be used, then everything needs to be changed to a different format as well.

[...]

Issue #1113, opened 2020-07-30
Allow importing indexed png as masks


ericoporto:
Currently importing room masks is restricted to `.bmp` files. This has not been always like this. At some point in the past it was possible to import Room masks as indexed png files. Unfortunately I am having a hard time tracing when this was changed. I only noticed because recently I had to edit some old rooms in my game that had the walkbehind png masks and I noticed I could not import them anymore. 

I noticed the February 2019 build of 3.4.3 allows png import but it requires the png to have exactly 256 colors in it's palette. I swear there was an older version I could import a mask with less colors in the png as long as it was indexed, but I can't seem to find it anymore (I tried the latest 3.4.1 and the result was the same). But now, it needs to be not only 256 colors but also a bmp.

ivan-mogilko:
If this was allowed but no longer works, then this should be considered a bug.

ericoporto:
There's a very good chance it was by mistake because I noticed today in older versions it could import and throw error after `.jpeg` images. I still haven't found the version that allowed lower than 256 colors palettes, but I will keep looking.

AlanDrake:
It wasn't by mistake #548

ivan-mogilko:
So, the #555 only changed the filter in the file dialog, but it does not seem to change anything in regards to loading.

ivan-mogilko:
I now remembered I changed the _filter in file dialog_ #555 , because I thought that AGS cannot import PNGs as room backgrounds (judging by user having problem with that).

ivan-mogilko:
Anyhow, this may be a matter of correctly converting source PNG to 8-bit bitmap, if this is done the support could be restored properly.

ericoporto:
> because I thought that AGS cannot import PNGs as room backgrounds (judging by user having problem with that)

You mean masks right? It always could import png as room backgrounds.

> Anyhow, this may be a matter of correctly converting source PNG to 8-bit bitmap, if this is done the support could be restored properly.

A png can have a palette and be indexed by it and the palette can have up to 256 colors (8-bit). A png can also have a bigger color depth and no palette. When I meant png I meant the ones that have palette.

ericoporto:
I think currently an exported mask can't be reimported - didn't work for me at least. The exported mask has fewer colors on it's palette than 256 - it must be exactly 256 for importing, doesn't accept less colors.

ivan-mogilko:
[...]

PR #1138, opened 2020-12-05
Refactored room editor to use graphics from C# instead of AGS.Native

persn:
Related to #469 

In short I have
1. Centralized certain room operations used by the room editor to the `IRoomController`, instead of globally used native invocations from `NativeProxy`. This is affects the public API for editor plugin developers.
2. Wrote a small framework as `System.Drawing.Bitmap` extensions for drawing images. I shopped around for potential third party frameworks but drawing on 8-bit indexed images is generally not supported.
3. Replaced a lot of `NativeProxy` invocations related to `IRoomController` with C# code. There's still some room code that uses `NativeProxy`, primarily related to loading/saving data from .crm files. This will be replaced in a future pull request.
4. Deleted dead AGS.Native code related to rooms.

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -601,7 +601,8 @@ namespace AGS.Editor.Components
             Room room = (Room)parameter;
             _agsEditor.RegenerateScriptHeader(room);
             List<Script> headers = (List<Script>)_agsEditor.GetAllScriptHeaders();
-            _agsEditor.CompileScript(room.Script, headers, null);
+            _agsEditor.CompileScript(room.Script, headers, null, true);
+            ((IRoomController)this).Save();
 
             return null;            
         }
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1669,6 +1669,7 @@ namespace AGS.Editor.Components
             if (maskType != RoomAreaMaskType.None)
             {
                 Bitmap mask8bpp = _maskCache[maskType];
+                ColorPalette paletteBackup = mask8bpp.Palette;
 
                 // Color not selected mask areas to gray
                 if (_grayOutMasks)
@@ -1700,6 +1701,7 @@ namespace AGS.Editor.Components
                 attributes.SetColorMatrix(colorMatrix, ColorMatrixFlag.Default, ColorAdjustType.Bitmap);
 
                 g.DrawImage(mask8bpp, drawingArea, 0, 0, mask8bpp.Width, mask8bpp.Height, GraphicsUnit.Pixel, attributes);
+                mask8bpp.Palette = paletteBackup;
             }
         }
 
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1524,6 +1524,21 @@ namespace AGS.Editor.Components
             _loadedRoom.Modified = true;
         }
 
+        void IRoomController.DeleteBackground(int background)
+        {
+            if (_loadedRoom == null)
+            {
+                throw new InvalidOperationException("No room is currently loaded");
+            }
+
+            // TODO Replace with bitmap deleting from disk with C# when room is open format
+            lock (_loadedRoom)
+            {
+                _nativeProxy.DeleteBackground(_loadedRoom, background);
+            }
+            _loadedRoom.Modified = true;
+        }
+
         Bitmap IRoomController.GetMask(RoomAreaMaskType mask)
         {
             if (_loadedRoom == null)
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -867,7 +867,7 @@ namespace AGS.Editor.Components
                 if (mask != room.MaskResolution)
                 {
                     room.MaskResolution = mask;
-                    NativeProxy.Instance.AdjustRoomMaskResolution(room);
+                    ((IRoomController)this).AdjustMaskResolution();
                     return true;
                 }
             }
@@ -1119,7 +1119,7 @@ namespace AGS.Editor.Components
                 if (Factory.GUIController.ShowQuestion("The new mask resolution is smaller and this will reduce mask's precision and some pixels may be lost in the process. Do you want to proceed?") != DialogResult.Yes)
                     return;
             }
-            _nativeProxy.AdjustRoomMaskResolution(_loadedRoom);
+            ((IRoomController)this).AdjustMaskResolution();
         }
 
         protected override void AddNewItemCommandsToFolderContextMenu(string controlID, IList<MenuCommand> menu)
@@ -1615,6 +1615,37 @@ namespace AGS.Editor.Components
             {
                 ((IRoomController)this).SetMask(RoomAreaMaskType.Regions, bmp);
             }
+            _loadedRoom.Modified = true;
+        }
+
+        void IRoomController.AdjustMaskResolution()
+        {
+            if (_loadedRoom == null)
+            {
+                throw new InvalidOperationException("No room is currently loaded");
+            }
+
+            using (Bitmap bg = ((IRoomController)this).GetBackground(0))
+            {
+                int baseWidth = bg.Width;
+                int baseHeight = bg.Height;
+                int scaledWidth = baseWidth / _loadedRoom.MaskResolution;
+                int scaledHeight = baseHeight / _loadedRoom.MaskResolution;
+
+                foreach (RoomAreaMaskType maskType in Enum.GetValues(typeof(RoomAreaMaskType)))
+                {
+                    if (maskType == RoomAreaMaskType.None) continue;
+
+                    using (Bitmap mask = ((IRoomController)this).GetMask(maskType))
+                    using (Bitmap scaled = maskType == RoomAreaMaskType.WalkBehinds
+                        ? mask.Scale(baseWidth, baseHeight) // Walk-behinds are always 1:1 of the primary background
+                        : mask.Scale(scaledWidth, scaledHeight)) // Other masks are 1:x where X is MaskResolution
+                    {
+                        ((IRoomController)this).SetMask(maskType, scaled);
+                    }
+                }
+            }
+
             _loadedRoom.Modified = true;
         }
 
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -609,7 +609,7 @@ namespace AGS.Editor.Components
             Room room = (Room)parameter;
             _agsEditor.RegenerateScriptHeader(room);
             List<Script> headers = (List<Script>)_agsEditor.GetAllScriptHeaders();
-            _agsEditor.CompileScript(room.Script, headers, null, true);
+            _agsEditor.CompileScript(room.Script, headers, null);
             ((IRoomController)this).Save();
 
             return null;            
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1604,6 +1604,20 @@ namespace AGS.Editor.Components
 			g.ReleaseHdc(hdc);
 		}
 
+        void IRoomController.CopyWalkableAreaMaskToRegions()
+        {
+            if (_loadedRoom == null)
+            {
+                throw new InvalidOperationException("No room is currently loaded");
+            }
+
+            using (Bitmap bmp = ((IRoomController)this).GetMask(RoomAreaMaskType.WalkableAreas))
+            {
+                ((IRoomController)this).SetMask(RoomAreaMaskType.Regions, bmp);
+            }
+            _loadedRoom.Modified = true;
+        }
+
 		bool IRoomController.GreyOutNonSelectedMasks
 		{
 			set { _nativeProxy.GreyOutNonSelectedMasks = value; }
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1720,12 +1720,12 @@ namespace AGS.Editor.Components
             {
                 if (maskType == RoomAreaMaskType.None) continue;
 
-                Bitmap mask = _maskCache[maskType];
-                Bitmap scaled = maskType == RoomAreaMaskType.WalkBehinds
-                    ? mask.Scale(baseWidth, baseHeight) // Walk-behinds are always 1:1 of the primary background
-                    : mask.Scale(scaledWidth, scaledHeight); // Other masks are 1:x where X is MaskResolution
-                mask.Dispose();
-                _maskCache[maskType] = scaled;
+                using (Bitmap mask = _maskCache[maskType])
+                {
+                    _maskCache[maskType] = maskType == RoomAreaMaskType.WalkBehinds
+                        ? mask.ScaleIndexed(baseWidth, baseHeight) // Walk-behinds are always 1:1 of the primary background
+                        : mask.ScaleIndexed(scaledWidth, scaledHeight); // Other masks are 1:x where X is MaskResolution
+                }
             }
 
             _loadedRoom.Modified = true;
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1540,6 +1540,10 @@ namespace AGS.Editor.Components
             {
                 throw new InvalidOperationException("No room is currently loaded");
             }
+            if (bmp == null)
+            {
+                throw new ArgumentNullException(nameof(bmp));
+            }
 
             _loadedRoom.Width = bmp.Width;
             _loadedRoom.Height = bmp.Height;
@@ -1596,6 +1600,10 @@ namespace AGS.Editor.Components
             if (_loadedRoom == null)
             {
                 throw new InvalidOperationException("No room is currently loaded");
+            }
+            if (bmp == null)
+            {
+                throw new ArgumentNullException(nameof(bmp));
             }
 
             // TODO Replace with bitmap saving to disk with C# when room is open format
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1517,6 +1517,7 @@ namespace AGS.Editor.Components
                 throw new InvalidOperationException("No room is currently loaded");
             }
 
+            SaveImages();
             _nativeProxy.SaveRoom(_loadedRoom);
             _loadedRoom.Modified = false;
         }
@@ -1528,10 +1529,7 @@ namespace AGS.Editor.Components
                 throw new InvalidOperationException("No room is currently loaded");
             }
 
-            // TODO Replace with bitmap loading from disk with C# when room is open format
-            return _backgroundCache.Any()
-                ? (Bitmap)_backgroundCache[background]?.Clone()
-                : _nativeProxy.GetBitmapForBackground(_loadedRoom, background);
+            return _backgroundCache[background].Clone() as Bitmap;
         }
 
         void IRoomController.SetBackground(int background, Bitmap bmp)
@@ -1545,20 +1543,38 @@ namespace AGS.Editor.Components
                 throw new ArgumentNullException(nameof(bmp));
             }
 
-            _loadedRoom.Width = bmp.Width;
-            _loadedRoom.Height = bmp.Height;
-            // TODO Replace with bitmap saving to disk with C# when room is open format
-            _nativeProxy.ImportBackground(
-                _loadedRoom, background, bmp, _agsEditor.Settings.RemapPalettizedBackgrounds, sharePalette: false);
+            Bitmap newBmp = (Bitmap)bmp.Clone();
+            _loadedRoom.Width = newBmp.Width;
+            _loadedRoom.Height = newBmp.Height;
+            _loadedRoom.ColorDepth = newBmp.GetColorDepth();
+
+            if (_loadedRoom.ColorDepth == 8)
+            {
+                ColorPalette palette = newBmp.Palette;
+                foreach (var color in _agsEditor.CurrentGame.Palette.Where(p => p.ColourType == PaletteColourType.Locked))
+                {
+                    palette.Entries[color.Index] = color.Colour;
+                }
+                newBmp.Palette = palette;
+
+                newBmp.SetGlobalPaletteFromPalette();
+            }
 
             if (background >= _backgroundCache.Count)
             {
-                _backgroundCache.Add((Bitmap)bmp.Clone());
+                _backgroundCache.Add(newBmp);
+                _loadedRoom.BackgroundCount++;
             }
             else
             {
                 _backgroundCache[background]?.Dispose();
-                _backgroundCache[background] = (Bitmap)bmp.Clone();
+                _backgroundCache[background] = newBmp;
+            }
+
+            // If size or resolution has changed, reset masks
+            if (newBmp.Width != _loadedRoom.Width || newBmp.Height != _loadedRoom.Height)
+            {
+                ClearMaskCache();
             }
 
             _loadedRoom.Modified = true;
@@ -1571,13 +1587,9 @@ namespace AGS.Editor.Components
                 throw new InvalidOperationException("No room is currently loaded");
             }
 
-            // TODO Replace with bitmap deleting from disk with C# when room is open format
-            lock (_loadedRoom)
-            {
-                _nativeProxy.DeleteBackground(_loadedRoom, background);
-            }
             _backgroundCache[background]?.Dispose();
             _backgroundCache.RemoveAt(background);
+            _loadedRoom.BackgroundCount--;
             _loadedRoom.Modified = true;
         }
 
@@ -1588,11 +1600,7 @@ namespace AGS.Editor.Components
                 throw new InvalidOperationException("No room is currently loaded");
             }
 
-            // TODO Replace with bitmap loading from disk with C# when room is open format
-            Bitmap cached;
-            return _maskCache.TryGetValue(mask, out cached) && cached != null
-                ? (Bitmap)cached.Clone()
-                : _nativeProxy.ExportAreaMask(_loadedRoom, mask);
+            return _maskCache[mask].Clone() as Bitmap;
         }
 
         void IRoomController.SetMask(RoomAreaMaskType mask, Bitmap bmp)
@@ -1606,10 +1614,8 @@ namespace AGS.Editor.Components
                 throw new ArgumentNullException(nameof(bmp));
             }
 
-            // TODO Replace with bitmap saving to disk with C# when room is open format
-            _nativeProxy.SetAreaMask(_loadedRoom, mask, bmp);
             _maskCache[mask]?.Dispose();
-            _maskCache[mask] = (Bitmap)bmp.Clone();
+            _maskCache[mask] = bmp.Clone() as Bitmap;
             _loadedRoom.Modified = true;
         }
 
@@ -1770,7 +1776,7 @@ namespace AGS.Editor.Components
 
             for (int i = 0; i < _loadedRoom.BackgroundCount; i++)
             {
-                _backgroundCache.Add(((IRoomController)this).GetBackground(i));
+                _backgroundCache.Add(_nativeProxy.GetBitmapForBackground(_loadedRoom, i));
             }
 
             foreach (RoomAreaMaskType mask in Enum.GetValues(typeof(RoomAreaMaskType)))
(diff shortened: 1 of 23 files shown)


Editor/AGS.Types/Room.cs
```diff
@@ -379,7 +379,22 @@ public Interactions Interactions
...
+            switch (mask)
+            {
+                case RoomAreaMaskType.WalkBehinds:
+                    return 1.0; // walk-behinds always 1:1 with room size
+                case RoomAreaMaskType.Hotspots:
+                case RoomAreaMaskType.WalkableAreas:
+                case RoomAreaMaskType.Regions:
+                    return 1.0 / MaskResolution;
+                default:
+                    return 0.0;
```
ivan-mogilko, line 393: Should not this throw exception instead of returning 0.0? As this happens in case of unknown mask type was passed.

ivan-mogilko:
In regards to "Copy walkable area mask to regions" command, does anyone know what that is for?

I recall this being a backward compatible hack for some ancient game versions, but I don't think it's useful today? If not it's best to just cut this out for ags4.
In the worst case a user may export/import mask for the same purpose, and that even is more versatile operation as mask could be applied to any type.

ivan-mogilko:
I've started testing these changes, and there are few problems I meet immediately:

* the room is not getting redrawn when you first open it, at all, it's like the panel retains previous surface until you move the window, or scroll the room (if it scrolls), or move zoom slider.
* after you make it redraw itself, it's just filled with black background, nothing else, regardless of editing modes.
* I cannot test any actual operations because of above, but only trying moving the room's scrollbar, I noticed that it became _significantly_ slower compared to current ags4 branch. I used [my own game with a very big scrolling room](https://github.com/ivan-mogilko/ags-lastfurious) for this test.
On the contrary, loading this room is much faster.
Maybe the fact that it does not draw properly affects this, so must retest when it displays actual room contents.

Editor/AGS.Types/Room.cs
persn: Yes probably, I was copying from `RoomStruct::GetMaskScale(RoomAreaMask mask)`, but I'll rewrite it

persn:
I agree that the "Copy walkable area mask to regions" button is oddly specific. Let's decide on whether we want to drop it from ags3 or ags 4 and then I'll take it out

persn:
I'll need to sit down and consider how to approach the performance of the drawing.

ivan-mogilko:
> I agree that the "Copy walkable area mask to regions" button is oddly specific. Let's decide on whether we want to drop it from ags3 or ags 4 and then I'll take it out

Both really, IMO it's useless now with import/export mechanic. Also one could add copy/paste mask commands later to do same thing in memory.

persn:
> I am concerned about having this overspecific function in the IRoomController interface too.

We decided against removing the button, but maybe I should still remove this functionality from the `IRommController` interface?

ivan-mogilko:
> We decided against removing the button, but maybe I should still remove this functionality from the IRommController interface?

I forgot about that bit. Yes I agree ofcourse.

persn:
@ivan-mogilko Ready for review again

> the room is not getting redrawn when you first open it, at all, it's like the panel retains previous surface until you move the window, or scroll the room (if it scrolls), or move zoom slider.

Could not reproduce this. But I discovered a bug where I didn't handle the enum case `RoomAreaMaskType.None``. It would crash when you opened the room editor without the room editor state having saved a previously selected mask. Maybe that's what you experienced.

> after you make it redraw itself, it's just filled with black background, nothing else, regardless of editing modes.
EDIT: ah, it finally drew background when I switched to one of the mask mode

See above

> but only trying moving the room's scrollbar, I noticed that it became significantly slower compared to current ags4 branch

I did some benchmarking with the `StopWatch`, and I found the performance issues to have 2 causes.
1. I was doing pixel-by-pixel alpha adjustment. I discovered that you can adjust a color matrix for dotnet `Bitmap` which is significantly less expensive. So now this is reduced to 0ms.
2. Loading images from unmanaged memory into managed memory is very expensive as it turns out. I measured times up to 1300-1400ms on your game with my CPU. This is probably going to get even more expensive when we load the images from disk so I decided to cache the images in managed memory.

In overall this appears to draw faster than it did before, with the expense of slightly longer room load. Benchmarks for `RoomSettingsEditor.bufferedPanel1_Paint()`:
Pixels | Branch | Time (ms)
---|---:| ---
240x320||
&nbsp; | ags4 | 9ms
&nbsp; | refactor | 39ms
&nbsp; | refactor-fix | 4ms
2784x2816||
&nbsp; | ags4 | 41ms
&nbsp; | refactor | 1300ms
&nbsp; | refactor-fix | 10ms

ivan-mogilko:
The room not displaying problem is now fixed.

There's a difference from usual AGS behavior in what seems to be area 0 of the mask actually drawn over room background as a dark grey overlay, which should not normally happen as area 0 is "eraser"/"no area".

Paint tools seem to perfom faster (not fully certain), but still not enough to make "freehand" work well. It takes editor few seconds to actually display a change after I hold and move "freehand tool" around, and it loads up 100% of a cpu core.

ivan-mogilko:
Also, frankly, I feel like all the "Mutate" method group is a wrong direction here. Maybe I am missing something, but what prevents us from keeping a compatible 32-bit bitmap at all time while user is editing the room, and only converting at room load (from 8-bit to 32-bit) and and room save (from 32-bit to 8-bit)?

persn:
> Also, frankly, I feel like all the "Mutate" method group is a wrong direction here. Maybe I am missing something, but what prevents us from keeping a compatible 32-bit bitmap at all time while user is editing the room, and only converting at room load (from 8-bit to 32-bit) and and room save (from 32-bit to 8-bit)?

You're probably right. The Bitmap extensions API is a result of mental fatigue after dealing with a lot of difficulties figuring out how to convert a 32-bit image to a 8-bit image, and also the other way around without transforming data. I'll redesign it to be simpler and more performant

I'll look into the other things you reported

persn:
> Would not it be more logical to have a mask bitmap in BaseAreasEditorFilter directly, and only do GetMask/SetMask when loading/saving room?

I tried this and ran into an obstacle because having the mask bitmap directly inside `BaseAreasEditorFilter` works very poorly when the mask is drawn inside `IRoomController.DrawRoomBackground`. The drawing routine can't access the cached mask obviously.

I considered this for a while, but I eventually remembered that in this PR I introduced a `IRoomController.Save()` method, which is typically invoked whenever the user has initiated a room save. I think the solution should be to move the saving of the backgrounds and masks from over to the save method. Then any uses of `IRoomController.SetBackground` and `IRoomController.SetMask` will set the bitmaps to the image cache, and no changes will be committed to disk until the save method has been invoked.

persn:
Ready for review again.

* Saving images to disk is now happening at the same time as saving the room properties, set and get mask operates on a cache.
* Added custom handling for simple drawing directly on a indexed image, instead of transforming a bitmap to 32-bit and back.
* Bunch of minor improvements like throwing exception on null arguments and such.

> There's a difference from usual AGS behavior in what seems to be area 0 of the mask actually drawn over room background as a dark grey overlay, which should not normally happen as area 0 is "eraser"/"no area".

Maybe I'm misunderstanding you here? But that is what happens on a copy of AGS 3.5 I have,

image

ericoporto:
Hey, for some reason the room editor bg is blurred for me :/

image

morganwillcock:
@ericoporto do use display scaling and override it for the editor(s)?

persn:
> Hey, for some reason the room editor bg is blurred for me :/

I see the problem now thank you, it's the same as @ivan-mogilko was talking about. The problem is that the masks doesn't have alpha channel, they have black pixels where the transparency would go. I just need to remap the color for palette index 0 and it should be fine.

persn:
Try now

ericoporto:
image

Left side is AGS without this and right side is this branch.

So the gray tone is darker instead of lighter - but this is minimal. What bothers me is that everything is blurred.

> @ericoporto do use display scaling and override it for the editor(s)?

I haven't particularly set anything since I just downloaded both.

image

persn:
> Didn't any of these properties worked?

I honestly can't see any blur when trying on my machine with my game so I don't know. I'll download your game later and test it out myself. Alternatively you could just quickly test it yourself? Just throw those line of codes in at the top of `void IRoomController.DrawRoomBackground(...)` in `RoomsComponent.cs`. If the quick hack works I'll figure out where to put those codelines later

ericoporto:
I put these:
```C#
                e.Graphics.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.NearestNeighbor;
                e.Graphics.SmoothingMode = System.Drawing.Drawing2D.SmoothingMode.None;
```

At line 342 and 343 of RoomSettingsEditor.cs and it fixed!

persn:
That looks decent, if you want you can push the branch online and I'll cherry pick it off your branch

ericoporto:
I am too lazy for that, but thanks 😁

persn:
Okay I pushed the fix

ericoporto:
Hey, so, I have a bad news. This version of the room editor is totally broken on Wine. The old one ran fine on Linux but this new one doesn't work:
- scrolling has artifacts, but this is minor
- drawing masks, the coordinates are random, you click one place, the place clicked is elsewhere.

Occasionally I get the error:

```
Error: Index was outside the bounds of the array.
Version: AGS 3.99.99.0

System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at AGS.Editor.BitmapExtensions.FillIndexedRectangle(Bitmap bmp, Int32 color, Point p1, Point p2, Double scale)
   at AGS.Editor.BaseAreasEditorFilter.DrawFilledRectangleOntoMask()
   at AGS.Editor.BaseAreasEditorFilter.MouseUp(MouseEventArgs e, RoomEditorState state)
   at AGS.Editor.RoomSettingsEditor.bufferedPanel1_MouseUp(Object sender, MouseEventArgs e)
   at System.Windows.Forms.Control.OnMouseUp(MouseEventArgs e)
   at AGS.Editor.FreePanControl.OnMouseUp(MouseEventArgs e)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ScrollableControl.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
```

This is using Wine 6.

https://user-images.githubusercontent.com/2244442/106975752-ddf9df00-6735-11eb-866f-d3492edb1058.mp4

rofl0r:
tbh, i don't even see the point of this PR. it replaces well-tested well-performing code with an implementation in another language that's hard to make fast. is it just for a personal preference of implementation language, or an exercise to try out refactoring features of an IDE, or something else ? the PR has no explanation whatsoever about why OP thinks its an improvement.

ericoporto:
It reduces friction to move away from C++/CLR the MSVC exclusively Windows part for managed to cpp. You don't use Wine rofl0r, just go away.

rofl0r:
> You don't use Wine rofl0r

i actually do use it, but i couldn't get the .net part working, that's why i'm stuck with ags 2.72 for editing.
but my comment isn't about it not working in wine, i was thinking this since this PR was opened some weeks ago but i've bitten my tongue so far...

ivan-mogilko:
> You don't use Wine rofl0r, just go away.

@ericoporto please don't talk to people like that, this is becoming a bad habit to have such replies in our project.

> It reduces friction to move away from C++/CLR the MSVC exclusively Windows part for managed to cpp. 

To be honest, while I sort of understand the direct reasons for this pr, at the same time I am in big doubts about its usefullness in the larger scale. Maybe this is because I am missing something. Unfortunately there's still no written design goal for this project ( which should have been created long ago) so it's often unclear where we are going, and whether we are going into same direction, or see the future variant of the editor and engine in similar way. I am relying that others know what they are doing, but I get uneasy feeling sometimes.

To clarify, what is done here, is this because C++/CLR is bad and/or it's bad to have it in this particular project? I recall there was a discussion on this years ago, where people stated that they cannot move editor to Mono while there's a native assembly. Is this still stand true?

The thing that bothers me most in this context, is that in my honest opinion transfering this functionality to C# could be a mistake in the long term. Similar thing had been done with the game file writing several years ago, and I now think that was a wrong move (and that part should be cut out as a part of #1188).

To elaborate, since certain time I lean towards idea of having the engine to perform scene rendering in the editor (room and gui at least). Either by having parts of it built in a DLL and then constructing a scene in it by passing data, or path to game files, etc, or by literally running it and passing window handle to draw on. This is by no means a trivial task, and I might not see all of the potential issues with this. But I thought if we could eventually pull that of, that would give us number of extra benefits:
* Much less code duplication (we already have parts of drawing duplicated in 2 languages).
* More authentic look, as user will see scenes exactly as they will be in the game. Even now editor does not show many things as they are suppose to look like: it does not scale objects in the room, does not colorize them in certain regions, etc (and it does not have all the logic necessary to do all of that). And if engine will be able to launch its plugins - then we'll have even extra rendering performed by engine plugins - in the editor.
* Following above, no need to invent .NET solutions for each quirk that engine already does (some of these may require more libraries in the editor).
* We might end up having a "edit mode" inside the engine, which may also be usable outside of the editor.

morganwillcock:
I think it is probably worthwhile to consider that:

* the existing setup also fails to represent the room as it is in the game
* the existing C++ code that is called via the proxy isn't providing any kind of bridging to the Engine, rather it is on the only feasible way to read the data due to premature resource packing by the Editor workflow
* the use of the native proxy is already hampering data operations and is the only thing that stops the spritecache being rebuilt from source files
* I don't use WINE, but I also don't want to use Windows. The Editor already runs cross-platform but it cannot do any editing operations because of the reliance on the native proxy
* the tight coupling of the Engine and Editor makes its more difficult to refactor both sides - since this is AGS4 that seems an advantage and this is also reflected in the other suggested development tasks. I accept that the current plan needs ~~writing~~ expanding upon but I don't think anyone will argue for the exact opposite of the general consensus to split out the tooling
* since this is AGS4 it seems the best place to try something new and see how it pans out, especially since I don't think the change has any impact on project data at all
* it would be pretty good if the engine could eventually be used for editing, or be used to render the previews, but that is functionality that would either optionally or fully replace these changes. I don't see how anything here impedes development of any alternative room editor in the future

At the very least I would hope stays around as a feature branch with CI builds, because anyone looking to improve (rewrite) the Editor needs some scope for flexibility.

ivan-mogilko:
> since this is AGS4 it seems the best place to try something new and see how it pans out, especially since I don't think the change has any impact on project data at all

From what I get, removing native parts leads to independency on storage in native code, and thus makes it easier for C# programmer to implement a move to different project format. E.g. the ticket #469 the @persn mentioned in the pr's description.

ivan-mogilko:
It's funny reading #469 now, which I wrote 2.5 years ago myself, mentions that everything for rooms should be done in C#...
Guess at that time it looked like the most straightforward way to decouple everything.

ericoporto:
Moving the Room Editor in Engine is trivial with a proper GUI like ImGui. Communicating with the Editor would require a lot of effort or minor effort depending on familiarity with networking between Engine and Editor. Everyone here appears to hate using libraries for some reason, so both the communication and the GUI is not trivial at all if the intent is being cross platform. The NIH is high here.

Removing the CLR/C++ parts is trivial using SWIG as long Windows programmers can live with compiling the Editor in two steps - once the Native and later the C# with the dll/so/dylib in place. Files are slow in Windows, I don't think completely removing the Native.dll is the way, objects in memory are faster. 

People tend to give up open source projects after they spent two months working on a PR that could have been refused at infancy if this was everyone else's opinion. I want the non binary rooms because they are better to work with in Source Control software. If the error in this PR can be solved by proper debugging I prefer to keep people happy. People here are toxic towards any language that are not "my favorite language ™️".

It's all very demotivating.

rofl0r:
> Everyone here appears to hate using libraries for some reason

that's completely untrue. AGS at current uses at least 10 libraries, some of them being in-tree and some of them external. the thing is that every library dependency has a cost: it usually needs to be kept up-to-date, may have bugs, may not compile for some platform, needs to be fetched, needs to be built, and generally makes the build process more complicated, the resulting binary bigger, and maintenance more difficult. additionally every in-tree library will make the git repository bigger, and thus will require more time to download/checkout, especially when it's a huge library that often needs to be updated.

> Moving the Room Editor in Engine is trivial with a proper GUI like ImGui

ah, seems you already have your new favorite library focused.
but what would be the purpose ? the gui is already in place, in this case it's just a matter of embedding the rendering result of the engine in the editor and communicate option changes like "don't display masked out areas" if the user toggled some setting.
having yet another gui insider the room display area would certainly not improve things.

> Removing the CLR/C++ parts is trivial using SWIG

hmm? what could SWIG achieve here that can't be already achieved using some minor changes in the build process ?

> People tend to give up open source projects after they spent two months working on a PR that could have been refused at infancy if this was everyone else's opinion.

this could be avoided if those people started a discussion first about what they want to achieve or at least clearly state in the PR description why they think an improvement is necessary and how their proposal actually improves things, instead of just changing things without explanation and assuming it will be merged because nobody has the guts to say "no".

ericoporto:
Rofl0r, you don't have guts. You want code to freeze in time. You are AGAINST C++, you have a license that reflects your opinion, so whenever you say that one should write in C++ something that was in other language I just know you are trolling.

ivan-mogilko:
I am now sorry for bringing this question up.

I may actually agree with what @morganwillcock said above:
> the existing setup also fails to represent the room as it is in the game
> the existing C++ code that is called via the proxy isn't providing any kind of bridging to the Engine,
> it would be pretty good if the engine could eventually be used for editing, or be used to render the previews, but that is functionality that would either optionally or fully replace these changes. I don't see how anything here impedes development of any alternative room editor in the future

I was going to accept this PR anyway as soon as it gets fixed, so I'll accept it when the newly found problem with WINE is fixed, and if testing will show that perfomance problem is now fixed too.

---

However, to clarify -

> I want the non binary rooms because they are better to work with in Source Control software. 

I mentioned this before, but this pr is not directly required to non binary rooms. It's more a cleaning up work because we were going to implement new room source format in C# anyway, and it will be definitely more straightforward to work on that if it's all in one language and one "part" of the program. In theory that could be accomplished by reusing native part too, but that would create extra burden for both development and maintenance.

> People here are toxic towards any language that are not "my favorite language"

This is not even related to the points I brought above. It was not about language, it was mainly about duplicating work that engine is already responsible for.

morganwillcock:
I gave it a quick try and performance seemed good, but there does seem to be an issue with the drag selection not aligning to the image. Here it is at 600% zoom:

pixel_snap

Selections seem to be half a pixel out and then jump when the mouse button is released.

morganwillcock:
The room edges have the same issue, so possibly it is the background image itself that is half a pixel out. If this is hardware accelerated then that might be an expected problem of texel -> pixel conversion.

ivan-mogilko:
> Selections seem to be half a pixel out and then jump when the mouse button is released.

I've seen this with old code a long while ago, it could also be a precision loss with room zooming.

ericoporto:
Found the culprit for the error on Wine.
https://user-images.githubusercontent.com/2244442/107146836-507ee080-6929-11eb-9501-82903daf1d8e.mp4

Line 18 of BitmapExtensions.cs, the culprit is `Image.GetPixelFormatSize(bmp.PixelFormat)`

This is always returning 32 but should return 8 for the indexed bitmap of the masks.

persn:
> Found the culprit for the error on Wine 6.

Cool thanks, setting up Wine has been a pain for me in the past so I didn't want to do that. I may be able to make `BitmapExtensions.GetColorDepth(..)` more robust so it works on Wine 6 too though. If you still have Wine 6 accessible, could you tell me what is the value of `bmp.PixelFormat`? It should be `PixelFormat.Format8bppIndexed` probably.

> half pixel offset can be fixed with the following:

Okay I'll push it and you and @morganwillcock can verify if it works. It's possible it might be a math error too though. The relevant code is `BitmapExtensions.FillIndexedRectangle(...)` if anyone wants to help spot it, assuming it exists.

> Occasionally I get the error

Reproduced by using draw rectangle, clicking on the canvas, and dragging outside of the picture to the bottom right corner. I'll take a look at it

ericoporto:
image

@persn I am so so really sorry. I think I had a mistake in my Wine Prefix.

I tried to rollback to Wine 6.0 and couldn't reproduce the error anymore. I created a new wineprefix and made a new wine6 prefix and cannot reproduce the error with Wine 6.0.

I recently was experimenting for another issue (#1113 ) and I think I actually had 3.20 at the time I reported the error on Wine. Sorry for the huge false alarm.

There's no error in any recent wine.

persn:
@ericoporto No worries

Ready for review again I guess

Editor/AGS.Types/Interfaces/IRoomController.cs
```diff
@@ -21,26 +21,76 @@ public interface IRoomController
...
+        /// and overlays one of the room masks onto it.
+        /// </summary>
+        void DrawRoomBackground(Graphics g, Point point, int backgroundNumber, double scaleFactor, RoomAreaMaskType maskType, int maskTransparency, int selectedArea);
+        /// <summary>
+        /// Scales all the room's masks according to the <see cref="Room.MaskResolution"/>,
+        /// execept for the <see cref="RoomAreaMaskType.WalkBehinds"/> which retains the
+        /// resolution of the background image. If the <see cref="Room.MaskResolution"/>
+        /// has a resolution of 2 it will halve the masks in both dimensions.
+        /// </summary>
+        void AdjustMaskResolution();
```
persn, line 89: Maybe this shouldn't be a part of the public interface? Maybe do an automatic scale when room mask resolution is changed instead

morganwillcock:
There still seems to be a selection problem, although the pixel offset is now fixed. It is more obvious if you drag a box from right to left, but the boundary jumps when the mouse button is released. Also the wrong cursor type appears until the cursor is moved again.

pixel_snap-02

persn:
Math error possibly. What is the width and height of the room? Also what is the Mask resolution?

The cursor thing is odd though, as far as I know I haven't touched any code related to cursors. Does it not happen on the ags4 branch?

persn:
Actually I just tested the cursor thing and it happens on ags4 too. Maybe open a seperate bug for that one. I'm already doing so much in this branch so I don't think I should do more

ericoporto:
@persn how are you drawing the preview code? What's the difference between that and the one that actually draws in the "bitmap" ?

persn:
By preview you mean the rectangle that is visible when dragging the mouse cursor? I haven't touched that code

Editor/AGS.Types/Interfaces/IRoomController.cs
ivan-mogilko: > Maybe do an automatic scale when room mask resolution is changed instead

Could you elaborate on this?

ivan-mogilko: One thing I will agree with is that the way function is made right now seems too obscure. It had to be called manually after you change Room.MaskResolution property, which is not good. Maybe make it accept new resolution factor and then call it and let it set that factor to the room inside the implementation instead?

ericoporto:
It looks like a difference between a less than and a less than or equal , the code can't be the same since the drawn thing has the mask scalling.

If you look here you can see how it should go about (note the less than or equal sign)

https://github.com/adventuregamestudio/lib-allegro/blob/c69e7bb48cee6e6b8730f557feb688cf2526bd55/src/gfx.c#L631

I think the intersects on your Bitmap drawing code doesn't work like you think it works, so you get the less than effect.

Editor/AGS.Types/Interfaces/IRoomController.cs
persn: When the mask resolution for a room is changed in the room editor it will invoke the `AdjustMaskResolution()` and scale down or up the actual bitmaps, except for walkbehind mask. The bitmaps will physically change size so if your mask resolution is 1, and then you set mask resolution = 2 for a room with 320x240 dimensions the masks will become 160x120 on disk.

I'm thinking that maybe this isn't the kind of feature we want to expose to a plugin developer?

persn: > Maybe make it accept new resolution factor and then call it and let it set that factor to the room inside the implementation instead?

Good idea actually. I think this might be the way to go

persn:
@ericoporto Thanks

Editor/AGS.Types/Interfaces/IRoomController.cs
persn: Changed it

persn:
@morganwillcock I hope you have time to verify if the rectangle draw looks better

ericoporto:
https://user-images.githubusercontent.com/2244442/107702502-1733d100-6c99-11eb-8609-243dfbca6dfb.mp4

Recorded a bit of usage. Changed mask resolution and things. It feels it works. I noticed sometimes the room was with the selected layer like it was not selected - when I deselected the room editor - but I don't know if this is a bug that is already there or not. Can't notice any jumping out of place thing.

persn:
> I noticed sometimes the room was with the selected layer like it was not selected - when I deselected the room editor 

Sorry but this doesn't make sense to me. Could you rephrase what the problem is?

morganwillcock:
Similar issues with the line tool, and the line isn't straight.

pixel_snap-03

ericoporto:
> Could you rephrase what the problem is?

@persn if you watch the video at 0:45, the room gets deselected, and later I try to draw expecting to be at layer 2, but my tool is somehow the eraser - you can see the layer selector hasn't changed. I don't know if it's always being like this or not, but felt weird.

persn:
> if you watch the video at 0:45, the room gets deselected, and later I try to draw expecting to be at layer 2, but my tool is somehow the eraser - you can see the layer selector hasn't changed. I don't know if it's always being like this or not, but felt weird.

This is also happening in the `master` branch, so I'm not going to do anything about it. Someone should open a bug issue about it. It looks like the selected area changes when you change area in the properties pane, however when you select the room itself from the property pane, the "breadcrumb" (don't know what we call it) doesn't update correctly

persn:
@morganwillcock try now

morganwillcock:
It is still differerent and effectively curved at one end. You can test it on the background in the template. Here is the top-left corner, joining 2 stars together...

Release version:
release

PR version:
pr

persn:
Aha. It has a 2 pixel merge at the start there, I see though that in the release version it has the same 2 pixel merge, just in the middle of the line instead. There's probably some difference in the numbers rounding off from before and after. I'll take a look

persn:
@morganwillcock Try now

ivan-mogilko:
Testing this after a while - drawing masks is now fast again. Although tbh I still don't get some things in new code which seems like extra operations when painting is done, but I did not investigate much and maybe this will be improved later.

I noticed that region colors are slightly different from original implementation: when mask transparency is at minimum inactive regions are painted dark grey while originally they were light grey, and with max transparency active region looks much more bright in tone. But idk if that matters.

ivan-mogilko:
I also think the line drawing issue reported by @morganwillcock is fixed (tried connecting these two stars in default template).

persn:
> I noticed that region colors are slightly different from original implementation: when mask transparency is at minimum inactive regions are painted dark grey while originally they were light grey, and with max transparency active region looks much more bright in tone. But idk if that matters.

I forgot to mention this earlier I see, but I could never figure out exactly what the original code did when applying the mask transparency. As far as I can tell I'm using the same math https://github.com/adventuregamestudio/ags/blob/ags4/Editor/AGS.Native/agsnative.cpp#L745-L760

However the fix I just pushed should make the grayscale brighter and much more closer to the original.

ericoporto:
You use 0->255, old code used 255->0, I think

persn:
I don't think that's the case, I already tried inverting the color but it got ugly white.

ericoporto:
Uhm, found this bit above in the lines you sent @persn , not sure but they look suspicious :)

```
   // Blank out the temporary palette, and set a shade of grey
    // for all the hotspot colours
    for (int aa = 0; aa < 256; aa++) {
      int lumin = 0;
      if ((aa < MAX_ROOM_HOTSPOTS) && (aa > 0))
        lumin = ((MAX_ROOM_HOTSPOTS - aa) % 30) * 2;
      thisColourOnlyPal[aa].r = lumin;
      thisColourOnlyPal[aa].g = lumin;
      thisColourOnlyPal[aa].b = lumin;
    }
```

persn:
The math there is a little funky because when it passes index 30 it'll start recycling already used gray tones.
`(50 - 1) % 30 == (50 - 31) % 30 == 19`
I'm not sure if that is a bug though, it's probably a relic from a time when the room hotspot limit was lower? I carried over the math as it is into this PR, I don't want to tamper with it since it's unrelated issue and I'm already doing a lot of work here

ivan-mogilko:
I think this may be merged, unless someone notices another math error with drawing lines etc? But according to my last test these were fixed.

Editor/AGS.Editor/Components/RoomsComponent.cs
```diff
@@ -1545,20 +1543,38 @@ void IRoomController.SetBackground(int background, Bitmap bmp)
...
             }
             else
             {
                 _backgroundCache[background]?.Dispose();
-                _backgroundCache[background] = (Bitmap)bmp.Clone();
+                _backgroundCache[background] = newBmp;
+            }
+
+            // If size or resolution has changed, reset masks
+            if (newBmp.Width != _loadedRoom.Width || newBmp.Height != _loadedRoom.Height)
```
AlanDrake, line 1574: Ah... so this is why masks are never resetted. This check is always false because _loadedRoom.Width/Height have already been changed to that same size.

But even fixing that issue, ClearMaskCache() disposes of the masks without recreating them, so we later get null pointer exceptions.

PR #1141, opened 2020-12-06
CI fixes for ags3 sdl2 part1: Windows build

ericoporto:
I would like to do the fixes needed for CI build in separate PRs per OS to make it easier to understand the changes

- removes full allegro mentions from cirrus pipeline in Engine and Native build tasks.
- adds environment variables for SDL2 and SDL_sound libraries.
- Windows Dockerfile now uses Cirrus CI contrib cmake image as base
- Windows Dockerfile no longer fetches allegro
- Windows Dockerfile now brings SDL2 binary dependences and builds SDL_sound library.

This builds the engine succesfully - you just need `SDL2.dll` in the same directory as the `acwin.exe` for it to work!

https://cirrus-ci.com/github/adventuregamestudio/ags/pull/1141

**PROBLEMS:** right now AGS Native is failing to build in the CI with the following error:
```
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um\winnt.h(2483): 
error C2338: Windows headers require the default packing option. Changing this can lead to memory corruption. 
This diagnostic can be disabled by building with WINDOWS_IGNORE_PACKING_MISMATCH defined. 
[C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Editor\AGS.Native\NativeDLL.vcxproj]
```

I decided to open the PR because I was having a hard time tracking down this - I am not very experienced with Windows.

This error has a possible fix on Stack Overflow: https://stackoverflow.com/questions/56570681/static-assertion-failed-with-windows-headers-require-the-default-packing-option

In my Windows 10 Machine with VS2019, **I don't get that error**. The only error I got was a missing `allegro.h` which was fixed by adding `..\..\libsrc\allegro\include` under  `<AdditionalIncludeDirectories>` in `Editor/AGS.Native/NativeDLL.vcxproj`.

In the container, cl.exe is Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24210 for x86
In my machine, cl.exe is Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24245 for x86

The docs mentions "The CLR will also not respect align": https://docs.microsoft.com/en-us/cpp/dotnet/how-to-migrate-to-clr?view=msvc-160

There's a warning [here](https://docs.microsoft.com/en-us/cpp/build/reference/zp-struct-member-alignment?view=msvc-160):
> Memory corruption may occur if the /Zp setting is changed inside the Windows SDK headers. The headers aren't affected by any /Zp option you set on the command line.

So I don't really get it how the errors appears if they were not supposed to be affected by `/Zp#`

---

_For context:_ **I don't know if this is related yet**, but Cirrus CI has recently changed their base images in docker hub ([here](https://hub.docker.com/r/cirrusci/windowsservercore/tags?page=1&ordering=last_updated)) which had some conversation going on it in this issue https://github.com/cirruslabs/docker-images-windows/issues/16 . I wanted to  make a new image as base with VS instead [here](https://github.com/ericoporto/docker-minimal-dev-ags/blob/main/Dockerfile), but even using my image as a base locally in my computer, I still have the error when building AGS Native in the _container_. 

A note, I have the container of this PR on my computer and I can navigate the files in it, so if someone has ideas of things to look into I can poke the files and give answers.


ericoporto:
List of files giving packaging mismatch compilation errors right now in AGS Native project:

- `agsicons.cpp`
- `agsnative.cpp`
- `MSSCCI.cpp`
- `NativeMethods.cpp`
- `ScriptCompiler.cpp`
- `..\scintilla\src\PlatWin.cxx`
- `..\scintilla\src\ScintillaWin.cxx`

ericoporto:
Found a much more interesting error, Debug builds don't use `/Zp4` at all! In fact they build but eventually they give a different error:

```
ResourceCompile target) -> 
         TRACKER : error TRK0005: Failed to locate: "rc.exe". 
The system cannot find the file specified. 

[C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Editor\AGS.Native\NativeDLL.vcxproj]
```

It appears it's some misconfiguration with SDK! Just need to figure out what exactly is it!

AlanDrake:
rc.exe, is the Microsoft platform SDK Resource Compiler, isn't all that stuff supposed to be included in the PATH variable?

ericoporto:
I use the build steps in the gist here to reproduce the issue in my container: https://gist.github.com/ericoporto/e260e4b46df3d36fe62264bda4ac17c7

I have to change the first env var to Debug and I actually run the commands by copy pasting. My guess is it's something wrong on the installation in the container, maybe some package installation order matters. But I think (current guess) is `rc.exe` should come from Win 8 SDK and fails because it instead looks into a wrong Win 10 SDK path.

ericoporto:
Editor and Acwin.exe should build now. This is no longer WIP and can be put under review. 

MacOS will be a separate PR, and it will be a short one (it's small code fixes and no touches in the CI). Linux will follow in a separate PR too (it's a ton of things!).

ivan-mogilko:
> Android I have it working but no idea how to ressurect old behavior, so I think it will be easier just having it in ags4 with new behavior and let someone else try their ways with Android in ags3.

Could you elaborate, what is this "old" and "new" behavior?

ericoporto:
The old and new mention are strictly about touch to mouse translation in the port.

The Allegro Android port has configurable mouse behavior, and the behavior uses gestures to provide extra buttons, for instance, finger hold on the screen doesn't translate to mouse left button holding down and instead it process the gesture of finger hold as a different action - I think it's mouse right click. This prevents some functionality that continuously check mouse button state in AGS Script to direct run on the Android port, most notably, the Drag and Drop module doesn't work.

SDL2 by default translates finger tap as left click and finger down as mouse left button down - but has no default mapping of other mouse buttons. It gives additional touch handling events if one wants to manually handle it. I believe the best approach is to port these specifics in some way to ags script so the developer can handle multitouch from Script, which would allow handling things like an ags script and GUI joystick overlay or other forms of multitouch interactivity on the screen. 

Previous Android port mouse behavior can possibly be restored but it's not a simple task - which would not be a problem for AGS4 since script API changes for touch handling are possible and could be standard to all ports.

I can open a ticket - I also have done experiments with APIs I can upload for testing.

ivan-mogilko:
> SDL2 by default translates finger tap as left click and finger down as mouse left button down - but has no default mapping of other mouse buttons. It gives additional touch handling events if one wants to manually handle it

Well, guess it may be possible to make some custom behavior by handling sdl touch events then.

Is Android working with sdl port already in ags3, or we'd need to additionally apply those changes you've shown in your branch? I am becoming bit confused again with ags3/ags4 branches, and would like to stick to ags3 sdl port for now. After it's done in ags3 and builds reliably at least on Windows and Linux, we will merge sdl to ags4 and manage its own issues there.

ericoporto:
OK! I can open later today a ticket for AGS3 SDL2 ports. I will walk around the different ports and the changes. I want to open a Touch on AGS4 ticket too just to prevent discussion from one to leak in the other.

> Well, guess it may be possible to make some custom compatible behavior by handling sdl touch events then.

This is definitely possible and doing so would work with touch not only on Android but where there's a touch device.

ivan-mogilko:
Looks like I somehow forgot to remove allegro.props from everywhere, but you also left it in Engine project - is there reason for that?
I think the file itself should be removed as it's no longer used.

ivan-mogilko:
I also have a request: when you are writing commit message, please make an extra empty line between the header and further text: this makes only first line a commit "title", and rest of text becomes "description".
https://chris.beams.io/posts/git-commit/#separate

If you don't do that then all your commit text becomes a "title", and is displayed as one huge line when you look at git history.

ericoporto:
I can amend the commit messages to rewrite them so it's in this format! 

> Looks like I somehow forgot to remove allegro.props from everywhere, but you also left it in Engine project - probably an oversight?

I forgot about it there, I can remove it!

My worries with this PR:
- Does the build results on Cirrus-CI actually work on Windows 7? I lack a Win7 machine to test.
- I changed the alignment for release builds on AGS Native, this didn't seen to affect anything on my tests, but is there a hard reason for using `/Zp4` before? Without the change it failed to build in the container.

ericoporto:
Removed allegro.props and removed it from the Engine.App.vcxproj and also fixed the first commit message. :)

ivan-mogilko:
> I changed the alignment for release builds on AGS Native, this didn't seen to affect anything on my tests, but is there a hard reason for using /Zp4 before?

This may be remains of some old rule we missed and did not remove earlier. I suspect it was set because in the old versions ags wrote whole data structs directly from memory into stream instead of doing this member by member, and therefore having exact alignment was important to keep data formats. Now it should not be necessary.

I am not able to test CI build because it keeps failing with "agent not responding" again. But I can build everything myself locally.

ericoporto:
Hey, it appears it's the size of the Windows container, I will try some tricks to lower it's size without changing what needs to be in it.

ericoporto:
We used to use `vcpkg` to build Xiph libraries but some change of vcpkg requirements made it no longer build in the Docker image we had, I tried to make a an image with more of VS but the image was too big to work in the gke, timeouting when downloading the docker image.

For now I switched to build the Xiph libraries in Azure DevOps and am pushing it directly to NuGet: https://github.com/ericoporto/xiph-for-ags

I am pulling these binaries when making the Docker image, which allow use a smaller docker image.

ivan-mogilko:
Merged #1151 and into sdl2 branch, could you fix the file conflict now?

ericoporto:
Yes, I will fix it today. :)

ericoporto:
It's building now on Windows parts of the CI! I rebased and amended the commit messages since now the Xiph prebuilt was already merged.

Because the SDL_Sound here has been updated to a newer version, the size of the `acwin.exe` build will change a little when compared to the one on the forums.

I noticed 4 warnings when building `acwin.exe` for Release (SDL Sound is Release too) in the CI that I don't know if they are meaningful:

```
     4>SDL_sound.lib(snd_fx.obj) : warning LNK4049: locally defined symbol _rand imported [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Solutions\Engine.App\Engine.App.vcxproj]
     4>SDL_sound.lib(load_abc.obj) : warning LNK4049: locally defined symbol _rand imported [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Solutions\Engine.App\Engine.App.vcxproj]
     4>SDL_sound.lib(load_abc.obj) : warning LNK4049: locally defined symbol _srand imported [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Solutions\Engine.App\Engine.App.vcxproj]
     4>SDL_sound.lib(load_abc.obj) : warning LNK4049: locally defined symbol __time64 imported [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\Solutions\Engine.App\Engine.App.vcxproj]
```

PR #1143, opened 2020-12-06
Update AGS.Editor.Full.sln to fix Compiler Release build

ericoporto:
- Mixed Platforms build of AGS.Editor.Full.sln uses Release_MD for
  Compiler and Compiler2 projects.

Alternatively it could build Common.lib instead of Common_MD.lib on `Release` directory.


ivan-mogilko:
I missed this one, but what is Mixed Platforms used for?

ericoporto:
I have no idea! But it's the default for the AGS.Editor.Full.sln !

ericoporto:
(maybe it's because some stuff has Win32 as a target and the C# code has Any  CPU as target

persn:
Yeah that's probably why.

ericoporto:
https://social.msdn.microsoft.com/forums/vstudio/en-US/81c72e8b-6335-4bf4-b7c0-b5c322edcaee/mixed-platforms-vs-any-cpu

> if you have a C# project and a C++ project, typically "Mixed Platforms/Debug" will map to "Any CPU/Debug" for the C# project and "Win32/Debug" for the C++ project.

PR #1149, opened 2020-12-10
CI: Add counter to force regeneration of Windows container

morganwillcock:
Should hopefully get the Windows builds working again...

### DIFF
--- a/ci/windows/Dockerfile
+++ b/ci/windows/Dockerfile
@@ -1,3 +1,6 @@
+# increment number to modify file hash but keep the same build configuration
+# 1
+
 # base will already have Chocolatey installed
 FROM cirrusci/windowsservercore:2019
 


morganwillcock:
It looks like the intermediate containers used are too prone to breaking change to rely on long term, so I think this either needs to be done from scratch and we cache our own containers built using the lowest level source or we find another place to run the builds that has some sort of stability guarantee.

ericoporto:
I created two base images here on dockerhub:
https://hub.docker.com/repository/docker/ericoporto/min-ags-dev-env
-> Tried to achieve minimum size but for some reason is slightly bigger than the other image. On the pro side, prefers chocolatey packages which are easier to understand.

https://hub.docker.com/repository/docker/ericoporto/vsmanagedcpp
-> This one follows what I understood is recommended by MS and is also smaller in size.

ericoporto:
Unfortunatelt Cirrus or GKE network runs at dial up speeds and it takes more than an hour and timeouts when downloading `min-ags-dev-env` -> https://cirrus-ci.com/task/5908932738678784

PR #1150, opened 2020-12-12
AGS4 - Dockerfile changes to fix Windows CI builds

ericoporto:
- some RUN commands reduced in rearrange to minimize layers (reduce docker size)
- no vcpkg bootstrapping, pulls prebuilt xiph libraries from nuget
  (vcpkg was not building with current VS tooling in this image)
- docker image size lowered a bit and is now 11GB

AGS3 SDL2 Agent Failures when asked about to Cirrus CI support by mail was reported due to failure in pulling the images in less than 15 min, causing agent timeout, due to image size - and I think due to network speeds.

I am guessing the failures in this branch is similar, so this PR aims to reduce the resulting image size slightly by combining layers (RUN commands). Additionally, the temp directory is cleared if needed before pushing each layer.

Unfortunately vcpkg doesn't build in this environment currently and we used it to build Xiph libraries (Ogg, Theora and Vorbis), so to counter this problem I am pulling [prebuilt ones I pushed on NuGet](https://github.com/ericoporto/xiph-for-ags).

---

If this works and is accepted, it can be replicated for AGS3 branch too.

### DIFF
--- a/ci/windows/Dockerfile
+++ b/ci/windows/Dockerfile
@@ -1,38 +1,48 @@
 # base will already have Chocolatey installed
 FROM cirrusci/windowsservercore:2019
 
-# no temp folder exists by default
-RUN mkdir %TEMP%
+# if no temp folder exists by default, create it
+RUN IF exist %TEMP%\nul ( echo %TEMP% ) ELSE ( mkdir %TEMP% )
 
 ARG VISUALCPP_BUILD_TOOLS_VERSION=14.0.25420.1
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Maybe do this on ags3 branch first? At least that still can be merged "up".

ericoporto:
I opened #1151 , this can be closed if wanted. :)

ivan-mogilko:
Merged #1151 and into ags4 too.

PR #1152, opened 2020-12-13
AGS3  SDL2  ci macos fixes

ericoporto:
Fixes to build on MacOS
- adds unix headers where missing in allegro
- fixes so SDL2 is picked up by SDL2_sound correctly
- updates to newer SDL2_sound and no longer overrides SDL2_sound CMakeLists.txt

With this, AGS is sucessfully built for MacOS in the CI. WHat we had in place for the CI in MacOS still works, so no actual changes where made there.

TODO: Figure out why initial window is sized as the native resolution. FullScreen works good.


ivan-mogilko:
Merged #1151 , will see if it makes this PR pass.

libsrc/allegro/src/unix/ufile.c
```diff
@@ -48,6 +48,9 @@
 
 #define PREFIX_I "al-unix INFO: "
 
+#ifndef _WIN32
```
ivan-mogilko, line 51: I don't think _WIN32 is necessary, this is unix/ufile.c, not supposed to be included on win.

ivan-mogilko:
Regarding changes in allegro sources, at least two of these headers - <unistd.h> and #include <fcntl.h> - are included in alucfg.h, which in turn included to alconfig.h and allegro.h, and therefore supposed to end up included in the *.c files.

Also, files sys/types.h and sys/stat.h are already included in ufile.c, under ALLEGRO_HAVE_DIRENT_H macro check.

If missing declarations are reported by compiler, maybe current includes are not done right, or some flags missing?

ivan-mogilko:
Regarding changes in allegro sources, at least two of these headers - <unistd.h> and #include <fcntl.h> - are included in alucfg.h, which in turn included to alconfig.h and allegro.h, and therefore supposed to end up included in the *.c files.

Also, files sys/types.h and sys/stat.h are already included in ufile.c, under ALLEGRO_HAVE_DIRENT_H macro check.

If missing declarations are reported by compiler, maybe current includes are not done right, or some flags missing?

libsrc/allegro/src/unix/ufile.c
ericoporto: Right, I will remove it!

ericoporto:
> Ah, it's macos, so it's including alosxcfg.h instead. Since you are using unix sources for mac too (judging by comment in cmakelists), maybe missing includes <unistd.h> and <fcntl.h> should be added there, similarily to how they are in alucfg.h. 

Uhm, thanks! Yes, it should be there. I will add it in the iOS ones too. Those should go in all non Windows platforms as far as I can tell. 

> But ALLEGRO_HAVE_DIRENT_H is already declared in alosxcfg.h, so no idea what's wrong there. Perhaps double check that platform detection works?

Right, I will. 

Hey, I noticed [alconfig.h](https://github.com/adventuregamestudio/ags/blob/ags3--sdl2/libsrc/allegro/include/allegro/internal/alconfig.h) is not guarded by `ifndef define` so it's included only once, should I guard it too?

ericoporto:
Should be good now. Also rebased the branch so the checks run on top of previous merge.

Moved the unix missing header <unistd.h> and <fcntl.h>  to [alosxcfg.h](https://github.com/adventuregamestudio/ags/pull/1152/commits/00b9836934d70fbf643e0f0b9283484fc5dbd873)

PR #1159, opened 2020-12-20
AGS3 SDL2: Debian fixes to produce AGS Linux Package

ericoporto:
- builds a debian package and produces the Linux packaging
- statically link SDL2_sound in Linux make builds since it doesn't have a package for now in debian
- fixes so CMake builds have the minimum dependencies as possible in Linux

The CMake changes are not relevant for the Linux Package we ship with the editor, but the changes significantly reduces the ldd output of ags (for CMake specific build) so I thought they would be useful to add.

With this PR we should be able to complete the Editor packaging task in the CI system but not the release, since for that we need to do the Android SDL2 port.


Engine/Makefile-defs.linux
```diff
@@ -38,7 +38,7 @@ SDL2_CFLAGS = $(shell sdl2-config --cflags)
 SDL2_LIBS = $(shell sdl2-config --libs)
 
 CFLAGS += $(SDL2_CFLAGS)
-LIBS += $(SDL2_LIBS) -lSDL2_sound
+LIBS += $(SDL2_LIBS) -Wl,-Bstatic -lSDL2_sound -Wl,-Bdynamic
```
rofl0r, line 41: i don't think this is necessary. if -lSDL2_sound is in the linker command it will pick up whatever library version (static or shared) that's available.

ci/linux/Dockerfile
```diff
@@ -51,20 +50,23 @@ RUN apt-get install \
...
+  libxxf86vm-dev \
+  libwayland-dev \
+  libxkbcommon-dev 
 
 # Build and install CMake
 ARG CMAKE_VERSION=3.14.5
 RUN curl -fLsS "https://github.com/Kitware/CMake/releases/download/v$CMAKE_VERSION/cmake-$CMAKE_VERSION.tar.gz" | tar -f - -xvzC /tmp && \
   cd /tmp/cmake-$CMAKE_VERSION && \
-  ./configure --system-curl && make && make install && \
+  ./configure --system-curl && make -j6 && make install && \
```
rofl0r, line 61: what's with the -j6 vs -j4 below? a user who wants -j options can export MAKEFLAGS=-j4 before running the script. note that compiling C++ code with debug info can lead to huge memory usage, so using e.g. -j6 by default may cause the build to fail if there are only like e.g. 4GB RAM available on the build environment.

ci/linux/Dockerfile
```diff
@@ -76,5 +78,5 @@ RUN cd /tmp && \
   cd /tmp/SDL_sound  && \
   mkdir /tmp/SDL_sound/build && \
   cd /tmp/SDL_sound/build && \
-  cmake ..  && make && make install && \
+  cmake -DSDL2_DIR=/usr/local/lib/cmake/SDL2 ..  && make -j4 && make install && \
```
rofl0r, line 81: why does the path to SDL2 need to be hardcoded here?

debian/ags+libraries/hooks/B00_copy_libs.sh
```diff
@@ -17,23 +17,19 @@ BIT=32
...
-    allegro/4.4.2/alleg-alsamidi.so \
-    allegro/4.4.2/modules.lst; do
+    libvorbisfile.so.3; do
       cp -L /usr/lib/$TRIPLET/$library $BINDMOUNT/data/lib$BIT
   done
-
+  
+  for library in \
+    libSDL2-2.0.so.0; do
+      cp -L /usr/local/lib/$library $BINDMOUNT/data/lib$BIT
```
rofl0r, line 29: why would you have SDL2 installed in /usr/local and everything else in /usr ? looks like a configuration error when compiling/installing SDL2.

debian/control
```diff
@@ -4,9 +4,8 @@ Priority: optional
 Maintainer: Tobias Hansen <tobias.han@gmx.de>
 Build-Depends: debhelper (>= 8.0.0),
                pkg-config,
-               liballegro4-dev,
-               libaldmb1-dev,
                libfreetype6-dev,
+               libsdl2-dev (>= 2.0.9),
```
rofl0r, line 8: is there a specific reason you request the >= 2.0.9 version of SDL2 ?

ci/linux/Dockerfile
ericoporto: This reduces the build time on our CI system, I tested. The Dockerfile is guided to our CI system.

ci/linux/Dockerfile
ericoporto: SDL2_DIR has to be specified if SDL2 is configured using CMake generated makefile instead of ./configure, before installation. The CMake build support all flags relative to reducing direct linking libraries in SDL2, so if I change this in the future things won't fail mysteriously. This also specifies which SDL2 I want to be linked with SDL2_sound in an obvious way, which is not the system one, it's the one I built.

debian/control
ericoporto: SDL2_sound won't work with lower versions of SDL2. This means it could work in systems with Debian Buster and beyond but won't work with system SDL2 in Debian Jessie. Feel free to test.

debian/ags+libraries/hooks/B00_copy_libs.sh
ericoporto: Other libraries are from Debian Jessie repository. SDL2 is built from source.

Engine/Makefile-defs.linux
ericoporto: We never should want the dynamically linked SDL2_sound since it has no package available for it, this is also how the rules have been written on Debian packaging. I can remove the dynamic library for it from the Docker library but removing here will cause people that try to build this makefile to mimic the AGS package having to remind and remove from the Makefile. Default behavior of the Makefile should statically link with SDL2_sound.

debian/ags+libraries/hooks/B00_copy_libs.sh
ericoporto: Also, no one should ship system SDL2 for others, SDL2 from package management comes configured to prefer linking directly instead of using dlopen. While this may work great in "my computer ™️" it's best to let SDL2 figure out what exists or not in the computer (alsa, pulseaudio, oss, framebuffer, ...).

ci/linux/Dockerfile
morganwillcock: As long as it is a real shell it should probably be `-j$(getconf _NPROCESSORS_ONLN)` if you want to go faster but allow people to run it manually without issues. That way the CPU count is limited to what Docker is allocated, which is under the control of the CI system or the user who is running it at home.

ivan-mogilko:
From what I understand, except for Makefile-defs.linux, these changes are only affecting prebuilt debian binaries?

ci/linux/Dockerfile
ericoporto: I can edit and push force as this tonight.

ericoporto:
We use the debian build process indirectly for making the Linux package that is shipped with the Editor!

I also added some updates on the CMake builds which are not particularly relevant for the releases we have right now but should be useful for anyone wanting to build and ship AGS from CMake builds on their own.

ci/linux/Dockerfile
ericoporto: For reference, our Docker image is built with 15GB RAM and 4 cores on Cirrus-CI.

Engine/Makefile-defs.linux
rofl0r: how about installing only the static library from the self-built sdl2_sound then ? that would remove the need for this hack, and allow other linux users to link with whatever library flavor they seem fit (some distros install only shared libs, in which case this would cause a build failure).

ericoporto:
With last two commits, once Cirrus CI builds, you can pick the editor from Editor Packaging task and successfully build a Linux game with the AGS3 SDL2 Engine. I will need to be away from the computer for some time, so if it's not possible to merge like this, feel free to cherry-pick or edit the code.

ivan-mogilko:
Regarding changes to linux makefile -

> how about installing only the static library from the self-built sdl2_sound then ? that would remove the need for this hack, and allow other linux users to link with whatever library flavor they see fit (some distros install only shared libs, in which case this would cause a build failure).

I think this is how it should be done, if it's doable. If that does not work perhaps make a switch in makefile to let users choose the way they want to link library?

ericoporto:
First, I should be back at home on the weekend and have a computer I can code!

@ivan-mogilko it's totally doable. I consider you to be the main user of the Linux Makefile, but you gave me two options. Which one you prefer? The suggested approach by rofl0r or a switch to be passed along with make?

ivan-mogilko:
> I consider you to be the main user of the Linux Makefile, but you gave me two options. Which one you prefer? The suggested approach by rofl0r or a switch to be passed along with make?

Erm, I am _definitely_ not the main user of linux makefile, there are many other people who might be using it, including @rofl0r for instance.

If I understand correctly, the process of making these prebuilt binaries for debian is creating a sort of custom enviroment, where you can install things unrelated to real system. In that sense the roflor's suggestion sounds better, so I'd put it on first place. The second method which I mentioned is an alternative for the worst case (when the first did not work for some reason).

ericoporto:
It should be just a drop of the commit that adds the Makefile change and an edit of the commit of the Dockerfile adding two switches to pass when the CMake file generates the Makefile for SDL2_Sound. Very doable.

rofl0r:
> It should be just a drop of the commit that adds the Makefile change and an edit of the commit of the Dockerfile adding two switches to pass when the CMake file generates the Makefile for SDL2_Sound. Very doable.

even simpler solution could be to just rm $destination/lib/libSDL2_sound.*so* after make install of SDL2_sound in the dockerfile.

ericoporto:
> even simpler solution could be to just rm $destination/lib/libSDL2_sound.*so* after make install of SDL2_sound in the dockerfile.

Went with this approach!

Removed the Makefile change and it appears to work!

PR #1160, opened 2020-12-27
Implement --tell-gameproperties instead of putting game info in --tell-config

### LINKED ISSUE #1154: --tell-config returns undocumented values for [misc]
The following values are not documented in https://github.com/adventuregamestudio/ags/blob/master/OPTIONS.md but are returned as part of the configuration data when using `--tell-config`:

```ini
[misc]
defaultres
game_height
game_width
gamecolordepth
letterbox
notruecolor
```

Presuambly the values which begin with 'game' are not actually configuration values?

Tested with 3.5.0.27

ivan-mogilko:
Resolves #1154

Moved "meta" settings from "--tell-config" to "--tell-gameproperties", improved value naming, added couple of more things.

Embedded winsetup receives a merged config map with "--tell-gameproperties" contents.
These were not really a true part of user config, thus supporting old names is not necessary.

All values printed by "--tell-gameproperties" go in "gameproperties" section, and are following:

* title - game title;
* guid - game guid;
* legacy_uniqueid - random generated 32-bit integer, that worked as UID in old games (may be ignored if guid is present);
* legacy_resolution - deprecated resolution type (integer), should be "8" for all new games;
* legacy_letterbox - bool, whether "letterbox by design" setting is on (considered obsolete);
* resolution_width - native resolution width;
* resolution_height - native resolution height;
* resolution_bpp - native resolution bits per pixel (8, 16, 32);
* render_at_screenres - have 3 values:
    -1 - determined by user config;
    0 (false) - locked at disabled
    1 (true) - locked at enabled

### DIFF
--- a/Engine/main/engine.cpp
+++ b/Engine/main/engine.cpp
@@ -1440,16 +1440,15 @@ static void engine_print_info(const std::set<String> &keys, ConfigTree *user_cfg
     {
         data["filepath"]["exe"] = appPath;
         data["filepath"]["cwd"] = Directory::GetCurrentDirectory();
-        data["filepath"]["startup"] = usetup.startup_dir;
-        data["filepath"]["datadir"] = ResPaths.DataDir;
+        data["filepath"]["datadir"] = Path::MakePathNoSlash(ResPaths.DataDir);
         if (!ResPaths.DataDir2.IsEmpty())
         {
-            data["filepath"]["datadir2"] = ResPaths.DataDir2;
-            data["filepath"]["audiodir2"] = ResPaths.AudioDir2;
-            data["filepath"]["voicedir2"] = ResPaths.VoiceDir2;
+            data["filepath"]["datadir2"] = Path::MakePathNoSlash(ResPaths.DataDir2);
+            data["filepath"]["audiodir2"] = Path::MakePathNoSlash(ResPaths.AudioDir2);
+            data["filepath"]["voicedir2"] = Path::MakePathNoSlash(ResPaths.VoiceDir2);
         }
-        data["filepath"]["savegamedir"] = GetGameUserDataDir().FullDir;
-        data["filepath"]["appdatadir"] = GetGameAppDataDir().FullDir;
+        data["filepath"]["savegamedir"] = Path::MakePathNoSlash(GetGameUserDataDir().FullDir);
+        data["filepath"]["appdatadir"] = Path::MakePathNoSlash(GetGameAppDataDir().FullDir);
     }
     String full;
     IniUtil::WriteToString(full, data);
--- a/Engine/main/engine.cpp
+++ b/Engine/main/engine.cpp
@@ -151,9 +151,26 @@ void engine_setup_window()
     platform->SetGameWindowIcon();
 }
 
+// Fills map with game settings, to e.g. let setup application(s)
+// display correct properties to the user
+static void fill_game_properties(StringOrderMap &map)
+{
+    map["title"] = game.gamename;
+    map["guid"] = game.guid;
+    map["legacy_uniqueid"] = StrUtil::IntToString(game.uniqueid);
+    map["legacy_resolution"] = StrUtil::IntToString(game.GetResolutionType());
+    map["legacy_letterbox"] = StrUtil::IntToString(game.options[OPT_LETTERBOX]);;
+    map["resolution_width"] = StrUtil::IntToString(game.GetDefaultRes().Width);
+    map["resolution_height"] = StrUtil::IntToString(game.GetDefaultRes().Height);
+    map["resolution_bpp"] = StrUtil::IntToString(game.GetColorDepth());
+    map["render_at_screenres"] = StrUtil::IntToString(
+        game.options[OPT_RENDERATSCREENRES] == kRenderAtScreenRes_UserDefined ? -1 :
+        (game.options[OPT_RENDERATSCREENRES] == kRenderAtScreenRes_Enabled ? 1 : 0));
+}
+
 // Starts up setup application, if capable.
 // Returns TRUE if should continue running the game, otherwise FALSE.
-bool engine_run_setup(ConfigTree &cfg, int &app_res)
+bool engine_run_setup(const ConfigTree &cfg, int &app_res)
 {
     app_res = EXIT_NORMAL;
 #if AGS_PLATFORM_OS_WINDOWS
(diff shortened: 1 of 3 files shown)


morganwillcock:
>     * render_at_screenres - have 3 values:
>       -1 - determined by user config;
>       0 (false) - locked at disabled
>       1 (true) - locked at enabled

Is this moved out of the 'graphics' section or is this from the 'disabled' section? If the latter, doesn't this split one of the 'disabled' settings away from the others? According to the options document there are many more, unless I am reading it incorrectly. From how I read it I have it noted that these values may or may not be present:

```ini
[disabled]
filters
hqx
linear
none
render_at_screenres
speechvox
stdscale
```

This is also the first time I've seen a reference to `-1` as a valid value for this setting. I tried adding adding this to a game config:
```ini
[disabled]
render_at_screenres=0
```
and then
```ini
[disabled]
render_at_screenres=1
```

Every time the `--tell` data reported the value as -1 (and changing the value in the user config made no difference either):
```
[gameoptions]
guid = {a1122974-442b-45a1-b016-7169486fced5}
legacy_letterbox = 0
legacy_resolution = 8
legacy_uniqueid = 893468
render_at_screenres = -1
resolution_bpp = 32
resolution_height = 200
resolution_width = 320
title = BASS
```

`--tell-gameoptions` doesn't seem to work (compare to above):
```
acwin.exe --console-attach --tell-gameoptions D:\AGS\BASS\Compiled\Windows

Adventure Game Studio v3.5 Interpreter
Copyright (c) 1999-2011 Chris Jones and 2011-2020 others
ACI version 3.5.0.27

Installing exception handler
Initializing allegro
[gameoptions]
guid =
legacy_letterbox = 0
legacy_resolution = -1
legacy_uniqueid = 0
render_at_screenres = -1
resolution_bpp = 0
resolution_height = 0
resolution_width = 0
title =
```

Also, maybe 'gameoptions' would be the wrong terminology. 'render_at_screenres seems' to be an option, but if everything else is metadata then maybe if should be called 'gamemeta' or 'gamemetadata'.

ivan-mogilko:
Fixed --tell-gameoptions not working, it was not marked as needed a game preload.

> Is this moved out of the 'graphics' section or is this from the 'disabled' section?

Neither, it's a separate option from the game (General Settings) which tells whether this setting should be locked to certain value or read from config.
The one in config 'graphics' section is a user setting that can be either true or false. The 'disabled' section is historically filled fully by hand in the config, there's no alternative to this. The 'disabled' section may or may not contain this option.

`This is also the first time I've seen a reference to -1 as a valid value for this setting. I tried adding adding this to a game config:`

This is not a value for game config, this is a value of the game option.

> Also, maybe 'gameoptions' would be the wrong terminology. 'render_at_screenres seems' to be an option, but if everything else is metadata

So far, these are all options, defined in General Settings in game project.

morganwillcock:
> This is not a value for game config, this is a value of the game option.

OK. So this is a value which hasn't appeared in the 'tell' output anywhere before?

> So far, these are all options, defined in General Settings in game project. I don't really know what "meta data" would mean in this context. But I can rename it to meta if it's more suitable.

OK, it makes more sense if this directly relates to those settings.

> Maybe also "gameproperties", as they are not player's options.

Yes, maybe that name is the best one.

I'll try to test it some more tomorrow.

ivan-mogilko:
> OK. So this is a value which hasn't appeared in the 'tell' output anywhere before?

It could appear in --tell-config as a combination of config options (normal "render_at_screenres" value and in "disabled" section), simply because engine was appending this to the contents of config file in order to pass to winsetup. Now it should not be doing that, so the --tell-config will display only settings read from files. At least I hope it does... (guess this is the desired behavior?)

morganwillcock:
Comparing output of tell with current release version:
```diff
@@ -1,7 +1,8 @@
+Logging to warnings.log
 [config-path]
 default = d:/AGS/BASS/Compiled/Windows/acsetup.cfg
-global = C:\Users\MWILLC~1\SAVEDG~1\.ags/acsetup.cfg
-user = C:\Users\MWILLC~1\SAVEDG~1/BASS/acsetup.cfg
+global = C:/Users/MWILLC~1/SAVEDG~1/.ags/acsetup.cfg
+user = C:/Users/MWILLC~1/SAVEDG~1/BASS/acsetup.cfg
 [config@graphics]
 driver = D3D9
 filter = StdScale
@@ -19,11 +20,9 @@
 [config@misc]
 antialias = 0
 cachemax = 131072
-defaultres = 8
 game_height = 200
 game_width = 320
 gamecolordepth = 32
-letterbox = 0
 notruecolor = 0
 shared_data_dir =
 titletext = BASS Setupzzz
@@ -44,6 +43,23 @@
 [engine]
 name = Adventure Game Studio run-time engine
 version = 3.5.0.27
+[filepath]
+appdatadir = C:/PROGRA~3/Adventure Game Studio/BASS
+cwd = c:/Users/MWILLC~1/DOWNLO~1
+datadir = d:/AGS/BASS/Compiled/Windows/
+exe = c:/Users/mwillcock/Downloads/acwin.exe
+savegamedir = C:/Users/MWILLC~1/SAVEDG~1/BASS
+startup = d:/AGS/BASS/Compiled/Windows
+[gameproperties]
+guid = {a1122974-442b-45a1-b016-7169486fced5}
+legacy_letterbox = 0
+legacy_resolution = 8
+legacy_uniqueid = 893468
+render_at_screenres = -1
+resolution_bpp = 32
+resolution_height = 200
+resolution_width = 320
+title = BASS
 [graphicdriver]
 0 = D3D9
 1 = OGL
```
There is an immediate problem that now the output is including "Logging to warnings.log", but everything else looks OK and the slash direction on the config paths now all go the same direction, so I guess you've fixed that too?

ivan-mogilko:
> There is an immediate problem that now the output is including "Logging to warnings.log"

No idea how it got there, I will find out.

> and the slash direction on the config paths now all go the same direction, so I guess you've fixed that too?

That was probably fixed by other commits in the master, I was doing some changes to how paths are dealt with in the engine recently.

ivan-mogilko:
I pushed a fix to master (4a0c503a00476d4477a1e8d8c851fd6c1373a09b), because it's not related to this PR, now it should not print random line to stdout when --tell command is processed.

morganwillcock:
Nothing in the data looks wrong, other than there is an inconsistency in how directories are indicated, e.g. 'datadir' ends with a directory separator (/) but other paths which are directories don't have it.

And I guess I can say that the key names relating to directories are not consistent in when they represent a directory and the key name ends in 'dir' (e.g. 'startup' is for a directory but isn't called 'startupdir'), but perhaps it is too late to change those if they have already been in use.

ivan-mogilko:
[filepath] was added to master branch few days ago 3c5e29431842c77993c29a128dbce3e48c987679.
"startup" is probably a typo. I can fix this and anything else there.

ivan-mogilko:
Added a small fix, also removed "startup" key, because it's not an "official" concept at the moment, just something I store in the engine for the reference.

morganwillcock:
I've tested it again and it looks fine now, assuming that the "Logging to warnings.log" issue is fixed separately.

PR #1161, opened 2021-01-01
Editor: replace DeserializeIgnoreAttribute with [Obsolete] properties

ivan-mogilko:
I was not complete sure about this so made a pr to hear opinions.

In the past I created this DeserializeIgnoreAttribute for making lists of deprecated class properties, but at that time I did not know whole deserialization procedure (or attributes) very well, so took this what seemed most straightforward approach.

But I never liked how it turned out. The use of this attribute seem bad for two main reasons:
* you got to apply it to the class's declaration above its name, sometimes multiple instances (one per property) which makes it look weird;
* it's not very obvious and may be therefore inconvenient to maintain.

In this pr I remove DeserializeIgnoreAttribute completely, and instead return previously deleted properties, tagged with [Obsolete] and [Browsable(false)] atributes, and a formal get method.
One one hand, this makes them more visible and one may arrange them in class as necessary (add comments, etc). On another hand, this reserves an opportunity to handle their loading and conversion if such need suddenly arises; this also makes their handling more uniform with other obsolete properties that already do such conversion.

According to deserialization, those properties which are readonly won't be loaded anyway and skipped without error.
I added a rule to serialization to skip [Obsolete] properties (similar to [AGSNoSerialize]), so they are never written again.


persn (approved):
This looks good I think, the attribute doesn't seem to have been used much anyway, and people probably wouldn't use it if they needed it because they didn't know it existed.

There's probably ways to make the usage of the attribute less clumsy, but no reason to think about it if we're going to purge it.

PR #1162, opened 2021-01-05
Editor: Removed the copy walkable to region button

persn:
Removed it from the hotbar and all related functionality.

This was oddly specific, it's unlikely that it's used often enough that it warrants a dedicated button. Users who wants to replicate this feature without button can export and import manually.

Related to #1138

### DIFF
--- a/Editor/AGS.Editor/NativeProxy.cs
+++ b/Editor/AGS.Editor/NativeProxy.cs
@@ -317,11 +317,6 @@ namespace AGS.Editor
             _native.DrawFillOntoMask(room, mask, x1, y1, color);
         }
 
-        public void CopyWalkableAreaMaskToRegions(Room room)
-        {
-            _native.CopyWalkableMaskToRegions(room);
-        }
-
 		public bool GreyOutNonSelectedMasks
 		{
 			set { _native.SetGreyedOutMasksEnabled(value); }
--- a/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
@@ -15,7 +15,6 @@ namespace AGS.Editor
         protected const string DRAW_FREEHAND_COMMAND = "DrawFreehand";
 		protected const string DRAW_RECTANGLE_COMMAND = "DrawRectangle";
 		protected const string DRAW_FILL_COMMAND = "DrawFill";
-        protected const string COPY_WALKABLE_AREA_MASK_COMMAND = "CopyWalkableMaskToRegions";
         protected const string IMPORT_MASK_COMMAND = "ImportAreaMask";
         protected const string EXPORT_MASK_COMMAND = "ExportAreaMask";
         protected const string UNDO_COMMAND = "UndoAreaDraw";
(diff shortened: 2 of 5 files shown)


morganwillcock:
> export and import manually

This isn't a direct equivalent because you need to use a paint program in the middle, which might not be too bad, but the mask format is specific enough that I'm not sure someone trying it will be successful if they go looking to find their own paint program.

If you wanted to apply a lighting effect to where you walk then you would be looking to press this button. To be honest I'm pretty hazy on what the master branch is for now though, so I've no idea where this is targeted to end up.

ivan-mogilko:
> This isn't a direct equivalent because you need to use a paint program in the middle, which might not be too bad, but the mask format is specific enough that I'm not sure someone trying it will be successful if they go looking to find their own paint program.

This is supposed to be an equivalent in theory. What would be a use of paint program if you only do export area and import immediately after? If import/export does not work correctly as it is then it should be fixed as its own issue.

Alternatively, we could implement copy/paste command for generic copying any mask to any mask through the memory, which ofcourse will be easier for user.

> If you wanted to apply a lighting effect to where you walk then you would be looking to press this button.

TBH I haven't thought about this use, initially I thought this is some outdated command for old games, therefore suggested to remove it.

Knowing this example, it might be better to replace with something like copy&past mask command instead of removing this completely.

persn:
> This isn't a direct equivalent because you need to use a paint program in the middle

Which piece of essential information am I missing here? What is the purpose of involving a paint program for a copy-paste functionality?

morganwillcock:
I thought the button exported the current walkable area but apparently it just overwrites all regions with all walkable areas, so you can ignore my objection.

ivan-mogilko:
> I thought the button exported the current walkable area but apparently it just overwrites all regions with all walkable areas, so you can ignore my objection.

Well, this is how I understood it, and it still makes sense imo.

ivan-mogilko:
Speaking of which, maybe there's sense to have two pairs of commands: copy/paste all mask and copy/paste whole area.

Or is it better to have a separate ticket for this?

morganwillcock:
If this is was ags4 I wouldn't have mentioned anything, because I also think the button needs to go to be replaced by something else (maybe masks shouldn't even be bitmaps at this stage). But if this is for something that will turn into a release then workflow on this particular feature is now worse than before, even though I agree that the feature is bizarrely specific and would force you to draw your masks in a certain order to be used. Copy and paste of all or some layers would solve the problem, but I think for the release branch this has to be a block for release rather than just another feature for the wish list.

Also, for people running AGS on WINE the mask export doesn't work properly. I think it is an issue of format selection within the native image encoder used and not an AGS problem, but if the mask copy button avoids the encoder then it probably works around the issue.

persn:
> Or is it better to have a separate ticket for this?

Yes. This is new a new feature. Might also be a good idea to start a poll or something on the forums to check if people actually want this.

ivan-mogilko:
@persn alright, sorry then, perhaps only remove this in ags4 as a part of #1138 and open ticket for ags4 to make better copy commands?

ivan-mogilko:
> If this is was ags4 I wouldn't have mentioned anything, because I also think the button needs to go to be replaced by something else (maybe masks shouldn't even be bitmaps at this stage). But if this is for something that will turn into a release then workflow on this particular feature is now worse than before

Alright, this was a wrong suggestion for now.

@persn I guess then it's best to keep it unchanged in your pr (#1138), and consider different controls later. But idk if it's high priority (depends on how many people actually need copy/past masks). Not that is difficult to implement though.

Issue #1148, opened 2020-12-09
Fixes for AGS3 SDL2 port


ericoporto:
Issue to aggregate remaining fixes for AGS3 SDL2 port (now merged into master).

### All Ports

- Sound config: device(?)  (moved to #1565)
- Threaded audio option is now working differently (disable? or keep it still because Emscripten port requires one) (moved to #1566)
- "Software" SDL Renderer output hinting config (moved to #1567)

---

### Windows

- Update winsetup (sound config, perhaps something else?) (moved to #1565)

---

### Linux

- [ ] Restore setting icon.xpm, see #1410 explaining the issues met.

---

### MacOS

- [ ]  Not exactly a task, but the port needs more testing. 
- [ ]  Update build instructions. (cmake generates xcode project or cmake builds directly, mention SDL2 framework)

---

### Android

- [ ]  Touch to mouse translation auto translated from SDL2 is not similar to old Android backend behavior. On a plus it allows drag and drop module to work, finger hold on screen is readable by mouse.IsButtonDown(): https://youtu.be/lbdth4SHbSE.

---

### iOS

iOS support is a bit tricky (for me), SDL2 supports but I lack the device to test it, so for now I think the main task is finding someone who is interested in picking it up. If no one picks it up, I can do it on an emulator.

https://hg.libsdl.org/SDL/file/bc96d65e47da/docs/README-cmake.md
https://hg.libsdl.org/SDL/file/bc96d65e47da/docs/README-ios.md

mapped on #1505

---
## Completed Tasks

<details><summary>Click to Expand - Finished Tasks hidden for readability</summary>
<p>

~~I marked with one asterisk `*` what is reasonably done but hasn't been merged yet, and with two asterisk `**` what the fix is made but isn't completely working yet and hasn't been merged.~~ Probably not relevant anymore, but kept for history.

### All Ports

- [x] missing proper sound cancellation, [see comment]()https://github.com/adventuregamestudio/ags/issues/1148#issuecomment-761581778
- [x] Sound subsystem and thread has to be explicitly stopped on exit before shutting down backend and deleting data, otherwise there may be errors.
- [x] CI/CMake update: don't use DirectX SDK anymore. (#1291)
- [x] CI/CMake update: don't require DirectX SDK and SDL2/SDL_sound for editor. (#1291)
- [x]  Remove `hqx` filter mention from command line arguments (#1293)
[...]

Issue #1096, opened 2020-06-24
a path towards moving from Allegro to SDL


ericoporto:
[**`ags/wiki/SDL-Migration-proposition.md`**](https://github.com/adventuregamestudio/ags/wiki/SDL-Migration-proposition)

The document is in the wiki so people can write in and clarify! If you are worried about losing your edit, do it using git, the document is `SDL-Migration-proposition.md` on the root directory.
- https: `git clone https://github.com/adventuregamestudio/ags.wiki.git`
- ssh: `git clone git@github.com:adventuregamestudio/ags.wiki.git`

---

I wrote above a [document](https://github.com/adventuregamestudio/ags/wiki/SDL-Migration-proposition) proposing a path toward migrating from Allegro, meaning, removing Allegro dependency completely, and using instead mostly SDL. It may turn out we need something else too.

Since last ags3-sdl2 port did not get merged, this tries to detail a plan, with **ags4** in mind and focus on complete removal of Allegro.

Please comment ~~, also accepting people to write the code~~. My central idea was to have code flowing towards this goal into AGS repo.

ivan-mogilko:
I may be overthinking, but I don't like this sentence:
>  also accepting people to write the code. My central idea was to have code flowing towards this goal into AGS repo.

This is not the kind of task to be made by any multiple contributors, there must be a general plan and focused developer(s) to overwatch the process. It's better to have 1-2 people working on this to keep things organized.

ericoporto:
I think a minimum of two people is required. I expressed badly, but writing code requires, other than the code: fix vcxproj files, fix makefiles, fix cmake files, testing and fixing initially in two (Windows and Linux?) , than three (MacOS), than 4 (Android), depending how libraries are inserted building or including these libraries (which means adding in the three dockerfiles), ... Even if is one for doing everything and the other to help testing. 

I think it's important to decide if we are building statically or dynamically (should game.exe accompany a SDL.dll ?) and how SDL gets in AGS project (code in repo or reference). It's important to note SDL may also go into editor due to Native.dll.

JanetGilbert:
[...]

PR #1165, opened 2021-01-11
SDL2 Port: building instructions

ivan-mogilko:
* Updated Windows readme
* Updated Debian (linux) readme.
* TODO: MacOS


ivan-mogilko:
@ericoporto could you check please?

ericoporto:
Since you are mentioning props environment variables for SDL (like `AGS_SDL_INCLUDE`), maybe also adding the ones for XIPH would be useful.

On SDL_Sound, maybe mentioning that AGS Windows Engine is 32-bit/x86, so it needs to match. Same for SDL2.dll, it's the 32 bit version - I remember once dualnames made a plug-in available in the forums and a lot of people got bugs on their own because they didn't knew they had to download the x86 version and not the x64 one.

Also not sure it's worth mentioning, but for the Engine, the CMake project builds for Windows without having to download or configure anything, at least I think it fetches everything that is needed.

ivan-mogilko:
> Since you are mentioning props environment variables for SDL (like AGS_SDL_INCLUDE), maybe also adding the ones for XIPH would be useful.

So, this made me revisit list of linked libraries again, and I realized that:
- we link xiph libs (ogg, vorbis, theora) from outside, but include headers from within our repository (Common\libinclude). I completely forgot about that. Maybe this was done for compliance with old allegro-based libraries. Anyway, I did not touch these headers now. But this means you need to only configure library locations, AGS_LIBOGG_LIB etc.
- vorbisfile is not necessary to link (anymore?).
- because engine is no longer using other DirectX libs than d3d9, there's probably no need to have a manual DirectX SDK at all, as d3d9 library may be taken from Windows SDK. Hence I removed directx.props from MSVS projects.
- 
@morganwillcock could you please double check the last point, because I may miss something too?

ericoporto:
> vorbisfile is not necessary to link (anymore?).

I don't think it's too, but I wasn't sure, this is why I kept it when building the debian package on Linux.

I am avoiding changing too much until AGS3 and SDL2 ports are merged since most of what I have in mind (and coded) for it is actually ags4 material (things that modify script API or game building in the Editor). But I understand this requires getting the mobile ports working, which is still work in progress.

ivan-mogilko:
I amended instructions for debian, as much as I could figure that out. Ofcourse these are mostly limited to systems like debian and ubuntu, but list of dependencies at least should be similar everywhere (I hope?).

For SDL_Sound I used the hint by rofl0r he posted here: https://github.com/adventuregamestudio/ags/issues/1096#issuecomment-731591824

debian/README.md
```diff
@@ -8,23 +8,32 @@ The following packages are required to build AGS. The versions in
...
+SDL_Sound library installation
+----------------------------------
+At the time of writing SDL_Sound 2.* did not have a proper release and almost no linux distro provides it.
+Until that is resolved, we recommend to download particular revision archive using following url:
+
+    https://hg.icculus.org/icculus/SDL_sound/archive/9262f9205898.tar.bz2
+
+Before building the library, perform following command on its source directory:
+
+    sed 's|FILES SDL_sound.h DESTINATION include|FILES src/SDL_sound.h DESTINATION include/SDL2|' -i CMakeLists.txt
```
morganwillcock, line 34: To be compatible with more sed implementations, this should probably be written as:
```
sed -i.bak 's|FILES SDL_sound.h DESTINATION include|FILES src/SDL_sound.h DESTINATION include/SDL2|' CMakeLists.txt
```
...assuming that the presence of the backup file doesn't cause a problem.

ericoporto: This sed is unneeded in the latest code from mercurial.

ericoporto: See : https://hg.icculus.org/icculus/SDL_sound/archive/997e90562b35.tar.gz

ivan-mogilko: Hmm interesting, this might be the relevant part of Makefiles.txt:

<pre>
install(TARGETS ${SDLSOUND_INSTALL_TARGETS}
        RUNTIME DESTINATION bin
        LIBRARY DESTINATION lib${LIB_SUFFIX}
        ARCHIVE DESTINATION lib${LIB_SUFFIX})
install(FILES src/SDL_sound.h DESTINATION include/SDL2)
</pre>

So it does install into SDL2 dir now.

ericoporto:
On Debian derivatives, building with CMake should be:

Install sdl2 build dependencies:
```
sudo apt-get build-dep libsdl2-2.0-0
```

then do CMake build:

```
cd ags
mkdir build-debug
cd build-debug
cmake .. -DCMAKE_BUILD_TYPE=Debug
cmake --build . --parallel
```

Every other dependencies (including Xiph) are figured out at CMake configure time and fetched as needed.

ivan-mogilko:
If using CMake, which other steps from instructions may be omited? E.g. are these still necessary or unnecessary with cmake:

> Build the package and install it with gdebi:

    fakeroot debian/rules binary
    sudo gdebi ../ags_3~git-1_*.deb

ivan-mogilko:
Which parts of the instructions does Cmake method cover? Does it only relates to regular build, or it can create Debian/Ubuntu package (deb) too?

ericoporto:
The CMake is building a different AGS where all dependencies are either statically built or figured at runtime with dlopen, this is a SDL2 feature that works well with Steam, so it's not built towards packaging as a deb. It's alright ommiting and I can just correct the CMAKE.md later. The only steps should be the one I mentioned in my previous comment though, it will produce a binary standalone ags executable as a result. 

Release build can be done by passing `-DCMAKE_BUILD_TYPE=Release` instead of the suggested Debug on. The Release build uses LTO and takes a bit more time to complete, so for development it's better to work with the Debug one.

ivan-mogilko:
Alright, I will merge current instructions then, unless anyone sees any errors?

morganwillcock:
> Alright, I will merge current instructions then, unless anyone sees any errors?

The props variables are currently referenced in the CI build:
https://github.com/adventuregamestudio/ags/blob/ags3--sdl2/.cirrus.yml#L9-L16
https://github.com/adventuregamestudio/ags/blob/ags3--sdl2/.cirrus.yml#L165-L169

As well as removing the unused ones from the CI build, the actual pre-build steps which setup those directories should also be removed, although it probably doesn't break anything to have them there so it can be a follow-up task if the CI builds of this PR are actually working.

ivan-mogilko:
I added a note on DirectX in the checklist here: #1148 
BTW, SDL2 and _sound also are not needed for compiling editor itself, so these dependencies may be removed too.

debian/README.md
rofl0r: > For SDL_Sound I used the hint by rofl0r he posted here: #1096 (comment)

yep, this was fortunately fixed in exactly the manner i proposed, but there are other issues stil lingering: https://github.com/sezero/SDL_sound/commit/e2e479f4a9deaeae02c8ac3039860b3348adb48f#commitcomment-45911235 but it seems @sezero is working on it. :+1:

ericoporto: @rofl0r , the version wrong when installing the default branch is because the version on the CMakeLists.txt is wrong (look for the word `SOVERSION` in it and you will see it's being set to 1 when it should be a 2, and the `VERSION` is being set to 1.9.0 instead of 2.0.0). Probably this will be updated once it's released only.

PR #1166, opened 2021-01-11
Ubuntu20fix

rofl0r:
fix compilation on ubuntu 20.04, some more sensible makefile cleanups

### DIFF
--- a/Engine/Makefile-defs.linux
+++ b/Engine/Makefile-defs.linux
@@ -19,7 +19,6 @@ CFLAGS := -O2 -g \
 
 CXXFLAGS := -std=c++11 -Werror=delete-non-virtual-dtor $(CXXFLAGS)
 
-LIBS := -Wl,-Bdynamic
 LIBS += $(FT_LDFLAGS)
 LIBS += $(shell pkg-config --libs allegro)
 LIBS += $(shell pkg-config --libs x11)
--- a/Engine/Makefile-defs.linux
+++ b/Engine/Makefile-defs.linux
@@ -33,12 +33,13 @@ else
   LIBS += $(shell pkg-config --libs vorbis)
 endif
 LIBS += $(shell pkg-config --libs vorbisfile)
-LIBS += -ldl -lpthread -lm
 
 ifneq ($(USE_BUILT_IN_LIBSRC), 1)
 ALDUMB :=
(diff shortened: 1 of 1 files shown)


rofl0r:
i'm not too familiar with the linux binary packaging process, but in case the build host used for it (cirrus ci?) has libdumb / aldumb installed as both dynamic and static libraries, the .so files copied into the linux archive will lack those for dumb currently. in that case the best fix  would be (analoguous to the SDL2 ci fix PR ##1159 by @ericoporto) to rm the dynamic library version before the build, so the static libs get picked up and the bundles are identical to how they were.

ericoporto:
Deleting a file downloaded from the package manager should not be a thing, removing something installed from source is different. This will make it not possible to make a Linux package from a local machine - or risk breaking other software. This is a wrong approach, it's best to do a command line switch for dynamically if you must. If you are not interested in this then you need to ship the dynamic library with the AGS Linux package and ensure proper use of rpath.

ericoporto:
This PR also fixes a very weird error in debian packaging.

The release of Linux for Editor in the CI is the Linux Packing, it should work normally.

So, for years we have shipped aldumb and libdumb twice?

See here: https://github.com/adventuregamestudio/ags/blob/master/debian/ags%2Blibraries/hooks/B00_copy_libs.sh

ivan-mogilko:
> So, for years we have shipped aldumb and libdumb twice?

Could've been someones misunderstanding, as there were more differences between platforms in the past.
But yes, dumb seem to has been included in Makefile-objs for linux: https://github.com/adventuregamestudio/ags/blob/master/Engine/Makefile-objs#L74

ericoporto:
> Could've been someones misunderstanding, as there were more differences between platforms in the past

Yeahp, as far as I can tell it runs alright in a clean machine too,  and ldd appears to be ok. If someone else could check it out just to verify I am not seeing things, but I do believe it's correct and it was some past misunderstanding.

ivan-mogilko:
@legluondunet would you be able to test this using rofl0r's branch, and tell if it works for you?

legluondunet:
> @legluondunet would you be able to test this using rofl0r's branch, and tell if it works for you?

Hello, AGS compilation success here on Ubuntu 20.04.
Very nice work!
You could consult full compilation log here:
https://paste.ubuntu.com/p/4gkFBTQw9k/

PR #1173, opened 2021-01-14
New compiler import autoptr bugs

fernewelten:
Address the following bugs:
- In `import` statements, `autoptr` entities aren't processed correctly, their pointer isn't implied but should be.
- All `autoptr` types are displayed with a spurious trailing `" "` in messages.

In particular, `String` is affected.

Fix the code so that the bugs go away.


ivan-mogilko:
@fernewelten , could you please tell which kind of bug this addresses? Is it something like `String arr[10];` not working?

ivan-mogilko:
Following kind of script was fixed:
<pre>import String s;

function test() {
    s = "text";
}
</pre>

PR #1182, opened 2021-01-19
When user interface is disabled, only grey/black out GUI buttons that have an associated click action.

fernewelten:
Buttons that don't have any click action are, in effect, just labels. The user can't react to them in the first place. So when the interface becomes disabled, the user should not get the message that they _now_ cannot react to them. 

This PR changes the UI so that GUI buttons whose ClickAction is NoAction are treated as labels.

(Why would an AGS coder use a button without any click action instead of a label in the first place? Because GUI labels can't show sprites. So if you want to denote with a spiffy sprite, e.g., whether Ego can currently fly or not, then a label won't do and you need a button.)

### DIFF
--- a/Common/gui/guibutton.cpp
+++ b/Common/gui/guibutton.cpp
@@ -89,26 +89,29 @@ bool GUIButton::IsClippingImage() const
 
 void GUIButton::Draw(Bitmap *ds)
 {
-    bool draw_disabled = !IsGUIEnabled(this);
+    // A non-clickable button is, in effect, just a label.
+    // When the GUI is disabled, the user should not get the message that
+    // the button is now unclickable since it had never been in the first place.
+    bool const button_is_clickable =
+         ClickAction[kMouseLeft] != kGUIAction_None ||
+        ClickAction[kMouseRight] != kGUIAction_None;
+
+    bool const draw_disabled =
+        !IsGUIEnabled(this) &&
+        button_is_clickable &&
+        gui_disabled_style != GUIDIS_UNCHANGED &&
+        gui_disabled_style != GUIDIS_GUIOFF;
 
     check_font(&Font);
-    // if it's "Unchanged when disabled" or "GUI Off", don't grey out
-    if (gui_disabled_style == GUIDIS_UNCHANGED ||
-        gui_disabled_style == GUIDIS_GUIOFF)
-    {
-        draw_disabled = false;
-    }
+
     // TODO: should only change properties in reaction to particular events
     if (CurrentImage <= 0 || draw_disabled)
         CurrentImage = Image;
 
-    if (draw_disabled && gui_disabled_style == GUIDIS_BLACKOUT)
-        // buttons off when disabled - no point carrying on
-        return;
-
-    // CHECKME: why testing both CurrentImage and Image?
-    if (CurrentImage > 0 && Image > 0)
+    // No need to check Image after the assignment directly above
+    if (CurrentImage > 0)
         DrawImageButton(ds, draw_disabled);
+
     // CHECKME: why don't draw frame if no Text? this will make button completely invisible!
     else if (!_text.IsEmpty())
         DrawTextButton(ds, draw_disabled);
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Do buttons with no click actions still display "pushed" graphic when clicked?

ivan-mogilko:
> Do buttons with no click actions still display "pushed" graphic when clicked?

So, I tested this, and the putton still may be pushed even if action = No Action.

My point being, if you want to treat a button with no action as a "graphical label", this should perhaps be consistent, and it should not be pushable as well.

Regarding the code itself, a little nitpick, but could you call variable `button_is_clickable` something else, because buttons already have "Clickable" property with different meaning.

rofl0r:
if a label with graphics is desired, why not simply add a graphical_label gui control ?

ivan-mogilko:
> if a label with graphics is desired, why not simply add a graphical_label gui control ?

I would suggest that is better in the long term, except much more things to code in both engine and editor.

rofl0r:
if i was to implement that, i'd just change label to support an additional graphic and add a property graphical=no (or graphics=null) to the traditional label variant, whereas the graphical variant would have set it to yes.

ivan-mogilko:
> if i was to implement that, i'd just change label to support an additional graphic and add a property graphical=no (or graphics=null) to the traditional label variant, whereas the graphical variant would have set it to yes.

That is also an option.

Anyhow, I think the purpose of this pr is to correct already existing behavior rather than introduce a new one. Frankly I think it is even suitable for ags3, because displaying disabled state on a button with no action is wrong.

fernewelten:
So what do you want me to implement?
- [ ] Only treat those buttons as labels that don't have any associated `PushedImage`.
- [ ] When buttons have `ClickAction` set to `NoAction` then any associated `PushedImage` is ignored. Those inoperable buttons only show their `Image`. If the user clicks onto them, they still don't show their pushed graphic. 

How do we treat `MouseoverImage`? Likewise?

ivan-mogilko:
Actually, I am not sure about whole thing at all. What if user made a no action button for a purpose to work as a pushable button that does nothing? I don't know why, for some obscure reason. Such scenario is still valid, as button is intended to work as button.

fernewelten:
@rofl0r 

I can well imagine touching up labels, especially in ags4. But let's keep that separate and do this one first. 

We might envision having a sound that is automatically played whenever a button is clicked, in the same way that view graphics can have a sound that is played whenever the graphic is displayed. But again, that's an idea that should be further discussed and then put into another PR.

fernewelten:
> Actually, I am not sure about whole thing at all anymore. What if user made a no action button for a purpose to work as a pushable button that does nothing?

Then they could set `ClickAction` to `RunAction` and associate an empty event function.

ivan-mogilko:
Actually, I am not sure about whole thing at all anymore. What if user made a no action button for a purpose to work as a pushable button that does nothing? I don't know why, for some obscure reason. Such scenario is still valid, as button is intended to work as button.

ivan-mogilko:
> > What if user made a no action button for a purpose to work as a pushable button that does nothing?
> 
> Then they could set `ClickAction` to `RunAction` and associate an empty event function.

But this is a very obscure rule.

ivan-mogilko:
> I can well imagine touching up labels, especially for ags4. But let's keep that separate and do this one first.

The question is then, why do this change to a button at all in ags4? Buttons were used to display sprites mostly because there was nothing else. So, if we are doing something in context of a next big version, would not it be a better choice to add that missing control type instead?

ivan-mogilko:
Okay, sorry, I need to think a little more about this.
Maybe "No action" was originally intended for buttons that just display a graphic, and this was made for simplicity sake (to reuse same control).

ivan-mogilko:
@fernewelten 
Alright, sorry for all the multiple comments.

I am coming to a thought that regardless of how we change the button now it will still be not fully right, because this object never had fully consistent behavior for its two main uses, and relied on careful setup by user.
Your main suggestion here is to fix one of these uses, which makes sense at least until we further redesign gui controls.

Therefore, I remove my proposal about disabling push state for such button, let user handle this the way they want. After all, if they dont want it to have any action *and* dont want button to be pushed, they may as well set Clickable false, and problem is solved.

So, if you just fix code a little according to my first comment, I will accept this PR.

PR #1183, opened 2021-01-19
AGS 4: extended meta format for main game data

### LINKED ISSUE #828: future backwards compatible modifications to game file layout
while working on [agsutils](https://github.com/rofl0r/agsutils), i constantly run into errors with newer games due to modifications to the game file headers. hunting these down requires a lot of effort, single-stepping through ags engine and my code at the same time.

e.g.:
https://github.com/rofl0r/agsutils/commit/5f042f82ab511f2a893c8415c58f884848796d1c
https://github.com/rofl0r/agsutils/commit/6d679e24178bc6e47e1520c260a19d5a39984cd0

i would suggest a method to keep gamefile modifications backwards compatible.
current gamefiles have an array of 17 reserved ints:

```cpp
    in->ReadArrayOfInt32(reserved, NUM_INTS_RESERVED);
```

for future engine versions, we could use int1 of the struct to signify total bytes of extensions added (`int total_extension_bytes`), and the extensions itself would be placed at the end of the existing header, i.e. straight after views are read.
the next int in the reserved block (`extension_count`) could signify how many extensions are added.
each extension itself could then start with an int `extension_length` signifying the total length of this single extension, followed by a string of a hardcoded length of e.g. 16 containing the name of the extension. for example "ExtSpriteInfo".
the engine can then determine whether it has code for handling "ExtSpriteInfo" extensions, and if not just skip the extension_length bytes to the next extension and so on. the data in the block after the extension name could be entirely extension-specific that way.

a tool like agscriptxtract would simply read `total_extension_bytes` from reserved[0] and skip the bytes after parsing the rest of the header.

ivan-mogilko:
Resolves #828 (hopefully).

The idea is to enforce the rule that any new change or addition to the main game data should be appended as an extension to the end of the current format. So far we were just adding these changes in the midst of existing format randomly, which complicates format parsing for external tools, and imho just generally a bad thing for long term maintenance.

A support is added for reading and writing a list of "blocks" appended after the "classic" format. _The blocks meta format is shared with the one from room files_, trying to keep things consistent if possible. The only addition is the replacement of 1-byte numeric ID with a 16-chars long string ID. I am thinking of doing same ID change for room format, again for consistency and convenience.

The proposal is to keep this way of adding new data while we keep current "raw" binary format and until we implement something better (like compressed text format, or one of these contemporary serialization libs -- just giving random examples here without any implication).

---

Extension format description.

Each block starts with the following header:
- 1 byte - an old-style unsigned numeric ID:
    where 0 would indicate following string ID,
    and 0xFF indicates end of extension list.
- 16 bytes - string ID of an extension. An empty string (with \0 in the first byte) indicates end of extensions.
- 8 bytes - length of extension data, in bytes.

Then goes regular data.
Hence the total blocks size is (1 + 16 + 8 + data length) bytes.

After the block is read the stream position supposed to be at (start + 25 + length of data), where "start" is a first byte of the block header. If it's further - the function returns with error. If it's not far enough - the function logs a warning and skips remaining bytes.

Extensions are read until meeting empty ID string, which indicates end of extension list.
If end of file is found earlier that is considered an error.

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -22,6 +22,7 @@
 #include "ac/dynobj/scriptaudioclip.h"
 #include "core/asset.h"
 #include "core/assetmanager.h"
+#include "debug/out.h"
 #include "game/main_game_file.h"
 #include "gui/guimain.h"
 #include "script/cc_error.h"
@@ -85,6 +86,12 @@ String GetMainGameFileErrorText(MainGameFileErrorType err)
         return "Format version of plugin data is not supported.";
     case kMGFErr_PluginDataSizeTooLarge:
         return "Plugin data size is too large.";
+    case kMGFErr_ExtUnexpectedEOF:
+        return "Unexpected end of file.";
+    case kMGFErr_ExtUnknown:
+        return "Unknown extension.";
+    case kMGFErr_ExtBlockDataOverlapping:
+        return "Block data overlapping.";
     }
     return "Unknown error.";
 }
@@ -483,10 +490,18 @@ HGameFileError ReadSpriteFlags(LoadedGameEntities &ents, Stream *in, GameDataVer
     return HGameFileError::None();
 }
 
+static HGameFileError ReadExtBlock(LoadedGameEntities &ents, Stream *in, const String &ext_id, soff_t block_len, GameDataVersion data_ver)
+{
+    return HGameFileError::None();
+}
+
 HGameFileError ReadGameData(LoadedGameEntities &ents, Stream *in, GameDataVersion data_ver)
 {
     GameSetupStruct &game = ents.Game;
 
+    //-------------------------------------------------------------------------
+    // The classic data section.
+    //-------------------------------------------------------------------------
     {
         AlignedStream align_s(in, Common::kAligned_Read);
         game.GameSetupStructBase::ReadFromFile(&align_s);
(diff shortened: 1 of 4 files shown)


rofl0r:
> 1 byte - unsigned numeric block id, this is mostly for compat with legacy block format, where 0xFF indicates end of extensions list, and other values may be either used or ignored.

if this one byte is currently only used to signify "end of extensions", then maybe we could just as well use the string id with first byte set to '\0' instead.

ericoporto:
Hey, after this is merged, it would be cool to add an entry in the wiki about how the meta format works: https://github.com/adventuregamestudio/ags/wiki

ivan-mogilko:
> > 1 byte - unsigned numeric block id, this is mostly for compat with legacy block format, where 0xFF indicates end of extensions list, and other values may be either used or ignored.

> if this one byte is currently only used to signify "end of extensions", then maybe we could just as well use the string id with first byte set to '\0' instead.

I did an experiment, but did not like how it came out.
The problem is that I'd really wish to share same format rule for room and main game file, and room files already have blocks identified by a 1-byte number, where 0xff means end of the list. It may be possible to check first byte in header both as potential numeric id and a first byte of a string, but this complicates the format rule. Or maybe I just failed to figure out a good way to join them.
In the end I decided to keep this extra 1 byte for simplicity sake:
* if it's 0xff - that's end of the extensions list (like it's been in the room file before);
* if it's > 0 - that's an old-style numeric id;
* if it's == 0 - next follows a 16-byte string id.

Added same logic to the room file, so it's now allowed to have blocks identified with a string, while supporting old-style blocks.

ivan-mogilko:
Amended with simple read/write helpers, also added new format support in the editor.

ivan-mogilko:
> Hey, after this is merged, it would be cool to add an entry in the wiki about how the meta format works:

Added wiki page: https://github.com/adventuregamestudio/ags/wiki/Extending-game-and-room-data-format

PR #1186, opened 2021-01-20
AGS 3: ensure to check special key combos prior to in-game keys

### LINKED ISSUE #1176: bug: when started with --fps, CTRL-E is passed to the game
CTRL-e, which can toggle between max FPS and game-set default is supposed to be caught by the engine and not passed to the game. but this is not what happens. therefore in the game "Abducted - 10 Minutes" (available on AGS website, 0.9 MB), during the intro is played, the CTRL-E is ignored. one has to wait (or skip) until intro finished in order to make it work.

ivan-mogilko:
Fixes #1176.

These are key combinations that are not part of the game itself, therefore they should be given a priority and checked regardless of current game state.

### DIFF
--- a/Engine/main/game_run.cpp
+++ b/Engine/main/game_run.cpp
@@ -296,7 +296,7 @@ static int get_active_shifts()
 
 // Runs service key controls, returns false if service key combinations were handled
 // and no more processing required, otherwise returns true and provides current keycode and key shifts.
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Ok, I tested this once more and it seems working, so merging....

Issue #1187, opened 2021-01-26
android dockerfile: required file no longer exists


rofl0r:
for the file mentioned here
https://github.com/adventuregamestudio/ags/blob/master/ci/android/Dockerfile#L72 i get a 404.

+ wget https://www.apache.org/dist/ant/binaries/apache-ant-1.10.7-bin.tar.gz
HTTP request sent, awaiting response... 404 Not Found
2021-01-25 18:53:37 ERROR 404: Not Found.

the oldest version on that ftp directory is 1.10.9

it's odd that CI still works with that failing in the dockerfile...

rofl0r:
thinking about it, i could imagine that cirrus has the image of the docker container cached, since it would take a long time to regenerate this for every commit. would be interesting to know whether it would still work with ant 1.10.9, though as it appears that apache is removing downloads of older binary releases, this file should be hosted somewhere else ASAP in case it is compatible.

ericoporto:
It's indeed cached, but it should work 1.10.9, if you want to try, you can change that version number to the one that produces the valid URL, and do a PR, this should trigger a rebuild of the Docker. Don't worry, even if something else fails this won't invalidate the cache of our current Dockerfile. The caching uses a hash of the Dockerfile as it's index and you can also pull images locally for inspection - if you look into the Cirrus-CI screen you can see the URL for the hash image used. Cirrus doesn't drop cache usually - unless we ask for.

Android through Ant though is unsupported by AOSP, Google and our CI provider - you can look into Cirrus-CI documentation for Android and it explains how Android apps should be built there. It's useful to migrate the build tools because this allows taking advantage of proper test runners and following modern Android practices are required to get apps accepted in Google Play store, which is a source of income for game developers.

rofl0r:
> Android through Ant though is unsupported by AOSP, Google and our CI provider - you can look into Cirrus-CI documentation for Android and it explains how Android apps should be built there. It's useful to migrate the build tools because this allows taking advantage of proper test runners and following modern Android practices are required to get apps accepted in Google Play store, which is a source of income for game developers.

lol, i was actually looking into that to create my own suckless toolchain to create android apps without java for my own use (without app store), and was positively surprised that ags uses a custom framework setup that supports old android versions, unlike most other ci docker files.

> Don't worry, even if something else fails this won't invalidate the cache of our current Dockerfile.

cool, will file a PR then.

ericoporto:
[...]

PR #1189, opened 2021-01-26
android dockerfile: bump ant to oldest version still available

rofl0r:
ant 1.10.7 was removed from apache server.
if this PR results in a successful build, the now used ant version 1.10.9
should be archived ASAP and made available from another URL via a mirror.

addressing #1187


ericoporto:
The error with MacOS I think is a fluke but it may be because Cirrus-CI moved on and is offering only Catalina and BigSur. I haven't checked how are things for building from source on Catalina and need to check on that.

ericoporto:
Done, task re ran successfully :)

Issue #877, opened 2019-07-23
exorbitant size/bloat of new ags games


rofl0r:
most new games that come out are 100s of MB big. a big share of it are speech and audio files, which the authors usually ship in studio quality, i.e. 44.1KHz, 16bit stereo.
as an example, the new game metaphobia ships in a .rar archive that's > 400MB.
by simply converting the mp3 and ogg files in it to lower bitrate (while pertaining pretty good quality, using the [instructions in agsutils' README](https://github.com/rofl0r/agsutils/blob/master/README#L28)) my [repack here](http://ftp.barfooze.de/pub/sabotage/tmp/Metaphobia_repack.7z) is "only" 80MB.

but that's still a lot. the game uses a resolution of 320x200, so it should be tiny. it seems all graphics are stored in 32bit color (and maybe even uncompressed), instead of 8bit and using a palette. At least that's what i assume, because i was debugging slow rendering speed and noticed that all allegro bitmaps in use (including the screen surface) used costly 32bit ops.

so my question is:
could this be improved by the editor, for example by counting the number of colors of all actors in a given scene as well as the backgrounds, and then creating a palette and 8bit images automatically, and store indexed images ? or does the user even already have to use custom palettes, and the editor just fails to generate proper 8bit indexed images ?

also, how can we make the game authors aware that they can produce a game that's 1/5th of the size simply by resampling their audio tunes a little ?

and can the engine use 8 bit or 16 bit mode by default for faster blitting speeds ?

sonneveld:
I don't think converting to 8bit is the answer. I'd much rather we encourage 32bit for everything as it much easier to render under OpenGL and provides alpha channels. 8bit and 16bit requires conversion to 32bit anyway under accelerated graphics drawing.

I've had success with disabling the simple sprite compression and storing all resources in a zip instead of the clib format. Even simple deflate compression cut down Unavowed from 2.5 gig to 1.5 gig. There's newer compression formats such as zstd which might help further.

There's also the issue of some games bundling the Windows and Linux data together. It seems some authors might not realise that the Windows exe is bundled already.

ivan-mogilko:
This is most ironic, because community members were suggesting removing 16-bit mode from engine for years (and some even suggested removing 8-bit mode). All the game templates were converted from 16-bit to 32-bit recently because 16-bit setting caused confusion between newcomers who tried to use sprites with alpha channel in their games (or had their sprites loosing colors).

[...]

PR #1190, opened 2021-01-28
AGS3-SDL2 Android Port

ericoporto:
Hey, I am hitting some troubles but since it's working I decided to do a PR. 

### Main changes

- OpenGL ES2 compatible renderer, needed for modern phones and useful for future WebGL implementation.
- Automated standard build process
- Bonus 64bit Android compatibility
- Renamed Launcher to Player because Launcher means something different in Android
- ListView deprecated -> using RecyclerView
- A ton of deprecated things are replaced by androidx (Jetpack) equivalents
- Player written in Kotlin, since it's the now recommended language to use for Android
- Moved from old Eclipse project to new Android Studio project (also ant -> gradle)
- producing Release and Debug versions of artifacts 
  - my reasoning is around 4x performance difference between Debug and Release of the APP, but debug is useful for dev

### Building and Development

- Android Studio 4.1.2 or later
- CMake 3.16.3 installed or other (update library/runtime/build.gradle if using a different version, 3.14.5 is the minimum)
  - On Windows, make sure CMake is added to PATH
- [Ninja](https://github.com/ninja-build/ninja/releases), any newer version will do (Android doesn't use Makefiles in this way, but ninja is faster)
  - On Windows, put the extracted `ninja.exe` in a directory named ninja somewhere and add it to PATH

Note: as soon as CMake 3.18 hits Android Studio it and ninja won't be required to install separately, it's on the beta channel, but not on stable yet. The selected version matches the one on Cirrus-CI distro. The result would be having only Android Studio as requirement for building! (my hope is this makes things better for people developing from Windows)

<details>
On Windows, apparently the first time run has to be the `Android/library/runtime` or it will complain about licenses

On Windows, if using the beta, CMake and NDK can be installed from the menu.

On Windows, if using the beta, CMake version can be set under the build.gradle file in the runtime project to match

Things should just build, manual sync can be done by File->Sync Project with Gradle Files
</details>

Both Kotlin, Java and Cpp code can be edited simultaneously in Android Studio now with full debug support. You can follow the variable going to native which makes things easier.

An internet connection is required since the first time gradle sync runs it will fetch dependencies. 

### TO-DOs

- Other current SDL2 port bugs we are facing are also present here.
- Touch hasn't been redone yet, so this is using SDL2 emulated mouse. 
  - found out that if I ignored `android.cfg` and use `control_enable = 1` on `acsetup.cfg` I get relative mouse movement, but I don't know why! This mode through android.cfg would support longClick is left click
  - ~~`config_mouse_control_mode` is the variable in `android.cfg` but don't know how to pass it to `engine_post_gfxmode_mouse_setup`~~ figured and commited.
- Still need to modify the Single Game project so it can work with this.
- Need to NOT use External Storage since it's deprecated [see 1](https://developer.android.com/about/versions/11/privacy/storage), [see 2](https://medium.com/androiddevelopers/android-11-storage-faq-78cefea52b7c), [see 3](https://commonsware.com/blog/2019/06/07/death-external-storage-end-saga.html)
- Running Last Save from the AGSPlayer ui is broken but I don't know why
- Delete parts of unused code in `acpland.cpp`


Engine/platform/android/acpland.cpp
```diff
@@ -611,40 +610,40 @@ bool ReadConfiguration(char* filename, bool read_everything)
 
 //    ReadInteger((int*)&psp_return_to_menu, "misc", "return_to_menu", 0, 1, 1);
 
-    ReadString(&psp_translation[0], cfg, "misc", "translation", "default");
+    ReadString(&psp_translation[0], cfg, (char*) "misc", (char*) "translation", (char*) "default");
 
-    ReadInteger((int*)&psp_config_enabled, cfg, "misc", "config_enabled", 0, 1, 0);
+    ReadInteger((int*)&psp_config_enabled, cfg, (char*) "misc", (char*) "config_enabled", 0, 1, 0);
```
ivan-mogilko, line 618: I noticed these changes in previous variants, probably dating back to sonneveld's work, and hoped this won't be in pr.

There's no reason for ReadInteger and similar functions to demand char* pointer, as it is not supposed to change the string. The correct way is instead to fix the ReadInteger and other function and change their "char*" arguments to "const char*".

rofl0r:
hmm... wouldn't it make more sense to get the sdl2 port working with the current toolset first, rather than throwing everything over board and jumping on the latest toolset?
the update to latest could then be done as another step, so people requiring support for e.g. old android versions would at least have one point known-working in the commit history to check out.

on that topic, what's the oldest android version that can run the agsplayer created from this branch using the all-new toolset ?

Engine/platform/android/acpland.cpp
ericoporto: Right, I will look into doing it this way.

ericoporto:
I believe the oldest in theory is KitKat 4.4, and the oldest SDL2 can go is 4.1. I verified on my Galaxy Tab 4.0 that it didn't ran on it, I successfully ran on my Android 6, Android 7 and Android 10 devices. I also have a Android 8 but haven't had full test on it. I am hitting trouble on 11 on emulator that I believe is the unsupported storage, since I know you guys love backwards compatibility I wanted to do the PR before it gets refused because I switched the storage model.

Engine/platform/android/acpland.cpp
ericoporto: I believe I updated the right places! Can you check if it's correct?

ivan-mogilko:
I am very confused by above comments, but I don't know that much about Android.

> About version support, I believe the oldest in theory is KitKat 4.4, and the oldest SDL2 can go is 4.1. I verified on my Galaxy Tab 2 (the 2012 model), which is 4.0 that it didn't ran on it, I successfully ran on my Android 6, Android 7 and Android 10 devices

I already forgot, what was the previous minimal support, for the reference?
I have one Android 5.1 device which I may use to test this later.

> This basically means a directory named ags under documents instead of any directory - or something similar that satisfies the new docs on this topic.

To confirm, you are talking about place where the games have to be installed?

ericoporto:
> I have one Android 5.1 device which I may use to test this later.

This will be super helpful! Please test and report results. Android Studio should help to debug in case there are problems!

> To confirm, you are talking about place where the games have to be installed?

Yes, so basically today we give permission for the App to access ALL FILES on Android. But this permission has gone away in newer Android APIs, so this is not a viable permission to use anymore. Google has MEDIA things (Podcast, Music, Images, ...) which I couldn't find a match and also has Documents, which is weird but I think is the best one to match previous behavior. We can't use Downloads too.

> TBH, reading your post I became unsure if you are talking of something that's already in place in this PR, or something that you want to do in the future?

Wanting is a strong word. But we HAVE to do this or it won't work in newer devices according to the documentation. I have looked into but this will lead to changes in the App.

Other alternative method is having a directory for the App itself, so in the file explorer you would share the game file with AGS, and it would pickup this stream and save in it's own directory. The advantage of this would be access to the Downloads folder, even though indirectly, it's a valid workflow. 

Both methods above together are also a valid option. I have experimented with the sharing interaction (it's how is done in love2d) but I haven't experimented yet with using the Documents directory - I found the docs lacking, I can do, but I need a bit more time to work on it, but fundamentally it changes the app, so the reason of the PR.

> Theoretically, what consequences does this have to existing users?

They need to move the game files from internal storage to Documents directory. You may ask about using a real life external SD card, and to that I don't know yet, the sd card readers on all my phones that have one are broken so I couldn't experiment with this. But external SD card is not internal storage, it uses a different permission - since the OS doesn't "live" there, so I believe these are not deprecated or affected but will have to exist explicitly.

rofl0r:
> the oldest SDL2 can go is 4.1

actually i just tried to build an SDL project with the toolchain ndk-r16b etc as used by the existing dockerfile, and it fails during build of SDL2's java files, as they use features introduced in API level 18 (which IIRC is corresponding to android 4.3). i filed a bug for that, but unless that is fixed the docker ci toolchain needs to be updated to 18 anyhow. from my initial experiments though it looks as if the build system used in the past is a lot nicer than current one. the old one uses ant and gnu make behind the scenes, whereas the new one uses gradle, cmake and ninja, i.e. a clusterfuck of extremely bloated applications, and a much newer openjdk is required as well.

ericoporto:
Literally the recommended options by Google.

The previous option has no package management.

The world moved on, grow up.

Engine/platform/android/acpland.cpp
ivan-mogilko: Yes it's fine now.

ivan-mogilko:
@ericoporto 
> If the preferences are done using the existing GUI from AGS we just need to implement a combobox component.

Frankly i am confused and don't know how all this is related to android port now, but want to mention that combobox may be emulated in AGS using 2 buttons and a listbox on separate GUI (basically, create a combo box from its components) :-).

ericoporto:
So basically I thought when doing the code in this PR that these Android particularities (android.cfg and it's GUI for configuration), since they only exist in the Android port, they receive less attention. I don't think it's a good idea to change this now (in ags3) but I wanted to figure out something in ags4 that could work for everything. The work is one that requires thought on what goes in acsetup.cfg and how the interface should look. I didn't want to distract this PR by going this way. 

I got this idea because I had difficulties with the feature "load last savegame", it uses a codepath not used on other platforms, as far as I can tell _appears_ to be correct, but doesn't work, and I don't know why.

Other thing I had difficulties is touch, relative mouse mode is set on `engine_post_gfxmode_mouse_setup` and I could not figure out how should `config_mouse_control_mode` (which should set relative mouse mode) be passed to there. Should I just `extern` and read the variable in `engine_post_gfxmode_mouse_setup` ?

For now, I tried to maintain the UI, I just wanted something as close as what existed for the user but having an updated codebase that is ready for the current Android Studio, runs on current devices and that offers a runtime compatible with Play Store restrictions. 

For ags3, I think maintaining `android.cfg` is best. But in ags4 it would be best to move touch and orientation details to `acsetup.cfg`.

ivan-mogilko:
> Other thing I had difficulties is touch, relative mouse mode is set on engine_post_gfxmode_mouse_setup and I could not figure out how should config_mouse_control_mode (which should set relative mouse mode) be passed to there. Should I just extern and read the variable in engine_post_gfxmode_mouse_setup ?

I think the most straightforward way now is to have everything gathered in GameSetup struct, as the instance of this struct is already accessed there. This struct is also filled after reading config.

ericoporto:
> I think the most straightforward way is to have everything gathered in GameSetup struct, as the instance of this struct is already accessed there. This struct is also filled after reading config.

Can you tell me where this happens? I could not find where this happens. 😕

ivan-mogilko:
The config is being applied in `void apply_config(const ConfigTree &cfg)`. This is where it takes a key/value map and reads values from it:
https://github.com/adventuregamestudio/ags/blob/master/Engine/main/config.cpp#L444

This map is filled from various sources: command line, file, even adding options in code backward support. But in the end all gathered values come to one point there.

ericoporto:
So if I understood I would add at `override_config_ext`, a read of `config_mouse_control_mode` , writing it's value at the `control_enable` key, and things should work right? I got the exact relative mouse mode behavior when I manually set `control_enabled = 1` in `acsetup.cfg`. I will try later.

ivan-mogilko:
I'd say that ideally it should not even be done like that, instead we might merge contents of android.cfg with standard config, and then read android-specific options also in `apply_config` (or replace them with more generic ones). But I don't remember how and when android config is read, so can't tell right away what should be done for that.

ericoporto:
I added the relative mouse mode handling in `Engine/main/config.cpp`.

Now, about LongClick or LongPress, the SDL way is to read the actual touch events and handle this on the cpp side, manually, which I think is under `Engine/ac/sys_events.cpp`. The other way would be to add an `OnLongClickListener` on the `SDLSurface`, but I think this would be hackish...

I thought something like a `mouse_button_state` is set to `MouseBitRight`, without modifying mouse position, if `on_sdl_touch_down` is followed by a `on_sdl_touch_up` after a time without a `on_sdl_touch_move`. I am not sure though if you already had something in mind for touch events there on `sys_events.cpp` or if you thought about placing them differently, so decided to ask.

ivan-mogilko:
No, I had no plans for touch events, barely know which kind of actions are there.

> I thought something like a mouse_button_state is set to MouseBitRight, without modifying mouse position, if on_sdl_touch_down is followed by a on_sdl_touch_up after a time without a on_sdl_touch_move.

So, "long press" is translated to right mouse click? One thing I am not aware of, what defines the minimal duration of long press which distincts it from "short press"?

rofl0r:
a question about the addition of the glm library: is its code only needed for android, or other platforms too? if so, would we need to add it to master (ags3) branch too for feature-parity?

ivan-mogilko:
> a question about the addition of the glm library: is its code only needed for android, or other platforms too? if so, would we need to add it to master (ags3) branch too for feature-parity?

In this pr it replaces matrix functions in OpenGL renderer unconditionally, therefore will be used when compiling other ports. I would not say it's *required* for them at the moment, but since it's added anyway I think we're planning to use it for some other math operations in the future engine.

Current ags3-sdl branch is temporary and will be merged into master as soon as main set of ports are working (I think android is the last remaining).

rofl0r:
> Current ags3-sdl2 branch is temporary and will be merged into master as soon as main set of ports are working (I think android is the last remaining).

ah, so master becomes ags-3.6-development. what will happen to the current allegro-based 3.5 code.

ivan-mogilko:
> ah, so master becomes ags-3.6-development. what will happen to the current allegro-based 3.5 code.

Well, it will remain in history. I could create a tag or branch right before the merge to mark the last point before sdl2 port.

Actually, I am still considering to make a minor 3.5.1 release with all these fixes added to master in past few months.

ericoporto:
The renderer is really important for WebGL, it's the one I am using in the webassembly version of the Engine. It has also been tested with JS builds but it is faster in WASM. This renderer is also helpful for iOS until I have time to write a proper Metal renderer.

For now I would just do WebGL for ags4 though.

I really wanted to also make a vulkan renderer one day  and glm is needed there too. I believe better performance can be achieved on Android with Vulkan.

ericoporto:
Added the single game launcher made by @monkey0506 and did minor changes for it to work with the SDL2 Android port.

rofl0r:
> The renderer is really important for WebGL

isn't that something for AGS4 ? i thought it's been decided that ags3.x development from now on consists only of bugfixes (after SDL2 port works).

> actually i just tried to build an SDL project with the toolchain ndk-r16b etc as used by the existing dockerfile, and it fails during build of SDL2's java files, as they use features introduced in API level 18 (which IIRC is corresponding to android 4.3). i filed a bug for that

actually i spent a couple days to investigate the issue and it turned out that SDL2 port can indeed be compiled with the current docker/ci toolchain. the trick is to set the SDK version for build to 26, but older android versions work regardless. here's the bug report for details: https://bugzilla.libsdl.org/show_bug.cgi?id=5515

rofl0r:
> Added the single game launcher made by @monkey0506 and did minor changes for it to work with the SDL2 Android port.

yay, another 10.000 lines added, including a copy of half of google's sdk.

ericoporto:
Google actually hasn't updated these libraries, so I can't use upstream

ivan-mogilko:
> Added the single game launcher made by @monkey0506 and did minor changes for it to work with the SDL2 Android port.

Please explain, what is this launcher and what is it used for?

Also, in general, if you are porting someone elses code into the project, I think it's a good idea to mention them in commit description, with a github tag if they have one. For example in this comment I reference its based on sonneveld's work: 62d39e3f394619c5ae8b219b085c4554feea1062

ericoporto:
@ivan-mogilko it's the latest version of the Tea for Two code that is referenced here:

https://www.adventuregamestudio.co.uk/forums/index.php?topic=55681.0

I cleansed of my own games and things, I will reedit the commit to correct the README so it's more updated. I just thought that I have been maintaining these in my repos for a long time now and could put it here so I could make it nicer. I can properly refer in the commit too, I can also remove. In my mind it made sense to be here instead of my account.

ivan-mogilko:
@ericoporto 
Excuse me, but I am still missing a chunk of knowledge here.
Do I understand correctly that this "single game launcher" is like a template for creating a proper APK for an individual game?

ericoporto:
> Do I understand correctly that this "single game launcher" is like a template for creating a proper signed APK for an individual game?

Yes. In a perfect world, it would be simpler but Google has requirements towards exactly how an app bundle should be. An app has a size limit (100MBs but it has some gotchas), so instead of everything going in the Apk you actually have AGS with the name of the game in one package and the rest of the resources in a different file, an obb, and then either the Google Play Store delivers both or it delivers the APK and then you get the obb. Then the obb gets mounted and you get to play the game.

Things are a bit more complicated though, nowadays you actually submit an app bundle which for the Apk part and the Play Store figures which architectures your app supports and generates optimized builds of your app for each types of phone. And also everything I told about the obb is deprecated since this year and now the way it works is we use a Play Asset Library - which is a Google Play only thing for doing exactly what was done before with obbs, but in a way Google can read the assets and optimize the assets for each phone - it gives tooling for this which we won't use but we could. Also I need the other storage type to switch to this and I will need to figure a way to make it optional since one may want to not have anything to do with Google Play.

But like, after May there's just no way to release on Play Store without integration with the Play Asset store - and commercial AGS Games are always bigger than 100MBs.

Also Google really doesn't recommend compression since it's compressed on transport and nowadays you should access mounted assets directly instead of copying to a cache - but they don't forbid doing so for now.

> Do you refer to default android readme, or there's a separate one? In latter case could you give a link to it?

Sure, here's my full history:

https://github.com/ericoporto/mythsuntold_dungeonhands
https://github.com/ericoporto/vacaroxa_hellspuppy
https://github.com/ericoporto/teafortwo

The first one actually had a game so one could build and actually run the thing. The other ones don't have binary parts and are more with the goal of allowing people to fork them - in the spirit of monkey's original one. I actually have some more but they are private - they are of others.

> Are these files necessary for anything that we build in our repository? I know we were building a) engine libs to include into APK and b) universal game launcher which is a java (now kotlin?) app to startup any game from device file system.

No, these are more similar to the Windows, MacOS and iOS port that are made for shipping game instead of being a game player. 

> Does this "single game launcher" require any of our source files directly to be built?

Not sure what you mean, it needs the runtime, which could be a library, but it must be in sync so it works. So the runtime would need to be sinced to Maven Central if the single game launcher is not here.

The Android port we had was the only one that wasn't able to directly ship games.

rofl0r:
> and commercial AGS Games are always bigger than 100MBs.

this is due to using bloated resources and super-high screen resolutions as mentioned in #877 , and probably also current ags making it too easy to do sloppy game engineering: older engines enforced a number of hardcoded limits which made the developer think twice whether he wants to use a resource or at least check whether there's unused stuff he could remove to free up some slots.
even in 2021 it should be possible to make good-looking adventure games when using 16 bit color instead of 32 and a compressed image format, or even 256 colors with palette, and using midi music or at least transform music resources from 24 bit studio quality to something reasonable.

ivan-mogilko:
The thing is, I have no idea what this "single game launcher" is, and we never properly discussed this before. It came as a surprise to me. This is why I wished to have as simple explanation as possible. Unfortunately there's still too much information above to process, and from a quick glance much of this information is details which are not important right now, like about size, compression, etc... I dont think I need to know that at the moment.

What I'd like to know is, first of all, if this is a completely new thing to our repository? If yes, then I believe it should definitely not be a part of this PR, because this PR is dedicated to adjusting Android port to the new SDL2 backend. At the very least it should be made a separate PR later. IMO it's not good to have too many things in one pr, as it increases number of things we have to agree on, try out, and so on.

Then, I understand that it has to be in sync with our runtime ofcourse, what I wanted to know is whether it's reliant on actual sources. I need to know this to see a full picture of how the projects are interconnected.

@rofl0r 
> even in 2021 it should be possible to make good-looking adventure games when using 16 bit color instead of 32 and a compressed image format, or even 256 colors with palette, and using midi music or at least transform music resources from 24 bit studio quality to something reasonable.

I really want to ask you not to bring this up here, firstly because it's not relevant in any way, secondly because it's all very arguable and different from how people actually make games. Let alone the fact that we do not impose such restrictions on them.

ericoporto:
@ivan-mogilko I will remove and PR separately then - but I did wrote about it on the readme before. If you can please tell me what is need from you to get the PR accepted.

ivan-mogilko:
I guess it's about updating two things: the engine libraries which are built for android, and existing game launcher APK. Making sure that engine runs and works on Android with the new backend.

I realize that do not pay enough attention to the state of Android port though, to which I am sorry. If something _extra_ is necessary for Android support I wish there was a separate ticket open for that so at least everyone had a chance to be properly informed.

> but I did wrote about it on the readme before

I guess I missed it... not sure which readme you are refering to though.

rofl0r:
i think ivan wants you to not use an improvement PR as a backdoor to bring in other unrelated changes (that you desire), e.g. in this case the PR is supposed to make android build working (which could be made in less than 500 changed lines as my experiments with the existing docker toolchains showed), yet it adds a new library in-tree (70Kloc) and a new launcher that you find "nice" (10 Kloc), and uses all-new android infrastructure instead of existing one.

ericoporto:
Not sure what to say, the Android port doesn't allow shipping games - it doesn't build to "hey it's my game play this", like literally all other ports.

ericoporto:
@ivan-mogilko if you go to my branch and look into the Android directory there's a text file explaining things.

ivan-mogilko:
I'd like to make this clear, that I do not know Android well enough to judge about toolchains and similar topics. You may compare me to half blinded person here, at least this is how I feel when dealing with this part of the project. For the same reasons I do not feel confident to deny something, and dont think I am in right to unless there's an obvious problem.

> Not sure what to say, the Android port doesn't allow shipping games - it doesn't build to "hey it's my game play this", like literally all other ports.

Yes I am fully aware of that, and this was in plans to improve. But as its own thing, not a part of SDL update.
We would also need to discuss how to integrate this with the Editor.

rofl0r:
> The old Android port was in the past "Run Android Studio"

actually the ags engine built by the ci was specifically set up to work without any GUI, and it uses quite up-to-date components to do it. the SDK is version 26.1.1, only NDK is a little bit old in that it uses GCC 4.9. i actually tried latest android studio stable (4.1, iirc) in ubuntu 20.04 which i installed in VM for this purpose, and it was buggy beyond hope, i never got even a simple SDL2 hello world compiled with it, while it was trivial using AGS' docker setup.

ericoporto:
@ivan-mogilko can you be specific about what are YOUR requirements? And do YOU plan to maintain the Android port?

ivan-mogilko:
> can you be specific about what are YOUR requirements?

Requirements for what?

> And do YOU plan to maintain the Android port?

I hoped to not do that, because I am already not fast enough to do everything that has to be done, or even what I wanted to do...

ericoporto:
@ivan-mogilko I really don't have a real Android 5 phone at hands so you hold the device I need to verify how it works. It may not even run, which is fine, but this is valuable input! Like I really appreciate your help testing it! It would mean the world to me if you can next weekend, find some time to play around with it, the CI should be building the Engine correctly, so it should be possible to install from there. I think that if you try you will be able to tell me what you like or not and we can work a concrete list of acceptance for this PR if you are then comfortable with it.

In the meantime, I will drop the commit you requested and move to a different branch - I will actually use for a game I am making anyway.

rofl0r:
> I really don't have a real Android 5 phone at hands so you hold the device I need to verify how it works.

as i mentioned in the libsdl bugzilla (link above that nobody looked at) you can run android-x86 in any PC emulator:

https://osdn.net/projects/android-x86/downloads/65697/android-x86-5.1-rc1.iso/

ericoporto:
@rofl0r it does run on emulated Android 5 but sometimes people's devices are different. The example that most trips me is that a bunch of functions in the API don't follow symlinks, and directories change a lot between manufacturers, or like, Open GL support is not completely there (the previous port had issues on this area). There are other issues but I don't know much else than "I got the error in the debugger and had to find a different way to do the same thing".

ivan-mogilko:
Sorry for delay.

I started with testing new AGS Player app. It runs, and hardware renderer now can use shaders correctly.

The biggest problem I found to this moment is the menu. In the previous application you could long-press the Back button to bring a menu, which allowed you to:
- bring up virtual keyboard;
- configure (virtual) hotkeys;
- exit game;
In AGS Player long-pressing Back buttons does nothing, and I found no way to shutdown the current game unless the game itself provides such function. Ofcourse real games do, but unfortunately some dummy test games I created in the past in order to test rendering problems do not have full proper in-game GUI. So I had to abort them by hand from the list of running apps.

Other than that, differences and issues I noticed:
* Launcher's background changed from black to lighter theme, which is ok;
* Ags games in the list now have full paths to their data noted after the title (good for testing, I guess);
* when I run AGS Player from android menu again it brings me to the game selection. At first I thought it would let me run several ags games in parallel, but unfortunately touching anything just opens same running game again.
* When I ran ags game for the very first time, it displayed a hint that I may "switch from fullscreen" by swiping from top... or something like that. Unfortunately I missed what it said, and swiping anyhow did not give any result.

There also was a strange occurance, a cursor stopped responding properly after switching in and out the game multiple times, but it got fixed after I shut it down and restarted, and I cannot reproduce this yet.

So, I guess it's fine so far, except I'm missing this menu. Did I miss a way to call it, or if there's no one, do you have plans on implementing something similar in the future?

I am going to try building Android app myself now following new instructions, will take me some time though.

ericoporto:
> In the previous application you could long-press the Back button to bring a menu, 

Oh, yes, I started adding that but stopped but it should be possible. In the old Monkey single game I think back was Esc when I played, so I wanted to allow both behaviors, because I think the old one allowed both behaviors - just not sure how yet.

> Launcher's background changed from black to lighter theme, which is ok

Android at some point added dark/light mode on top menu, so it actually has two themes and follows what the system is set to - either light or dark.

> Ags games in the list now have full paths to their data

It was just an idea to help me understand things.

> when I run AGS Player from android menu again it brings me to the game selection. At first I thought it would let me run several ags games in parallel, but unfortunately touching anything just opens same running game again.

This should not happen at all, but I think I reproduced what you did. You did not kill AGS App and tried to load the player, this brought you to the main Activity. Problem is Runtime activity is already running, so you can only bring the Runtime activity back. So my guess is some way to detect that the Runtime activity is running and bring it instead of the main activity. Probably if you had the back menu you would have exited the app by using it.

> Did I miss a way to call it, or if there's no one, do you have plans on implementing something similar in the future?

I haven't added it yet, but I will now. 

Thanks for testing!

ivan-mogilko:
After few hickups I was able to build my own APK with the Android Studio on Windows. It works, but the overall process felt weird because I didn't understand 99% of what was happening :-).

Couple of things to note.

1. For Windows at least instructions should mention to put path to Ninja tool in PATH enviroment variable, or CMake won't find it. That seems kind of obvious, but someone may forget or simply not realize what the CMake error message tells them.

2. Output ags player apk is called something like app-debug.apk, I wonder if it should be agsplayer.apk or similar instead.

3. Following files were generated during the build, and not tracked by git. Should these be added to .gitignore or not?
<pre>
Android\agsplayer\.idea\.gitignore
Android\agsplayer\.idea\.name
Android\agsplayer\.idea\compiler.xml
Android\agsplayer\.idea\gradle.xml
Android\agsplayer\.idea\jarRepositories.xml
Android\agsplayer\.idea\misc.xml
</pre>

4. I found Android Studio itself kind of weird at times. Well, I never used it before. For instance - it was not clear if it catches any configuration changes or not. For example, I forgot to install necessary CMake version, and it kept telling me that it is not found even after I updated. I had to reboot system to make it realize there's a new version. Then again, maybe it's a Windows thing.

ericoporto:
> 4. I found Android Studio itself kind of weird at times. Well, I never used it before. For instance - it was not clear if it catches any configuration changes or not. For example, I forgot to install necessary CMake version, and it kept telling me that it is not found even after I updated. I had to reboot system to make it realize there's a new version. Then again, maybe it's a Windows thing.

It's possibly the environment variables if you are not on Windows 10. I think previous Windows required logoff and logon to refresh those. Important note is both Ninja and CMake come with Android Studio soon, so it should not require anything else - from Android Studio 4.2 onwards. The 4.1 and older come with older version of these tools so this is why it needs them externally. I use the 4.2 on Windows (it's Beta now but it should release to stable quite soon!).

Gradle for building everything, it has an Android plugin for building things Android, and is what Android Studio uses by default. It abstracts a bunch of stuff, but basically, for the cpp parts, gradle loads a CMake toolchain file that configures a lot of things, and gradle also uses information for things like minSdk and other flags and applies the NDK equivalents, CMake then generates a Ninja file which is run and passes the building to Clang. You can set the architectures to build on the build.gradle for the runtime in `abiFilters`. I usually let it be just the platform I am debugging/coding so it only builds for that specific platform.

3. I think it's better to gitignore them, yes!

2. It's because I had no idea what to name it but it's an easy fix on gradle, will do!

1. I will note the above and also add a note that depending 
on the version of Android Studio this is not needed.

ivan-mogilko:
@ericoporto could you tell which state this pr is, do you plan to change/fix anything further?

ericoporto:
@ivan-mogilko sorry, I got distracted, but I tried again today and got to where I stopped.

I can't figure out how to capture the back button without modifying the SDL Activity. Basically, the Back Button is captured in a onKey override in the surface itself (not the SDL Activity, something in it...), and so the only way to capture the back press is from the cpp part.

So on `sys_events.cpp`  (if it worked it could be moved to acpland, but ignore this for a moment), the way to capture is this:

```c++
    case SDL_KEYDOWN:
#if AGS_PLATFORM_OS_ANDROID
        // capture Android BACK Button
        if(event.key.keysym.sym == SDLK_AC_BACK) {
            // we are going to call onBackPressed in AGSRuntimeActivity.java!
            JNIEnv* env = (JNIEnv*)SDL_AndroidGetJNIEnv();
            jobject activity = (jobject)SDL_AndroidGetActivity();
            jclass clazz(env->GetObjectClass(activity));
            jmethodID method_id = env->GetMethodID(clazz, "onBackPressedNative", "()V");
            env->CallVoidMethod(activity, method_id);
            break;
        }
#endif
```

The idea here is to call `onBackPressed` function from the AGSRuntimeActivity, and the above code actually works, if I put the debugger on a void method named `onBackPressedNative` (named so it won't clash with `onBackPressed`), well, the method actually gets called! The problem is, for some reason I can't figure out yet, instants after this, we call `update_polled_stuff_if_runtime()` and this makes SDL crash. I don't know why, and this doesn't happen if I don't try to invoke methods from java directly at that specific time.

I also foundout that SDL has `SDL_SetTextInputRect` and `SDL_StartTextInput` for showing the keyboard directly from the Engine and it would support any OS with software keyboard. But going this route I don't get the other menu.

So basically what I can do is hack around SDL and prevent it from capturing the back key in first place, and handle the thing cleanly on Java, and then on the future in AGS4 we don't need this hack because by then we have proper multi-touch support in AGS and we can do better. But once the game gets running, I can't seem to be able to run Java functions directly from AGSRuntimeActivity without crashing.

ivan-mogilko:
First of all, if restoring menu functionality would take additional effort, then it may be left for later. We will definitely have to beta test and fix SDL port more anyway.

If you could tidy up branch (if necessary) and split single player addition into a separate PR (which I'd like to try out too as soon as I get spare moment), then I'll accept this PR already. Then we can finally merge master and sdl2 branch together (and hopefuly merge this into ags4 too to sync them all).

In regards to comment above, I have to say I don't know much about what the "Activity" and "Surface" are, and what their purpose is, except these may be Android system classes(?), so it's difficult to judge or suggest anything atm.

Something I may note:

* In the old Android port back key was handled inside its own library rather than the engine. This is seen here: https://github.com/adventuregamestudio/ags/blob/283c1d5/Android/library/src/com/bigbluecup/android/AgsEngine.java#L365
Class AgsEngine extends Activity, so this is Android Activity implementation, if I understand right.
And yes, it's the `dispatchKeyEvent` that you also mentioned. I haven't yet investigated the new code, if something changed there in this regard, so idk if it's still possible to do same thing or not.

* I heard that `SDL_SetTextInputRect` allows to display a text input field on screen. But this does not sound like direct equivalent to virtual keyboard, as was not about displaying text field, but rather a way to send key messages into the game.  Unless SDL_SetTextInputRect actually also causes a software keyboard to appear on screen on Android?

ivan-mogilko:
Another thing... I was reviewing some code again, and being honest I still feel very uneasy about addition of glm library.

This is not because I don't like the library itself, it has its uses, but it's intended for programs relying heavily on 3D math, and AGS does not and won't in the nearby future. It feels very wrong adding a large math library while we only need like 3-4 matrix functions.
Plus, there are already separate matrix functions for Direct3D renderer, and it's not clear whether it will be convenient to reuse glm for it (as we mentioned earlier, D3D has inverted matrix format).

Say, what if I'll make a commit adding one header with our own matrix functions currently necessary for our OpenGL/D3D renderers?

We may still consider glm later if somehow AGS will demand more sophisticated 3d math.

ericoporto:
I am tired of arguing, and I am moving away from AGS.

ivan-mogilko:
I was not arguing I was suggesting. If it's too difficult, I could accept it as it is.

I really fail to understand why is everything that I ask or suggest cause this kind of reaction. If I did something wrong, then I am sorry about that.

ivan-mogilko:
If you have decided to leave for good anyway, are we allowed to use the commits from your PR?

rofl0r:
in case erico is serious about this, i can try an intiial android port based on the existing CI infrastructure and my related experiments. once it's working in CI we can still decide which of the additional features this PR presents can be adopted,

argh, you just merged, so forget it....

rofl0r:
decision making progress in ags: if someone throws a tantrum, just click the merge button to prevent hurt feelings

ivan-mogilko:
I'm sorry. I know this project is far from perfect, and many additions to it may be seen as non-optimal. But it often takes too much time to progress anywhere, and it may be better for majority of end-users to take existing solution.

rofl0r:
well, only 6 hrs passed since erico closed this so there was barely time to propose a better option.

ivan-mogilko:
> well, only 6 hrs passed since erico closed this so there was barely time to propose a better option.

There were years before that, and several months while SDL2 branch was created in this repository.

rofl0r:
loosing the work could have been prevented by pulling the branch from erico remote, but alas, it's too late (unless you forcepush HEAD to where it was 30 mins ago)

ivan-mogilko:
That's not the point... we were waiting for SDL port for too long. I was hoping to release a testable version so that more end-users could try it out.
There's a working variant. I don't know how long would it take for anyone else (e.g. you) to present an alternative. Furthermore, I don't know if you will be willing to work on other things android-related further. You did not seem to show any interest in android port earlier.

rofl0r:
i've mentioned in one of the first comments here that i've been working with the initial CI setup, and at some point reported that i got SDL2 working with the existing toolchain. i could probably throw a working setup for AGS together in a day or two, but you're right in that my android interest is not very extensive, and future maintenance would likely have been someone else's job. i had basically offered to do the work in order to prevent the 90K of bloat thrown into this PR from getting merged.

ivan-mogilko:
I believe that other contributors expressed a wish to move to Android Studio project in the past too, so probably this is just a matter of time when it's added.

The "single game player" is a completely separate thing that again probably should have been added anyway in this form to allow end-users build their signed APKs easier in a traditional way.

ivan-mogilko:
@rofl0r I looked around other game engines in the meantime, and they all seem to have moved to Gradle projects at some point. Hence I assume that's a generally preferred today.

That said, I must agree that we might have had a point at which SDL version is built using old toolchain, for backup and compatibility reasons, before moving forward to newer Android Studio versions. Having that was my original intention back when I just started pulling in sonneveld's SDL work, but then I got distracted by something else.

If building using old toolchain is a relatively simple thing to do, guess I could redo the branch and reorganize commits in different order. But I would apply new projects on top anyway.

rofl0r:
> they all seem to have moved to Gradle projects at some point. Hence I assume that's generally preferred today.

i don't really think it's preferred, it's just what google dictates. and it's a huge mess - android studio downloads one version but then to build a project you need a different one that's not compatible with the project settings - it's a huge mess and i wasted around 3 days trying to get my demo SDL app working, without success.
otoh when i updated ant in CI to last available binary version, it just worked without issues. it's also a lot faster. gradle took like 20x longer for every task then ant.

> If building using old toolchain is a relatively simple thing to do, guess I could redo the top of the sdl branch and reorganize commits in different order. But I would apply new projects on top anyway.

well, it was relatively easy before this commit was merged. now we have 10 KLOC of google's sdk copied inside for this new launcher app and it's almost guaranteed that won't build with ant.
the only thing i can see having success is to rewind the sdl2 branch to before the merge, but then what's the point if you later overwrite it again? we'd have only a single commit in history that could be built with the ant-based build system.

ivan-mogilko:
> i don't really think it's preferred, it's just what google dictates. 

Well, maybe it's the reason people want to have it then. I don't know all the details of course.

>  and it's a huge mess - android studio downloads one version but then to build a project you need a different one that's not compatible with the project settings

I have a guess that it might be possible to work without Studio itself, there are command-line tools available for download, and it might be possible to download necessary SDK and Gradle separately. I might have to research this more closely later...

PR #1191, opened 2021-01-29
AGS4: BlendMode remake

ivan-mogilko:
This is mostly a copy of work by @AlanDrake (#1124). As much as I understood he did not have time to fix up his pr, so I picked his commits, did some cleanup, and also added few more things for consistency.

* Implements BlendMode type, which presents number of traditional sprite blend modes: Add, Darken, Lighten, Multiply, Screen, Burn, Subtract, Exclusion, Dodge;
* Added BlendMode property to Character, RoomObject, GUI, Overlay both in script and in the editor (except for Overlays, which cannot be created in editor).
* Implemented DrawingSurface.BlendImage and BlendSurface functions, which work like DrawImage and DrawSurface but accept BlendMode as an extra argument.

Known issues:
* BlendImage does not work if both sprites have alpha channel. This is actually a silly oversight because Alan's code did not provide software blender setup for this variant. There might be a way to make ones (by adjusting present ones?).
* Linear Burn and Dodge are approximations on D3D9/OGL, to work properly they might require writing a shader (In fact, maybe in the end we may switch completely to shaders here).
* (Android) Darken and Lighten don't work because GLES2 lacks GL_MIN/GL_MAX, but maybe GLES3 supports them
(then again, we may switch to shaders later).
* GUIControl seem to be the only remaining drawable object type that does not have BlendMode. This is unfortunately due to how its drawing on parent GUI is currently done in the engine (they are always drawn using simple software paint right to the parent's bitmap). I thought that adding blend operations there will result in suboptimal implementation. So I'd leave this until gui rendering is improved. A workaround for now is to create separate GUI for individual controls, and set their blending mode instead.
* EDIT: Ah, another object which is missing is ... mouse cursor. I am wondering where in API to add such property or function that sets blendmode. For instance, maybe it could be added to a cursor type (where cursor image etc are)? A workaround for now is again to make a gui with blending to work as a cursor.

Other references:
* posted by Alan in his pr: http://blog.onebyonedesign.com/unity/photoshop-blend-modes-for-unity3d/

### DIFF
--- a/Engine/ac/overlay.cpp
+++ b/Engine/ac/overlay.cpp
@@ -151,7 +151,7 @@ void Overlay_SetBlendMode(ScriptOverlay *scover, int blendMode) {
     if (ovri < 0)
         quit("!invalid overlay ID specified");
 
-    screenover[ovri].blendMode = blendMode;
+    screenover[ovri].blendMode = (BlendMode)blendMode;
 }
 
 //=============================================================================
@@ -207,7 +207,7 @@ size_t add_screen_overlay(int x, int y, int type, Bitmap *piccy, bool alphaChann
     return add_screen_overlay(x, y, type, piccy, 0, 0, alphaChannel);
 }
 
-size_t add_screen_overlay(int x, int y, int type, Common::Bitmap *piccy, int pic_offx, int pic_offy, bool alphaChannel, int blendMode)
+size_t add_screen_overlay(int x, int y, int type, Common::Bitmap *piccy, int pic_offx, int pic_offy, bool alphaChannel, BlendMode blendMode)
 {
     if (type==OVER_COMPLETE) is_complete_overlay++;
     if (type==OVER_TEXTMSG) is_text_overlay++;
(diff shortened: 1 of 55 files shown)


Engine/ac/overlay.cpp
```diff
@@ -138,6 +138,24 @@ ScriptOverlay* Overlay_CreateTextual(int x, int y, int width, int font, int colo
...
+
+    return screenover[ovri].blendMode;
+}
+
+void Overlay_SetBlendMode(ScriptOverlay *scover, int blendMode) {
+    int ovri = find_overlay_of_type(scover->overlayId);
+    if (ovri < 0)
+        quit("!invalid overlay ID specified");
+    if ((blendMode < 0) || (blendMode >= kNumBlendModes))
+        quitprintf("!SetBlendMode: invalid blend mode %d, supported modes are %d - %d", 0, kNumBlendModes - 1);
```
AlanDrake, line 154: You're missing blendMode in the variadic parameters, so the error is incorrectly displayed (same for the other instances)

ivan-mogilko: ah shoot.

AlanDrake:
Looks good, to me

PR #1196, opened 2021-02-01
Editor: enforce a rebuild-all when running after upgrading the project

AlanDrake:
Long overdue. When upgrading projects, the next run/compile should always do a "rebuild all". This will save us from bogus error reports, and that strange invalid character exception from the new compiler.

Now that I'm writing this I realize  it would be better to serialize this extra value in case the user closed the project without running, although adding parameters to the xml would be annoying if we want to backport this as a simple quality-of-life improvement.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -100,7 +100,7 @@ namespace AGS.Editor
          * 2: 3.4.0.1    - WorkspaceState section
          * 3: 3.5.0.11
         */
-        public const int LATEST_USER_DATA_XML_VERSION_INDEX = 3;
+        public const int LATEST_USER_DATA_XML_VERSION_INDEX = 4;
         public const string AUDIO_VOX_FILE_NAME = "audio.vox";
 
         private const string USER_DATA_FILE_NAME = GAME_FILE_NAME + USER_DATA_FILE_SUFFIX;
@@ -1235,6 +1235,8 @@ namespace AGS.Editor
 
             Utilities.EnsureStandardSubFoldersExist();
 
+            forceRebuild |= _game.WorkspaceState.RequiresRebuild;
+
             if (PreCompileGame != null)
             {
 				PreCompileGameEventArgs evArgs = new PreCompileGameEventArgs(forceRebuild);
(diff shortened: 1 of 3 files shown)


persn (commented):
Maybe `AGSEditor.CurrentGame` would be a more logical place to put this flag? It's the current game that needs rebuilding no?

Otherwise it looks good

Editor/AGS.Editor/AGSEditor.cs
```diff
@@ -36,6 +36,8 @@ public class AGSEditor : IAGSEditorDirectories, ISourceControlIntegration
         public event ExtraOutputCreationStepHandler ExtraOutputCreationStep;
 		public event GetSourceControlFileListHandler GetSourceControlFileList;
 
+        public bool RequiresRebuild = false;
```
persn, line 39: Property please

```suggestion
        public bool RequiresRebuild { get; set; } = false;
```

Editor/AGS.Editor/AGSEditor.cs
```diff
@@ -1235,6 +1237,8 @@ public CompileMessages CompileGame(bool forceRebuild, bool createMiniExeForDebug
 
             Utilities.EnsureStandardSubFoldersExist();
 
+            if (RequiresRebuild) forceRebuild = true;
```
persn, line 1240: I believe this does the same, alternative way of writing that's more concise. Not very important so feel free to ignore if you can't be bothered

```suggestion
            forceRebuild |= RequiresRebuild;
```

persn (commented):
Forgot some inline comments

ivan-mogilko:
> although adding parameters to the xml would be annoying if we want to backport this as a simple quality-of-life improvement.

We have more than one xml. There's a Game.agf.user which is specifically meant for similar purpose, I believe.
Currently it contains only "last build configuration".

AlanDrake:
Alright, moved it into WorkspaceState with a property.
Seems to be working as intended.

ivan-mogilko:
One thing missing is to advance LATEST_USER_DATA_XML_VERSION_INDEX. That is format index for WorkspaceState (similar to format index for game data).

I propose to make a jump to nearest 10th, just in case to leave several reserved indexes for backward compatible branch.

Issue #1195, opened 2021-01-31
New compiler: minor mistake with forward struct declaration


ivan-mogilko:
@fernewelten 

If you make a forward managed struct declaration, when doing actual declaration later you may omit "managed" keyword and compiler does not care.

<pre>
managed struct SomeType;

import void Func(SomeType* obj);

struct SomeType {
    // contents
};
</pre>
Above works well, and struct is indeed treated as a managed one.

Should this be more strict and demand indentical declaration in each case?

---

Also, in my opinion, the error message for non-managed struct forward declaration is not clear.

E.g. if you have
<pre>
struct SomeType;
</pre>
compiler reports:
`Error: Forward-declared 'struct's must be 'managed'`

I think this message may be occasionally misinterpreted by user, as if they should just add "managed" keyword to make it work.
What about saying "Only managed structs may be forward-declared" instead?

fernewelten:
Hi Ivan,
In my opinion, for the first you've uncovered a bug, and the compiler should require the managed keyword when the forward declaration has been managed. What do you think would be better?

For the second, good point; I'll fix it.

ivan-mogilko:
> In my opinion, for the first you've uncovered a bug, and the compiler should require the managed keyword when the forward declaration has been managed. What do you think would be better?

Yes I also agree, because otherwise this may lead to peoples' confusion when you have same struct declared differently in different places.

fernewelten:
Turns out that the code had a similar problem with respect to 'autoptr' and 'builtin'.

Fixed in #1197

PR #1197, opened 2021-02-02
New compiler: Forward struct declaration and actual declaration must match with respect to 'autoptr', 'builtin', and 'managed'

fernewelten:
This PR fixes the bug reported in issue #1195.

If a forward declaration of a struct has any of the qualifiers 'autoptr', 'builtin', or 'managed' then the actual declaration needs them, too. If OTOH the forward declaration doesn't, then the actual declaration may not have them.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -4715,6 +4715,52 @@ AGS::ErrorType AGS::Parser::HandleEndOfFuncBody(Symbol &struct_of_current_func, 
     return kERR_None;
 }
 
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
This works, but I noticed two new things:

1) For some reason `autoptr` keyword cannot be applied to user struct, demands that struct is builtin. Why is this made a restriction?
BTW, what is the meaning of `autoptr` now if compiler allows to omit `*` with any struct?

2) When I try to use `new` with the builtin struct, it reports following error:
>  Error: Expected '[' after the built-in type 'SomeType'
I think this may be unclear for users, would it be possible to change to stating explicitly that you cannot create instances of builtin type, only dynamic array with pointers?

PR #1200, opened 2021-02-02
New compiler: Disallow local struct or enum type definitions

fernewelten:
Typical code (should yield an error):
```
void Test1() {
    struct MyStruct {
        int a;
    };
}
```


ivan-mogilko:
This works, except I think that word "when" is unnecessary in the error message, "within a function" or "inside a function" would be sufficient.

Same as for another pr, there are some conflicts after merging other fixes, could you rebase again?

PR #1201, opened 2021-02-03
Editor: adjustments for room panning controls

ivan-mogilko:
For #613, adjust panning & zoom controls as suggested in comments.

* added PanButtons property to FreePanControl and use that instead of hardcoding which buttons to use;
* pass another event (on key up) through from the parent EditorContentPanel class to Room editor pane to be able to react to released keys there;
* changed room panning from Ctrl+Shift+LClick to Space+LClick; Middle mouse button also works.
I did not really know how to better organize this. FreePanControl is not guaranteed to have focus, so it's not possible to catch press and release events there. What I did: made RoomSettingsEditor do that instead, and toggle FreePanControl's PanButtons to allow and disallow Left Click as Space is pressed and released.
If it's bad, I can maybe add methods for passing key pressed/release in FreePanControl, but that seemed a bit ugly.
* Changed room context menu call to Shift+RClick. I also tried to implement suggestion from comments and use middle button "when mouse did not move", but it worked very unreliably. Maybe I did not do something right.
* New controls for scroll/zoom:
    - wheel w/o modifiers: scroll vertically;
    - wheel + Shift: scroll horizontally;
    - wheel + Ctrl: zoom

### DIFF
--- a/Editor/AGS.Editor/GUI/TabbedDocumentManager.cs
+++ b/Editor/AGS.Editor/GUI/TabbedDocumentManager.cs
@@ -367,6 +367,10 @@ namespace AGS.Editor
                 }
                 _flipThroughPanesIndex = 0;
             }
+            else if (_currentPane != null)
+            {
+                return _currentPane.Control.KeyReleased(key);
+            }
             return false;
         }
 
--- a/Editor/AGS.Editor/Panes/FreePanControl.cs
+++ b/Editor/AGS.Editor/Panes/FreePanControl.cs
@@ -9,12 +9,30 @@ namespace AGS.Editor
     public class FreePanControl : BufferedPanel
     {
         private const int WS_EX_COMPOSITED = 0x02000000;
+        private MouseButtons _panButtons = MouseButtons.Middle; // which buttons trigger panning
         private int _lastX, _lastY; //used to prevent useless mouse move events
         private Point _panGrabPoint = Point.Empty;
+        private MouseButtons _panGrabButton = MouseButtons.None; // which button was used last
         public event EventHandler PanGrabbed;
         public event EventHandler PanReleased;
         public bool IsPanning { get; private set; }
         public Cursor PanCursor { get; private set; } = Cursors.Hand; // We need a better hand cursor, using the "link" hand for now
+
+        /// <summary>
+        /// Which mouse buttons should be panning the image.
+        /// </summary>
+        public MouseButtons PanButtons
+        {
+            get { return _panButtons; }
+            set
+            {
+                _panButtons = value;
+                if (!_panButtons.HasFlag(_panGrabButton))
+                {
+                    OnPanRelease();
+                }
+            }
+        }
 
         // Enabling Composite makes scrolling smoother
         protected override CreateParams CreateParams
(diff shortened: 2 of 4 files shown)


ericoporto:
Scroll Wheel and Scroll Wheel with shift doesn't work for me.
- Scroll Wheel alone: does nothing
- Scroll Wheel + Shift: Scrolls both horizontal and vertical on diagonal

It would be cool if the zoom was centered on the mouse position when using ctrl+scroll wheel with the mouse on the room - Google Maps does this. I don't know the math to do it.

The build check that failed was a bug in Cirrus CI, it should have been fixed by now.

ivan-mogilko:
They work, but apparently something breaks as soon as you click away, and if you click on a room it starts working again...

ericoporto:
OK, I figured. My Windows 10 has the feature "_Scroll Inactive Windows when I hover them_" activated by default. If I disable it, then the scroll wheel works with the said caveats, but it works. But with this feature enabled, it breaks.

ivan-mogilko:
Actually, I found under debugger that even in normal circumstances simple mouse wheel (without modifiers) does not cause a room panel to get event, so it might be default behavior that scrolls it vertically. I'd need to look more into this.

ivan-mogilko:
Oh. Apparently, mouse wheel sends key events "up" and "down" (eq to arrow keys), and this is how room gets scrolled by default.

Is this how .NET works, or there is a setting that controls that in a form or application?

This makes things more complicated. Especially considering if an object is selected then mouse wheel will actually move it up and down, which I never paid attention to before...

ivan-mogilko:
@ericoporto 
I think I fixed these wheel behavior, these were combination of two problems: certain controls intercepting mouse wheel events when they have focus (namely background combo box and buffered panel itself), and my new wheel handling did applied reverse vertical scroll.

ivan-mogilko:
> It would be cool if the zoom was centered on the mouse position when using ctrl+scroll wheel with the mouse on the room - Google Maps does this. I don't know the math to do it.

The math is a second problem, the first problem is to define the logic...

So, after experimenting a minute with google maps, I think the rule is that *the point under cursor must remain under cursor (at the same position) after zoom*.

ericoporto:
Hey it works @ivan-mogilko ! Even with the "Scroll Inactive Windows when I hover them" setting on in Windows 10!!!

ivan-mogilko:
> Even with the "Scroll Inactive Windows when I hover them" setting on in Windows 10!!!

Oh, this is good to know, I had no idea if this will still be a problem.

I'm currently working on the zooming math, I think the difficult part was to guess how it should work, because I could not figure out until watched google map example very closely...

ivan-mogilko:
Added zooming into mouse cursor.

ericoporto:
Sorry, I got an error, do you know what it means?

image
<details>
Error: Value of '370' is not valid for 'Value'. 'Value' should be between 'minimum' and 'maximum'.
Parameter name: Value
Version: AGS 3.99.99.0

System.ArgumentOutOfRangeException: Value of '370' is not valid for 'Value'. 'Value' should be between 'minimum' and 'maximum'.
Parameter name: Value
   at System.Windows.Forms.ScrollProperties.set_Value(Int32 value)
   at AGS.Editor.RoomSettingsEditor.ChangeZoom(Int32 zoom, Point anchor)
   at AGS.Editor.RoomSettingsEditor.sldZoomLevel_Scroll(Object sender, EventArgs e)
   at AGS.Editor.RoomSettingsEditor.SetZoomSliderToMultiplier(Int32 factor)
   at AGS.Editor.RoomSettingsEditor..ctor(Room room)
   at AGS.Editor.Components.RoomsComponent.CreateRoomSettings(DockData previousDockData)
   at AGS.Editor.Components.RoomsComponent.LoadDifferentRoom(UnloadedRoom newRoom)
   at AGS.Editor.Components.RoomsComponent.LoadRoom(String controlID)
   at AGS.Editor.Components.RoomsComponent.ItemCommandClick(String controlID)
   at AGS.Editor.Components.BaseComponentWithFolders`2.CommandClick(String controlID)
   at AGS.Editor.ProjectTree.ProcessClickOnNode(String nodeID, MouseButtons button)
   at AGS.Editor.ProjectTree.projectTree_NodeMouseDoubleClick(Object sender, TreeNodeMouseClickEventArgs e)
   at System.Windows.Forms.TreeView.OnNodeMouseDoubleClick(TreeNodeMouseClickEventArgs e)
   at System.Windows.Forms.TreeView.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
</details>

I can't load the Room Editor anymore. I tried creating a new game from the Sierra Template and the error persisted.

ivan-mogilko:
Ah yes I can reproduce this when creating a new blank room.

ivan-mogilko:
@ericoporto updated with a fix.

ericoporto:
works beautifully! Can't find any bugs.

Issue #1092, opened 2020-06-17
Separate Windows Forms from AGS Types


ericoporto:
There are some files in AGS Types that use `System.Windows`.  They are below.

```
EditorFeatures/MenuCommand.cs  // uses only to describe shortcut keys
EditorFeatures/EditorContentPanel.cs // actual UserControl, no way to separate!
GUIControl.cs                       // uses for Clipboard
Interfaces/IScriptEditorControl.cs     // Has one property using Control, hard to separate
Interfaces/Docking/IDockingContainer.cs // uses for a property of type `ContextMenuStrip`
PropertyGridExtras/PropertyTabEvents.cs // a Forms.Design.PropertyTab, no way to separate!
PropertyGridExtras/SpriteFileNameEditor.cs // a FileNameEditor, no way to separate here
PropertyGridExtras/BuildTargetUIEditor.cs // ItemCheckEvent everywhere, no way to separate!
PropertyGridExtras/StringListUIEditorControl.Designer.cs // is a designer!
PropertyGridExtras/StringListUIEditor.cs // uses IWindowsFormsEditorService, hard to separate!
PropertyGridExtras/StringListUIEditorControl.cs // it's a WindowsFormControl!
PropertyGridExtras/PropertyTabInteractions.cs // a Forms.Design.PropertyTab, no way to separate!
```

If there's no dependency on `System.Windows`, AGS.Types can be used directly in the command line tools too, in non Windows OSes!

The reason behind this, is to enable usage of Command Line tools to build ags games.

I believe some of it is due to how `Game.agf` has both information on the Editor configurations and the Game project itself. Alternatively, if this was separated (let's say `Game.agf` and `Game.user.agf`) then there would be two AGS Types thing, one for the Project and the other for the Editor panels and configs not belonging to the project. This task is quite big so decided to open an issue to see what the situation here. Note there is no need to separate or change much if there's a way to remove `System.Windows` from above.

The following files have `using System.Windows` in them but not actually use it, probably just there from copy pasting, and could be removed now.
```
BaseFolderCollection.cs 
PropertyGridExtras/ViewUIEditor.cs
EditorFeatures/ContentDocument.cs 
Interfaces/IEditorComponent.cs
PropertyGridExtras/ScriptFunctionUIEditor.cs
PropertyGridExtras/SpriteSelectUIEditor.cs 
PropertyGridExtras/CustomPropertiesUIEditor.cs 
[...]

PR #1204, opened 2021-02-09
(Test) AGS 4: use standard ColorDialog as a color picker for properties

ivan-mogilko:
This is frankly just a quick experiment, and I don't know if it looks good and convenient to use. It seems working though.

Introduced UITypeEditor wrapper over classic ColorDialog. Set it as an editor for each game property of type Color. This replaces default color picker called by property grid, which is not good for AGS Editor at all.

Caveats:
* ColorDialog is from System.Windows.Forms, if that matters. EDIT: made the UITypeEditor class in AGS.Types to use delegate instead of directly referencing Windows.Forms namespace.
* ColorDialog is modal and appears at the center of the screen, as opposed to default non-modal color picker that was displayed right next to the property.
* I made it save "custom colors" array in a static variable, so it will persist during editor session, but it's not serialized in project files.

PS. On second thought, I myself don't like this variant very much either. It might be better to find a proper color picker control with a RGB circle or similar, which would be compatible with UITypeEditor or already implemented as one.

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -66,6 +66,8 @@ namespace AGS.Editor
 
         private string _timerScriptName;
         private string _timerSearchForText;
+        // Custom color table for the ColorDialog
+        private int[] _customColors = null;
 
         private GUIController()
         {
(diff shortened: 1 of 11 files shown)


persn:
> ColorDialog is from System.Windows.Forms, if that matters.

@ericoporto is trying to get rid of `System.Windows.Forms` references directly from AGS.Types #1092 so this may potentially cause extra work work for him.

ivan-mogilko:
It may be possible to call actual implemetation through a delegate, as done in another editor: 
https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Types/PropertyGridExtras/CustomResolutionUIEditor.cs#L10

persn:
That sounds good

ivan-mogilko:
I was thinking, even if it's not ideal, this code change is quite small. Then if someone makes a better dialog they would just replace it in this delegate function.

ericoporto:
This new color picker is easier to use and easier to replace! The old one was not good, and required going to the color panel to find a good color - playing around with the sliders - and then copy pasting.

ivan-mogilko:
Ah, I did not realize the color sample was missing, added your code now.

ivan-mogilko:
Ok, it's fine if I merge it then?

ericoporto:
I am fine with merging, it's a good change!

PR #1206, opened 2021-02-13
Script API: implement Overlay.ZOrder

### LINKED ISSUE #1202: AGS 4: Add Overlay.ZOrder property
Following might be not hard to implement, but I was curious to hear other opinions, and see if maybe I am missing anything.

### Brief recap

Overlays are primitive dynamically created game objects that contain an image, has a position, and don't have any behavior on its own (may be manipulated by the engine states though). Image may be either taken from a pre-existing sprite, or generated as a block of text (aka "textual overlays"). Position may be referring either screen or room coordinates; latter used mainly to display speech over character's heads; but overlays are always drawn on "gui layer", above the room. User-created overlays are always screen-related.

Overlays have two Z positions on screen: behind and above GUI. This choice is hard-coded:
* custom overlays (user-created, background speech) are displayed behind GUI;
* speech overlays (text & portraits) are displayed above GUI (I think this was made so there could be character speech describing inventory items, and similar purpose).
* overlays are not sorted among themselves, their drawing order is matching their order of creation (last created is drawn above all).

There are two big limitations with overlays now:
* User cannot z-order them among themselves or other game objects.
* They cannot be put inside the room (and because they have no z-order they won't be as useful there).

This ticket is about the first problem, the second is best to be dealt afterwards as its own task, because it has its own nuances.

### Proposal

1. Add integer ZOrder property with meaning similar to GUI.ZOrder.
2. GUI overlays are *sorted among the GUIs* according to [same rules](https://github.com/adventuregamestudio/ags-manual/wiki/GUI#guizorder).

The sorting is ascending, those with lower z-order are drawn behind and with higher z-order above (in accordance to existing rules).

*Special overlays and backward compatibility*

To maintain existing behavior special overlays (ones created by the game's own logic) must have z-order to either appear behind or above all GUI. This could be done either by setting very large hard-coded values (+/-), or by adjusting their z-order everytime a GUI's one is set. The first solution obviously may not work if user set even larger ZOrder for GUI, but the second probably would make sense only for backward compatible mode (if we have one in ags4), because in the new games users may want to rearrange these special overlays too.

Speaking of which, we may also add properties which define default z-order for special overlays, because it may be inconvenient or impossible for user to set one for these overlays right after their creation.
For example, Speech.DefaultZOrder property could be used to set default speech and portrait z-order, and so forth.

ivan-mogilko:
Resolves #1202.

This implements Overlay.ZOrder property in compliance with other similar properties from GUI and GUIControl.

Overlays are now z-sorted along with the GUI on screen.

To keep default and backward compatible behavior, newly created overlays have their zorder set:
* For custom overlays and QF4 speech style overlays: to INT_MIN, so that they are beyond any GUI.
* For the text and portraits overlays: to INT_MAX, so that they are above any GUI.

User may adjust Overlay.ZOrder anytime after their creation, so long as there's an access to overlay object in script.

What to note on potential problems?

It must be kept in mind that overlays are not interactable, therefore mouse always "clicks through" them and activates whatever is behind. Just something to remember or maybe write in the manual, to prevent user confusion.

Some of the special overlays still cannot be referenced in the script, which imo is a separate issue that may be resolved by future changes.

I'd mention two good candidates for improvement here:
* An abiility to get actual Overlay objects of speech. I had a [PR in the past](#1127), but closed it being not very happy with it. I could restore it though, in hopes that overlays become easier to manipulate in the future.
* An ability to set default z-order for these overlays, so that user could set them once per game instead of having to catch overlay creation somehow and override z-order.

Is there anything I'm missing?

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -942,6 +942,8 @@ builtin managed struct Overlay {
   import attribute BlendMode BlendMode;
   /// Gets/sets the transparency of this overlay.
   import attribute int Transparency;
+  /// Gets/sets the overlay's z-order relative to other overlays and on-screen objects.
+  import attribute int ZOrder;
 #endif
 };
 
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -158,7 +158,7 @@ struct SpriteListEntry
     IDriverDependantBitmap *bmp = nullptr;
     int transparent = 0;
     int x = 0, y = 0;
-    int baseline = 0;
+    int zorder = 0;
     // Tells if this item should take priority during sort if z1 == z2
     // TODO: this is some compatibility feature - find out if may be omited and done without extra struct?
     bool takesPriorityIfEqual = false;
@@ -793,7 +793,7 @@ static void clear_sprite_list()
     sprlist.clear();
 }
 
-static void add_to_sprite_list(IDriverDependantBitmap* spp, int xx, int yy, int baseline, int trans, bool isWalkBehind,
+static void add_to_sprite_list(IDriverDependantBitmap* spp, int xx, int yy, int zorder, bool isWalkBehind, int trans = 0,
     BlendMode blendMode = kBlend_Normal)
 {
     if (spp == nullptr)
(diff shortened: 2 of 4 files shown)


ivan-mogilko:
Well, since no one sais anything, I guess it's okay. It's a simple enough change too but it extends an object logic. Of course there's always a chance some problem no one thought about will sprang out...

PR #1212, opened 2021-02-21
Script API: add access to blocking speech overlays and latest skip reason

ivan-mogilko:
This is a slightly different remake of #1127.

This PR exposes couple of overlays that were already internal objects but could not be accessed from script in any way, plus adds couple of complementary things.

### Speech.TextOverlay and Speech.PortraitOverlay
`readonly attribute Overlay *Speech.TextOverlay` returns special overlay created for the blocking speech, or null if no such text is on screen.
`readonly attribute Overlay *Game.PortraitOverlay` returns special overlay created for the speech portrait, or null if no such portrait is on screen.

Because of how blocking speech works, these overlays will be only available in `repeatedly_execute_always` and `late_repeatedly_execute_always` script callbacks.
But these are valid overlays in every other aspect: you can use its properties, read or change position, for example.
Calling Remove() function is also legal, and will correctly stop the current blocking speech (the engine was already programmed to rely strictly on overlay's presence).

When the speech is removed, script object is detached from actual overlay and invalidated: its Overlay.Valid property will be returning false, and attempt to call any other property or function will result in script error. This is essential to know if you store Overlay pointer in a variable for later use.

**On practical use**

To clarify, at the moment there's may not much uses of changing their properties per se, but that may be rather a limitation of overlay struct.<br>
The most significant is simply the capability to read these overlay pointers and thus *detect their presence*. Or store them in a variable and later compare with previously saved pointers and thus *detect when the speech changes*. For every speech line there's a new overlay generated by the engine. In practice this gives user an ability to know when speech appeared, changed to next line, or ended - something that was previously not possible without writing wrapper functions over Say command.
Not going to say if this is better or more convenient compared to having your custom Say functions, but I believe this is still a reasonable thing to have, as these are existing game objects. Exposing them will give an opportunity to add some quick handling of built-in speech without writing fully custom speech system.

As noted above, this also adds a completely new method of skipping built-in speech: by calling its overlay's Remove() function.

Which brings us to...

### eSkipNone

`eSkipNone` option added for SkipSpeechStyle. When set, speech may only be skipped by a script command (i.e. - calling Speech.TextOverlay.Remove()).<br>
This complements similar mode added for Wait and cutscenes (see #1084).

### Game.BlockingWaitSkipped

`readonly attribute int Game.BlockingWaitSkipped` returns a code telling how the last blocking wait action was ended. This code is equivalent to the return value of the Wait* functions, introduced previously (#1084). At the moment this property is updated by three sorts of skippable blocking actions: Say*, Display* and Wait*.

The purpose of such property is to be able to write a generic code for detecting last skip reason - without wrapping over every blocking Say etc call. One could e.g. check that in late_repeatedly_execute_always to see if anything was skipped in game last frame.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2292,6 +2292,8 @@ builtin struct Speech {
 #ifdef SCRIPT_API_v399
   /// Gets the overlay representing displayed blocking text, or null if no such text none is displayed at the moment.
   import static readonly attribute Overlay* TextOverlay;
+  /// Gets the overlay representing displayed portrait, or null if it is not displayed at the moment.
+  import static readonly attribute Overlay* PortraitOverlay;
 #endif
 };
 
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2157,6 +2157,10 @@ builtin struct Game {
   /// Gets the number of cameras.
   import static readonly attribute int CameraCount;
 #endif
+#ifdef SCRIPT_API_v399
+  /// Gets the code which describes how was the last blocking state skipped by a user (or autotimer).
+  import static readonly attribute int BlockingWaitSkipped;
+#endif
 };
 
 builtin struct GameState {
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2289,6 +2289,10 @@ builtin struct Speech {
   import static attribute bool            UseGlobalSpeechAnimationDelay;
   /// Gets/sets whether voice and/or text are used in the game.
   import static attribute eVoiceMode      VoiceMode;
+#ifdef SCRIPT_API_v399
+  /// Gets the overlay representing displayed blocking text, or null if no such text none is displayed at the moment.
+  import static readonly attribute Overlay* TextOverlay;
+#endif
 };
 
 #ifdef SCRIPT_API_v3507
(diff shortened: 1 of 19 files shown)


ericoporto:
If overlays gain a scalling property that allows scaling down (like characters can have lower resolution than game), then this could be used to improve resolution of game speech text.

ivan-mogilko:
> If overlays gain a scalling property that allows scaling down (like characters can have lower resolution than game), then this could be used to improve resolution of game speech text. (Sometimes pixely text is a bit hard to read).

No, this is not related, the text must be rendered using different resolution for this, not just scaled with overlay or other game object.

ericoporto:
Ah, yes, I thinked wrong about it. Yes! I used the character solution before because it's easy to carry it around! XD then used the scaling option on winsetup that alows this. But now thinking about I may skip this by just making a high resolution game and using cameras to "zoom" the game. :)

ivan-mogilko:
BTW I haven't thought what will happen if you remove portrait overlay instead.

ivan-mogilko:
> I like to modify the screen showing GUIs when a character is saying something, right now I loop through all characters to check if someone is saying something, if I understood this gives me a single place to check for this, which is very cool.

Yes. This whole thing was originally done for a user who wanted more precise control over how input skips speech. So with this he was able to detect whether the speech timed out on its own, and other similar cases.

ericoporto:
> BTW I haven't thought what will happen if you remove portrait overlay instead

If the portrait is animated, does it remove and creates new overlays as they get animated?

ivan-mogilko:
Tested removing portrait and it leads to a hilarous bug when portrait overlay dissapears but portrait continues it animate inside the text overlay.

ivan-mogilko:
I made a mistake by not thinking this through beforehand, but on second thought portrait right now is a part of built-in speech "object" with its own behavior, so made removing it has same effect as removing text overlay.

Issue #830, opened 2019-06-11
Script API: GetTranslation returns non-const string type


ghost:
GetTranslation() function accepts and returns old **string** type instead of managed **String**.

These "strings" declared in script has a limit of 200 chars, but this is not a case here, and they are rather treated as "char*" pointer that points to *some* memory (may be internal engine's memory).

So the char limit may not be an issue here, but the safety is. Because, if you enable old string API, you can legitimately do this:

<pre>
StrCat( GetTranslation("Some text"), "random text" );
</pre>

This chunk of code actually writes into *some memory*. It could be memory allocated by the translation dictionary, or, if there's no translation ready, it could be string literal allocated by the script (storing "Some text").

**EDIT:**
Perhaps one of the following things may be done here:
1) Change the return type to "const string", assuming the script compiler can deal with that.
2) Alter existing implementation to return a copy of a string stored in special buffer of at least 200 chars wide (an official limit of **string** type) which would at least make return value safer. But problem with that is that a) we cannot have multiple buffers as these will be non-reference counted and there's no way to keep track of the ones we return from engine into the script, b) if we return same buffer everytime, may there be a chance that script stored previous return value in a const string variable? does ags script support that in user script?
3) Provide new variant of this function that returns managed String. Unsafe version will be kept for backward compatibility.

ivan-mogilko:
Hmm, maybe adding "const" qualifier to return value would be sufficient. Assuming it works with ags script ofcourse.

ivan-mogilko:
> Hmm, maybe adding "const" qualifier to return value would be sufficient. Assuming it works with ags script ofcourse.

That would be most desirable solution, but unfortunately existing compiler is unable to correctly parse "const string" return type.
Thus either one would need to fix old compiler (I don't think I'll want to do this), or leave this to use with the new compiler; in turn that means that this will either be only fixed in ags4, or new compiler would have to be backported to ags3.

ivan-mogilko:
Resolved in ags4. This is impossible to fix in optimal way in ags3 unless either old compiler is fixed, or new compiler is backported.

rofl0r:
> old compiler is fixed

what would be necessary for that, simply correct parsing of const string as return type, or also setting some attributes for the engine ?

ivan-mogilko:
> what would be necessary for that, simply correct parsing of const string as return type, or also setting some attributes for the engine ?

Only parsing, as engine does not distinct and does not care about const and nonconst values.

[...]

PR #1215, opened 2021-02-22
script compiler: implement parsing of const function return type

rofl0r:
the following now works:

    import const string get_version();
    const string baz(const string b) {
        return b;
    }
    const string bar() {
	return "bar";
    }

this should be sufficient to resolve AGS issue 830

https://github.com/adventuregamestudio/ags/issues/830

using const string variables doesn't work still:

   const string muh = "muh";

-> cannot assign value to string, use StrCopy


ivan-mogilko:
Thank you, this seems to be working fine.

Issue #1219, opened 2021-02-24
New compiler: corrupted XML messages while running from editor


AlanDrake:
@fernewelten 
I'm experiencing random Editor exceptions from badly formatted XML messages coming from the Engine, when using the new compiler.
They look like there's some corruption happening randomly in the portion regarding script names:

Broken  examples:
```
<?xml version="1.0" encoding="Windows-1252"?><Debugger Command="EXIT">\n<EngineWindow>5445382</EngineWindow>   <ScriptState><![CDATA[in "", line 215
from "", line 330\n]]></ScriptState> </Debugger>
```
```
<?xml version="1.0" encoding="Windows-1252"?><Debugger Command="EXIT">  <EngineWindow>1314858</EngineWindow>   <ScriptState><![CDATA[in "ModalDialog.asc", line 58\nfrom "çé\nˆèé\n¸ëé\nhíé\n(îé\nàà\n"b…nøæ", line 136\n]]></ScriptState> </Debugger>
```

What the latter looks like when not sending broken XML:
```
"<?xml version=\"1.0\" encoding=\"Windows-1252\"?><Debugger Command=\"EXIT\">  <EngineWindow>1250534</EngineWindow>   <ScriptState><![CDATA[in \"ModalDialog.asc\", line 58\nfrom \"GlobalScript.asc\", line 136\n]]></ScriptState> </Debugger>"
```

It happens rarely, and it's the same kind of error I was experiencing from running upgraded projects without a full rebuild.
I suspect there's something causing memory corruption of the area containing script name tokens, but I'm not sure how to investigate further.

ivan-mogilko:
Well, first of all we'd need to find whether names are corrupted by compiler and saved like that, or they are corrupted within engine during runtime.

fernewelten:
On a hunch, possibly, the script name might sometimes be unset in some way for the **very first** line number metacode that is emitted after the script name changes. OTOH, the line numbers in your examples seem to be too high for that.

I'll dig around. Unfortunately, intermittent bugs are hard to hunt.

fernewelten:
I might have found the bug. That code wasn't as easy to verify at a glance as I would have liked. So I reformulated the passage to make it clearer.  There were similar passages in the vicinity, making it likely that similar bugs could hide there, too. So I took the occasion to also re-formulate those passages that were similar to the buggy one.

ivan-mogilko:
[...]

PR #1220, opened 2021-02-25
New compiler: Memory allocation bug when the section changes

fernewelten:
This addresses #1219.

In some rare cases, memory allocation for section names went awry. The passage cc_compiledscript.cpp#L236 and following was probably responsible for the bug, it reserves space proportionate to `codesize` (the size of the generated Bytecode) instead of the number of sections. This would fail when very little code is generated for a lot of sections, which possible but rarely the case. So this would explain why the bug only happens rarely. 

I reformulated that passage and the similar passages of that file, introducing some named constants, to make it easy to verify at a glance whether they are set correctly. I rewrote ´ResizeMemory()` to calculate with array elements instead of bytes and renamed some variables to (hopefully) make it clearer what they mean.

cc_compiledscript.cpp#L78 : The assignment `EmitLineNumbers = emit_line_numbers;` is duplicate.

### DIFF
--- a/Compiler/script2/cc_compiledscript.cpp
+++ b/Compiler/script2/cc_compiledscript.cpp
@@ -23,26 +23,26 @@ void AGS::ccCompiledScript::PopReg(CodeCell regg)
     OffsetToLocalVarBlock -= SIZE_OF_STACK_CELL;
 }
 
-ErrorType AGS::ccCompiledScript::ResizeMemory(size_t const needed, size_t const min_size, void *&start, size_t &allocated)
+ErrorType AGS::ccCompiledScript::ResizeMemory(size_t const el_size, size_t const needed, size_t const minimum, void *&chunk, size_t &allocated)
 {
     if (allocated >= needed)
         return kERR_None; // nothing to be done
 
-    size_t new_size = (allocated < min_size) ? min_size : allocated;
-    while (new_size < needed)
-        new_size += new_size / 2;
+    size_t new_element_number = (allocated < minimum) ? minimum : allocated;
+    while (new_element_number < needed)
+        new_element_number += new_element_number / 2;
 
-    void *new_start = realloc(start, new_size);
-    if (!new_start)
+    void *new_chunk = realloc(chunk, el_size * new_element_number);
+    if (!new_chunk)
     {
         // Note, according to the STL, the old block is NOT freed if allocation fails
-        free(start);
-        start = nullptr;
+        free(chunk);
+        chunk = nullptr;
         return kERR_InternalError;
     }
 
-    start = new_start;
-    allocated = new_size;
+    chunk = new_chunk;
+    allocated = new_element_number;
     return kERR_None;
 }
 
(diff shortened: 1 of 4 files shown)


AlanDrake:
I'll give this fix a run

Compiler/script2/cc_compiledscript.cpp
```diff
@@ -84,19 +82,22 @@ AGS::ccCompiledScript::~ccCompiledScript()
...
         return globaldatasize; // nothing to do
 
     // The new global variable will be moved to &(globaldata[offset])
     size_t const offset = (globaldatasize < 0) ? 0u : static_cast<size_t>(globaldatasize);
 
-    void *chunk_start = globaldata;
+    void *globaldata_chunk = globaldata;
+    size_t const el_size = sizeof(globaldata[0u]); // size of 1 element, i.e. 1 char
+    size_t const needed = offset + value_size;
+    size_t const minimum_allocated = 100u;
```
rofl0r, line 96: what's the purpose of these hardcoded numbers ? shouldnt we allocate just as many bytes as we need ?

ivan-mogilko:
TBH I find other thing strange. Never looked much into the new compiler code, but I had an understanding that it's written in C++; why are there manual memory (re)allocations, and not e.g. std::vectors?

AlanDrake:
After merging this fix locally, I haven't experienced the exception even once.
Before I used to trigger the issue once every 1-2 hours while re-launching the game every 3-4 minutes to test small changes.

I think if everything keeps working for two more days then it's safe to say it was fixed.

fernewelten:
For context: The old code did this in essentially the same way, but wrongly -- it would leave a memory leak in the case of an unsuccessful reallocation. Static analysis kept nagging at at that code, so after quite some time I ended up fixing the d*mn thing to make the nagging go away. I tried to contain the ugliness by moving it into a dedicated function. But then of course, I managed to put a bug into the bug fix by calling that function with a wrong parameter in one place. Sorry about that.

The class (struct) `ccScript` is used to interface the Engine. Ivan's right, it would be **much** cleaner to use  `std::vector`s to pass all that data; memory deallocation would be handled automatically as soon as the object is discarded. But this code is on the periphery and interfaces other code. Someone somewhere is going to `free` those blocks of memory later on, so I don't have much alternative to `malloc`ing them here.

As for the magic numbers, the basic logic for that was already in the old code: The arrays concerned basically grow one entry as a time. For instance, the `fixups` array grows one fixup as a time. It would be inefficient to constantly `realloc` whenever a fixup is added, so instead, the array is started with a number of entries, and whenever the current capacity is reached, the array is extended by some multiplicative factor (I think I put in 10 entries and a factor of 1.5). The specific values aren't that important. If they happen to be a bit too small then there'll be another reallocation or two. Since each extension is a factor larger than the previous extension, the number of allocations grows logarithmic to the size of the array, which is low.  This can't waste more memory than the initial allocation or half the needed entries, whichever is larger.

ivan-mogilko:
> The class (struct) ccScript is used to interface the Engine. Ivan's right, it would be much cleaner to use std::vectors to pass all that data; memory deallocation would be handled automatically as soon as the object is discarded. But this code is on the periphery and interfaces other code. Someone somewhere is going to free those blocks of memory later on, so I don't have much alternative to mallocing them here.

I really believe this should be rewritten into something more compliant with C++. ccScript could be split if necessary, it's a rather simple struct so code duplication would be minimal. Engine could later made use the new struct with std::vectors as well. Then only old compiler will keep using old struct (so long as we keep old compiler as a backup that is).

But I guess this is going to be separate task for another time.

PR #1221, opened 2021-02-25
Room serialization

persn:
Related to #469 

Room entities was lacking serialization handling so I added it, with test cases as proof of concept since no actual uses this yet. They will be used in the next PR related to converting the rooms to open format, but this was big enough change I felt it warranted an isolated PR.

The QA I did here was to make sure that the serialization and deserialization handled all the properties that being read/written in `load_crm_file(...)` and `save_crm_file(...)` from `agsnative.cpp`, and I wrote tests that handles these properties from xml. Anything left out can probably be added in the next PR.

I realize there's an ongoing discussion about using yaml instead, but we already have lots of convenient code for handling xml and I feel like adding yaml and handling it for ags should be a separate issue that maybe covers more than the room format. For now I use xml, we always change it later.

### DIFF
--- a/Editor/AGS.Editor.Tests/Types/RoomHotspotTests.cs
+++ b/Editor/AGS.Editor.Tests/Types/RoomHotspotTests.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Drawing;
+using System.IO;
+using System.Xml;
 using AGS.Types;
 using NSubstitute;
 using NUnit.Framework;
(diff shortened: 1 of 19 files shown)


ivan-mogilko:
Question: ID property marked as [AGSNoSerialize] for one types of objects but not the others, is this intentional or oversight?
And for these objects that do not save ID, does it mean that their ID will be defined by their node order in xml?

persn:
> ID property marked as [AGSNoSerialize] for one types of objects but not the others, is this intentional or oversight?

Oversight, I fixed now

> And for these objects that do not save ID, does it mean that their ID will be defined by their "physical" node order in xml?

Yes

ivan-mogilko:
Don't know if this matters, but RoomObject also dont have [AGSNoSerialize] on ID.

> > And for these objects that do not save ID, does it mean that their ID will be defined by their "physical" node order in xml?

> Yes

What is the benefit from that?

ivan-mogilko:
Something I'd like to add for the reference. Originally most items in the project tree were fixed by the order of their creation (and therefore their ID). However in AGS 3.2.2 tzachs added subfolder support, and ability to freely drag items withing their list. Dragging does not change ID. This created a certain discrepancy between order of these items in the tree and their order in the game (based on ID property). Game objects may be currently accessed through arrays in script, where they are arranged according to these IDs.

I just tested again to be sure, and dragging items around makes them saved in different order in XML, but IDs stay correct. This may result in items arranged not in the order of their IDs inside XML.

Reliance on numeric ID and order in the list may be seen as a bad or unsafe thing by some, and maybe this needs to be changed and made more consistent somehow, but while such behavior exists, I feel uneasy seeing as room objects are saved without ones. If for example there will be similar lists of room items in the editor, won't it cause problems?

persn:
> What is the benefit or purpose of this approach?

I think we need to keep in mind that when we open up the format and start providing standalone CLIs for compiling we will have to support more use cases where game developers are working from their favorite text editing tool (VS Code, Atom, etc.) and are manipulating data by hand. As such having as little data to work with as possible will make things generally more pleasant, less to read, less to write. Additionally I believe that not having more data on file consumes less space (although in this case probably marginally), and reduces chances of annoying merge conflicts when working with version control.

> but would the format be safer with ID present?

I can't think of a reason why. If anything having the ID opens up headache use cases where users accidentally set multiple entities to the same ID, or sets it to an ID which is outside of the many limits that the room code operates with. `MAX_OBJECTS`, `MAX_HOTSPOTS`, `MAX_WALKABLE_AREAS`, etc.

> but this means inconsistency with other game objects which serialize their id

Not sure if that matters. Also I can argue that this way of handling the ID is consistent for how it's already handled in the `agsnative.cpp`

```
	for (size_t i = 0; i < thisroom.ObjectCount; ++i) 
	{
		RoomObject ^obj = gcnew RoomObject(room);
		obj->ID = i;
        ...
	for (size_t i = 0; i < thisroom.HotspotCount; ++i) 
	{
		RoomHotspot ^hotspot = room->Hotspots[i];
		hotspot->ID = i;
        etc...
```

> I just tested again to be sure, and dragging items around in project tree 

I'm aware of this, but the room sub entities aren't displayed in the project tree. They're presented as a flat list in the room editor breadcrumb menu.

>  If for example there will be similar lists of room items in the editor (where these e.g. may be freely rearranged in subfolders), won't it cause problems?

No because the room number is still being serialized as the unit test shows, specifically this is being serialized because of the reasons you already mentioned with the ordering in the folders, but also because the room ID can actually be changed from the editor.

It's the sub entities of the room which are presented in the breadcrumb crumb which doesn't serialize their IDs, and those properties are readonly from the editor side. In my mind it doesn't make sense to disallow the user writing to ID from the editor, and then they can just go to the disk and overwrite it manually. I think that if we change our mind and decide that Walkable areas and such should have read/write IDs, we can introduce the field for serializing then.

ivan-mogilko:
> > If for example there will be similar lists of room items in the editor (where these e.g. may be freely rearranged in subfolders), won't it cause problems?

> No because the room number is still being serialized as the unit test shows, specifically this is being serialized because of the reasons you already mentioned with the ordering in the folders, but also because the room ID can actually be changed from the editor.

Sorry, I phrased this incorrectly, by "room items" there I meant room objects.

persn:
> Sorry, I phrased this incorrectly, by "room items" there I meant room objects, not rooms.

Ah that changes the context a little. My desired outcome is that if we go the route of wanting to organize the sub-entities/room-objcets/whater-you-call-them in folders and whatever else we can introduce the ID serialization then I think?

ivan-mogilko:
> Ah that changes the context a little. My desired outcome is that if we go the route of wanting to organize the sub-entities/room-objcets/whater-you-call-them in folders and whatever else we can introduce the ID serialization then I think?

Yes, after you mentioned that I guess that's true, because so far the editor's UI was the main reason there was this issue. I cannot think of a reason why anyone editing xml by hand would intentionally place objects in a "wrong" order.

But overall this is a complicated situation, and I wish we could find a way to make object's arrangement at design and runtime consistent in the future.

persn:
> But overall this is a complicated situation, and I wish we could find a way to make object's arrangement at design and runtime consistent in the future.

Yeah I agree

PR #1229, opened 2021-03-01
AGS 4: Object Rotation

ivan-mogilko:
**WARNING:** early-WIP, don't merge! for testing, review and suggestions.

This adds support for free game object graphic rotations. Internally Direct3D and OpenGL graphic drivers use tranforms to rotate a texture thus rotation will be hardware accelerated, just like scaling is now.

In script API `Rotation` property is added to following objects: Character, GUI, Object, Overlay, and Camera. (Yes, you may rotate camera too, which naturally summs with individual object rotations)
Rotation is `float` and is set in *degrees*, because imo this will be easier for majority of users. It may be converted to radians internally whenever necessary (e.g. Direct3D expects one in rads, but OpenGL uses degrees).
Images are rotated strictly around their center (also see todo below).

**IMPORTANT:** this is *strictly graphic rotation*, so to speak. It does not change, for example, where Character's point of connection with walkable areas is. It also does not affect collision box (which will still be aligned with room axes).

**NOTE:** I did not add this to GUIControl, because as of now these are drawn directly on GUI bitmap (this was also a reason I decided not to add BlendMode to them, and maybe why @AlanDrake did not too). I imagine it may be better to revamp controls drawing before supporting more visual styles for them. As with blendmodes it's possible to workaround this by creating separate GUI with single control on it and rotate that gui.

Visually, this already works, and for objects that don't require anything beyond (i.e. Overlays) this may be used already.
Interaction *does not work* yet (see TODO below). Clicking will work as if objects were not rotated.

<details> <summary>Screenshots:</summary>

rotated

rotated2
</details>

### TODO

1. Interaction, or rather hit detection is the major problem to work on next. My plan is to add an internal struct with 2 parameters: bounding box (AABB) and transformation matrix, and assign it to all interactable objects. Bounding box will be used for non-pixel perfect hit detection and just to know approximate region the object occupies on screen, while matrix used for pixel-perfect detection, as we will need to rotate hit coordinates. Viewport-camera pairs will also need these, as camera may be rotated.
2. Software renderer does not have sprites rotating implemented yet (it will be simply ignored). I'd leave it until the end because it may require special approach.
3. Minor problems, such as property value range and validation.
4. Should Viewport be rotateable on its own? This may be also a thing, as Camera may be connected to multiple viewports, so logically you could have a rotated camera in 2 viewports where one of the viewports is also rotated on its own.
6. Pivot property (the point around which image is rotated). It's possible to add, even though thinking about it made me wonder how the whole thing will affect convenience of object positioning. For simpliest example, if you have an overlay, and rotate its image around 0,0, then Overlay's position may end up visually in the bottom-right corner of an image. But maybe that's expected.


messengerbag:
Sweet! Did you already add this improved rotation capability to DynamicSprite?

Maybe I'm missing something, but why does `int Character_GetRotation(CharacterInfo *chaa)` (character.cpp 1571) return an int? Shouldn't it be a float?

ivan-mogilko:
> Did you already add this improved rotation capability to DynamicSprite?

I did not have such plans in this time. Dynamic sprites are currently rotated, resized and flipped by editing bitmap directly. Changing that would require to rewrite their internal logic, where in hardware accelerated mode they would only modify their transformation matrix, but if user requested for drawingsurface they would have to apply all the accumulated changes at once.

ericoporto:
the piixel perfect check I think uses this [`is_pos_in_sprite`](https://github.com/adventuregamestudio/ags/blob/40dd9a6ee9348f15ea7b85557b5eb7e9a484e826/Engine/ac/object.cpp#L486) function, which I think does reverse transformation on the point of interest to figure out if no transformation was applied to the sprite, would it hit the bitmap.

ivan-mogilko:
Yes, this function implementation has to be replaced with generic test which works for all rquired objects and uses full transform matrix (probably).

ivan-mogilko:
Added quick & dirty test of using matrix for hit detection on GUI: now rotated GUIs can be actually interacted.

PS. i love matrixes, although have no idea how they work :).

ivan-mogilko:
Kind of covered most of the functional basics, but this still need some reorganization + polishing.

[GIF](https://i.imgur.com/PQuApHe.gif)

ivan-mogilko:
Note: this feature mostly works, but right now I am delaying for two reasons:

* there are few design choices including script API which I'd like to think about again (maybe discuss with more people on forums);
* maybe will wait until SDL2 backend is merged into ags4, to reduce number of things to redo after.

ivan-mogilko:
Rebased and updated.

Note that I changed property names: GUI, Overlay and Camera still have **Rotation**, but Character and Object now has **GraphicRotation**, while "Rotation" property name is reserved for these two in case we introduce full entity rotation (which would include collider box, walkable area connection etc).

Currently most work is complete. Both visual and interaction works with rotation. All interaction now uses matrix math, as that is supposedly faster than using trigonometric functions for each hit test, also it's more uniform (trigonometry of course is still used when setting these matrixes up).

The biggest remaining part is software renderer. This pr may be merged without it since we have a WIP branch anyway, and may add rotation support in software renderer later.

Minor missing element is range checking, I think user-set angles should be brought to 0-360 degree range.

ivan-mogilko:
Pushed an update with preliminary software rotation support. Currently this only works for characters and objects, but these were perhaps most difficult to do, as there's alot of messy code related to their image caches. I'm still not certain if I covered everything though, this is something to find out as the new feature gets tested...

There's currently a drawing bug: when sprite is rotated the intermediate bitmap is not cleared to transparent pixels. But I left this temporarily because ironically this made it easier to test (I think something similar could be made as an official debugging feature at some point).

I made an important change to how sprite coordinates are sent to 3D renderers (D3D, OpenGL): now engine passes object's origin coordinates instead of precalculated coordinates of the top-left corner of the sprite in the room. That is done for consistency. Software renderer draws prepared transformed bitmaps (already scaled and rotated), so it has to know the top-left corner. But 3D renderers do all transformations on their own, so it seemed more convenient to just pass object's position + "origin" parameter; "origin" - is a factor of relative sprite offset, for example (0.5, 1.0) for character and (0.0, 1.0) for room object.

I suppose the code could be refactored further to keep most of these differences inside renderer's implementation, or at least in a separate chain of functions, but there's too much to change now.

One last thing, but also important. I suddenly realized that my Rotation property rotates counter-clockwise (that's just how 3D renderers work so I relied on their way), but existing DynamicSprite.Rotate rotates clockwise..... So now I am wondering if I should actually make the property rotate clockwise too. Because there should be consistency in this.

AlanDrake:
Clockwise is usually what is expected

ivan-mogilko:
Only software camera rotation left.
But seriously this might be the last big thing I do for visual effects until the drawing code gets a good cleanup...

ivan-mogilko:
Added software camera rotation, and switch rotation meaning to clockwise.

ivan-mogilko:
Alright, it's working. The software renderer may have few issues, but these may be fixed later.

Issue #1222, opened 2021-02-27
Segmentation fault in ogg_page_serialno when restoring saved game


morganwillcock:
**Describe the bug**
It appears to be possible to have a saved game which is unable to be restored, giving a segmentation fault trying to restore the audio state. This definitely occurs with 3.5.0.30  down through 3.5.0.27. 3.5.0.26 doesn't want to load it, probably for other reasons (the message displayed in the game is not clearly readable because of the font and resolution) so I haven't tested anything lower than that.

The problem appears to be related to the `vorbisfile` library that is shipped with the engine, so the problem exists by default for games which are built by the AGS Editor. Moving the library into a sub directory means the engine doesn't find it:

```
cd lib64
mkdir disabled
mv libvorbisfile.so.3 disabled/
```

...now the saved game can be restored with no issues, as long as the OS can provide a working vorbisfile library. As far as I know the library files which are shipped with the engine are still just a copy of the Debian 8 ones, copied with `B00_copy_libs.sh`, but I've not had this happen before. I'm using Debian 9.

The problem doesn't occur with every saved game. I've triggered it twice, probably after saving somewhere between 20 and 30 times whilst using the supplied libvorbisfile.so.3 file. The same saved game file can be restored on the Windows engine with no problems.

**AGS Version**
3.5.0.30

```
Thread 1 "ags" received signal SIGSEGV, Segmentation fault.
0x00007ffff7573c03 in ogg_page_serialno () from /home/mwillcock/Build/ags/Engine/data/lib64/libogg.so.0
(gdb) backtrace
#0  0x00007ffff7573c03 in ogg_page_serialno () from /home/mwillcock/Build/ags/Engine/data/lib64/libogg.so.0
#1  0x00007ffff6ef1b1d in ov_pcm_seek_page () from /home/mwillcock/Build/ags/Engine/data/lib64/libvorbisfile.so.3
#2  0x00007ffff6ef1d7c in ov_pcm_seek () from /home/mwillcock/Build/ags/Engine/data/lib64/libvorbisfile.so.3
#3  0x000055555563dad6 in alogg_seek_abs_msecs_ogg (ogg=<optimized out>, msec=msec@entry=11) at libsrc/alogg/alogg.c:330
#4  0x000055555577607c in MYSTATICOGG::play_from (this=0x555555ebce70, position=11) at media/audio/clip_mystaticogg.cpp:183
[...]

PR #1236, opened 2021-03-13
Linux: package newest versions of Xiph libraries

morganwillcock:
Adds build of libogg, libtheora, and libvorbis into the Linux build Docker container, with installation into /opt. Installing here means that the existing Debian hook script still gives the same result, but newer versions are available to be copied into the archive which is distributed as the Linux version of the engine.

The CI task which builds intermediate archives for each architecture will now look to replace the named Xiph libraries, if present. i.e. lib64/libvorbis.so.0 will be replaced by dereferencing /opt/lib/libvorbis.so.0 when lib64/libvorbis.so.0 exists

This should be seen as a short term change to work around issues found within older Debian libraries, ideally final archive archive contents need to be verified to be in the correct state but this needs further improvements to the build system.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -76,7 +76,16 @@ build_linux_debian_task:
       DEB_BUILD_OPTIONS="rpath=$RPATH_PREFIX$bit" fakeroot debian/rules binary
       sed -E "/^BI(NDMOUNT|T)=/d" debian/ags+libraries/hooks/B00_copy_libs.sh | BIT=$bit BINDMOUNT=$(pwd) sh
       ar -p ../ags_${version}_$arch.deb data.tar.xz | unxz | tar -f - -xvC data --transform "s/.*ags/ags$bit/" ./usr/bin/ags
-      cd data && tar -cvzf ../data_$arch.tar.gz *
+      cd data && \
+      (cd lib$bit && find . \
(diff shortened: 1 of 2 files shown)


morganwillcock:
I've verified that the libraries are updated in the archive:
https://api.cirrus-ci.com/v1/artifact/task/6668805150605312/binaries/ags_3.5.0.31_linux.tar.gz
...and tested that the crash is fixed.

Fixes #1222

rofl0r:
tbh, i'd find it more appealing if those would all build the static libs only (--enable-static --disable-shared) and the cirrus task/docker be changed to not install apt's version of those libs, then add -L/opt/lib to LDFLAGS and -I/opt/include to CFLAGS/CXXFLAGS for the build.
that way the shared libraries wouldnt need to be copied around and the ags binary would only link the parts of the code that's actually needed (with shared libs you always get *everything*)

morganwillcock:
Because of the previous reliance on the hook script, the engine binary is being extracted from the result of the Debian build process which is meant to use other Debian packages as dependencies rather than use static linking. So I agree that the change is slightly convoluted and not ideal, but given the amount of restructuring to separate concerns and the fact that the current setup is actually crashing,  I think this is OK in the short term.

PR #1238, opened 2021-03-14
Engine: Fix positioning with custom text window GUI

criezy:
This was a regression introduced with commit 41d156ef63 (from pull request #1104). Only one of the two calls to draw_text_window_and_bar in _display_main were changed to pass the local adjustedXX and adjustedYY variables. Games using the other unmodified call (when `asspch` is 0) have their text boxes shifted to the bottom and right as a result.

For example in Order of the Thorne: The King's Challenge, we get
image

And with the fix in this pull request we get
image

The Blackwell Converge is another game impacted:

Current master:
image

With the fix in this pull request:
image


ivan-mogilko:
This looks like a silly oversight, thank you.

PR #1250, opened 2021-04-11
New compiler: Compile-time expressions, `++` and `--` operators, `readonly` parameters

fernewelten:
This PR adds several features which were interlinked, so I found it easiest to do them jointly.

__Compile-time expressions__

Integer or float expressions that consist of constants and that don't have any side effects will be evaluated at compile time. You can use a compile-time expression in lieu of a literal wherever a literal is expected. This will work with all the operators that have a symbol and allow code such as

```
enum Flags
{
	kBlue = 1 << 0,
	kSquare = 1 << 1,
	kSmelly = 1 << 2,
};

#define kRows 7
#define kColumns 8
#define kOffset 1
int Array[kRows * kColumns + kOffset];
```

`enum` constants can be part of constant expressions and will be evaluated at compile time, but `readonly` variables currently will not. I haven't implemented that because `readonly` variables might still be modified, e.g., when they are exported and an importing module does __not__ declare them readonly. The compiler isn't smart enough to detect something like this.

Function calls will never be evaluated at compile-time.

This will work for integers and floats, but not for strings.

__Increment and decrement operators__

`++` and `--` have become operators. You can use them in C style within expressions, either prefix or postfix. As in C, `i--` will evaluate to `i` and decrement the variable afterwards.

Typical code:
```
for (int i = 0; i < 10; ++i)
    if (Components[i]-- > 0)
        player.Say("Adding yet another component to the flux capacitor!"); 
```

I haven't implemented `=` etc. as operators. If users clamour for it, I'm willing to offer that. But in C, it is a very easy mistake to code `if (i = 5)` and mean a comparison instead of an assignment. Currently, AGS will catch such a mistake because '=' can't be in an expression, which is a nice feature IMO. 

__readonly parameters__

There's been clamouring for `const` parameters to mean parameters that are readonly within the function body. But unfortunately, `const` is a keyword with a very specific meaning in AGS.

However, parameters that can't be modified are analogous to local `readonly` variables. And so I've implemented `readonly` for function parameters.

Typical code:

```
function ObjectOpen(readonly Object *o) 
{
    o = null; // will yield a compile-time error
}
```


ivan-mogilko:
> There's been clamouring for const parameters to mean parameters that are readonly within the function body. But unfortunately, const is a keyword with a very specific meaning in AGS.

I remember we discussed this but never finished the discussion. I tried to argue that `const` has pretty logical meaning in AGS.

Could you elaborate your point here? Because I believe it must be `const`, not `readonly`, as readonly is pure attribute's type that means "no setter".

ivan-mogilko:
Sorry, nevermind, I deleted my previous comment.

Although I still think it should have been `const` from the start, but considering how `readonly` is already used everywhere already since probably AGS 2.72 (if not earlier) would make a big change silly. Function parameters seem to be the only place where `readonly` was not allowed for some obscure reason.

fernewelten:
So, I wasn't around at the time. How __did__ `readonly` come into existence? Did Chris Jones decide on that? Was there once a type system aficionado in the team?

Our type system does seem kind of very elaborate for an adventure game engine. We've got `readonly`, `writeprotected`, and `const`, when the main use of the Engine is to just make Fragly close a door or pick up a dragon egg. 

So as a grand recap, here's the C# view of `readonly`, as far as it pertains to __variables__ (not attributes):
> In a field declaration, readonly indicates that assignment to the field can only occur as part of the declaration or in a constructor in the same class. A readonly field can be assigned and reassigned multiple times within the field declaration and constructor.  A readonly field can't be assigned after the constructor exits.  (This seems to be the original view of the concept; Microsoft has since added onto this by inventing `readonly struct`s and such. Cf. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly)

The upshot is that `readonly` is a characteristic of a __variable__. A readonly variable is still a variable. Memory will be assigned to it at runtime. Only it's a variable that you can't write to once it is initialized. 

In contrast, here's what C# says about `const`:
> Constant fields and locals aren't variables and may not be modified. ... Don’t create a constant to represent information that you expect to change at any time.  ... [B]ecause compilers propagate constants, other code compiled with your libraries will have to be recompiled to see the changes.(cf. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/const)

In other words, as I understand it: When you do `const int X = 2;`, the field `x` is __not__ a variable, it's another designation for `2`.  No memory will be reserved for `X`  at runtime. Instead, wherever the definition is in scope and an `X` is found, the compiler will immediately go, "Ah yes, they mean a 2 here. Let's convert that to 2 and carry on."

So as a recap, what's the AGS view on it? 
- As far as I understand from what's in the old compiler, when you do `readonly int X = 4;` you define a variable in the C# sense. Memory will be reserved on the stack or in the `global` space, and the compiler will prevent assignments to `X`. 
- When you try to do `const int X = 4;`, you get a compiler error. 

That's what happens in the new compiler, too. 

Now, __function parameters__.

In general, function parameters are local variables that have the special feature that they are created and initialized by the function __caller__. They are variables in every other respect: The function can use parameters just like variables, and memory is reserved for them on the stack just like for variables.

So that's why I implemented "readonly" over "const" for now: It's what would have been correct in a local variable declaration. 

But if the consensus is different, I'm certainly willing to change this on short notice.

We've also got struct declarations, and within those, attribute declarations and proper field declarations, with the yet additional keyword `writeprotected`. That's a quagmire on its own. As far as I know, the current engine functions don't even use `writeprotected` variables anywhere. 

Can we just deprecate it and instead introduce a `private` section for variables that may __neither__ be read __nor__ written from the outside?

ivan-mogilko:
> So, I wasn't around at the time. How did readonly come into existence? Did Chris Jones decide on that? Was there once a type system aficionado in the team?

Before AGS 3.2.2 there was no team :) it was Chris Jones working alone, perhaps some of the long term AGS users affected his decisions. But I don't know history that far, maybe searching on the forums somewhere around 2.71-2.72 release time could shed light on this (if it's worth it).

`readonly` exists and works for almost everything (except function params) in 2.7, I think it was added along with OO-style API (Character, Object etc structs) for struct member access regulation.

But yes, I do think that `readonly` is strange, as it defines different kinds of behavior depending on situation. When used with variables it is practically "const", when used with attributes it means "no setter function". Plus, struct variables declared as `readonly` are not necessarily "const" in practice, as they also may be changed by the engine.

Sorry, I have to read and reply to the rest of your comment later.

ivan-mogilko:
BTW, reading the "const" article linked above, I found an important passage comparing const and readonly in C#:

> The readonly keyword differs from the const keyword. A const field can only be initialized at the declaration of the field. A readonly field can be initialized either at the declaration or in a constructor. Therefore, readonly fields can have different values depending on the constructor used. Also, although a const field is a compile-time constant, **the readonly field can be used for run-time constants**, as in this line: public static readonly uint l1 = (uint)DateTime.Now.Ticks;

This differs from C/C++ of course where it will be `const` in both cases. So, true, in case we decide to stick closer to C# principles *and* keep `const`, then `const` should only be a compile time calculated value, while readonly used for actual variables that cannot be changed in certain scope.

> There might perhaps be some strange need for variables that you can't write to in the C# universe, but I can't imagine why I'd urgently need that in AGS code

Well, a random example of using readonly variables is how const may be used in C/C++: if you have some complicated code you may sometimes want to declare precalculated variables as const to not let them change anywhere later by mistake.

ivan-mogilko:
@fernewelten 
Testing this, I found there's a difference of how readonly function parameters are treated, which leads to a bug.

For example:
<pre>
function ReadonlyTest1()
{
    readonly int a = 0;
    readonly int b = a;
    int c = a;
    return a;
}
</pre>
All of this works.

But, if we do this:
function ReadonlyTest2(readonly int i)
{
    readonly int a = i;
    int c = i;
    return i;
}
Every single line will cause compiler error, either `Cannot assign a type 'const int' value to a type 'int' variable` or `Type mismatch: cannot convert 'const int' to 'int'`.
So I cannot even assign readonly parameter to readonly local variable.

fernewelten:
That was me confusing `const` parameters (must only be assigned to other variables of type `const`) and `readonly` parameters (may not be changed but may be assigned to other variables).

The newest changes should have fixed it. I hope. All your example assigns and returns should work now.

ivan-mogilko:
Yes, this works now.

But it has to be rebased after merging Array.Length pr unfortunately.

fernewelten:
Rebased the PR. 
Warning: I force-pushed the changes.

PR #1251, opened 2021-04-11
Engine: Fix finding unused overlay custom ID

criezy:
This was a regression from cb7ab658a.

The custom IDs are supposed to be `> OVER_CUSTOM`, but since that commit it can be `OVER_CUSTOM` (for example if the `screenover` vector is empty). We need the loop to have one more iteration than the number of overlays to be sure we find an unused one.

This issue caused some overlays to use `OVER_CUSTOM ` as their ID, which can then result in the "SetTextOverlay internal error: inconsistent type ids" as the ID can gets changed in `add_screen_overlay` when it thinks it need to try to find an unused ID again.

### DIFF
--- a/Engine/ac/overlay.cpp
+++ b/Engine/ac/overlay.cpp
@@ -288,7 +288,7 @@ size_t add_screen_overlay(int x, int y, int type, Common::Bitmap *piccy, int pic
 {
     if (type == OVER_CUSTOM) {
         // find an unused custom ID; TODO: find a better approach!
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Thank you. Although this code is so weird that I could not remember what it means. There's even a TODO about changing it.

rofl0r:
why is this PR targeting ags4 and not ags3 ?

ivan-mogilko:
Because this change only exists in ags4.

Issue #1156, opened 2020-12-15
Engine start writes user config


morganwillcock:
Just running a game writes a user configuration file, which then overrides any change made to the game's configuration file.

This is extremely confusing since:
* editing the game configuration (e.g. windowed mode) would work before starting the game
* editing the game configuration (e.g. windowed mode) would not work after the game was started once

```
$ cat /home/mwillcock/.local/share/ags/DaggerHell/acsetup.cfg
cat: /home/mwillcock/.local/share/ags/DaggerHell/acsetup.cfg: No such file or directory

$ ~/Downloads/DAGGERHELL/DAGGERHELL
Adventure Game Studio v3.5 Interpreter
Copyright (c) 1999-2011 Chris Jones and 2011-2020 others
ACI version 3.5.0.27

Initializing allegro
Initializing game data
Located game data file: /home/mwillcock/Downloads/DAGGERHELL/data/DaggerHell.ags
Opened game data file: game28.dta
Game data version: 49
Compiled with: 3.4.3.1
Setting up game configuration
Data directory: /home/mwillcock/Downloads/DAGGERHELL/data/
Setting up window
Initializing TTF renderer
Initializing keyboard
Initializing mouse: number of buttons reported is 3
Install timer
Sound settings: digital driver ID: 'AUTO' (0xffffffff), MIDI driver ID: 'AUTO' (0xffffffff)
Trying to init: digital driver ID: 'SDL2' (0x53444c32), MIDI driver ID: 'DIGI' (0x44494749)
Installed digital driver ID: 'SDL2' (0x53444c32), MIDI driver ID: 'DIGI' (0x44494749)
Install exit handler
Initialize legacy path finder library
Game GUI version: 118
Requested script API: v3.4.1 (5), compat level: v3.4.1 (5)
Game title: 'DaggerHell'
Checking for disk space
Initializing MOD/XM player
Game native resolution: 320 x 200 (32 bit)
Graphic settings: driver: D3D9, windowed: yes, screen def: max, screen size: 0 x 0, match device ratio: ignore, game scale: max_round
Requested graphics driver 'D3D9' not found, will try existing drivers instead
Running OpenGL: 3.0
Mouse control: off, base: 1.000000, speed: 1.000000
Initialize sprites
Audio is processed on the main thread
Engine initialization complete
Starting game
***** ENGINE HAS SHUTDOWN

$ cat /home/mwillcock/.local/share/ags/DaggerHell/acsetup.cfg
[graphics]
[...]

Issue #975, opened 2019-10-24
Find solution for the problem of user config "secretly" overriding default cfg


ghost:
This is written after this: #789, but the problem exists since introduction of a user config which is written in a "safe writeable" directory (currently: in the saves dir), and somewhat worsened when engine began to save particular options upon exit (options that could be changed at runtime, such as translation for instance).

The problem is this: unless user is aware that this extra config file is saved, he or she may be confused when trying to change a config option in the file inside game dir only to see that it does not work. This is specifically essential on Linux, where we do not support builtin setup program.
Even if user is aware of that, it may be simply tiresome to find config file for particular game every time.

Note that since AGS 3.5.0 engine can tell where configs are stored if you run it like `ags --tell-configpath` or print current effective config if you do `ags --tell-config`m which may help diagnosing the situation. But that don't change the fact that you still have to keep this in mind and search for the user config.

So, some kind of solution is needed to ease this situation for developers and players. I'll leave this ticket open for suggestions.

**See Also**: #961

i30817:
I don't think there is a good solution in the absence of a config program for naive users. The main problem is that it's the unaware user that will have this problem and making them 'aware' after the fact when AGS runs with a timestamp of option override detection is both intrusive at best (a sdl text saying some settings were replaced) and most likely a false positive from the game distribution if it wasn't the user itself screwing up.

legluondunet:
same issue here:

```
$ ./ags64 --tell-configpath
[config-path]
default = /home/legluondunet/Bureau/tmp/ags/ags_3.5.0.23_linux/data/acsetup.cfg
global = /home/legluondunet/.local/share/ags/acsetup.cfg
user = /home/legluondunet/.local/share/ags/Maniac Mansion Deluxe/acsetup.cfg
```

but in fact the Global config take control over the default config file in game path.
Is it not possible to ask ags to ignore other config files when when there is one in the game data path?

ericoporto:
@legluondunet I am making [a graphical setup program](https://github.com/ericoporto/agsconfig), the idea is that a developer could ship it along with their game, if they choose to, while there's no default one provided by the AGS Engine - I lack in skills to figure out and code a proper one in the engine itself.

ivan-mogilko:
> Is it not possible to ask ags to ignore other config files when when there is one in the game data path?

It should be possible, #961

ivan-mogilko:
Posted by @legluondunet in #1163 (closed as duplicated and moved here)

[...]

Issue #961, opened 2019-10-19
Engine mode to have everything saved and read in the game dir


ghost:
In the past multiple users have expressed a wish to be able to make engine save everything in the game directory similarily to how AGS worked originally; this means: user config, game saves, any dynamic data game writes.

As of 3.5.0 there are two separate options for saves path (user_data_dir) and "shared data" path (shared_data_dir), these may be used to remap file saving into game dir, but even with these set engine will still write user config elsewhere (currently - into the common savegame dir).

The suggestion is to introduce a simple boolean config option that would switch engine defaults from using common writeable locations to the game directory for literally everything that engine saves and reads for this game.

AlanDrake:
Or maybe not the game directory but a subfolder/s

morganwillcock:
On Windows this could be pretty confusing, because unprivileged writes to a privileged location will be silently redirected to a per user location (within the user profile). If this feature is added, to work as intended it would need to explicitly bypass the redirection, in-case the game has been copied to a location where the user cannot write, and also gracefully handle being denied access.

ivan-mogilko:
Posted by @legluondunet in #1163

> I'm writing a game script on Lutris (games platform for Linux) which will use AGS.
I need support about an issue.
I put a config file in main game folder, acsetup.cfg.
But when AGS launch it copies this config file in /home/your-username/.local/share/ags/your_game/.
Then AGS only take care of the cfg file located in /home/your-username/.local/share/ags/your_game/.
Modifications done on acsetup.cfg in main game folder are no more taken into account.
Is there a way to tell AGS to not copy config file in user profile and just take in account the config file in main game folder?
Our installs are locale, we don't want to use user profile folder, Lutris should manage AGS option in game folder (or other place we could choose).

ivan-mogilko:
Not exactly how I envisioned this originally, but there is now the option to keep user config in game dir, and save/appdata paths were already customizable. (#1252, 109a9d985f498ed31308e2e97e1b97fa306c87c0)

PR #1252, opened 2021-04-11
Engine: implemented command line options for config and user dir paths

ivan-mogilko:
In response to #961 and #975.

Implemented following:

 --conf - defines explicit and only config file path;
 --user-data-dir - defines user data directory for game files (aka $SAVEGAMEDIR$);
 --shared-data-dir - defines shared data directory for game files (aka $APPDATADIR$);

Example of use:

`ags --conf ./my.cfg --user-data-dir ./saves --shared-data-dir ./appdata`

NOTES:

`--conf` forces engine to ignore all other config files.

`--user-data-dir` and `--shared-data-dir` override corresponding options from config file.

---

The biggest issue I have with `--conf` is that so far it does not have a clear rule in regards to *saving config back*. In other words, must it be considered a read-only special config, or actually "user config" with all the rules applying?
A while ago I've disabled saving config by the engine on exit because it was confusing for number of reasons (see #1156, also related to #975).
But, if we ever enable that, or, for instance, let game developers change and save config from within the game script, what should be the rule in regards to file passed as `--conf`?
I have a concern that if unresolved now then in the long run this may become a cause of more logical conflicts.

### DIFF
--- a/Engine/main/main.cpp
+++ b/Engine/main/main.cpp
@@ -321,6 +321,10 @@ static int main_process_cmdline(ConfigTree &cfg, int argc, char *argv[])
         //
         // Config overrides
         //
+        else if ((ags_stricmp(arg, "--user-data-dir") == 0) && (argc > ee + 1))
+            cfg["misc"]["user_data_dir"] = argv[++ee];
+        else if ((ags_stricmp(arg, "--shared-data-dir") == 0) && (argc > ee + 1))
+            cfg["misc"]["shared_data_dir"] = argv[++ee];
         else if (ags_stricmp(arg, "--windowed") == 0)
             force_window = 1;
         else if (ags_stricmp(arg, "--fullscreen") == 0)


rofl0r:
> The biggest issue I have with --conf is that so far it does not have a clear rule in regards to saving config back. 

i suspect if the user uses command line option to force a specific option, it would be unexpected to have it overwritten by engine.
for example i was surprised when AGS started last game i played in fullscreen, even though i put windowed manually into config, probably because it was in fullscreen when i quitted it. i'd rather that windowed stays the default unless i change it.

ivan-mogilko:
Well, what I am trying to say, this option in pr, it tells to explicitly use certain config for startup. That's fine on its own, but this does not cover another use case when players wanted not only read but also write config in game dir. Because I disabled writing config by engine (on exit), right now this may happen only with embedded setup program (on Windows), or any standalone setup app that would use engine's rules of finding user config.
Assuming in the future we may add support for writing config by script command, that would also cause similar problem.

On a somewhat related note, right now there's already at least one script command that may change game setup: it's the System.Windowed that lets game switch from windowed to fullscreen. Making abstract assumption that a game uses a custom config file, in theory it may even switch on start. I guess that's a separate issue, but wanted to mention this, because there is also an old feature request to support other graphic settings in script.

ivan-mogilko:
@legluondunet , could you tell, given the above does current solution work for you?

I've been also thinking maybe add an option that tells to use default config's location (in game dir) as user config location in all operations (so long as it's writeable).

legluondunet:
1) About read/write config, I think it will be better if ags would write options in config file on the fly.
2) Another thing, what's the matter if I start  a game ike this:
ags -conf ./game.cfg
and game.cfg file does not exit, ags will create it with default ags values right?
3) What's the matter if the folder where is located config/saves files is read only?

ivan-mogilko:
> 1. About read/write config, I think it will be better if ags would write options in config file on the fly.

Could you please elaborate, what situation you are refering to?

> 2. Another thing, what's the matter if I start  a game ike this:
>    ags -conf ./game.cfg
>    and game.cfg file does not exit, ags will create it with default ags values right?

The latest engine does not create or write anything.
If we take slightly earlier version, it wrote only *certain* settings on exit (the settings that could be changed by script commands).
If we take embedded setup program (it works only on Windows), that would create user config with default values, but that config would be located according to standard path rules.
In its current implementation `--conf` does not directly correspond to user config, it does not replace its location; that's why I proposed to add another option that actually would do that, either instead or along with `--conf`.

> 3. What's the matter if the folder where is located config/saves files is read only?

Back when the engine was writing config on exit it would silently fail. The embedded setup would display an error message.

Any changes to above behavior may be a subject of discussion of course.

legluondunet:
In summary if the config file is read in read only that suits me very well, as long as some games does not complain.
If read-only config file cause an issue for some games, it will be another issue to open.

ivan-mogilko:
Well, I'm still not fully certain if we have same understanding of this.
Right now `--conf` does not impose any additional rules. Like, it does not tell engine "not to write into this file under any circumstances". It only tells "read config from there on startup, and from nowhere else".
If we assume that engine will be writing user config again at some point, and if you pass a file that is NOT the one which engine or setup program is considering user config location, then they won't write in it.


On another note, I have a dumb question. Does it matter if the command options for user data and shared data dir have precisely same name as in config file? I just realized in config they have underscores between name parts, but I made cmd args with dashes. Actually, we have a bit of a mess there, not all command arg names correspond to config 1 to 1.

rofl0r:
> Does it matter if the command options for user data and shared data dir have precisely same name as in config file? I just realized in config they have underscores between name parts, but I made cmd args with dashes.

traditional command line tools all use dashes instead of underline for wordbreaks, so that sounds good as is.

ivan-mogilko:
@legluondunet I am sorry, I should have checked your older comments in #975 to recall your use case.

Basically you have a game platform inside Linux, and are using your own setup utility to write game config inside game directory.
Your problem was that engine was writing a user config in user .local/share etc, overriding your settings.
This should not longer happen as of [AGS 3.5.0.31](https://github.com/adventuregamestudio/ags/releases/tag/v.3.5.0.31), where I disabled engine writing config on exit. Idea is that it is disabled *temporarily* until there's a good method to control this behavior, such as a restricting where it saves this config.

I made a mistake having several tickets refering to related problems but not updating them in time.

So, what I will do now, I will change this PR to only add `--user-data-dir` and `--shared-data-dir` as these correspond to existing options and may be still useful. For now I will not be adding any `--conf` or similar options, because these are simply not necessary (at least not on Linux where we do not have an embedded setup program).

The problem with writing config back remains (it still exists on Windows), but I will address this separately.

legluondunet:
I must admit that I have a hard time following you, in the end you are not going to add a command line option to tell AGS where the config file is located? 
Like
ags --config-path /your/path/to/config.file
or
ags --config- /your/path/to/config.file
or 
ags --base-path  /your/path/to/config.file and savefiles
?

ivan-mogilko:
> in the end you are not going to add a command line option to tell AGS where the config file is located?

The question is, what is the use case? Do you actually need this ability to tell engine explicit config location? Is that config is outside of game directory somewhere for example?

ivan-mogilko:
> I must admit that I have a hard time following you, in the end you are not going to add a command line option to tell AGS where the config file is located?

Sorry, let me rephrase this. Yes, I am not going to add this command line option, because I came to conclusion that it's not necessary in your case, as your problem was related to engine writing user config in a different location. But because that behavior is disabled this should no longer be a problem.

I may be mistaken (again) of course, in which case I ask to elaborate on the situation.

legluondunet:
I need to understand...
So after this PR applied, where AGS will search for game config file? 
Where it will search for savefiles?
English is not my native language, thank you for your understanding.

ivan-mogilko:
It will be the same. My question was, do I need to change that?

Because the latest engine _does not_ write any of these configs. So the only way they may be created is a) by user's hand b) by some setup application.

ivan-mogilko:
> Where it will search for savefiles?

By default it searches for them in the user's folder. But you may also set "user_data_dir" in config. After this PR you may also pass `--user-data-dir` in command line to define the save game location.

ivan-mogilko:
Alright. I made a mistake.

Perhaps you have a need to restrict engine to very particular config. So I will just add such command. Then if this would conflict with new engine behavior or some games later we may find a solution.

I apologize for all the confusion; I am thinking about several problems at a time.

legluondunet:
>Because the latest engine does not write any of these configs anymore. So the only way they may be created is a) by user's hand b) by some setup application.

ok, let's go for this scenari, should be enough to avoid config file confusion for users.
thank you to take time to explain me in details.

ivan-mogilko:
Ok.

I also apologize for all the confusion, I was thinking about several problems at a time.

rofl0r:
> For now I will NOT be adding any --conf or similar options, because these are simply not necessary (at least not on Linux where we do not have an embedded setup program).

what? that logic seems to be reversed, especially because we dont have a setup program we may want to pass the config file location to engine (for example using a *custom* frontend)

i'm still interested into having the --config option so i can finally write a wrapper around ags that always uses the config from current directory and i can finally stop to go hunting for the right config file.

ivan-mogilko:
My point above was that unless you create a config in user location, the engine will use the one in the game dir anyway. There would not be a need to hunt for config elsewhere.

But yes, I can add the command for explicit config location. The question that I've been asking is: should it be just a config for reading at startup, or it should replace user config in all aspects. As in, assuming engine would like to write config back, should it use the path provided with this command?

ivan-mogilko:
Added `--conf` f47c365473abe9ffd4e269c563e9bb14ff428985

rofl0r:
> As in, assuming, hypothetically, engine would like to write a modified config back, should it use the path provided with this command?

i still fail to see the purpose of having the engine write a config back, if not to document what its default values are.

but in case this gets the default again we could add a switch --config-readonly (boolean switch) or --ro-config-path=./acsetup.cfg which sets both readonly to true and sets the path.

ivan-mogilko:
> i still fail to see the purpose of having the engine write a config back

For instance, there was a feature request of editing config from within the game, so that if player changes settings in game menus it would be saved and restored on next run.

I should also probably mention that since 3.5.0 there's now `--tell-configpath` command that prints locations of all configs as expected by the engine to stdout. In theory this may be used by setup program if it needs to follow engine's default rules.

rofl0r:
`--tell-configpath` is an improvement and i was actually refering to that when i talked about `hunting down` the location, because you always have to look into all 3 configs to figure out from where the option you don't want comes from...

> so that if player changes settings in game menus it would be saved and restored on next run.

hmm that sounds as if the option they actually wanted is a script function called SaveConfigToDisk()

Issue #637, opened 2019-02-18
Add Length property for arrays


alkhimey:
Currently, the workaround for not being able to "ask" an array of it's length is to keep track of the length using separate variable. This is inconvenient and can lead to nasty bugs when the length is changed, but the variable is not.

Instead of:

    int a[] = new int[3];
    readonly int len = 3; // should be same as the length of a
    f(a, len);

    function(int[] a; int len) {
        for (i=0; i < len; i++) {
            ...
        }
    }

We could:

    int a[] = new int[3];
    f(a);

    function(int[] a) {
        for (i=0; i < a.Length; i++) {
            ...
        }
    }

fernewelten:
As I see it, there are two main ways to do it.

__One way__ would be to get the Engine to help. It already knows the size of dynamic memory blocks and uses this for memory overrun checks (opcode `SCMD_DYNAMICBOUNDS`). So we could add an opcode that reads the memory size and returns it into a register. The compiler would then translate all calls to `array.Length` into code that uses the new opcode.

So the disadvantage is that the Engine would need to be extended.

__The other way__ would do all the logic on the compiler side, with the existing opcodes: 
- Whenever a program defines a dynamic array `a`, the compiler would secretly add a definition for `a#Length`. 
- Whenever the program does `a = new Struct[size];`, the compiler would generate the equivalent for `a#Length = size;`.
- Whenever a program defines a function with a dynamic array parameter, the compiler would add a secret length parameter. And when the program passes `a` to the function, the compiler would pass `a#Length` to the secret extra parameter. 
- Whenever a program defines a function that returns a dynamic array, then the corresponding bytecode would return the array in register AX and its size in BX. The caller of such a function would read BX after it returns and use this value to keep track of the size of the returned array.

The disadvantages: 
- All existing system functions that either accept dynamic array parameters or return dynamic arrays would need to be rewritten to accommodate the new calling convention.  
- __Both__ the old and the new compiler would need to be modified in parallel.  Otherwise, since the calling convention for functions is changed, the old compiler couldn't compile _any_ program any longer, no matter whether that program uses `a.Length`.

ivan-mogilko:
Ermmm... yeah, I would go with engine's approach :).
I mean, dynamic arrays are already *managed objects*, so that's natural that engine participates in handling them.

ivan-mogilko:
Hey, but, in fact, there's another 3rd option:

* don't add any new opcodes;
[...]

PR #1256, opened 2021-04-13
New compiler: `.Length` pseudo-attribute for dynamic arrays

fernewelten:
This PR adresses #637.

It adds the `Length` attribute for dynamic arrays which will return the length of any dynamic array at runtime. 
Specifically, if `DA` is a dynamic array, then:
- `DA.Length` will return the the total number of all the fields.
- ~~`DA.Length[0]` will return the number of dimensions, which is always `1` because currently our dynamic arrays only have 1 dimension~~
- ~~`DA.Length[i]` for a positive value of `i`  will return the size of the `i`th dimension (where the first dimension is number 1, not number 0).~~ ~~So currently, `DA.Length[1]` will yield the same as `Da.Length`.~~   ~~If there isn't such a dimension, in particular for a negative `i`, you'll get a runtime out-of-bounds error.~~

~~Currently, our dynamic arrays don't support multiple dimensions, but this may change some day. In the meantime, we don't need to tell anybody about the bracketed parameter if we don't want to. :-)~~

This PR uses an idea of @ivan-mogilko and adds an Engine function to provide the core functionality. This function is only imported and called from the compiler when needed. 

The Engine is new territory for me, so:
**please check the Engine code for whether I've adhered to any conventions that may be in place, whether anything is missing or wrong, whether the code blends in tidily, etc.**

I've currently only done this for dynamic arrays. This is probably the most important use case because you can pass a dynamic array to a function and then, within that function, the code will probably want to know how large that dynamic array is. ~~I intend to implement the `Length` attribute for static arrays, too, for consistency, but make another PR for it.~~

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -1747,6 +1747,14 @@ AGS::ErrorType AGS::Parser::ParseFuncdecl_HandleFunctionOrImportIndex(TypeQualif
 // An extender func param, if any, has already been resolved
 AGS::ErrorType AGS::Parser::ParseFuncdecl(size_t declaration_start, TypeQualifierSet tqs, Vartype return_vartype, Symbol struct_of_func, Symbol name_of_func, bool no_loop_check, bool &body_follows)
 {
+    // __builtin_
+    // 123456789a
+    if (0 == _sym.GetName(name_of_func).substr(0u, 10u).compare("__Builtin_"))
+    {
+        Error("Function names may not begin with '__Builtin_'");
+        return kERR_UserError;
+    }
+
     ErrorType retval = ParseFuncdecl_DoesBodyFollow(body_follows);
     if (retval < 0) return retval;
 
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
> * `DA.Length[0]` will return the number of dimensions, which is always `1` because currently our dynamic arrays only have 1 dimension
> * `DA.Length[i]` for a positive value of `i` will return the size of the `i`th dimension (where the first dimension is number 1, not number 0).

Ungggg.... this idea makes me feel very uneasy, because it's unexpected and counterintuitive.

fernewelten:
> Ungggg.... this idea makes me feel very uneasy, because it's unexpected and counterintuitive.
  
Well, you're dead right there. It's a violation of the rule, “what's written on it must be in it”. The attribute is called, “Length”, and so it should return a length and not a number of dimensions, no matter what the argument is.  What about splitting the thing up?
- ~~DA.Length[0]~~ `DA.Dimensions` will return the number of dimensions, which is always 1 because currently our dynamic arrays only have 1 dimension.
- `DA.Length[i]` for a non-negative value of `i` will return the size of the `i`th dimension, where `i` is zero-based, not one-based. A negative `i` will trigger a run-time error.

fernewelten:
(By the way, we Germans have the nice name "Nutella rule" for that rule because at the turn of the 1980s a nougat spread manufacturer had plastered the commercial slogan „Nur wo Nutella draufsteht, ist auch Nutella drin“  (Only where it says "Nutella" on it, is there's Nutella in it.) all over the country.)

rofl0r:
```
    DA.Length will return the the total number of all the fields.
    DA.Length[0] will return the number of dimensions, which is always 1 because currently our dynamic arrays only have 1 dimension
    DA.Length[i] for a positive value of i will return the size of the ith dimension (where the first dimension is number 1, not number 0). So currently, DA.Length[1] will yield the same as Da.Length. If there isn't such a dimension, in particular for a negative i, you'll get a runtime out-of-bounds error.
```
this is unlike anything i've ever seen in other PLs.
first, using the array syntax `length[]` is odd, shouldn't it be like a method call?
in other PLs' array.length() would return the number of elements, and array[0].length() the size of the invidual element (however i dont see a good reason why this would even be needed, since we don't have malloc() and friend in ASC).

ivan-mogilko:
@rofl0r 
> this is unlike anything i've ever seen in other PLs.
first, using the array syntax length[] is odd, shouldn't it be like a method call?

this proposes Length as a property, which in AGS script may be indexed. E.g. compare to Game.AudioClips[] for example.

rofl0r:
> Game.AudioClips[] for example

isn't that a regular array though ?

> you meant each array dimension should be returning it's own Length instead?

yes

> I did not understand your comment about malloc though.

well e.g. in C usually you need to size of a single element (like sizeof struct foo) to do low-level stuff, such as allocate memory, copying memory, and the like. in a dynamic language like ASC one would instead just resize the array if more items are needed, leaving the low-level details to the engine.

ivan-mogilko:
> > Game.AudioClips[] for example
> 
> isn't that a regular array though ?

No, it's an indexed attribute, these still have getter and setter methods behind, but they have slightly different naming: `geti_NAME` and `seti_NAME` instead of `get_NAME` and `set_NAME`, and accept extra index argument. There are some parallels with [C# indexers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/), except in C# you must apply these to type rather than a property of type.

fernewelten:
We can simply cut all this out and leave this at `DynamicArray.Length`. Then this would simply become a parameterless attribute.

It would solve the problem at hand -- querying the length of a dynamic array that was passed as a parameter in a function body.

We could do this and move on and worry about multidimensional dynamic arrays if and when we get them.

ivan-mogilko:
@fernewelten tested, and it seem to work fine.

Could you squash the commits so that there are not redundant code addition/removal?

ericoporto:
is Length on autocomplete?

ivan-mogilko:
> is Length on autocomplete?

Probably no.

It's an interesting problem how to add this, because it's not a part of declared type, and also depends on compiler version.

PR #1262, opened 2021-04-20
Tools: crm2ash - generates room script header

### LINKED ISSUE #1147: Tool: Generate script headers from room files (CRM)
### Summary

A stand-alone tool which may be run from command-line (no GUI), which loads room's CRM format and generates script header necessary for compiling script for that room.
Presumably C++ is better choice here, as it's more portable than C# , and also there's already a CRM-loading code in the engine/editor that may be reused.

**Input:**

* a compiled room file (*.crm);
* path to output directory or name of output script file;

**Output:**

* generated room script header; the filename could be either taken from cmdline args or designed using some rule.

---

### Details

CRM file loading code:
https://github.com/adventuregamestudio/ags/blob/master/Common/game/room_file.h
https://github.com/adventuregamestudio/ags/blob/master/Common/game/room_file.cpp

*NOTE:* You do not have to read all the room blocks. Also I believe it may be possible to avoid linking allegro library, because it's only necessary for reading background and mask bitmaps, and you do not need these to generate scripts.

Script header generation process:
https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Tasks.cs#L569
(AppendRoomObjectsAndHotspotsToHeader function)

---

### See Also

#1146

ivan-mogilko:
Resolves #1147.

This is a proof-of-concept variant of the tool which parses compiled room files (CRM) and generates script header for them.
As it seemed to be the simpliest to make, I decided to try this tool first.

**crm2ash** stands for - CRM to ASH...

*Input:* compiled ags room (*.crm)
*Output:* generated script header with room declarations.

Usage: `crm2ash room.crm room.ash`

It links several utility cpps from the engine's Common folder. ~I hoped to reuse room file reading code too, but it has unnecessary dependencies and may need splitting before it's possible to do without bringing more stuff. So for now just copied necessary bits.~

Builds on Windows and Linux. For MSVS I added a new "Tools" solution. For Linux there's a Makefile. I am not a big expert in writing these, so copied one from ags itself, maybe few things were unnecessary.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -77,6 +77,7 @@ target_sources(common
     game/plugininfo.h
     game/room_file.cpp
     game/room_file.h
+    game/room_file_base.cpp
     game/room_file_deprecated.cpp
     game/room_version.h
     game/roomstruct.cpp
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -41,67 +41,6 @@ namespace AGS
 namespace Common
 {
 
-RoomDataSource::RoomDataSource()
-    : DataVersion(kRoomVersion_Undefined)
-{
-}
-
-String GetRoomFileErrorText(RoomFileErrorType err)
-{
-    switch (err)
-    {
-    case kRoomFileErr_NoError:
-        return "No error.";
-    case kRoomFileErr_FileOpenFailed:
-        return "Room file was not found or could not be opened.";
-    case kRoomFileErr_FormatNotSupported:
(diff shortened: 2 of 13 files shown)


ivan-mogilko:
Simplified the code, reusing more of the common engine files.

Renamed the tool to **crm2ash**, because it generates *.ash out of *.crm... hopefully this is not too ugly for a name, at least it matches what the tool does.

ivan-mogilko:
The tool is working and suppose is suitable as a first version.

Issue #1188, opened 2021-01-26
Tool: Build main game data from Game.agf


ivan-mogilko:
A stand-alone tool which may be run from command-line (no GUI), which parses Game.agf xml and write a main game data in binary format.
Presumably C++ is better choice here, as it's more portable than C# , and also there's already a game-loading code in the engine that may be reused.

*NOTE:* this task is exclusively in writing this tool, adjusting Editor is a separate task, so no need to concern yourself with that if you are doing this. In fact it's best to assume that Editor will not be present to ensure tool's work result has no reliance on it.

**Input:**
* an xml file in AGS 3.* project format (path to file);
* path to output file;

**Output:**
* file containing main game data in binary format (options, global objects, and so on).

---

### Details

The C++ code which reads game data may be found in a dedicated h/cpp pair:
https://github.com/adventuregamestudio/ags/blob/master/Common/game/main_game_file.h
https://github.com/adventuregamestudio/ags/blob/master/Common/game/main_game_file.cpp
Relevant functions are OpenMainGameFile() and ReadGameData().

The C++ code that writes same data was once present in the AGSNative dll sources, but it was very "dirty", and it was removed after C# variant was implemented (it was too bad to keep, and one can recreate it using reading code anyway).
In C# writing is performed by DataFileWriter class: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/DataFileWriter.cs
But if the tool would be written in C++ I'd recommend using C++ reading functions as the main reference.

In fact, it will be fine to put data writing in the same main_game_file.cpp, for consistency, and link it to the tool.

TODO: more info on how this data is stored in Game.agf

To be amended.

---

### See Also

#1146 - task on another tool which also parses Game.agf but for different purpose.
General overview of the game building process: https://github.com/adventuregamestudio/ags/wiki/AGS-Game-Build-process-(3.5.*)

ericoporto:
I am reading the C# implementation 

https://github.com/adventuregamestudio/ags/blob/4934cd0abc705941804e05b4c4d4feb1a5d4aca2/Editor/AGS.Editor/DataFileWriter.cs#L1384

It appears it would also need a dir containing the binary of the built scripts as input of the tool, since the script binaries are later stored inside the game28.dta structure.

---

Some notes here... The GameSetupStructBase code is already replicated in both C# and C++ (no idea if we ever test the c++ code path, is this a case of matching data and savegame?)

[...]

PR #1264, opened 2021-04-25
Tool: agf2autoash, agf2glvar - generate scripts out of Game.agf

### LINKED ISSUE #1146: Tool: Generate script files from Game.agf
### Summary

A stand-alone tool which may be run from command-line (no GUI), which parses Game.agf xml and generates _script files_ necessary for game compilation.
Presumably C++ is better choice here, as it's more portable than C#.

*NOTE:* this task is exclusively in writing this tool, adjusting Editor is a separate task, so no need to concern yourself with that if you are doing this. In fact it's best to assume that Editor will not be present to ensure tool's work result has no reliance on it.

**Input:**
* an xml file in AGS 3.* project format (path to file);
* path to output directory;

**Output:**
* _AutoGenerated.ash - contains declarations (imports) of global game objects, arrays, macros and so on;
* _GlobalVariables.ash - contains imports of global variables found in "GlobalVariables" xml section;
* _GlobalVariables.asc - contains definitions and export statements for "GlobalVariables";
*Optionally?*
* dialog scripts, each in a separate file (to be converted and merged by a different tool as a later step) - verbatim content of `<Dialogs> - <Dialog> - <Script>` xml sections, excluding `<![CDATA[` tag; need to figure out how to call these temporary files to avoid conflicts.

---

### Details

**_AutoGenerated.ash**

Process could be found in the Editor code:
https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Tasks.cs#L526
(RegenerateScriptHeader function)

Note that the header takes global declarations from game data itself (read from Game.agf), but also room declarations from the given room. This is because room scripts need both.
My proposal is to *ignore room declarations* for the purpose of this tool, and only make it work with global ones. The generated _AutoGenerated.ash may then be used everywhere, and for room scripts there will be a separate tool (or separate task) which reads room files (CRM) and creates additional headers per room.

**_GlobalVariables.ash/asc**

Process could be found in the Editor code:
ash: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Components/GlobalVariablesComponent.cs#L89
asc: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Components/GlobalVariablesComponent.cs#L57
(UpdateScriptHeader and GenerateGlobalVariablesScriptModule functions)

To be amended.

---

### See Also

#1007 - a previous ticket on subject, I guess it may be treated as a separate task for replacing Editor's code that does all the above with calls to standalone tool(s) when it's ready.

ivan-mogilko:
Resolves #1146 .

**agf2autoash** - stands for "AGF to Autogenerated ASH" - parses Game.agf (xml) and generates main game header with all the entities and arrays declarations.

_Input:_ game project (*.agf).
_Output:_ generated script header with game declarations.

Usage: `agf2autoash Game.agf _AutoGenerated.ash`

**agf2glvar** - parses Game.agf and generates ash/asc with user's global variables declarations and initialization.

_Input:_ game project (*.agf).
_Output:_ generated script header and body.

Usage: : `agf2glvar Game.agf _GlobalVariables.ash _GlobalVariables.asc`

XML parsing is done by a [tinyxml2 library](https://github.com/leethomason/tinyxml2) - a C++ library consisting of 1 header and 1 cpp file. I placed it inside the repository in this PR, but guess it may be done as a submodule too, since it's present on github (idk if worth the hassle, and we have other lib sources there already).
*NOTE:* tinyxml2's readme sais it works exclusively with utf-8; Editor still saves user text in ANSI encoding in game.agf. This may not be a problem if the library reads the element values as-is not trying to convert it to something, but it's best to double check this.

AGF reading is done by a small library of "parser classes" which are currently barebones, as this tool required mostly entity script names. Their composition is not final, and likely change when more sophisticated tools are made (e.g. #1188).

Builds on Windows and Linux. For MSVS it uses "Tools" solution. For Linux there are separate Makefiles per tool.


ivan-mogilko:
Added separate **agf2glvar** tool which generates global variables from same project file (Game.agf).
(Updated description in the first comment)

rofl0r:
i'd be interested to see output of the latter tool, especially contents of .asc file

ivan-mogilko:
> i'd be interested to see output of the latter tool, especially contents of .asc file

Here's an artificial example:

[test--ashgen.zip](https://github.com/adventuregamestudio/ags/files/6373267/test--ashgen.zip)

Contains:
- Game.agf,
- autogen.ash - game object declarations,
- glvar.ash/asc pair - global variables.

To clarify, these "global variables" are created in a separate dialog inside the editor, this is done purely for convenience of beginner users who may have trouble with properly declaring and importing/exporting them.

rofl0r:
> Here's an artificial example:

thanks. interesting, i wasn't aware of special game_start func. oddly enough, your game last n'furious contains a different version of it in gamescript5,8,14-18.s and globalscript.s (as disassembled by agsutils).

btw autogen.ash contains this which looks like a bug `#define {STATUSLINE} FindGUIID("{STATUSLINE}")`

ivan-mogilko:
> btw autogen.ash contains this which looks like a bug #define {STATUSLINE} FindGUIID("{STATUSLINE}")

True, this carried over from C# formatting which uses {x} for placeholders, I fixed it in the latest commit.

ivan-mogilko:
Made an update to this.

Restructured code, now there's a clear separation between reading from xml to temporary data structs and generating scripts. Script generation is done using structs instead of continiously reading xml document. I think this makes these functions more flexible.

Fixed missing Views in auto header.

Issue #1008, opened 2019-12-12
Editor: store Dialog scripts outside of Game.agf


ivan-mogilko:
Among strange design decisions, Dialog scripts are not stored in separate file(s), but as a huge item group inside Game.agf (xml). If script compilers and other tools will be standalone programs, we must move them outside in their respective file or group of files.

This may be also simply convenient to user, because right now there's no way to edit them outside of editor except for searching for them and hacking XML document.

An important nuance is, that dialog scripts is a super-set of AGS script. It supports regular commands, but also has it's own. Therefore standalone script compiler won't be enough to compile them, there also has to be a preprocessor to replace special dialog commands with the proper AGS script commands. (Currently this is done by an embedded converter inside Editor program)

On another hand, maybe the final workflow could support both dialog scripts in special format and in the ags script format.

ericoporto:
`.asd` as the suggested extension name.

ivan-mogilko:
> .asd as the suggested extension name.

The last letter of `asc` & `ash` indicate whether they are script body or header, which may probably cause inconsistency with `asd`.

Another option is `.dsc` as in "Dialog SCript" (as opposed to ".asc" as in "Ags SCript").

ericoporto:
Linux doesn't care for extensions, it uses magic numbers to determine a filetype primarily.

File command is the main implementation of this:
https://github.com/file/file

Note that magic numbers are not needed if the file is not binary (like .scm) and is instead a text file. If the file is a text file, then extensions are useful sugar to determine a filetype, for an external IDE that needs to know how to color a text file, like Atom when loading `.ash` and `.asc` files. But there are no limits for extension or naming of the file (like `.gitignore`). Files that start with a dot are also considered hidden by file explorers.

morganwillcock:
I'd probably prefer `.asd` so that an editor can be configured to match with `as{c,d,h}`. Maybe the trickier choice is deciding on the file encoding.

I think this change may actually be quite complicated though, since once the content is outside of the XML file it could be modified independently of it. So unfortunately you can't really just write the text to a file and say the job is done. More likely it would be necessary to define the dialog script as something that the script editor can open to get all of the same reload behaviour and save behaviour as regular script or script header, which might mean abstracting out anything that doesn't now apply to all types of editable scripts, and anything that checks 'all scripts' (like the game writer/compiler) would need to differentiate too.

I think this is worth doing though, but whoever tries just needs to make sure that it is treated no differently to another editable text file.

[...]

PR #1269, opened 2021-05-01
Tool: agf2dlgasc - generate ags scripts out of Game.agf's dialog scripts

ivan-mogilko:
Related to #1008, but does not resolve it (apparently I forgot to create a separate ticket about dialog conversion tool).

agf2dlgasc - stands for "AGF to Dialog ASC" - parses Game.agf (xml) and generates a merged dialog script converting dialog scripts found in the game into standard ags script format.

*Input:* game project (*.agf).
*Output:* generated script body.

Usage: `agf2glvar Game.agf __DialogScripts.asc`

---

For the better context, Dialog Scripts are separate entities in AGS currently, written in their own dialect. Prior to AGS 2.72 they were also processed separately. Today they are converted into regular script and then compiled along with the rest of game scripts.

Here's an article on dialogs: https://github.com/adventuregamestudio/ags-manual/wiki/Settingupthegame#conversations
Scrolling down a bit you will see the syntax explanation.

Conversion is done by replacing special dialog commands with ags script commands, and dialogs themselves are wrapped in regular ags functions with names matching certain convention. This is how it's done in the Editor:
https://github.com/adventuregamestudio/ags/blob/ags3/Editor/AGS.Editor/Utils/DialogScriptConverter.cs

For this standalone tool I practically copied algorithm from DialogScriptConverter.cs. The output from the tool supposed to be nearly identical to the Editor, perhaps with minor identation differences.

---

For now this tool works by reading everything from Game.agf. I had thoughts about maybe making a separate utility for extracting dialog scripts into files, but then decided not to. Because of #1008 our project format may change in a way that I cannot predict right now. So it's perhaps best to keep working with AGF until #1008 will be actually in works and situation becomes more clear.

I did only basic tests so far, but there are numerous dialog commands which have to be all checked. In the end we of course should have automatic tests.

### DIFF
--- a/Common/util/memorystream.cpp
+++ b/Common/util/memorystream.cpp
@@ -1,0 +1,156 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+#include "util/memorystream.h"
+#include <algorithm>
+#include <string.h>
+
+namespace AGS
+{
+namespace Common
+{
+
+MemoryStream::MemoryStream(const std::vector<char> &cbuf, DataEndianess stream_endianess)
+    : DataStream(stream_endianess)
+    , _cbuf(&cbuf.front())
+    , _len(cbuf.size())
+    , _buf(nullptr)
+    , _mode(kStream_Read)
+    , _pos(0)
+{
+}
+
+MemoryStream::MemoryStream(const String &cbuf, DataEndianess stream_endianess)
+    : DataStream(stream_endianess)
+    , _cbuf(cbuf.GetCStr())
+    , _len(cbuf.GetLength())
+    , _buf(nullptr)
+    , _mode(kStream_Read)
+    , _pos(0)
+{
+}
+
+MemoryStream::MemoryStream(std::vector<char> &buf, StreamWorkMode mode, DataEndianess stream_endianess)
+    : DataStream(stream_endianess)
+    , _len(buf.size())
+    , _buf(&buf)
+    , _mode(mode)
+    , _pos(buf.size())
+{
+    _cbuf = (mode == kStream_Read) ? &buf.front() : nullptr;
+}
+
+MemoryStream::~MemoryStream()
+{
+}
+
+void MemoryStream::Close()
+{
+    _cbuf = nullptr;
+    _buf = nullptr;
+    _pos = -1;
+}
+
+bool MemoryStream::Flush()
+{
+    return true;
+}
+
+bool MemoryStream::IsValid() const
+{
+    return _cbuf != nullptr || _buf != nullptr;
+}
+
+bool MemoryStream::EOS() const
+{
+    return _pos >= _len;
+}
+
+soff_t MemoryStream::GetLength() const
+{
+    return _len;
+}
+
+soff_t MemoryStream::GetPosition() const
+{
+    return _pos;
+}
+
+bool MemoryStream::CanRead() const
+{
+    return (_cbuf != nullptr) && (_mode == kStream_Read);
+}
+
+bool MemoryStream::CanWrite() const
+{
+    return (_buf != nullptr) && (_mode == kStream_Write);
+}
+
+bool MemoryStream::CanSeek() const
+{
+    return CanRead(); // TODO: support seeking in writable stream?
+}
+
+size_t MemoryStream::Read(void *buffer, size_t size)
+{
+    if (EOS()) { return 0; }
+    soff_t remain = _len - _pos;
+    assert(remain > 0);
+    size_t read_sz = std::min((size_t)remain, size);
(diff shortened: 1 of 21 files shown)


ivan-mogilko:
Rebased and restructed code to be consistent with the previous two tools.

I think this is generally working, but I haven't tested all the possible dialog commands yet.

rofl0r:
hmm, i thought it would be more useful to have 2 steps here:

- extract dialog scripts in textual form
- turn dialog script into ags script

agsutils already supports the former (with game28.dta files) so having the latter available as a separate tool would allow to use your tool with those too.

ivan-mogilko:
Having two tools would make sense, but I mentioned my concerns about this above:
> For now this tool works by reading everything from Game.agf. I had thoughts about maybe making a separate utility for extracting dialog scripts into files, but then decided not to. Because of #1008 our project format may change in a way that I cannot predict right now.

I guess current tool could be split up into two, except then we'll create new format for intermediate dialog scripts, which may or not match future dialog script format. If it won't then the future versions of the tool would need to support both.

> agsutils already supports the former (with game28.dta files) so having the latter available as a separate tool would allow to use your tool with those too.

I just need to mention that the converter currently uses extra information from Game.agf:
* list of character names;
* number of dialog options;
* couple of extra game options; but these are necessary only for the newest games and not present in old compiled games where dialog scripts are found.

The first two are required for the script validation, to check that such characters actually exist. Guess there are two options here: the tool may accept a parameter disabling this validation, and/or accept this information in some alternate way.
As soon as the script gets into the script compiler any wrong character references will be detected, but not the dialog option numbers (as they are not named, and simply a literal number).

ivan-mogilko:
It has to be decided what the conversion tool input is. Whether the tool is creating whole merged dialog script or converts 1 dialog at a time. In the first case it needs to have either a path to where dialog texts are in separate files, and a naming rule (e.g. `dialogNNN.txt`). In the second case there has to be another operation that merges them all.

The final merged script requires additional chunk of script prepended on top, that is always constant (at least for now).
https://github.com/ivan-mogilko/ags-refactoring/blob/ags3tools--agf2dlgasc/Tools/data/dialogscriptconv.cpp#L28

rofl0r:
> I just need to mention that the converter currently uses extra information from Game.agf:

oh. here's an example of dialogscript extracted of larry2 via agsutils (dialogscript000.ads):
```
// dialog script file
@S  // dialog startup entry point
return
@1  // option 1
narrator:"Sure," you reply, "always time for a little nip with some seedy-looking stranger!"
run-script 1
stop
@2  // option 2
narrator:"Thanks anyway, buster, but I'm on the wagon."
stop
@3  // option 3
narrator:"Do you have a license for that breath of yours?" you ask him.
narrator:He ignores you. "How 'bout a little nip?"
stop

@4  // option 4
```

so without additional contextual info it would be impossible to turn this into .asc ?

ivan-mogilko:
> so without additional contextual info it would be impossible to turn this into .asc ?

It is possible, like I mentioned this information is used only for validation of character and dialog names.

rofl0r:
> Dialog scripts have character names in lowercase, also with the starting "c" letter is stripped for historical reasons.

for same game above, agscriptxtract produces following "builtinscriptheader.ash": https://0x0.st/-BGO.ash

iirc one of the tools you already committed to repo creates a header that's quite similar. couldn't this header be used (included) from the generated asc file rather than resorting to poke at more xml of agf ?

> Also, I don't know what is the use case of converting dialogs extracted from the old compiled game file.

well i guess for someone trying to fix a bug in the old game it could be more convenient to work with .asc rather than .ads. but you're right: i can produce script file source only for older game versions that store these in encrypted form, newer ones don't. so you're right in that code produced for recent engine version that uses different function names this would not be super helpful.

btw is each dialogscript emitted into a separate .asc file, or are all of them combined into a single one ? there seems to be only one compiler dialogscript per game.

ivan-mogilko:
I made a decision to add this tool as it is now.

There are too many problems that require working on, and I am already lagging behind my plans.

In the future this tool may be remade, split, complemented with others, etc.

rofl0r:
> It's a question of the form in which it receives this list. In theory it may be read by parsing agf, or by parsing a script header, or by parsing a compiled game data, or by receiving an explicit list as txt, and so on.

well, my idea was since the header already exist, it could just be included so nothing at all has to be parsed (apart from dialogscript code).
e.g. output looks like

```c
#include "builtinscriptheader.ash"
// code of converted script follows, using ids like cEgo etc assuming header provides them
```

ivan-mogilko:
> well, my idea was since the header already exist, it could just be included so nothing at all has to be parsed (apart from dialogscript code).

I do not think we understand each other here...

What I was talking about is the problem of *finding* the real character script name *during* the conversion.

For example, in a dialog script you have:

`roger: I say something`

Here "roger" has to be converted to the real character script name. But where to find it? You got to parse *something* that has a list of existing characters. Only after you have that list you may replace "roger" with "cRoger", or anything else that matches, according to the rules.

*After* the converted script is generated, any necessary headers may be attached of course.

rofl0r:
> For example, in a dialog script you have:
> roger: I say something

ah, i was under the impression you could always just prepend 'c' and turn first char into uppercase. nevermind then.

messengerbag:
For AGS 4, I'd suggest getting rid of all of that nonsense entirely. Entities should have one consistent name throughout, and no part of the script should have to guess at variations with/without systems-Hungarian prefixes.

ivan-mogilko:
@messengerbag it's best to open a separate suggestion ticket, or add this proposal here: #1008

PR #1277, opened 2021-05-15
misleading set outline in fonts.cpp

ericoporto:
was not really setting outline type but only usage of this function is in the LucasFan-Font fix and it uses the value FONT_OUTLINE_AUTO as parameter so this commit doesn't actually change anything other than correctnes.


ivan-mogilko:
I guess this is remains of some quick refactoring.

PR #1278, opened 2021-05-17
Functional open rooms

persn:
Reference #469

I still need to cleanup some code and do other improvements, but the most important mechanics are in place so I'm creating a draft PR so people can begin testing.

image

- When game is loaded into the editor an upgrade process will trigger. The code is handled from the editor but I imagine it'll be replaced with a CLI tool for processing rooms or something similar in the future. The upgrade process will extract room data and images and save them to disk under a directory `Rooms/{room-number}/`
- Room editor is refactored to read data directly from the new assets in the `Rooms/` directory. .crm files are mostly ignored until compile time.
- .crm files will now be generated from the data found in the `Rooms/` directory when rooms are saved/compiled, so with this PR these files should be safe to .gitignore.
- File watcher code from script editor has been refactored into a utility class with reusable file watching helper code. This means that the script editor file watching should also be tested probably.
- Fixed some minor bugs with room xml serialization
- Removed transparency from mask palette color 0 (eraser area) to not be transparent. It ruined the image when exporting and reimporting masks as .png files. I was never able to figure out exactly what's happening but most software I tested with gets quirky when 8-bit pngs have transparency. A minor rant about my hair pulling experience here https://github.com/adventuregamestudio/ags/issues/1113#issuecomment-794506794

PS: Remember to take backup of your game before upgrading

### DIFF
--- a/Editor/AGS.CScript.Compiler/Preprocessor.cs
+++ b/Editor/AGS.CScript.Compiler/Preprocessor.cs
@@ -40,7 +40,7 @@ namespace AGS.CScript.Compiler
 		public string Preprocess(string script, string scriptName)
 		{
 			StringBuilder output = new StringBuilder(script.Length);
-			output.AppendLine(Constants.NEW_SCRIPT_MARKER + scriptName + "\"");
+			output.AppendLine(Constants.NEW_SCRIPT_MARKER + scriptName.Replace(@"\", @"\\") + "\"");
 			StringReader reader = new StringReader(script);
 			string thisLine;
 			_scriptName = scriptName;
--- a/Editor/AGS.Editor.Tests/ScriptCompiler/PreprocessorTests.cs
+++ b/Editor/AGS.Editor.Tests/ScriptCompiler/PreprocessorTests.cs
@@ -345,5 +345,15 @@ Display(""test"");
             Assert.That(preprocessor.Results.Count, Is.EqualTo(1));
             Assert.That(preprocessor.Results[0].Code, Is.EqualTo(ErrorCode.EndIfWithoutIf));
         }
+
+        [TestCase("room2.asc", "room2.asc")]
+        [TestCase(@"Rooms\2\room2.asc", @"Rooms\\2\\room2.asc")]
+        public void FormatsScriptName(string scriptName, string expected)
+        {
+            IPreprocessor preprocessor = CompilerFactory.CreatePreprocessor(AGS.Types.Version.AGS_EDITOR_VERSION);
+            string res = preprocessor.Preprocess(string.Empty, scriptName);
+            AssertStringEqual(res, $@"""__NEWSCRIPTSTART_{expected}""
+");
+        }
     }
 }
--- a/Editor/AGS.Editor.Tests/Types/RoomTests.cs
+++ b/Editor/AGS.Editor.Tests/Types/RoomTests.cs
@@ -213,7 +213,7 @@ namespace AGS.Types
             int topEdgeY, int bottomEdgeY, int number, string description)
         {
             string xml = $@"
-            <Room>
+            <Room Version=""{Room.LATEST_XML_VERSION}"">
                 <MaskResolution>{maskResolution}</MaskResolution>
                 <BackgroundCount>{backgroundCount}</BackgroundCount>
                 <GameID>{gameId}</GameID>
@@ -255,6 +255,8 @@ namespace AGS.Types
             XmlDocument doc = new XmlDocument();
             doc.LoadXml(xml);
             _room = new Room(doc.SelectSingleNode("Room"));
+
+            Assert.That(Room.LATEST_XML_VERSION, Is.EqualTo(doc.DocumentElement.Attributes["Version"].Value));
 
             Assert.That(_room.MaskResolution, Is.EqualTo(maskResolution));
             Assert.That(_room.BackgroundCount, Is.EqualTo(backgroundCount));
@@ -316,6 +318,7 @@ namespace AGS.Types
 
             XmlDocument doc = _room.ToXmlDocument();
 
+            Assert.That(doc.DocumentElement.Attributes["Version"].Value, Is.EqualTo(Room.LATEST_XML_VERSION));
             Assert.That(doc.SelectSingleNode("/Room/MaskResolution").InnerText, Is.EqualTo(maskResolution.ToString()));
             Assert.That(doc.SelectSingleNode("/Room/BackgroundCount").InnerText, Is.EqualTo(backgroundCount.ToString()));
             Assert.That(doc.SelectSingleNode("/Room/GameID").InnerText, Is.EqualTo(gameId.ToString()));
--- a/Editor/AGS.Editor.Tests/Types/RoomTests.cs
+++ b/Editor/AGS.Editor.Tests/Types/RoomTests.cs
@@ -170,6 +170,21 @@ namespace AGS.Types
             Assert.That(_room.BottomEdgeY, Is.EqualTo(bottomEdgeY));
         }
 
+        [TestCase(RoomAreaMaskType.Hotspots, Room.MAX_HOTSPOTS)]
+        [TestCase(RoomAreaMaskType.Regions, Room.MAX_REGIONS)]
+        [TestCase(RoomAreaMaskType.WalkableAreas, Room.MAX_WALKABLE_AREAS)]
+        [TestCase(RoomAreaMaskType.WalkBehinds, Room.MAX_WALK_BEHINDS)]
+        public void GetsMaskMaxColor(RoomAreaMaskType mask, int expected)
+        {
+            Assert.That(Room.GetMaskMaxColor(mask), Is.EqualTo(expected));
+        }
+
+        [Test]
+        public void GetsMaskMaxColorThrowsExceptionWithIllegalMask()
+        {
+            Assert.Throws<ArgumentException>(() => Room.GetMaskMaxColor(RoomAreaMaskType.None));
+        }
+
         [TestCase(1, RoomAreaMaskType.WalkBehinds, 1.00)]
         [TestCase(2, RoomAreaMaskType.WalkBehinds, 1.00)]
         [TestCase(1, RoomAreaMaskType.Hotspots, 1.00)]
--- a/Editor/AGS.Editor.Tests/Types/RoomTests.cs
+++ b/Editor/AGS.Editor.Tests/Types/RoomTests.cs
@@ -217,7 +217,17 @@ namespace AGS.Types
                 <Properties/>
                 <Number>{number}</Number>
                 <Description xml:space=""preserve"">{description}</Description>
-                <Interactions/>
+                <Interactions>
+                    <Event Index=""0"">room_LeftEdge</Event>
+                    <Event Index=""1"" />
+                    <Event Index=""2"" />
+                    <Event Index=""3"" />
+                    <Event Index=""4"" />
+                    <Event Index=""5"" />
+                    <Event Index=""6"" />
+                    <Event Index=""7"" />
+                    <Event Index=""8"" />
+                </Interactions>
                 <Messages/>
                 <Objects/>
                 <Hotspots/>
@@ -249,6 +259,16 @@ namespace AGS.Types
             Assert.That(_room.Number, Is.EqualTo(number));
             Assert.That(_room.Description, Is.EqualTo(description));
 
+            Assert.That(_room.Interactions.ScriptFunctionNames[0], Is.EqualTo("room_LeftEdge"));
+            Assert.That(_room.Interactions.ScriptFunctionNames[1], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[2], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[3], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[4], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[5], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[6], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[7], Is.Null);
+            Assert.That(_room.Interactions.ScriptFunctionNames[8], Is.Null);
+
             File.Delete("TestScript.asc");
         }
 
@@ -276,6 +296,9 @@ namespace AGS.Types
             _room.BottomEdgeY = bottomEdgeY;
             _room.Number = number;
             _room.Description = description;
+
+            _room.Interactions.ScriptFunctionNames[0] = "room_LeftEdge";
+
             XmlDocument doc = _room.ToXmlDocument();
 
             Assert.That(doc.SelectSingleNode("/Room/MaskResolution").InnerText, Is.EqualTo(maskResolution.ToString()));
@@ -295,6 +318,18 @@ namespace AGS.Types
             Assert.That(doc.SelectSingleNode("/Room/BottomEdgeY").InnerText, Is.EqualTo(bottomEdgeY.ToString()));
             Assert.That(doc.SelectSingleNode("/Room/Number").InnerText, Is.EqualTo(number.ToString()));
             Assert.That(doc.SelectSingleNode("/Room/Description").InnerText, Is.EqualTo(description.ToString()));
+
+            var interactions = doc.SelectSingleNode("/Room/Interactions").SelectNodes("Event");
+            
+            Assert.That(interactions[0].InnerText, Is.EqualTo("room_LeftEdge"));
+            Assert.That(interactions[1].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[2].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[3].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[4].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[5].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[6].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[7].InnerText, Is.EqualTo(""));
+            Assert.That(interactions[8].InnerText, Is.EqualTo(""));
         }
     }
 }
--- a/Editor/AGS.Editor.Tests/Types/RoomTests.cs
+++ b/Editor/AGS.Editor.Tests/Types/RoomTests.cs
@@ -1,5 +1,6 @@
 ﻿using NUnit.Framework;
 using System;
+using System.IO;
 using System.Xml;
 
 namespace AGS.Types
@@ -224,6 +225,8 @@ namespace AGS.Types
                 <WalkBehinds/>
                 <Regions/>
             </Room>";
+            Directory.CreateDirectory(Path.Combine("Rooms", $"{number}"));
+            File.WriteAllText(Path.Combine("Rooms", $"{number}", $"room{number}.asc"), "Test placeholder");
             XmlDocument doc = new XmlDocument();
             doc.LoadXml(xml);
             _room = new Room(doc.SelectSingleNode("Room"));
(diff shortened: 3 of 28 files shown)


ivan-mogilko:
Was wondering about this progress only yesterday :-).

persn:
Bugs found

1: Editor will crash on Objects or Character view in Room Editor when the sprite it's trying to display is anything but a 32-bit image.

2: Editor will show blue cup (or whatever is sprite number 0 in your project) for every room object because of incorrect upgrade

ivan-mogilko:
Sorry for delay, I must find some time to test this soon.

persn:
> Sorry for delay, I must find some time to test this soon.

No rush, there still things I need to do. I just opened the PR so that if people who are dying to test this can do so ;)

persn:
What's the sensible way of checking if the upgrade should be performed? Right now it just checks if the game project directory has a folder called "Rooms", and if it's missing it'll perform the upgrade. This is kind of lazy though as there might be a directory with that name there already for whatever reasons.

ivan-mogilko:
> What's the sensible way of checking if the upgrade should be performed? Right now it just checks if the game project directory has a folder called "Rooms", and if it's missing it'll perform the upgrade. This is kind of lazy though as there might be a directory with that name there already for whatever reasons.

First thing that comes to mind is a project version (LATEST_XML_VERSION_INDEX), this is what we check now. Of course that may seem bit off, as rooms are not part of Game.agf, but I guess we may say that in the current project structure Game.agf controls everything else too, it's like a meta information about the game source.

persn:
> First thing that comes to mind is a project version (LATEST_XML_VERSION_INDEX), this is what we check now. Of course that may seem bit off, as rooms are not part of Game.agf, but I guess we may say that in the current project structure Game.agf controls everything else too, it's like a meta information about the game source.

I'll try that, thank you

ericoporto:
@persn please rebase to latest ags4 and push force, previous CI had some bugs that will fail your build.

persn:
@ericoporto I'll do it later unless there's a very important reason I have to do it right away

ericoporto:
No important reason other than it would be much easier to test.

ivan-mogilko:
I have a question: this PR seem to enforce PNG format for all room images: backgrounds and masks. Is this intended to restrict these to png, or it is a temporary state?

persn:
.PNG because AFAIK it's the most conveniently available compressed lossless format. It's not "intended" to restrict anything, but it does restrict it right now simply because of convenience, it was the easy way.

ivan-mogilko:
Things I noticed so far (will post more as I find anything else):

1. If I create "Rooms/ROOM_NUMBER" subfolder in the old project, and keep this subfolder open in the Windows Explorer, then try opening the project in the new editor, the editor will fail to upgrade the game with "Path access denied" exception and refuse to open the project. I think we had similar problem with compiling the game.
2. If I delete room script file (*.asc) by hand, trying to open this script from the editor will result in exception followed by "Serious error" message. Perhaps it could create a new empty file if script is missing (with a warning)?
3. If I double click on the room editor opens both the room editor and script tabs. Idk if it's intended? Additionally, when this happens both documents are marked as "modified" for some reason (and editor asks if I like to save them when I close them). This does not happen if I open room script separately.
4. If after p3 you save the room, editor sometimes tells me that the script was modified externally for some reason. Regardless of whether I choose to reload or not it will keep showing me this notification for about 6-8 times more.

ivan-mogilko:
In regards to the data.xml, would it be possible to add a header similar to the one in Game.agf, with encoding and document tag with format version? I think latter might be useful for future updates, while encoding might be helpful if we change to unicode support - so that we know what enconding/locale to convert game texts from.

ivan-mogilko:
May be bit far fetched, but I realized something about the update process. If I understood right, if there existed a "Rooms" folder you rename it to "Room_Backup". If this folder contained sprite or audio sources - the paths to them will break. So editor will also need to run checks on all the existing sprites and audio files and remap paths if necessary.

persn:
@ivan-mogilko I think I've addressed everything you listed now except for..

> May be bit far fetched, but I realized something about the update process. If I understood right, if there existed a "Rooms" folder you rename it to "Rooms_backup". If this folder contained sprite or audio sources - the paths to them will break. So editor will also need to run checks on all the existing sprites and audio files and remap source paths if necessary.
EDIT: maybe we could have a generic procedure that does that - in case we update/migrate other folders.

I've been thinking about this and I tried implementing it but it became complicated because I would likely need to distinguish between absolute paths and relative paths, preserve them and so on. The possible problems that I was able to think of grew fast, so I don't think that this should be a part of this PR, but maybe it should be an isolated issue that need to be resolved on the AGS4 branch

ivan-mogilko:
I tried this out, and following two issues are still not fully fixed:

> 2. If I delete room script file (*.asc) by hand, trying to open this script from the editor will result in exception followed by "Serious error" message.

This is resolved only if you click on the room first, then it tells that the empty file is created. But if you first click on room.asc in the project tree, then it first displays the "Unhandled exception" message again, before opening an empty file.

Additionally, I noticed that the missing file is not created on disk until you type something into it first. Even if you just hit the "save" this empty script will not save to file for some reason. Idk if this is a new bug, or has something to do with how Scintilla works.

> 5. ... if I actually modify the script externally, it keeps showing me this notification for about 6-8 times more, regardless of what I choose (reload or not).

This unfortunately is not fixed and still happening every time.

I also have a question about room's xml format, currently there's no meta information, such as format version. Do you think we need it, and if not, how will we deal with upgrading the room data?

ivan-mogilko:
More things I found, deleting parts of the room on disk, then trying to build the game, makes Editor display unhandled exception with "Index out of range":

<pre>
в System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   в System.Collections.Generic.List`1.get_Item(Int32 index)
   в AGS.Editor.Components.RoomsComponent.SaveCrm() в D:\Dev\ags\main\Editor\AGS.Editor\Components\RoomsComponent.cs:строка 2171
   в AGS.Editor.Components.RoomsComponent.AGS.Types.IRoomController.Save() в D:\Dev\ags\main\Editor\AGS.Editor\Components\RoomsComponent.cs:строка 1589
   в AGS.Editor.Components.RoomsComponent.SaveRoomOnThread(Object parameter) в D:\Dev\ags\main\Editor\AGS.Editor\Components\RoomsComponent.cs:строка 631
</pre>

ivan-mogilko:
Don't know what is the intended behavior, but deleting a room won't delete the room files.


Another thing, I don't know what this relates to , but MSVS keeps reporting following (handled) exception from time to time, which I don't remember seeing before:

<pre>
Exception thrown: 'System.ObjectDisposedException' in mscorlib.dll
Additional information: Safe handle has been closed
If there is a handler for this exception, the program may be safely continued.
</pre>

persn:
> This is resolved only if you click on the room first, then it tells that the empty file is created. But if you first click on room.asc in the project tree, then it displays the "Unhandled exception" message again, before opening an empty file.

Won't fix. This is not a regression introduced in this PR as it also happens on the `master` branch. Feel free to open an issue for it though, and maybe it can get fixed for 3.x version as well.

> Additionally, I noticed that the missing file is not created on disk until you type something into it first. Even if you just hit the "save" this empty script will not save to file for some reason. Idk if this is a new bug, or has something to do with how Scintilla works.

Same as above.

> This unfortunately is not fixed and still happening every time.

I don't understand how this is possible.. If I understand what you're saying the editor is not in focus but you still get the messages? The code says to not use the editor when it's not in focus. Can you record what you're doing and post it here so there's no miscommunication on what is happening?

> I also have a question about room's xml format, currently there's no meta information, such as format version. Do you think we need it, and if not, how will we deal with upgrading the room data?

I have no idea how to handle this, don't have much experience on that. I can just add it if we can agree on the spec.. Is a simple incrementing number like `<Room Version="1">` sufficient?

> More things I found, when the room has multiple background frames, deleting some of them from disk makes Editor display unhandled exceptions with "Index out of range" when you try opening them in the room editor, or rebuilding the game:

I added a fail safe check that validates room background count from xml with actual number of images found and loaded. It should trigger when the room is loaded.

> Don't know what is the intended behavior, but deleting a room won't delete the room files.

Seems like I forgot to rewrite this functionality, should be fixed.

> Another thing, I don't know what this relates to , but MSVS keeps reporting following (handled) exception from time to time, which I don't remember seeing before:

A quick glance and I can't see a single occurence. Do you do anything specific to make them show?

persn:
@AlanDrake Did you post to the right thread? I can't tell how your comment relates to this PR

AlanDrake:
> @AlanDrake Did you post to the right thread? I can't tell how your comment relates to this PR

Yes, it's about the script name marker being inserted in the headers, since it now can have a relative path it may end up looking like this:
`__NEWSCRIPTSTART_Rooms\8\room8.asc`

As you can see the backslashes need to be escaped, so you can just replace that line with my first suggestion. The issue with the new compiler was resolved, so it's safe now.

ivan-mogilko:
@persn 

>> This is resolved only if you click on the room first, then it tells that the empty file is created. But if you first click on room.asc in the project tree, then it displays the "Unhandled exception" message again, before opening an empty file.

> Won't fix. This is not a regression introduced in this PR as it also happens on the master branch. Feel free to open an issue for it though, and maybe it can get fixed for 3.x version as well.

Tested and opened a ticket: #1419

---

> I don't understand how this is possible.. If I understand what you're saying the editor is not in focus but you still get the messages? The code says to not use the editor when it's not in focus. Can you record what you're doing and post it here so there's no miscommunication on what is happening?

Recorded video: 
[openroomtest001.zip](https://github.com/adventuregamestudio/ags/files/7220773/openroomtest001.zip)

--- 

>> I also have a question about room's xml format, currently there's no meta information, such as format version. Do you think we need it, and if not, how will we deal with upgrading the room data?

> I have no idea how to handle this, don't have much experience on that. I can just add it if we can agree on the spec.. Is a simple incrementing number like `<Room Version="1">` sufficient?

Yes, I think if it were a version number similar to the Game.agf, that would work.

persn:
> Tested and opened a ticket: #1419

Thanks

> Recorded video:

Thanks, I'm not exactly sure what is happening, but I pushed an experimental change that should make the file watcher hopefully more stable. Please try it out for me

> Yes, I think if it were a version number similar to the Game.agf, that would work.

Should be fixed now

ivan-mogilko:
> > Recorded video:
> 
> Thanks, I'm not exactly sure what is happening, but I pushed an experimental change that should make the file watcher hopefully more stable. Please try it out for me

No, I'm afraid this is still happening.

What about others, am I the only person who gets this effect?

ivan-mogilko:
I think I understand what's going on.

You have this in the FileWatcher:
<pre>
        private void OnMainWindowActivated(object sender, EventArgs e)
        {
            if (_changed)
            {
                LoadFile();
                _changed = false;
            }
        }
</pre>

It looks like this method is getting called over and over again as soon as I click on any button in the question dialog, before the `_changed` has a chance to reset.

persn:
@ivan-mogilko ah great. That could've taken me a long time to figure out so thank you. If the problem is what you say it is maybe a lock would be just as good? Could you try that?

```csharp
       private static readonly _activatedWindowLock = new object();

       private void OnMainWindowActivated(object sender, EventArgs e)
        {
            lock (_activatedWindowLock)
            {
            if (_changed)
            {
                LoadFile();
                _changed = false;
            }
            }
        }
```

ivan-mogilko:
@persn the lock does not work, because apparently the consecutive OnMainWindowActivated calls are made within same thread,  this is the stack:

<pre>
 	AGSEditor.exe!AGS.Editor.FileWatcher.OnMainWindowActivated(object sender, System.EventArgs e) Line 145	C#
 	AGSEditor.exe!AGS.Editor.GUIController._mainForm_OnMainWindowActivated(object sender, System.EventArgs e) Line 848	C#
 	AGSEditor.exe!AGS.Editor.frmMain.frmMain_Activated(object sender, System.EventArgs e) Line 69	C#
 	System.Windows.Forms.dll!System.Windows.Forms.Form.OnActivated(System.EventArgs e)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Form.Active.set(bool value)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Form.WmActivate(ref System.Windows.Forms.Message m)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Form.WndProc(ref System.Windows.Forms.Message m)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Control.ControlNativeWindow.OnMessage(ref System.Windows.Forms.Message m)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Control.ControlNativeWindow.WndProc(ref System.Windows.Forms.Message m)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.NativeWindow.Callback(System.IntPtr hWnd, int msg, System.IntPtr wparam, System.IntPtr lparam)	Unknown
 	[Native to Managed Transition]	
 	user32.dll!75a162fa()	Unknown
 	[Frames below may be incorrect and/or missing, no symbols loaded for user32.dll]	
 	user32.dll!75a16d3a()	Unknown
 	user32.dll!75a16de8()	Unknown
 	user32.dll!75a16e44()	Unknown
 	ntdll.dll!77c0011a()	Unknown
 	user32.dll!75a22fc5()	Unknown
 	user32.dll!75a22fb0()	Unknown
 	System.Windows.Forms.ni.dll!51b63318()	Unknown
 	[Managed to Native Transition]	
 	System.Windows.Forms.dll!System.Windows.Forms.Application.ThreadWindows.Enable(bool state)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Application.ThreadContext.EnableWindowsForModalLoop(bool onlyWinForms, System.Windows.Forms.ApplicationContext context)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Application.ThreadContext.EndModalMessageLoop(System.Windows.Forms.ApplicationContext context)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.Application.EndModalMessageLoop()	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.MessageBox.ShowCore(System.Windows.Forms.IWin32Window owner, string text, string caption, System.Windows.Forms.MessageBoxButtons buttons, System.Windows.Forms.MessageBoxIcon icon, System.Windows.Forms.MessageBoxDefaultButton defaultButton, System.Windows.Forms.MessageBoxOptions options, bool showHelp)	Unknown
 	System.Windows.Forms.dll!System.Windows.Forms.MessageBox.Show(string text, string caption, System.Windows.Forms.MessageBoxButtons buttons, System.Windows.Forms.MessageBoxIcon icon)	Unknown
 	AGSEditor.exe!AGS.Editor.GUIController.ShowQuestion(string message, System.Windows.Forms.MessageBoxIcon icon) Line 176	C#
 	AGSEditor.exe!AGS.Editor.FileWatcher.ShouldLoad() Line 200	C#
 	AGSEditor.exe!AGS.Editor.FileWatcher.LoadFile() Line 186	C#
>	AGSEditor.exe!AGS.Editor.FileWatcher.OnMainWindowActivated(object sender, System.EventArgs e) Line 149	C#

</pre>

ivan-mogilko:
@persn Do you have an idea of what is left to be done here? Is anything else requires testing?

ivan-mogilko:
> Manually changing the encoding to utf-8 without BOM makes the room load properly, but then if you change something and save it, it returns to Windows-1252, so perhaps there's a wrong flag somewhere in the room number change that is not setting the expected encoding?

Note: the editor is currently saving everything game data (xml & scripts) in the default code page, which would be "Windows-125*" most often. I think it's correct to keep things consistent, and only switch to UTF-8 when every other component of data supports UTF-8 (#1297).

ivan-mogilko:
Sorry, I did not realize this is not a draft anymore. I will make few more tests with my own projects just in case, and then merge this.

ivan-mogilko:
Found new bug:

1. Create new game using any template;
2. Delete existing room 1.
3. Create new room with the same number 1.
4. The editor will immediately report that the background and masks "has been modified externally", and whether I want to reload it.
5. Clicking on "yes" results in unhandled exception.

ivan-mogilko:
Another thing, I might be wrong, but it seems like the editor _locks the input file_ when adding a room background.

I was editing a background in a paint software, and saved it elsewhere (not in the project folder). After importing this background into the room I edited the same picture more and tried to save it, but the paint software reported error accessing the file. It kept doing so until I closed the editor.

ivan-mogilko:
In regards to watching for the external changes, changing background image in Rooms/N/... is not applied in the room editor until I switch to another room and back. I was not sure how this is supposed to work.

ivan-mogilko:
Another problem: it appears that when you draw room masks externally, they dont have enough `paletteDrawing.Entries[]` array size. E.g. if you only used 2 colours there, then `paletteDrawing.Entries[]` will be size of 2.
However the code in `IRoomController.DrawRoomBackground` expects this array to have all 256 entries, which causes "index out of bounds exception".

ivan-mogilko:
@persn this pr seem to be stalling, which is unfortunate.

Recalling the last issues I found (recent comments above):
* creating room of the same index as recently deleted one during same editor session; results in "background/mask was modified externally" warning that does not work correctly.
* editor possibly locks input file when importing background/mask (unsure if it's this PR's fault, or bug was present before).
* changing room background/mask externally is not automatically applied until you reload the room in the editor.
* editing masks in the Rooms folder may cause exceptions due to image palette's size not matching code assumptions.

If everyone who uses ags4 branch will agree that above problems may be tolerated for a while, we may perhaps merge this, which also would let other people test and fix these bugs?

ivan-mogilko:
It's best if the branch is rebased first though, it may be somewhat outdated by the game format, and also has conflicts in code it seems.

persn:
Before we merge this I'd like to do some adjustments to the room convertion logic as I have learned more about C# async logic, and I discovered that there's a potential deadlock when you're developing a WinForms app.

persn:
@ivan-mogilko I was going to try and do some last minute fixes, however after rebasing I'm getting these error messages in my test project when loading a project into the editor. I tested on `master` as well and the same thing happen there. Presumably it finds some sprite files that doesn't do what it expects but I'm unsure what to do about it because the error message isn't very friendly. Should I open a bug about this?

image

In any event I have rebased this PR now it could be merged if we like.

ivan-mogilko:
@persn this is probably related to the sprite storage enhancements we did several months ago.

Could you send your test project for me to look into?

ivan-mogilko:
@persn I confirm, there's a bug when _saving_ 16-bit sprites in 3.6.0. You may notice e.g. sprite id 2 in your game, its image gets halved.

ivan-mogilko:
Pushed a fix to master and ags4 (b51408b7192013f8baf7c54ffeab5e77128a8bda)

persn:
Seems to work better now. Thanks

persn:
> creating room of the same index as recently deleted one during same editor session; results in "background/mask was modified externally" warning that does not work correctly.

Should be solved now

persn:
> editor possibly locks input file when importing background/mask (unsure if it's this PR's fault, or bug was present before).

It was not present in master. Should be fixed

persn:
> changing room background/mask externally is not automatically applied until you reload the room in the editor.

Cannot reproduce this at all. This has always worked flawlessly for me so I guess it'll be difficult to debug. This is probably an assignment for someone else.

For the record the intended behavior is that the file watcher should update on any change for the room that is open in the room editor. The room editor will update the picture once the draw event has triggered.

Just to rule out any issues you might want to try forcing the draw event to trigger after you made an external change. Put a print line inside the `IRoomController.DrawRoomBackground` in `RoomsComponent.cs` so you're sure it has been executed. Fiddling with the zoom slider or something should make it happen.

ivan-mogilko:
I tried this out again, and found couple of new problems:

1. Convertion to open format is not triggered for older ags4 projects, because it converts only versions before AGS_4_0_0_XML_VERSION_INDEX. I guess you need to increment the version and make "open rooms project" version constant.
2. The room xmls are saved as Encoding.Default.WebName; but now we support utf-8. You could use Game.TextEncoding for the reference instead.
3. ASCII / Unicode conversion does not convert the new rooms, but that's a separate issue, as it was written to work with room.crm files. In general, I think we may make ags4 to be unicode only, unless someone finds a reason not to; and then run ascii->utf8 conversion also at opening the project along with the room format conversion, etc.
4. A real problem with room editing though: changing background in the editor does not resize the masks: they stay in the same resolution, which also breaks in-editor mask painting (the drawing is performed at wrong offsets).

AlanDrake:
Point 4 is solved by  #1435 and has been broken in wait of the open room to be merged because we thought it was happen quickly.

ivan-mogilko:
Ok, I went ahead and merged that; fixed the first two small problems myself.
I will look into p3 soon too.

PR #1280, opened 2021-05-21
workaround invalid OpenGL ES2 functions, need a proper fix

ericoporto:
OpenGL ES 2.0 doesn't have neither `glColor4f` or [`glTexEnvi`](https://www.khronos.org/registry/OpenGL-Refpages/es1.1/xhtml/glTexEnv.xml) used in the code. 

This PR simply guards these with ifdef until a proper code can be figured out, so that the Android port can be built and the CI doesn't fail at packaging the editor. Note that if using the normal blend mode, things will work alright.

I think overall the blend modes could be reimplemented with shaders but this is not trivial since the color to blend need to be in a texture (see https://stackoverflow.com/a/17666532 and https://www.shadertoy.com/view/XdS3RW). Curiously the [gl blend ](https://docs.gl/es2/glBlendFunc)functions ARE available in Open GL ES 2.0, it's just the mentioned functions above that aren't.

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -1172,12 +1172,16 @@ void OGLGraphicsDriver::_renderSprite(const OGLDrawListEntry *drawListEntry, con
         case kBlend_Multiply:
         case kBlend_Burn: // burn is imperfect due to blend mode, darker than normal even when trasparent
             // fade to white
+#if !AGS_OPENGL_ES2 // glTexEnvi and glColor4f are not available on OpenGL ES2, need to rewrite the code here
             glTexEnvi(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_ADD);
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
There are couple of more cases right above your change, guarded by `#ifdef GL_MIN` and `#ifdef GL_MAX`, do you think these may share same macro check?

ericoporto:
Both of these doesn't appear to be available in OpenGL ES 2.0 too, but the current check does work with it, and seems more precise. But it could be changed if it brings more clarity.

ivan-mogilko:
Well, guess it's not necessary then.

ivan-mogilko:
Ah, I missed this at first: could you let `default:` line outside these checks, as switch may need to keep a default label anyway.

ericoporto:
Stealthly rewrote so that the `#if` `#endif` guards don't hide the `default:`. 😀

PR #1284, opened 2021-05-21
Fixed alphabetical sorting of elements in Game.agf

### LINKED ISSUE #1282: Game.agf file randomly reorders lines
**Describe the bug**
When saving the game project, the Game.agf file will sometimes change, apparently at random, by reordering some lines/tags. For example:

```
    <AudioClips>
        <!-- Bunch of nested tags -->
                  <AudioClip>
                    <SourceFileName xml:space="preserve">D:\Music\AGS\agsawards2015\x. Acceptance Speech No Space Quest.mp3</SourceFileName>
                    <ScriptName>aIntro</ScriptName>
```
Randomly changed to:

```
    <AudioClips>
        <!-- Bunch of nested tags -->
                  <AudioClip>
                    <ScriptName>aIntro</ScriptName>
                    <SourceFileName xml:space="preserve">D:\Music\AGS\agsawards2015\x. Acceptance Speech No Space Quest.mp3</SourceFileName>
```
(As you can see, the only change is that the `<SourceFileName />` and `<ScriptName />` tags change place.)

This adds junk to the version history and makes it harder to diff versions of the file in source control. 

**AGS Version**
Issue observed with 3.4.3

**To Reproduce**
1. Keep an AGS project in source control (git), making sure that Game.agf is treated as text
2. Save the game project and commit it.
3. Save the project again, check for changes.

**Expected behavior**
The Game.agf file should always be saved with tags in the same order, so that the file only changes when there is an actual change in content.

persn:
Fixes #1282 

Makes the Game.agf more reliable for version control users

### DIFF
--- a/Editor/AGS.Types/SerializeUtils.cs
+++ b/Editor/AGS.Types/SerializeUtils.cs
@@ -84,7 +84,7 @@ namespace AGS.Types
         /// <param name="writer"></param>
         public static void SerializePropertiesToXML(object obj, XmlTextWriter writer)
         {
-            IEnumerable<PropertyInfo> properties = obj.GetType().GetProperties().OrderBy(p => p.Name);
+            IEnumerable<PropertyInfo> properties = obj.GetType().GetProperties().OrderBy(p => p.Name, StringComparer.Ordinal);
 
             foreach (PropertyInfo prop in properties)
             {
--- a/Editor/AGS.Types/SerializeUtils.cs
+++ b/Editor/AGS.Types/SerializeUtils.cs
@@ -5,6 +5,7 @@ using System.Reflection;
 using System.Text;
 using System.Xml;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


morganwillcock:
Doesn't this need to be much more specific about what is doing the sorting to ensure that different locales (cultures) don't return the results in a different order?

Maybe explicitly specify to use `StringComparer.Ordinal` ?

persn:
Not sure, but I can just add it anyway because it doesn't cost us anything

ivan-mogilko:
May they actually return in different order? The property names are all strictly in latin chars.

ivan-mogilko:
So, this is okay to merge now?

persn:
I don't think there's an actual risk, but it's easy to stay safe. I think this is OK to merge unless anyone else thinks otherwise

Issue #1287, opened 2021-05-22
Alt+Tab on FullScreen Direct3D game hangs on 3.6.0.X


ericoporto:
**Describe the bug**
This bug has been reported on the forums as crashing here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=58976.msg636635407#msg636635407

I actually don't have a crash on my Windows 10 PC, instead when I alt+tab a fullscreen Direct3D game, it just hangs with the window staying in the background, but I can't alt+tab back to it and the only way to close it is by killing using the task manager

**AGS Version**
Built directly from the master branch https://github.com/adventuregamestudio/ags/tree/3d0353e36ef55b14fffe3f7515d74adf095415a1

**Game**
Here's an example game, but it happens with any on this engine build (3.6.0.4 release happens too) [game.zip](https://github.com/adventuregamestudio/ags/files/6525699/Release.zip)

**To Reproduce**
- Make sure the Direct3D is in use (check with Ctrl+V). If not, set using winsetup.
- make sure the game is fullscreen (Alt+Enter if needed)
- then, alt+tab 
- the game will hang in the background, you can't alt+tab back to it's window

**Expected behavior**
It's expected to alt+tab to any other window on the desktop and alt+tab back

**Desktop:**
 - OS: Windows
 - Version 10

ivan-mogilko:
I get a different error when running on Windows 7:

> Error: Unable to lock texture

Good thing that the problem is reproducible in some way.

ivan-mogilko:
Ah. AGS does not stop running the game and keep redrawing even when window has no focus. It could be that suspending the alt+tabbed game is not working in SDL version yet.

Maybe Direct3D does not like that in fullscreen mode. But I don't remember how it worked in previous versions when Multitasking mode was on. Might be worth comparing, maybe there was some condition that prevented renderer from working that is now not active.

ericoporto:
> AGS does not stop running the game and keep redrawing even when window has no focus

Uhm. AGS has `System.HasInputFocus` in the script API to be able to detect if the game is focused or not, does this means it should keep running just not drawing?

ivan-mogilko:
AGS has two modes: multitasking and non-multitasking, they are controlled by [SetMultitaskingMode](https://github.com/adventuregamestudio/ags-manual/wiki/Globalfunctions_General#setmultitaskingmode). In multitasking mode engine supposed to run game loops when alt+tabbed, otherwise it should suspend game execution.
[...]

PR #1288, opened 2021-05-22
fixes hang when alt+tab fullscreen direct3d game

ericoporto:
Probable fix of #1287 

My hypothesis: We are calling Reset on the direct3D device passing the d3dpp structure, and we are setting the Windowed property in it to false, but we are creating our window using `SDL_WINDOW_FULLSCREEN_DESKTOP`, which creates a "fake" fullscreen with a bordless window, so as far as Direct3D is aware, we have a Window, even in FullScreen.

https://docs.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-reset

### DIFF
--- a/Engine/platform/windows/gfx/ali3dd3d.cpp
+++ b/Engine/platform/windows/gfx/ali3dd3d.cpp
@@ -579,7 +579,7 @@ int D3DGraphicsDriver::_initDLLCallback(const DisplayMode &mode)
   // THIS MUST BE SWAPEFFECT_COPY FOR PlayVideo TO WORK
   d3dpp.SwapEffect = D3DSWAPEFFECT_COPY; //D3DSWAPEFFECT_DISCARD; 
   d3dpp.hDeviceWindow = hwnd;
-  d3dpp.Windowed = mode.Windowed;
+  d3dpp.Windowed = TRUE;
   d3dpp.EnableAutoDepthStencil = FALSE;
   d3dpp.Flags = D3DPRESENTFLAG_LOCKABLE_BACKBUFFER; // we need this flag to access the backbuffer with lockrect
   d3dpp.FullScreen_RefreshRateInHz = D3DPRESENT_RATE_DEFAULT;


ivan-mogilko:
> but we are creating our window using SDL_WINDOW_FULLSCREEN_DESKTOP, which creates a "fake" fullscreen with a bordless window

If this is what Direct3D renderer does, then it's a mistake, it should create proper fullscreen by default.

ivan-mogilko:
Well, I can merge this as a temporary fix, if you confirm that it fixes the problem.

ericoporto:
I think the bordless "fake" Fullscreen is better, the other one has the "crash changes desktop resolution" bug.

ericoporto:
It fixes in my computer but the person that reported didn't have a "hang" like I had, they reported a proper crash, which I could not reproduce. :(

ivan-mogilko:
> I think the bordless "fake" Fullscreen is better

Engine needs to support both in the end.

Regarding this fix, I am not certain if SDL window type really conflicts with Direct3D, because Direct3D supposed to handle modes itself.
Have you tried setting window flags to SDL_WINDOW_FULLSCREEN instead to see what happens?

ericoporto:
> Have you tried setting window flags to SDL_WINDOW_FULLSCREEN instead to see what happens?

I had not. But I just tried... And now I got the crash!

image

ivan-mogilko:
Last question though, did fullscreen mode work correctly in before 3.6.0 (in allegro versions)?

ericoporto:
> did fullscreen mode work correctly in before 3.6.0 (in allegro versions)?

Yes, it works with 3.5.0.X versions of the engine in my Windows 10 computer, I can alt+tab from a fullscreen window and back without problem.

ivan-mogilko:
Hm, after some thinking, I'd rather not apply such fix, but separate this into two tasks:
* implement proper switch between "exclusive fullscreen" and "desktop fullscreen" (full size borderless window) and support both modes;
* find out what is actually wrong with the exclusive fullscreen mode.

I'd begin with investigating how this flag is used in SDL2 source code, to make sure that it actually affects *something* (and what) in case when we do not create SDL_Renderer (and we only do that for the new software mode).
Besides sdl window, it may be also some mistake made in Direct3D class when porting to sdl2.

ericoporto:
https://github.com/libsdl-org/SDL/blob/157c3f8097403c08cc9747082de37276fb2ab28c/src/render/direct3d/SDL_render_d3d.c#L291-335

ivan-mogilko:
SDL_Renderer is only created for software mode, it is not called for other renderers (Direct3D/OpenGL).

ericoporto:
Yeah, I understand, I mean we can copy some of it's SDL Window handling.

ericoporto:
I am closing this since I understand a different approach is desired.

ericoporto:
I am leaving it reopened, but it can be closed if a different decision is made - software is hard. :/

ivan-mogilko:
I'll add it as a temporary fix.

PR #1289, opened 2021-05-22
3.5.1: Android: search for games using proper engine's algorithm

### LINKED ISSUE #379: Android Launcher: game search logic is incorrect
AGS launcher app for Android uses an incorrect method for detecting AGS game files: it looks for files named "ac2game.dat" which have Windows executable properties inside of them. (I think this was made for simplicity, at the times when proper method was not well known).

Correct way to find AGS game data files should be copied from the native engine; see [FindGameData](https://github.com/adventuregamestudio/ags/blob/master/Common/game/main_game_file.cpp#L134) for the reference.

ivan-mogilko:
Resolves #379.

Based on the work of @sonneveld (Nick Sonneveld) with minor modifications and fixes. Original commit: https://github.com/sonneveld/ags/commit/1b92d9d42d99a401fa02b6cbf0b4849633e581f9

The old method checked for Windows executable header and retrieved program version info to make sure the file contains Windows engine in certain version range. This made *some* sense many years ago when ports were mostly meant for running older games attached to Windows engine exe. But it no longer does.

The new method calls engine's own function FindGameData() - the one that engine uses itself when scanning directory for a game.

NOTE: there were other improvements done by sonneveld to game search in the android launcher, but I'd leave these for 3.6.0. This particular fix is most important. In 3.6.0 the new launcher works differently, and I am not yet sure if this fix could be applied same way.

NOTE2: In sonneveld's work the app was requesting Manifest.permission.WRITE_EXTERNAL_STORAGE, but I don't know if that is really necessary here, as we do not write anything when searching for games (do we?).


ericoporto:
> In 3.6.0 the new launcher works differently, and I am not yet sure if this fix could be applied same way. 

It would require some changes since class loading happens in SDL Activity, so it requires it to be created. Instead the old PE class could be renamed to a helper class and it could also load the engine and SDL native libraries so have that function available before the activity for usage by the launcher.

ivan-mogilko:
Looks like working, tested with game.ags from Compiled/Data (not attached to acwin.exe).

PR #1290, opened 2021-05-22
unified gfx support definitions to a single header 

ericoporto:
- unified OpenGL availability check to a single file.
- prefer AGS_HAS_DIRECT3D over AGS_PLATFORM_OS_WINDOWS for Direct3D checks

**problem:** There are currently 4 different files that need to be updated on the platform support for OpenGL, and also different macros for telling about OpenGL support and Direct3D support. This code change unifies these to a single file.

---

**QUESTIONS:** 
- I placed the new header in `Engine/gfx`, should this header be in `Common/gfx` instead? 
- Should I not have created a new HEADER and instead placed it on `Common/core/platform.h` ?
- We have MacOS OpenGL in some parts but not others, theoretically MacOS dropped OpenGL support in their newer versions, so I didn't include it in the new macro, but if it should have been, then I can add it back. I don't have a new MacOS available for me to test right now, on the VM the OpenGL renderer indeed fails but it may be something on it being a VM.


ivan-mogilko:
> Should I not have created a new HEADER and instead placed it on Common/core/platform.h

Yes, maybe just put them in platform.h for now? These are macros and not data declaration so nothing unnecessary is exposed.

ericoporto:
I will rework the PR so it's all in `Common/core/platform.h` and place the additional mentioned macros there.

ericoporto:
I adjusted the PR to instead move the macros to `Common/core/platform.h`, I also moved `AGS_OPENGL_ES2` and `AGS_SEARCH_FOR_GAME_ON_LAUNCH` macro definitions to there.

ivan-mogilko:
Regarding the MacOS, I don't know much abiut it myself but [this article sais](https://methi1999.github.io/2020/08/19/opengl.html) that it was deprecated in macOS 10.14 which is *less than 3 years old*. I don't know what are the traditions of system update among macos users, but I'd imagine there may be still people with that and older systems, esp. if they are installed on laptops?

I'd leave it as it is, because OpenGL loads dynamically and there's a fallback mechanism to software renderer anyway.

ericoporto:
Hey, I just found an error in the OGL driver, when the transparency shader fails to compile, there's no "backup plan", so it will write down the errors on the command line but not step back and try a different driver because the command that calls it actually expect errors in the shader to **throw** (it's in a try catch). So I can add back MacOS and works on my Mojave, but the newer versions I don't have one to test except in the VM, where some OpenGL works (just enough to not crash earlier) but the not enough for the shaders. I think I could add MacOS back in but I need to fix up the checks in the OGL driver so it steps back if the shaders fail.

ivan-mogilko:
Regarding the shaders, originally that was designed so that renderer keeps running even if shaders don't compile, only posting a warning. At least that worked with tint & light shaders on Android.
I have not yet investigated how it works with sonneveld's change in the sdl port.

ericoporto:
If the shaders fail to work the renderer won't work at all, I will open a ticket about it.

messengerbag:
> Regarding the MacOS, [this article sais](https://methi1999.github.io/2020/08/19/opengl.html) that it was deprecated in macOS 10.14 which is _less than 3 years old_. I don't know what are the traditions of system update among macos users, but I'd imagine there may be still people with that and older systems, esp. if they are installed on laptops?

Yes, in particular as that update also stops Wine from working, so many people with an interest in running Windows programs may hold off on it.

ericoporto:
@messengerbag on wine you should probably just use Directx since wine will translate it to OpenGL - on Linux it translates to vulkan I think.

messengerbag:
@ericoporto, I wasn't talking about running AGS through Wine, but commenting on whether renderers that only work up to MacOS 10.13 remain relevant. And given that 10.14 breaks a lot of software that many people rely on (such as Wine), so that there are reasons to remain on 10.13, I think the answer is yes.

Though at the same time, the fact that Wine is not an option in MacOS 10.14 and up means that having working Mac-native builds for those platforms becomes even more crucial.

PR #1291, opened 2021-05-23
CI fixes and updates

### LINKED ISSUE #1275: SDL2: particular OGG clip does not play
Following short OGG clip plays fine in AGS 3.5.1 but does not play in AGS 3.6.0: [dial1.zip](https://github.com/adventuregamestudio/ags/files/6468860/dial1.zip)

I did only brief debugging so far, and this leads down to the SDL_sound library's `Sound_Decode`, and then to `VORBIS_read`, which fails at some point.

This was tested with the AGS 3.6.0.3 built with our CI, so SDL_sound revision is fetched according to the CMake script.

Firstly may be worth trying this with the latest library revision, and if it's still not working then report to the library developers.

ericoporto:
**our Windows Dockerfile no longer builds**, since the [end of April](https://docs.microsoft.com/en-us/lifecycle/announcements/sha-1-signed-content-retired), [Visual Studio 2015 C++ Build Tools are no longer available through the web installer](https://docs.microsoft.com/en-us/answers/questions/387496/can39t-install-visual-cpp-tools-2015-on-windows-se.html), and downloading the new offline installer for it [from VS dowloads](https://my.visualstudio.com/Downloads?q=visual%20c++%20build%20tools&wt.mc_id=o~msft~vscom~older-downloads) requires password, so to avoid modifying the build too much, I am using [a docker image I pushed on dockerhub months ago that contains exactly it](https://hub.docker.com/repository/registry-1.docker.io/ericoporto/min-ags-dev-env). I [tried to pull from an alternate download on archive.org](https://gist.github.com/ericoporto/52290685e12b1c84f4b16c6c6055b73e) but it took just too much time to download it.

- this PR makes the Windows Dockerfile build again
- it removes Directx from the image and as dependency in the CMake scripts, since it's no longer required. The Directx installer we used is also no longer available on MS website.
- additionally, updates SDL_sound to latest commit available (all builds). (fixes #1275)
- Moves from Xiph libraries to GitHub Xiph mirror because of TLS incompatibilities with Debian Jessie's curl (see https://github.com/xiph/theora/issues/16).
- removes wrongly placed environment variables in our Editor build in the CI.
- fixes win bsdtar bug that kept failing when packaging the editor for release
---

please wait cirrus-ci build this and confirm this PR works.


ivan-mogilko:
This fails when building the Editor, from the logs it looks like `rc.exe` is missing (microsoft resource compiler)?

ericoporto:
Yeahp, I will look into fixing, it's probably related to have multiple VS installations, so I need to be explicit with the editor, I will look into it.

ericoporto:
@morganwillcock when you have time, can you take a look into this PR? I will left it as draft until then.

morganwillcock:
I think this is the only option left to use the existing CI configuration, so if it works it is fine with me. I think the only real change would be that previously the caching of the built container would only have been dependent on the Dockerfile contents, but now if we wanted to update something in your container the Dockerfile needs to also change in order to trigger the rebuild - so as long as the container version number changes it should be fine.

Worst case we switch to something else, I've started looking at build system changes so hopefully things get easier to build and not harder.

ericoporto:
Thanks Morgan! Yeah, in the meantime I am also looking to understand better how the Vs 2019 build tools work. @ivan-mogilko can you confirm the binaries generated in the CI are fine on Win7? It should be, but I don't have a Win7 machine anymore to test.

ericoporto:
Updated SDL_sound on the Linux Dockerfile too, had forgotten about it! Had to do a bit more there since the curl in it is too old.
Also bsdtar on Windows isn't reliable from piped input when it's a zip file (it's fine for tar.gz), so made it write to disk instead, probably will fix all those "missing `AGS.Control.dll`" bugs we keep getting on our CI. 
Updated the description on top to reflect that.

morganwillcock:
> Also bsdtar on Windows isn't reliable from piped input when it's a zip file (it's fine for tar.gz), so made it write to disk instead, probably will fix all those "missing AGS.Control.dll" bugs we keep getting on our CI.

It is likely a good idea to switch this although you should probably quote the paths just in-case the CI runner ever happens to use a temporary path with a space in it. e.g. `"%TEMP%\binaries.zip"` instead of `%TEMP%\binaries.zip`

ericoporto:
Ah, yes, I can quote. I ended up in a rabbit hole find out it's a limitation when a zip has multiple files, the tar.gz works because of it's two step approach.

ivan-mogilko:
> can you confirm the binaries generated in the CI (in this PR) are fine on Window 7

Yes, it works.

ericoporto:
if you want to, I have rebased these same commits in the branch here: https://github.com/ericoporto/ags/tree/ci_cmake_ags4_fixes

ivan-mogilko:
I'll just merge the master there, any changes specific to ags4 may be appended separately.

ericoporto:
Ah, cool, it should automatically merge then, here I had no conflicts!

PR #1292, opened 2021-05-23
3.5.1: reintroduced a suspend mode in WaitForNextFrame()

ivan-mogilko:
The problem is briefly mentioned here: https://github.com/adventuregamestudio/ags/issues/1287#issuecomment-846428137

Previously the engine was using allegro timers that were suspended by default when the player switched away from the game. That stopped game from updating and redrawing itself.

The commits f2ab428a596c67d950f2aed5fbd335f6ccc72e3f and 15489ff2c97762f7c40a8374d5b50241830bcd74 broke that behavior. Here we reintroduce it, now implementing the timer lock mechanic directly in the engine. This is a bit naive implementation but somewhat in line with the older code (e.g. see https://github.com/adventuregamestudio/ags/blob/release-3.5.0/Engine/ac/timer.cpp#L40).

### DIFF
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -101,6 +101,7 @@ RoomStruct thisroom;
 
 volatile int switching_away_from_game = 0;
 volatile bool switched_away = false;
+volatile int game_update_suspend = 0;
 volatile char want_exit = 0, abort_engine = 0;
 GameDataVersion loaded_game_file_version = kGameVersion_Undefined;
 int frames_per_second=40;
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
In regards to 3.6.0 (SDL2-based), unfortunately this fix is not enough.

In allegro version the "switch out" callback is run on a separate thread, and therefore can intercept a game update thread at the moment between two updates (frames).

In its current state in 3.6.0 all sdl events are processed on the same thread as game update, which means that if "switch out" event is handled the process will still be inside a game frame update and it will continue at least until the next frame. Some renderers may be fine with this, but Direct3D will be crashing in fullscreen. To fix this we need to have at least relevant sdl events processed on a separate thread. Perhaps, ideally the events are processed on main thread and the game update has its own. In any case a larger rewrite is necessary.

ivan-mogilko:
In regards to my above comment, I found it's actually possible to solve this in 3.6.0 with a minimal addition, which hopefully will work for now.

Issue #1227, opened 2021-02-28
Allow packaging arbitrary files


ericoporto:
My usecase is text files, like in-game books (`my-book.txt`). Maybe we could limit this if there's some dangerous use.

Basically, I would like to leverage AGS File.Open API but instead of reading from disk, I would like to read from non-exposed game resources (the own `.ags` game or maybe something like `.vox` packages) - I just don't would want people to **easily** read (`the-end.txt`) or tamper with my game, and currently the only way to do this is to write some script that generates AGS Script strings from text files in a script that is part of the game.

This would ease the process of revision and fixing of these text files that are part of the game for me.

ivan-mogilko:
This was even done for a test before, but user who requested did not respond and feature got dropped. I'm not even fully certain if there's no ticket for this already somewhere...

Absolutely trivial to do in the engine, all you have to do is to use common AssetManager::OpenAsset function. That will automatically search in every registered resource location. There are certain rules for where non-sound files are searched (voxes are skipped), mostly done for compatibility reasons, as AGS still does not have a fully proper "virtual filesystem", but these may be worked around if necessary.

There are two questions remaining:

1. How to address these files. E.g. there's currently $INSTALLDIR$ which means "data directory", and you are only allowed to perform read operations there. We may simply use this, then it will mean simply "search in game data". The tag name would seem a bit off then though. *Or* we may introduce new tag, like $RESOURCE$ for searching strictly in packages, except that would contradict virtual filesystem concept.
2. How to package these. Unfortunately game project still has inconsistent rules about how resources are included. So guess if we work with that we need to either a) have a  special named folder for these files, where game compiler would take them automatically (I used this method in the mentioned test), b) create a new node in project tree and basically make it work similar to audioclips.

ericoporto:
1. In my mind I had `$RESOURCE$` tag way. I know the engine magically picks up rooms if they are outside in some way but I don't know if this is related or not to things like `$INSTALLDIR$`. The need is read-only, yes!
2. A node in AGS project explorer like audio clips is the way I had on my mind. This also gives visual confirmation that things are going to get picked up.

My use case is really a bunch of in-game books and readable items that are easier to write and revise the text as a bunch of `.txt` files.

ivan-mogilko:
I think it's worth to clarify, there are two ways to approach all this. One is to treat these user added files as a new type of resource with its own rules (how to package, how to open, etc).
[...]

PR #1294, opened 2021-05-26
AGS 4: Support reading from game assets in script

ivan-mogilko:
For #1227 (does only engine part of the job).

This supports reading resource files ("assets") in script. Resources are retrieved using engine's own AssetManager, thus the script has access to everything that engine has. This method allows to open files only in read mode.

The opened file stream is modified to restrict reading and seeking to the range of offsets: this ensures that the script won't read past the end of resource even if it's retrieved from the larger file (package).

To specify game resource a new path token is introduced: "$DATA$".

Following script functions support this token:

**File.Open** - opens resource file for reading (if script asks to open for writing it will throw an error and return null pointer, as usual).
Example: `File* f = File.Open("$DATA$/myfile.dat");`

**DynamicSprite.CreateFromFile** - will try to create a sprite from the bitmap in game resources.
Example: `DynamicSprite* dspr = DynamicSprite.CreateFromFile("$DATA$/custom.bmp");`

**ListBox.FillDirList** - will fill list box with all found resource names using wildcard pattern.
Example: `MyListBox.FillDirList("$DATA$/room*.crm");` - this shows a list of all room files in the game.

### DIFF
--- a/Common/core/assetmanager.cpp
+++ b/Common/core/assetmanager.cpp
@@ -282,10 +282,11 @@ Stream *AssetManager::OpenAsset(const String &asset_name, const String &filter, 
     AssetLocation loc;
     if (GetAsset(asset_name, filter, false, &loc, open_mode, work_mode))
     {
-        Stream *s = File::OpenFile(loc.FileName, open_mode, work_mode);
+        Stream *s = work_mode == kFile_Read ?
+            File::OpenFile(loc.FileName, loc.Offset, loc.Offset + loc.Size) :
+            File::OpenFile(loc.FileName, open_mode, work_mode);
         if (s)
         {
-            s->Seek(loc.Offset, kSeekBegin);
             if (asset_size)
                 *asset_size = loc.Size;
         }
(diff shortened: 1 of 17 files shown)


ericoporto:
So, I tried the following, I saved a poem in a `poem.txt` file, renamed it to `poem.wav`, and imported with the Editor.

```
function hGlowingOrb_AnyClick()
{
  File* f = File.Open("$DATA$/au000001.wav",eFileRead);
  if(f == null) { Display("no file found"); return; }
  while(!f.EOF) { Display(f.ReadRawLineBack()); }
}
````

It works if I build the `.exe` and run it, but it doesn't work if I run in the debug mode in the Editor.

ivan-mogilko:
That was because when running in debug mode it does not pack sounds into the game but reads clips from audiocache, and audiocache is registered using special filter to maintain classic search rules... anyway, I made it ignore these filters when opening assets with $DATA$ token.

ericoporto:
Yeah! Now both works! And I could draw a bitmap I got from an "audio" file!

ivan-mogilko:
Well, after this there's still a task of adding custom files in the project, but meanwhile this workaround with substituting one of the existing resource types may also be used.

PR #1296, opened 2021-05-28
CI fixes and updates (ags4)

ericoporto:
Exactly the same commits in #1291 , to confirm it's working in the ags4 branch (need to check when CI builds). 

Not sure if the best is to merge through here or some other way, curiously I can't automatically merge when I target my own fork (!?), so I wasn't sure how to test in CirrusCi on my fork.


ericoporto:
ok, apparently I can't open two PRs with same commits since it breaks something on CirrusCi

PR #1302, opened 2021-06-01
AGS 3: Google test project for the Common library & Engine

### LINKED ISSUE #664: Remake engine tests as google test project (similar to compiler)
There's a bunch of semi-automatic tests written inside the engine, which are currently simply ran at startup if program is built with _DEBUG flag.

They should be moved out and remade as a google test project, similar to the one we have for script compiler. Then we will be able to configure them as a job for one of our CI tools.

These are mostly basic utility class and function tests, so should not be a big problem in extracting them out (they don't need actual game to run).

ivan-mogilko:
Resolves #664

This reimplements engine tests as google tests, similar to already existing compiler test project. Most of the existing tests are intended to test Common library classes (string, etc), for these I created a Common.Test project. EDIT: There's a separate solution for test projects now.

The projects builds a separate console application that runs tests.


ericoporto:
Conceptually, if the Engine builds acwin.exe but the test fails, do you wish the pipeline to fail or proceed?

Technically we can actually leave as is (adding Google test) and just store the binary of the test.exe and run in a different pipeline if it's desired that the everything else proceeds (like packaging the Editor).

ivan-mogilko:
I think the tests should run everything, is it possible to declared debug symbol for them always? If not, maybe I could hide these methods behind another macro.

ericoporto:
> I think the tests should run everything, 

The tests will run everything in the Debug build, and not everything in the Release build, but we build and run both.

> is it possible to declared debug symbol for them always?

Not easily, because we use the already built Common lib in both the Engine and the Test, so we would need to build Common twice for this.

> If not, maybe I could hide these methods behind another macro.

The string methods won't work hidden behind another macro since they are already built. Alternatively better to just not hide behind `AGS_PLATFORM_DEBUG` said methods. The alternative is to separate in a different solution, but I fear this will be too boring to maintain. Other idea is to not build the test in Release

image

ivan-mogilko:
I now think maybe I should not link whole Common, but only particular files, then I can hide these functions behind some TEST macro instead of DEBUG. These tests also do not test library in whole, but particular classes and functions.

ericoporto:
I don't understand, are those three inline methods used anywhere else but in these tests? Can't they just be in the test file itself? It's best to leave as either a test of common or engine, since this will make easier for additional tests one wants to write.

ivan-mogilko:
> I don't understand, are those three inline methods used anywhere else but in these tests? Can't they just be in the test file itself?

They are strictly for debugging and testing, but they access private String members.

ericoporto:
> I don't think it makes sense to not run these tests in release, as they are related to internal string work.

I looked how this accessibility was dealt in #1089 but in C# we had reflection :(

C++ has `friend` as the only apparent way other than inheritance to access the private methods information - e.g. making a TestableString from String with said additional methods.

I guess your way works... Don't know how to properly name the test code unit though. :/

ericoporto:
I feel really stupid, this fixes: https://github.com/ericoporto/ags/commit/51ab6f804fcf41b360664c1f1556378395005b43

Switch to a different macro and simply define there. I tried to define debug but it got overwritten by `core/platform.h`, this is why I changed the name.

ivan-mogilko:
I'm sorry, I rethought some things, and made a separate solution instead. This way you only need to fetch googletest module for this particular solution. The project itself is kept, but now does not link to the full common lib, but individual units that has to be tested.

Common/core/platform.h
```diff
@@ -102,6 +102,10 @@
     #define AGS_PLATFORM_DEBUG  (0)
 #endif
 
+#if defined(AGS_PLATFORM_DEBUG)
+    #define AGS_PLATFORM_TEST   (1)
+#endif
+
```
ericoporto, line 108: These three lines don't make much sense since AGS_PLATFORM_DEBUG is always defined. Perhaps you meant the code below?
```
#ifndef AGS_PLATFORM_TEST
#define AGS_PLATFORM_TEST (AGS_PLATFORM_DEBUG)
#endif
```

ericoporto (approved):

Common/core/platform.h
ivan-mogilko: No, I meant rather 

```
#if AGS_PLATFORM_DEBUG
#define AGS_PLATFORM_TEST (1)
#endif
```

ericoporto: Can the define be also in the vcxproj (in some of the million configs in it)? So it can be set in a release build of the tests.

ivan-mogilko:
Okay, I moved the last tests.

Common/core/platform.h
ivan-mogilko: I added this macro to both configs in the test projects, as intended.

ivan-mogilko:
@ericoporto do you know what is require to actually run these tests by CI? I could merge this now (unless there's some problem) and the deal with the cmake lists and everything else necessary later.

ericoporto:
Getting the submodule, building the source and running the test .exe file. Similar to the compiler test task: 
https://github.com/adventuregamestudio/ags/blob/4cfd049e6b45426f4c9a0db79f28632d5f57503d/.cirrus.yml#L298-L309

I assumed either I or Morgan would do this after you merged this PR.

PR #1303, opened 2021-06-02
Tools: trac - compiles game translation (TRS -> TRA)

### LINKED ISSUE #1299: Tool: translation compiler
As a step of decoupling game compilation procedure from the Editor, we need a stand-alone tool that is run from the command-line, parses a translation source (TRS) file and writes a compiled translation (TRA) file in binary format.
Should be written in the similar line with the existing tools, in C++ (#1262, #1264, #1269).

Name suggestion: `trac` (TRAnslation Compiler).

_NOTE:_ this task is exclusively in writing this tool, adjusting Editor is a separate task, so no need to concern yourself with that if you are doing this. In fact it's best to assume that Editor will not be present to ensure tool's work result has no reliance on it.

**Input:**
* a TRS file (AGS translation source);
* path to output file;

**Output**
* a TRA file (AGS compiled translation)

---

### Details

The algorithm is relatively simple, most of TRS is just pairs of text lines that form a key/value map, with a few exceptions (comments, options).
Current implementation of reading and writing TRS in the Editor may be found here:
https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Types/Translation.cs
Implementation of the TRA writer in the Editor may be found here: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Components/TranslationsComponent.cs#L92
User documentation of TRS format:
https://github.com/adventuregamestudio/ags-manual/wiki/Translations

The existing Tool code in the current master branch:
https://github.com/adventuregamestudio/ags/tree/master/Tools
MSVS solution for standalone tools:
https://github.com/adventuregamestudio/ags/blob/master/Solutions/Tools.sln

ivan-mogilko:
Resolves #1299

This tool compiles TRS (translation source file) into TRA (runtime translation file).

**trac** - stands for "TRAnslation Compiler".

*Input:* translation source (`*.trs`) and game identification parameters (name, uniqueid).
*Output:* runtime translation (`*.tra`).

Usage: `trac name.trs name.tra --gamename MyGame --uniqueid 12345`

---

There's a noteable issue that the TRA needs game title and uniqueid field to be set: engine uses these to match translation with the game. Normally editor would get these from the game data, but here I thought that it may be overkill to parse Game.agf for two values. Also, I thought hypothetically that we may add an option to force translation without game check.
These two arguments may be omited, in which case they will be saved as an empty string and 0 respectively.
**EDIT:** one alternative is to support reading these from TRS (there are already few extra options read from there), in which case TRS will be fully independent from the rest of the game data. But this won't work if it comes from older project of course - it will have to be updated in such case. Editor may be adjusted to write this information into TRS on generation.

Another thing, it may not be easy to compare output with the one produced by the editor. The main bulk of data is a dictionary of key/value string pairs and there's no rule to how these should be ordered. Editor currently saves them in the order they are present in .NET implementation of Dictionary class, which is an unordered map. Without using exactly same hashing etc algorithms I don't think it's possible to match the same order, and I don't believe it's essential to do so.

PS. Oh, and the output strings are encrypted by the way. Again, maybe we could support a switch that enables/disables this encryption for the easier debugging (but the engine will have to be aware of that too).

### DIFF
--- a/Common/game/tra_file.cpp
+++ b/Common/game/tra_file.cpp
@@ -12,14 +12,19 @@
 //
 //=============================================================================
 #include "game/tra_file.h"
+#include <string.h>
 #include "ac/wordsdictionary.h"
 #include "debug/out.h"
 #include "util/string_compat.h"
+#include "util/string_utils.h"
 
 namespace AGS
 {
 namespace Common
 {
+
+const char *TRASignature = "AGSTranslation";
+
 
 String GetTraFileErrorText(TraFileErrorType err)
 {
@@ -96,10 +101,9 @@ HTraFileError ReadTraBlock(Translation &tra, Stream *in, TraFileBlock block, sof
 HTraFileError ReadTraData(PfnReadTraBlock reader, Stream *in)
 {
     // Test the file signature
-    const char *tra_sig = "AGSTranslation";
     char sigbuf[16] = { 0 };
     in->Read(sigbuf, 15);
-    if (ags_stricmp(tra_sig, sigbuf) != 0)
+    if (ags_stricmp(TRASignature, sigbuf) != 0)
         return new TraFileError(kTraFileErr_SignatureFailed);
 
     while (!in->EOS())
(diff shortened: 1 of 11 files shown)


rofl0r:
> PS. Oh, and the output strings are encrypted by the way.

why?

ivan-mogilko:
> > PS. Oh, and the output strings are encrypted by the way.
> 
> why?

This is the original format. Probably to prevent players from reading game texts from the game files and getting hints and spoilers.

PR #1308, opened 2021-06-06
Winsetup icon to better supported functions

ericoporto:
- SetClassLongPtr is compatible with 32 and 64 bit
- LoadIcon has been superseded by LoadImage function


ivan-mogilko:
I think `(LONG)` cast there should also be replaced with `(LONG_PTR)`, according to SetClassLongPtr's prototype.

ericoporto:
Aha! True! Let me fix this.

PR #1309, opened 2021-06-08
fixes for agf2dlgasc.exe

ericoporto:
Hey, I tried to use `agf2dlgasc.exe`, but had an issue with `option-on` and `option-off` not being recognized. Additionally, I adjusted one statement that missed a C string conversion - judging from other calls in the same file.

Also I didn't fix here, but I noticed the option `DialogScriptSayFunction` from the `Game.agf` file is not considered - that one that allows using `Character.MySay()` instead of `Character.Say()`.

### DIFF
--- a/Tools/data/dialogscriptconv.cpp
+++ b/Tools/data/dialogscriptconv.cpp
@@ -389,7 +389,7 @@ String DialogScriptConverter::ProcessOptionOnOff(const String &line, const char 
         return "";
     }
 
-    const char *option_str = mr[3].str().c_str();
+    String option_str = mr[3].str().c_str();
     int option_num;
     if (StrUtil::StringToInt(option_str, option_num, 0) == StrUtil::kNoError)
     {
--- a/Tools/data/dialogscriptconv.cpp
+++ b/Tools/data/dialogscriptconv.cpp
@@ -343,7 +343,7 @@ String DialogScriptConverter::ProcessCmdGotoDialog(const String &line)
         }
     }
 
-    CompileError(String::FromFormat("Dialog not found: %s", dialog_name));
+    CompileError(String::FromFormat("Dialog not found: %s", dialog_name.GetCStr()));
     return "";
 }
 


ivan-mogilko:
Did this fix actually work? I'm asking because String has a constructor that accepts const char, and in theory StringToInt should implicitly convert const char* to String. So this situation looks strange to me.
(That said we may as well add a helper version later which accepts const char to avoid extra conversions)

> Also I didn't fix here, but I noticed the option `DialogScriptSayFunction` from the `Game.agf` file is not considered - that one that allows using `Character.MySay()` instead of `Character.Say()`.

These are suppose to be read from AGF, maybe there's some fault in the logic.
https://github.com/adventuregamestudio/ags/blob/master/Tools/data/agfreader.cpp#L269

ericoporto:
> Did this fix actually work? I'm asking because String has a constructor that accepts const char*, and in theory StringToInt should implicitly convert const char* to String

Yes, I get 0 without it in all of the option-on and option-off. Here's a [`Game.agf`](https://github.com/ericoporto/i_rented_a_boat/blob/main/I_rented_a_boat/Game.agf) if you want one to try it!

> These are supposed to be read from AGF, maybe there's some fault in the logic.

Yeah, I followed with the debugger, but I am afraid I don't understand how the xml reader should work. The problem is `ReadGameSettings()` is not working. You can try with above Game.agf too.

ericoporto:
Thanks for the fix in master! Now it correctly uses `MySay`.

ivan-mogilko:
Ah! This is stupid. It's not because of StringToInt.

This is because the line `mr[3].str().c_str()` creates a temporary string object that is deleted immediately after. And saving its result in "const char*" basically is useless.

ericoporto:
Ha! Good catch. I am fine closing this PR and waiting for your fix!

ivan-mogilko:
Well, it's a suitable fix.

ericoporto:
reworded the commit so it correctly reflects what is happening.

Issue #187, opened 2014-09-20
Compiler Inconsistency: Can compile structs with function in global scripts but not in room scripts


edmundito:
It turns out that in a global script/script module source you can declare a struct with methods in them such as:

``` C++
// Module.asc
struct MyStruct {
  import function MyFunction();
};

function MyStruct::MyFunction() {
}
```

Trying to do the same thing in a room script will cause a script error because the compiler cannot properly find MyFunction:

> Variable 'MyStruct::MyFunction' is already imported

ivan-mogilko:
Confirmed to still present in 3.5.0.16.

ivan-mogilko:
Should have looked into this earlier. The reason is not the compiler itself, but that room scripts are compiled with an additional option:

https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Native/ScriptCompiler.cpp#L76

ivan-mogilko:
I sort of *think* this might have been to prevent occasional mistakes in the times when there were no script modules but only a global script and global header. Then this could have been a solution for users mistakenly placing imported function implementation in the room script. But that's just a guess.

I think we may remove this, because it makes no sense now.

ivan-mogilko:
Fixed by e08de235c7154a82c681709aa69208a3a52972b8

fernewelten:
As far as I understand the issue with "overriding imports":

Let's do variables first and then functions.

In general, you have a module body `B.asc` and the corresponding module header `B.ash`: 
* When the body of the module is compiled, then the header file and the body file are compiled together.
* When all the other entities are compiled, then only the header file is included in front of the entities.

In the former case, the header file will have lots of `import` variables, but the engine should not import the variables really: To the contrary, the body file will provide local definitions for them, and the engine should work with these local versions. 

However, those variables need to be EXported in this case so that they can be available to other modules. So whenever a body file provides local definitions for a variable that has been `import`ed, then this local definition will “override” the `import` statement of the header file and convert it into an `export` instead.

Now, rooms are a special case. The AGS tries to economize on memory usage, and so it compiles each room into a separate “room program”. Whenever the player character changes rooms at runtime, the program for the old room is swapped out and the program for the new room is swapped in. This means that all the variables in room files are __not__ available to __other__ room files.

The room programs are assembled by joining (amongst others): 
* `globalscript.ash` (the header)
*  `globalscript.asc` (the body)
* and the respective room script.

[...]

PR #1310, opened 2021-06-09
adds a command line front end to the script compiler 

ericoporto:
Adding a command line front end for the script compiler. Named `agscc`, currently this is the output of `agscc --help`:

```
agscc v0.1.0 - A command line compiler for preprocessed ags scripts
Copyright (c) 2021 AGS Team and contributors

Usage: agscc <input.asc> <output.o> [OPTION...]
-H, --Headers HEADER  Header Files in order
-r, --roomscript      Use if is a room script
-h, --help            Print usage
```

If I modify the code I will update usage above. Only for Windows (for now), builds using a Solution and vcxproj, they are placed on the `Solutions` directory.

There are probably many errors, and I didn't write with any expectation of merging, I think at minimum having something to play can help writing an issue for what a script compiler should have - and also how to organize the files. :)

I was looking the original compiler files and could not figure how to get the editor from the following line:

https://github.com/adventuregamestudio/ags/blob/daf70322bf8677534d31570b7253a94ea16d5758/Editor/AGS.Native/ScriptCompiler.cpp#L71

(for testing I used a hacky tool I made to patch crm room files with the script obj output)

---

List of flags in the compiler today (here because they need to be in the interface)

https://github.com/adventuregamestudio/ags/blob/daf70322bf8677534d31570b7253a94ea16d5758/Common/script/cc_options.h#L22-L29

@rofl0r gave following suggestions for the name of the flags which we could use: 
```
 exportall (1/0) - export all functions automatically
 showwarnings (1/0) - printf warnings to console
 linenumbers (1/0) - include line numbers in compiled code
 autoimport (1/0) - when creating instance, export funcs to other scripts
 debugrun (1/0) - write instructions as they are processed to log file
 noimportoverride (1/0) - do not allow an import to be re-declared
 lefttoright (1/0) - left-to-right operator precedance
 oldstrings (1/0) - allow old-style strings
```

Note though that `noimportoverride` doesn't exists anymore in the editor so we should remove it.

Funnily in general settings not all compiler options are in the compiler area: https://adventuregamestudio.github.io/ags-manual/Settingupthegame.html#general-settings

#### adding preprocessor

There are certain conveniences gained by adding preprocessor support, we may need that here too. On the interface this means setting preprocessor macros. Here are two options on this from @rofl0r's list:

```
-D macro[=val] : define preprocessor macro
-E : run preprocessor only (create file with .i extension)
```

Though note that in this case a preprocessor has to be written.

---

I also have been thinking about passing individual headers through command line and am not sure this is the best way to go, I imagine any medium game would have like 60 different parameters for headers alone.
<details>

```
private void DefineMacrosAccordingToGameSettings(IPreprocessor preprocessor)
		{
			preprocessor.DefineMacro("AGS_NEW_STRINGS", "1");
			preprocessor.DefineMacro("AGS_SUPPORTS_IFVER", "1");
			if (_game.Settings.DebugMode)
			{
				preprocessor.DefineMacro("DEBUG", "1");
			}
			if (_game.Settings.EnforceObjectBasedScript)
			{
				preprocessor.DefineMacro("STRICT", "1");
			}
			if (_game.Settings.LeftToRightPrecedence)
			{
				preprocessor.DefineMacro("LRPRECEDENCE", "1");
			}
			if (_game.Settings.EnforceNewStrings)
			{
				preprocessor.DefineMacro("STRICT_STRINGS", "1");
			}
            if (_game.Settings.EnforceNewAudio)
            {
                preprocessor.DefineMacro("STRICT_AUDIO", "1");
            }
            if (!_game.Settings.UseOldCustomDialogOptionsAPI)
            {
                preprocessor.DefineMacro("NEW_DIALOGOPTS_API", "1");
            }
            // Define Script API level macros
            foreach (ScriptAPIVersion v in Enum.GetValues(typeof(ScriptAPIVersion)))
            {
                if (v == ScriptAPIVersion.Highest)
                    continue; // skip Highest constant
                if (v > _game.Settings.ScriptAPIVersionReal)
                    continue;
                preprocessor.DefineMacro(_scriptAPIVersionMacros[v], "1");
            }
            foreach (ScriptAPIVersion v in Enum.GetValues(typeof(ScriptAPIVersion)))
            {
                if (v == ScriptAPIVersion.Highest)
                    continue; // skip Highest constant
                if (v < _game.Settings.ScriptCompatLevelReal)
                    continue;
                preprocessor.DefineMacro(_scriptCompatLevelMacros[v], "1");
            }
        }
```

</details>

### DIFF
--- a/Compiler/CMakeLists.txt
+++ b/Compiler/CMakeLists.txt
@@ -42,6 +42,16 @@ source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} PREFIX "Source Files" FILES ${COMM
 
 add_library(AGS::Compiler ALIAS compiler)
 
+add_executable(agscc main.cpp)
+set_target_properties(agscc PROPERTIES
+        CXX_STANDARD 11
+        CXX_EXTENSIONS NO
+        C_STANDARD 11
+        C_EXTENSIONS NO
+        )
+
+target_link_libraries(agscc PUBLIC AGS::Compiler)
+
 if(AGS_TESTS)
     add_executable(
             compiler_test
--- a/Compiler/main.cpp
+++ b/Compiler/main.cpp
@@ -1,0 +1,151 @@
+#include "script/cs_compiler.h"
+#include "script/cc_options.h"
+#include "script/cc_error.h"
+#include "util/filestream.h"
+#include "util/file.h"
+#include "util/textstreamreader.h"
+#include "util/string_compat.h"
+
+using namespace AGS::Common;
+
+/* needs this or linker fails */
+extern int currentline; // in script/script_common
+
+std::pair<String, String> cc_error_at_line(const char* error_msg)
+{
+    return std::make_pair(String::FromFormat("Error (line %d): %s", currentline, error_msg), String());
+}
+
+String cc_error_without_line(const char* error_msg)
+{
+    return String::FromFormat("Error (line unknown): %s", error_msg);
+}
+
+
+const char *HELP_STRING = "Usage: agscc <input.asc> <output.o> [OPTION...] \n"
+"-H, --Headers HEADER  Header Files in order\n"
+"-r, --roomscript      Use if is a room script\n"
+"-h, --help            Print usage\n";
+
+int main(int argc, char* argv[])
+{
+    bool isRoomScript = false;
+    int h_i_start = 0;
+    int h_i_end = 0;
+    printf("agscc v0.1.0 - A command line compiler for preprocessed ags scripts\n"\
+        "Copyright (c) 2021 AGS Team and contributors\n");
+
+    for (int i = 1; i < argc; ++i)
+    {
+        const char *arg = argv[i];
+        if (ags_stricmp(arg, "--help") == 0 || ags_stricmp(arg, "/?") == 0 || ags_stricmp(arg, "-?") == 0)
+        {
+            printf("\n%s\n", HELP_STRING);
+            return 0; // display help and bail out
+        }
+        if (ags_stricmp(arg, "--roomscript") == 0 || ags_stricmp(arg, "-r") == 0)
+        {
+            isRoomScript = true;
+        }
+
+        if (ags_stricmp(arg, "--Headers") == 0 || ags_stricmp(arg, "-H") == 0)
+        {
+            h_i_start = i;
+        }
+        else if(h_i_start != 0)
+        {
+            h_i_end = i;
+        }
+    }
+    if (argc < 3)
+    {
+        printf("Error: not enough arguments\n");
+        printf("%s\n", HELP_STRING);
+        return -1;
+    }
+
+    const char *src = argv[1];
+    const char *dst = argv[2];
+    printf("Input script file: %s\n", src);
+    printf("Output script obj: %s\n", dst);
+
+    std::vector<String> headers;
+    if (h_i_start != 0 && h_i_end > h_i_start)
+    {
+        for(int i=h_i_start+1; i<=h_i_end; i++) {
+            headers.emplace_back(argv[i]);
+        }
+    }
+
+    //-----------------------------------------------------------------------//
+    // Configure compiler
+    //-----------------------------------------------------------------------//
+    ccSetSoftwareVersion("version"); // need to set the version here somehow
+
+    ccSetOption(SCOPT_EXPORTALL, 1);
+    ccSetOption(SCOPT_LINENUMBERS, 1);
+    // Don't allow them to override imports in the room script
+    ccSetOption(SCOPT_NOIMPORTOVERRIDE, isRoomScript);
+
+    ccSetOption(SCOPT_LEFTTORIGHT, 1 /*LeftToRightPrecedence*/);
+    ccSetOption(SCOPT_OLDSTRINGS, 0 /*EnforceNewStrings*/);
+
+    ccRemoveDefaultHeaders();
+
+    //-----------------------------------------------------------------------//
+    // Read input files
+    //-----------------------------------------------------------------------//
+    std::vector<String> heads;
+    for(const auto& header: headers)
+    {
+        Stream *in = File::OpenFileRead(header);
+        if (!in)
+        {
+            printf("Error: failed to open header for reading: %s\n", header.GetCStr());
+            return -1;
+        }
+        TextStreamReader sr(in);
+        heads.push_back(sr.ReadAll());
+        in->Close();
+
+        ccAddDefaultHeader((char*) heads.back().GetCStr(), (char*) header.GetCStr());
+    }
+
+    String script_input;
+    if(src)
+    {
+        Stream *in = File::OpenFileRead(src);
+        if (!in)
+        {
+            printf("Error: failed to open script for reading: %s\n", src);
+            return -1;
+        }
+        TextStreamReader sr(in);
+        script_input = sr.ReadAll();
+        in->Close();
+    }
+
+    //-----------------------------------------------------------------------//
+    // Compiler script
+    //-----------------------------------------------------------------------//
+    ccScript* script = ccCompileText(script_input.GetCStr(), src);
+    if ((script == nullptr) || (ccError != 0))
+    {
+        printf("Error: compile failed at %s, line %d : %s\n", ccCurScriptName, ccErrorLine, ccErrorString.GetCStr());
+        return -1;
+    }
+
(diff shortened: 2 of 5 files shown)


ivan-mogilko:
It would be really good to clarify in the PR description that this is a standalone launcher for existing compiler, because right now there's no explanation on what it is and why it's needed for anyone who is not familiar with the current development context.

Regarding the options, tbh I was thinking to propose "-I" as in "include", as in theory it does not have to be a "header" file, but you may glue literally any file. But that's a matter of discussion.

More importantly, the "is room" option is not needed, and I removed the SCOPT_NOIMPORTOVERRIDE from the Editor some time ago, because it was restricting room scripts from implementing imports: #187

ericoporto:
> In general it's good to first make a list of requirements for a tool, about its input data, settings etc, and its output, so that everyone knew what is needed and what to expect.

I totally agree, but I wanted to have this as a small sketch to help figuring these requirements. Also this was essentially free code since I had written it a long time ago - just with slightly different organization. It's alright throwing it away 😃 

About the preprocessor, I think it could go in the same tool, probably with an option to run just the preprocessor (so it can be tested), it has to match the exact behavior of the one we have in the editor because we want our code completion to match the results. Anyway, but I hadn't written a preprocessor, so just `cl.exe /D my.asc` for now 😬 

I will set the PR as draft to better reflect. Still, having a sketch that builds and runs may help figuring other tools. :)

rofl0r:
uhm, could we keep this compatible with my [https://github.com/rofl0r/ascc](ascc) please ? or even better, use my already working compiler frontend. i carefully modeled the command line options to be familiar with existing C/C++ compiler users.

```
usage: ./ascc [OPTIONS] FILE.asc
compiles FILE.asc to FILE.o
OPTIONS:
-A API version : specify API version. this affects which functions
   and data items are made available by global header agsdefns.sh.
   supported values are:
   350 341 340 335 334 330 321 261
   default: 350
-a ABI version: same as above, but restricts the emitted bytecode
-f option=value: set a compiler option. recognized options:
   exportall (1/0) - export all functions automatically
   showwarnings (1/0) - printf warnings to console
   linenumbers (1/0) - include line numbers in compiled code
   autoimport (1/0) - when creating instance, export funcs to other scripts
   debugrun (1/0) - write instructions as they are processed to log file
   noimportoverride (1/0) - do not allow an import to be re-declared
   lefttoright (1/0) - left-to-right operator precedance
   oldstrings (1/0) - allow old-style strings
   checkbounds (1/0) - emit array boundary check instructions. makes code slower but could be helpful for debugging.
   explicitret (1/0) - emit explicit return 0 value in non-void functions closing without return
   oldnegate (1/0) - emit old inefficient code loading a negative literal
   pointerhack (1/0) - enable hack that allows pointers/strings and ints to be used interchangeably. required for emulation of pointer-to-pointer until it is properly implemented for some low-level code. not recommended for use in AGS.
   free-standing (---) - do not include systemheaders
-i systemheaderdir : provide path to system headers
   this is the path containing implicitly included headers (atm only agsdefns.sh).
-H 1.ash[:2.ash...] : colon-separated list of headers to include.
   order matters. AGS editor automatically includes all headers
   it finds in the game dir, but here you need to specify them.
-o filename : explicitily specify output file name
-P preprocessor : use `preprocessor` on files before running the built-in
   one on them. example: -P "cpp -P". note that you if you use a
   standard C preprocessor, linenumber directives need to be supressed
   as in the example cpp command.
-D macro[=val] : define preprocessor macro
-E : run preprocessor only (create file with .i extension)
-W : warn about non-fatal errors parsing system headers
-g : turn on debug info
-S : write textual assembly instead of binary (requires agsdisas)
```

ericoporto:
Can you change it's license?

rofl0r:
ascc uses the AGS license already.

ericoporto:
Thanks! Can we just pick code from it directly? Specially for the command line parsing, since I think you have solved a lot of things with it already.

ivan-mogilko:
I'm not against compatibility with that program, although I was confused by the API level handling and wanted to take time and investigate how it's done and why.

Frankly speaking I doubt about using the main.cpp code directly, because it looks like written in a C style and ours is a C++ project (well, to be fair, old compiler was written much like that), and there may already be non-portable things like `getopt` which have to be either copied from posix libraries or rewritten.

The presence of a new @fernewelten 's is another factor, in time we probably will have to make sure that both compilers support same options...

---

I think this should be done in a more organized way, for example:

1. If there are improvements or useful additions to the old compiler, like writing more efficient byte code, these could be merged separately (as they don't depend on how compiler is run). At least have them documented first as a list with references to commits or something.

2. If I understand correctly, the preprocessor is actually an original AGS code in C++ which was later deleted from repository? It's been too long and I was not sure if one existed, but it may (should) be restored of course. There are few things specific to AGS preprocessing, like linenumber reset, and also `#ifver` / `#ifnver` which may be met in some module code (this was an old way to test API).

3. In terms of a compiler launcher, yes, we need to have an elaboration on the API and headers rule, and basically gather full picture of how it's suppose to work with these options, for the reference. I.e. I already see there are modes that run only preprocessor and output files with different extensions.

ericoporto:
> If I understand correctly, the preprocessor is actually an original AGS code in C++ which was later deleted from repository?

I am not sure, we use a C# preprocessor that runs in the editor, but if you look into the current `Compiler` directory you will see some macro related code, but the files were placed there in 2012, I don't know if the preprocessor was somewhere else before.

ivan-mogilko:
I found, this big "cleanup" commit removed it: 507def9d1914b5ec958388bab3d639cdb7624341
So the old preprocessor should exist before that commit.

ivan-mogilko:
I tried testing this with a trivial script with just 1 function inside, but compiler fails with "undefined token `function`" error.
I may only assume something is missing, either some options have to be added, or preprocessor was supposed to add some special keywords to the script, or else.
Comments are failing too, maybe they are dealt with preprocessor in AGS.

ivan-mogilko:
Oh, just for the reference, I did a small test with this, and unfortunately it's too inconvenient to use without a preprocessor, because comments are not parsed correctly or not at all by the compiler itself, and so it's impossible to add default api header (and you cannot have comments in your code).

(Also you cannot have `function` in the code, because it's defined as `#define function int`, and define does not work without preprocessor either)

I had to declare the functions inside the script to complete a successful test:
<pre>
int a;
import void Display(const string message, ...);

int on_key_press(int k)
{
    Display("works");
}
</pre>

rofl0r:
> Just for the reference, I did a small test with this, and unfortunately it's too inconvenient to use without a preprocessor

ftr, in my agscc project i've ripped out the last c++ preprocessor version from ags git history and embedded it into the project.

ericoporto:
@ivan-mogilko for clarity, HOW did you do a test?

About the preprocessor you can use Cl.exe until one is written.

Also, any ideas on passing the headers? I think the order of them don't matter, but they still need all to be passed and once the project starts to get big the list gets big... It becomes not very practical to pass one by one on the command line...

ivan-mogilko:
> @ivan-mogilko for clarity, HOW did you do a test?

Simply, something like `agscc.exe path-to-game-dir/room1.asc path-to-game-dir/room1.o`

> Also, any ideas on passing the headers? I think the order of them don't matter, but they still need all to be passed and once the project starts to get big the list gets big... It becomes not very practical to pass one by one on the command line...

Errmmmmm.... the order *does* matter of course, just like in C/C++, because the headers may use declaration from the previous headers too.

Maybe we could accept a list of these headers in a text file? Passed with an alternate command parameter (I don't know, to make something up `-H` passes headers and `-h` passes list file).

ericoporto:
if the order matters, how did you just link Display at run time from the middle of the code?

ivan-mogilko:
> if the order matters, how did you just link Display at run time from the middle of the code?

Sorry, I don't understand the question...

The order is relevant, for example script header 1 may define a custom type (enum, etc), and script header 2 may declare a function returning that type.

ericoporto:
Ah, right, now it makes sense, thanks!

ivan-mogilko:
Well, is not that what (system) scripts are for?

Tbh at the moment I'm concerned mostly with having all the basics first, and later we may figure something for extra convenience.
E.g. there could be a list of modules that compiler's frontend could parse and basically figure out which headers to include, but that's only a random idea.

ericoporto:
Agreed, let's get this shipped and going with all tools!

Issue #1312, opened 2021-06-11
ags4 new compiler string on struct error


ericoporto:
**Describe the bug**
I can't store a string in a non-managed struct when using the new compiler, it gives me no errors, but once the game loads I get the following error:

Error: internal error: MEMWRITEPTR argument is not dynamic object

**AGS Version**
AGS 4, [98b56ebb7ae722a6d387aa88c509a281382994f1](https://github.com/adventuregamestudio/ags/tree/98b56ebb7ae722a6d387aa88c509a281382994f1) | Cirrus artifact: [task/5983374398455808](https://cirrus-ci.com/task/5983374398455808)

**Game**
[compilerbugags4.zip](https://github.com/adventuregamestudio/ags/files/6635244/compilerbugags4.zip)

**To Reproduce**
1. Select the new script compiler
2. Put the following code in a script module:
```AGS Script
struct Stru {
  String Txt;
};

void game_start()
{
  Stru a;
  a.Txt = "Cause bug!";
}
```
4. Build, run and see error

**Expected behavior**
It should not cause error.

**Screenshots**
![image](https://user-images.githubusercontent.com/2244442/121618546-556db100-ca3d-11eb-8936-c8efe68593cc.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 10

ivan-mogilko:
@ericoporto not related to the bug, but will it be easy to add new compiler tests to CI too? (#1217)

ericoporto:
Ah, it is easy, yes! I didn't saw that issue. I will add.

fernewelten:
> Error: internal error: MEMWRITEPTR argument is not dynamic object

I'm having a look at it ...

ericoporto:
The PR fixed for me, but as requested I opened a PR (#1314) to add tests for the compiler and found a failing test.

fernewelten:
> found a failing test.

Addressed in #1315.

ivan-mogilko:
This ticket is resolved by #1313.

PR #1313, opened 2021-06-11
New Compiler: Fix bug when assigning a string literal to a String

fernewelten:
This fix addresses #1312.

When a `string` literal is assigned to a `String`, a `string`-to-`String` conversion should be emitted before the assignment is emitted. Wasn't when compiling the following AGS code:
```
struct Stru {
  String Txt;
};

void game_start()
{
  Stru a;
  a.Txt = "Cause bug!";
}
```

Root cause: Parameters for `ConvertAXStringToStringObject()` were named differently in `parser.h` and `parser.cpp`; because of this confusion, this function was called incorrectly in one place.

This PR fixes the bug and adds a googletest.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -4189,7 +4189,7 @@ AGS::ErrorType AGS::Parser::AccessData_AssignTo(ScopeType sct, Vartype vartype, 
         return kERR_None;
     }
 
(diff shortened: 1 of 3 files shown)


ericoporto:
( Note on failures on the CI today, "Cloud Build experiencing high CPU quota usage", I believe this may fail to start some containers. Unfortunately I can't find any more info and the status page for Google Cloud doesn't have the an entry for this in their graphs: https://status.cloud.google.com/ )

PR #1314, opened 2021-06-12
CI: add compiler2 tests in ags4 branch

### LINKED ISSUE #1217: AGS4 CI: run new compiler tests
There's already "compiler_tests" job running, but it runs old compiler tests, and we have new ones in the ags4 branch.

Need to add new compiler tests, I guess preferably as a separate job to make it easier to distinguish errors.

ericoporto:
fix #1217 

- builds and tests new compiler in CI in ags4 branch
- fix small typo in the test vcxproj for the new compiler

**NOTE: there's a test current failing in `cc_parser_test_1.cpp`, under `TEST_F(Compile1, FloatInt2)`**

I could allow failure tolerance if desired (https://cirrus-ci.org/guide/writing-tasks/#failure-toleration) but it would still break flow on the specific task.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -315,11 +315,16 @@ ags_windows_tests_task:
   env:
     BUILD_CONFIG: Release
   get_submodule_script: git submodule update --init Common/libsrc/googletest
-  build_compiler_test_runner_script: >
+  build_legacy_compiler_test_runner_script: >
     "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
     cd Solutions &&
     msbuild Compiler.Lib.sln /p:PlatformToolset=v140 /p:Configuration=%BUILD_CONFIG% /p:Platform=Win32 /maxcpucount /nologo
-  run_compiler_tests_script: Solutions\.test\Release\Compiler.Lib.Test.exe
+  run_legacy_compiler_tests_script: Solutions\.test\Release\Compiler.Lib.Test.exe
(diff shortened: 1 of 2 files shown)


fernewelten:
> there's a test current failing in `cc_parser_test_1.cpp`, under `TEST_F(Compile1, FloatInt2)`

Well, it shouldn't. This test tests whether `int Var = !2.0;`makes the compiler balk, and the compiler __should__ balk. I ran the whole test suite before sending of the PR, including this test, and there were no errors in my environment.. 

So it seems that the CI test has uncovered some inconsistency. What's off?

ericoporto:
> 1 Byte /Zp1". Once I changed them to have the same alignment

Oh, I think I remember this error at some point in the past of zp alignment, we had something like this at some point with native dll build I think. I notice you build locally using debug builds but we build in the CI the release builds - since it's what ships in the Editor, not sure if this has some influence.

ericoporto:
Found the history here: https://github.com/adventuregamestudio/ags/pull/1141

Trying to decipher what this all means.

ericoporto:
I reproduced in my machine, but not sure what it is yet...

image

fernewelten:
So here we go. I switched to the Release build, and after slightly fixing `Compiler2.Lib.Test.vcxproj", I successfully recompiled the project. Then I ran the test suite. 

Voilà! __Now__ I can replicate that all the tests succeed except for `FloatInt2`.

Now to find out what's wrong. ...

ericoporto:
Yey! Awesome!!! I was here looking at the code without understanding a thing! Thanks!

Issue #738, opened 2019-04-29
Continue compiling scripts if there's error in only 1 one


ivan-mogilko:
Consider you have a lot of script modules and rooms, and changed something in upper script which would cause many other scripts to break. You would like to see all of the new errors, but instead compiler will stop on first one.

I do not know how feasible is to make ags compiler to continue compiling same script after 1 error, but that must be possible to continue with other scripts at least, because compilation of a script body should not depend on compilation of other script bodies.

tzachshabtay:
You can also maybe compile scripts in parallel (when there's no dependency) to speed up the process.

fernewelten:
We can and it would be well worth working on because it would at least halve the compiling time. But then we must  be careful about shared memory.

Currently, the compiler isn't an object at all, it's a bunch of functions. These functions work with a global symbol table, declared `Symboltable sym;`, and the code is shock full of expressions such as `sym.get_type(baz)` etc. If two threads work with the same symbol table at the same time, especially when compiling different sources, they will definitely clash.

fernewelten:
>but that must be possible to continue with other scripts at least, because compilation of a script body should not depend on compilation of other script bodies.

For each source to be compiled, the Editor calls the  native method `CompileScript()` which makes the compiler run. Whenever a compiler run fails, it throws an exception. The loop over the sources is wrapped in one giant `try`, this is why the first failed compiler run will abort *all* compilations. 

See code excerpt below, in source `AGSEditor.cs:930`. This might be the point of the code that needs to be changed. 
```
try
{
	[…]
	foreach (Script script in GetInternalScriptModules())
	{
		CompileScript(script, headers, errors, false);
		[…]
	}

	foreach (ScriptAndHeader scripts in _game.RootScriptFolder.AllItemsFlat)
	{
		[…]
		CompileScript(scripts.Script, headers, errors, false);
		[…]		
	}

	CompileScript(dialogScripts, headers, errors, false);
	[…]
}
catch (CompileMessage ex)
{ […] }
```

fernewelten:
For reference, the new compiler works on its own copy of the symbol table, so running several compilers in parallel would work in this respect. 

[...]

PR #1319, opened 2021-06-16
New compiler: Receive and show compiler warnings in Editor

fernewelten:
Please test before integrating
I tried hard to keep all the other behaviour as before, but you never know.
-----
The new compiler has generated warnings under certain circumstances that were invisible because they didn't reach the Editor. Typical example: A header file has:
```
import function Foo(float f = 0);
```
The old compiler treats '0' as a special case and lets this through. The new compiler does likewise but warns that this default value is an integer and so putting `0.0` would be preferable.

This PR makes those compiler warnings show up in the Editor. Warnings don't terminate the compilation run (but errors do, of course.)
As a side effect, this PR improves compilation times a bit.

-----
Specific changes:
New compiler:
- `ccCompileText2()`, add parameter for messages
- Move message handler into dedicated `cc_message_handler.cpp/.h`
- cs_compiler.h, change "THIS IS PROPRIETARY CODE" comment to Artistic License verbiage.

Editor / AGS.Native:
- `CompileScript()` , add parameter for messages; don't throw on compiler error
- `CompileScript()`, call new compiler in a way that gets warnings as well as error (old compiler still works as before)

Editor / AGS.Editor:
- Don't throw on compiler error.
- Aggregate messages in `CompileScripts()` instead of `CompileMessages()`
- RoomsComponent.cs, adapt to new behaviour of `CompileMessages()`

### DIFF
--- a/Compiler/script2/cs_compiler.cpp
+++ b/Compiler/script2/cs_compiler.cpp
@@ -14,17 +14,12 @@
 
 // A  wrapper around cc_compile(), in order to squeeze the C++ style parameters 
 // through the limited means of Managed C++ (CLR) into the C# Editor.
-ccScript *ccCompileText2(char const *script, char const *scriptName, long const options)
+ccScript *ccCompileText2(std::string const &script, std::string const &scriptName, long const options, MessageHandler &mh)
 {
-    // All warnings and the error (if present) end up here.
-    // TODO: This is what will need to be sqeezed through the interface
-    // into the editor so that the editor can display the warnings, too
-    MessageHandler mh;
-
     ccCompiledScript *compiled_script =
         new ccCompiledScript(FlagIsSet(options, SCOPT_LINENUMBERS));
 
-    compiled_script->StartNewSection(scriptName ? scriptName : "Unnamed script");
+    compiled_script->StartNewSection(scriptName.empty() ? scriptName : "Unnamed script");
     int const error_code = cc_compile(script, options, *compiled_script, mh);
     if (error_code < 0)
     {
--- a/Compiler/script2/cs_compiler.h
+++ b/Compiler/script2/cs_compiler.h
@@ -1,22 +1,22 @@
 /*
-** 'C'-style script compiler
-** Copyright (C) 2000-2001, Chris Jones
-** All Rights Reserved.
-**
-** This is UNPUBLISHED PROPRIETARY SOURCE CODE;
-** the contents of this file may not be disclosed to third parties,
-** copied or duplicated in any form, in whole or in part, without
-** prior express permission from Chris Jones.
-**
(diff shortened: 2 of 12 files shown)


ivan-mogilko:
Perhaps a note for the future, but imo it's best to do multiple commits when there are unrelated changes, like in this case the parallel compilation is not directly related to compiler warnings and may have its own issues. I don't think it's a "side effect" either.

Something I'd like to clarify, is it necessary to cancel compilation when errors are found in one of the scripts? There was this ticket  #738 proposing compiling everything if possible to report all found errors at once. That seem to be a logical next step as you gather all messages from all the scripts anyway.

I commented out the error check in the compilation foreach as an experiment, (and removed condition `if (!messagesToReturn.HasErrors)` which prevented adding more messages when there are already errors -- not sure why this is necessary...) and it seems working well.

ivan-mogilko:
So, I guess my only question remaining is, why is there this condition that does not let add any messages if there are errors found in the list?
https://github.com/adventuregamestudio/ags/pull/1319/commits/f554ac1bc3a577bb4b1cf80af04ddf63c8d92521#diff-9efc75afc36570774adb0d848e6daf642e283532476159f114054657502f9979R1004

This condition is only applied to the new compiler's results, but not to the old compiler.

ericoporto:
Oh, so this is the PR that made ags4 compile things so fast.

I am curious why this doesn't make it fast when using the old compiler, is it because since it runs as a library and the old compiler uses a lot of globals, there may be issues compiling things in parallel with it? 

I ask because as I mostly use ags3 for maintaining modules mostly, and I went back to my project after a hiatus and the time it takes to build everything everytime I change something in a header is way too much and being spoiled by ags4 I am looking if this is something that can be backported.

fernewelten:
> I am curious why this doesn't make it fast when using the old compiler in ags4

It isn't coded that way.

The old compiler uses globals extensively, in particular the symbol table lies in global static memory. Also, the old compiler uses some functions such as `cc_error()` that rely on global variables. I haven't checked whether bad things happen when you call the compiler in parallel, but I haven't been able to rule that out, either. 

On the other hand, the new compiler is encapsulated and only (*) uses local memory. The scaffolding around the compiler is responsible for calling the "global" functions. For instance, the compiler doesn't call `cc_error()` itself, but collects its error messages and passes them to the caller in a parameter. Then the caller is the one that will call `cc_error()`. So I trust that bad things won't happen when the compiler is called in parallel.

The coding checks whether the new compiler is called or the old one, and only with the new compiler does it attempt to run in parallel.

(*) The new compiler does write to some few global variables while it works, in particular to `currentline`, but those values are never used, the caller writes them anew before calling "global" functions that use them, so it doesn't matter if those variables get clobbered in parallel runs.

PR #1320, opened 2021-06-17
New compiler: Bug fix: Disallow "this" parameter for static extender function

fernewelten:
The new compiler let through:
```import static int Foo(this Bar *);```

Judging from the `static` keyword, the function `Foo()` must be static, but in this case it doesn't have any `this` pointer, which collides with `this Bar *` part. So this declaration is nonsensical.  This PR makes the compiler balk if whenever encounters that kind of code.

(The correct way of declaring a static extender function would be `import int Foo(static Bar);`).


ivan-mogilko:
By the way, I noticed that compiler allows to declare `static` imports outside of the struct.

Issue #1297, opened 2021-05-28
Unicode support


ivan-mogilko:
# Unicode support

This ticket's purpose is to gather a list of components in the editor and engine which will be necessary to adjust for unicode support, as well as any potential issues.

## Game data

Game data contains text as byte arrays. There's nothing preventing to write and interpret these as utf8 strings. The length fields will have to keep their meaning of _size-in-bytes_ because there's no direct conversion between length in bytes and string length in utf8.

It's preferable to remove remaining arbitrary length limits in data format, esp. if they are small, because non-latin unicode text requires more memory. What was barely enough for latin names may be not enough for non-latin ones.

We probably may need an identification of the text format in the game data. Of course we could use just a format number, but still it may be good to have text format id too for safety, and in case we will need to support variants in the future.
Each data file that may contain text should have it: main game data, room file, compiled translation (TRA), compiled script too.
(**UPDATE:** After (#1321) TRS/TRA formats support encoding hints.)

## Editor and tools

As a .NET application Editor itself is already unicode compatible at least in terms of UI.
The script editor is under question though and should be checked.

The most important part is link between editor's fields and native game data and serialization functions. This is where most strings are converted from unicode into byte-char ANSI strings.

Another point of interest is translation compilator. TRS source files themselves are just text, and could be saved in utf8. But current TRA compiler does not understand utf8 and writes faulty data.
(**UPDATE:** After (#1321) TRS/TRA formats support encoding hints, which translation compiler writes and engine is taught to check.)

Font preview in the editor should be redone. Currently it simply draws first 256 characters of the font onto bitmap and displays that bitmap. That would be not optimal neither convenient for a unicode font preview.

## Script format and compiler

Script compiler must support scripts in utf8. It may require an option telling whether to expect ascii or utf8 source and convert if necessary (provided this option indicates an ANSI codepage).
But I think even in ascii mode the utf8 scripts (without BOM) may be maintained correctly so long as there's no unicode characters in the code itself. The *texts* (in double quotes) are likely viewed just as arrays of chars by compiler, so they may be accepted and written as-is.

Speaking of utf8 scripts, there's an interesting question of whether we need to also support unicode in the script names, because editor will support it in property fields by default. We could if it's not too hard to do, then users would be able to give non-latin names to the code symbols... I won't call this a priority though. But while this is not done we'll have to keep converting script names to ascii when generating script headers.

## Script API

[...]

PR #1321, opened 2021-06-20
Partial Unicode support in the engine (UTF-8 translations)

### LINKED ISSUE #662: ags3 linux: can't open datafiles in path with utf8 chars
to reproduce:
- launch linux
- extract this game: https://0x0.st/zouS.rar (agsarchives.com/bicycle/mmm/MMM\ 016\ -\ Meteor\ Family\ -\ Die\ Rückkehr\ des\ Meteoriten.rar) with `7z x filename.rar`
- cd into the resulting directory
- `rm *.lnk ; mv Files/* .`
- ags mmgui.exe

results in 
```
Game data file could not be found. Search path used: '/home/rofl/games/ags/mmm/Maniac Mansion Mania 016 - Meteor Family - Die R�ckkehr des Meteoriten/mmmgui.exe'
ERROR: Unable to find game data files. The necessary files are either missing or are of unsupported format.Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]
```
it seems the utf8 sequence `R\303\274ckkehr` is transformed into `R\374ckkehr` during the code leading to search_for_game_data_file (filename=...)
    at main/engine.cpp:303

i see 2 issues here:
1) the obvious bug of modifying the path string
2) the usage of a full path rather than just `.` for the directory or even just the plain filename without any directory components, as passed on the command line

ivan-mogilko:
Addresses #1297
Resolves #662 (or so I hope)

This implements partial unicode (UTF-8, to be precise) support in the engine, with an effort to maintain backwards compatibility at the same time.

Supports loading translations saved as UTF-8 (*without* BOM), with an encoding hint (see below). Unicode in the game data itself (game properties and script) is not yet supported: these would require additional changes to editor, and maybe to game format.

Obviously, for displaying unicode text one would have to provide a proper unicode font (not the hacky ANSI-codepage-compatible fonts that are normally used with AGS).

Also file paths are now strictly UTF-8, which suppose to resolve #662.

---

### Implementation notes

The biggest problem with backward compatibility is that ASCII text in AGS is not a strict ASCII, as engine and editor let game devs to use extended char values (127 - 255) and write texts in corresponding ANSI codepage (of course that requires fonts compatible with that codepage). For this reason we cannot just replace all string functions with utf-8 equivalents so long as we want to load old games.

Current solution utilizes text functions provided by allegro library, because they may switch meaning depending on text format setting. Thus same engine code may work for both ascii and utf-8, with the few exceptions.

When the game is loaded the text format is set to ASCII (current default). When a translation is loaded the engine will check its encoding hint and switch to either ASCII or UTF-8 (ASCII is applied also if no hint is found).

Previously I was thinking about having to switch encoding every time when the engine has to work with filepaths (and then back), but in the end decided to take alternate route. Instead I adjusted allegro library code (which is embedded in the engine now), and made all the file functions always use UTF-8 unconditionally. This is done through macros to reduce changes in code.

---

### Translation

TRS (translation source) now supports two additional options: "Encoding" and "GameEncoding" along with the existing ones in the header. For "Encoding" the only supported value really is "UTF-8", and any other, or no value, is interpreted as "ASCII".

"GameEncoding" is a hint telling how to interpret original game texts. This is only useful if the game itself contains extended ASCII. In this case they have to be converted to match UTF-8 keys from TRA. Having "GameEncoding" hint helps engine to do so.
"GameEncoding" is suppose to be a C locale name, most usually in a form of ".125x", where the number specifies the Windows codepage set on the system when the game data was compiled.

Example of TRS header with Encoding param:
<pre>
// ** Translation settings are below
// ** Leave them as "DEFAULT" to use the game settings
// The normal font to use - DEFAULT or font number
//#NormalFont=3
// The speech font to use - DEFAULT or font number
//#SpeechFont=3
// Text direction - DEFAULT, LEFT or RIGHT
//#TextDirection=DEFAULT
// Text encoding hint
//#Encoding=UTF-8
// Source text encoding hint
//#GameEncoding=.1251
</pre>

Translation compiler is adjusted accordingly to write these values into the TRA file.

---

### Expected behavior

Any existing game or older translation file should still load and display as before.

Editor should update UTF-8 TRS file correctly with new texts, not damaging existing ones.

Engine is capable of starting a game in a directory which path contains unicode characters, on any system.

Loading UTF-8 translation will make following unicode-aware:
* Text display (at least with TTF fonts, but I suspect WFN will work too if it contains glyphs in proper indices);
* String API (also see notes below);
* Text input in the TextBox control.

---

### Script API

Script String class API was changed only in implementation for now, and every "index" or "length" value is interpreted in characters (as opposed to bytes). In my opinion this is what will be commonly expected by users (raw array of chars may be created by other means in script).

Still there are functions that accept or return character as `char` - that is "byte", - I did no changes to these yet.

---

### Potential TODO

* Use Wide-char file/path functions on Windows (get rid of the ugly DOS paths); **EDIT:** *not critical, maybe I will do this separately;*
* BOM detection in TRS files for translation compiler; **EDIT:** *not critical*
* Text parser made unicode-aware; **EDIT:** *there is a larger issue with the parser: its words are not included into translation; that problem should be addressed on its own.*
* String API working with unicode characters (not yet sure if and how this is doable in ags script); **EDIT:** *perhaps better leave for a separate task.*

---

### Potential issues

* There is a group of languages with special rules to gluing characters together, examples are Arabic and Persian, but maybe some more. This is currently not supported. I'm unaware if our current font library (alfont) can do that, but probably not. In any case this is not going to be a part of this PR, and requires additional work.
* If the game contains text with extended ASCII (127 -255) and you load UTF-8 translation, then the original untranslated text containing these extended ascii will not be displayed correctly. The solution now is to provide *any* translation for these lines, even if you just copy the original one.

---

Test game I made for myself:
[test--unicode-source.zip](https://github.com/adventuregamestudio/ags/files/6681700/test--unicode-source.zip)
[test--unicode-game.zip](https://github.com/adventuregamestudio/ags/files/6681701/test--unicode-game.zip)

The game source contains several TRS files which may be copied, filled with different translations and built into TRA.
If you cannot use the Editor, you may also try using stand-alone tool `trac`, which may be built either with a [Makefile](https://github.com/adventuregamestudio/ags/tree/master/Tools/trac) or with Tools.sln [MSVS solution here](https://github.com/adventuregamestudio/ags/blob/master/Solutions/).

In the game itself:
* Space key - starts String API test
* T key - brings up GUI with text box, which fills a list by pressing Enter
* D key - brings up GUI with list box, filled by game directory contents (this is to test dirs with unicode characters)
* S key - plays a sound - this is to test that engine loads up audio file from the other pack/dir, again in case there are unicode characters in directory name, to be certain that nothing is broken there.

### DIFF
--- a/Common/ac/keycode.h
+++ b/Common/ac/keycode.h
@@ -19,6 +19,7 @@
 #define __AGS_EE_AC__KEYCODE_H
 
 #include "core/platform.h"
+#include "core/types.h"
 
 #define EXTENDED_KEY_CODE ('\0')
 #define EXTENDED_KEY_CODE_MACOS ('?')
@@ -243,6 +244,18 @@ enum eAGSKeyCode
     case 425: __allegro_KEY_NUMLOCK
     case 426: __allegro_KEY_CAPSLOCK
     */
+};
+
+
+// Combined key code and a textual representation in UTF-8
+struct KeyInput
+{
+    const static size_t UTF8_ARR_SIZE = 5;
+
+    eAGSKeyCode Key = eAGSKeyCodeNone;
+    char        Text[UTF8_ARR_SIZE] = { 0 };
+
+    KeyInput() = default;
 };
 
 
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -205,6 +205,8 @@ void unescape_script_string(const char *cstr, std::vector<char> &out)
         cstr++;
     }
     // Replace all other occurrences as they're found
+    // NOTE: we do not need to decode utf8 here, because
+    // we are only searching for low-code ascii chars.
     const char *off;
     for (off = cstr; *off; ++off)
     {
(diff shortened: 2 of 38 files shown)


ivan-mogilko:
For the reference, I found couple of ags users who are testing this atm.

Last updates:
* Editor taught to update UTF-8 TRS files without corrupting them.
* TRS/TRA now will have an optional hint about game source encoding. I put it there to also support case when translation is added for the already existing older game without recompiling it with the new AGS. This hint will be used to help matching ANSI source texts with UTF-8 TRA dictionary keys.

ivan-mogilko:
I think this is working as described; my tester said she had it working with several languages too.

Old games with extended ASCII texts also load fine as much as I can tell.

So I'll be merging this, unless someone sees any serious problem here.

Issue #1265, opened 2021-04-26
Feature request: Add FreeBSD to build targets


PaddyMac:
The current version of AGS for FreeBSD is 3.4.4.1 from 2019. I'd like to have a newer version available. I've been tinkering with it, and after lying to the build system and making it think it's building on Linux, everything works fine 0n 3.5.0.31 if I use GCC up until the final linking phase when I get some undefined reference errors. There will need to be a little extra work to get it working with Clang which is the FreeBSD native compiler. In any case, I'm not much of a programmer, but I'd certainly be willing to put in some effort on my part to get this working. Somewhat tangentially, could you make the build system accept the use of more system libraries? Other than Allegro, it doesn't look like you're using customized versions of external libraries. I mention this partly because I tried compiling the 3.6 series of AGS, and I ran into issues compiling the included SDL2 library which I also ran into previously when trying to manually compile an unpatched SDL2 on FreeBSD. So this issue could be avoided entirely if it were possible to use the system library instead of a bundled one.

rofl0r:
if you use Engine/Makefile, only strictly required library components are built - to my knowledge that's only the embedded freetype copy. i've been using that makefile since forever with system-provided allegro4 library, there was never a need for a especially patched version. you should however use allegro 4.4.3 as it has all a number of bugfixes and features to support AGS better.
as for SDL2 version aka 3.6: the code uses libsdl-sound in the SDL2 incarnation, which has so far not seen a release (however one was recently promised to debian maintainers in a mail that has been linked somewhere here). you gotta build that on your own from mercurial/git and install it systemwide, but apart from that you can use Engine/Makefile just fine.

> Add FreeBSD to build targets

i don't think a freebsd-specific build target needs to be added, it should build just fine using the makefile, unless there are some `#ifdef __linux__` or similar defines, in which case maybe the "linux target" should be remodeled as a "generic posix" target.

i guess a good start would be if you report the specific errors you're experiencing so those can be examinated on a case-by-case basis.

ivan-mogilko:
@PaddyMac 
I'd like to clarify, you are refering to Cmake scripts and want to add FreeBSD to our Cmake scripts, is that correct?

As far as I know it's Cmake that fetches our patched version of Allegro 4 for Linux, even though today this may not be completely necessary (at least from the point of functionality). AGS should work well with their official latest release (4.4.3.*) which already contains all the patches we were sending to them over last couple several years.
[...]

PR #1324, opened 2021-06-21
adds FreeBSD platform

ericoporto:
I ended up adding a platform driver named Generic Unix, but I specialize it for FreeBSD so I can have a `eOSFreeBSD`. I am not sure if this is the best approach, in my head it makes easier to add more platforms. 

I only used CMake since it's written to build the most portable binary as possible - so it's better for actually shipping binary games.

For system packages is better that the person packaging it writes their own build aware of whatever packages are available in the specific FreeBSD version. Unfortunately no packager ever returned any of my e-mails, so this may take a while.

Still, the system specifics defines in headers are now alright for FreeBSD.

I don't plan to provide CI builds for FreeBSD.


Common/core/platform.h
```diff
@@ -32,36 +34,48 @@
     #endif
 
     #if TARGET_OS_SIMULATOR || TARGET_IPHONE_SIMULATOR
-        #define AGS_PLATFORM_OS_WINDOWS (0)
```
rofl0r, line 35: i don't know about AGS stance on this, but in other projects i contribute too the general consensus is to not add gratuitous whitespace changes to unrelated changes.

ericoporto: They are not gratuitous, I will need the additional space soon and I don't know yet when this will be merged.

ivan-mogilko: I don't quite understand though, what are these spaces for?

ivan-mogilko:
If I understand correctly, this should resolve #1265 ? Wanted to reference it here just in case.

Common/core/platform.h
ericoporto: I just wanted to avoid merge conflicts with the Emscripten change in case this takes a while - it may not be trivial to setup a FreeBSD VM.

ericoporto:
Yes, but not fully, the user clearly has an expectation to be able to download from ports but I haven't been able to reach any of the people that did the commits from that port. See the commit history below:

https://www.freshports.org/games/ags

Ideally someone would need to contact @mat813 or @t6 which are the maintainers (at least I think, neither answered e-mails).

Common/core/platform.h
ivan-mogilko: My point was, you do not have to do this in Emscripten pr either, there's no real need to adjust the lengths of all the lines if you add a new one which is slightly longer.

ivan-mogilko:
The biggest question I have in regards to this change is, what is the difference between AGSGenericUnix and AGSLinux platform driver? May it make sense to make AGSLinux also a child of AGSGenericUnix?

ericoporto:
So, this is what I was most curious, right now there's no difference, and I think it could, except that the old CD support (does someone actually ever used that?) actually is there on Linux, but I have removed it from Generic Unix - I doubt anyone will ever test and verify, when I imagine sbc like Dragonfly picking it up or NetBSD.

Other thing is I temporarily left XDG support there (because it IS available on FreeBSD as far as I can tell), but I don't think XDG is something in all Unixes, only those that adhere to Freedesktop. 

So in my head it's like:

base (which could use SDL stuff) -> Generic/Posix Unix -> Unix with Freedesktop -> GNU/Linux

Common/core/platform.h
ericoporto: I will rebase and these will not be in this PR.

rofl0r:
i'd model it like so:
basically there's windows vs posix  (which means nothing else than "generic unix") code, and a os selection which turns on or off via macro certain platform specific code

e.g.
```c
#ifdef __linux__
#define OS_LINUX
#define PLATFORM_POSIX
#define PLATFORM_SUPPORTS_CDA
#define PLATFORM_SUPPORTS_XDG
#endif
```
etc.

> Other thing is I temporarily left XDG support there (because it IS available on FreeBSD as far as I can tell), but I don't think XDG is something in all Unixes, only those that adhere to Freedesktop (may need to be moved around).

tbh not even linux is so clear-cut, basically "linux" means nothing more than "uses linux kernel" but everything else can be different for every single distribution.

ericoporto:
@rofl0r do you know if there's a way to detect XDG support? Better if at compiler time but alright if runtime, so we could something like, use these folders but if can't use these?

Also, should I separate the generic_unix from the tiny FreeBSD code? My mentality on leaving together is because the macros are defining what will build and what won't be built, so one could lump it on together when building and the compiler will select what makes sense.

I think I can add the `PLATFORM_SUPPORTS_CDA` directly on core/platform.h too. I don't like extremely long files but there it really saves time finding things together - I even thought on moving the library loading selection there but left out because it has 3 states instead of 2.

ivan-mogilko:
> I think I can add the PLATFORM_SUPPORTS_CDA directly on core/platform.h too. I don't like extremely long files but there it really saves time finding things together

I don't think it's too long at the moment, but in the worst case we may split it into 2, where one has platform detection and second has options defined depending on the platform.

ericoporto:
Found something interesting, SDL simply checks if a HOME environment variable exists (https://github.com/libsdl-org/SDL/blob/main/src/filesystem/unix/SDL_sysfilesystem.c)

I am thinking on having Generic Unix without libcda, and have it do the XDG test but slightly better at fallback and then have both Linux and FreeBSD build on it, switching the OS enum, and have Linux have libcda.

Engine/CMakeLists.txt
```diff
@@ -419,6 +419,8 @@ target_sources(engine
     media/video/video.h
     platform/base/agsplatformdriver.cpp
     platform/base/agsplatformdriver.h
+    platform/base/agsplatform_unix.cpp
+    platform/base/agsplatform_unix.h
     platform/base/sys_main.cpp
```
rofl0r, line 424: the cpp file would need to be added to Engine/Makefile too i guess ?

ericoporto: so, the base platform apparently has some magic that makes unnecessary, because platform/base is already added.

https://github.com/adventuregamestudio/ags/blob/811f6c126fcff1cc5a0db8b6e0a1b9fca2550235/Engine/Makefile-objs#L30

but check the commit below, that one would be needed if there was a bsd makefile (I think)

ericoporto:
@rofl0r I am not particularly confident on the names I chose for things, do you think the hierarchy makes sense?

rofl0r:
> @rofl0r I am not particularly confident on the names I chose for things, do you think the hierarchy makes sense?

well, it's not quite as i lined out but looks OKish so far. i only don't like the "freedesktop platform" bit, which makes it sound as if freedesktop is a hugely important thing, when this is really just about XDG home directory layout. i'd just call it something like "uses_xdg", "has_xdg", "with_xdg", "supports_xdg" or something like that.

> the base platform apparently has some magic that makes unnecessary, because platform/base is already added.

oh great, i forgot we use a glob there.

ericoporto:
Uhm, maybe `AGS_PLATFORM_IS_XDG_UNIX` ?

ivan-mogilko:
Since you have a "AGSPlatformUnix" there, why not call it just AGS_PLATFORM_UNIX?

There are no other subtypes there at the moment, and we don't know if we will have any in the future.

ivan-mogilko:
I think [rofl0r's original suggestion](https://github.com/adventuregamestudio/ags/pull/1324#issuecomment-865115436) was to consider XDG to be rather an optional component rather something that defines whole platform type?

What is the purpose of this base class, which functionality it shares to all common children?

What if we just call it "desktop unix" as opposed to mobile (e.g. Android). And then you may toggle XDG support with additional macro inside the class if required.

ivan-mogilko:
Sorry nevermind, I removed some of my previous comments. Apparently I misunderstood what XDG really is.

TBH personally I would just call it "desktop unix" for now, to distinct from mobile and as something that would be clear to most developers.
But if "XDG unix" is a more correct term for this, okay then.

ericoporto:
Yes, I agree, this is why I didn't want the name to be *just* Unix, thought it would be misleading.

ivan-mogilko:
@ericoporto if I understand correctly, this PR is practically ready?
One thing, could you also rename `AGSPlatformUnix` to have XDG in the name to match the platform definition?

ericoporto:
it's a bit long but it's done :grimacing:

PR #1325, opened 2021-06-21
Platform additions emscripten

ericoporto:
these are platform specific checks that are used in the Emscripten port, I am adding them here in case they are easier to understand than FreeBSD ones to avoid future conflicts.


ivan-mogilko:
I honestly don't think it's necessary to adjust all these macro defines, you could simply add a new one even if it's longer than the others - that would be fine.

PR #1330, opened 2021-06-27
Engine: support scripts as separate assets

ivan-mogilko:
**Summary**

Support loading scripts as separate asset files. Allow to *not* have scripts inside game28.dta and room.crm.

This does not change any data formats, only reinterprets one existing flag in the main game data (as explained below).

The purpose of this is to simplify the data layout in the game package, and the packaging process itself, as this will require at least 1 step less for game and room compilation (compiled script injection into the respective data blob).

**Details**

This allows engine to support game distribs where scripts are not embedded into the game28.dta and roomN.crm blobs, but are present as separate asset files inside the game package.

First of all, the engine will now treat `game.load_compiled_script` flag as an indication whether the scripts are found inside main game data file or not, and won't immediately raise an error if it's 0 (false).

After the main game data is loaded the engine will look up for script files of hardcoded names for Global and Dialog scripts, and then for the script modules. These are optional. If they are not found no error is raised. If they exist then they will be loaded, replacing any scripts found in the main game data if these were also present.

There are two hardcoded script file names: **"GlobalScript.o"** and **"DialogScript.o"**.

For the rest of the script modules, engine first tries loading an asset called **"ScriptModules.lst"** which is supposed to contain simple textual list of filenames *separated by linebreaks*. These files are then loaded as script modules in the given order.
This is not hardcoded, but the suggested names for these modules is again **"scriptname.o"**.

Regardless, GlobalScript must exist anyway. If it's not found at all (neither in main data blob nor as a separate asset), the error is raised.

Similar change is done for the room scripts. Except room scripts are an optional block in the room format, so there was no need for any flag.

Room script should be called **"roomN.o"** where N is a room's number.

**PS.** Naturally these assets may also be loaded from the disk, as any other game data in AGS.

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -71,8 +71,6 @@ String GetMainGameFileErrorText(MainGameFileErrorType err)
         return "Failed to deserialize custom properties schema.";
     case kMGFErr_InvalidPropertyValues:
         return "Errors encountered when reading custom properties.";
-    case kMGFErr_NoGlobalScript:
-        return "No global script in game.";
     case kMGFErr_CreateGlobalScriptFailed:
         return "Failed to load global script.";
     case kMGFErr_CreateDialogScriptFailed:
@@ -802,17 +800,18 @@ HGameFileError ReadGameData(LoadedGameEntities &ents, Stream *in, GameDataVersio
     game.read_interaction_scripts(in, data_ver);
     game.read_words_dictionary(in);
 
-    if (!game.load_compiled_script)
-        return new MainGameFileError(kMGFErr_NoGlobalScript);
-    ents.GlobalScript.reset(ccScript::CreateFromStream(in));
-    if (!ents.GlobalScript)
-        return new MainGameFileError(kMGFErr_CreateGlobalScriptFailed, ccErrorString);
-    err = ReadDialogScript(ents.DialogScript, in, data_ver);
-    if (!err)
-        return err;
-    err = ReadScriptModules(ents.ScriptModules, in, data_ver);
-    if (!err)
-        return err;
+    if (game.load_compiled_script)
+    {
+        ents.GlobalScript.reset(ccScript::CreateFromStream(in));
+        if (!ents.GlobalScript)
+            return new MainGameFileError(kMGFErr_CreateGlobalScriptFailed, ccErrorString);
+        err = ReadDialogScript(ents.DialogScript, in, data_ver);
+        if (!err)
+            return err;
+        err = ReadScriptModules(ents.ScriptModules, in, data_ver);
+        if (!err)
+            return err;
+    }
 
     ReadViews(game, ents.Views, in, data_ver);
 
--- a/Common/game/main_game_file.h
+++ b/Common/game/main_game_file.h
@@ -56,7 +56,6 @@ enum MainGameFileErrorType
     kMGFErr_TooManyCursors,
     kMGFErr_InvalidPropertySchema,
     kMGFErr_InvalidPropertyValues,
-    kMGFErr_NoGlobalScript,
     kMGFErr_CreateGlobalScriptFailed,
     kMGFErr_CreateDialogScriptFailed,
     kMGFErr_CreateScriptModuleFailed,
--- a/Engine/ac/room.cpp
+++ b/Engine/ac/room.cpp
@@ -448,6 +448,26 @@ static void update_all_viewcams_with_newroom()
     }
 }
 
+// Looks up for the room script available as a separate asset.
+// This is optional, so no error is raised if one is not found.
+// If found however, it will replace room script if one had been loaded
+// from the room file itself.
+HError LoadRoomScript(RoomStruct *room, int newnum)
+{
+    String filename = String::FromFormat("room%d.o", newnum);
+    std::unique_ptr<Stream> in(AssetMgr->OpenAsset(filename));
+    if (in)
+    {
+        PScript script(ccScript::CreateFromStream(in.get()));
+        if (!script)
+            return new Error(String::FromFormat(
+                "Failed to load a script module: %s", filename.GetCStr()),
+                ccErrorString);
+        room->CompiledScript = script;
+    }
+    return HError::None();
+}
+
 // forchar = playerchar on NewRoom, or NULL if restore saved game
 void load_new_room(int newnum, CharacterInfo*forchar) {
 
@@ -486,6 +506,11 @@ void load_new_room(int newnum, CharacterInfo*forchar) {
         (thisroom.GameID != game.uniqueid)) {
             quitprintf("!Unable to load '%s'. This room file is assigned to a different game.", room_filename.GetCStr());
     }
+
+    HError err = LoadRoomScript(&thisroom, newnum);
+    if (!err)
+        quitprintf("!Unable to load '%s'. Error: %s", room_filename.GetCStr(),
+            err->FullMessage().GetCStr());
 
     convert_room_coordinates_to_data_res(&thisroom);
 
--- a/Engine/game/game_init.cpp
+++ b/Engine/game/game_init.cpp
@@ -124,6 +124,8 @@ String GetGameInitErrorText(GameInitErrorType err)
         return "Too many plugins for this engine to handle.";
     case kGameInitErr_PluginNameInvalid:
         return "Plugin name is invalid.";
+    case kGameInitErr_NoGlobalScript:
+        return "No global script in game.";
     case kGameInitErr_ScriptLinkFailed:
         return "Script link failed.";
     }
@@ -442,6 +444,8 @@ HGameInitError InitGameState(const LoadedGameEntities &ents, GameDataVersion dat
     // NOTE: we must do this after plugins, because some plugins may export
     // script symbols too.
     //
+    if (!ents.GlobalScript)
+        return new GameInitError(kGameInitErr_NoGlobalScript);
     gamescript = ents.GlobalScript;
     dialogScriptsScript = ents.DialogScript;
     numScriptModules = ents.ScriptModules.size();
--- a/Engine/game/game_init.h
+++ b/Engine/game/game_init.h
@@ -40,7 +40,8 @@ enum GameInitErrorType
     kGameInitErr_EntityInitFail,
     kGameInitErr_TooManyPlugins,
     kGameInitErr_PluginNameInvalid,
-    kGameInitErr_ScriptLinkFailed
+    kGameInitErr_NoGlobalScript,
+    kGameInitErr_ScriptLinkFailed,
 };
 
 String GetGameInitErrorText(GameInitErrorType err);
--- a/Engine/main/game_file.cpp
+++ b/Engine/main/game_file.cpp
@@ -39,6 +39,7 @@
 #include "gfx/blender.h"
 #include "core/assetmanager.h"
 #include "util/alignedstream.h"
+#include "util/textstreamreader.h"
 #include "ac/gamesetup.h"
 #include "game/main_game_file.h"
 #include "game/game_init.h"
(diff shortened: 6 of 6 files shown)


ericoporto:
So, this means one day I will be able to hit save on the room editor when I haven't finished writing the script and not get annoying errors?

ivan-mogilko:
> So, this means one day I will be able to hit save on the room editor when I haven't finished writing the script and not get annoying errors?

Hmm, I did not think about that. Maybe this will be a consequence too.
Same will happen because of the room format change too (#1278).

ericoporto:
It would be nice to have a config to always prefer packaged scripts to avoid someone making an evil with the intent of pulling game assets.

rofl0r:
> evil script with the intent of pulling game assets

what do you think could such evil script do ?

ericoporto:
Extract all images without even needing an external tool.

rofl0r:
you'd already need an external tool, at least external script compiler, to create that "evil script stealing images", in which case it would be a lot simpler to just use `agsprite` tool to rip the images.

> Assets are licensed from the artists to the gamedev with the purpose of making the game, not for general use - which would be more expensive.

if someone rips images from say "call of duty", can he then legally use those images for other works? nope. just because someone can technically get to licensed content, it doesn't change the license of the content.

ivan-mogilko:
There was at least one instance of another user asking for this in the past. I would see this as a separate issue of course, as this is related not only to scripts but any assets.

Frankly I've been always reluctant to do this, as on one hand there are already multiple tools extracting resources from AGS games (some of them exist for decades), so if the person is dedicated they will find what they want. On another hand changing this priority will close certain opportunities like quickly patching the game without repackaging, or providing new translations with new font files.

I feel like this is a serious matter that has to be discussed openly first.

At the same time I'm 100% sure that if we do this, someone will immediately make a version of an engine where this is reverted.

ericoporto:
I think if you ask around if people want their games edited by a third party you will find gamdevelopers have the same opinion as me. 

> At the same time I'm 100% sure that if we do this, someone will make a version of an engine where this is reverted

Well, sure, but good luck to that person maintaining this. Do you prefer one fork for EACH gamedev reverting this instead? 

If you don't want to check this bit for this switch for some reason, then we should start thinking about shipping two engines with the Editor, the one for hacks, and the one that actually gets shipped with the game.

ivan-mogilko:
> I think if you ask around if people want their games edited by a third party

The question is not whether they want their games edited by someone or not. Their games may already be edited, because all the necessary tools exist.

Do you realize that as soon as we provide the full collection of standalone tools with the Editor, *anyone* will be able to repack the game data in anyones game? So all this protection will just fly into the window. The only way to protect your game from editing (more or less) is to create a closed-source data format.

ivan-mogilko:
Well, I mean, we can do this of course. Most people probably won't notice the difference, those who used this will hate.

fernewelten:
When I learned how the (old) compiler worked, I needed to know the __exact__ inputs that would be passed to the compiler. So I made a small stub at the start of the compiling function that wrote the input to a file called `scriptname`-`timestamp`.`asc` before processing it further. That was an effort of  ~5 lines. 

In the same vein, packaging a game isn't the same thing as a copy protection: The Engine must by needs be able to unbundle and read all the assets. So it's probably very easy to modify it in such a way that it writes all the assets out to some file before processing them or acting on them. As soon as there is some way to run a game with your own Engine, out goes the "bundling" copy protection.

The evil unbundler only needs to patch __one__ Engine __once__ with some file writing calls: Even if our Engine gradually evolves, the patched evil Engine will be still functional enough for a long time to do the evil writing work; afterwards it can gladly crash because the deed has been done. And when the old Engine isn't functional enough any longer at long last, you can probably simply cherry-pick the old changes onto the newest Engine version, and chances are quite high that git will integrate them without even needing manual intervention.

We might perhaps implement some public key / private key encryption system to protect the integrity of the files. The Engine would only be able to DEcrypt them, but to ENcrypt a changed file, the developer's private key would be needed. 

But even then, that would not prevent an evil decrypter from unpacking everything, changing things around and then packing it all again with a one-time encryption key that they make up for the occasion.

ericoporto:
_graph of what I said:_
- Tools (our or third party) -> fair game
- The engine alone that's in the released game -> no

> public key / private key encryption 

Let's not go there.

ivan-mogilko:
My primary question here is: what's the point of changing anything if that won't lead to the goal (game protection), but only annoy a number of people? Or increase the project maintenance.

ivan-mogilko:
In regards to this PR though, is it acceptable as is (so long as it works as expected), or is this concern about safety too high to let it be merged at the moment?

If it's okay, I'd like to propose moving the game protection discussion to the dedicated ticket, because there are already several variants suggested above (like providing separate engines for release and debugging).

ivan-mogilko:
@ericoporto I gave this a quick thought, and even if we go with "release" engine build, I'd rather also have an option in game settings for this, as a backup plan in case someone wants to support game modding or something.

I.e. the engine flag may define the starting priority setting here:
https://github.com/adventuregamestudio/ags/blob/a3c6e7cfded18daadab00474ff6d148904581f08/Engine/main/main.cpp#L109


but then it loads game data and adjusts according to what is specified in a dedicated OPT_* value.

rofl0r:
is there some kind of mechanism to prevent the engine from trying to open these separate files over and over?
i have this debug log from latest PR improving ci_find_file:

```
fast path /home/rofl/games/ags/Strangeland-GOG/game.ags
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
fast path /home/rofl/games/ags/Strangeland-GOG//speech.vox
fast path /home/rofl/games/ags/Strangeland-GOG//speech.vox
fast path /home/rofl/games/ags/Strangeland-GOG//speech.vox
fast path /home/rofl/games/ags/Strangeland-GOG//audio.vox
fast path /home/rofl/games/ags/Strangeland-GOG//audio.vox
fast path /home/rofl/games/ags/Strangeland-GOG//audio.vox
file not found /home/rofl/games/ags/Strangeland-GOG/ game28.dta
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ GlobalScript.o
file not found /home/rofl/games/ags/Strangeland-GOG/ DialogScript.o
file not found /home/rofl/games/ags/Strangeland-GOG/ ScriptModules.lst
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt0.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt1.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt2.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt3.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt4.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt5.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt6.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt7.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt8.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt9.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt10.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt11.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt12.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt13.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt14.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt15.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ agsfnt16.ttf
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ acsprset.spr
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ sprindex.dat
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ room41.crm
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ room41.o
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound59.sfx
file not found /home/rofl/games/ags/Strangeland-GOG/ room1.crm
fast path /home/rofl/games/ags/Strangeland-GOG//game.ags
file not found /home/rofl/games/ags/Strangeland-GOG/ room1.o
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound456.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound288.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound34.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound261.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound261.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound261.sfx
fast path /home/rofl/games/ags/Strangeland-GOG//Sounds/sound34.sfx
```

ivan-mogilko:
> is there some kind of mechanism to prevent the engine from trying to open these separate files over and over?

If you mean *.o files, global scripts are only looked up once, and room.o are looked up whever a room is loaded. Is this an unwanted behavior?

rofl0r:
well, i just noticed in the output above that engine tries to load first any files that are inside .ags/.exe file from fs, and when it fails from .ags pack, since any failed try involves 2 chdir() and comparison with any filename in those directories <strike>using stat syscall on any file</strike>(i just fixed this in my PR)

ivan-mogilko:
This is how engine works by default: it first always checks for a file on disk and then inside the package (this was discussed above in this thread).

Since there was already a request for changing this behavior by a game option, if this is desired we may also have a config or a command-line option to set opposite reading order.

rofl0r:
well, my though was that there may be a convention like "if globalscript.o isn't found, then don't try to load any other .o/.crm files from fs from now on"

ivan-mogilko:
> well, my thought was that there may be a convention like "if globalscript.o isn't found, then don't try to load any other .o/.crm files from fs from now on"

This priority read rule existed since the beginning of AGS, and support for separate script files did not change anything in this regard.
I do not think that the abscence of one asset, even if it's more important that the others, justifies such change in behavior. E.g. there may be no globalscript.o, but there could be a room.o added for a test. Other assets are completely unrelated from scripts.

PR #1335, opened 2021-06-30
SDL_sound has moved to github (fix CI failing)

ericoporto:
- SDL_sound is not on icculus mercurial website

icculus switched SDL_sound off from mercurial and is now pushing to his GitHub, so we can't fetch from the old place anymore. Link to new home:

https://github.com/icculus/SDL_sound

PR changes Docker/CMake to fetch from the new place. 

Since previous commit number was now invalid with the system change (hg->git), I took the opportunity to update to the newer version, but there's no significant change/improvements to the formats we use more often.

### DIFF
--- a/CMake/FetchSDL_Sound.cmake
+++ b/CMake/FetchSDL_Sound.cmake
@@ -1,7 +1,7 @@
 FetchContent_Declare(
     sdlsound_content
-    URL https://hg.icculus.org/icculus/SDL_sound/archive/4b78007f035f.tar.gz
-    URL_HASH MD5=ff2448f5a789ae2a07636c243d8f8780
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Oh so that's why it was failing yesterday.

ivan-mogilko:
@ericoporto could you clarify, is the SDL_sound built with timidity part for midi to wav support, or not?

ericoporto:
by default SDL_sound builds with everything. But I don't understand... 

> timidity part for midi to wav support

what does this means?

ivan-mogilko:
That's what timidity's readme sais:
> TiMidity is a MIDI to WAVE converter.

ericoporto:
Ah, it does. Or it should, but I haven't tried to use. Timidity requires the sound fonts and a timidity.cfg file. 

SDL_sound switched from using the rewritten version from before to a different version that should be more complete - I think the version that was being used in SDL_mixer, but not sure.

PR #1342, opened 2021-07-04
CI add existing tests on Linux and MacOS CMake pipelines

ericoporto:
I am adding the existing tests to the CI.

I think it's almost done, but I would like to wait the CI to run everything.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -41,7 +41,12 @@ build_linux_cmake_task:
     arch=$(dpkg --print-architecture)
     filename=${BUILD_TYPE}_$arch
     mkdir build_$filename && cd build_$filename
-    cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE .. && make
+    cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DAGS_BUILD_COMPILER=1 -DAGS_TESTS=1  .. && make
+  test_linux_script: |
+    arch=$(dpkg --print-architecture)
+    filename=${BUILD_TYPE}_$arch
+    cd build_$filename
+    ctest --output-on-failure
   binaries_artifacts:
     path: build_*/ags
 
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Why is it necessary to replace strings like
<pre>
" text \
   more text "
</pre>
with
<pre>
" text "
" more text "
</pre>
?

ericoporto:
It was giving me an error with gcc 9 on Ubuntu, but I didn't investigate further.

ericoporto:
The PR does what it was supposed to do, but the bug it's catching doesn't happen in ags codebase as far as I can tell (t_int64 in a struct after a short). If merged as is it won't fail the pipeline since there's no one depending on that particular build.

ivan-mogilko:
> any idea why the aligned stream test would fail on 32 bit Debian?

For the reference, is it a little endian or big endian debian?

AlignedStream was created to take care of reading and writing existing game formats, where structs were saved raw from memory, with padding. So to keep the data format engine has to use AlignedStream.
It was *supposed* to work everywhere producing same result, using byte conversion functions that are defined differently depending on a platform. If it does not work somewhere then it should be fixed of course.

Ideally we should get rid of it in ags4 (unless this isn't already done). It should not be used in saves since the new save format was introduced in 3.4.0.

> And it's weird considering there's a proper function to serialize array of bytes.

Well, it *is* a proper function to serialize array of bytes. It's argument may be remade into `char*` to emphasize what it does, if that matters.

ericoporto:
> For the reference, is it a little endian or big endian system?

It's little endian, and the `TEST_BIGENDIAN` is also not defined.

ivan-mogilko:
To clarify more on AlignedStream maybe, the way it calculates padding *should not depend on running system*. It is made purposedly to emulate padding as done by *the old engine* on a 32-bit Windows, because that is how all the old games were serialized.

ivan-mogilko:
I see you added detailed asserts, to clarify, it does everything right until `i64b` member and then fails on it?

ericoporto:
> it does everything right until i64b member and then fails on it?

Yes, exactly. I poked around a lot but I am out of ideas.

ivan-mogilko:
Ok, trying to avoid any possible confusion, so to reiterate main points:

* AlignedStream is made to automatically calculate and read raw memory data written by very particular engine compiled for particular system (old AGS windows engine). This is purely for backward compatibility (to match existing file formats).
* In the test the memory data is written by *current system* (whatever the test is running on).

So, probably this is a wrong test to run on any random system.

I think more correct test would be to not do all that TrickyData writing in code, but instead define a hard-coded array with data filled as if it were written on 32-bit Windows with padding etc, and feed that array to Aligned Stream.
That is to make sure that AlignedStream works with exact data form it was designed for.

ericoporto:
> I think more correct test would be to not do all that TrickyData writing in code, but instead define a hard-coded array with data filled as if it were written on 32-bit Windows with padding etc, and feed that array to Aligned Stream.

With this in mind, I would suggest:

- moving the writing part of the method that does the writing to a new method, and make so the memory gets written to a file (say `alignedstreamtest.bin`)
- run this writing method once in a Windows computer, pickup the resulting bin file, and commit it in the repository. 
- hide this new method behind a `#if 0` macro.
- redo the remaining reading part of the Test method to read the data from this file.

Alternatively, serialize this binary file to a text file and copy the array in the code to avoid reading files in the test.

ivan-mogilko:
> Alternatively, serialize this binary file to a text file and copy the array in the code to avoid reading files in the test.

I'd rather have this than a file.


Hmm, if that fails than there might be something wrong with types and sizeof.
This may be a silly though, but, can we check how int64_t is declared on that failing system?

ericoporto:
> This may be a silly thought, but, can we check how int64_t is declared on that failing system?

I have it mounted on my machine, but I don't know how to do this without an IDE :grimacing: . Is there a code that I can compile that will tell something useful?

ivan-mogilko:
> I have it mounted on my machine, but I don't know how to do this without an IDE grimacing . Is there a code that I can compile that will tell something useful?

maybe this:
<pre>
printf("%zu", sizeof(int64_t));
</pre>

ericoporto:
```
./hello 
8, 8
```
from:

```C++
#include <cstdint>
#include <stdio.h>
int main() {
printf("\n%zu, %zu\n", sizeof(int64_t), alignof(int64_t)); return 0;
}
```

built with `g++ hello.cpp -std=c++11 -o hello`

ivan-mogilko:
Well, same on 32-bit Windows.

In theory padding also depends on how variables are following each other in the struct. Maybe there was a difference there.

But yes, it's best to fix the test to work with a precreated array. This problem may be curious to investigate on its own, but it simply is not a priority for this project.

BTW, there are no int64 values in the old formats, so they are never read like this in practice.

ericoporto:
Seems like it worked.

PR #1346, opened 2021-07-07
[DO NOT MERGE] Add Emscripten platform

ericoporto:
This is a draft pull request. I think it's getting close to have a possible WebGL platform for AGS that people could use.

But the code probably needs adjustments to be merged. It's alright, but I need help understanding what would need to be changed.

Here's the hightlights of things that could or not be problems

- Single thread: Emscripten support pthreads and JS has multithreading features, but having multiple threads today with current in place wasm solutions in Browsers require that the website serves two additional headers. Unfortunately, the usual hosts for this kinda of build doesn't serve these (itch.io, gamejolt, GitHub, ...). Currently this means supporting single thread audio in AGS (at compile time).
  - the pretty code for this would require (#1349).
- Alt+tab/out of focus: I had to hide the existing mechanism for this because it halted the webpage running the game, it does work now.
- It's a new platform and I need a directory to place the html file needed for it's generation, a build script and a markdown file that will describe how to build.
  - Should I add a toplevel directory named Emscripten or create a new dir like "Platforms" and put files in a directory inside this?

The CI for this is not here yet, I will make it soon.

A very small number of people already used experimental versions of the web port (not necessarily the last one) for their games, I think there's interest in it:
https://itch.io/games/made-with-adventure-game-studio/platform-web

Notes: 
- wasm doesn't have a real start and end of process, so we can't boot AGS again if it fails because variables won't be clean on the second time - asset library for example will fail a null assert.
- on MacOS, [Safari uses Angle](https://trac.webkit.org/wiki/AngleforWebGL), so the OpenGL renderer will run fine (iOS still has "real" WebGL)
- I will only build existing tests later, they will require nodejs to run the wasm and tests on the command line to verify things are built correctly. Right now this is a little tricky and would require building things twice, the second time for the tests and engine.
- ~~there's maybe something wrong on our Sound driver, according to OpenAL ref though it's all good~~
  - ~~runs fine with Android OpenAL or a regular desktop OpenAL~~
  - ~~MojoAL on the desktop had to be patched to avoid crashing https://github.com/adventuregamestudio/ags/commit/679fc7f516bdfa61cd015d4504117a1a22bf1957~~
  - ~~Emscripten OpenAL patch for this is https://github.com/ericoporto/emscripten/commit/5a7c723301222bc669412447ba13d7f09012be5e~~ THE PATCH HAS BEEN MERGED UPSTREAM!!! 😄🎈🎈

Edit:

Started working on the CI for this. It will take some time since I need to patch emsdk in a custom version and create a docker image - upstream Emscripten recent commits broke things and I can't build basic things anymore currently...


ericoporto:
Since the OpenAL patch has been merged upstream (https://github.com/emscripten-core/emscripten/pull/12664), building with the latest version of everything now at least works fine on the desktop! Need to retest on mobile though! :)

PR #1348, opened 2021-07-08
Tools: agspak, agsunpak - pack and unpack game packages (and templates)

### LINKED ISSUE #1336: Allow building game templates from the command line
Currently creating an AGS game template requires booting up AGS and clicking through menus. If it was possible to build/export game templates from command line we could automate our own templates builds in https://github.com/adventuregamestudio/ags-template-source.

This could be a switch on AGS Editor itself though since I think it would be _us_ the users of this functionality, no need to make it too complicate. Just enough to help with fixing template issues.

ivan-mogilko:
This provides a pair of tools for distributing with the Editor. Game packaging requires only agspak, but working with the game and room templates may require agsunpak as well.

Maybe can resolve #1336 (need to double check that everything is covered).

---

**agspak** - generates an ags game pak.

*Input:* directory with files.
*Output*: pack file (*.ags).
*Options:*
- `-p <MB>` - generate multiple parts, each limited by N megabytes.

Usage: `agspak dir_path output.ags -p 10`

**NOTES:**
- currently only uses the latest package format version. That might be enough for the IDE.
- packs every file from the given input directory. This means that necessary files may have to be copied somewhere first.
- with the size limit it cannot split the assets themselves, so if an asset file is larger than the limit it will still be added, but the package part will only contain that asset alone.

**TODO:**
- tidy/reorganize the code;
- support picking only particular files by command?
- support appending pak to another file (like *.exe);
- support recursive subdirectories (may be required for templates);
- support writing in specified format version? not a priority for us though.

---

**agsunpak** - extracts an ags game pak.

*Input:* pack file (*.ags).
*Output*: output directory.

Usage: `agsunpak game.ags out-dir`

**NOTES:**
- creates subdirectories if they are present in an asset name (some templates have these).

**TODO:**
- tidy/reorganize the code;
- double check that can unpack multiple lib parts.


ericoporto:
You removed `ComparePaths` which is used in tests. Also, non-related directly to this PR, but we should really start thinking about having Path class that can be serialized as CString instead of using string for path manipulation.

PR #1350, opened 2021-07-08
ci_find_file/unix: add fast-path for file lookup

rofl0r:
this achieves 2 major improvements:

- if the directory and filename are correct, only a single stat
  syscall needs to be done to check whether the file exists.
  no need to open the directory (1 syscall) and check whether
  each one of the entries matches the filename (1 syscall each).
  for perspective, each syscall requires to enter the kernel,
  a full contex switch, and therefore at least 1000 cpu cycles.

- if the filename contains a directory separator to dive into
  a subdir, the existing method fails.
  this allows to use e.g. PlayMP3File("sounds/sound25.ogg")
  successfully.

### DIFF
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -115,7 +115,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
     fix_filename_slashes(filename);
   }
 
-  if(directory && filename) {
+  if(directory && filename && !strstr(filename, "..")) {
     char buf[1024];
     snprintf(buf, sizeof buf, "%s/%s", directory, filename);
     lstat(buf, &statbuf);
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -115,6 +115,15 @@ char *ci_find_file(const char *dir_name, const char *file_name)
     fix_filename_slashes(filename);
   }
 
+  if(directory && filename) {
+    char buf[1024];
+    snprintf(buf, sizeof buf, "%s/%s", directory, filename);
+    lstat(buf, &statbuf);
+    if (S_ISREG(statbuf.st_mode) || S_ISLNK(statbuf.st_mode)) {
+      diamond = strdup(buf); goto out;
+    }
+  }
+
   if (directory == nullptr) {
     char  *match    = nullptr;
     int   match_len = 0;
@@ -174,6 +183,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
   fchdir(dirfd(prevdir));
   closedir(prevdir);
 
+out:;
   free(directory);
   free(filename);
 
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -115,6 +115,13 @@ char *ci_find_file(const char *dir_name, const char *file_name)
     fix_filename_slashes(filename);
   }
 
+  // the ".." check here prevents file system traversal -
+  // since only in this fast-path it's possible a potentially evil
+  // script could try to break out of the directories it's restricted
+  // to, whereas the latter chdir/opendir approach checks file by file
+  // in the directory. it's theoretically possible a valid filename
+  // could contain "..", but in that case it will just fallback to the
+  // slower method later on and succeed.
   if(directory && filename && !strstr(filename, "..")) {
     char buf[1024];
     snprintf(buf, sizeof buf, "%s/%s", directory, filename);
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -164,12 +164,12 @@ char *ci_find_file(const char *dir_name, const char *file_name)
 
   if (chdir(directory) == -1) {
     fprintf(stderr, "ci_find_file: cannot change to directory: %s\n", directory);
-    goto out;
+    goto out_pd;
   }
 
   if ((rough = opendir(directory)) == nullptr) {
     fprintf(stderr, "ci_find_file: cannot open directory: %s\n", directory);
-    goto out;
+    goto out_pd;
   }
 
   while ((entry = readdir(rough)) != nullptr) {
@@ -187,6 +187,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
   }
   closedir(rough);
 
+out_pd:;
   fchdir(dirfd(prevdir));
   closedir(prevdir);
 
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -85,7 +85,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
 }
 
 #else
-/* Case Insensitive File Find */
+/* Case Sensitive File Find - only used on UNIX platforms */
 char *ci_find_file(const char *dir_name, const char *file_name)
 {
   struct stat   statbuf;
@@ -131,7 +131,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
 
     match = get_filename(filename);
     if (match == nullptr)
-      return nullptr;
+      goto out;
 
     match_len = strlen(match);
     dir_len   = (match - filename);
@@ -152,17 +152,17 @@ char *ci_find_file(const char *dir_name, const char *file_name)
 
   if ((prevdir = opendir(".")) == nullptr) {
     fprintf(stderr, "ci_find_file: cannot open current working directory\n");
-    return nullptr;
+    goto out;
   }
 
   if (chdir(directory) == -1) {
     fprintf(stderr, "ci_find_file: cannot change to directory: %s\n", directory);
-    return nullptr;
+    goto out;
   }
-  
+
   if ((rough = opendir(directory)) == nullptr) {
     fprintf(stderr, "ci_find_file: cannot open directory: %s\n", directory);
-    return nullptr;
+    goto out;
   }
 
   while ((entry = readdir(rough)) != nullptr) {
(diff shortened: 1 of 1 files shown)


ericoporto:
Why using goto in a C++ codebase?

rofl0r:
> Why using goto in a C++ codebase?

because it's valid C++ ?

> If you can't refactor your code use the do while(false) pattern and break to keep the stack making sense.

the change you propose is much more intrusive, as it adds another indentation level for all code that follows, and makes the diff 50+ lines rather than 10.

ericoporto:
You can modify more lines, it will be code that builds. It will do the same in your architecture and not blow up my wasm port.

rofl0r:
> You can modify more lines, it will be code that builds. 

yeah, but it's not only intrusive, it also breaks git features like git blame.

> not blow up my wasm port. Now I need to PR a New version of this for wasm because there's no way to goto there.

emscripten/wasm don't support goto ? first time i hear this. do you have a link with more info about that?

ericoporto:
https://github.com/WebAssembly/design/issues/796

rofl0r:
> WebAssembly/design#796
>Everything that uses goto has to be refactored. This is one of the reasons I can't build everything from source or build old things from source that had goto. This is why I bring some stuff from Emscripten ports.

this seems to be a restriction of the wasm bytecode format itself, certainly emscripten itself can deal with goto as it is used in a lot of codebases, for example extensively even in musl libc which emscripten uses as its libc implementation.

rofl0r:
```
$ global -xg goto | wc -l
1278
```
current master branch has 1278 uses of goto in different sources, so if emscripten can't deal with it your effort is prone to fail anyway, this one goto doesn't change the overall situation.

rofl0r:
just found this comment by kripken (emscripten lead dev) confirming that emscripten can deal just fine with goto: https://github.com/WebAssembly/design/issues/796#issuecomment-245642856

ericoporto:
I strongly disenchourage using goto.

I also recommend using `dir_name == nullptr && file_name == nullptr` to be clear of intentions instead (like in the lines above) of the cryptic `directory && filename`.

ivan-mogilko:
I don't have much problem with this change if it keeps working everywhere for now.

> This entire file though will probably get axed soon since it has already other problems

Then it may be okay to put one fix even if it's temporary.

> I also recommend using dir_name == nullptr && file_name == nullptr to be clear of intentions instead (like in the lines above) of the cryptic directory && filename.

To be fair, we do not have this requirement for the code style. "directory && filename" is not cryptic, we use this pointer check in many many places in the code, even the newest code; I usually write in a same way for instance.

That said the rest of this particular file uses `==` explicit comparison.

> There's also a magic number 1024 there as size of a buffer... What does it limits? It's a mystery

Why is that a mystery? it's used in a sprintf one line below, and that's it.

ivan-mogilko:
By the way, since we started talking about `goto`, looking around this function, am I seeing things or there are potential memory leak cases?

The return statements on lines 155, 160, 165, they may leave local buffers `directory` and `filename` allocated but not freed.

rofl0r:
i found goto to be used in:
```
Common/libsrc/freetype-2.1.3/*
Compiler/script/cs_parser.cpp
Editor/scintilla/src/*
Engine/ac/dialog.cpp
Engine/ac/route_finder_impl_legacy.cpp
Engine/libsrc/apeg-1.2.1/audio/layer2.c
Engine/libsrc/apeg-1.2.1/audio/layer3.c
Engine/libsrc/apeg-1.2.1/audio/readers.c
Engine/libsrc/apeg-1.2.1/getbits.c
Engine/libsrc/apeg-1.2.1/getpic.c
Engine/libsrc/apeg-1.2.1/mpeg1dec.c
Engine/libsrc/apeg-1.2.1/ogg.c
Engine/platform/util/pe.c
Tools/data/dialogscriptconv.cpp
libsrc/allegro/src/blit.c
libsrc/allegro/src/file.c
libsrc/allegro/src/fli.c
libsrc/allegro/src/rotate.c
libsrc/allegro/src/unicode.c
```

> There's also a magic number 1024 there as size of a buffer... What does it limits? It's a mystery

it's big enough under most circumstances to hold a filepath, but not enough to cause issues with huge stack buffers (eventually overflowing thread stacks) and if it isn't the code will fail gracefully and use the non-fast-path implementation below.

> goto is dangerous, so I'd double check that there is nothing left deinitialized.

it's not really dangerous, but i already checked the code for these issues.

> the code is hardcoding / as file separator instead of using append_filename.

the code is only used on non-case-insentive filesystems, so only on platforms that have '/' as directory separator.

> You also forgot the AGS_PLATFORM_DEBUG check that is expected to print debug messages.

what?

> I strongly disenchourage using goto. There's no need at all for it, it's a simple refactor to not use it. If you can't refactor to use inline function, use the do while false break pattern.

once out of rational arguments, you refrain to "disencourage" use without arguments... alright.
the need for it is to not have code duplication and not to clutter unrelated code with gratuitous whitespace changes and breaking git blame functionality for all affected lines.

> I also recommend using dir_name == nullptr && file_name == nullptr to be clear of intentions instead (like in the lines above) of the cryptic directory && filename.

it's not cryptic, it's actually more clear than having the gratuitous comparisons which make the code longer for no reason to do the same and cause more strain and effort for the reader.

> The return statements on lines 155, 160, 165, they may leave local buffers directory and filename allocated but not freed.

i'll take a look and fix those in a followup commit, if indeed an issue.

ivan-mogilko:
> > You also forgot the AGS_PLATFORM_DEBUG check that is expected to print debug messages.
> 
> what?

I think @ericoporto may be referring to fprintf calls. Apprently only one of them was wrapped in AGS_PLATFORM_DEBUG 
check as seen [here](https://github.com/adventuregamestudio/ags/blob/8c088f61a3180d70591a98dcb36b189e5f6cb2e6/Common/util/misc.cpp#L172), and the rest are not. But this is how the code in our repository is, and this pr does not change anything there.

ivan-mogilko:
> i found goto to be used in:
> 
> ```
> Engine/ac/dialog.cpp
> Tools/data/dialogscriptconv.cpp
> ```

Not much important, but they do not have `goto` in real code, they just mention this dialog script command (and there's goto in an old commented code).

Common/util/misc.cpp
```diff
@@ -115,7 +115,7 @@ char *ci_find_file(const char *dir_name, const char *file_name)
     fix_filename_slashes(filename);
   }
 
-  if(directory && filename) {
+  if(directory && filename && !strstr(filename, "..")) {
```
ivan-mogilko, line 118: I don't think this is a correct check. It's possible to have a file name containing two dots.

rofl0r: yeah, but in that case we fallback to the chdir/opendir approach below

ivan-mogilko: Could you add a small comment above noting this?

rofl0r: sure!

ivan-mogilko: thank you )

ericoporto:
Case Sensitivity: No need to add the additional comment clause at right, it's false, MacOS is Unix and Case Insensitive by default: https://support.apple.com/guide/disk-utility/file-system-formats-available-in-disk-utility-dsku19ed921c/mac

Unfortunately this can't be fixed here in this PR, AGS should be able to workout Case Sensitivity at runtime.

Also this function used to print when it found a diamond in debug and now it doesn't.

rofl0r:
the comment says "only used on UNIX platforms", not "on ALL UNIX platforms"

ivan-mogilko:
> - using AGS own string type would make this code much cleaner and remove any need for goto at no performance impact at all.

This will likely require refactor whole function, and also adjust its uses, as the code that calls it expects a buffer that got to be freed.

ivan-mogilko:
@rofl0r I think your memory leaks fix is incomplete, there may be some things left unitialized.

There is a case where `prevdir` is storing handle from `opendir`, and there's also a call to `chdir` that is not getting reverted if you break execution in the middle.

ivan-mogilko:
About what you wrote in PR description:

> if the filename contains a directory separator to dive into
a subdir, the existing method fails.
this allows to use e.g. PlayMP3File("sounds/sound25.ogg")
successfully.

From what I understand, your added code only deals with the case-matching situation. Does this mean that if there's `PlayMP3File("SOUNDS/SOUND25.ogg");` instead - it will still fail?

rofl0r:
> There is a case where prevdir is storing handle from opendir, and there's also a call to chdir that is not getting reverted if you break execution in the middle.

good catch, it's fixed now

> the fact that it uses chdir is annoying, this makes this function harder to use in a multithreaded enviroment... if we ever rewrite it, chdir should not be used imho.

absolutely correct. using chdir is a real hack and codesmell.

> Does this mean that if there's PlayMP3File("SOUNDS/SOUND25.ogg"); (while the path & file are lowercase) instead - it will still fail?

yes, fastpath code (and that usecase) only works when the case is 100% correct.

ivan-mogilko:
@ericoporto I am okay with `goto` there mostly because it's a small library function and purely C-style with no C++ objects around. This pr is also a small fix, it does not change the function much (and it fixed few potential memory leaks now too).

But it definitely better be rewritten at some point in the future, it's not good for number of reasons. Some cases won't even work with it, not even after this fix (I mentioned one example above).

PR #1351, opened 2021-07-08
import error: print all missing imports for current script module

rofl0r:
rather than bailing out after the first.
this helps to evaluate missing script functions much quicker, than
with current approach. it's helpful when e.g. trying to implement
stubs for a proprietary binary plugin.

### DIFF
--- a/Engine/script/cc_instance.cpp
+++ b/Engine/script/cc_instance.cpp
@@ -1638,7 +1638,7 @@ bool ccInstance::ResolveScriptImports(PScript scri)
     }
 
     resolved_imports = new int[numimports];
-    int errors = 0;
+    int errors = 0, last_err_idx;
     for (int i = 0; i < scri->numimports; ++i) {
         if (scri->imports[i] == nullptr) {
             resolved_imports[i] = -1;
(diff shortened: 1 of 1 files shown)


rofl0r:
i actually wanted to use cc_warning for this pr, but no such function exists. i wasn't quite sure which debug log mode to use, so i'm happy for any better ideas and will amend PR accordingly.

ericoporto:
The plugin you are hacking around is open source. When I had to do what you claim was your problem with a proprietary plugin I did what everyone before me has done, used a hexeditor and extracted the stubs.

rofl0r:
> The plugin you are hacking around is open source. When I had to do what you claim was your problem with a proprietary plugin I did what everyone before me has done, used a hexeditor and extracted the stubs.

the plugin in question offers > 100 function of which only about 20-30 are used, so its much more efficient to implement only those that are used.

but that's not relevant to this PR, which simply improves error reporting by being more detailed.

ericoporto:
Is it possible to accumulate the things that didn't load? Do not change the code yet. I think on Windows the cc_error is an actual message box with the error, with the change that information is now gone.

rofl0r:
> I think on Windows the cc_error is an actual message box with the error, with the change that information is now gone.

you check by running a game that uses a plugin dll (e.g. strangeland.exe) after you removed or renamed the plugin.

ericoporto:
Before PR:
image

After PR:
image

I think it would be better if at least one error got stored in a string to make the error a bit more useful.

rofl0r:
a valid concern then. i pushed an update to address it.

ericoporto:
This looks good!

@ivan-mogilko after this is merged we should think about having a function to retrieve last logged error message - I don't know if it's stored somewhere, but this is not the first time I see something that could use one.

ivan-mogilko:
@ericoporto 
> @ivan-mogilko after this is merged we should think about having a function to retrieve last logged error message - I don't know if it's stored somewhere, but this is not the first time I see something that could use one.

Could you elaborate on this, do you mean the one logged with Debug::Printf, or something else?
I might mention, there's a separate HError struct, which is basically a thin wrapper over Error* object pointer (which is simply null in case of no error); I've been using it to return errors from a number of functions lately.

rofl0r:
> @rofl0r , I read the code and have no idea why it changed to ^0, do you have any guess?

it's displaying a different function, first was open, this is close

ivan-mogilko:
Ah. Should not it rather print the *first* unresolved import though, not the last?

rofl0r:
> Ah. Should not it rather print the first unresolved import though, not the last?

it would be slightly more code to implement and i figured last is just as helpful as first for this case

ivan-mogilko:
Alright.

PR #1353, opened 2021-07-08
Allow using OpenGL on MacOS

ericoporto:
- modify OGL Renderer to fail when shader creation fails so Software renderer is used instead.

I moved things around to account for shader failing to compile, which should not be a problem - it should not fail on real hardware but I do have a MacOS VM that doesn't have good OpenGL support and it does. 

I am not sure if the code changes are acceptable, it's mostly changing function signatures to pass the error forward. Please verify.

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -208,7 +208,7 @@ void OGLGraphicsDriver::SetTintMethod(TintMethod method)
   _legacyPixelShader = (method == TintReColourise);
 }
 
-void OGLGraphicsDriver::FirstTimeInit()
+bool OGLGraphicsDriver::FirstTimeInit()
 {
   String ogl_v_str;
 #ifdef GLAPI
@@ -222,12 +222,15 @@ void OGLGraphicsDriver::FirstTimeInit()
   OGLGraphicsDriver::InitSpriteBatch(0, _spriteBatchDesc[0]);
 
   TestRenderToTexture();
-  CreateShaders();
+  _shaders_created = CreateShaders();
   _firstTimeInit = true;
+  return _shaders_created;
 }
 
 bool OGLGraphicsDriver::InitGlScreen(const DisplayMode &mode)
 {
+  if(!_shaders_created) return false;
+
   SDL_Window* window = sys_get_window();
   if (window == nullptr || (SDL_GetWindowFlags(window) & SDL_WINDOW_OPENGL) == 0)
   {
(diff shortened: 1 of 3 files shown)


Engine/gfx/ali3dogl.h
```diff
@@ -255,6 +255,7 @@ class OGLGraphicsDriver : public VideoMemoryGraphicsDriver
     bool _smoothScaling;
     bool _legacyPixelShader;
 
+    bool _shaders_created {};
```
ivan-mogilko, line 258: What does this `{}` do in declaration? I noticed nick sonneveld was adding these too, but not for all the members.
Perhaps I'm missing something about the new C++ standard.

ericoporto: It's the default initialization, in this case means false. I can rewrite so it's written false instead.

ivan-mogilko: Tbh I am not well aware of how this works, in which circumstances will this value be applied?
The class already has a defined constructor, is this default initialization happens before the constructor?

ivan-mogilko: PS. that's not a problem for this PR, I was rather curious.

ericoporto: I found a bit of writing on this here: https://en.cppreference.com/w/cpp/language/value_initialization

If there's a value in the list initializer then it doesn't do narrowing conversion.

ivan-mogilko:
There's a logical mistake in this change, you check for `_shaders_created` in `InitGlScreen`, but that method is called before the shaders are created (in `FirstTimeInit`).

It does not look like `_shaders_created` is necessary, as the driver will fail and bail anyway if they are not.

Engine/gfx/ali3dogl.cpp
```diff
@@ -556,41 +561,47 @@ void main()
...
 
-void CreateTransparencyShader(ShaderProgram &prg)
+bool CreateTransparencyShader(ShaderProgram &prg)
 {
-  CreateShaderProgram(prg, "Transparency", default_vertex_shader_src, transparency_fragment_shader_src);
-
-  prg.MVPMatrix = glGetUniformLocation(prg.Program, "uMVPMatrix");
-  prg.TextureId = glGetUniformLocation(prg.Program, "textID");
-  prg.Alpha = glGetUniformLocation(prg.Program, "alpha");
+  if(CreateShaderProgram(prg, "Transparency", default_vertex_shader_src, transparency_fragment_shader_src)) {
```
ivan-mogilko, line 566: In regards to this, and other similar changes below: I really suggest to use the "early exit" rule, instead of putting multiple lines into an extra nesting.

That is, -
<pre>
if ( !condition ) return;
// do something on success
</pre>

That way you know that the function proceeds below only on success, and don't have to change nesting too much too.

ericoporto:
I am closing and reopening this PR because GitHub is not picking up my newer commits.

PR #1354, opened 2021-07-09
find file redone using AGS String and libraries.

ericoporto:


ivan-mogilko:
Ugh... it's not necesarily correct to just replace const chars with strings...

This function still uses SetCurrentDirectory, and your code returns the error that was fixed previously, when it may return from a function without changing current directory back.

ivan-mogilko:
@ericoporto if you are seriously up to rewriting this, please put the new function under a new name in one of the util/ modules, for example among the [File functions](https://github.com/adventuregamestudio/ags/blob/master/Common/util/file.h). It's best to do this in a first commit to make the removed code better visible in the git history.

ivan-mogilko:
Pushed similar code: 4e2df949ff36bbb6d08e402586dc8c2bb02ecad9

based on the latest rofl0r's fixes (without change cur dir), and have few simplifications. Most noteable is that Windows' part is reduced simply to "ConcatPaths", because there was nothing else there in a nutshell.

PR #1355, opened 2021-07-09
ci_find_file: fix omission to check lstat return value, improve fastpath

rofl0r:
if lstat() doesn't return success, it's undefined behaviour to access
the uninitialized memory of the stat buffer.

fast path was improved to kick in in almost 100% of cases, namely:
when directory is null and filename starts with a slash (absolute path)
use the absolute path for lookup, when directory is null and filename
doesn't start with a slash (relative path), use current directory.

### DIFF
--- a/Common/util/misc.cpp
+++ b/Common/util/misc.cpp
@@ -122,11 +122,15 @@ char *ci_find_file(const char *dir_name, const char *file_name)
   // in the directory. it's theoretically possible a valid filename
   // could contain "..", but in that case it will just fallback to the
   // slower method later on and succeed.
-  if(directory && filename && !strstr(filename, "..")) {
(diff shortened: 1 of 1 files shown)


rofl0r:
btw, in my opinion the code

```c
  if ((prevdir = opendir(".")) == nullptr) {
    fprintf(stderr, "ci_find_file: cannot open current working directory\n");
    goto out;
  }

  if (chdir(directory) == -1) {
    fprintf(stderr, "ci_find_file: cannot change to directory: %s\n", directory$
    goto out_pd;
  }
```
doesn't make any sense, as directory is then passed to opendir anyway, and there's nothing preventing opendir from working on either absolute or relative paths.. i'll remove the code locally for testing and report back whether its safe to remove it.

PR #1360, opened 2021-07-13
Allow using OpenGL on MacOS (fail and use Software if it can't)

ericoporto:
- modify OGL Renderer to fail when shader creation fails so Software renderer is used instead.
- modify shader version to be compatible (https://en.wikipedia.org/wiki/OpenGL_Shading_Language)
  - MacOS has at least 2.1 (version 120), but not OpenGL ES2 (version 100)

exactly the same as #1353

### DIFF
--- a/Common/core/platform.h
+++ b/Common/core/platform.h
@@ -142,7 +142,8 @@
                         AGS_PLATFORM_OS_IOS        || \
                         AGS_PLATFORM_OS_LINUX      || \
                         AGS_PLATFORM_OS_EMSCRIPTEN || \
-                        AGS_PLATFORM_OS_FREEBSD)
+                        AGS_PLATFORM_OS_FREEBSD    || \
+                        AGS_PLATFORM_OS_MACOS)
 #define AGS_OPENGL_ES2 (AGS_PLATFORM_OS_ANDROID || AGS_PLATFORM_OS_EMSCRIPTEN)
 
 // Only allow searching around for game data on desktop systems;
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -208,7 +208,7 @@ void OGLGraphicsDriver::SetTintMethod(TintMethod method)
   _legacyPixelShader = (method == TintReColourise);
 }
 
-void OGLGraphicsDriver::FirstTimeInit()
+bool OGLGraphicsDriver::FirstTimeInit()
 {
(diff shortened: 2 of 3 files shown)


ericoporto:
This moves CreateShaders from `FirstTimeInit` to `SetDisplayMode`. 

The idea if OpenGL shaders are correctly created, the renderer is used. If they fail for whatever reason, then it will fail the `SetDisplayMode` function which will propagate the failure to the callers. This will make AGS try other things until it tries the Software renderer, which works, and the game starts.

ericoporto:
I can't use the C++ preprocessor directives inside the here document (the multilines started with `r"EOS(`) so I placed the strings of the "header" that sets the version outside of it.

ivan-mogilko:
> This moves CreateShaders from FirstTimeInit to SetDisplayMode.

Could you elaborate on why this was necessary?
Before the shaders were created only once when OpenGL first initialized, now they will be recreated and destroyed when the display mode changes. Do shaders depend on the display mode settings, or there was some other reason for this?

ericoporto:
> > This moves CreateShaders from FirstTimeInit to SetDisplayMode.

> Could you elaborate on why this was necessary?

The only requirement for shader creation is it has to be called after either `gladLoadGL` or `gladLoadGLES2Loader`. 

I moved out of `FirstTimeInit` to try to solve the logical problem of shader creation (https://github.com/adventuregamestudio/ags/pull/1353#issuecomment-876912171). I can change back `FirstTimeInit` to bool and try to have it there. 

I don't know if I can create a shader in a context, have the context deleted and then create a new context and use the shader in this new context - in my mind it should be fine, and if it's fine, then it's fine to move back to FirstTimeInit with the signature change.

ivan-mogilko:
Maybe I am missing something, but what was the original problem there? It looked like it's fine in FirstTimeInit, because context is created *before* FirstTimeInit.

Unless there's a situation where it is not?

ericoporto:
I could not find such scenario, so I switched back to `FirsTimeInit()` and made so shader creation errors propagate. It seems to work!

ivan-mogilko:
I double checked the code, and *theoretically* this problem may occur if there's already an SDL window with SDL_WINDOW_OPENGL flag created before the driver is initialized.

This is only a theoretical possibility now, because in practice our renderers destroy their windows when they are deleted.
Maybe this problem could be adressed separately, for example, by creating new window only in FirstInit too.

PR #1367, opened 2021-07-17
Small fixes for building with CMake+MSVC on Windows

ericoporto:
Just some small corrections I picked building with CMake on Windows
- link shlwapi when building Common alone 
- hasNegateLiteral is being used without being initialized error pops up when trying to build Compiler Tests with CMake on Windows

Additionally, better always get SDL2 from GitHub on CMake builds, because it has better uptime than the SDL website.


ivan-mogilko:
> hasNegateLiteral is being used without being initialized

This is a very serious error, I wonder did it lead to some bugs, or never happened in practice due to how the function works...

PR #1370, opened 2021-07-19
Add compiler frontend 

ericoporto:
broke the compiler command line interface in a few parts - for testability.

Currently it looks like this: 
```
.\agscc.exe --help
agscc v0.1.0 - A Compiler for AGS Script
Copyright (c) 2021 AGS Team and contributors
Usage: agscc [options] <INPUT.asc>
-A <version>                 Script API Version               (default:Highest)
-C <version>                 Script API Compatibility version (default:Highest)
-H, --Headers <H1>[:<H2>...] Header Files in order  (; as separator in cmd.exe)
-D <macro>[=<val>]           Define <macro> to <val> (or 1 if <val> omitted)
-E                           Only run the preprocessor
-fshowwarnings[=0]           Print warnings to console              (default:1)
-fexportall[=0]              Exports all functions automatically
-flinenumbers[=0]            Include line numbers in compiled code
-fautoimport[=0]             Instancing exports funcs to other script
-fdebugrun[=0]               Log instructions as they are processed
-fnoimportoverride[=0]       Do not allow import to be re-declared
-fforceobjectbasedscript[=0] Enforce Object based scripting         (default:1)
-flefttoright[=0]            Left-to-right operator precedence      (default:1)
-fforcenewstrings[=0]        Enforce new strings                    (default:1)
-fforcenewaudio[=0]          Enforce new audio system               (default:1)
-foldcustomdialogopt[=0]     Use old custom dialog API
-g                           Generate debug information
--list-api-versions          Returns supported Script API Versions
-o <OUT.o>, --output <OUT.o> Place output in specified file.  (default:INPUT.o)
--override-version <VERSION> Overrides editor version
-h, --help                   Print this usage message
```


ericoporto:
I think it's ready for review!

Common/util/cmdlineopts.h
```diff
@@ -0,0 +1,50 @@
...
+    struct ParseResult {
+        bool HelpRequested = false;
+        std::vector<std::pair<String, String>> OptWithValue{};
+        std::vector<String> PosArgs;
+        std::multiset<String> Opt;
+    };
+
+    // opt_params_with_values are the options that always take an argument.
+    // example: `-l Val` or `--long Val` are passed with {"-l","--long"}.
+    ParseResult Parse(int argc, char *argv[], std::set<String> opt_params_with_values);
```
ivan-mogilko, line 43: Judging by the use, `opt_params_with_values` should be passed by reference to avoid making a copy, and be constants, as it is not supposed to change inside the function, so:
`const std::set<String> &opt_params_with_values`

Common/util/cmdlineopts.cpp
```diff
@@ -0,0 +1,91 @@
...
+        const bool is_opt_with_value = opt_params_with_values.find(arg) != opt_params_with_values.end();
+        if(!is_opt_with_value)
+        {
+            bool cont = false;
+            // check if it's written like `--optVal` instead of `--opt Val`
+            for(const auto& opt_param_with_value : opt_params_with_values)
+            {
+                if(arg.StartsWith(opt_param_with_value))
+                {
+                    String val = arg.GetCStr();
```
ivan-mogilko, line 66: I think GetCStr here is not necessary.

Common/util/cmdlineopts.cpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    for (int i = 0; i < args.size(); i++)
+    {
+        String arg = args[i];
+
+        if (arg[0] != '-')
+        {
+            parseResult.PosArgs.emplace_back(args[i]);
```
ivan-mogilko, line 53: Perhaps use `arg` here, as it is already created earlier.

Common/util/cmdlineopts.h
ericoporto: Ah, good observation! I didn't notice it, will change!

Common/util/cmdlineopts.cpp
ericoporto: It isn't, I will remove it. Thanks!

ericoporto: done

ericoporto:
Hey, I think I may have gotten this reversed but I am not sure (it's the compatibility level macros that is possible to set on the Editor)

https://github.com/ericoporto/ags/blob/add-compiler/Compiler/compiler.cpp#L111-L117

Compiler/script/cc_macrotable.h
```diff
@@ -9,9 +9,12 @@ struct MacroTable {
     char*macro[MAXDEFINES];
     void init();
     void shutdown();
-    int  find_name(char*);
-    void add(char*,char*);
+    int  find_name(const char*);
+    bool contains(const char *);
+    char* get_macro(const char *);
```
ivan-mogilko, line 14: I think "get_macro" should also return `const char*` if it's for reading. For editing, I would not recommend such function, because it returns a pointer to *some* buffer, but without telling its size nor letting to increase it.

ivan-mogilko: Also, tbh, the whole MacroTable struct could be replaced with a std::map<string, string> probably.

I may be mistaken here, but it looks like it's not used anywhere but the preprocessor, which you are rewriting from scratch anyway; if that's correct, then replacing with std::map won't be difficult to do.

Compiler/preproc/preprocessor.cpp
```diff
@@ -0,0 +1,429 @@
...
+    public:
+        explicit StringBuilder(size_t size) {
+            _str.Reserve(size);
+        }
+
+        void WriteLine(const String& str) override { _str.Append(str); _str.Append(li_end); }
+        bool IsValid() const override { return true; }
+        void WriteChar(char c) override {  _str.AppendChar(c); }
+        void WriteString(const String &str) override { _str.Append(str);}
+        void WriteFormat(const char *fmt, ...) override { va_list argptr; _str.Append(String::FromFormatV(fmt, argptr)); }
```
ivan-mogilko, line 36: This is not going to work, you declare `argptr` but not initialize it.

For the reference, this is how it's done: https://github.com/adventuregamestudio/ags/blob/008b2d2c61420f6f2c996317a0eb56a49e708095/Common/util/string.cpp#L650

Compiler/preproc/preprocessor.cpp
```diff
@@ -0,0 +1,429 @@
...
+                if (length == 0) return String("");
+                return _str.Mid(idx,length);
+            } else {
+                return ReadAll();
+            }
+        }
+        String ReadAll() override { size_t idx = _idx; _idx = _str.GetLength(); return _str.Mid(idx, _str.GetLength() - idx); }
+    };
+
+    static bool Contains(std::vector<String> v, const String &s) {
```
ivan-mogilko, line 72: Formally speaking, here the vector should be passed by a const reference `&` to avoid copying.

Also probably the loop may be replaced with `std::find(v.rbegin(), v.rend(), s) != v.rend();`

Compiler/preproc/preprocessor.cpp
```diff
@@ -0,0 +1,429 @@
...
+                    break;
+                case ErrorCode::LineTooLong:
+                    cc_error("Line too long (max line length = " STRINGIFY(MAX_LINE_LENGTH) ")");
+                    break;
+                case ErrorCode::EndIfWithoutIf:
+                    cc_error("#endif has no matching #if");
+                    break;
+                case ErrorCode::IfWithoutEndIf:
+                    cc_error("Missing #endif");
+                    break;
```
ivan-mogilko, line 94: I suggest adding a `default` case here, for asserting that a wrong error code is not passed in.

(Usually i just add "Unknown error" message to print something).

Compiler/preproc/preprocessor.cpp
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            output.WriteString(line.Mid(0, i));
+
+            if (i < line.GetLength())
+            {
+                bool precededByDot = false;
+                if (i > 0) precededByDot = (line[i - 1] == '.');
+
+                line = line.Mid(i, line.GetLength() - i);
```
ivan-mogilko, line 341: Not much important, just noticed this randomly, and wanted to note that you can write simply `line.Mid(i)` here. When the `count` argument is not passed the Mid will take everything to the end.

Hmm, also the whole `line = line.Mid(i);` may be done as `line.ClipLeft(i);`

Common/util/cmdlineopts.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace AGS
+{
+namespace Common
+{
+
+namespace CmdLineOpts
+{
+
+ParseResult Parse(int argc, char* argv[], const std::set<String> &opt_params_with_values)
```
ivan-mogilko, line 27: Ah, I did not notice this earlier, but the `argv` parameter might also be `const`, that would also help with tests where you are forced to cast `const char*` to `char*`.

This is a separate issue, but we also have this mistake in the engine code.

ericoporto: this one has to match the signature of our main, which doesn't have const.

ivan-mogilko: You may convert non const char* array to const char* array, see my "edit", I've tested that in the engine code.

Compiler/compiler.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+    if(comp_opts.PreprocessOnly)
+    {
+        Stream *out = File::CreateFile(comp_opts.OutputObjFile.c_str());
+        if (!out || !(out->CanWrite())) {
+            std::cerr << "Error: failed to open for writing: " << comp_opts.OutputObjFile << std::endl;
+            return -1;
+        }
+        script_pp.Write(out);
+        out->Close();
```
ivan-mogilko, line 217: `out` will not be deleted here. In practice this is not serious, because you are exiting right away, but in any other situation this will lead to a memory leak.

In the engine we usually have `delete out`, because stream calls Close in destructor, but engine's code is old and often not using modern C++. The right thing here would be to declare `std::unique_ptr<Stream>` instead, that would deal with both closing and deletion.

ivan-mogilko:
Ok, there's a lot of things that may be tested here. It works in a simple test, but I'd need to make myself some good test case to do this properly. Since we already have a tool that generates symbols for a game, maybe it will be possible to compile real game script.

Compiler/script/cc_macrotable.h
ericoporto: so, unordered_map gives me all sort of weird errors with the hashing function and ags String type - works fine with std::strings. Using map it works fine with AGS Strings, is map fine, or it HAS to be unordered map?

Compiler/preproc/preprocessor.cpp
ericoporto: thanks, I didn't knew that. Just fixed.

Compiler/script/cc_macrotable.h
ivan-mogilko: I forgot to mention, we have to provide a hashing function for a String class. It's in utils/string_types.h, simply including it should work.

As for selection between ordered and unordered, iirc ordered may be better when you constantly adding and removing items, and unordered is better when you add them once and then mostly keep looking up.

ericoporto:
Hey, I did the adjustments requested in a single commit to be easier to track changes, but I do plan to break up and squash the adjustments in the original commits when possible to reduce amount of commits in the history. I took the opportunity to delete cs_prepro because I don't think it's used.

Compiler/compiler.cpp
ericoporto: I did exactly it using unique_ptr https://github.com/adventuregamestudio/ags/pull/1370/commits/a922cee7f09b669c4b0073e75061cc14048ca579#diff-0cd44e2bb5fd3e84785ae6ad43582a662c32f9c080dae17bf0794315f2daf1eeL236-R241

Common/util/cmdlineopts.cpp
ericoporto: Thanks for the `const char *const argv[]`, I couldn't figure this! Did the change here:

https://github.com/adventuregamestudio/ags/pull/1370/commits/a922cee7f09b669c4b0073e75061cc14048ca579#diff-8b3b8298bbda3a3eee264e0d8994d6294883e3edc45b3b1981a24c8663cf504fL43-R43

Compiler/preproc/preprocessor.cpp
ericoporto: Added the default here: https://github.com/adventuregamestudio/ags/pull/1370/commits/a922cee7f09b669c4b0073e75061cc14048ca579#diff-32411dcb2b7765703f7e4bc2ce6ca1f198d06ede7d9c7e2bcb02612b79c29615R93-R102

But notice not all messages go through this switch currently, I wanted to, but could not figure how to put all the messages in one place and deal with the arguments some error message take currently.

Compiler/preproc/preprocessor.cpp
ericoporto: thanks, this made things really short!

https://github.com/adventuregamestudio/ags/pull/1370/commits/a922cee7f09b669c4b0073e75061cc14048ca579#diff-32411dcb2b7765703f7e4bc2ce6ca1f198d06ede7d9c7e2bcb02612b79c29615R73-R74

ericoporto:
put all previous reviews in place and changed commit history so that the fixes are in original commits.

ivan-mogilko:
The MSVS project still makes "Compiler.exe" instead of "agscc.exe".

That aside, I tested compiling full game script using [this project](https://github.com/ivan-mogilko/ags-camdemo), and looks like it is working.

<details>
<summary>Used following batch script on Windows</summary>
```
setlocal enabledelayedexpansion
set BIN=%1

mkdir temp
set WORK=temp

copy Game.agf %WORK%
copy agsdefns.sh %WORK%
copy *.ash %WORK%
copy *.asc %WORK%
copy *.crm %WORK%

cd %WORK%

rem Game Auto headers
%BIN%\agf2autoash.exe Game.agf _AutoGenerated.ash
%BIN%\agf2glvar.exe Game.agf _GlobalVariables.ash _GlobalVariables.asc
%BIN%\agf2dlgasc.exe Game.agf DialogScript.asc

rem Room Auto headers
for /r %%f in (room*.crm) do %BIN%\crm2ash.exe "%%f" "%%~nf.ash"

rem Prepare modules
set AUTO_HEADERS=agsdefns.sh:_AutoGenerated.ash
set MODULES=_GlobalVariables Utils CameraTricks DragDrop ObjectPool KeyboardMovement GlobalScript
set MODULE_HEADERS=

(for %%m in (%MODULES%) do (
	SET ASC=%%m.asc
	SET OUT=%%m.o
	SET MODULE_HEADERS=!MODULE_HEADERS!:%%m.ash
	%BIN%\agscc.exe !ASC! !OUT! -H %AUTO_HEADERS%!MODULE_HEADERS!
))

%BIN%\agscc.exe DialogScript.asc DialogScript.o -H %AUTO_HEADERS%!MODULE_HEADERS!

(for /r %%f in (room*.asc) do (
	SET ASC=%%~nf.asc
	SET ASH=%%~nf.ash
	SET OUT=%%~nf.o
	%BIN%\agscc.exe !ASC! !OUT! -H %AUTO_HEADERS%!MODULE_HEADERS!:!ASH!
))
```
</details>

I've been able to test above without repacking whole game now thanks to #1330.

As you may notice, the only step which is left not automated is the ordered list of script modules (and active rooms actually).

I'm thinking of writing a very simple tool for exporting miscellaneous stuff from Game.agf, which would write such lists (that may come handy for other purposes too).

ericoporto:
Hey, pushed the rename of the target on vcxproj.

I have been thinking of modifying the `comments` column in the [wiki here](https://github.com/adventuregamestudio/ags/wiki/AGS-Game-Build-process-(3.5.*)) to put the tools we have to see what's missing.

ivan-mogilko:
Regarding the exe name, I don't think it should be different depending on configuration, this is usually how libraries are named, but not executables.

ericoporto:
I switched the name to be the same in all configs when using vcxproj files

ivan-mogilko:
I think we may merge this.
* The tool is formally working;
* It's an alternative but not a part of a Editor workflow yet, so even with mistakes it won't break anything;
* Fully testing this will require a lot of time, which is better to do in steps, perhaps involving volunteer testers from AGS community.

@ericoporto could you squash last two commits please?

ericoporto:
Squashed the commits!

PR #1372, opened 2021-07-24
3.6.0: Support reading from game assets in script + Editor supports packing custom files

### LINKED ISSUE #1227: Allow packaging arbitrary files

ivan-mogilko:
Fully resolves #1227.

This PR does two things:
1. Adds support for reading resource files ("assets") in script. This is a backport of #1294 for ags3. I recommend reading 1294's description to understand how this works.
2. In the Editor adds new property to General Settings called "Package custom data folder". If this property is filled, Editor will gather all files from the corresponding subfolder (recursively, including all files in all nested subfolders), and package these files into the game along with all the standard files.

Regarding p2:
* Resources packaged like this are assigned names containing full relative path. For example: `Data/subdir/myfile.dat`. The topmost dir name is also kept - to easily avoid filename conflicts with the rest of the game files.
* To be precise, the path slashes are forced to strictly forward style. This may be important when opening these files in script; although we might adjust engine's behavior to support both ways when opening resources from a package (frankly, I don't remember if it already does that).
* These files may be opened in script using a new filepath token "$DATA$" (as described in #1294), for example:
```
File* f = File.Open("$DATA$/Data/subdir/myfile.dat", eFileRead);
```
* To clarify once more, just in case, $DATA$ token only allows to open files for reading.

### DIFF
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
@@ -33,7 +33,10 @@ namespace AGS.Editor
             }
         }
 
-        private string[] ConstructFileListForDataFile()
+        /// Creates a list of game resources as a list of tuples:
+        /// - first tuple's element is resource's name,
+        /// - second is real file path
+        private Tuple<string, string>[] ConstructFileListForDataFile()
         {
             List<string> files = new List<string>();
             Environment.CurrentDirectory = Factory.AGSEditor.CurrentGame.DirectoryPath;
@@ -59,7 +62,14 @@ namespace AGS.Editor
                 }
             }
             Utilities.AddAllMatchingFiles(files, "*.ogv");
-            return files.ToArray();
+
+            // Regular files are registered under their filenames (w/o dir)
+            var assetList = new List<Tuple<string, string>>();
+            foreach (var f in files)
+            {
+                assetList.Add(new Tuple<string, string>(Path.GetFileName(f), f));
+            }
+            return assetList.ToArray();
         }
 
         private void CreateAudioVOXFile(bool forceRebuild)
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
@@ -63,11 +63,23 @@ namespace AGS.Editor
             }
             Utilities.AddAllMatchingFiles(files, "*.ogv");
 
+            List<string> userFiles = new List<string>();
+            string user_dir = Factory.AGSEditor.CurrentGame.Settings.CustomDataDir;
+            if (!string.IsNullOrWhiteSpace(user_dir))
+            {
+                Utilities.AddAllMatchingFiles(userFiles, user_dir, "*", true, SearchOption.AllDirectories);
+            }
+
             // Regular files are registered under their filenames (w/o dir)
             var assetList = new List<Tuple<string, string>>();
             foreach (var f in files)
             {
                 assetList.Add(new Tuple<string, string>(Path.GetFileName(f), f));
+            }
+            // User files are registered under their relative paths
+            foreach (var f in userFiles)
+            {
+                assetList.Add(new Tuple<string, string>(f, f));
             }
             return assetList.ToArray();
         }
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
@@ -33,10 +33,12 @@ namespace AGS.Editor
             }
         }
 
+        /// <summary>
         /// Creates a list of game resources as a list of tuples:
         /// - first tuple's element is resource's name,
         /// - second is real file path
-        private Tuple<string, string>[] ConstructFileListForDataFile()
+        /// </summary>
+        private Tuple<string, string>[] ConstructFileListForDataFile(CompileMessages errors)
         {
             List<string> files = new List<string>();
             Environment.CurrentDirectory = Factory.AGSEditor.CurrentGame.DirectoryPath;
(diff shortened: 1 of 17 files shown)


ivan-mogilko:
Hmm, maybe we could also allow to input a comma separated list of folders in that field?

ericoporto:
I tested and verified it works. 

I thought it would look nicer if it was a node in the Explore Project menu, which usually points out the resources, but the sheer amount of work to create a new Component in AGSEditor/Components and a new type in AGS Types just to store this string made me realize I prefer like this right now and maybe one day on the future it's worth making it it's own entry there if this is something more people are interested.

ivan-mogilko:
I forgot to mention, as per last commit this setting now 
* allows to input multiple directories, separated by commas;
* tests that directories are relative to game project, and not one of the standard folders (Compiled, AudioCache etc), to avoid including wrong stuff.
* if same directory is entered multiple times, or user input sudirectories of other custom directories, these duplicates will be ignored.

ivan-mogilko:
@persn if you have a bit of spare time, could you check the changes I made to the Editor here? There are not big, but i wonder if there's something that could've been done easier in C#.

persn:
I can check on Sunday at earliest

Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
```diff
@@ -59,7 +64,78 @@ private string[] ConstructFileListForDataFile()
                 }
             }
             Utilities.AddAllMatchingFiles(files, "*.ogv");
-            return files.ToArray();
+
+            // Regular files are registered under their filenames (w/o dir)
```
persn, line 68: `assetList` gets iterated over twice, once in foreach and second in `ToArray()`. It's kind of wasteful so maybe try to avoid it with lazy executing code, there's 2 ways to do that, Linq or yield keyword.

Additionally it might be a good idea to change `ConstructCustomFileListForDataFile(...)` to return empty collection on failure instead of `null` because then you can concat it

Linq
```csharp
return files
    .Select(f => new Tuple<string, string>(Path.GetFileName(f), f));
    .Concat(ConstructCustomFileListForDataFile(errors))
    .ToArray();
```

yield, in order for this to work the return value needs to be changed to `IEnumerable<Tuple<string, string>>`
```csharp
foreach (var f in files)
    yield return new Tuple<string, string>(Path.GetFileName(f), f);

foreach (var f in ConstructCustomFileListForDataFile(errors))
    yield return f;
```

Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
```diff
@@ -59,7 +64,78 @@ private string[] ConstructFileListForDataFile()
...
+                }
+                if (Utilities.PathIsSameOrNestedAmong(testDir, usedCustomDirs.ToArray()))
+                    continue; // don't report, simply skip
+
+                Utilities.AddAllMatchingFiles(userFiles, dir, "*", true, SearchOption.AllDirectories);
+                usedCustomDirs.Add(Path.Combine(curDir, dir));
+            }
+
+            // User files are registered under their relative paths
+            var assetList = new List<Tuple<string, string>>();
```
persn, line 133: Can condense this a little

```csharp
return userFiles.Select(f => new Tuple<string, string>(f, f)).ToArray();
```

Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
```diff
@@ -59,7 +64,78 @@ private string[] ConstructFileListForDataFile()
...
+
+        /// <summary>
+        /// Creates a list of custom game resources as a list of tuples:
+        /// - first tuple's element is resource's name,
+        /// - second is real file path
+        /// </summary>
+        private Tuple<string, string>[] ConstructCustomFileListForDataFile(CompileMessages errors)
+        {
+            string userOpt = Factory.AGSEditor.CurrentGame.Settings.CustomDataDir;
+            if (string.IsNullOrEmpty(userOpt)) return null;
```
persn, line 88:
```suggestion
            if (string.IsNullOrEmpty(userOpt)) return Enumerable.Empty<Tuple<string, string>>();
```

Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
```diff
@@ -59,7 +64,78 @@ private string[] ConstructFileListForDataFile()
...
+
+            string curDir = Factory.AGSEditor.CurrentGame.DirectoryPath;
+            string[] restrictedDirs = new string[Factory.AGSEditor.RestrictedGameDirectories.Length];
```
persn, line 91:
```
string[] restrictedDirs = Factory.AGSEditor.RestrictedGameDirectories.Select(d => Path.Combine(curCir, d)).ToArray();
```

Editor/AGS.Editor/BuildTargets/BuildTargetDataFile.cs
```diff
@@ -59,7 +64,78 @@ private string[] ConstructFileListForDataFile()
...
+                // don't allow paths that match or inside restricted folders,
+                // or folders that were already used to make a list of assets
+                string testDir = Path.Combine(curDir, dir);
+                if (Utilities.PathIsSameOrNestedAmong(testDir, restrictedDirs))
+                {
+                    errors.Add(new CompileWarning(
+                        string.Format("Cannot use restricted location as a custom data source: {0}", dir)));
+                    continue;
+                }
+                if (Utilities.PathIsSameOrNestedAmong(testDir, usedCustomDirs.ToArray()))
```
persn, line 125: If `Utilities.PathIsSameOrNestedAmong` input is changed so it accepts IEnumerable<string> as second parameter you can pass any collection (array, list, queue, etc.) without having to reallocate them to another data structure.

persn (approved):
I was able to carve out some time. I'm marking this as approved because good enough for production. I have some suggestions using linq to make the code more readable and/or efficient. However they're all just suggestions and if you don't want to include them for any reason it's not a big deal.

I may have more ideas but the review window is now so full of my comments it's hard to get a good overview, I could take a second look at it later

ivan-mogilko:
@persn Amended with almost all of your suggestions.

One that I could not change was about changing AddAllMatchingFiles's IList argument to List, because there are already IList variables passed into it on some occasions.

PR #1373, opened 2021-07-26
CMake: allow building Tools from CMake

ericoporto:
This add a new toggle to build Tools in CMake - which when on also builds Compiler related stuff.

The toggle is OFF by default, and I am not turning the CI on this yet (we don't have a plan for actually shipping the Tools yet, so no need to turn on now). 

Btw, some Makefiles for tools are broken, `multifilelib.cpp` actually has a typo and is `mutifilelib.cpp` (missing first `l` character), I wasn't sure if I should fix with `git mv` the file name or change the Makefiles to carry the typo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,8 @@ project(AGS
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/CMake")
 
 option(AGS_TESTS "Build tests" OFF)
-option(AGS_BUILD_COMPILER "Build compiler library" OFF)
+option(AGS_BUILD_TOOLS "Build Tools" OFF)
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
> Btw, some Makefiles for tools are broken, multifilelib.cpp actually has a typo and is mutifilelib.cpp (missing first l character), I wasn't sure if I should fix with git mv the file name or change the Makefiles to carry the typo.

Never noticed that...
I pushed the `git mv` change, along with corrections to all the project files that had it referenced.

ivan-mogilko:
I think Tools/Cmakelist.txt in this pr now has to be fixed, because it still references wrong name.

ericoporto:
I fixed the `Tools/Cmakelist.txt` and picked two small things when trying to build them on GCC in my Ubuntu 20.04 machine.

PR #1375, opened 2021-07-28
Tool: agfexport - export miscellaneous data from the game source

ivan-mogilko:
agfexport - parses the game project file (*.agf), and exports various stuff depending on command.
This tool is meant to group simple tasks which do not require complex generation of anything, only outputting data for use along with the other tools and processes.

*Input:* game project (*.agf).
*Output:* a text file.

Usage: `agfexport Game.agf <COMMAND> [<OPTIONS>] somefile.txt`
Currently supported commands are:
* `--script-list` - exports ordered list of script modules (in the order of their dependency, or headers inclusion)
* `--room-list` - exports list of active rooms in the form of `roomNNN.crm:DESCRIPTION`.

These provide last remaining information necessary to write a system script that compiles all game scripts from the source without specifying script files by hand.

More could be added later if necessary.

For example of potential extensions, this tool could be exporting dialog scripts and additional info needed for the dialog converter to process them into ags script, then we would be able to replace current *agf2dlgasc* tool with one that does conversion using these files and not reading AGF document directly.

### DIFF
--- a/Common/util/string.cpp
+++ b/Common/util/string.cpp
@@ -522,6 +522,27 @@ void String::AppendChar(char c)
     }
 }
 
+void String::AppendFmt(const char *fcstr, ...)
+{
+    va_list argptr;
+    va_start(argptr, fcstr);
+    AppendFmtv(fcstr, argptr);
+    va_end(argptr);
+}
+
+void String::AppendFmtv(const char *fcstr, va_list argptr)
+{
+    fcstr = fcstr ? fcstr : "";
+    va_list argptr_cpy;
+    va_copy(argptr_cpy, argptr);
+    size_t length = vsnprintf(nullptr, 0u, fcstr, argptr);
+    ReserveAndShift(false, length);
+    vsprintf(_cstr + _len, fcstr, argptr_cpy);
+    va_end(argptr_cpy);
+    _len += length;
+    _cstr[_len] = 0;
+}
+
 void String::ClipLeft(size_t count)
 {
     if ((_len != 0) && (count > 0))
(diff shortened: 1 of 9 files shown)


ericoporto:
I don't if it should be here or not, but script API and compat level is something I need for input in the compiler

ivan-mogilko:
Oh, right, it may output a list of compilation flags read from the game settings.

ivan-mogilko:
I have a bad thought that maybe I made a mistake with the tools. Instead of having separate tools for generating a game header, and another for global variables script, we could perhaps have one tool for exporting everything necessary for a compilation from Game.agf; basically, preparing all intermediate data necessary for compilers at once.

I probably should not be making any changes now to not make any hasty decisions nor waste time on this, but maybe this could be revised later, when we have a full toolchain ready and tested.

Anyway, in regards to agfexport, this should do at least 2 things:
1. Export info to help compile the game script;
2. Export info to help package the game (because Game.agf contains some directives about that too).

> Edit: also kinda feel this is the one tool that would be useful to be able to output to stdout too, to be able to fill variables in the command line with the outputs of it

I will make file output at first to see what form is convenient, then it will be more clear whether output to stdout is necessary.
With files it's more direct, even though you have to have a "cleanup" step to wipe these out.

ericoporto:
So I got a spurious notification from this PR from I think a bug in GitHub, so I started looking into what was that happened with the build tools

> Instead of having separate tools for generating a game header, and another for global variables script, we could perhaps have one tool for exporting everything necessary for a compilation from Game.agf

This seems sensible, like, having a single tool responsible to understand the Game.agf format and have it be able to generate that can be fed into other tools. This reduces the surface for testing things.

ivan-mogilko:
The commits in this branch seem to contain some useful generic code, I may push it to the master separately.

PR #1378, opened 2021-08-01
Tool: crmpak - compiled room's (re)packer

ivan-mogilko:
After #1330 we may store compiled scripts separate from the room files (*.crm), but then I realized that there will still be rooms with these scripts inside remaining in the ags3 game project. If we, for instance, begin to package scripts separately in the future, there will be redundant script data left inside these rooms.

So here comes the tool that deals with that, but I decided to make it more universal just in case.
(The blocks are searched all the same, so there's no difference)

---

crmpak - imports, exports or deletes chosen blocks from the compiled room file (*.crm).

Usage: `crmpak [OPTIONS] [<in-room.crm> <COMMAND> [<CMD_OPTIONS>]]`

Options:
* `  --tell-blockids` - print a list of the known block ids;
* 
Commands:
* `-d <blockid>` -  delete: remove a block from the compiled room;
* `-e <blockid> <file>` - export: write a block into this file;
* `-i <blockid> <file>` - import/inject: add/replace a block with this file contents;
* `-l` - list: print id of all blocks found in the room;
* `-x <blockid> <file>` - extract: remove a block and save it in this file (export + delete);

Command Options:
* `-w <out-room.crm>` - for all commands but '-e': write the resulting room into a new file; otherwise will modify the input file;

Block IDs may be given both [as numbers](https://github.com/adventuregamestudio/ags/blob/4d8fa9fe6cbe63f6770777f3853e221117c94744/Common/game/room_file.h#L55) or [their string aliases](https://github.com/adventuregamestudio/ags/blob/4d8fa9fe6cbe63f6770777f3853e221117c94744/Common/game/room_file_base.cpp#L83).

**Examples:**

`crmpak room1.crm -x 7 room1.o`
`crmpak room1.crm -x CompScript3 room1.o`

These two commands are equivalent, they extract a compiled script from the `room1.crm`, save it into `room1.o`, and then rewrite the room without it.

`crmpak room1.crm -d CompScript3`

Deletes the script from the room. NOTE: it's interesting that the engine won't complain of its abscence.

`crmpak room1.crm -i CompScript3 newscript.o -w newroom.crm`

Writes a new room as `newroom.crm`; it will have all content from `room1.crm`, but the script will be added/replaced from `newscript.o`.

### DIFF
--- a/Common/game/room_file.h
+++ b/Common/game/room_file.h
@@ -116,6 +116,10 @@ HRoomFileError WriteRoomData(const RoomStruct *room, Stream *out, RoomFileVersio
 // Reads room data header using stream assigned to RoomDataSource;
 // tests and saves its format index if successful
 HRoomFileError ReadRoomHeader(RoomDataSource &src);
+// Writes room data header
+void WriteRoomHeader(Stream *out, RoomFileVersion data_ver);
+// Writes a room data ending
+void WriteRoomEnding(Stream *out, RoomFileVersion data_ver);
 
 // Type of function that writes single room block.
 typedef std::function<void(const RoomStruct *room, Stream *out)> PfnWriteRoomBlock;
--- a/Common/game/room_file.h
+++ b/Common/game/room_file.h
@@ -73,7 +73,9 @@ enum RoomFileBlock
     // Script names of the room objects
     kRoomFblk_ObjectScNames = 9,
     // End of room data tag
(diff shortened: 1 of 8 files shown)


ericoporto:
Hey, two ideas:
- telling all possible block IDs
- telling the available block IDs in current room.crm file

PR #1380, opened 2021-08-02
Script API: AudioClip.PlayOnChannel()

ivan-mogilko:
This adds AudioClip.PlayOnChannel() function:

`AudioChannel* AudioClip.PlayOnChannel(int chan, AudioPriority=SCR_NO_VALUE, RepeatStyle=SCR_NO_VALUE);`

The purpose of this function is to provide a lower-level control over how the clips are put onto the logical "channel" slots. Normally this distribution is controlled by a number of settings involving audio type reservation and clip priority, but there are circumstances when users want to have their own logic, yet there's no direct way to apply that. With this function it might be possible to script your own audio mixer freely.

The necessary changes for this are minimal, as this script function simply calls an existing internal one that does the necessary thing, circumventing channel calculation.

Notable issues: channel 0 is reserved for voice-over. Technically it may be used, but there's some engine logic that stops any other sound on it, probably because no actual speech is on.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2087,6 +2087,10 @@ builtin managed struct AudioClip {
   import AudioChannel* PlayFrom(int position, AudioPriority=SCR_NO_VALUE, RepeatStyle=SCR_NO_VALUE);
   /// Plays this audio clip, or queues it if all channels are busy.
   import AudioChannel* PlayQueued(AudioPriority=SCR_NO_VALUE, RepeatStyle=SCR_NO_VALUE);
+#ifdef SCRIPT_API_v360
+  /// Plays this audio clip, explicitly putting it on the particular channel.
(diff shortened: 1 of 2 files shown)


fernewelten:
> there are circumstances when users want to have their own logic, yet there's no direct way to apply that.

This is especially interesting for PlayQueued(). If you want to line up clips to be queued, you want to be able to specify __what__ queue exactly. 

For instance, suppose you have an animated spaceship warp. The engines start running and at a certain variable point, the spaceship fizzes away.

So you set up a queue: A "ramp  up" sound, then in a loop a "running engine" sound. 
Start the queue playing, at the same time start the animation playing.
When the variable point has come where the engine fizzes away, you interrupt and stop the sound queue (it will be busy playing the engine noise in a loop) and then play the fizz away noise. 

You can do this with music because there usually is just 1 music channel. So all music will automatically queue up one behind the other. 

But you can't do this with sound which usually has more than 1 channel: It's impossible to specify that the "engine running" noise must specifically queue up behind the "ramp up" noise, even if an idle channel is available.

PR #1381, opened 2021-08-02
New compiler: Join juxtaposed string literals

fernewelten:
Make the compiler join string _literals_ (!) that are separated by whitespace, if any. This is an operation that happens at compile time, not at runtime.

The same way it is done in `C` or `C++`.

Typical code:
```
player.Say(
    "The rain it raineth on the just ""And also on the unjust fella; "
    "But chiefly on the just because "     "The unjust hath the just's umbrella.");
```

Generates the same code as
```
player.Say(
    "The rain it raineth on the just And also on the unjust fella; But chiefly on the just because The unjust hath the just's umbrella.");
```

### DIFF
--- a/Compiler/script2/cs_scanner.cpp
+++ b/Compiler/script2/cs_scanner.cpp
@@ -9,7 +9,9 @@
 
 #include "cs_scanner.h"
 
+//                                                      123456789a1234567
 std::string const AGS::Scanner::kNewSectionLitPrefix = "__NEWSCRIPTSTART_";
+size_t const AGS::Scanner::kNewSectionLitPrefixSize = 17u;
 
 AGS::Scanner::Scanner(std::string const &input, SrcList &token_list, ccCompiledScript &string_collector, SymbolTable &symt, MessageHandler &messageHandler)
     : _ocMatcher(*this)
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
This change seem to have broken error reporting: whenever there's an error in script, the location looks like this:

> NewScript1.ash__NEWSCRIPTSTART_NewScript.ash__NEWSCRIPTSTART_compiler.ash__NEWSCRIPTSTART_GlobalScript.ash__NEWSCRIPTSTART_Chains.asc(28): Unexpected symbol 'asdasdasd'

ericoporto:
You can run the command line compiler and tell it to run only the preprocessor, and you will get the resulting file.

In the original preprocessor this string marker is [here](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.CScript.Compiler/Preprocessor.cs#L43) : 
```			output.AppendLine(Constants.NEW_SCRIPT_MARKER + scriptName + "\"");
```

I have a guess that switching the last part of the string to `"\";"` might fix this but need to test.

fernewelten:
@ivan-mogilko Your hunch is probably correct; I'll have a look and fix that.

By the way, do you have any idea why we do section changes in such a quirky way? Users that will code `player.Say("__NEWSCRIPTSTART_")` will get an error message and wonder what is happening.

As far as I know, the __traditional__ way of doing this was to leave a "#line" directive in the code after preprocessing, cf. for instance https://docs.microsoft.com/en-us/cpp/preprocessor/hash-line-directive-c-cpp . That is, instead of putting 
```
"__NEWSCRIPTSTART_GlobalScript.asc"
``` 
at the start of a line, there would be 
```
#line 1 "GlobalScript.asc"
``` 
at the start of the line.  

The advantage of that traditional directive is that preprocessors can sometimes add extra lines into the program, and if they do, they can use that same `#line` directive to synchronise the line numbering of the source code afterwards, by emitting, e.g. `#line 417 "GlobalScript.asc"` 

What was the rationale for inventing something different in AGS? Do you happen to know?

rofl0r:
> What was the rationale for inventing something different in AGS? Do you happen to know?

probably cj wasn't aware of it, and there's SCMD_LINENUM for debugging which has only a single numeric value. the NEWSCRIPTSTART thing is to produce a new script "section" which has filename and start/end offset in script bytecode so the debugger can map the SCMD_LINENUM to a filename depending from where it was called.

ivan-mogilko:
No, I don't know, I was not around when all this happened, it could be what rolf0r said.

We may switch to a more convenient form for the new compiler of course.

Compiler/script2/cs_scanner.cpp
```diff
@@ -544,14 +545,34 @@ AGS::ErrorType AGS::Scanner::ReadInStringLit(std::string &symstring, std::string
...
-                return kERR_None; // End of string
+                return kERR_None;
             }
 
             // Another string literal follows
+            // Tentatively read ahead to check whether it's a new section marker
+            pos_before_skip = _inputStream.tellg();
+            lineno_before_skip = _lineno;
+            Get(); // Eat leading '"'
+            char *tbuffer = new char[kNewSectionLitPrefix.length() + 1];
```
ivan-mogilko, line 562: Randomly noticed this, I think allocating memory could be avoided since kNewSectionLitPrefix is a constant string.
Even if you cannot use its "length()" there because its not `constexpr` (and you don't want to make this string a literal), you still could use a buffer of large enough fixed size to accomodate it.

ivan-mogilko, line 562: I think you may avoid repeatedly allocating extra buffer here, since kNewSectionLitPrefix is constant, having a large enough buffer allocated on stack could be enough.

ivan-mogilko:
Concatenation is working, and do not notice any problem during execution now.

Compiler/script2/cs_scanner.cpp
fernewelten: It's not a code I'm the most proud of. My trouble was with the `std::istring::get()` call: it takes a `char *` in the first argument, and I didn't find a variant that takes a `std::string` instead. Bummer. So at first I tried to force a string into the function call:
```
std::string buffer;
buffer.reserve(…); // so that its internal buffer is long enough
… get(&(buffer[0]), …) …
```
but after that sequence, the content __looked__ okay but `buffer.length()` reported the wrong length. Well okay, it's probably a dirty coding to begin with, so let's behave and not do that, Peter.  🙄

So that's when I allocated them their d*mn `char[]`, moved the result into a std::string and freed the `char[]` directly afterwards. 

Okay, after a night's sleep, I'll try to get that more elegant.
Okay, I'll come up with something more elegant.

fernewelten:
(Squashing for better clarity.)

PR #1385, opened 2021-08-05
Windows: support UTF-8 paths in the engine

ivan-mogilko:
After #1321 unix ports should support utf-8 paths naturally, but on Windows standard library functions (as well as WinAPI) accepting `const char*` paths do not handle utf-8, which means we have to use wide-string api for unicode paths.

This pr changes all the file related functions to convert our strings to wide-char paths and use corresponding wide-string api on Windows.

Supposedly, after this change -
* DOS paths (aka "short paths") are no longer used anywhere in the engine;
* Windows engine should correctly interpret unicode paths passed into command line (well, it probably did before, but just to be sure...);
* Hopefully utf-8 paths in config file and winsetup will also work.

### DIFF
--- a/Common/util/directory.cpp
+++ b/Common/util/directory.cpp
@@ -27,11 +27,13 @@ namespace Directory
 
 bool CreateDirectory(const String &path)
 {
-    return mkdir(path.GetCStr()
-#if ! AGS_PLATFORM_OS_WINDOWS
-        , 0755
+#if AGS_PLATFORM_OS_WINDOWS
+    WCHAR wstr[MAX_PATH_SZ];
+    MultiByteToWideChar(CP_UTF8, 0, path.GetCStr(), -1, wstr, MAX_PATH_SZ);
+    return (CreateDirectoryW(wstr, NULL) != FALSE) || (GetLastError() == ERROR_ALREADY_EXISTS);
+#else
+    return (mkdir(path.GetCStr(), 0755) == 0) || (errno == EEXIST);
 #endif
-        ) == 0 || errno == EEXIST;
 }
 
 bool CreateAllDirectories(const String &parent, const String &sub_dirs)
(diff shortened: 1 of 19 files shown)


ericoporto:
On the Editor, when writing the default config, it uses `Encoding.Default`:

https://github.com/adventuregamestudio/ags/blob/b6c0a796251bea51a6655782d2cefbf12d3d330d/Editor/AGS.Editor/AGSEditor.cs#L1600

According to MS Docs below, it's better we explicitly set the encoding to utf8.

https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=net-5.0#System_Text_Encoding_Default

ivan-mogilko:
> According to MS Docs below, it's better we explicitly set the encoding to utf8.

Of course, the Editor will have to be adjusted to write utf-8.

ivan-mogilko:
@ericoporto Editor writes all the project and game files in ansi codepage right now, this is a separate problem, also noted in the [unicode ticket](#1297).

But also in regards to above, the code you linked writes user project settings (Game.agf.user file).

The default config is written here: https://github.com/adventuregamestudio/ags/blob/v.3.6.0.5/Editor/AGS.Editor/AGSEditor.cs#L1516

ericoporto:
Ah, I followed the wrong code then. But yes, I meant `acsetup.cfg` that is shipped, since you mentioned it expects utf8 paths in config.

ivan-mogilko:
I'm not sure what encoding do these functions - `WritePrivateProfileString` - use, maybe they just write C-strings as-is. Need to test this. But also, these could be replaced with IniFile class from the common, or maybe similar class written in C# later.

ivan-mogilko:
Rebased to account for 7be9337211a937a604b6ad63d8d0946499b3fb15

ivan-mogilko:
For the reference, I don't think I want to add any changes to the Editor in this PR, rather make sure that the windows engine can work with the unicode paths properly. The editor may be adressed separately.

I'm still testing & fixing this, because there are some mistakes.

ivan-mogilko:
After latest fixes, the win engine is confirmed to work with unicode paths correctly.

Also, WinSetup is now a unicode-aware dialog, so it will also display and save unicode paths correctly.

ericoporto:
Hey, I tried with some games and directories and this appears to work with me. It's nice to see the full name of the directories when using `--tell-all` instead of the `USERNA~1/` from before.

PR #1386, opened 2021-08-08
Compiler: remove deprecated lefttoright flag

ericoporto:
fix the failing builds on the CI, I forgot I let the compiler front end building in it.


Compiler/main.cpp
```diff
@@ -208,7 +207,7 @@ ParsedOptions parser_to_compiler_opts(const ParseResult& parseResult)
                 continue;
             }
             if(flag_name == "lefttoright") {
-                compilerOptions.Flags.LeftToRightPrecedence = flag_value;
+                printf("Warning: lefttoright flag is deprecated");
```
ivan-mogilko, line 210: Probably should end with `\n`.

ericoporto: fixed!

PR #1388, opened 2021-08-08
CMake: Add crmpak to CMake build

ericoporto:
- also fixes for building with it

### DIFF
--- a/Tools/crmpak/main.cpp
+++ b/Tools/crmpak/main.cpp
@@ -24,7 +24,7 @@ public:
 static void print_known_blockids()
 {
     for (int i = kRoomFblk_FirstID; i < kRoomFblk_LastID; ++i)
-        printf("%d:%s\n", i, GetRoomBlockName((RoomFileBlock)i));
+        printf("%d:%s\n", i, GetRoomBlockName((RoomFileBlock)i).GetCStr());
 }
 
 HError print_room_blockids(RoomDataSource &datasrc)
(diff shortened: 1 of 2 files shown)


Tools/crmpak/main.cpp
```diff
@@ -153,7 +153,7 @@ int main(int argc, char *argv[])
     // Open the room, export list of block ids ('l' command).
     //-----------------------------------------------------------------------//
     RoomDataSource datasrc;
-    HError err = OpenRoomFile(in_roomfile, datasrc);
+    HError err = static_cast<PError>(OpenRoomFile(in_roomfile, datasrc));
```
ivan-mogilko, line 156: Why is this fix necessary, but there's no error reported by any of the build on CI?

ericoporto: The tools aren't being built in the CI currently. I thought it would only be necessary to build when there's a plan to how they are going to be released. If you want I can turn on for the CMake builds.

ivan-mogilko: So, this fails on one of the non-msvs compilers, like gcc?

ericoporto: It failed to me on Clang, you can probably use it even on Windows.

ericoporto: I don't understand why you seem to want an HError but made a function that returns something else.

PR #1389, opened 2021-08-12
Android single game package

ericoporto:
Fixes for building the an Android App 

- Reads files directly from Android Asset Manager
- Compatible with Play Asset Delivery
- Removes OBB since it's no longer supported
- Updated instructions in the README

Take time to review. There's one commit that removes one verification for Android only in `Engine/main/engine.cpp` which I am not confidant - the problem is it checks for files only, but the game is inside Android Asset Manager.


ivan-mogilko:
To elaborate, these are not only fixes for *building* android app, these are also updates for *running* android app that is compiled into APK or another android-specific package.

The new stream type is using Android's NDK and reads data not from "FILE" but from APK or other packs.

ericoporto:
The Android log library link addition commit is because we currently use what is "linked" (in CMake terms) by SDL2 (in 2.0.12) but in most recent SDL2 versions (2.0.14 and 2.0.16), it's "linking" libraries are now private, meaning we have to be explicit of the ones we want in AGS, I tested, and nothing changes EXCEPT this Android log library.

ivan-mogilko:
Finally got time to check this out; I can only judge the changes to the engine, and the only thing I don't quite like is adding of macro checks around a code in [this commit](https://github.com/ericoporto/ags/commit/e55e21a0b849cf52e39267c9ff7a23f9bfa65bf5).

What does cmdGameDataPath equal to in this case, is it a path to APK or something unrelated?

ericoporto:
It's the `path/to/mygame.ags`. I don't like that commit either which is why I marked it as hack. The problem is it needs some fallback way to check the assets in the Android Asset Manager too, after if it didn't found it in the path it think it is, it's because the file is inside.

An idea I had was , in Android platform checks, to simply try to open the AAssetStream before checking for the files, and if it didn't found, then log a debug message and proceed to go on with the file related functions.

ivan-mogilko:
> Other idea would be having some platform driver function that would do the check, doing nothing and returning false for base platform and doing aasset search for Android's case, and using such function in this place, before common file operations take place. It's possible the check should happen before even entering such function too.

That would be a much nicer solution imo; but something I don't yet understand, why does it have a path to *.ags when it runs from APK? What passes this argument into the engine then?

ericoporto:
> For example, on Windows when you just run *.exe then it does not have any directional argument, and proceeds to checking for attached data.

It would be theoretically possible to do something similar but it will still require a function in the platform driver since it would require listing files in the assets and figuring out which file is the game package. I don't know where this would happen though - literally in the code. 

> why does it have a path to *.ags when it runs from APK? What passes this argument into the engine then?

The runtime library was originally made for the multi game launcher and then repurposed for the single game launcher, so the thing that passes the path is the Single game Main Activity.

ivan-mogilko:
> It would be theoretically possible to do something similar but it will still require a function in the platform driver since it would require listing files in the assets and figuring out which file is the game package. I don't know where this would happen though - literally in the code.

What if you do that platform call under "#else" if `AGS_SEARCH_FOR_GAME_ON_LAUNCH` is not set? Like here: https://github.com/adventuregamestudio/ags/blob/2132ba833b9d4508ed3f81a65f531b88f6087cb5/Engine/main/engine.cpp#L993

> The runtime library was originally made for the multi game launcher and then repurposed for the single game launcher, so the thing that passes the path is the Single game Main Activity.

Is it possible to make single game launcher to *not* pass any redundant command arguments into the engine? That would solve this problem.

ivan-mogilko:
Alternatively, you could have a platform function that optionally accepts a game file name (*.ags), but checks it along with (prior or after) checking files on disk, around here:
https://github.com/adventuregamestudio/ags/blob/2132ba833b9d4508ed3f81a65f531b88f6087cb5/Engine/main/engine.cpp#L967-L973

ericoporto:
@ivan-mogilko , pushed a sketch commit trying to build on top of what I understood from the observations

ivan-mogilko:
I think it's okay for now.
Looking at the Android's "AAssetDir" api, this could probably be later remade into a android-specific implementation of a filefind iterator here: <https://github.com/adventuregamestudio/ags/blob/master/Common/util/directory.h#L52>

then we will have one interface to search in both directory and APK (and more platform implementations could be added later).

ericoporto:
That `FindFile` class is really interesting. I honestly had not seen it before, I will think about it. iOS has something similar with the contents of the bundle but from what I could grasp so far you can work with regular file API and nothing special is needed.

PR #1392, opened 2021-08-17
Ci CMake: Upgrade SDL2 to 2.0.16

ericoporto:
- SDL version updated to **2.0.16**
  - Android sdl project upgraded

I am leaving this as draft to give time to check for regressions with newer SDL2. If regressions are found, we need to report to upstream to get things fixed in time of ~~2.0.18~~ 2.0.20.

---

The items below have been already merged through PR #1398 

- Updates Xiph libraries package 
  - only change is libogg to 1.3.5 which doesn't actually matter, but it syncs version accross systems.
- SDL_sound version updated to newer commit 
  - fix build with MSVC and SDL 2.0.16, see https://github.com/icculus/SDL_sound/issues/25
- fixes for working with 2.0.16
  - explicitly link to android lib (same case of previous log library)
  - Wayland was disabled in SDL2 for our previous version (during build checks) but now is explicitly disabled in our Docker build
  - Since Debian Stretch is the new oldoldstable we may eventually upgrade to it in the future (would give Wayland support)
  - more info in https://github.com/libsdl-org/SDL/issues/4645

### DIFF
--- a/Android/library/runtime/src/main/java/org/libsdl/app/HIDDeviceManager.java
+++ b/Android/library/runtime/src/main/java/org/libsdl/app/HIDDeviceManager.java
@@ -7,6 +7,7 @@ import android.bluetooth.BluetoothAdapter;
 import android.bluetooth.BluetoothDevice;
 import android.bluetooth.BluetoothManager;
 import android.bluetooth.BluetoothProfile;
+import android.os.Build;
 import android.util.Log;
 import android.content.BroadcastReceiver;
 import android.content.Context;
@@ -275,6 +276,7 @@ public class HIDDeviceManager {
             0x15e4, // Numark
             0x162e, // Joytech
             0x1689, // Razer Onza
+            0x1949, // Lab126, Inc.
             0x1bad, // Harmonix
             0x24c6, // PowerA
         };
@@ -353,9 +355,18 @@ public class HIDDeviceManager {
 
     private void connectHIDDeviceUSB(UsbDevice usbDevice) {
         synchronized (this) {
+            int interface_mask = 0;
             for (int interface_index = 0; interface_index < usbDevice.getInterfaceCount(); interface_index++) {
                 UsbInterface usbInterface = usbDevice.getInterface(interface_index);
                 if (isHIDDeviceInterface(usbDevice, usbInterface)) {
+                    // Check to see if we've already added this interface
+                    // This happens with the Xbox Series X controller which has a duplicate interface 0, which is inactive
+                    int interface_id = usbInterface.getId();
+                    if ((interface_mask & (1 << interface_id)) != 0) {
+                        continue;
+                    }
+                    interface_mask |= (1 << interface_id);
+
                     HIDDeviceUSB device = new HIDDeviceUSB(this, usbDevice, interface_index);
                     int id = device.getId();
                     mDevicesById.put(id, device);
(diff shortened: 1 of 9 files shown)


ericoporto:
I also started to play around with 2.0.18, but I found a problem with Android CI, I opened a PR for Cirrus Docker images.

https://github.com/ericoporto/ags/tree/ags-sdl-2.0.18

ivan-mogilko:
Seeing as multiple versions of SDL2 mentioned above (2.0.18/2.0.20), it is important to keep in mind that some systems (like linux distributions) may not yet have these in their stable releases. In such case it may be better to postpone an upgrade, as requiring higher versions may make it inconvenient or impossible for the users of these systems to use AGS engine.

ericoporto:
I do not agree with this assesment @ivan-mogilko , I think the binaries we ship (as in, the ones that comes with the Editor) should use the newer SDL2 when available. We can ALSO build with an older SDL2 in the CI if you want to check API compatibility with an old version, but the newer versions of SDL2 have better compatibility with newer systems while keeping compatibility with older systems too. In the example case of Linux, we have Wayland compatibility only on the newer versions of SDL2.

ivan-mogilko:
So, do I understand this correctly, that so long as we keep sdl2 API in check, the engine should keep working with previous sdl2 builds (to some point)?
Is there anything else that may depend on sdl2 version, like we may need to write different code somewhere to accomodate for a newer behavior, but that the older versions do not need?

IIRC, one big problem is that sdl_sound is still not released "officially", so it's likely not available as a "proper" system lib on many systems. I guess that's a separate issue, but when there will finally be the release, we will also have to look out for its API compatibility too...

Since we're at this, could you tell, does our cmake link to a system sdl2 or fetches and builds selected version by default on Linux? Or is there a switch between these two modes?

ericoporto:
> So, do I understand this correctly, that so long as we keep sdl2 API in check, the engine should keep working with previous sdl2 builds (to some point)?

Yes, basically SDL2 API doesn't change much, it has new additions, but it never breaks the existing ones. Some things are added/fixed without altering API at all (like the SDL2 render backends, newer gamepad controls, and other device oriented things).

It's possible to test SDL2 version at build time too, but not at runtime - afaik. For instance, say we want to have a "System.OpenURL" based on SDL2 Open URL command (which is safer than people using AGSShell plugin), we could not build the feature if the used SDL2 is a version before 2.0.14.

Other than this I don't think there's much effort, basically avoid newer APIs in critical places - like in the previous example it doing nothing is an ok behavior, it probably won't alter much of the resulting game.

> Is there anything else that may depend on sdl2 version

As long as we are in 2.0.X I don't expect API/behavior to change in SDL2 side. 

> Since we're at this, could you tell, does our cmake link to a system sdl2 or fetches and builds selected version by default on Linux? Or is there a switch between these two build modes?

Currently it's fetching SDL2 from the same version we are using elsewhere. The exception is Emscripten PR which fetches the latest version in Emscripten (which is currently SDL2 latest master I believe), this is just because with Emscripten is better to sync libraries with the ones from Emscripten ports and Emscripten system libraries.

The CMake script currently doesn't use or link with any system libraries, it's completely self contained.

Is the idea of adding it so package maintainers can author their scripts using our CMake script instead of the Makefiles?

ericoporto:
Btw I don't think we NEED to upgrade the SDL2 version, this is why this is a draft, but I would like to be compatible and test with newer versions. For this sometimes I need to modify AGS and sometimes I need to modify SDL2, I am just tracking things to ensure things are compatible.

ericoporto:
I am closing this. While I do believe upgrading SDL2 version is necessary, I think I will do this in a way with build checks for previous version too, and it will take me a bit of time to properly figure it out.

Issue #967, opened 2019-10-21
In graphic renderers replace SpriteBatch list with a SceneNode tree


ghost:
Sprite batches were added in 3.5.0. SpriteBatch is a parent entity of a group of sprites, which defines their global render transformation: offset, scaling and rotation (latter not supported in practice yet), and few others. This transformation is merged with each child sprite's own.
SpriteBatch may be thought as a "directory" for sprites, which applies shared settings.
Currently sprite batches are stored sequentially in a linear list, and rendered in the same order.

This system has limitations, and for some situations it may be more convenient to allow child sprite batches inside the parent. Also there's currently no expliciy way to end the list of sprite batches (it's reset automatically when final Render is called).

Thus the suggestion is following:

1. Have a generic struct called "Scene Node", "Node" or "Scene". The SpriteEntry is either derived of that Node struct or merged with one (whichever seem convenient).
2. Remake list of sprite batches into the tree, where each node may be a parent of more nodes, and you may have sprite entries and "group" nodes on the same level.
3. When deciding on the final sprite's transform, SpriteEntry/Node's transformations are applied from bottom to top (logically, actual matrix order is renderer-specific).
4. The preferred API is not clear at this point, and there is also #969 that may affect its selection, if get accepted. But I believe that there should be a way to mark beginning and ending of the scene, for explicitly indicating when we begin sending data for the next render step and when we finish doing so. This should make it clear throughout the engine code where scene begins being constructed and where it ends.

ericoporto:
What is the usage of an 'EndSpritreBatch'?

Also, in the SceneNode tree, suppose we have two viewports in a game, and they, point to the same camera, or point to a different camera. I am trying to understand the order the transformations are supposed to be applied, is each viewport a node, and both has all things in the room as children, and the window itself is a root node? I am having some difficulty understanding the tree without a drawing :/

ivan-mogilko:
> What is the usage of an 'EndSpriteBatch'?

Hypothetically, tell renderer that you finished adding items to current sprite batch.

> Also, in the SceneNode tree, suppose we have two viewports in a game, and they, point to the same camera, or point to a different camera. I am trying to understand the order the transformations are supposed to be applied, is each viewport a node, and both has all things in the room as children, and the window itself is a root node? I am having some difficulty understanding the tree without a drawing :/

The tree is simply a way to organize multiple children in the same node. For an individual object the tree structure may be ignored, and instead you have a line which goes through all its parents up to the root node, for example:

[...]

PR #1393, opened 2021-08-19
[WIP/TEST] Draw GUI controls as separate textures

ivan-mogilko:
This is only a test for now, so marked as wip.

The GUI controls were historically not rendered as separate objects added into the list of sprites, but instead drawn on a GUI surface. On one hand this is very simple, on another - a) engine has to redraw whole GUI when anything on it changes, b) this does not let to apply same graphical properties and effects as for other objects (accelerated transformation, shaders and so on).

This pr changes this, giving each gui control its texture and passing them through the same procedure as every other object on screen.

In terms of perfomance, in low-res games there is no much difference. In high-resolution games there's a noticable perfomance improvement (especially in unlimited fps mode), but it turned out to be not that big. There were other perfomance fixes recently, which restricted GUI redraws to only when something really changes on them, and these seem to had much more impact.
Perhaps it will be interesting to make tests with larger amount of changing controls, like multiple animated buttins in high resolution, and so on.

In any case, for now this particular change does not seem to be seriously necessary for ags3 branch (3.6.0 version). This may come handy when we want to support individual control scaling and other transformations, and/or rendering controls in the window resolution (similar to the character sprites in the corresponding mode).

---

Current issues:

1. Control contents (text, image) are clipped by the size. This is different from how the engine currently works, where text may freely exceed control frame (due to how they are drawn). Personally I think content clipping is how it is normally supposed to be, but regardless it would have to have non-clipping mode to support old behavior. This may be done by telling the "extended metrics" of the controls, precalculating the size their contents would take.
2. On contrary, the controls themselves are supposed to be clipped by their parent GUI, but with this change they are not. The solution may be to define each GUI children as a sprite batch limited to GUI-sized area when these sprites are passed into the renderer.
3. Historically GUI had 3 modes of blending controls (named "classic", "additive" and "proper"). Only "proper" would be supported by OpenGL and Direct3D without additional changes. The "classic" perhaps may be done by marking controls' sprites "opaque", but "additive" mode would require a shader. Temporarily in the first two old modes controls are drawn the old way (on GUI bitmap).


ivan-mogilko:
> On contrary, the controls themselves are supposed to be clipped by their parent GUI, but with this change they are not. The solution may be to define each GUI children as a sprite batch limited to GUI-sized area when these sprites are passed into the renderer.

This may be solved using a simplier solution within 3.6.0 ([example](https://github.com/adventuregamestudio/ags/pull/1393/commits/d3a2a897a2e19aef1f3f883486c603c8cd4e090d)), but will anyway require completing #967 for ags4, because guis support additional transformations there.

ericoporto:
Could GUIs be like viewport and camera for the GUI Controls they own? 

I don't know if it's possible since we can't ZOrder Viewports, my guess is it's not feasible with the current things in the graphics part.

ivan-mogilko:
> Could GUIs be like viewport and camera for the GUI Controls they own?

The difference is that viewports dont have their own graphic, but GUIs do. So there's no "texture" passed for viewport, but there is for gui. Also guis are sorted among overlays too.

> I don't know if it's possible since we can't ZOrder Viewports

Viewports have ZOrder property, but it works only among viewports atm. Sorting them among GUI would require rewriting how list of drawable items is constructed, so probably will be possible only after #967.

> having this as is (without #967), would it make merging things to ags4 harder

Merging code itself would be not difficult, but it simply won't work in ags4; the fallback method would work, but that will defeat the purpose of this feature. So if i do that, someone will have to later fully reimplement control texture rendering in ags4.

ivan-mogilko:
Okay, i did something similar to #967 here, now there's BeginSpriteBatch and EndSpriteBatch, and they support sub-levels. At least this works with GUI controls.

ericoporto:
Just a question, not something for now, just thinking an idea here, can it check after a frame ended that no begin call was left missing an end and then log a warning somewhere?

ivan-mogilko:
So, the rendering controls as textures currently has restrictions, which should be clear from the code:
https://github.com/adventuregamestudio/ags/pull/1393/commits/dadd6d81fb9160623a1ca8c9f4ed01b1d41a271e#diff-1d31b249a940535166d842cb1b562d67f46362d7a99e5ea8aa127f942fe24b6dR2286-R2293

These restrictions may theoretically be removed with additional changes later.
For software renderer, there's additional concern that using the old way may be faster at least in low-mid game resolutions.

ivan-mogilko:
So, the rendering controls as textures currently has restrictions, which should be clear from the code:
https://github.com/adventuregamestudio/ags/pull/1393/commits/667bdc939648d705fc385dedaae5e2cd80cf3b99#diff-1d31b249a940535166d842cb1b562d67f46362d7a99e5ea8aa127f942fe24b6dR2286-R2293

These restrictions may theoretically be removed with additional changes later.
For software renderer, there's additional concern that using the old way may be faster at least in low-mid game resolutions.
In regards to clipping, this is now solved for the Slider, the only control that must be drawn without clipping for now, as we decided previously. Removing this restriction is a matter of precalculating unclipped text size for labels and buttons and list boxes; that may be easy, but it's better to do some refactor and ensure this calculation is done only when the text and other properties change, and not each frame.

ivan-mogilko:
Got good news from @tarnos12, who has a 1920x1080 game, and this change significantly improves a scene with multiple moving gui buttons. So all this definitely was not in vain.

ivan-mogilko:
Rebased and squashed commits.

Added functions that calculate control content sizes for "non-clipping" mode; this allows to also draw these controls as textures.
This increases the code size, but not *too* much. On the other hand this will make the new render mode enhance running older high-resolution games too. I think this may be cut from the ags4 though, if we decide to keep gui controls strictly clipped.

ericoporto:
Hey, I think someone is missing in the change from the commit here: https://github.com/adventuregamestudio/ags/pull/1393/commits/0156847541c02777f6ca0b672667ea732e8845da

The error at building the Editor complains of s missing NotifyParentChanged.

ivan-mogilko:
In regards to inventory's white rectangle, the same bug is present in tag v.3.6.0.22.

ericoporto:
Hey, reading the top post I don't understand if you plan on merging this or not, but if you do, vga256 on the forums is making a game that uses lots of GUIs, so if there's any issue I am sure it will be picked quickly. So once this is merged, probably would be nice to get a new release not long after.

ivan-mogilko:
Updated the pr description.

ericoporto:
I have a minor issue in the GUIs of I Rented a Boat. Left is from this PR, right is from master (which is the intended look)
image

ivan-mogilko:
I guess the "has alpha" flag may be not set correctly on the control textures.

ericoporto:
Damn, find another thing, still on I Rented a Boat. If you load it, in the options menu, you can click to alternate between Windowed and Fullscreen. With the current master, this works alright, but with this PR I get the following error:

image

ivan-mogilko:
> Damn, found another thing, still on I Rented a Boat. If you load it, in the options menu, you can click to alternate between Windowed and Fullscreen. With the current master, this works alright, but with this PR I get the following error:

So, apparently this happens when buttons change visibility. I.e. similar code:
https://github.com/ericoporto/i_rented_a_boat/blob/main/I_rented_a_boat/Menu.asc#L42-L46

ericoporto:
when visibility changes, are any members of the `guiobjbmp` vector removed?

ericoporto:
Hey, I tested some more games and I could not finding anything now. I think it's safe to merge! 😃

ivan-mogilko:
I realized that made a mistake with the changes to sorting algorithm ([this commit in particular](https://github.com/adventuregamestudio/ags/pull/1393/commits/bbae9ce641f5b9ab26e8ca797c2f7efdf21ee0fe)); it works only while certain conditions are met. I need to rethink this and make differently.

ivan-mogilko:
Ok I think it's done now; although gathering overlay + gui + gui control textures has a little "unexpected" order of actions, but in general the code became more straightforward (imho). It's far from perfect, but this is trying to work within the existing system...

PR #1397, opened 2021-08-23
Reimplement room mask display as a persistent overlay

### LINKED ISSUE #724: Redesign runtime debug overlay

ivan-mogilko:
Resolves #724.

1. "Show walkable areas" debug command (`Debug(2, 0)`) is reimplemented from displaying an opaque mask over a freezed game into persistent half-transparent overlay that does not pause the game.
2. `Debug(2, n)` now has a meaningful second parameter which defines a mask type, values are: 0 -none, 1 -hotspot, 2 -walk-behind, 3 -walkable, 4 -regions.
3. `Debug(2, n)` command now works as a on/off toggle, so calling it will same mask type consecutively will turn it on and off.  This also helps with backward compatibility, because old games that have this command in script would not have a way to turn masks off.
4. "Show character walking paths" debug command (`Debug(5, n)`) is similarily reimplemented as a half-transparent overlay, and working in a toggling manner.

**NOTE:** In many old games this command is run on "Ctrl+A"; you may have to run them in debug mode though (`ags game --test`)

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -11,7 +11,6 @@
 // http://www.opensource.org/licenses/artistic-license-2.0.php
 //
 //=============================================================================
-
 #include <algorithm>
 #include <cmath>
 #include "aastr.h"
@@ -106,7 +105,7 @@ RGB palette[256];
 
 COLOR_MAP maincoltable;
 
-IGraphicsDriver *gfxDriver;
+IGraphicsDriver *gfxDriver = nullptr;
 IDriverDependantBitmap *blankImage = nullptr;
 IDriverDependantBitmap *blankSidebarImage = nullptr;
 IDriverDependantBitmap *debugConsole = nullptr;
@@ -122,6 +121,9 @@ std::vector<CachedActSpsData> actspswbcache;
 // GUI surfaces
 std::vector<Bitmap*> guibg;
 std::vector<IDriverDependantBitmap*> guibgbmp;
+// For debugging room masks
+RoomAreaMask debugRoomMask = kRoomAreaNone;
+IDriverDependantBitmap *debugRoomMaskDDB = nullptr;
 
 bool current_background_is_dirty = false;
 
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Also reimplemented display of walking paths; it's run as `Debug(5, n)`, where n is character's index, and this is strangely not documented in the manual currently.

ericoporto:
> this is strangely not documented in the manual currently.

I honestly didn't knew about `Debug(5,n)`, added a mention there just to not forget it's existence.

ivan-mogilko:
I'm not entirely happy with the way I made these commands work atm, I will think a bit more, maybe make it work more conveniently in the new versions first and then figure out some trick to keep this usable in the old games.

>  I am impressed it works with multiple cameras!

yes, thankfully this happens automatically now, because I pass the debug overlay as a part of the room sprite list.

ericoporto:
> I'm not entirely happy with the way I made these commands work atm

If it's only the cmd to switch between masks and not the implementation on how they draw, perhaps this could be left for improvement later. As is, it already makes life easier when using multiple cameras, plus you can activate and play the game normally, it's really useful as is!

ivan-mogilko:
So, I ended up with having `Debug(2, n)` command, where you have to pass which mask type exactly you want to see.
Values are: 0 -none, 1 -hotspot, 2 -walk-behind, 3 -walkable, 4 -regions.
The command works as a toggle between on/off, so calling `Debug(2, 3)` twice in a row will turn them on and then off.

More details are in the edited PR description.

PS. I was also thinking about maybe implementing hard-coded hotkeys for this enabled only in the test mode, to have debug overlays working regardless of whether they are in script or not, but maybe this could be done later.

ericoporto:
I think it works fine as is, I think Debug commands are less problematic to change later, so it would be good to roll this into a release to get more feedback :)

PR #1405, opened 2021-08-26
New compiler: Better error messages for type mismatches

fernewelten:
Less confusing type descriptions:
– e-g-. "Can't convert **an array** to 'int'", instead of printing a complex type expression.

Better pinpointing:
– In expressions, name the specific operator that has the error and (if applicable) whether the error is the left-hand side or right-hand side 
– In function calls, name the specific function and the number of the parameter that has the error
This should make it easier to find the error location if the AGS code has a complicated expression.

Typical examples:
```
int y[];  
++y;
```
> ` Argument of '++': Can't convert an array to 'int' `

```
import function DoIt(float a, float b);
… DoIt(4, 4.0); …
```
> `Parameter #1 of call to function 'DoIt': Can't convert 'int' to 'float' `

```
int x, y;
float fl[];
int res = (x + y) * (x - fl);  
```
> `Right-hand side of '-' term: Can't convert an array to 'int'`

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -2992,7 +2992,7 @@ AGS::ErrorType AGS::Parser::AccessData_FunctionCall_PushParams(SrcList &paramete
                 kKW_String == _sym.VartypeWithout(VTT::kConst, param_vartype))
                 WriteCmd(SCMD_CHECKNULLREG, SREG_AX);
 
-            std::string msg = "Parameter #<num> of function <func>";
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
@fernewelten I apologize for a big delay. Testing this PR, I noticed that there a small typo in the error message produced by the third code example, it reads:
`Right-hand side of '-' term:: Can't convert an array to 'int'`
notice the double "::".

ivan-mogilko:
@fernewelten could you please rebase your branch, it looks like you merged instead and now it pulls 50 extra commits from somewhere else.

PR #1408, opened 2021-08-27
Engine: Link in such a way that functions can call all imported (!) functions, no matter the module sequence

fernewelten:
Rationale:
- It makes life easier for coders, they have less reason to juggle the module sequence. 
- This is a **prerequisite** for implementing any sort of a function pointer or callback mechanism. For callbacks, it's kind of the point that we can call a function without knowing anything about it, in particular without having to guarantee that it has been defined in an earlier module. 

AFAICS existing programs don't need to be changed in any way for this PR; they should continue to work without recompiling.
AFAICS neither the new nor the existing compilers need to be changed for this PR.

The effect is achieved by changing the sequence of the linking steps.
- First, create the instances for the scripts.
- Then, resolve references to imports and the 'IMPORT' fixups.
- Then, create the forks.

### DIFF
--- a/Engine/ac/room.cpp
+++ b/Engine/ac/room.cpp
@@ -945,6 +945,12 @@ void compile_room_script() {
         quitprintf("Unable to create local script: %s", ccErrorString.GetCStr());
     }
 
+    if (!roominst->ResolveScriptImports(roominst->instanceof))
+        quitprintf("Unable to resolve imports in room script");
+
+    if (!roominst->ResolveImportFixups(roominst->instanceof))
+        quitprintf("Unable to resolve import fixups in room script");
+
     roominstFork = roominst->Fork();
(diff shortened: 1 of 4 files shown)


AlanDrake:
A change like this was long needed.
I gave it a quick spin with a couple of projects, and the only case where I managed to trigger an error was when I moved Globalscript all the way up, but it was due to an imported variable, not a function.

ivan-mogilko:
I'm sorry, i got really distracted last week; will try to double check this again in the following days and merge if everything is fine. Maybe we could even copy it to 3.6.0 too, as this does not depend on compiler and may be quite beneficial.

AlanDrake:
I've been running a build with this PR for twelve days, as far as functionality I'd say it works as expected.

PR #1410, opened 2021-09-02
SDL2 engine: window icons

ivan-mogilko:
This reimplements game window icons for Windows and Linux in the new engine.

Added `SDL_Surface *CreateWindowIcon()` method to the platform driver, the idea is that it either creates and returns SDL_Surface with an icon, or sets an icon itself and returns null, or just returns null if does not support this. If the return value is not null, then engine passes the SDL_Surface object into `SDL_SetWindowIcon` and disposes it.

---

On Windows this method simply assigns an icon from resources; I figured it's better to make a pair WinAPI calls than bother generating sdl image from a resource.
There was another reason though; at first I tried the hard way for a test's sake: load icon, extract pixels, convert into SDL_Surface format, and so on. But it seemed that SDL2 does something wrong, or I did, or perhaps there are more requirements for this to work, because the icon appeared as if it were downgraded in color depth.

---

On Linux, we were embedding an XPM file previously, which allegro4 loaded using X11 libs. I am unsure whether it's wanted or not that we link x11 to the engine again. If we do, then I think I could copy original allegro4 code that set up an icon ([one may check it out here](https://github.com/liballeg/allegro5/blob/4.4/src/x/xwin.c#L2243)), or use something similar.
For now I went a "hard" way here and wrote a primitive XPM parser that reads its array and writes into RGBA buffer.

The problem here is, that although this worked when I test on Windows, similar code has no result when running on Linux... the icon is simply not set (I see a gray question mark instead on Ubuntu). I'd leave this PR as wip for a little while, and do more tests and debugging at the same time.

**PS.** BTW, maybe this code should rather be in the parent AGSPlatformXDGUnix class, thus shared with FreeBSD? Or, in theory, with manual XPM loading, same icon could be used on any platform, even Windows.

### DIFF
--- a/Engine/platform/base/agsplatformdriver.h
+++ b/Engine/platform/base/agsplatformdriver.h
@@ -25,6 +25,8 @@
 #include "debug/outputhandler.h"
 #include "util/ini_util.h"
 
+struct SDL_Surface;
+
 namespace AGS
 {
     namespace Common { class Stream; }
@@ -103,6 +105,9 @@ public:
     virtual void UnRegisterGameWithGameExplorer();
     // Adjust window size to ensure it is in the supported limits
     virtual void ValidateWindowSize(int &x, int &y, bool borderless) const {}
+    // Either set window icon using system API directly, or create a SDL_Surface
+    // for the SDL backend to set an icon instead.
+    virtual SDL_Surface *CreateWindowIcon() { return nullptr; }
 
     virtual int  InitializeCDPlayer() = 0;  // return 0 on success
     virtual int  CDPlayerCommand(int cmdd, int datt) = 0;
--- a/Engine/platform/base/sys_main.cpp
+++ b/Engine/platform/base/sys_main.cpp
@@ -14,6 +14,7 @@
 #include "platform/base/sys_main.h"
 #include <SDL.h>
 #include <SDL_syswm.h>
+#include "platform/base/agsplatformdriver.h"
 #include "util/geometry.h"
 #include "util/string.h"
 
@@ -149,10 +150,10 @@ void sys_window_set_title(const char *title) {
 
 void sys_window_set_icon() {
     if (window) {
-        // TODO: actually support getting icon from resources and converting into SDL_Surface.
-        //  - on Linux we had platform/linux/icon.xpm
-        //  - on Windows we had standard embedded resource under ID = 101
-        SDL_SetWindowIcon(window, nullptr);
+        SDL_Surface *icon = platform->CreateWindowIcon();
+        if (!icon) return; // no icon
+        SDL_SetWindowIcon(window, icon);
+        SDL_FreeSurface(icon);
     }
 }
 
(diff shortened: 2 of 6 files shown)


ivan-mogilko:
Tried to replicate allegro4 code to set an icon using Xlib functions, and also failed. For some reason a call to XSetWMHints always returns "BadRequest (invalid request code or no such operation)".

The code I used (in the Linux's platform driver class):
```
#include <SDL.h>
#include <SDL_syswm.h>
#include <X11/Xlib.h>
#include <X11/xpm.h>

<...>

// Get X display and window from SDL
SDL_SysWMinfo wminfo;
SDL_VERSION(&wminfo.version);
SDL_GetWindowWMInfo(sys_get_window());
Display* display = wminfo.info.x11.display;
Window window = wminfo.info.x11.window;

// This works: change window's title for a test
XStoreName(display, window, "test window title");

int result = 0;
char err[256] {0};
XWMHints wmhints;
wmhints.flags = IconPixmapHint | IconMaskHint;
result = XpmCreatePixmapFromData(display, window, (char**)icon_xpm, &wmhint.icon_pixmap, &wmhint.icon_mask, NULL);
XGetErrorText(display, result, err, sizeof(err));
// no error here, also icon_pixmap and icon_mask have valid values assigned
Debug::Printf("XpmCreatePixmapFromData = %d, %s", result, err);
result = XSetWMHints(display, window, &wmhints);
XGetErrorText(display, result, err, sizeof(err));
// Returns code 1 (BadRequest)
Debug::Printf("XSetWMHints = %d, %s", result, err);
```

This code is almost the same as was used in allegro4; so probably I am missing something.

ericoporto:
@ivan-mogilko I would not worry much with the Linux icon. If the Linux computer adhere to Freedesktop, the icon used is the one installed, see [Freedesktop icon spec](https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s07.html). It also needs a `.desktop` file that will point to it, [see here in arch wiki](https://wiki.archlinux.org/title/desktop_entries).

I also don't understand why doing platform specific icons instead of just using [SDL functions](https://wiki.libsdl.org/SDL_SetWindowIcon).

ivan-mogilko:
> @ivan-mogilko I would not worry much with the Linux icon. If the Linux computer adhere to Freedesktop, the icon used is the one installed, see [Freedesktop icon spec](https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s07.html). It also needs a `.desktop` file that will point to it, [see here in arch wiki](https://wiki.archlinux.org/title/desktop_entries).

It worked with the previous version though.

> I also don't understand why doing platform specific icons instead of just using [SDL functions](https://wiki.libsdl.org/SDL_SetWindowIcon).

What do you mean by "platform specific icons", icon images, or methods to set these? I explained the reasons for choosing the methods in the first post.

ivan-mogilko:
I ended up pushing first commit that adds general method of setting an icon and Windows implementation (67f1d33c58fcf77037d82b57a8c6a269a0f3665c), but the one for linux requires more research, so I will be dealing with it separately.

PR #1414, opened 2021-09-18
Editor: Sound indicator for view frames

ericoporto:
Sound indicator in case the frame has a sound ([requested here](https://www.adventuregamestudio.co.uk/forums/index.php?topic=58788)).

<details>
 <summary>Sound Indicator looks like this</summary>

</details>


AlanDrake:
The sound indicator could be put on a corner of the frame itself

ericoporto:
I think for the sound indicator I will try @AlanDrake suggestion, probably on bottom right corner of the image where usually it will be a pink background which should work alright with the sound icon. :)

ericoporto:
sound indicator experiment

Ignore the zoom, just trying to see if you guys think the sound indicator works on bottom right corner.

AlanDrake:
Hmm, I used a random online icon editor, no idea why it picked an inverted color it looked transparent in there 😕

ericoporto:
The inverted color is a thing, it's like when your mouse becomes a caret, usually the caret cursor is draw using the inverted color, it will reverse colors below. I liked it like this, so I thought it was intended.

morganwillcock:
I'm not sure it is a good idea to potentially cover content in the image unless you can toggle the sound icons on and off. I'd probably prefer to have it underneath as an icon (⏲3 🔊) but with the current frame layout I'm not sure there is enough space if there was scaling applied to the interface itself. The problem with a toggle is that then it introduces a potential need for a preference of the default state and saving the current state, and potentially someone will ask for that to be per-view.

So for the sound icon I worry that this is just going to trigger additional requirements based on feedback. The zoom slider seems a safe and useful addition.

Side note: I don't think it is obvious that "DLY" means delay and indicating DLY:0 seems redundant (in the same way that there is no icon for frames without sound).

ivan-mogilko:
> The problem with a toggle is that then it introduces a potential need for a preference of the default state and saving the current state, and potentially someone will ask for that to be per-view.

I strongly doubt per-view setting make sense. This icon is not a part of view, but a part of the view editor appearance. And of course we never have to comply to any given user request.
On the other hand, the toggle may not be a bad idea, in my opinion. If toggle existed, and had a keyboard shortcut, this would let user turn it on and off when editing particular views easily.
We already have editor preferences and workspace settings, so there's a way to save this setting both for an editor and for a project if necessary.

> So for the sound icon I worry that this is just going to trigger additional requirements based on feedback.

Sound icon is added as a response to multiple user request, because as of now there's no visual indication to which frames have a linked sound.

> Side note: I don't think it is obvious that "DLY" means delay and indicating DLY:0 seems redundant (in the same way that there is no icon for frames without sound).

I disagree, delay is a primary animation parameter, that matter for each frame, while sound is an option and used sparingly.
DLY may be renamed to something, or even replaced with a "time-related" icon. In any case, I think it takes a little time to get used to what it means.

ericoporto:
There's a low chance the sound icon will cover something because of how scaling is done, the image of the frame has to be square to have pixels covered. I tried before with the icon on the bottom, but I am convinced it's best in the frame itself, it gives a much better ownership sense - specially when resizing is done.

morganwillcock:
> I tend to disagree, delay is a primary animation parameter that matters for each frame, while sound is an option and used sparingly.

> There's a low chance the sound icon will cover something because of how scaling is done, the image of the frame has to be square to have pixels covered. I tried before with the icon on the bottom, but I am convinced it's best in the frame itself, it gives a much better ownership sense - specially when resizing is done.

I think there are a lot of assumptions being made what the use case of the view is. There is no reason that delay is relevant when all someone is interested in is using linked audio at a constant speed or any guarantee that the dimensions of the sprite will be taller than they are wide.

AlanDrake:
Well, we could also make a smaller icon that's just an inverted outline and squash all complaints that way.

morganwillcock:
> I like being able to read delay because I need it.

I'm not saying to make it unimportant. But I don't think hiding modified values in a sea of 0s helps to understand what delays are already set.

<table>
<tr><td></td><td>1</td><td>2</td><td>3</td><td>4</td><td>5</td><td>6</td><td>7</td></tr>
<tr><td>1</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>2</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>3</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>4</td><td>DLY: 0</td><td>DLY: 8</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>5</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>6</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
<tr><td>7</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td><td>DLY: 0</td></tr>
</table>

AlanDrake:
Ah, I see now what you meant. Yeah it could be displayed only when it's not zero. But that has nothing to do with slider and and soundframes, we can fix it anytime.

ericoporto:
image

image

Trying to convey the difference. Also removed Delay info temporarily, no need to think about that now.

ivan-mogilko:
Idk if it's just me, but now I wonder if the delay and sound icon will become more readable if they were on top rather than the bottom of a frame...

ivan-mogilko:
Perhaps it would be actually better to merge animation preview zoom with the inventory preview zoom in one pr, and address sound indication separately?

ericoporto:
ok, I will pull the zoom from here to the other and leave this one to figure out later.

ericoporto:
@AlanDrake do you think you can make a pixel drawing for the delay icon? I would like to adjust that too. Something as small as the sound indicator icon you did. I want to remove the `DLY: 0` from all views and only signal when it's different using an appropriate icon and the value.

AlanDrake:
@ericoporto sure, I'll come up with something soon

ericoporto:
image

It's lookin like this for now

AlanDrake:
Experiments:
immagine

Any preference?

ericoporto:
image

I really like the one you selected to put right below the frame. I like the idea of the tiny ones, but they look reaaaally tiny in my monitor. Can you make them in .ico format?

AlanDrake:
Looks good, but I'd put the clock on the left side of the number

ericoporto:
But then the number ownership is ambiguous, as is, it's owned/related to the clock.

ivan-mogilko:
> But then the number ownership is ambiguous, as is, it's owned/related to the clock.

Is not number related to the clock?
Reading the previous conversation, the clock means "delay" and the number is length of delay?

ericoporto:
> Or do you mean that it's unclear if it refers to the clock or to the sound icon?

If the number is in the middle of both it's unclear, if it's at left of the clock, and the clock is between the number and the audio icon , you read number->clock or clock->number if your culture reads left to right or right to left, being better at establishing that the number and the clock are somehow related.

ivan-mogilko:
> you read number->clock or clock->number if your culture reads left to right or right to left

But this is not how i read it, personally I read : icon = value.

ivan-mogilko:
Perhaps it would be more clear if the icon and number were outlined to show they are related?

Also, idk if this is planned, but if the frame has no delay the the sound icon is in the middle, but if it has a delay then it's more to the right.

ericoporto:
> Also, idk if this is planned, but if the frame has no delay the the sound icon is in the middle, but if it has a delay then it's more to the right

It's in the code and in the image on the top 😬 

> Perhaps it would be more clear if the icon and number were outlined to show they are related? Or there would be some kind of separator between them and the next icon; like a vertical line.

That's too much information, I think as is is clear enough. The person is meant to edit the properties through the property panel anyway, this is more of a visual indicator.

ivan-mogilko:
Well, I can merge this if it's fine, there will always be an opportunity to polish this.

ericoporto:
Pushed a tiny two pixel adjustment

image

I think it's good enough now, if you are good, it's mergeable!

ivan-mogilko:
I found one silly issue: apparently these hints are not restricted to the size of the frame. If you type a big delay number, the number and icons begin to expand over to other frames.
viewframehints

ericoporto:
I can make it so anything beyond 99 becomes +99 and you have to read the property for the full information. Other than this not sure what can be done. Isn't this a problem in current implementation too?

selmiak:
this is very good looking. I'd always align the sound icon to the right side though.
can we have svg images that scale with the new zoom factors?

selmiak:
and the option to quickly mute the sound on click would be cool, but I know it is just information display, convenience should be next :)
same for the delay, "muting" it on click and later even editing the values there would be fun. maybe in ags4...

Issue #710, opened 2019-04-10
Zoom slider in the Views panel?


HectorBometon:
Just a little suggestion for minor improvements: I think a zoom slider in the Views, similar like the one on the Rooms, could be great. 

Also, Delay text under frames on dark theme is hard to read.

![improvements](https://user-images.githubusercontent.com/49480354/55872991-bc0d9080-5b8e-11e9-9125-ac1e643ed6f9.png)

selmiak:
I proposed something similar for the inventory item thumbnails as issue #575 

So maybe there are more thumbnails that should be zoomable to more than 100% for our nice pixelart. in the sprite manager for example....

ericoporto:
The source code already has a [`FRAME_DISPLAY_SIZE`](https://github.com/adventuregamestudio/ags/blob/ba575681a33ce0c69f28cf4b5e62b948127af1d0/Editor/AGS.Editor/Panes/ViewLoopEditor.cs#L53) variable, but there are some additional magic numbers that need to be figured out for varying it's value not break the appearance.

ericoporto:
The view frames zoom feature is now in, the preview pane I was not sure how to deal with it design wise.

ivan-mogilko:
@ericoporto 
> The view frames zoom feature is now in, the preview pane I was not sure how to deal with it design wise.

May it be possible to simply scale the preview frame following same existing slider, similar to how it's done for the cursor and items prevew? That might work for the time being.

selmiak:
what do you mean design wise? If the preview should have a separate size slider or use the same scaling the rest of the views?
I'd say whatever is easier to implement and then we'll see. But the preview should scale in one way or the other too of course.
The only advantage a separate zoom control would give is that the display is not *that* jumpy when dragging the slider, but I can live with that, the player does not have to see this. another idea would be to have a "lock scaling" checkbox above the preview to avoid that, and when unticked it jumps to master scale and is a bit jumpy but that is okay.

ericoporto:
I got the view preview zoom working here: https://github.com/ericoporto/ags/tree/view-preview-zoom

I just need to workout a detail when the view preview is used in the character screen - it's zoom there is defaulting to 100% and also zooming a lot is not increasing the size of the area if possible like in the Inventory screen. :/

ivan-mogilko:
> I got the view preview zoom working here: https://github.com/ericoporto/ags/tree/view-preview-zoom

[...]

PR #1415, opened 2021-09-19
Editor: Zoom Slider in Inventory and View panels

### LINKED ISSUE #575: inventory items: Make inventory items images zoomable
this makes it easier to see these usually small images better and even easier to precisely place the click hotspot for the mousecursor.
2x or 3x zoom level could be enough already.
0,5x zoom could be handy when for whatever reason someone has huge mousecursor gfx.

ericoporto:
adds a zoom slider for inventory item images. You can also hold CTRL to scroll with mouse wheel.
A zoom slider is also added to the View Editor!
Edit: and also to the Cursor editor!

image
image

fixes #575 (and maybe #710)

### DIFF
--- a/Editor/AGS.Editor/Panes/InventoryEditor.cs
+++ b/Editor/AGS.Editor/Panes/InventoryEditor.cs
@@ -11,13 +11,42 @@ namespace AGS.Editor
 {
     public partial class InventoryEditor : EditorContentPanel
     {
-        private int IMAGE_SCALE_FACTOR = 2;
         private InventoryItem _item;
+        private int _previousItemCursorImage;
+        private int _previousItemImage;
+
+        void UpdatePanelSizes()
+        {
+            if (_item != null)
+            {
+                Size itemSpriteSize = Utilities.GetSizeSpriteWillBeRenderedInGame(_item.Image);
+                Size itemCursorSpriteSize = Utilities.GetSizeSpriteWillBeRenderedInGame(_item.CursorImage);
+                itemSpriteSize.Width = Scale(itemSpriteSize.Width);
+                itemSpriteSize.Height = Scale(itemSpriteSize.Height);
+                itemCursorSpriteSize.Width = Scale(itemCursorSpriteSize.Width);
+                itemCursorSpriteSize.Height = Scale(itemCursorSpriteSize.Height);
+
+                pnlInvWindowImage.AutoSize = false;
+                pnlInvWindowImage.MaximumSize = itemSpriteSize;
+                pnlInvWindowImage.MinimumSize = itemSpriteSize;
+                pnlInvWindowImage.Size = itemSpriteSize;
+                panelScrollAreaCursor.AutoScroll = true;
+                pnlCursorImage.AutoSize = false;
+                pnlCursorImage.MaximumSize = itemCursorSpriteSize;
+                pnlCursorImage.MinimumSize = itemCursorSpriteSize;
+                pnlCursorImage.Size = itemCursorSpriteSize;
+                panelScrollAreaImage.AutoScroll = true;
+
+                _previousItemCursorImage = _item.CursorImage;
+                _previousItemImage = _item.Image;
+            }
+        }
 
         public InventoryEditor()
         {
             InitializeComponent();
             Factory.GUIController.ColorThemes.Apply(LoadColorTheme);
+            UpdatePanelSizes();
         }
 
         public InventoryEditor(InventoryItem itemToEdit) : this()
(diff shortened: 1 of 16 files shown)


ivan-mogilko:
I think my biggest criticism to this and #1414 would be that the zoom levels are becoming inconsistent throughout the application.

The Room and GUI zoom sliders have a same range of 25-600% and increment of 25%. They also show current zoom level on a label, so it's easy to understand how far you are from the real size.

Sprite zoom is 100-800% with increment of 100%.

Now, Views have zoom of 100-300%, if I calculated correctly, with increment of 50% (judging by loop pane multiplying slider value by 0.5). And Inventory pane zoom is 100-800% with increment of 100% again.

I'd suggest to bring these to some consistency maybe, at least with the increments and zoom control displaying current zoom level.

ericoporto:
About the sprite manager, I think it SHOULD be different, I modeled it to Windows Explorer, so I don't really want to change it - it works exactly as I think it should.

But on the other points it's fair and I will look into addressing. In the other PR, can you comment about the sound indicator? 

Also I was thinking about joining PRs - not sure yet, but maybe I will attempt to create a new Zoom control to share among elements.

Editor/AGS.Editor/Panes/InventoryEditor.cs
```diff
@@ -48,22 +77,40 @@ public void RefreshData()
...
-                if ((_item.HotspotX > 0) && (_item.HotspotY > 0))
+                if ((_item.HotspotX > -1) && (_item.HotspotY > -1))
                 {
-                    e.Graphics.DrawLine(Pens.LightGreen, (_item.HotspotX - 2) * IMAGE_SCALE_FACTOR, _item.HotspotY * IMAGE_SCALE_FACTOR, (_item.HotspotX + 2) * IMAGE_SCALE_FACTOR, _item.HotspotY * IMAGE_SCALE_FACTOR);
-                    e.Graphics.DrawLine(Pens.Blue, _item.HotspotX * IMAGE_SCALE_FACTOR, (_item.HotspotY - 2) * IMAGE_SCALE_FACTOR, _item.HotspotX * IMAGE_SCALE_FACTOR, (_item.HotspotY + 2) * IMAGE_SCALE_FACTOR);
+                    Pen penGreen = new Pen(Color.LightGreen, 1);
+                    Pen penBlue = new Pen(Color.Blue, 1);
+
+                    // Create rectangle.
+                    Rectangle rectH = new Rectangle(
```
persn, line 100: If you feel the parameters are not obvious what they are you can specify this with a syntax feature instead of comments

```suggestion
                    Rectangle rectH = new Rectangle(
                        x: (_item.HotspotX - 1) * zoomSlider.Value,
                        y: (_item.HotspotY ) * zoomSlider.Value,
                        width: 3 * zoomSlider.Value,
                        height: zoomSlider.Value);
```

persn (approved):
Looks good

ivan-mogilko:
@ericoporto I recall there were plans to merge in a zoom control in Views too. Is this still planned, or this pr is done and may be merged?

ericoporto:
The PR is not done, I haven't yet figured how to keep same control across the "zooms"/make it more consistent. It works and does what it says, but I don't know yet how to not divert from other zoom types.

ericoporto:
@ivan-mogilko I made it visually more similar. I do think that the inventory zoom must be a bit bigger since a person must be able to select a single pixel in the cursor to be able to mark where is the cursor anchor. This is not a need elsewhere, so other places are fine with a different maximum zoom.

I also am not sure the 100% number makes sense in the VIEW since it's relative to the surface of the view and not the actual sprite.

I tried to make a custom control for the zoom slider, but it did not worked very well, so I kept at the trackbar and separate labels.

ivan-mogilko:
>  I want to try again to make a single winforms zoom control, so I placed it as draft. It will probably look similar in the end but with less copy and pasted code.

That would be great to have a similar control, but if that does not work having at least similarily looking may be fine as the first iteration.

Also, you edited out the original comment, where you say that various panels may use various maximum zoom, and that they zoom different things.; and I tend to agree to that actually. That made me think that we may rather need a zoom control which does not set zoom limits relative to the original image, but is instead based on a relation between a user's screen and image's resolution. That way, for example, the image's pixel may be made same relative size everywhere.

ericoporto:
I placed a new zoom control

image
image

ivan-mogilko:
The slider works; but the editable field is confusing:
* typing any random value will snap it to one of the slider's notches, so it's not clear why typing is allowed at all.
* you can edit % sign... well, it reappears, but it's still weird.
* typing a value outside of the slider's range causes Editor to display the "unhandled exception" window.

ivan-mogilko:
Oh, I found another strange problem: in this branch the transparent "magic pink" color in Views is displayed as black.

ericoporto:
> Oh, I found another strange problem: in this branch the transparent "magic pink" color in Views is displayed as black.

Wait, can you post a picture? It should still be magic pink! It's for me at least...

ivan-mogilko:
Here's the project I was testing with:
[test--360-musicdisplay.zip](https://github.com/adventuregamestudio/ags/files/7875592/test--360-musicdisplay.zip)

it has one view, which has sprite 0 in all frames. In master branch this sprite is shown as pink, but in this branch it becomes black for some reason.

ericoporto:
AAAHH, it's not the magic pink! Your sprite zero is a 1x1 pixel black dot!

ivan-mogilko:
There's a following problem with the Cursor and Inventory items: if zoom is set to max, the image may become larger than the group box surrounding it, and scrollbars appear, even though there's still enough space around these group boxes. This looks pretty weird.

Is there an easy way to resize these group boxes as their contents change?

ericoporto:
I can set the outer thing to dock fill layout, so it will occupy the whole panel. It does look a bit curious if the panel is very large.

selmiak:
This is very cool and I like how it is coming along. Thanks a lot for your efforts eri0o!
But since this is early on let's do it right! What bothered me about the cursor image was how fiddely it is to precisely set the cursor hotspot with the mouse on a really small image. Now setting the zoom level has that level of annoying fiddeling I fear (I haven't tried it yet but the really small increments look like that). So is there any reason we cannot make room for that functional zoom slider to make it easy to use? Like this

149635165-d5e471e2-dcf8-417e-9b17-695f10c6d5c9

is this slider created by the IDE or can we style it? I can do that in PS and make it look good.

can we also have a - and a + button on the left and the right and clicking those jumps from one increment to the next for another nice way of working with it. I can design those too when I know the image dimensions needed :)

ericoporto:
You can type in the place the zoom value is shown to specify a zoom value. You can also use the keyboard arrows to modify the zoom by a single step. I previously had added the minus and plus buttons but thought it was too ugly/noisy, so I removed.

The zoom is a single new custom control, but it's made of the Trackbar Windows Forms component, so it's visual aspect is tied to it for now - it can be resized, of course. 

I think a possible advancement is to add a new control that is the both the zoom control and the panel with the scroll area in a single thing, so it would be possible to cut some more copy pasted code in all zooms - and then pick up the zoom features currently in the Room Editor.

selmiak:
sure, get it out and get more feedback, though dragging the slider bigger could be added already to that release :)

So I read this that you are working on abstracting this and making it possible to style that slider in the future? Once this is possible send me a forum PM or so and I'll send you some nice images :)

PR #1416, opened 2021-09-20
Ags4: Editor UI refresh

### LINKED ISSUE #1077: Color theme don't allow replace or delete color themes
The new color theme option in the editor don't allow replace color themes.

**AGS Version**
3.5.0_24

**To Reproduce**
Import a new color theme with the same name that already exist (example VisualStudioDark). You will need to change the name of the theme, but it will be stacked. This is bad when you are doing a test of the color theme.

**Expected behavior**
Color themes with the same name replaced or option to delete color themes from the list.

AlanDrake:
This is a WIP for improved theme support.
Features:
- simpler theme format with fallback global values - [DOWNLOAD THEMES](https://www.dropbox.com/s/6hlc00l36a4nz88/Themes_v3.zip?dl=0)
- added a few themeable elements
- VS2015 Blue Theme (other variations and original base theme can be changed in preferences)
- implemented Scintilla's ScrollWidthTracking for auto horizontal scrollbar sizing
- small improvements to default theme

The expanded theme code has simplified some areas, and I made it backward compatible-ish (a few renamed styles won't load without throwing errors).
This is not the perfect solution I'd like to have implemented, but it's a little better than we currently have, both for source code and for theme authoring.

Other areas of improvement:
- [x] Theme handling: Button to open the theme folder so we can close #1077 without unnecessary implementations
- [x] The possibility to choose which (dockpanel) theme to apply from the preferences

Abandoned:
- ~VS2015 theme customization: unlike the original VS2005Theme which uses the `Skin` property to colorize the interface, newer ones have a different set of properties~ Couldn't see a way to change theme colors at runtime. DockPanelSuite docs are misleading in mentioning easy ways of making themes through *.vstheme files, too bad it looks like you have to compile them.
  By choosing VS2005 theme, the old style Skin will still be applied.

EDIT: Updated after the new scintilla landed on master/ags4 branches
EDIT: Implemented preference setting to choose DockPanelSuite themes

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -93,6 +93,10 @@
     <Reference Include="WeifenLuo.WinFormsUI.Docking, Version=3.0.6.0, Culture=neutral, PublicKeyToken=5cded1a1a0a7b481, processorArchitecture=MSIL">
       <HintPath>..\..\Solutions\packages\DockPanelSuite.3.0.6\lib\net40\WeifenLuo.WinFormsUI.Docking.dll</HintPath>
     </Reference>
+    <Reference Include="WeifenLuo.WinFormsUI.Docking.ThemeVS2015, Version=3.0.6.0, Culture=neutral, PublicKeyToken=5cded1a1a0a7b481, processorArchitecture=MSIL">
+      <HintPath>..\..\Solutions\packages\DockPanelSuite.ThemeVS2015.3.0.6\lib\net40\WeifenLuo.WinFormsUI.Docking.ThemeVS2015.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AGSEditor.cs" />
--- a/Editor/AGS.Editor/Entities/EditorPreferences.cs
+++ b/Editor/AGS.Editor/Entities/EditorPreferences.cs
@@ -77,6 +77,19 @@ namespace AGS.Editor.Preferences
         Windowed = 2
     }
 
+    [Flags]
+    public enum BaseTheme
+    {
+        [Description("VS2005")]
+        VS2005 = 0,
+        [Description("VS2015BlueTheme")]
+        VS2015BlueTheme = 1,
+        [Description("VS2015DarkTheme")]
+        VS2015DarkTheme = 2,
+        [Description("VS2015LightTheme")]
+        VS2015LightTheme = 3
+    }
+
     public class ColorThemeTypeConverter : BaseListSelectTypeConverter<string, string>
     {
         protected override Dictionary<string, string> GetValueList(ITypeDescriptorContext context)
@@ -762,6 +775,25 @@ namespace AGS.Editor.Preferences
             {
                 if (value < 6) value = 6;
                 this["ScriptTipFontSize"] = value;
+            }
+        }
+
+        [Browsable(true)]
+        [DisplayName("Base Theme")]
+        [Description("Select which base theme the editor should be using.")]
+        [Category("Editor Appearance")]
+        [UserScopedSettingAttribute()]
+        [DefaultSettingValueAttribute("VS2015BlueTheme")]
+        [TypeConverter(typeof(EnumTypeConverter))]
+        public BaseTheme BaseTheme
+        {
+            get
+            {
+                return (BaseTheme)(this["BaseTheme"]);
+            }
+            set
+            {
+                this["BaseTheme"] = value;
             }
         }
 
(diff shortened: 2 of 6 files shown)


ericoporto:
Hey, the Scintilla stuff here should probably be dropped. Also it may be a good idea to add a new repository in the organization for Editor V3 and V4 themes. In particular I think the new theme format is better for newer panels in the editor.

AlanDrake:
Yes, I'll be redoing this soon since we moved on a better version of Scintilla.

AlanDrake:
Alright, and this is up to speed as well.

Although I've based this branch on ags4, the same changes should work fine for ags3. Should we upgrade the theme format there as well?
We could omit using the VS2015 theme so we keep a visual separation betwen 3.x and 4.x, but the rest of the current changes are stuff that both benefits.

I might work on Scintilla multi-cursors in a different PR.

ivan-mogilko:
> I might work on Scintilla multi-cursors in a different PR.

I think it's a good idea to split up Themes and Scintilla updates, as they are unrelated.

AlanDrake:
I refreshed the branch and added a few new commits

AlanDrake:
Alright, I don't think I'm going to do anything more with this PR.

ericoporto:
I think most of the code you commented could just be deleted, specially those that have been replaced by new helper methods. I think it's alright to do it in a new cleanup commit too.

ivan-mogilko:
> I hampered with one of your changes, specifically the one that only upscales a small sprite to 2x max. Problem was that to avoid the ugly blurry upscale we must use all the space for at least one side. Also, the outline, being overimposed at lower resolutions could cover too much and looked bad when blurred.
It could be reimplemented in a different way, but IMO limiting to just 2x may not be enough in some cases.

By the way, this is a problem in 3.6.0 too.

AlanDrake:
@ivan-mogilko all these changes are an improvement to 3.6 as well.
I could retarget to master if you want, maybe excluding the dockpanel theme selector. Dunno.

ivan-mogilko:
Necessary fixes may be backported to 3.6.0 separately. But we need a list of those that are applicable.

AlanDrake:
You can treat the commits as a list of features, I'd just omit the additional dockpanel themes. Maybe I'll rebase again to squash a few thing to reduce the noise.

I can make a PR for master real quick, it's all going to be merged back into ags4 anyway.
I've been testing these changes since before the issue was opened, so they shouldn't bring in any issue. 

As for theme format changes that may cause a couple of colors to be ignored while loading old themes:
`script-editor/text-editor/selected`
was expanded into
- script-editor/text-editor/selected/background
- script-editor/text-editor/selected/foreground

`script-editor/text-editor/caret`
was expanded into
- script-editor/text-editor/caret/caret-fore
- script-editor/text-editor/caret/caret-line-back
- script-editor/text-editor/caret/caret-line-back-alpha  (int value)

ericoporto:
I read the Layout.xml available in AppData/Local/AGS and trying to make sense of it I get that the dockpanel is about proportions of things and may not have an absolute height. You can see there things like the `CallStackPanel` and `FindResultsPanel`, that are docked in panes, which are nested in dockwindows, which have proportions. But I could not make sense of it from reading the structure there, just commenting in case it makes sense to you.

AlanDrake:
The xml layout is not relevant, there should be some property somewhere I can access to figure out how many things are docked in the same panel.
That's all I would need to update the bottom padding appropriately. Although, I also wonder why on earth it isn't shrinking automatically since it's not like the scrollable component height has any issue changing while resizing the area.

ivan-mogilko:
I have doubts about the replacement of "Import theme" with "Open folder", whether that really is making theme management easier, as this makes user to manually copy files around their computer. Normally, I'd expect a software to provide some kind of a list where you can view, add and remove themes with UI; and "open location" is an extra option rather than a replacement for this.

AlanDrake:
I don't know, opening the folder solves all problems, delete, rename, copy, and it comes with zero extra work.
Even if we fix the import button, the user will still be forced to pinpoint the folder to rename or delete themes, but manually, wasting time searching the forums to figure out the path.
Speaking of FileSystemWatcher, I could refresh the list on focus instead. The way we've handled watchers in 4.x is also incomplete, as we aren't tracking deletes/renames/move operations. I need to open a ticket as a reminder.

Anyway, I'll open a PR for master omitting that commit and the other unnecessary stuff.

ivan-mogilko:
> Even if we fix the import button, the user will still be forced to pinpoint the folder to rename or delete themes, but manually, wasting time searching the forums to figure out the path.

Why not add buttons for these actions?

AlanDrake:
Temporarily returned to draft as I'll need to address a few conflicts after we merge the ags3 version.

AlanDrake:
Now that the bulk of changes have been merged to master and back to ags4, the only change here is the new selectable dockpanel themes.
If nobody has any objection I will merge it this weekend.

PR #1417, opened 2021-09-21
Ci release 3.5.1 fixes

ericoporto:
Windows related changes:
- Editor: Native use default Zp alignment (it fails to build the editor in release otherwise)
- CirrusCI: add path to Win SDK or it fails to find `rc.exe` (required to build the Editor)
- Windows Docker: DirectXSDk from MS is not available, use directx sdk from archive.org
- Windows Docker: VS2015 from MS is not availabe, use base image with preinstalled VS2015
If these changes are a problem, we can use the backup Windows Docker image from [here](https://hub.docker.com/repository/docker/ericoporto/ags35x_allegro4) instead.

Additional changes:
- Linux Docker: use XIPH from GitHub for older tls support (Debian Jessie has no newer tls)
- Linux Docker: autoconf necessary for theora build from GitHub sources
- Android Docker: apache ant old version is not available, upgrade it
- Android Docker: use release-3.5.1 branch instead of master

please wait CI finish building to verify these changes are correct.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -184,6 +184,7 @@ build_editor_task:
     copy C:\Lib\irrKlang\*.dll Editor\References\ &&
     cd Solutions &&
     cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\8.0\Lib\Win8\um\x86;!LIB!" &&
+    set "PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;!PATH!" &&
     msbuild AGS.Editor.Full.sln /p:PlatformToolset=v140 /p:Configuration=%BUILD_CONFIG% /p:Platform="Mixed Platforms" /maxcpucount /nologo"
   ags_editor_pdb_artifacts:
     path: Solutions/.build/*/*.pdb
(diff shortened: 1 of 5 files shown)


morganwillcock:
I'm not 100% sure that the CI builds verified this is working correctly.

I didn't really know what the relevance of TLS versions was in order understand the need to switch to downloading from GitHub, but this command doesn't look correct:
```
./autogen.sh --disable-encode --disable-examples --disable-oggtest --prefix=/opt
```
The arguments are meant to be passed to `configure` and script above doesn't process arguments. I don't think this will build correctly because the configuration will be skipped.

morganwillcock:
Actually I looked at the wrong autogen.sh and it does invoke configure in this instance, so don't worry. Although I'm still interested in how TLS is relevant.

ericoporto:
Ah, it's the headers on xiph servers, here is the issue when I realized the change:

https://github.com/xiph/theora/issues/16#issuecomment-850088490.

The problem is Jessie has a version of OpenSSL that is just too old. Since it's not even the oldoldstable anymore, we should look into switching to a less old version eventually - on the master branch, the benefit would be Wayland with SDL2. :)

PR #1420, opened 2021-09-27
Engine: fixed panning SDL implementation

AlanDrake:
Uses OpenAL spatial audio, which only applies to mono-channel clips.

But what I'd rather do, at least for AGS 3.6 given the backward compat, would be getting hang of the two channels of the stereo clip, and manually set the gain in order to reimplement the original behavior.
For this we  supposedly need to call `alGenSources()` with 2 sources, if I understand the documentation correctly.

### DIFF
--- a/Engine/media/audio/audio_core.cpp
+++ b/Engine/media/audio/audio_core.cpp
@@ -19,6 +19,7 @@
 // safer slot look ups (with gen id)
 // generate/load mod/midi offsets
 
+#include <math.h>
 #include "media/audio/audio_core.h"
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@AlanDrake could you give more insight in the problem, because right now it's impossible to understand why is this change necessary from the existing description?

AlanDrake:
It's simple. if we want to pan stereo clips we need to to do it manually, because OpenAL only does 3D spatial audio with mono clips, everything else is played as-is.
In this PR I fixed what prevented the current spatial audio implementation from working, so panning finally works (with mono clips)

PR #1421, opened 2021-09-29
Android: expand file-related implementations based on AAssetManager

ivan-mogilko:
This adds more Android-specific implementations of the file system functions based on Android's AAssetManager; which is a kind of virtual filesystem from the engine's perspective.
This PR covers testing that file (aasset) exists, and searching for files (aassets) inside the pseudo-directory of an android package.
Android port will first search in the real filesystem as usual, then - using AAssetManager.
This hopefully fixes the problem of engine ignoring game data packed as individual AAssets in APK (translation files, etc).

* Better handling of our engine's AAssetManager reference: query it once, keep same ref, then dispose upon engine shutdown.
* AAsset variants of File::IsFile() and File::GetFileSize() functions.
* FindFile class's implementation that also searches through AAssets.

### DIFF
--- a/Common/util/android_file.cpp
+++ b/Common/util/android_file.cpp
@@ -69,6 +69,27 @@ AAssetManager* GetAAssetManager()
     return gAAssetManager;
 }
 
+bool GetAAssetExists(const String &filename)
+{
+    AAssetManager* mgr = GetAAssetManager();
+    // TODO: find out if it's acceptable to open/close asset to only test its existance;
+    // the alternative is to use AAssetManager_openDir and read asset list using AAssetDir_getNextFileName
+    AAsset *asset = AAssetManager_open(mgr, filename.GetCStr(), AASSET_MODE_UNKNOWN);
+    if (!asset) return false;
+    AAsset_close(asset);
+    return true;
+}
+
+soff_t GetAAssetSize(const String &filename)
+{
+    AAssetManager* mgr = GetAAssetManager();
+    AAsset *asset = AAssetManager_open(mgr, filename.GetCStr(), AASSET_MODE_UNKNOWN);
+    if (!asset) return -1;
+    soff_t len = AAsset_getLength64(asset);
+    AAsset_close(asset);
+    return len;
+}
+
 } // namespace Common
 } // namespace AGS
 #endif // AGS_PLATFORM_OS_ANDROID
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
@ericoporto would be nice if you check this out.

ericoporto:
It all looks fine. I haven't had time to run the code yet, but I read the changes. Just adding a note below in reference of my previous code.

I was getting the AAssetManager everytime because of the cryptic text in the [docs](https://developer.android.com/ndk/reference/group/asset#group___asset_1gadfd6537af41577735bcaee52120127f4):

> Note that the caller is responsible for obtaining and holding a VM reference to the jobject to prevent its being garbage collected while the native object is in use.

So I did the `env->NewGlobalRef` stuff which I thought was enough (about the Garbage Collection) and then also let the thing happen always just in case there was some activity lifecycle thing... 

Anyway, point being, I think the way you did is actually the correct way to handle that, so that is definitely an improvement.

ericoporto:
Just to be explicit: let's merge :)

ivan-mogilko:
Thank you; yet I'd really like to hear confirmation from the user(s) that this works, because I myself did not have an opportunity to test this on real device yet.

Perhaps we may provide native libs from this pr for the to try?

ericoporto:
Uhm, it's weird, have you tried this PR? You are not calling `InitAndroidFile` before using it, so it will fail the assert in the below function

```cpp
AAssetManager* GetAAssetManager()
{
    assert(gAAssetManager);
    return gAAssetManager;
}
```

ivan-mogilko:
Is this called from `Java_uk_co_adventuregamestudio_runtime_PreferencesActivity_readConfigFile`?

Perhaps InitAndroidFile could be called from there as well; alternatively GetAAssetManager may init its reference when it's called for the first time.

ericoporto:
> alternatively GetAAssetManager may init its reference when it's called for the first time.

I really like this approach, it only initializes it when needed!

Common/util/android_file.cpp
```diff
@@ -0,0 +1,74 @@
...
+        JNIEnv* env = (JNIEnv*)SDL_AndroidGetJNIEnv();
+        env->DeleteGlobalRef(gAAssetManagerRef);
+        gAAssetManagerRef = nullptr;
+        gAAssetManager = nullptr;
+    }
+}
+
+AAssetManager* GetAAssetManager()
+{
+    if (!gAssetManager) { InitAndroidFile(); }
```
ericoporto, line 67: there's a typo here!

ericoporto:
all good!

PR #1428, opened 2021-10-07
3.6.0: enhanced gfx mode config

ivan-mogilko:
This is another attempt at improving the way graphics mode is defined in user config. The existing one was made for AGS 3.4.0, and while it allows to freely setup (almost) any mode, there are annoying issues:

1) It's overloaded with config options; there are several options that have to be set, and which combination determines gfx mode: `screen_def`, `screen_width`, `screen_height`, less used `match_device_ratio`, and if you want to get window of xN size of the game then you also have to put proper number in the `game_scale_win` option.
2) It allows for inconsistent or redundant option combinations, for example if `screen_def` is `scaling` then `screen_width` and `screen_height` have to be ignored. If `screen_def` is `explicit` then it uses `screen_width` and `screen_height`, but if at the same time `game_scale_*` option is set to the scaling number, then the game may end up shown in a rectangle of smaller size inside the larger window.
3) Engine now supports toggle between windowed and fullscreen modes, but the config only allows to set up the initial one, and the other will be deduced using engine defaults, which is not necessarily what user wants. 

---

The new config style is meant to fix these issues. There are three changes:
1) The graphic mode is now defined by a *single* option that contains all possible variants in one string.
2) There are two options, one for windowed and another for fullscreen mode, called `window` and `fullscreen` respectively. Both are read by the engine and used as defaults when the user toggles between the modes at runtime.
3) `game_scale_*` options now only define the scaling *style* (`round`, `stretch` and `proportional`), but no longer define the scaling factor - that is derived from the window size.

In [graphics] section fullscreen mode is defined by the option "fullscreen" while the windowed mode is defined by the option "window". These option may have following values:

* WxH - explicit window size (e.g. "1280x720");
* xS - integer game scaling factor (e.g. "x4");
* default - 
   * in fullscreen mode that uses current desktop's resolution;
   * in windowed mode that tries to create largest possible window (for the current desktop size), while keeping game scaling style (max stretch, proportional stretch or max integer multiplier).

The example of new config:
<pre>
[graphics]
fullscreen=1280x720
window=x2
game_scale_fs=proportional
game_scale_win=round
</pre>

---

TODO:
Note that I called the windowed mode option "window", but that's mostly for backward compatibility with old configs which have "windowed" defining initial mode, that is "windowed=1/0".

But there's also a need to dinstinguish "desktop fullscreen" (a borderless window covering whole desktop) and "exclusive fullscreen" (for systems that support one), so that old option no longer suits us either. This is going to be changed next.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -95,6 +95,7 @@ namespace AGS.Editor
          * 3.6.0          - Settings.CustomDataDir;
          *                - Font.AutoOutlineStyle, AutoOutlineThickness;
          *                - Character.IdleDelay
+         *                - RuntimeSetup.FullscreenDesktop
         */
         public const int    LATEST_XML_VERSION_INDEX = 3060000;
         /*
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
That is true.

rofl0r:
i see you've added some code to parse legacy config, which made me think about the plausible scenario that a user might want to use ags 3.5.x and 3.6.x in parallel (for example one engine version might have some issue with a specific game). ags 3.5. will probably not be happy if it encounters config in new format. would it make sense to give a new filename to new-style config ? such as acsetup2.cfg

ivan-mogilko:
> ags 3.5. will probably not be happy if it encounters config in new format

I'm not reusing previously existing options, and iirc usually tried to keep it this way.

ivan-mogilko:
I have one irritating issue left here; since I added "native" value which defines game's resolution, I've been thinking that it could make sense to also use special value to define current desktop resolution for window or exclusive fullscreen mode. Previously I used "desktop" to define the "FULLSCREEN_DESKTOP" mode. Now I wonder if there may be another good name for it, while "desktop" could be used to create desktop-sized exclusive mode or max fitting windowed mode.

ivan-mogilko:
So, I did this: "desktop" value is used to define the window resolution, similar to "native", and "fullscreen_window" means the borderless window covering whole desktop. I'll leave it at that for now, but since 3.6.0 is still wip, this may change if nicer alternatives are found.

Note: had to rebase because there were bugs in the master.

Issue #1430, opened 2021-10-13
AGS4 / New compiler: `switch` can result in endless loop


fernewelten:
**Describe the bug**
The code below compiles into an endless loop.

**AGS Version**

AGS 4.0, commit 9719b05
Build https://cirrus-ci.com/build/4889336761024512

**Game**
n/a

**To Reproduce**
- Install the following code for the After FadeIn event of some room.
- Execute the room.
- Note that the Engine enters an endless loop.

```
function room_AfterFadeIn()
{
    int test = 2;
    switch(test)
    {
    case 1:
    {
        break;
    }
    
    case 2:
    {       
    }
    } // switch
}
```
**Expected behavior**
Code execution falls through the `switch`.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - Windows 10, German edition

**Smartphone (please complete the following information):**
 - n/a
 - 
**Additional context**
(none)

PR #1433, opened 2021-10-16
New compiler: Dead end tracking

fernewelten:
This also addresses #1430, see below near the end of the posting.

This PR track whether a point in code is guaranteed to be unreachable because `return`, `continue`, or `break` precedes it. Let's call such points “execution dead ends”.

The new compiler uses this information in several useful ways:

___Forgotten `break;` in `switch()` statements:___

It's __very easy__ to code a switch statement and forget the `break;` at the end of some cases. In this case, code execution will “fall through” to the following case.  The semantics allow this, but it's usually not what the programmer intends.
Typical code:
```
switch (weapon_status)
{
case kDamaged: player.Say("The weapon has damage"); break;
case kEnchanted:  player.Say("The weapon has a faint blue shine");// <-- problem here
default: player.Say("I see nothing unusual"); 
}
```
The compiler will try to detect whether the end of a case is an execution dead end and warn (!) the user when it can't determine that. So the compiler will flag the second case in the example above. (This is not an error; the compiler will continue compiling.)

Sequences of `case` labels won't yield any warning, of course:
```
switch (num)
   …  
case 2:
case 3:
case 4:
    return kSmallTeam;
    …
}
```

___Statements in execution dead ends___

Whenever the new compiler detects this, it'll warn that the code can't be reached.
Typical code:
```
while (true)
{
    …
    if (…)
        break;
    else
        continue;
    bool flag = false; // <-- unreachable
}
```

___Missing `return` statement___

Whenever a function returns another type than an integer type, then the end of its body must be an execution dead end. Otherwise, the function will default to returning a `0` which doesn't match the return type. When the compiler suspects that such a body end is reachable, it'll warn.

Typical code:
```
float Distance(Point *a, Point *b)
{
    if (a != null && b != null)
    {
        float diff_x = a.x - b.x;
        float diff_y = a.y - b.y;
        return Maths.Sqrt(IntToFloat(diff_x * diff_x - diff_y * diff_y));
    }
    // <-- This point can be reached!
}
```

___Code optimizations___

1. If the end of a function body is guaranteed to be an execution dead end, then the compiler needn't emit a default return. This is very common, a lot of functions end with
```
    return something;
}
```
(As a case in point, oodles of Googletests were generated more efficiently because of this dead code detection, this is the reason for the many changes in the files Compiler/test2/cc_bytecode_test_0.cpp and  Compiler/test2/cc_bytecode_test_1.cpp) . 

2. If the last case of a switch is guaranteed to be an execution dead end, then the compiler needn't protect the jump table with a preceding jump.

This also addresses #1430 : When I analysed that bug, I determined as the root cause that my compiler had done a sloppy job of finding out whether the last `switch` case ended on a jump-out, and got it wrong. Then the last `switch` case and the jump table catastrophically combined into an endless loop. 
To repair that properly, I decided on implementing robust execution dead end tracking.

### DIFF
--- a/Compiler/script2/cc_compiledscript.h
+++ b/Compiler/script2/cc_compiledscript.h
@@ -183,6 +183,8 @@ public:
     // Set the destination to the location given; default to current location in code
     void Set(CodeLoc cl = -1);
     inline CodeLoc Get() const { return _dest; }
+    // Whether the destination has been set
+    inline bool IsSet() const { return _dest != -1; }
     // Write a jump to the code location that I represent
     void WriteJump(CodeCell jump_op, size_t cur_line);
 };
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -299,6 +299,7 @@ AGS::SymbolTable::SymbolTable()
     AddKeyword(kKW_Enum, "enum");
     AddKeyword(kKW_Export, "export");
     AddKeyword(kKW_Extends, "extends");
+    AddKeyword(kKW_FallThrough, "fallthrough");
     AddKeyword(kKW_For, "for");
     AddKeyword(kKW_If, "if");
     AddKeyword(kKW_ImportStd, "import");
--- a/Compiler/script2/cc_symboltable.h
+++ b/Compiler/script2/cc_symboltable.h
@@ -177,6 +177,7 @@ enum Predefined : Symbol
     kKW_Enum,           // "enum"
     kKW_Export,         // "export"
     kKW_Extends,        // "extends"
+    kKW_FallThrough,    // "fallthrough"
     kKW_For,            // "for"
     kKW_If,             // "if"
     kKW_ImportStd,      // "import"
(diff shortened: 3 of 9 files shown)


ivan-mogilko:
I'm concerned about this example:

```
switch (weapon_status)
{
case kDamaged: player.Say("The weapon has damage"); break;
case kEnchanted:  player.Say("The weapon has a faint blue shine");// <-- problem here
default: player.Say("I see nothing unusual"); 
}
```

This is a mistake sometimes, but may be a wanted behavior, in which case warnings will become distracting. Could there be any syntax methods to tell the compiler that this "fallthrough" is intended by the coder?

fernewelten:
> Could there be any syntax methods to tell the compiler that this "fallthrough" is intended by the coder?

Oh yes, I forgot. There is:
If you code,
``´
switch (weapon_status)
{
case kDamaged: 
    player.Say("The weapon has damage"); 
    break;
case kEnchanted: 
    player.Say("The weapon has a faint blue shine");
    fallthrough;
default: 
    player.Say("I see nothing unusual"); 
}
```
... then the compiler will __not__ warn.

fernewelten:
Entered the function as shown below and hit F7:

image

PR #1449, opened 2021-11-14
AGS4 / New compiler: Attributes with extender syntax

fernewelten:
This PR adds attributes with extender syntax to the language. So it is possible to add, e.g., a `Weight` attribute to a `Character` even though the struct `Character` has been pre-defined.  This would result in the functions `Character::get_Weight()` and `Character::set_Weight()` to be declared, and the coder must then provide code for these functions, too, probably again using the extender syntax.
Extender attributes can be `readonly`, in this case the setter function isn't declared. 

Declare the attribute, in a header file:
```
attribute int Weight(this Character *);
```

Define the functions, in a code file:
```
#define WEIGHTPROP "Weight"
int get_Weight(this Character *)
{
    return this.GetProperty(WEIGHTPROP);
}
void set_Weight(this Character *, int weight)
{
    this.SetProperty(WEIGHTPROP, weight);
}
```

Use the attribute, e.g., in an event:
```
function hBench_Interact()
{
    if (player.Weight > 100)
        player.Say("I'd rather not sit on this frail bench; I'd break it!");
}
```

There are static attributes, too:
```
attribute int Foo(static Character);
…
if (Character.Foo > 10) return;
```

Indexed attributes:
```
attribute int Weapon[](this Character *);
attribute int Bar[](static Character);
```

### DIFF
--- a/Compiler/script2/cc_compiledscript.h
+++ b/Compiler/script2/cc_compiledscript.h
@@ -137,6 +137,7 @@ public:
     void Restore();
 
     inline bool IsEmpty() const { return _scrip.codesize <= static_cast<int>(_rememberedCodeLocation); }
+    inline CodeLoc CodeLocation() const { return _rememberedCodeLocation; }
 };
 
 // A storage for parameters of forward jumps. When at some later time,
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -5113,7 +5113,7 @@ AGS::ErrorType AGS::Parser::ParseVardecl_CheckAndStashOldDefn(Symbol var_name)
     return kERR_None;
 }
 
-AGS::ErrorType AGS::Parser::ParseVardecl(Symbol var_name, Vartype vartype, ScopeType scope_type, TypeQualifierSet tqs)
+AGS::ErrorType AGS::Parser::ParseVardecl(TypeQualifierSet tqs, Vartype vartype, Symbol var_name, ScopeType scope_type)
 {
     ErrorType retval = ParseVardecl_CheckIllegalCombis(vartype, scope_type);
     if (retval < 0) return retval;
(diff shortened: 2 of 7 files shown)


ivan-mogilko:
I tested this and it seems working. There are couple of notes:

1) When I declare a readonly attribute and try to set it, the compiler displays an error:

`!Attribute function 'set_Weight' not found in struct 'Character'`

Firstly note strange "!" at the beginning, but I think more importantly the error message may be quite confusing to users.

For the reference, old compiler reports this when trying to set readonly attributes:

`property 'System::AudioChannelCount' is read-only`



2) I don't know if this matters, but it's possible to declare a readonly externder attribute, and then separately declare an extender setter for that attribute:

<pre>
readonly attribute int Weight(this Character *);

void set_Weight(this Character *, int weight)
{
    this.SetProperty("Weight", weight);
}
</pre>

fernewelten:
>[I]t's possible to declare a readonly externder attribute, and then separately declare an extender setter for that attribute:

So far, this would work; 
When you do `player.set_Weight(17);` then this is just a random function call for the compiler. It checks whether the function is in the symbol table at this point, but it doesn't check whether the function begins with `set_` and there is an attribute called `Weight` and that attribute is readonly. 

I can put this in, what do you think?

fernewelten:
> Firstly note strange "!" at the beginning, 

That seems to be part of a bigger problem; it should be addressed, but perhaps not in this PR:

1. The file `cc_error.h`  declares the function `extern int ccError` and only has the explanation `// error reporting`. But when you look at the corresponding code, there's more to that. It seems that __user__ errors need to begin with a bang `!` whilst internal errors do not begin with a bang. This is undocumented in the header file.
cc_error.cpp:34
```
void cc_error(const char *descr, ...)
{
    ccErrorIsUserError = false;
    if (descr[0] == '!')
    {
        ccErrorIsUserError = true;
        descr++;
    }
```
2. This feature doesn't seem to be used in practice. When we call 'cc_error()' with a message that does __not__ start with a bang `!`, nothing special happens. There's no indication that this is a system error that's no fault of the user. OTOH, when the message does start with a bang, the bang is simply displayed.
3. My compiler gets this wrong systematically. It starts all __internal__ errors __with__ a bang and all __user__ errors __without__ a bang instead of the other way round.

ivan-mogilko:
Yes I think this should work, since trying to assign a readonly attribute will get an error regardless now.

---

In regards to `cc_error`, having an opportunity to distinguish error types is useful, even though it seems that these are only checked within the engine currently.

But that's not the ideal way to do so either. I'm thinking that maybe it could be refactored into e.g. having a struct with an error message and enum for type later. But as a start I may suggest to have two inlined helper methods, e.g. `cc_error_user` and `cc_error_internal` that both just call `cc_error` for now; if we use them inside compiler(s) this way we could have a better distinction between these error types. And later add a real implementation that passes error type id of some kind.

Issue #1441, opened 2021-10-26
Attempt to improve Font's Height handling in games


ivan-mogilko:
So I've been researching problems with TTF fonts display in ags recently (https://github.com/adventuregamestudio/ags/issues/888), and think that understand the issue much better now. I've pushed some changes that may help maintain backward compatible arrangement of the texts in the old games and at the same time add potential to further improve the situation in the new games. But I'm still not fully certain of what would be expected and convenient solution from the user's perspective; and that is why I am opening this ticket.

## Problem summary

AGS historically has hacks to the TTF font library, that have two primary consequences.

### Loaded font sizes

The way TTF fonts work is that they, as vector fonts, cannot translate 1:1 to any random pixel size. There's even a note about that in, for example, FreeFont documentation ([link](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_set_pixel_sizes)): *You should not rely on the resulting glyphs matching or being constrained to this pixel size*. I think this is because they can't scale to any random pixel size without loosing proportions.
What this means is that when the font library tries to init a TTF using "font size" as a request, the resulting size could be identical, less, or larger than requested.

Alfont library, which is a wrapper over FreeType that AGS uses, normally does following: if the font size does not match the requested one, then it will try to lower/increase the request until finding the closest possible actual size. You may observe this behavior [in the library code](https://github.com/adventuregamestudio/ags/blob/8a735b4/Common/libsrc/alfont-2.0.9/alfont.c#L431).

However, AGS contains [a hack](https://github.com/adventuregamestudio/ags/blob/8a735b4/Common/libsrc/alfont-2.0.9/alfont.c#L428) to this library that instead makes it choose the very first result and keep it. This leads to some fonts being notably larger than requested. E.g. a font imported with size 52 may in practice end up being 62 pixels high - that's about 1/5 more!

### Reported font sizes

To complicate the above, there was another hack that saved not the real font's size, but the requested font size. Which leads to the engine using this "nominal" requested height in any text or UI arrangement, and reports this height in script too (with calls to GetFontHeight, GetTextHeight, etc). This means that the fonts are not guaranteed to fit in precalculated borders, and game developers cannot programmatically learn the real height of the font either.

One known consequence is infamous issue of some fonts getting their bottom pixels "cut off", because the engine cannot correctly resize the texture it's drawing them on.

## Considering solutions

So, the question is, where do we want to go from here.

[...]

PR #1450, opened 2021-11-14
Editor: allow to import TTF by searching for a point size matching pixel height

ivan-mogilko:
For #1441.

As per @fernewelten 's suggestion, this gives user a choice when importing a TTF font in the project:
a) provide an explicit point size value;
b) provide a pixel height and ask to find the point size that results in the closest match to that height.

New dialog looks like this:
importttfdialog


ivan-mogilko:
Okay, I will merge this and see how it goes; after all this does not impose anything on data format, and simply is a different input dialog.

PR #1453, opened 2021-11-20
Support multiple Speech VOXes for game translations

### LINKED ISSUE #686: Expanding the use of the speech voice files (naming limits and translations)
Current situation with speech voice-over is:
* only 1 speech.vox pack is supported and read from;
* there's no automatic solution to adjust played voice clips depending on selected translation;
* voice clip filenames are restricted to NNNNXXXX.ext format, where NNNN is four first letters of character's script name and XXXX is a number 1-9999. I believe this name format may be a legacy of DOS support (8.3 filenames).

The name rule may cause conflicts between characters with similar script names and usually forces gamedev to edit script names to make first 4 letters unique.

Limit of 9999 clips per character may be less critical, but sometimes being ran into. (In such case they usually use dummy characters to increase number of slots. AFAIK they may also use this trick to merge different languages into one VOX.)

My suggestions are:

**1. Regarding multiple translations**

1.1 Create multiple VOX packages.

Game compiler will treat any subfolder inside Speech folder as a separate VOX content. Resulting VOX will be called "sp_N.vox" where 'N' is a subfolder's name.
Files from Speech folder itself go into "speech.vox" like they do now.

1.2 Support switching VOX package at runtime

There are two alternatives here: automatic and manual.

Automatic solution looks for the "sp_N.vox" with 'N' being equal to the current translation name. If "sp_N.vox" is not found, then looks for default "speech.vox".

Manual solution would require introducing new script commands to set translation explicitly for voice-over. This would also allow to have text in one language and voice-over in another language.

These two solutions may co-exist.

**2. Remove old naming rule**

Instead voice clip format should be "NX.ext", where N is full character's script name and X any number, let's say, fit into 32-bit integer.

@messengerbag has suggested an alternative to this: introduce new Character's property to define voice-name, which is equal to its script name by default.

ivan-mogilko:
Resolves #686 (except the part with expanded clip names, which I think may be moved to a separate ticket).

This adds a support for game having multiple voice-over packages, and being able to switch between them at runtime. The primary goal is to let game developers have the game translated with voice-overs in various languages.

Implementation details follow.

---

### Engine

Engine supports loading a different speech vox at runtime.

By default engine loads a `speech.vox`, if one exists.

When ordered to change to an alternate vox of a name "[name]", engine will look for a vox called `sp_[name].vox`. If this vox is found then it will be loaded. If not, engine falls back to the default `speech.vox`.

The assets in alternate voxes must have names `[name]/filename.ext` where `[name]` is the same "tag name" of a vox, as noted above. This naming rule goal is to mimic the disk structure, where default vox files are in the root and all alternate ones are inside subdirectories. This allows to also read these assets from disk if necessary without any extra hacks to path search. This also in a way corresponds to how the custom user files are named (see #1227, #1372).

Everything else works as usual, regardless of which vox is loaded.

---

### Script API

Implemented 2 new API items:

`bool Game.ChangeSpeechVox(const string name);` - changes the speech vox to the vox of the given name (the actual filename will be `sp_name.vox`, as explained above. An empty string may be passed to switch to the default speech.vox.
`readonly String Game.SpeechVoxFilename;` - gets the "name" of the currently loaded vox (without "sp_" prefix and no extension), an empty string for a default speech.vox.

---

### Editor

Editor now packs alternate speech voxes, by scanning the top-level directories inside the "Speech" folder. For each top-level subdirectory It creates a `sp_[name].vox` where `[name]` is the lowercase subdir name.

For example, if you have this in your game project:
* Speech
  * Francais
  * MyLang

The editor will create following voxes:
- speech.vox (with top-level files from the root "Speech" folder);
- sp_francais.vox (with top-level files from "Speech/Francais");
- sp_mylang.vox (with top-level files from "Speech/MyLang");

The assets in the *alternate* voxes will be called according to the aforementioned naming rules, that is `[dirname]/filename.ext`.

### DIFF
--- a/Common/core/assetmanager.h
+++ b/Common/core/assetmanager.h
@@ -83,7 +83,8 @@ public:
 
     // Add library location to the list of asset locations
     AssetError   AddLibrary(const String &path, const AssetLibInfo **lib = nullptr);
-    // Add library location, specifying comma-separated list of filters
+    // Add library location, specifying comma-separated list of filters;
+    // if library was already added before, this method will overwrite the filters only
     AssetError   AddLibrary(const String &path, const String &filters, const AssetLibInfo **lib = nullptr);
     // Remove library location from the list of asset locations
     void         RemoveLibrary(const String &path);
--- a/Editor/AGS.Editor/Components/SpeechComponent.cs
+++ b/Editor/AGS.Editor/Components/SpeechComponent.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Windows.Forms;
@@ -50,6 +51,12 @@ namespace AGS.Editor.Components
         private void _agsEditor_ExtraCompilationStep(CompileMessages errors)
         {
             MakeOneLipSyncDat(SPEECH_DIRECTORY, LIP_SYNC_DATA_OUTPUT, errors);
+            // Also try compile corresponding lipsync dat for each top-level subfolder
+            // inside the main Speech folder.
+            foreach (string dir in Directory.GetDirectories(SPEECH_DIRECTORY))
+            {
+                MakeOneLipSyncDat(dir, LIP_SYNC_DATA_OUTPUT, errors);
+            }
         }
 
         private void _agsEditor_ExtraOutputCreationStep(bool miniExeForDebug)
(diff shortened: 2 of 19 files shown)


ivan-mogilko:
Updated the PR, removed an automatic switch on changing translations and added new functions to script api: `Game.ChangeVoicePack` and `Game.VoicePakFilename`. See the first post for more details.

ivan-mogilko:
Ended up with `Game.ChangeSpeechVox` and `Game.SpeechVoxFilename`, because noticed there's already `IsSpeechVoxAvailable`, so decided to use same term.

PR #1460, opened 2021-12-05
3.5.1 hotfix: don't precreate common engine/game directories

### LINKED ISSUE #1459: Using alternate --conf, --shared-data-dir and --user-data-dir still creates (a blank) %ProgramData%\Adventure Game Studio\
**Describe the bug**
Using alternate `--conf`, `--shared-data-dir ` and `--user-data-dir` ([see documentation](https://github.com/adventuregamestudio/ags/blob/master/OPTIONS.md#configuration-file-options)) still creates (a blank) `%ProgramData%\Adventure Game Studio\` which is both futile and hurts the portability purpose of using these parameters in the first place.

**AGS Version**
3.5.1 patch 4

**Game**
All.

**To Reproduce**
1. Start any game using these parameters (make sure all of them use alternate locations)
2. Exit the game
3. See a blank folder in the aforementioned path

**Expected behavior**
Don't create this folder while using these alternate parameters, especially when it just means creating a blank folder.

 - OS: Windows
 - Version 10

ivan-mogilko:
Fixes #1459.

Opened the PR mostly to ensure it builds for all platforms and to provide a test build from the build server.
I will be testing this myself too, as there's a number of scenarios to check.

The problem was that platform drivers in AGS were unconditionally creating directories for game saves etc files at the standard paths. This changes them to not precreate any dirs, but only return the paths, letting engine create directories only if necessary. Thus not only standard dirs will not be created if user (player) provided their own in config, but they are not be created unless engine actually writes something. This complements the previous changes to the 3.5.1 engine.

For technical details, the paths are stored as a pair of "base dir"/"full dir", where "base dir" is considered not engine's business and not touched, only the part after "base dir" will be created when necessary.

### DIFF
--- a/Common/util/directory.cpp
+++ b/Common/util/directory.cpp
@@ -38,7 +38,8 @@ bool CreateAllDirectories(const String &parent, const String &path)
         return false;
 
     String fixup_parent = Path::MakeTrailingSlash(parent);
-    String sub_path = Path::MakeRelativePath(fixup_parent, path);
+    String fixup_child = Path::MakeTrailingSlash(path);
+    String sub_path = Path::MakeRelativePath(fixup_parent, fixup_child);
     String make_path = parent;
     std::vector<String> dirs = Path::Split(sub_path);
     for (const String &dir : dirs)
--- a/Engine/ac/file.cpp
+++ b/Engine/ac/file.cpp
@@ -284,33 +284,39 @@ FSLocation GetGameUserConfigDir()
     return dir.Concat(game.saveGameFolderName);
 }
 
-// A helper function that deduces a data directory either using default system location,
-// or user option from config. In case of a default location a path is appended with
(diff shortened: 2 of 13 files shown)


ivan-mogilko:
Tested on all platforms (except iOS).

ericoporto:
Hey, I am having a bit of trouble with the Android port. The code looks great, but the call to GetAllUsersDataDirectory/GetUserSavedgamesDirectory only happens after the reset saveslot is created, so I am getting an error with "Unable to write in the savegame directory.". This happens when using the single game launcher because even though the base directory is correct, it hasn't been applied yet, so instead it tries to write in the asset package and fails - at least I think this is what I am seeing, I may be a bit rusty, so not sure...

ivan-mogilko:
@ericoporto I will look into this.

ericoporto:
Sure, I will look at it again today to write more easily to reproduce steps. :)

PR #1465, opened 2021-12-30
New compiler: Make sure that all parameters that follow a default parameter have a default, too

fernewelten:
Typical code:
```
import function Func(int i = 5, float j);
```

Parameter `j` must have a default because `i` has a default. So the code above should yield an error.  Adding an appropriate check.

I'm taking the opportunity to re-order and group together all Googletests concerning function header declarations.


ivan-mogilko:
I think the error message is not worded correctly:

> Parameter #2 of function 'FuncWithDefaults' follows a default parameter and so must have a default, too

Probably should say either "follows an optional parameter and so must be optional" or "follows a parameter with default value...".

PR #1466, opened 2021-12-30
New compiler: Refactorings

fernewelten:
No functional changes.

– I've refactored the `Expect()` function and its calls. This function makes sure that the next code symbol is in a list of given symbols (or else it yields an error).
– Now that the compiler treats errors by throwing exceptions, we need to take special care that all memory that is allocated is deallocated afterwards – even when an exception is thrown. I've done this "the C++" way and replaced naked pointers `*` by `std::unique_ptr` wherever this could be a problem.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -187,6 +187,7 @@ Pointers are exclusively used for managed memory. If managed structs are manipul
 #include <fstream>
 #include <cmath>
 #include <climits>
+#include <memory>
 
 #include "util/string.h"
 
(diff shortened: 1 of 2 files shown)


ericoporto:
First commit is actually from #1465

Compiler/script2/cs_parser.cpp
```diff
@@ -1620,28 +1618,26 @@ void AGS::Parser::ParseFuncdecl(size_t declaration_start, TypeQualifierSet tqs,
     }
 
     // Stash away the known info about the function so that we can check whether this declaration is compatible
-    SymbolTableEntry::FunctionDesc *known_info = _sym[name_of_func].FunctionD;
+    std::shared_ptr<SymbolTableEntry::FunctionDesc> known_info{ _sym[name_of_func].FunctionD };
```
ivan-mogilko, line 1621: The commit's description mentions unique_ptr, which would be correct, but you are using shared_ptr here, which is redundant as the object is not shared and disposed soon after.

ivan-mogilko:
There's a small fix suggestion, also the branch has to be rebased after #1465 got merged.

PR #1467, opened 2021-12-30
3.6.0 Android Player: search for games using proper engine's algorithm

### LINKED ISSUE #379: Android Launcher: game search logic is incorrect

ivan-mogilko:
Resolves #379

**NOTE:** This was already fixed in 3.5.* branch by #1289, but that code could not be merged to 3.6.0 directly, because we have a new AGS launcher for Android (now called AGS Player) written in Kotlin with different file structure, plus there's a number of minor differences.

Based on the work of @sonneveld (Nick Sonneveld) with minor modifications and fixes. Original commit: https://github.com/sonneveld/ags/commit/1b92d9d42d99a401fa02b6cbf0b4849633e581f9

The old method checked for Windows executable header and retrieved program version info to make sure the file contains Windows engine in certain version range. This made *some* sense many years ago when ports were mostly meant for running older games attached to Windows engine exe. But it no longer does.

The new method calls engine's own function FindGameData() - the one that engine uses itself when scanning directory for a game.

PS. First time I wrote something in Kotlin, even though just few lines.


ericoporto:
Hey, you need to use Extern C in jni calls like this: https://github.com/ericoporto/ags/commit/d8a82e61a6455d552abf786b42a607412c262fa0

ivan-mogilko:
It appears that Java side must explicitly load "engine" library, at least for my device to work.

As the current solution requires a number of utility functions from the engine, I would keep this helper inside the "engine" lib for now, with the thought to split out a "nativehelper" lib containing only necessary sources somewhere in the future (not a big priority for the game launcher).

ericoporto:
I am out of home right now, but did it work with the last commit in your device? If it did then the native part of the helper library that is standalone can be removed for now. :)

ivan-mogilko:
Yes, this works now.

> Anyway, after the above change it works the same as before as far as I can tell.

From the user's perspective the difference will be that the unsupported/wrong *.ags files will not be added to the list: previous code in 3.6.0 added any *.ags file, but the new one actually checks if there's a game data inside.
The easiest way to test this is to create a dir with a dummy blank file called game.ags.

PR #1474, opened 2022-01-02
Editor: implement Layout -> Reset To Defaults command

### LINKED ISSUE #599: Editor: "reset layout" menu command
Sometimes it happens so that users break panels layout in the editor and cannot restore it on their own. For example they may drag the window outside of the screen (idk how that happens exactly, maybe there's a bug too). In such case the easiest solution is to reset layout. But atm the only way to do so is to delete Layout.xml in app data.

I suggest adding a ~~simple~~ command into "Window" menu that would reset panel layout to default.
EDIT: okay, it may be not so simple, as editor itself does not know what default layout is. Maybe this could be solved by embedding a default layout file into editor resources?

~Bonus tasks: save and restore custom layouts.~ this was already supported(?)

ivan-mogilko:
Resolves #599.

* A default layout xml is added as an embedded resource;
* Adds Window -> Layout -> Reset To Defaults menu command that loads default layout from resources;
* In case loading layout from file failed, also load default from resource (this also fixes the unhandled exception when load from file fails).


ericoporto:
this PR appears to be working as stated. The layout reset also closes open panes, but I think it's alright.

ivan-mogilko:
Ah. Soon after merging this I realized that because this closes the panel contents, this command may lead to loosing script changes.

So there has to be a warning and a question to confirm saving, similar to when the editor is being shut down.

ivan-mogilko:
Egh. I also found there's a strictly programmatic way to make default layout without an embedded layout resource. Apparently it was not coded correctly year ago, but it's a two lines fix.

Issue #666, opened 2019-03-05
AGSEditor /compile should turn alert into log messages


ericoporto:
So I am looking into building an AGS Game using a CI system.

Once you can get AGS installed in the system, it's as simple as running the following:
```
%AGSINSTALLDIR%\AGSEditor.exe /compile %AGSGAMEPROJECT%\Game.agf
```

If the game builds ok, everything is good. If when building, the game misses a plugin, even if harmless (eg: SpeechCenter), an alert window with an OK is popped up. Also if the game was made with a previous version of AGS, the ok alert window will pop up. And finally, in case of error, another window will pop up.

Ideally, when using a flag like `/compile` I would expect missing a plugin, or other warns to be simple log messages on console, and the build to proceed. Then if the build is ok, it should finish with 0, and if the build hit an error, finish with 1, and log the errors to console.

ivan-mogilko:
This is honestly the first time I learnt about this flag in all those years. No wonder there was no attempt to prevent anything modal from appearing.

Hmm, this may require some good thought and maybe redesign of some processes.
For example, a warning/error reporting interface could be passed in building functions, and let external user decide what to do with alert.

morganwillcock:
+1 for stand-alone game compiler

ericoporto:
Just to illustrate the use, here is Dungeon Hands:
- [Dungeon Hands GH Repo](https://github.com/ericoporto/dungeonhands)
- [Script to install AGS in clean env and build the game](https://github.com/ericoporto/DungeonHands/tree/master/Script)
- [Appveyor Log](https://ci.appveyor.com/project/ericoporto/dungeonhands)
- [Actual game built in CI](https://ci.appveyor.com/project/ericoporto/dungeonhands/build/artifacts)

ivan-mogilko:
Only FYI, as a hotfix we could try checking "compileAndExit" flag in GUIController.ShowMessage() methods. Maybe it's possible to dump the message to stdout instead if .NET can do that (but IIRC when running windowed app on Windows it does not write back to same terminal it was launched from).
Of course that won't cover other possible blocking events we may not know about yet.

ericoporto:
That hotfix works for me :]

ericoporto:
![image](https://user-images.githubusercontent.com/2244442/59315832-d698de80-8c91-11e9-9222-a1fa17944491.png)

[...]

PR #1476, opened 2022-01-03
Editor: intercept corrupt app config case to avoid getting stuck on load

### LINKED ISSUE #1457: Editor fails to launch if the user prefs file is corrupted
**Problem**

This has been reported few times already (e.g. [forum thread](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59358.0)), but if the preferences file gets corrupted for whatever reason Editor refuses to start, displaying "unhandled exception".

**Expected behavior**

Leaving the corruption problem aside, there has to be a way to launch the program even if the prefs cannot be loaded, using defaults.

**To reproduce**

1. Open current preferences file (should be in `%USERPROFILE%\AppData\Local\AGS\AGSEditor.exe_StrongName_mqpdv5kqypl2dezfmjnjxse4zvfvse2h\<VERSION NUMBER>\user.config`) with a text editor.
2. Put random garbage inside and save.
3. Try launching AGS Editor. Observe the exception.

ivan-mogilko:
Resolves #1457

If the application settings file is corrupt Editor threw exception at startup and got stuck at the splash screen.

Added a test in AppSettings constructor that catches any config exceptions and suggests that it should be reset to continue. As a second option it lets close the Editor killing the process (perhaps, may be useful if it was user who was editing the file and left it in an invalid state).

NOTE: I had to add System.Configuration reference to the project.


ericoporto:
I think it's better to delete and inform instead of offering the option of closing the Editor. People rarely read text pop-up messages, so if they click cancel without reading it will simply appear the editor closed for no reason. I understand manually editing the game project, but the config appears to be something someone would not usually do.

morganwillcock:
I'm not sure it is a good idea to have the check within the constructor of the derived class itself, there might be a risk that a change to the parent class would cause the file to be loaded earlier.

ivan-mogilko:
Variant 2:
* test for corruption in the AGSEditor's constructor; I was not certain how soon the Settings are used, so constructor is probably the safest time.
* Do not ask for confirmation, just report the problem and reset file.

Editor/AGS.Editor/AGSEditor.cs
```diff
@@ -324,6 +326,28 @@ public void Dispose()
...
+            // Test for a valid config file, in case of corrupt config
+            // delete the file and notify the user that the prefs will be reset.
+            try
+            {
+                ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.PerUserRoamingAndLocal);
+            }
+            catch (ConfigurationErrorsException ex)
+            {
+                string filename = ex.Filename;
+                MessageBox.Show("Editor's configuration is corrupt and cannot be loaded. " +
```
persn, line 340: Use `Factory.GUIController.ShowMessage` instead, it'll play nicer when we finally make improvements to editor running in headless mode as discussed in #666

ivan-mogilko: Fixed.

ericoporto:
I tested by going to mine `C:\Users\user\AppData\Local\AGS\AGSEditor.exe_StrongName_mqpdv5kqypl2dezfmjnjxse4zvfvse2h\3.6.0.13` dir and messing up the user.config, and it showed the message box saying it was going to delete and keep going and it deleted and kept going! When I tried the same with the previous version it got stuck and required using the task manager to kill the AGS process, so I say this PR is working!

PR #1478, opened 2022-01-04
Android: fix Preferences crashes when no android.cfg

### LINKED ISSUE #1471: Android Launcher: Preferences crash the app if android.cfg does not exist
**Describe the bug**

If `android.cfg` does not exist in the current directory selected by the launcher, trying to open Preferences will crash the app.

**AGS Version**
3.6.0

**To Reproduce**
1. Run the Android AGS Launcher ("AGS Player");
2. Open the settings menu ("..." in the top-right corner);
3. Choose "Select game folder" and setup a directory that either does not exist, or does not contain android.cfg.
4. Now choose "Preferences" from the same menu.
5. Observe the app crash.

**Expected behavior**

Preferences should not crash the app regardless whether android.cfg exists in the selected dir or not. If it's not readable, then Preferences should reset to defaults.

ericoporto:
Crash is caused by unintialized SDL.

fix #1471

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ option(AGS_BUILD_COMPILER "Build compiler" ${AGS_BUILD_TOOLS})
 option(AGS_NO_MP3_PLAYER "Disable MP3" OFF)
 option(AGS_NO_VIDEO_PLAYER "Disable Video" OFF)
 option(AGS_BUILTIN_PLUGINS "Built in plugins" ON)
+option(AGS_DEBUG_MANAGED_OBJECTS "Managed Objects Log" OFF)
 set(AGS_BUILD_STR "" CACHE STRING "Engine Build Information")
 
 
@@ -26,6 +27,7 @@ message(" AGS_BUILD_COMPILER: ${AGS_BUILD_COMPILER}")
 message(" AGS_NO_MP3_PLAYER: ${AGS_NO_MP3_PLAYER}")
 message(" AGS_NO_VIDEO_PLAYER: ${AGS_NO_VIDEO_PLAYER}")
 message(" AGS_BUILTIN_PLUGINS: ${AGS_BUILTIN_PLUGINS}")
+message(" AGS_DEBUG_MANAGED_OBJECTS: ${AGS_DEBUG_MANAGED_OBJECTS}")
 message("----------------------------------------")
 
 # Tools like ninja don't have a pseudo-terminal so compilers assume no coloured output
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
If i understood [the cause of the issue](https://github.com/adventuregamestudio/ags/issues/1471#issuecomment-1005247681) correctly, any random call to "OpenFile", or several other functions that use Android specific functions internally, may cause the same issue. If so, then this should be fixed inside OpenFile instead, or inside the Android related code (AAssetStream, GetAAssetExists and so on).

With current fix this is rather a small patch on one occurance of a problem we stumbled upon.

ericoporto:
I answered in the issue, but once the CI finishes the APK will be built. I also noticed that if we were using a try/catch exception would be a bit easier to handle but we use return null as indication of failure. :/

ivan-mogilko:
Hm, I was about to admit that it may be fine to have this as a hotfix before any larger changes.

But then I noticed that there's a `reset_configuration` variable that basically tells every ReadN function to not read anything but set default value. This is done in a kind of hacky way too, but is something.

It looks like if you call ResetConfiguration() function you may avoid reading a file and set everything as defaults instead. So *maybe* it would look better if you 
* test for the file existence
* if it does not exist, then call ResetConfiguration(). This may be done even on Java side.

> I also noticed that if we were using a try/catch exception would be a bit easier to handle but we use return null as indication of failure.

Which return (in which function) are you refering to?

ericoporto:
Did a change that would add a check for SDL initialization, and throw in case it can't load the AAssetManager.

ivan-mogilko:
The new code looks fine, and i confirm that it no longer crashes on device if android.cfg was not present.
If you could squash the commits it would be perfect.

On a side note, something i noticed, the android build seem to have DEBUG_MANAGED_OBJECTS enabled, that fills the log with all the thousands of managed object messages. Was this always like this or something changed in makefiles?

ericoporto:
> DEBUG_MANAGED_OBJECTS enabled, that fills the log with all the thousands of managed object messages. Was this always like this or something changed in makefiles?

There should be no Makefiles involved, if this is turned on its in the debug build on CMake for all systems, I can check this out later

ivan-mogilko:
Yes, it's best to have it off by default, because even in debug configuration they are too verbose, and rarely wanted.

BTW, SOUND_CACHE_DEBUG no longer is present in the engine code, because sound cache was removed when moving to SDL.

ericoporto:
I will remove the flag, I didn't knew it, just a sec

PR #1479, opened 2022-01-06
New compiler: Further refactorings

fernewelten:
No functional changes.

I'm still indulging in yak shaving to make it easier to implement delegates in the compiler.

1. When I started out, I considered attributes to be a special kind of variable - an attribute variable. It's not, it's a different category of type that needs different information in the symbol table and different code to process. In this way, it is similar to functions and the upcoming delegates that need special treatment, too. The implementation has already drifted this way; now I've streamlined it to make it easier to put delegates alongside.
2. Similarly for `const`: Apart from the type `const string`, the compiler uses `const` as a special category; it keeps track of (integer or float) literals using dedicated fields in the symbol table and it has dedicated compile-time processing for `const`s (for instance, `3 + 4` is converted to `7` at compile-time). This is already implemented, but I've streamlined it to make it easier to put delegates alongside.
3. When a delegate object is called, this looks exactly like a function call. So seeing that delegates will be implemented, I've cleaned up the function calling code beforehand.

### DIFF
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -55,7 +55,6 @@ std::map<AGS::TypeQualifier, AGS::Symbol> const &AGS::TypeQualifierSet::TQToSymb
         { TQ::kAttribute,       kKW_Attribute, },
         { TQ::kAutoptr,         kKW_Autoptr, },
         { TQ::kBuiltin,         kKW_Builtin, },
-        { TQ::kConst,           kKW_Const, },
         { TQ::kImport,          kKW_ImportStd, },
         { TQ::kManaged,         kKW_Managed,  },
         { TQ::kProtected,       kKW_Protected,  },
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -9,6 +9,8 @@
 AGS::SymbolTableEntry::~SymbolTableEntry()
 {
     // (note that null pointers may be safely 'delete'd, in contrast to 'free'd)
+    delete this->AttributeD;
+    this->AttributeD = nullptr;
     delete this->ConstantD;
     this->ConstantD = nullptr;
     delete this->ComponentD;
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
> No functional changes.

Testing this I found changes in behavior.

<pre>
function f (const int a)
{
}
</pre>

This worked before but no longer does, compiler error is:
> The only allowed type beginning with 'const' is 'const string' (did you want to use 'readonly'?)

Is this a correct behavior now?

fernewelten:
> Is this a correct behavior now?

I fear that this qualifies as a bug in the __unmodified__ compiler. Here's another short program:
```
void func(const int i) 
{                     
    i = 0;              
}  
```
The unmodified compiler eats this without balking, i.e., it ignores the `const` in the parameter list. OTOH, if we code `readonly` instead of `const`, it balks:
```
void func(readonly int i) 
{                     
    i = 0;              
}  
```
--> `Cannot write to readonly 'i'`

fernewelten:
So this is what the compiler tries to do:

- `const` in a statement defines a compile-time constant. Both `int` and `float` are possible. For instance, `const int i = 5 + 3;` declares that `i` is synonymous to the integer value 8. __No code is generated at this point__, this is only tracked in the symbol table, and whenever `i` is in the code from this point onwards until `i` gets out of scope, it is treated as if an 8 was written there. (As far as I know, C# does this, and it is in line with CJ's way of treating `readonly` below)

- As a special exception, `const string` is the type of an old-style string where its content must not be modified. 

As a bit of context, CJ's train of thought as far as I understand it: Everyone knows that parameters are passed "by value" in AGS. So normally, even if a function body changes a parameter, this will not modify the variable that had been passed to the function. But something different can happen if you pass a __pointer__ to a function. Then the pointer of the caller will still stay unchanged no matter what the function does, but the __thing that the pointer points to__ can be indeed modified by the function. (As generally known, this is the standard `C` way of passing variables by reference, `&variable` and `*parameter`.)

In AGS, old-style `strings` are passed to functions as pointers to the first byte. So the function can modify the content of a `string` passed to it in a way that the caller sees.

Trouble occurs when a function defines a `string` parameter and the caller passes a `String` to it. Then the `String` will be implicitly converted to a `string`, and a pointer to the first byte of the `string` is passed to the function. Then the function body can modify the contents of the `string`, and the result is that the  __contents__ of the `String` that the caller has passed will be modified And it's strictly forbidden to modify the contents of a `String`.

So specifically for `string` type parameters that might be passed a `String`, CJ needed a type where it is disallowed to modify its contents. And so `const string` was born. CJ didn't do this, but we've learned in the meantime that we also need `const string` to be a possible return type. 

- `readonly` in a statement defines a variable that must not be modified. Any type is possible. For instance, `const int i = 5 + 3;` defines a space on the stack or in global or `import` space that will be allocated right then, right here, and 8 will be put into this space. The compiler will balk at all attempts to change the variable directly. But it might be changed behind-the-scenes, e.g. if it is an import variable and a different module declares it as non-`readonly`.

- If a function declares a parameter as `readonly`, this parameter is treated within the function body in the same way that a `readonly variable` would be. That is, the function body may not modify it.

ivan-mogilko:
I really wish we had some kind of a cheat sheet at least, documenting new compiler changes.
There was a short list once, in the very first PR, but it's long buried deep in closed issues. And many more changes were added since.

It may be too early to do in the manual; but perhaps in the [repository wiki](https://github.com/adventuregamestudio/ags/wiki)?

fernewelten:
> It may be too early to do in the manual; but perhaps in the [repository wiki](https://github.com/adventuregamestudio/ags/wiki)?

I'm having a go at it.

PR #1483, opened 2022-01-07
Engine: AnimateCharacter with better error message for invalid loop error

### LINKED ISSUE #1482: clarification request for Error "invalid loop number specified"
**Describe the problem**
Error message clarification request: AnimateCharacter "invalid loop number specified"

image

**Suggested change**
Would it be possible to include the character and view number on this error message? It would save a *ton* of time.

Originally from forums here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59636.0

ericoporto:
AnimateCharacter "invalid loop number specified" didn't inform, either loop number, view or character.

image

Fix #1482

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -2125,8 +2125,10 @@ void animate_character(CharacterInfo *chap, int loopn,int sppd,int rept, int noi
         Character_UnlockView(chap);
         chap->idleleft=chap->idletime;
     }
-    if ((loopn < 0) || (loopn >= views[chap->view].numLoops))
-        quit("!AnimateCharacter: invalid loop number specified");
+    if ((loopn < 0) || (loopn >= views[chap->view].numLoops)) {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I may accept this, but the problem is much bigger that this one case: there is a multitude of similar messages, and they all have different report format, some having extra information, others not at all. For example, in other cases of view-related problems I was also adding valid ranges to the error message: https://github.com/adventuregamestudio/ags/blob/v.3.6.0.13/Engine/ac/global_object.cpp#L162


My believe is that the true solution would be to create a number of helper functions, reporting various types of issues with uniform formats.

ericoporto:
While I understand what you mean, for this particular case, I just copy pasted the message that is some lines above

https://github.com/adventuregamestudio/ags/blob/77990fb23773813785e61012590f550b04d304ff/Engine/ac/character.cpp#L2117-L2120

ericoporto:
The problem for sistematic error messages is now separated on issue #1486

PR #1485, opened 2022-01-07
Engine: audio core init may throw

ericoporto:
when running `audio_core_init()`, it may throw if **no audio device is available**. 

I discovered it today when I had my headphones disconnected from my computer and no speakers since I had just moved the desk. In this case, AGS from our master doesn't start. Verified in Windows 10.

Previous allegro based AGS show a windows message saying it couldn't initialize audio, and then proceed with the game. I thought a debug message is fine - if it isn't, I can add a **`platform->DisplayAlert()`**.

https://github.com/adventuregamestudio/ags/blob/master/Engine/media/audio/audio_core.cpp#L78-L82

This PR simply catches the thrown errors and disables the audio backend.

### DIFF
--- a/Engine/main/engine.cpp
+++ b/Engine/main/engine.cpp
@@ -383,7 +383,12 @@ void engine_init_audio()
     if (usetup.audio_backend != 0)
     {
         Debug::Printf("Initializing audio");
-        audio_core_init(); // audio core system
+        try {
+            audio_core_init(); // audio core system
+        } catch(std::runtime_error) {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
What will happen if user turns the headphones on during the game?

ivan-mogilko:
Ah. It's the audio_core_init() that throws, because it's written that way.

Alright, I will merge this, but I believe this situation has to be documented in its own ticket and investigated.

ericoporto:
I opened #1487 just to remember this.

PR #1489, opened 2022-01-09
Android Player: let user select dir through file picker

ericoporto:
Possible fix for #597 

Adds a "browse" button when selecting directory which will launch the default file picker to let the use select the directory. I have no idea what happens when there is no file picker, but I think most devices have one.

### DIFF
--- a/Android/agsplayer/app/src/main/java/uk/co/adventuregamestudio/agsplayer/FileUtil.kt
+++ b/Android/agsplayer/app/src/main/java/uk/co/adventuregamestudio/agsplayer/FileUtil.kt
@@ -1,0 +1,108 @@
+package uk.co.adventuregamestudio.agsplayer
+
+import android.annotation.TargetApi
+import android.content.Context
+import android.net.Uri
+import android.os.Build
+import android.os.storage.StorageManager
+import android.os.storage.StorageVolume
+import android.provider.DocumentsContract
+import androidx.annotation.Nullable
+import java.io.File
+import java.lang.Exception
+import java.lang.reflect.Array
+import java.lang.reflect.Method
+
+// lifted from https://stackoverflow.com/questions/34927748/android-5-0-documentfile-from-tree-uri/36162691#36162691
+
+object FileUtil {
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
So, testing this, on "Browse" I have a screen with a "Open from" list. But this list contains only 1 item called "Recent", which in turn does not show any items. So I cannot proceed.

How do these file pickers suppose to work on Android anyway?

ivan-mogilko:
EDIT: [redacted]

Oh, I actually was able to get better results, by choosing "Show SD card" from the menu. That was not very obvious, but now I can see all the internal storage. Without it I saw only "Recent" folder which was empty.

ericoporto:
The file picker and any related configs are per device and depends on whatever file manager each one uses. In my case I use [Amaze File Manager](https://github.com/TeamAmaze/AmazeFileManager), and it's the file picker and settings I see. :/

The other way to do this is to implement a basic file manager in the AGS app and use that instead. I looked for libraries for doing this but they all looked abandoned, like [this](https://github.com/BoardiesITSolutions/FileDirectoryPicker) and [this](https://github.com/passy/Android-DirectoryChooser), but they all seem to ignore the new Storage Access Framework (Android 11 and upwards), and the file picker can be made to work - once we switch to SDK 30 for building. The other way is handmading our own but I tried this other way first in case this was good enough.

ivan-mogilko:
I think this should work. This is just for one config action, so spending more time on this now will be counterproductive.

BTW, on related note, do you know how this path is saved? It gets reset every time I reinstall app, and because i have to test a lot of versions, setting it back becomes a chore. Is this something that we may control, or system controls when the settings are deleted along with the app?

ericoporto:
I think the path is part of the app settings, I can check it out later. If you never uninstall but instead just install over (with adb or Android Studio) it should actually preserve the settings. Are you uninstalling beforehand each time?

ivan-mogilko:
> Are you uninstalling beforehand each time?

Yes, installing over does not work for some reason.

PR #1490, opened 2022-01-09
Android: recursively iterate for games on Android Player

### LINKED ISSUE #77: Recursive gamedir
Hi!
Please add support for a recursive gamedir (on Android).
When I put my games into subfolders, they are not shown in AGS.

ericoporto:
fixes #77 

Will make Android Player recursively look for games when a directory is set

Based on the original fix from here: https://github.com/sonneveld/ags/commit/a4f3d4af65365c2764f7a3b70bbd415dea02f9f0

### DIFF
--- a/Android/agsplayer/app/src/main/java/uk/co/adventuregamestudio/agsplayer/MainActivity.kt
+++ b/Android/agsplayer/app/src/main/java/uk/co/adventuregamestudio/agsplayer/MainActivity.kt
@@ -200,6 +200,27 @@ class MainActivity : AppCompatActivity(), GameListRecyclerViewAdapter.ItemClickL
         }
     }
 
+    private fun getDirectoryTree(baseDirectory: String, path: String?): LinkedList<String> {
+        val result = LinkedList<String>()
+        var searchDirectory = File(baseDirectory)
+        if (path != null) {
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I confirm this works.

The question is, how it is desired to work. Unfortunately, the original ticket never had it explained.
Right now all the games are displayed in one list, and they have full path prepended.

E.g.
games/MyGame
test/MyTestGame

and so on.

This may be enough for now, but to make things cleaner, there got to be some game list redesign.

ericoporto:
There may be a way to remove the prepended directory and only add them when there's another game with same ending dir name, but I don't know how to deal with the reverse yet - in my case there were games copied directly from their Linux counterparts and they all ended with `/data`.

ivan-mogilko:
> in my case there were games copied directly from their Linux counterparts and they all ended with /data.

Oh, i did not realize that's the case.

Hmm, perhaps we could display the actual game name, as it may be preread from the data, like winsetup does; but this seem like a completely separate issue.

ericoporto:
> Hmm, perhaps we could display the actual game title, as it may be preread from the data

I will look into, it should be easy to add a function to the native helper library.

PR #1491, opened 2022-01-09
Refactor expressions

fernewelten:
No functional changes - again. 😀

So here's the last one for now, but it's an important one.

Looking at the parser header file, I realized that dozens of functions shared the same parameters. That's a big tip off that the parameters really belong together and so should probably be bundled into a class or struct.

And they do. The functions concerned all deal with generating code to evaluate an expression, and the parameters concerned are all aspects of the evaluation result:
- What kind it is
- What type it has
- Where it can be found
etc.

So I've bundled the parameters into a `struct EvaluationResult`. 

Moreover, the code had partly different functions for "expressions that must have side effects (such as a function call or an increment operation)" and for "expressions in general". I found that I could unite those functions by keeping track in the `EvaluationResult` whether side effects happened while it was calculated. 

This does make the code simpler. The downside is that some dozens of functions and their calls change, so there are lots of modified code lines.

Then I saw a problem with the mechanism that updates the MAR register. This updating is done "lazily": Instead of generating code whenever an offset needs to be added to MAR, all those offsets are accumulated in an object, and the instructions to load the MAR register are only generated immediately before the register needs to be used. 

I saw the problem with the object that manages those updates: This object was lugged around as a parameter of many functions. But that doesn't make sense: There's only one MAR register, so there is only one object managing this register (it must by needs always be the __same__ object), and so it makes more sense to keep it in a class variable, `_marMgr`, instead of passing it around in function calls.

Lastly, on a different topic, there are some opcodes that do Boolean operations, and the compiler recognized which ones do by looking at the specific opcode numbers. It's bad style to rely on "magic numbers". So I created a field in the symbol table for opcode entries; this field keeps the information whether the respective opcode is a Boolean one.


fernewelten:
(#648facf and #2ed4955  are identical except that I deleted a superfluous line out of the commit comment.)

ericoporto:
using `git rebase -i HEAD~4` in this branch you should be able to mark `2ed4955` with an `s` to merge it in the previous commit.

ivan-mogilko:
Depending on the git fronted you are using, there may be GUI methods to do the same.

ivan-mogilko:
@fernewelten tested on few projects and see no problems; but could you please remake the branch to not include commit duplication and merge commits in it?

fernewelten:
> could you please remake the branch to not include commit duplication and merge commits in it?

I'll rebase.
I've got a kind of emergency on my hands right now, but I'll get on to it ASAP, probably already this afternoon.

fernewelten:
(Solved this by discarding the unwanted commits and then force-pushing the branch)

PR #1495, opened 2022-01-11
3.6.0: optionally clip GUI control contents

ivan-mogilko:
Historically AGS does not clip GUI controls, which may result in texts drawn outside of the control rectangle. While users may get used to this, it's not correct in my opinion. Also in some cases (like with the list boxes) it may be difficult to work around.

This PR introduces a game setting, which tells whether to clip control content or not. It's enabled by default in the new 3.6.0 projects, but disabled for projects upgraded from before 3.6.0.

My thinking is that control clipping their contents (text, images etc) should be a default. I may imagine a couple of use cases when one may want to have text drawn outside of the button, for instance, but it seems these may be resolved by combining a button and a label, or having a custom sprite on a button. But I am ready to hear more opinions on this.

EDIT:
Theoretically, I may also add a per-control property, merging it with the Button's "Clip Image". Technically this should be trivial, as this is just an internal engine's flag that may already be applied to any control. But in this case I'm more worried that it will make game design complicated, as people will have to specifically remember to toggle this on or off per each thing that needs it.

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -77,6 +77,7 @@
 #define OPT_RENDERATSCREENRES 45 // scale sprites at the (final) screen resolution
 #define OPT_RELATIVEASSETRES 46 // relative asset resolution mode (where sprites are resized to match game type)
 #define OPT_WALKSPEEDABSOLUTE 47 // if movement speeds are independent of walkable mask resolution
+#define OPT_CLIPGUICONTROLS 48 // clip drawn gui control contents to the control's rectangle
 #define OPT_HIGHESTOPTION   OPT_WALKSPEEDABSOLUTE
 #define OPT_NOMODMUSIC      98
 #define OPT_LIPSYNCTEXT     99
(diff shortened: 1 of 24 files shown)


ericoporto:
Does adding the toggle, means #1393 is moved to ags4 and that the toggle is then removed in ags4? I do think that clipping is overall the expected behavior and the AGS behavior of not clipping GUI Controls is unexpected. Specially since the GUI itself is currently clipped - I leverage this behavior to code scrolling in GUIs.

ivan-mogilko:
I posted this in Discord and forums to hear opinions, so will delay this PR for a while (it's not urgent, and we are still working on 3.6.0).

So far one concern mentioned is that of the translations. Indeed, if automatically translated line is longer than expected, then it will be clipped. This will look evern worse if the "text alignment" is "centered". Currently it's drawn outside of the control.
Indeed this will be annoying, yet imho this is an issue of arranging controls for translations, because even without clipping this may cause text to mix with other ui elements and look bad or unreadable.

ivan-mogilko:
Well, there has been an abscence of interest to this, so guess I'll just push this in. In theory it will be possible to add per-control setting too (although I hope it won't be necessary).

PR #1504, opened 2022-01-17
Upgrade sdl sound

ericoporto:
This is just to keep more updated to current SDL Sound main branch, hopefully there will be a tag release someday.

This requires rebuilding our dockers (may take a little while), but Debian Jessie now complains of the certificate for sdl website, so I switched it's URL to GitHub source for now.

I also took the time to update the Windows and Linux README files since they contained links to the removed mercurial repository.


ivan-mogilko:
Tested with the new snapshot, seem like working.

PR #1507, opened 2022-01-19
3.6.0 Editor: fixed sprites never deleted from spritefile

ivan-mogilko:
The commit 393ef2c24bbc835d8e7a7229d40f638a6f172304 has introduced a bug, where saving spritefile would unintentionally restore all the sprites that were deleted during the session.

Sprites are not all stored in memory, and when the file is saved, these that were not loaded at the moment would be read from the *previous* version of the file, and copied into new one. Aftter certain changes the code was missing a proper check for non-existing sprites. Because of that all the sprites that were in the previous file but marked for deletion are still carried over.
And because their references are removed from the project, these would be non accessible, and user won't even know about the leftover data, unless they try to import new sprites and notice they cannot assign a presumably "empty" slot to them.

This PR fixes the problem, but also implements a cleanup procedure, where Editor would test for sprite data that does not have a proper reference inside a project, and mark it for deletion. This procedure is run right after the game is loaded.

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -480,6 +480,8 @@ int SpriteCache::SaveToFile(const String &filename, int store_flags, SpriteCompr
 
 HError SpriteCache::InitFile(const String &filename, const String &sprindex_filename)
 {
+    Reset();
+
     std::vector<Size> metrics;
     HError err = _file.OpenFile(filename, sprindex_filename, metrics);
     if (!err)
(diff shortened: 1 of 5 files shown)


ericoporto:
> This procedure is run right after the game is loaded.

Hey, just clarify the behavior, it does run, but it doesn't save, and saving the project right after won't do too, only after a modification is done (like deleting or importing a new sprite), then it removes the erroniously undeleted sprite.

ivan-mogilko:
Corrected, also added a message.

ericoporto:
Hey, if I use the version before the last one, the one that crashes on saving, I get the sprite cache in a state where it remove the sprites that are missing, but it also removes the ones that are in the sprite cache and in the project. It looks like this in the Sierra Template:

image

Issue #1477, opened 2022-01-03
Editor: Switch Preferences UI for a property grid (like General Settings)


ericoporto:
Currently the Editor preferences is a nicely handmade screen like this

![image](https://user-images.githubusercontent.com/2244442/147975612-73ae5dba-5b6e-4158-abf4-92e917c0c3dc.png)

It's functional, but adding an item in the preferences requires juggling around - I am currently adding some things in the bottom left corner there and I feel the next person will have some hard time on this screen...

I would suggest redoing it similar to general settings, which uses the property grid. This would probably require some rework, like the path search needs  to be possible in the properties, but it would make it less cumbersome to add or change preferences.

I would like some opinions on this.

persn:
I've been annoyed by the same thing in the past so I'm all for this. It would be best if changed this screen to change dynamically as we added new properties

ericoporto:
I was totally expecting a Preferences.cs on AGS Types and was very frustrated when I found out things weren't exactly like this xD

I can do the change btw, I just want to see if there's something that I am forgetting.

persn:
Adding an `IPreferences` to AGS.Types would be welcome at the same time. Interface can be extracted from `AppSettings` in `EditorPreferences.cs`, and any related types would also have to be moved.

ericoporto:
> Adding an IPreferences to AGS.Types would be welcome at the same time. Interface can be extracted from AppSettings in EditorPreferences.cs

I am looking at `AppSettings` but I don't think I understand how such interface would be. 

> any related types would also have to be moved.

They all appear to be defined in the same `EditorPreferences.cs`.

Also, in my thinking Recent Games would be somewhere else... Not sure where, but it and Recent Searches look like something that should be... Elsewhere 🤔

persn:
> I am looking at AppSettings but I don't think I understand how such interface would be.

Something like this I recon.
```csharp
public interface IPreferences
{
    TestGameWindowStyle TestGameWindowStyle { get; set; }
    StartupPane StartupPane { get; set; }
    // Etc.
    void Save();
}
```

Keep in mind that this would just be desireable, you don't actually have to do it for the sake of this issue

ivan-mogilko:
> Also, in my thinking Recent Games would be somewhere else... Not sure where, but it and Recent Searches look like something that should be... Elsewhere

[...]

Issue #1362, opened 2021-07-14
Read engine Log in the Editor


ericoporto:
**Describe the problem**
System.Log was added in the engine and it's a super useful feature! But we can't read this log ouput in the Editor.

**Suggested change**
Below is a way I got working by creating a console to read stdout/stderr directly. _(gif links to video)_
[![ezgif com-gif-maker](https://user-images.githubusercontent.com/2244442/125671792-e37ad72c-f99a-4ac8-bbf9-bfc8c13d7cc1.gif)](https://user-images.githubusercontent.com/2244442/125671607-b33ae047-1593-4d8a-ae3b-ba5160c42ad7.mp4)
Editor Build with Engine: **[AGS-4-Console-Log.zip](https://github.com/adventuregamestudio/ags/files/6818311/AGS-4-Console-Log.zip)**
https://github.com/ericoporto/ags/commits/feature-console

- in this approach there was a need to add flushing in the Windows printf statement in the engine
- code above on top of ags4.
- there's also opportunity here to use the console for errors instead of messagebox on the Editor (possible fix for #1034).
- the code is only POC, it has things not implemented and things that could be removed.
- this panel is inspired by Godot Output panel, which has an additional preference options that could be the default here:
  - Always Clear Output On Play
- I have no idea how to theme it to work with the existing themes...
- Console is just a RichTextBox winform control

I think figuring how to properly do this can take a while so I am opening this with the intent of getting ideas. 

I also am not sure a console is the way to go for logging, I can see how a way through debug messages would give things like filtering per type of log message in the editor (like Android Logcat panel), but using a console was much easier to implement.

ivan-mogilko:
This is interesting.

Have you checked the IAGSEditorDebugger though? One alternate approach, possible with less code changes, could be to send messages through it, if you register the debugger object as a new log sink (they are called "targets" in the current code, because I did not know correct term back then).

ericoporto:
Yes, I did check `IAGSEditorDebugger` but I could not understand how it worked...

I do think it's probably better, it would enable that each line had actual group information, log level, and well, other information available to the debugger. 

I didn't had much idea on how a proper panel to it would look, most IDEs I tried have simple text buffer as output. I know table like forms in other toolkits I tried get really slow above a certain number of lines, but I had not tested if this happens with winforms too.

ivan-mogilko:
> but I could not understand how it worked...

[...]

PR #1509, opened 2022-01-20
Editor: Android build feature

### LINKED ISSUE #1337: Editor: Android build target
The suggestion is to implement game building and packaging for Android into the AGS Editor.

I do not have full knowledge of how properly packaged Android APK for single game is done, so give a general task overview, with intent to fill the gaps later as necessary.

### Part 1: ensuring building by hand

1. The required output is a signed APK containing single game.
2. We have a single game APK template in the repository now, unless I'm mistaken it's here: https://github.com/adventuregamestudio/ags/tree/master/Android/mygame
3. We should be assuming prebuilt native engine libraries for this, so that users don't have to build them themselves.
3. Gather information of what steps are required in order to get from the compiled game data to the complete APK.
4. As these steps would require tools (Android SDK and more), make list of what we may have to write ourselves, distribute with the Editor, and what we do not distribute and leave for users to download themselves.

NOTE: Ideally this feature should rely on as less downloaded utilities as possible. I.e. if it's possible to not download full Android Studio, but only its command line tools.

### Part 2: integrating with IDE

After we have a list of things which make it possible to do by hand, we move to integrating with IDE.
1. Building for Android may require setting additional things up. For these we may need a separate property list editor/panel. There was an Editor plugin [created by monkey0506 several years ago](https://www.adventuregamestudio.co.uk/forums/index.php?topic=53910.0), it is already in C#. If license allows we may use it directly, or use it as a reference to write our own (as plugin or as a part of the editor itself).
2. Make a [BuildTarget](https://github.com/adventuregamestudio/ags/tree/master/Editor/AGS.Editor/BuildTargets) class for Android and make sure it appears in the list of possible targets in the project settings.
3. Configure installer to have "Android build package" as an optional component, similar to the "Linux package". This package will include prebuilt native engine libraries and any extra tools we had to write.
4. Adjust CI to let us package installer with this new component.

### LINKED ISSUE #1472: 3.6.0: Android Launcher: restore the in-game menu
The older Launcher app (prior to 3.6.0) has an in-game menu called by holding down the BACK key.
That menu includes following items:

* Toggle keyboard - would bring up on-screen keyboard letting player to pass key input into the game;
* Keys - would bring up a submenu with special desktop keys (F1 - F12) to pass into the game;
* Exit game - would ask for a confirmation and close the game if player accepts.

The new 3.6.0 does not include this menu. It's *very much* desired to have it back, either in the same form or any other that works. These options were essential for playing games that were not specifically adjusted for Android and have desktop controls which requires key input.

ericoporto:
This PR should be working now. Additional information in the [forum thread](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59772.0).

https://user-images.githubusercontent.com/2244442/150422308-c7d3c43f-6dcd-45c0-ac7c-7a068ccf7b6e.mp4

## Test instructions

1. Load the AGS Editor built from this codebase and on general settings you should see an Android build option
  a. If you see no Android build option, then you are missing the android bundle. 
  b. If you are building from source you need to download the android bundle, it's named as `AGS-3.6.0.16-android-proj-release.zip` and you can get it from the CI artifacts. It's just the single game project, with the libraries put under libs, and it should be placed in a directory named `Android/` in your Editor root folder.
2. Install Android Studio, preferably the last one, and set the following environment variable in your Windows
  a.  `ANDROID_HOME`, set it to where you installed the Android SDK (usually `C:\Users\USERNAME\AppData\Local\Android\Sdk`)
  b. `JAVA_HOME`, set it to the jre in the Android Studio installation (usually in `C:\Program Files\Android\Android Studio\jre`)
3. Reboot, or log out, just make sure that these are properly applied
4. Now load the AGS Editor again and hit build EXE

---

Leaving in draft since this has a lot of things missing, but just get things started here. Here are some things to do:

- [x] figure how to put keystore info in preferences or **people are going to push their key passwords to git** (maybe someone will solve #1477 ?)
- [x] Allow setting where the keystore is through the Editor (in preferences), this is now set in `Compiled\Android\mygame\local.static.properties` 
- [x] Actually apply the package name, this can be done through gradle nowadays (no renaming)
- [x] move the .aab or .apk file after it's built to root of Android dir.
- [x] Set whether to package game inside apk or in asset dir with .aab
- [x] figure a way to pass file filters through Property decorators..
- [x] CI to prepare the package I made above
- [x] Allow configuring ANDROID_HOME and JAVA_HOME through Editor preferences.
- [x] Check with `sdkmanager` that dependencies (android platform sdk and ndk for the versions used) are in place, and if not, ask to install them (not sure if needed, maybe gradle can request them)
- [x] Override back button for keyboard toggle and extra keys (fix #1472)
- [x] Write `android.cfg` file
- [x] Properly write the app name used in Android launcher
- [x] Add Keystore Generation
- [x] APP icons are picked up from `icons/android/` dir at the AGS Project root, which has to contain android mipmap icon dirs with `ic_launcher.png` files.
- [x] automatically close the gradlew log window (but keep it if there's an error)
- [x] do not use gradle daemon by default, but allow turning it on through preference to speed up building

image

---

## FAQ

**Q:** Can I not install Android Studio?
**A:** I am using the exactly same tools from Android Studio - it requires installing Android Studio. You could theoretically install only Android SDK, Bundle Tool, Java JRE and JDK but then you need to piece everything together and it will take roughly the same storage space, so it's not worth it.

---

Fix #1337

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -149,30 +149,15 @@ build_android_task:
     greedy: true
   git_submodules_script: git submodule update --init --recursive
   build_debug_library_and_runtime_script: |
-    pushd Android/agsplayer && ./gradlew assembleDebug && popd
+    pushd Android && ./and-build.sh build_debug && popd
   build_release_library_and_runtime_script: |
-    pushd Android/agsplayer && ./gradlew assembleRelease && popd
-    bsdtar -f android-libs-release.zip -acv --strip-components 8 Android/agsplayer/app/build/intermediates/stripped_native_libs/release/out/lib
+    pushd Android && ./and-build.sh build_release && popd
   rename_apks_script: |
-    version=$(awk -F"[ \"]+" '{ if ($1=="#define" && $2=="ACI_VERSION_STR") { print $3; exit } }' Common/core/def_version.h)
-    pushd Android/agsplayer/app/build/outputs/apk/debug/
-    for apk in $(find -maxdepth 1 -name "*.apk" -type f); do
-      mv -v $apk AGS-${version}-${apk#*-}
-    done
-    popd
-    mv Android/agsplayer/app/build/outputs/apk/debug/*.apk .
-    pushd Android/agsplayer/app/build/outputs/apk/release/
-    for apk in $(find -maxdepth 1 -name "*.apk" -type f); do
-      mv -v $apk AGS-${version}-${apk#*-}
-    done
-    popd
-    mv Android/agsplayer/app/build/outputs/apk/release/*.apk .
+    pushd Android && ./and-build.sh archive_apks && popd
   apks_artifacts:
     path: AGS-*.apk
   create_libs_archive_script: |
-    version=$(awk -F"[ \"]+" '{ if ($1=="#define" && $2=="ACI_VERSION_STR") { print $3; exit } }' Common/core/def_version.h)
-    bsdtar -f AGS-${version}-android-libs-debug.zip -acv --strip-components 8 Android/agsplayer/app/build/intermediates/stripped_native_libs/debug/out/lib
-    bsdtar -f AGS-${version}-android-libs-release.zip -acv --strip-components 8 Android/agsplayer/app/build/intermediates/stripped_native_libs/release/out/lib
+    pushd Android && ./and-build.sh archive_libs && popd
   libs_artifacts:
     path: AGS-*-android-libs-*.zip
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -238,6 +238,7 @@ windows_packaging_task:
     - build_windows
     - linux_packaging
     - build_emscripten
+    - build_android
   windows_container:
     dockerfile: ci/windows/Dockerfile
     os_version: 2019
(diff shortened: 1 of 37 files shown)


ericoporto:
> Is it feasible to connect the output handles on the build processes to aggregate in an Editor control, in a similar way to what you are showing in `#1362`

Actually in my first version I had a gradle panel that was using the console from the first version there that included a gradle stop button. I could also add a generic Console panel that could be reused whenever we need one. This most about figuring the design!

> are there separate log file outputs running too for when someone needs to investigate a problem?

Build logs are always generated though. I don't even know how to not generate them.

ivan-mogilko:
I'm only testing how it works from the user's perspective now, and there are couple of things to note:

1. In General Settings - Android section theres' "AppVersionName" property. We also have a "Version" property located in "Windows Game Explorer" section. May it have sense to merge these and move to e.g. "(Basic properties)" section, and reuse for every build that needs a game version string?
2. If user forgets to input "keystore" settings in the Editor Preferences, then building for android results in unelaborate error message about "paths being null or empty". May it be possible to test for this problem and throw an error with a more precise message? Or, perhaps, check that preferences set correctly before running the build.
3. I might have missed information about this, but why the resulting app file called `app-mygame-release.apk`, and does not use the game file name property from the Settings? Are there some Android rules about apk naming?

ericoporto:
I think all necessary features are in. I can modify any commits necessary. Full usage instructions are in the [forum thread](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59772.0).

ericoporto:
Please look if you can reproduce the graphical glitch Mehrdad had in the forums, I could not trigger that when I tested.

ivan-mogilko:
Tested a pair of compiled games. Don't see any immediate problems.

Couple of questions though:
1. Is there a way to setup the game's config, similar to "AGS Player"?
2. When installing apk I noticed that it requires "full network access" among other things. Why is that necessary?

ericoporto:
> When installing apk I noticed that it requires "full network access" among other things. Why is that necessary?

This is weird, the only thing it should require is sdcard write access, which is really not needed but I forgot to take out - I need to figure how to pull from single game without pulling out from the agsplayer. I will check this.

ivan-mogilko:
Alright, i think this is fine to use.

PR #1512, opened 2022-01-23
Script API: settable Hotspot.Name and Object.Name

### LINKED ISSUE #979: Script API: allow to change Object.Name at runtime
This is many years old request, users would like to be a able to change Object.Name and Hotspot.Name at runtime. These names are returned by GetLocationName, and sometimes need to be modified to reflect game state change.

Following Names are already settable:
* Character.Name
* InventoryItem.Name

Following has to be made settable:
* Object.Name
* Hotspot.Name

Things to keep in mind:

1. Be careful when modifying structs, as some _may_ have their first fields exposed in the old script API, and plugin API, in which case they cannot be replaced and their order cannot be changed.
2. The visited rooms are kept in memory as RoomStatus objects. When a room is first loaded, its data that is changeable is copied into the current room status (called `croom`) [see load_new_room](https://github.com/adventuregamestudio/ags/blob/v.3.6.0.14/Engine/ac/room.cpp#L606). Need to add the new object & hotspots names there too.
3. These fields should be added to the [game save format](https://github.com/adventuregamestudio/ags/blob/master/Engine/game/savegame_components.cpp).

---

UPDATE:

Regarding possible optimizations. Since not every game changes these names, or all of these names, and there may be a lot of total objects and hotspots accumulated after large number of rooms, we may consider to not save the names when they are identical to originals from the "static" game data.
Note that this is not critical, as it's unlikely that these names would take too much extra disk space (we are speaking of kilobytes or maybe tens of kilobytes in a bigger game).

So, speaking theoretically: because empty string is also a valid value it cannot be used to determine "use defaults" when storing this field. There are two options I see now:
* this has to be set as a flag in some flags field;
* alternatively, this may be saved as a custom property with some kind of "internal name" which otherwise cannot be reproduced. As custom properties are maps, they are only applied if the keys exist.

But like mentioned above, this may be an unnecessary optimization. The names will take place in memory anyway, and kilobytes of disk space is not too serious nowadays.

ivan-mogilko:
Resolves #979.

Hotspot and Object.Name property can now be set from script.

Any internal engine's action retrieving an object or hotspot's name should be using current name from the RoomStatus, rather than the default name from the room file.

These properties are now saved when the game is saved.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -144,7 +144,8 @@ enum GameDataVersion
     kGameVersion_350            = 50,
     kGameVersion_360            = 3060000,
     kGameVersion_360_11         = 3060011,
-    kGameVersion_Current        = kGameVersion_360_11
+    kGameVersion_360_16         = 3060016,
+    kGameVersion_Current        = kGameVersion_360_16
 };
 
 // Data format version of the loaded game
--- a/Common/game/roomstruct.cpp
+++ b/Common/game/roomstruct.cpp
@@ -178,13 +178,7 @@ void RoomStruct::InitDefaults()
     MessageCount    = 0;
 
     for (size_t i = 0; i < (size_t)MAX_ROOM_HOTSPOTS; ++i)
-    {
         Hotspots[i] = RoomHotspot();
-        if (i == 0)
-            Hotspots[i].Name = "No hotspot";
-        else
-            Hotspots[i].Name.Format("Hotspot %zu", i);
-    }
     for (size_t i = 0; i < (size_t)MAX_ROOM_OBJECTS; ++i)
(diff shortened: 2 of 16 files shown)


ericoporto:
Tested some old games and they seem to work unnafected by this change. :)

Also made a small test game and it appears to work.

I would only suggest that on the attribute "comment description" it changes from `/// Gets the name of the hotspot.` to `/// Gets/sets the name of the hotspot.`.

ivan-mogilko:
Had to rebase because there was a bug in master that prevented from loading saves in 3.5.1 games.
Also added one fix.

ivan-mogilko:
Rebased again. I confirm that after #1515 the hotspots no longer have redundant "hotspot x" names in the new rooms.

PR #1515, opened 2022-01-23
3.6.0: create blank room programmatically instead of using a binary resource

ivan-mogilko:
While working on #1512 I realized that the default room template is present in the Editor as an actual binary room file, which makes it quite inconvenient to upgrade along with AGS.

This pr removes the blank.crm from the editor, and instead generates an empty minimal room programmatically. I think this will make things easier at least until we have an open room format (#1278), after which we may review this situation.


ericoporto:
Ah, I was really curious, if the hotspot names you removed were going to be or not in blank.crm, but this is MUCH better. :)

Editor/AGS.Native/agsnative.cpp
```diff
@@ -3827,60 +3827,67 @@ PInteractionScripts convert_interaction_scripts(Interactions ^interactions)
...
+    if (!rs.RegionMask)
+        rs.RegionMask.reset(BitmapHelper::CreateClearBitmap(rs.Width / rs.MaskResolution, rs.Height / rs.MaskResolution, 0, 8));
+    if (!rs.WalkBehindMask)
+        rs.WalkBehindMask.reset(BitmapHelper::CreateClearBitmap(rs.Width, rs.Height, 0, 8));
+
+    calculate_walkable_areas(rs);
+
+    rs.BackgroundBPP = rs.BgFrames[0].Graphic->GetBPP();
+    for (int i = 0; i < 256; ++i)
        thisroom.Palette[i] = thisroom.BgFrames[0].Palette[i];
```
rofl0r, line 3873: hmm, what happened here with this `^M`. is that a DOS linefeed?

ivan-mogilko: Recommited.

ivan-mogilko:
By the way, another consequence of this change is that the blank room is created with the default background and masks size equal to game resolution, instead of hardcoded 320x200.
This may not be a big deal as normally one would import their own background, but may speed up the process of setting up a quick test room.

ericoporto:
Just for reference, in ags4 this has been improved to rely only on managed code in commit
 https://github.com/adventuregamestudio/ags/commit/e235690481f21d44e96e452e0bf4c4c342136a8c

PR #1516, opened 2022-01-24
New compiler: Fix some bugs in constant handling

fernewelten:
- `while (true)` should not short-circuit the `while` body
```
function game_start()
{
    int i = 5;
    while(true)
    {
        i--;
        if (i <= 0) break;
    }
    if (i ==  5)
        AbortGame("'while' was short-circuited");
}
```

- `do … while (false)` should not have a jump back to the start of the `do`

Can't really test that except by examining the generated code: If there __is__ a jump back to the start of the do, it won't ever be executed because the `do while` condition is `false`.

- `if (expr)` should ensure that type of the condition is `int`
```
float TestFunc()
{
    if (TestFunc()) // ← Should not compile
        return 7.0;
    return 0.0;
}
```


fernewelten:
(Had to redo this because one of the tests failed. Force-pushing to avoid spurious commits.)

PR #1518, opened 2022-01-25
Feature: Log in the Editor

### LINKED ISSUE #1362: Read engine Log in the Editor

ericoporto:
Targeting 3.6.1, since 3.6.0.X it has a cool logging feature added I made this so it's easier to reach it through the Editor. On Editor code side, the panel code is not great, I don't know how to set things like it's initial position or how to make it stay docked. 

Fix #1362 

This adds a new log output in the engine, debugger, and renames some references to the Editor in the engine code as debugger, as discussed in the issue, and since any other software could be a debugger that could talk through the same interface.

- AGS Types has a new type for storing the config of the debug, log level and log groups are added as enums
- Editor has a new component for the debugger, it's config is serialized/deserialized there in ~~Game.agf~~ Game.agf.user
- The panel when active, can be configured by using the properties pane
  - you can either change how the log is configured through command line (meaning clicking play button to test the game), or later through dynamic filters

image


ericoporto:
Hey, if I were to finish this thing up so we can have this, apart from modifying initialization so we can have the full boot log on the Editor, what else is needed?

I think the boot log of AGS is really useful to pickup errors when some device fails to initialize for some reason.

ericoporto:
> Could you post an explanation of these controls?

So previously I had two comboboxes, one for groups and other for log level, and you were supposed to select a group and then it's loglevel, and so on to change them. I unfortunately had way too much difficult dealing with the managed objects group, it clobbered the log no matter what I tried. So instead I decided to redo the whole thing and now I am only supporting two groups in the UI, Game and Script.

If in the future we decide we want more groups, then we can think on how to accomplish that, but for now, I think this suffices.

log_level_selection

So now, both comboboxes allow setting the minimal loglevel, one for the Game group and other for the Script group. I think these are the more useful ones in a day to day use of the engine.

I still don't know how to deal with theming, since AGS themes are panel specific (unless #1416 gets an ags3 version), and there's no public repository for themes that I could update the existing themes to take into account this extra panel.

ivan-mogilko:
I still don't have full information on how these controls suppose to work. Do they change the levels that will be shown from now on, or show full log from the beginning, but filtered?

> So previously I had two comboboxes, one for groups and other for log level, and you were supposed to select a group and then it's loglevel, and so on to change them. I unfortunately had way too much difficult dealing with the managed objects group, it clobbered the log no matter what I tried.

Perhaps I don't understand that issue, but won't it be possible to add a "disabled" log level, that would completely cut the group off?

In general, don't remember if I mentioned this before, but I may propose a separate dialog with a property grid, where all groups are listed and you may choose levels for each of them. This dialog may be shown by a simple "..." button located in the corner of the log panel.

ericoporto:
> Regarding these controls; do they change the levels that will be shown from now on, or show full log from the beginning, but filtered?

The log is accumulated to a buffer in the Editor, older log is discarded automatically and replaced by newer entries. At any point you can filter this buffer per group and get only the desired output. This is similar to how it works with logcat.

> Perhaps I don't understand that issue, is not there a way to fully disable the group? is this a problem with controls, or a problem with engine sending too much data at once?

Currently it's disabled (see Tasks.cs), but there's no way to change the log group config while the engine is running. Overall, when I played around there was just so much data flowing from Manobj log that it would slow the Editor to a crawl, making it really hard to debug the Editor in VS when testing a game.

Also, I am not sure what happens, but it appears that when there's a lot of data the debugger gets hard to use. It's curious that it did not happen in my first approach that used stdout directly, so I don't know what is it... Maybe it's the XML parser?

ivan-mogilko:
> Currently it's disabled (see Tasks.cs), but there's no way to change the log group config while the engine is running.

If it's necessary, engine may be also adjusted to change log levels at runtime, by command from the debugger.

Also, I thought that manobj group is disabled in the standard engine build, to improve perfomance. Is it enabled in the build that we distribute?

ericoporto:
> Also, I thought that manobj group is disabled in the standard engine build, to improve perfomance. Is it enabled in the build that we distribute?

It's not but it's there in the list of possible groups in the command line, can I throw an `#ifdef` there to not have it written there when it's not built along? I was testing with it because I thought I was supposed to support there but if I can ignore it I will.

> If it's necessary, engine may be also adjusted to change log levels at runtime, by command from the debugger.

I will add a command for this then.

ivan-mogilko:
> It's not but it's there in the list of possible groups in the command line, can I throw an #ifdef there to not have it written there when it's not built along? I was testing with it because I thought I was supposed to support there but if I can ignore it I will.

I believe it may be completely removed from the editor's log, at least for now. It's not very well suited for runtime debugging, and imho the debugging of managed object pool should be done differently, maybe by sending overall stats once per X ms, or something like that.

ivan-mogilko:
What is the state of this pr? I had an impression that it's practically working.

ericoporto:
Did you try to use it? If I log too much the debug seems to get into an unreliable state. It does have to be a lot though, but not sure how to workaround it.

ericoporto:
I looked into the properties window, going that route it's basically

- Add LogLevel to AGS Types Enums
- Add new EngineLog class in AGS Types to hold the configurable properties of the Editor
- ditch the log groups as enums approach, and instead have an individual property per log group in the new EngineLog class in AGS Types
- figure a way to store the instance of the EngineLog class, it will need to be serialized and also be accessible to everyone (LogBuffer, Tasks.cs, ...)
- LogPanel has to be refactored as Pane, so that it can be considered active to show the properties page

ericoporto:
@ivan-mogilko when you have some time, can you give this a try? I reworked so that the panel is only a log, and when it's selected, the property grid now shows it's properties. The log config is serialized and deserialized in it's own component.

ivan-mogilko:
> @ivan-mogilko when you have some time, can you give this a try?

I think i will be able to do this in the end of the week.

ivan-mogilko:
I quickly tried this out, did not go through the whole implementation.

My biggest question at the moment is, what were the reasons for making it a component. Is this to be able to show it on a property grid, or serialization, or else?

In regards to serialization, I believe it should write the setting into the workspace settings (Game.agf.user). Otherwise it will modify the game file every time you change log levels. I looked around and it seems for that you need to implement IPersistUserData interface. Unfortunately there does not seem to be a way to completely prevent editor from writing component's name into Game.agf, but if ToXml and FromXml are not overridden then it will be just an empty xml node (compare with e.g. "Welcome" component).

Another thing, in my opinion the "Debug log" menu item should be in the Window menu rather than the Help menu.

ericoporto:
> My biggest question at the moment is, what were the reasons for making it a component. Is this to be able to show it on a property grid, or serialization, or else?

To show it on a property grid.

> In regards to serialization, I believe it should write the setting into the workspace settings (Game.agf.user). ...  I looked around and it seems for that you need to implement IPersistUserData

Ah, that's doable, I can change to use that!

> Another thing, in my opinion the "Debug log" menu item should be in the Window menu rather than the Help menu.

I agree, but components get thrown there by default. **I could not figure how to place it in the Window menu**.

> in regards to commits, perhaps if you do "use single function to send messages" first, and "log output" second, then that ugly old code with sprintf/strlen will be avoided.

Ah, that is an easy change, I can do this.

ivan-mogilko:
I keep looking into this to see what may be improved.

There's a method `void PropertyChanged(string propertyName, object oldValue)` in IEditorComponent, that may be overriden to respond to the object's property change in the grid. I think it may be used instead of manually raising PropertyChanged in each setter in the DebugLog.

ericoporto:
Thanks! I looked for a method for this as I felt it probably already existed/was used somewhere but did not find at the time. I will rewrite to use it instead!

ivan-mogilko:
Another thing that I'd like to suggest is to make log group/level list into a struct, then you can have 2 objects with similar contents in the `DebugLog` instead of 2 sequences of properties with slightly different names.

To give an example:
```
    public class LogGroups
    {
        [TypeConverter(typeof(EnumTypeConverter))]
        [Description("Define the log level of Main engine, which gives information on general ags behavior.")]
        public LogLevel Main
        {
            get; set;
        }

        // Game log group
        [TypeConverter(typeof(EnumTypeConverter))]
        [Description("Define the log level of game related methods, which gives information on internal Engine API and behavior.")]
        public LogLevel Game
        {
            get; set;
        }

        // Script log group
        [DisplayName("Script")]
        [Description("Define the log level of messages generated by user script, using System.Log method.")]
        [TypeConverter(typeof(EnumTypeConverter))]
        public LogLevel Script
        {
            get; set;
        }

        // Sprite Cache log group
        [DisplayName("Sprite Cache")]
        [Browsable(false)]
        [Description("Define the log level of messages generated by sprite cache messages.")]
        [TypeConverter(typeof(EnumTypeConverter))]
        public LogLevel SpriteCache
        {
            get; set;
        }

        // Managed Objects log group
        [DisplayName("Managed Objects")]
        [Browsable(false)]
        [Description("Define the log level of messages generated by managed objects.")]
        [TypeConverter(typeof(EnumTypeConverter))]
        public LogLevel ManObj
        {
            get; set;
        }

        [DisplayName("SDL")]
        [Browsable(false)]
        [Description("Define the log level of messages generated by SDL backend.")]
        [TypeConverter(typeof(EnumTypeConverter))]
        public LogLevel SDL
        {
            get; set;
        }

        public LogLevel GetGroupLevel(LogGroup group)
        {
            switch (group)
            {
                case LogGroup.Main:
                    return Main;
                case LogGroup.Game:
                    return Game;
                case LogGroup.Script:
                    return Script;
                case LogGroup.SprCache:
                    return SpriteCache;
                case LogGroup.ManObj:
                    return ManObj;
                case LogGroup.SDL:
                    return SDL;
            }
            return LogLevel.None;
        }
    }
```

The members in the DebugLog class must be declared using "ExpandableObjectConverter" type converter: in such case user will be able to expand the object like a group of properties.
```
        private LogGroups _logOutput = new LogGroups();
        private LogGroups _logFilter = new LogGroups();

        [DisplayName("Log Output")]
        [TypeConverter(typeof(ExpandableObjectConverter))]
        public LogGroups LogOutput
        {
            get { return _logOutput; }
            set { _logOutput = value; }
        }

        [DisplayName("Log Filter")]
        [TypeConverter(typeof(ExpandableObjectConverter))]
        public LogGroups LogFilter
        {
            get { return _logFilter; }
            set { _logFilter = value;}
        }
```

Maybe something else is missing here, but it looks like a necessary minimum to work.

One problem remaining would be the serialization, because AGS cannot serialize inner classes like that. So either something needs to be adjusted for serialization, either in these classes/properties or in Serialize Utils, or a custom serialization is used, where you would manually create 2 xml nodes and pass these 2 objects to serialize method separately; this is similar to how Sprites data is serialized in the game.agf, if I remember correctly.

ivan-mogilko:
About this -

> > Another thing, in my opinion the "Debug log" menu item should be in the Window menu rather than the Help menu.

> I agree, but components get thrown there by default. I could not figure how to place it in the Window menu.

It does not look like there's a default for components, each component may request its own menu group. For example, FileCommands, Room and Script components register in the "File" menu.

But the Window menu is bit special, as other items are added outside of the component system, and if you simply add component's item there using same method (`_guiController.AddMenuItems`), the menu gets broken (the component's item does not show up, and neither do most of other panels). So this might require either an adjustment to the menu system, or a different approach.

ericoporto:
> Another thing that I'd like to suggest (at least to try) is to make log group/level list into a struct

Hey, you played with the code? I tried exactly your suggestion and it worked! The only curious thing is on the PropertyChanged event, is the name of the property we get is the property of the LogGroups, and there's nothing to disambiguate between them, but since I wasn't using this information, it does not matter. I did so in a separate commit to be easier to see it.

> But the Window menu is bit special, as other items are added outside of the component system, and if you simply add component's item there using same method (_guiController.AddMenuItems), the menu gets broken

An alternative would be pushing it in the Build menu, since it already holds the Debug/Run stuff!

ivan-mogilko:
Hmm, this is interesting, maybe I did something wrong so serialization did not work correctly for me.

Another thing about properties, in theory you can create properties dynamically too. Here's an interesting reference:
https://stackoverflow.com/questions/931644/net-property-grid-is-there-a-way-to-let-the-grid-manipulate-object-in-differen

Not suggesting to do so right away, but theoretically editor could query engine for the log groups instead of hardcoding them.

ericoporto:
> Hmm, this is interesting, maybe I did something wrong so serialization did not work correctly for me

Can you elaborate on this? I downloaded the AGS built from the CI to check if I had forgotten to push something, to try it, and the serialization is working with the generated AGS Editor.

> Not suggesting to do so right away, but theoretically editor could query engine for the log groups instead of hardcoding them

Uhm... I vaguely understand this, I already added a new browsable attribute to enums, so they could be leveraged for the enums of log groups that I want to hide (ManObj and SprCache, I really don't want to be available). It does look a bit beyond my knowledge right now though.

ivan-mogilko:
> > Hmm, this is interesting, maybe I did something wrong so serialization did not work correctly for me

> Can you elaborate on this? I downloaded the AGS built from the CI to check if I had forgotten to push something, to try it, and the serialization is working with the generated AGS Editor.

I meant my own experiments.

ericoporto:
Just splitted the commit `[Editor: split debug messages if they are joined](https://github.com/adventuregamestudio/ags/pull/1518/commits/31e069f3c3d48394b8ead3b21c009a48fe5bcc4e)` in case we wanted to apply it sooner. In theory, any such situation should crash the editor, but supposing it doesn't, it could in theory result in the Editor not getting a message from the Engine - because it was packed along with a previous message.

ericoporto:
This is probably done, but I don't know where to merge this, so I will keep it as draft until there's a destination.

ericoporto:
I think the gitignore commit (1583e14a3ae9e3324f9e00f7b33e0e4e156ba118) is fine to merge in master but is this something to push to ags4?

ivan-mogilko:
Since 3.6.0 is taking extra time to finish, and if this addition does not conflict with anything, or may be easily disabled, I think we could merge it into 3.6.0 too.

ericoporto:
I can rebase and revise this if we want to merge this, I just need a day or two for this. :)

Engine/debug/debug.cpp
```diff
@@ -89,6 +89,25 @@ int first_debug_line = 0, last_debug_line = 0, display_console = 0;
...
+    String messageToSend = String::FromFormat(R"(<?xml version="1.0" encoding="Windows-1252"?><Debugger Command="%s">)", command.GetCStr());
+#if AGS_PLATFORM_OS_WINDOWS
+    messageToSend.Append(String::FromFormat("  <EngineWindow>%d</EngineWindow> ", (int)sys_win_get_window()));
+#endif
+
+    for(const auto& tag_value : tag_values)
+    {
+        String tag_line = String::FromFormat("  <%s><![CDATA[%s]]></%s> ",
+                                             tag_value.first.GetCStr(), tag_value.second.GetCStr(), tag_value.first.GetCStr());
+        messageToSend.Append(tag_line);
```
ivan-mogilko, line 103: You could also use AppendFormat here, for a bit more perfomance.

Engine/debug/debug.cpp
```diff
@@ -108,13 +108,41 @@ void send_message_to_debugger(const std::vector<std::pair<String, String>>& tag_
     editor_debugger->SendMessageToEditor(messageToSend.GetCStr());
 }
 
+String ToString(int v)
```
ivan-mogilko, line 111: There's a common function in StrUtils that does this:
https://github.com/adventuregamestudio/ags/blob/24fdb20eadf56f9a6d9a432d15491f8741540623/Common/util/string_utils.h#L40

ivan-mogilko:
On a more close inspection, I have a serious concern about how the log is stored and displayed. Reading this code in LogPanel and LogBuffer, and seeing how the LogPanel reacts to filter changes and entry addition...

Do I understand correctly, that -
* the log is *fully* copied into the text box when the log filter changes?
* the log is also *fully* copied into the text box when any entry is added?

ericoporto:
You need to check the LogBuffer.cs, it's a circular buffer of a fixed size. New entries are appended there. The LogBuffer has a debounce it uses for when the LogBuffer should change it's output, that should smash together changes that are close in time, and yeah, it copies all of this buffer when it does print.

ivan-mogilko:
> The LogBuffer has a debounce it uses for when the LogBuffer should change it's output, that should smash together changes that are close in time

I could not understand this unfortunately, what does "smash together changes" means, and where can this be found in code?

> it copies all of this buffer when it does print.

Copying whole buffer contents when only 1 entry is added sounds like a lot of extra work...

I may have forgotten this, but is the displayed log basically limited to the size of the buffer? So, if it gets fulled, I can no longer scroll back to the top of the log?

ericoporto:
It works like this, when a new entry is added to the LogBuffer, the LogBuffer mark it's modified flag, and the LogBuffer kills it's debouncing timer if it has started and starts again. This means any entry in a 10ms interval, will be "ignored" and say we get an event only after no new entry is added in a 10ms interval - and then we get an update event.

But, this is not all, at the same time, the LogPanel has a timer that goes on for a fixed timer interval repeatedly and checks the modified flag of the LogBuffer, if it's on, it will also update - so we get intermediary updates.

https://github.com/adventuregamestudio/ags/blob/e5efbcadb91f8ce4c4597c18f1724dfc91a489c8/Editor/AGS.Editor/Panes/LogPanel.cs#L82

> Copying whole buffer contents when only 1 entry is added sounds like a lot of redundant work...

It's not that much because we get a limited amount of entries (4096), and the decoupling occurs if someone wants to implement a viewer that is not a textbox one day.

ivan-mogilko:
> > Copying whole buffer contents when only 1 entry is added sounds like a lot of redundant work...

> It's not that much because we get a limited amount of entries (4096)

Well, it is still alot of unnecessary work. But it is not only that. Since it copies whole text each time, the text box looses its current scroll position. If I try to read the log while the game runs, I mostly cannot do that, because the text gets replaced and it brings me back to the bottom all the time.

Is it possible to optimize this by changing how update event works? For example, if event could tell the range of entries, and the log panel would request only that range and append, instead of clearing and resetting whole test? Although tbh I don't know how text box works in that case (whether it also scrolls to the bottom or not).

ivan-mogilko:
Hm, okay, I see, there are two issues:

1. The potential performance issue, because the whole text is recreated even when only 1 log entry is added.
2. A very serious usability issue: when the Text property of the text box is set, it looses its current scroll position. If I try to read the log while the game runs, I mostly cannot do that, because the text gets replaced and it brings me back to the bottom all the time. This makes it impossible to use the log, unless I pause the game by a breakpoint, for example.

ivan-mogilko:
On a side note, above you mentioned that there are some things that could be applied separately from the log panel.
I just got a crash that you referenced, when the message from the engine contains 2 <xml> entries, because I ran old Editor with the newer engine (not sure if it could have happened earlier).

Could we merge these changes separately for now, because they are useful on their own?

I think first 3 commits are easily acceptable:
* split debug messages if they are joined
* gitignore change
* use a single function to send messages to editor debugger

ericoporto:
> Could we merge these changes separately for now, because they are useful on their own?

I opened #1866 for these. The rest is better left for ags4 or some other time.

ericoporto:
It appears this won't be the actual implementation used, so I am closing this to reduce noose. I will keep the branch in my fork in case it's needed later.

PR #1519, opened 2022-01-28
Script API: added Character.IdleAnimationDelay and Cursor.AnimationDelay

### LINKED ISSUE #1517: Remove hardcoded animation speed modifier for idle views and cursors
In the engine there are two obscure hardcoded modifiers to the animation speed:

1. Idle views, when animating their animation speed (frame delays) is actually calculated as `Character.AnimationSpeed + 5`.
as mentioned in this old forum thread: https://www.adventuregamestudio.co.uk/forums/index.php?topic=36779.0
and seen in this code: https://github.com/adventuregamestudio/ags/blob/v.3.6.0.15/Engine/ac/characterinfo_engine.cpp#L376-L379

2. Mouse cursors, when they have assigned views, increase each frame delay also by 5.
as seen in this code: https://github.com/adventuregamestudio/ags/blob/v.3.6.0.15/Engine/ac/draw.cpp#L2519

I don't know if these are coincidence, or both a part of some ancient design idea; but these are causing problems and may be a source of immense frustration to users, as these are obscure behavior modifiers, and cannot be directly changed (and not documented, i presume).

The proposal is to remove both of these mods and make idle view and cursor animations follow actual animation parameters.

For backward compatibility we may have a setting enabled for older games.

ivan-mogilko:
Resolves #1517.

**1.**
Adds `Character.IdleAnimationDelay` property in the script and editor properties.
The name is complying to previously added SpeechAnimationDelay (also related property is called AnimationDelay internally in the editor, not "speed").

For older games engine assigns this property to (Character.AnimationSpeed + 5) for backward compatibility.

**2.**
Adds `Cursor.AnimationDelay` property in the editor. Added 3rd argument to `Mouse.ChangeModeView` script function, now function is:
`void ChangeModeView(CursorMode, int view, int delay = SCR_NO_VALUE);`
Skipping `delay` argument (or passing SCR_NO_VALUE) will result in keeping previous delay.

For compatibility default delay value in the editor and game data is 5.

### DIFF
--- a/Common/ac/characterinfo.cpp
+++ b/Common/ac/characterinfo.cpp
@@ -11,9 +11,9 @@
 // http://www.opensource.org/licenses/artistic-license-2.0.php
 //
 //=============================================================================
-
-#include <string.h>
 #include "ac/characterinfo.h"
+#include <string.h>
+#include "ac/game_version.h"
 #include "util/stream.h"
 
 using AGS::Common::Stream;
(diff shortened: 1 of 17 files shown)


ivan-mogilko:
Added a method for setting cursor animation delay (see first post for description).

PR #1523, opened 2022-02-01
Feature: Web (Emscripten) build through the Editor

ericoporto:
This adds support for generating a web build through the Editor.

You will need to extract the contents of the generated artifact and place in the Editor directory in a folder named `Web/` 

I did additional changes to the CI, which are explained in the commit.

I have not wired yet the installer and release to actually use the artifact. I want to let the Emscripten builds happen a bit more on the CI before adding the dependency on the release process.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -179,15 +179,17 @@ build_android_task:
 
 build_emscripten_task:
   container:
-    image: ubuntu:focal
-  install_packages_script: |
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
Is there anything else required after editor completes compilation?

After building, in Compiled/Web I open index.html, it displays "Preparing... (0/1)", and does not progress (waited for several minutes).

ericoporto:
I have been thinking if I should or not make a mini webserver starter and stopper for the Editor so these are easier to test. I haven't read on, but I imagine it's something that's baked on .NET so should not be hard.

Issue #1525, opened 2022-02-03
`Hotspot::SetProperty` claims to be `GetProperty`


fernewelten:
**Describe the bug**
Here's a tiny one:

Whenever `Hotspot::SetProperty` doesn't find the property name in the schema, it claims to be `GetProperty`. 

**AGS Version**
Build 3.5.1.14

**To Reproduce**
```
function game_start()
{
    …
    hHotspot10.SetProperty("Fush", 99);
    …
}
```

**Expected behavior**
An error message that begins with "SetProperty".

**Screenshots**
The error message that I did get, beginning with "GetProperty".
![image](https://user-images.githubusercontent.com/41213694/152260477-91496ffe-1981-4dd7-a5b5-e1a061272799.png)

**Desktop (please complete the following information):**
 Windows 10, German edition

**Smartphone (please complete the following information):**
n/a

ivan-mogilko:
Hm, yes, this is because the error is reported not by particular script function, but by the function that accesses the property schema.

fernewelten:
By the bye: The function that accesses the property probably knows what it is looking for, so no need for "__no such__ property". This could probably be reworded,  "SetProperty: Did not find the property 'Fush' in the schema." So when the coder has used the description instead of the name by mistake, they can see what they've done wrong directly.

PR #1529, opened 2022-02-03
Reword error messages for property lookup

fernewelten:
Name the property. Don't guess and name the command that looks up the property. See #1525 for details.


ericoporto:
I think the new text is better and looks like a really smart approach. We should probably apply this to master branch too.

PR #1531, opened 2022-02-05
Refactor OpenAlDecoder and video playback

ivan-mogilko:
This is a serious refactor of the new audio playback and video playback, meant to simplify the code use and maintenance in the longer run.

Reasons:
* OpenALDecoder was introduced when moving to SDL2 backend with SDL_Sound and OpenAL libs. It is combining functionality of decoder and "player" (OpenAL source) in one object, which would be inconvenient if we need to replace one of these parts, or need to use only one of these parts elsewhere.
* AGS supports 2 video types (OGV Theora and Flic), and there are two completely separate playback implementations, which use different options, and may behave slightly differently (and have separate mistakes).
* The OGV playback is run inside the Apeg library, which not only decodes, but also runs the loop and plays sound on its own. Only final render is done by the callback in the engine's code. This makes it difficult to control.
* For sound Apeg now has its own implementation using OpenAL, which duplicates the audio code in the engine.
* Video playback uses alot of global variables, which would prevent from having multiple videos playing at the time.

Refactor in audio:
* OpenALDecoder is split into two classes: SDLDecoder which does decoding (uses SDL_Sound library) and OpenAlSource which plays the provided sound chunks using OpenAL.
* Now bit more sophisticated AudioCoreSlot object manages decoder and player components, retrieves data from one and passes into the second. The way this is organized permits optional replacement of either decoder or player if such need arises in the future; or adding any additional process in between.

Refactor in video:
* Made VideoPlayer class that implements options, playback loop, frame render, and audio playback using OpenAlSource; leaves video decoding to the derived classes.
* FlicPlayer is a child class that decodes frames using flic functions from allegro4.
* TheoraPlayer is a child class that decodes frames and audio using apeg library.
* Unified video flags and behavior. The script functions that have different input parameter meanings now convert these to the standard.
* Completely removed OpenAl usage from apeg sources. 
* Possible TODO: apeg sources may be further cleaned up, as parts of it related to playback loop are no longer necessary. We may strip everything that is not related to decoding.

NOTE: I tried to keep audio & video working at each commit, although some intermediate commits may contain minor mistakes.

NOTE: if someone has a concern that it's too much for one pr, i may move audio and video changes apart; but audio is only 1 commit and changes were started primarily for video code improvement.


ericoporto:
Hey, in the `#else` case of `video.cpp` (when the build option with no video is used), you need `void video_shutdown() {}` or the linker will fail on `video_shutdown()` when building AGS with `AGS_NO_VIDEO_PLAYER` option.

ivan-mogilko:
> Hey, in the `#else` case of `video.cpp` (when the build option with no video is used), you need `void video_shutdown() {}` or the linker will fail on `video_shutdown()` when building AGS with `AGS_NO_VIDEO_PLAYER` option.

So, this is what the error was, fixed now.

ericoporto:
More a question, there are some files here that we haven't touched for 11 years: https://github.com/adventuregamestudio/ags/tree/master/Engine/libsrc/apeg-1.2.1/audio

I believe the reason we haven't touched them is because we don't use them. Would it make sense to take this opportunitty to also delete these files?

ericoporto:
We should merge this. Things we can adjust later.

- adjust the `fli_speed` to a variable in the VideoPlayer object
- look into audio/video sync in the theora player
- remove unused files in apeg library

ivan-mogilko:
Rebased, squashed some commits, also this:
> adjust the fli_speed to a variable in the VideoPlayer object

PR #1532, opened 2022-02-05
Installer CI: add Web component to releases

ericoporto:
image

Adds the web build in releases, in the Editor archive under `Web/`, and as the Web component in the installer.

Edit: just squashed the commits with the adjustment to the review.

### DIFF
--- a/Windows/Installer/ags.iss
+++ b/Windows/Installer/ags.iss
@@ -52,6 +52,7 @@ ComponentMain=Main files
 ComponentEngines=Engines
 ComponentEngineDefault=Runtime engine for MS Windows
 ComponentLinuxBuild=Linux build component
+ComponentWebBuild=Web build component
 ; ComponentDemoGame=Demo Game
 InstallOptions=Install options
 InstallVCRedist=Install {#VcRedistName}
@@ -64,6 +65,7 @@ Name: "main"; Description: "{cm:ComponentMain}"; Types: full compact custom; Fla
 Name: "engine"; Description: "{cm:ComponentEngines}"; Types: full compact custom; Flags: fixed
 Name: "engine\default"; Description: "{cm:ComponentEngineDefault}"; Types: full compact; Flags: exclusive
 Name: "linux"; Description: "{cm:ComponentLinuxBuild}"; Types: full custom
+Name: "web"; Description: "{cm:ComponentWebBuild}"; Types: full custom
 ; Name: "demogame"; Description: "{cm:ComponentDemoGame}"; Types: full custom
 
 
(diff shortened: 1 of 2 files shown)


Windows/Installer/ags.iss
```diff
@@ -111,6 +113,8 @@ Source: "Source\Linux\ags64"; DestDir: "{app}\Linux"; Flags: ignoreversion; Comp
 Source: "Source\Linux\lib32\*"; DestDir: "{app}\Linux\lib32"; Flags: ignoreversion recursesubdirs createallsubdirs; Components: linux
 Source: "Source\Linux\lib64\*"; DestDir: "{app}\Linux\lib64"; Flags: ignoreversion recursesubdirs createallsubdirs; Components: linux
 Source: "Source\Linux\licenses\*"; DestDir: "{app}\Linux\licenses"; Flags: ignoreversion recursesubdirs createallsubdirs; Components: linux
+; Web build components
+Source: "Source\Web\*"; DestDir: "{app}\Web"; Flags: ignoreversion; Components: web
```
morganwillcock, line 117: If the file names are fixed it is safer to explicitly list each component to make sure that the installer build fails if one of them is missing.

ericoporto: Adjusted! Seems to be alright!

ericoporto:
sorry, clicked a thing by accident :0

ivan-mogilko:
Perhaps a separate issue, but there's one thing that bothers me.

With the installer user can select whether to install certain components (for Linux, Web build, etc).
With the zip archive user cannot do that. So far we only had 1 extra component for Linux build, but now we are going to have also Web and Android, and could also be one for mac in the future.
Do you think this may become a problem for some, and are there any good solutions?

ericoporto:
Not sure what you mean but like the user can delete the subdirectory and the rest of the Editor keeps working fine.

In the past I suggested to move these under a platforms dir with os-arch dirs inside, so it was more obvious they were pieces (like `Platform/Linux/amd64`, `Platform/Windows/x86`, ...).

It's possible to make something in the Editor that would download a component from GitHub Assets by request instead, using GitHub API, but I don't think we should go there yet - it's simply not that big. If we do go there, then the complicated part is basically if the person placed the Editor in like Program Files, I am not sure we can write this new download package there, instead we need to go to AppData (and probably Windows will redirect our file writing there if we don't), and then what is in this directory is either AGS version independent or we have to manage this. If we do want to have these packages installed side-by-side, we need to create something like the Visual Studio installer software that allows downloading VS packages in Program Files.

ericoporto:
Overall, we may do something about it eventually, but it may not be trivial and I do think if we do we should still keep offering a portable zip, with batteries included that one can pickup and use.

Issue #106, opened 2013-12-21
Rename psp_* variables


ghost:
There's a whole bunch of global variables in the engine named with "psp_" prefix after PSP port. Since most of them is used generally now, they should be renamed, with this prefix removed, to avoid confusion.

ericoporto:
https://github.com/adventuregamestudio/ags/blob/efa89736d868e9dfda4200149d33ba8637746399/Engine/gfx/gfx_util.cpp#L122-L128

this is one of the few `psp_` var on the outside of config and platform specific stuff, unfortunately it's in a hack. I tried to remove and run in a few android devices and it seems to not cause any disturbance, unfortunately I am not sure any of the games I tested could cause the need for such hack.

ivan-mogilko:
> https://github.com/adventuregamestudio/ags/blob/efa89736d868e9dfda4200149d33ba8637746399/Engine/gfx/gfx_util.cpp#L122-L128
> 
> this is one of the few `psp_` var on the outside of config and platform specific stuff, unfortunately it's in a hack. 

Lol, reading this hack, it makes no sense whatsoever. Because `surface_depth = ds->GetColorDepth();`, so it basically compares `ds->GetBPP() < ds->GetColorDepth()` which is *always true*, as GetBPP returns number of bytes, while ColorDepth returns number bits.

So I say, remove this. If there are any problems, they likely may be fixed differently.

ericoporto:
Some floating around psp_variables have been axed in recent commits

psp_gfx_renderer : https://github.com/adventuregamestudio/ags/commit/083d8496f2d7c0eedc31c068ddf27720dee4ff2b

psp_clear_cache_on_room_change : https://github.com/adventuregamestudio/ags/commit/7abbd673e910289511bcca3101abd03101599eb0

psp_video_framedrop: https://github.com/adventuregamestudio/ags/commit/6c10e6e5e16ec70f5d5986b23217dc5ebf13a70c 

I wonder if it makes sense to place the remaining ones in an object instead of using individual global vars since most of the remaining use cases is contained somewhat in their own platform driver code.

ivan-mogilko:
> I wonder if it makes sense to place the remaining ones in an object instead of using individual global vars

I think they may be put into GameSetup struct.

ericoporto:
> I think they may be put into GameSetup struct.

There are two TO-DOs in the GameSetup header that made me unsure if that's where these should go.

ivan-mogilko:
[...]

PR #1535, opened 2022-02-09
Engine: add clear cache on room change from Android port to config

ericoporto:
**I don't know if we want this or not**

This was a setting that was added to the psp port and to the mobile ports probably due to some memory restriction in these devices back in the day. It's still available in the android.cfg file. If we don't want, then the right PR is the one that removes the psp variable.

We probably need to figure out what goes and what doesn't for #106 and #104

### DIFF
--- a/Engine/main/main.cpp
+++ b/Engine/main/main.cpp
@@ -154,6 +154,7 @@ void main_print_help() {
         "Usage: ags [OPTIONS] [GAMEFILE or DIRECTORY]\n\n"
           //--------------------------------------------------------------------------------|
            "Options:\n"
+           "  --clear-cache-on-room-change Clears sprite cache on every room change\n"
            "  --conf FILEPATH              Specify explicit config file to read on startup\n"
 #if AGS_PLATFORM_OS_WINDOWS
            "  --console-attach             Write output to the parent process's console\n"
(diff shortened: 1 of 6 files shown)


Engine/main/main.cpp
```diff
@@ -313,6 +314,10 @@ static int main_process_cmdline(ConfigTree &cfg, int argc, char *argv[])
             play.takeover_from[49] = 0;
             ee += 2;
         }
+        else if (ags_stricmp(arg, "--clear-cache-on-room-change") == 0)
+        {
+            usetup.clear_cache_on_room_change = true;
```
ivan-mogilko, line 319: It's best to instead write a value into the `cfg` object, to let it be merged with the config and applied uniformly.

`INIwriteint(cfg, "misc", "clear_cache_on_room_change", 1);`

ericoporto: Thanks, I will do that then. I have left the PR as draft until I can adjust this.

ericoporto: There's a mix of usetup, cfg, and INIwrite usages in main.cpp, I don't understand what are the differences between each of these ways of doing things. Perhaps a comment on top of the file could clarify when to use what.

Issue #1527, opened 2022-02-03
Experiment with unicode mode in the Editor


ivan-mogilko:
When I was writing #1297 I was not at all sure about how many tasks and what effort overall it would take to move AGS to unicode.
But a success at featuring unicode mode in the engine (#1321) have resolved this problem by 1/2 (or maybe even 2/3).

After #1321 engine can now work in ASCII or UTF-8 text mode, and switch that mode freely at runtime. Currently this is only done when loading a translation which has a "encoding" switch, used as a signal for the engine. But there's no technical necessity to rely on translation file: that was done mostly because translations are the simpliest available data files that may be edited outside of the editor, and did not require much changes to compile from utf-8 text source.

I left this topic afterwards, being distracted by other priorities at the time, but kept thinking that *maybe* we are closer to the unicode support in the Editor than I initially assumed.

---

So I suggest doing an experiment. An important reminder though: as we still have to maintain backward compatibility here, the unicode support must be done through a switch. Presumably this switch may be located right in the General Settings, and set to "ANSI" mode for the older imported projects.

Let's briefly recap the Editor's situation from the point of text encoding. There are roughly 4 components which pass (text) data between each other:

1. Editor input controls. For *text* these are: property grid, and Scintilla text editor used for scripts. Because Editor is a .NET program it supports unicode, so property grid has to be working in unicode. I don't remember and not certain about Scintilla: it could be unicode-aware, or having an encoding setting. In the worst case it's ansi-only, which would suck, but I don't think that may be the case?...
2. "Native" data structures in the C++ library. Now, these work with pure char (byte) arrays and strings, which may contain anything, it's mostly a question of how to interpret that (see what was done for the engine).  The problem is *how the text data is converted when it's passed from .NET side* (and back). IIRC this is done using special functions that convert managed strings into native ones through the *chosen Encoding object*. Supposedly, if we replace one Encoding object with another (depending on a setting value), we may convert C# Strings into UTF-8 byte sequence nicely.
3. Data in compiled game files. Once more, if these are written by .NET side, then text data is converted using a chosen Encoding object. Replacing one should affect the encoding at destination. Those which are written by the native C++ part will keep the strings and char arrays as-is, no questions asked (and these were already encoded correctly in p2).
[...]

PR #1542, opened 2022-02-19
Editor: Unicode support for the game data (scripts, text properties)

ivan-mogilko:
For #1527 (and #1297)
Based on an [experimental branch](https://github.com/ivan-mogilko/ags-refactoring/commits/experiment--unicodeeditor). This is the first step for having unicode in the editor, and may be amended with further changes later.

This adds a support for saving game data in unicode (UTF-8) when working in the AGS Editor.
As 3.6.0 is still a backward compatible version, the support is done by replacing hardcoded ANSI encoding with a variable one throughout the program, in places where the game data is converted from source files into memory and saved back.

1. Added GameTextEncoding property to GeneralSettings. Currently the only choices are ANSI and UTF-8. In theory this could have a full range of explicit Windows-12xx codepages for selection, but for now it is the "undefined ansi", which corresponds to the "current system locale", as editor worked with it previously.
2. The value of the encoding is saved as OPT_GAMETEXTENCODING in the game data. Engine switches to corresponding mode on startup.
3. All the places in the Editor's code which previously used hardcoded "current system locale" now use variable encoding, depending on this game setting.
4. Adjusted Scintilla control (which we use to edit scripts) to work in either mode respecting the game setting.

Current limitations:

1. No automatic conversion of data is performed when user changes the text mode. Therefore this currently is suitable in two cases: for the completely new projects, and if the project contained only standard ASCII text (no extended ANSI characters).
2. No way to select exact ANSI codepage. There's a potential to add this ability, but i am not certain if this is necessary, as we should recommend moving to utf-8 and using ansi only as a backup plan when working with older projects.
3. There's no way to tell the encoding of the rooms and script modules, would these be transferred from another project.
4. Font preview in the Editor is mostly useless for unicode fonts, as it only displays first 256 slots. Well, it was like this for a while now.

Other notes:

* Please remember to use a proper unicode font when testing utf-8 mode. Any "correct" TTF will do, but here's a GNU unifont if you need one for a quick test: https://unifoundry.com/unifont/


ivan-mogilko:
Preliminary tests successful, but more testers are welcome.

Tested myself:
* game texts (game title, character names)
* room texts (object names)
* text in regular script (speech, messages)
* text in dialog scripts

Bugs found:
* typing non-latin letters in script, outside of the double quotes, causes autocomplete to crash.

ivan-mogilko:
So, there's some bad news. Scintilla editor appeared to be a much bigger obstacle for this issue. Although it formally supports UTF-8 mode, but its API currently lacks methods for deducing position in character indexes. Furthermore, there's a wrapper class written around Scintilla itself (called ScintillaWrapper), which does number of actions with it, and it basically uses same variables in both meanings (as a character index, and as internal position in bytes).

I guess all this is solveable with certain effort, but will take more time.

Taking a look at the contemporary Scintilla version, they already [got the necessary functions](https://github.com/jacobslusser/ScintillaNET/blob/f4a1be4df11580071b77dfe16b01893a738d1f03/src/ScintillaNET/LineCollection.cs#L96), but I have no idea how realistic it would be to upgrade the library to the latest stable version. As that may easily break the surrounding Editor's code.

In the worst case, maybe I could use their code as a reference.

ericoporto:
Scintilla is updated by #1416 , although not sure if the version upgrade there is sufficient.

ivan-mogilko:
> Scintilla is updated by #1416 , although not sure if the version upgrade there is sufficient.

No, it does not look like it makes any difference in regards to this problem.

ivan-mogilko:
Made an experimental branch with upgraded Scintilla: https://github.com/ivan-mogilko/ags-refactoring/tree/scintilla-upgrade-test

Had to change a lot of code in ScintillaWrapper, but most of it is straightforward replacing old API with new API syntax.
From what I understood, new ScintillaNET works in UTF-8 mode by default; it does not even have a direct way to change this to ANSI, other than accessing underlying Scintilla library. But it suddenly came to me that we may keep it UTF-8 at all times for simplicity sake; as the script text has to pass through encoding conversion before saving anyway, if the game is saved in ANSI it still may be done by converting from UTF-8.

ericoporto:
If you really want to reduce size, it may be possible to axe some lexers (like Fortran). I think having it all in the repository makes easier to fix potential bugs in this layer. 

If you think maintaining such code separately works, you can create a new repository in AGS organization and then place the code there - AgsScintillaNet or something like it. Then the project can consume it through submodule or nuget - in case of nuget I think an actual person has to submit the package in the end.

ivan-mogilko:
Opened #1544 for Scintilla. IMO it's best to first merge that one, as this pr leads to various mistakes in script editor when in unicode mode.

ivan-mogilko:
Rebased, and updated after Scintilla's changes.

NOTE: changed the way "text format" property works, so if you saved any test project with the previous version, you'd likely have to delete "GameTextFormat" entry by hand from the Game.agf.

ivan-mogilko:
Alright, i think this should do.

Issue #1034, opened 2020-01-17
Getting a script error while running game from editor doesn't show error message.


Khatoblepas:
**Describe the bug**
When testing the game in the editor, if you get a script error, instead of popping up an error message about a script error, it just closes the Output window and changes focus to the Call Stack instead. Using 256 colour mode with the software renderer in windowed mode.

**AGS Version**
3.5.0.23

**Expected behavior**
As with 3.2.1, if the game aborts with an error, a pop up window should appear with the script error bug, and change to the script editor to the highlighted line.

**Desktop (please complete the following information):**

    OS: Windows
    Version: 7

ivan-mogilko:
I do not observe this error. Does this happen in case of any error? What if you try non-8-bit game?

Khatoblepas:
I tried making a new project that was 32bit and it does the same thing. Testing by using AbortGame to force an error pop up - the game window doesn't close either.

ivan-mogilko:
Do you close the game by pressing "Stop" button on toolbar?

Is it correct, that it also does not open the script pane, nor highlights the script line?

Could you check, when this happens, in the File menu - are Open, Save and Exit commands disabled or stay active?

Khatoblepas:
They are disabled. I've found the source of the issue - the little error box appears to the left of the AGS window, When maximised on my main monitor, the error box opens off screen and can't be clicked. When I make the AGS window smaller or move it to the monitor to the right, the error box appears. Maybe the window is spawned in the wrong place relative to the main window?

ivan-mogilko:
Alright, so this is a multiple monitors issue.

ivan-mogilko:
![Screenshot from Khatoblepas](https://user-images.githubusercontent.com/1833754/73611297-99eb7000-45f1-11ea-8a10-307310b7ba23.png)

ivan-mogilko:
Showing error message is handled by this code in Scintilla: https://github.com/adventuregamestudio/ags/blob/dc9f95ad7aa36bf17dbdcad2903f1b81335367ce/Editor/AGS.Editor/Panes/ScintillaWrapper.cs#L408

perhaps, it does not account for multiple monitors, or has some oversights.

ericoporto:
A few lines below there's this:

`Point mainWindowOffset = this.PointToScreen(new Point(0, 0));`

[...]

PR #1544, opened 2022-02-24
Editor: upgrade Scintilla control to a contemporary version

ivan-mogilko:
May be required for #1527; also see https://github.com/adventuregamestudio/ags/pull/1542#issuecomment-1046346848

**Context**

[Scintilla](https://www.scintilla.org/) is a script editor control written in C++.
[ScintillaNET](https://github.com/jacobslusser/ScintillaNET) is a .NET wrapper and over a Scintilla.

We are currently using Scintilla 1.71, and ScintillaNET which exact version is unknown, but judging by the mention in the AGS documentation it's one released somewhere between years 2003 and 2006.

**Purpose**

1. Switching to a more up-to-date software, as the one we're using is at around 15 years old (or more).
2. Proper unicode support. Unfortunately, even though Scintilla itself has a basic support for UTF-8 format, an old ScintillaNET does not account for multibyte characters, and all of its API implementation assumes that 1 character = 1 byte. If we want to have a script editor working in a unicode mode, we'd have to write another layer around ScintillaNET, addressing this issue. The new ScintillaNET is fully unicode compatible, and in fact configures Scintilla to work in UTF-8 by default.

**Changes**

1. Imported Scintilla 3.21.1 (release of 12 September 2020) [from here](https://sourceforge.net/projects/scintilla/files/scintilla/3.21.1/). This is a "LongTerm" branch ([more info here](https://scintilla.sourceforge.io/LongTermDownload.html)) that keeps reliance on C++11 standard. Frankly, I don't think that sticking to C++11 is important for the Editor (as it may be for the Engine), but decided to do this for consistency. NOTE: when importing the code, only one "Cpp" lexer was added out of around 100, to save some space, because that's the only we are currently using.
2. Imported ScintillaNET 3.6.3 [from here](https://github.com/jacobslusser/ScintillaNET). This seem to be the latest stable version. The code is considerably bigger than the old version, but this may be because the old one was simply a barebones wrapper which only called Scintilla's `DirectMessage` function for everything. The new code has a more .NET style API, does number of conversions (for unicode, and others), and has larger comments coverage. EDIT: unexpected, but total pr change is -13k, could be because i did not import various complementary files this time.
3. 99% changes in the Editor are in [ScintillaWrapper.cs](https://github.com/adventuregamestudio/ags/blob/v.3.6.0.17/Editor/AGS.Editor/Panes/ScintillaWrapper.cs), and are due to the API changes.
4. The script editor itself now works in UTF-8 always. I think this should not be an issue, as editor converts script text using project's encoding when saving or passing into compiler. But worth double checking, of course.

**Potential issues**

Looking at the history of Scintilla and ScintillaNET files within the AGS repository, it seems these had several addtions by CJ and other contributors done many years ago. This should be kept in mind if some issues are found, as we may (or may not) have to restore some of these fixes.


ericoporto:
Apparently there is a way that we could unmangle that  `_Scintilla_DirectFunction@16` to just `Scintilla_DirectFunction` in msvc, using an exports.def file: https://www.codeproject.com/Articles/778255/DllExports-Common-Problems-and-Solutions

Could not yet test to verify it works.

ivan-mogilko:
@ericoporto added the fixes you mention.

In regards to dialog scripts, I found an old bug (?). Apparently the styling parser does not recognize tab as a valid indent for the regular script, unlike spaces, so everything after tab is colored as dialog speech.
In terms of compilation, tabs seem work as indents fine, so it's somewhere in styling.

ericoporto:
hey, it appears to be working for me now!

ivan-mogilko:
@ericoporto 
> Edit: I don't understood what you mentioned about tabs, I thought AGS automatically replaced tabs for two spaces - maybe I configured this ages ago and forgot how...

Check out the Preferences window: you may have spaces or tabs for indents.

ericoporto:
> pressing enter after a string containing them will move cursor few lines below

The last commit was a fix to this issue? I just tested after it and could not reproduce.

I just connected my PC to the tv to test having an additional monitor. The autocomplete appears correctly, and also triggering an error (e.g. animating an invalid loop) under debug raises the box in the correct place.

ivan-mogilko:
@ericoporto 
> I just connected my PC to the tv to test having an additional monitor. The autocomplete appears correctly, and also triggering an error (e.g. animating an invalid loop) under debug raises the box in the correct place.

could that mean that #1034 is fixed?

ericoporto:
I cannot reproduce that issue in neither 3.5.1 or current ags.

PR #1545, opened 2022-02-25
Engine: hide sound options that are not currently working

ericoporto:
image

 I am sure at some point we will figure what to do with the sound options in winsetup, but in the meanwhile they do nothing, which can be confusing. 

This PR hides the controls for the time being.

This was also my first time playing with winsetup, I used the visual editor in VS to edit it, and then read the changes through git diff and picked the essential changes and reapplied them manually. It was fairly quick but I am curious if there's some way to tell VS to not fuss in the other things in the resource file.


fernewelten:
What about “Enable threaded audio?” Doesn't that do anything either? I know that some of my games used to have bad audio hiccups without that checked and avoid those hiccups when it's checked. 

For instance, IIRC, Commando Raid, at the time when the Engine needs to switch to the super large maze room and calculate the maze. With threaded audio disabled, it gets hiccups, with it enabled, it avoids the hiccups.

ericoporto:
The audio stuff has been completely refactored and it's always thread now unless it's on a platform that doesn't support it well (for now only Emscripten builds).

ivan-mogilko:
Sorry, but i don't think this is necessary, and with big empty space this looks even more confusing imo. These options should be properly remade somehow. Likely, merged into 1 option. If SDL lets select different devices, then it should have a list, otherwise - a checkbox (sound on/off).

ivan-mogilko:
If you are planning to work on this further, I may mention a hint: unfortunately the standard ("native") window editor is horrible, as it offers very little help to arrange things. So, say, if you remove/replace something and have to also adjust the positions of controls below - then it may be simplier to subtract Y positions directly in the rc source.

ericoporto:
I think I can figure what to do regarding sound but not regarding Midi, it goes through the same backend as sound now. I would probably axe midi device from the config tree too and then we only have sound setting. I think the threaded option can be cut too, at least from the GUI.

ivan-mogilko:
Yes, the midi option was separate likely because allegro had a different path for midi sounds and initialized separately. The need to have 2 switches was lost long ago though.

PR #1548, opened 2022-03-01
Editor: convert game files when the text format changes

ivan-mogilko:
These are the final changes for #1527 (and maybe #1297 in general).

After #1542 is merged you can have a unicode support in the editor, however if you import an older project, and then switch to UTF-8, the text may become corrupted if it uses any characters but common latin.
This pr adds a conversion process run after changing the encoding setting in the editor. The conversion is done simply as:
* load the data using old encoding;
* save the data using new encoding;
repeated with each file.

Files processed are:
* all script files;
* room files;
* main project file (Game.agf).

Translations (TRS files) *are not affected*, as they each have their individual encoding setting.

No big code changes were necessary, as the Editor already converts texts to C# representation in memory using provided Encoding object. So this is mostly calling existing load/save functions in a loop, with some adjustments.

Additional changes:
* save encoding hint to the SCM (exported script module)
* save encoding hint to the room file. For the rooms, added a trivial extension id "ext_sopts", which contains a string key/value map. This extension may be used in the future to store any other optional metadata.

These hints are used by the Editor when importing script modules and rooms.

### DIFF
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -465,11 +465,13 @@ HError ReadRoomBlock(RoomStruct *room, Stream *in, RoomFileBlock block, const St
             String::FromFormat("Type: %d, known range: %d - %d.", block, kRoomFblk_Main, kRoomFblk_ObjectScNames));
     }
 
-    // Add extensions here checking ext_id, which is an up to 16-chars name, for example:
-    // if (ext_id.CompareNoCase("REGION_NEWPROPS") == 0)
-    // {
-    //     // read new region properties
-    // }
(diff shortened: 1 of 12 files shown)


ericoporto:
hey, when I export and import a script module, on import the script and header parts are reversed.

ivan-mogilko:
> when I export and import a script module, on import the script and header parts are reversed.

Fixed in master.

ivan-mogilko:
Ok I think i will add this. Let alpha/beta testers find if there's still any problems. (also it's not too hard to change)

PR #1553, opened 2022-03-05
Engine: fix sqrtf was not declared error in gcc 10

ericoporto:


ivan-mogilko:
Please, place the standard headers together at the top of the list; our include convention atm is:
* the header pair of current cpp
* standard headers
* the engine headers

ericoporto:
placed it there together with algorithm. :)

Issue #950, opened 2019-10-15
Script editor font and font size


ghost:
It is possible to change font size in the script editor using Ctrl + Mouse Wheel, but that could be "unintentional" (provided by Scintilla component). Problem is that it does not save, and does not apply to every script window.

My suggestion is to add at least font size (for script editor) to the editor preferences and maybe update it when user changes one inside script itself.

ivan-mogilko:
Resolved by #1557

PR #1557, opened 2022-03-06
Editor: Script Font and Size selection in Preferences

ericoporto:
Allow selecting which font to use for Script Editor in AGS Editor, and also the font size.

If it's decided we don't want this now, it can be closed. But I thought it was really nice to be able to switch the default font and font size. :)

image

### DIFF
--- a/Editor/AGS.Editor/Entities/EditorPreferences.cs
+++ b/Editor/AGS.Editor/Entities/EditorPreferences.cs
@@ -85,6 +85,14 @@ namespace AGS.Editor.Preferences
         }
     }
 
+    public class InstalledFontTypeConverter : BaseListSelectTypeConverter<string, string>
+    {
+        protected override Dictionary<string, string> GetValueList(ITypeDescriptorContext context)
+        {
+            return Factory.GUIController.InstalledFonts;
+        }
+    }
+
     public class RecentGame : IEquatable<RecentGame>
     {
         // default constructor is needed to serialize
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
I think these new properties should be in "Script Editor" category (currently: "Editor Appearance").

I noticed following bug: the left panel of the script window is not resized when the font changes, until you close and reopen the document. This may cause label numbers not be fully visible.

ericoporto:
Ah, good idea in the Script Editor category!

> I noticed following bug: the left panel of the script window is not resized when the font changes, until you close and reopen the document.

Ah, thanks, I didn't notice before! Will fix!

ivan-mogilko:
I think these new properties should be in "Script Editor" category (currently: "Editor Appearance").

I noticed following bug: the left panel of the script window is not resized when the font changes, until you close and reopen the document. This may cause label numbers not be fully visible.

ivan-mogilko:
Weird, my comment got posted again for some reason... (deleted duplicate)

ericoporto:
Adjusted the mentioned issues!

ericoporto:
Found an issue that could possibly be closed by this , #950

ivan-mogilko:
There's a styling problem with AutoComplete, but it's rather a problem with Scintilla. Unfortunately it does not have a separate style for autocomplete, and also uses some kind of cached "line height" value when displaying autocomplete list. This cached value does not update immediately on setting font and size, but only upon certain changes. This results in list displayed in its own font, but using script editor's line spacing, which increases with the assigned font size.

ivan-mogilko:
On additional note, the tooltip for function parameters is still using fixed font.

ericoporto:
Interesting, I did not notice that. The only way I can think right now without looking more in depth in scintilla code is to remove scintilla autocomplete and use a new one instead. Something like what is here: https://github.com/jacobslusser/ScintillaNET/wiki/Custom-Autocompletion

But before going this route I would look a bit more if it's possible to fix this in some way.

ivan-mogilko:
I wonder if we may use same font for autocomplete for the time being. This way we won't have to rely on hacks.
It looks a little strange, when you set large font in the editor and autocomplete still stays small.

Alternatively, we could try patching Scintilla, add a new Style for autocomplete, in similar fashion to it having its own style for the tooltips. Then add config options for all of these.

ivan-mogilko:
Does this look bad?

autocompletefonts

ericoporto:
The autocomplete looks alright for me but the tooltip I think it benefits from having a different font. Not sure if I am just used to default AGS. I need to check how things are VS and other IDEs - I am drawing a blank trying to remember, but I am not on my computer right now.

ivan-mogilko:
Well, tooltip has its own style, so it may be kept, and perhaps configured in preferences.

PR #1558, opened 2022-03-06
Build tools on CI

morganwillcock:
- Builds tools binaries on all 3 platforms which are configured to build
- Includes 2 source changes to tools source to account for variations in the compiler being used (as far as I can tell these are errors relating to string data types and the accidental creation of a circular namespace reference)
- Removes some push and pop commands from directory switching because they mislead about how the CI is operating

Note: Tool binaries are not tested and not uploaded as part of a release

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -167,17 +167,17 @@ build_android_task:
     greedy: true
   git_submodules_script: git submodule update --init --recursive
   build_prepare_script: |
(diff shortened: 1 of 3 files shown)


ericoporto:
It is indeed building, looks fine @morganwillcock !

morganwillcock:
It is related to that, but is also just to have them building to just check that they do actually build.

Issue #1550, opened 2022-03-03
Unicode support: in script API String class should work with unicode chars


ivan-mogilko:
As #1297 was generally complete, there's still an issue of String class API using 8-bit `char` type to return and append text characters.

The proposal is to adjust related functions and properties to work with `int` based type instead. This may be simply `int` for now, or a new type declared for "text character" as opposed to `char`=byte.

In AGS script `char` and `int` may be assigned to each other without explicit conversion (unlike int-float), therefore the existing scripts will continue to work.

The related members are:

- String.Chars[ ] (indexed attribute)
- String.AppendChar
- String.ReplaceCharAt

On a separate note, there's a question of String.Format's `%c` specifier. In C/C++ this means strictly 1byte-char, so this will work only for character codes <256 now. Need to figure out whether to add another one for printing unicode chars (int), or allow to use same specifier for unicode too. Technically, the script vm allows the latter, as `char`, `short` and `int` all are represented as 32-bit ints when passed as function arguments in ags script vm.

AlanDrake:
We'll also need to add String.Bytes[]/AppendByte/etc variants, so the ability to manipulate raw data is not lost.
Though, won't there be annoyances with backward compat?

AlanDrake:
For example, evil tricks like manually encoding custom data on savegame descriptions.
What about File.ReadRawChar() and File.WriteRawChar()? Would they handle characters or bytes?

ivan-mogilko:
> manually encoding custom data on savegame descriptions.

But you likely don't need "bytes" for that, you can use regular characters.

> What about File.ReadRawChar() and File.WriteRawChar()? Would they handle characters or bytes?

I don't know, probably bytes, because otherwise there will be no way to read byte-by-byte. Perhaps there should be another function for reading or writing full characters. Or maybe reading/writing a string of given length, thus one will be able to save string of length 1 if they need 1 char.

In general, I think i disagree with letting users to access internal String contents as bytes. That breaks the idea of String as a sequence of characters, and also locks String to particular format implementation. AGS script's String is not a C-string, not a raw memory buffer, it's more like a C# String. If you look into C# their String does not give free access to its bytes; it only allows to retrieve bytes as array, or set new string from byte array.

ivan-mogilko:
[...]

PR #1562, opened 2022-03-07
Add basic utf-8 utilities to Common, and fix extended ansi key codes in on_key_press callback

### LINKED ISSUE #1561: AGS doesn't handle dead keys (accents) anymore
**Describe the bug**
AGS 3.5.1 which was Allegro based, could handle [dead keys](https://en.wikipedia.org/wiki/Dead_key), meaning on Windows 10, I could type `'` and then `e` and get the `é` character with my keyboard configured for **English International**. 

After the SDL2 conversion, any dead key usage results in keycode 195 - after configuring the keyboard, type \` and `space` to load the console and try any combination of accent + vogal, and see it reported as keycode 195.

**AGS Version**
current master (https://github.com/adventuregamestudio/ags/commit/230795414b676a2c1e84ae72488af18b90f71692)

**Game**
[ags_dead_key_test.zip](https://github.com/adventuregamestudio/ags/files/8193302/ags_dead_key_test.zip)

**To Reproduce**
Type in the textbox in the above game any key using accents with English International using the 3.5.1 engine and verify it works.
Now use any build of the engine from current master and see it doesn't work anymore.

**Expected behavior**

**Screenshots**
Both screenshots are pressing `'` and `e` (é), `'` and `a` (á), `'` and `u` (ú), `'` and `o` (ó), `'` and `i` (í).

AGS 3.5.1
image

AGS master
image

**Desktop:**
 - OS: Windows
 - Version 10

**Additional context**
Text input in the example above is handled through `on_key_press`, the simplified version of the code is something like:

```AGS Script 
String input_text;
void on_key_press(eKeyCode keycode)
{
  input_text = input_text.AppendChar(keycode);
}
```

So it appears the keycode before was the ascii final key emitted by the OS.

- [3.5.1 `sys_events.cpp`](https://github.com/adventuregamestudio/ags/blob/release-3.5.1/Engine/ac/sys_events.cpp)
- [master `sys_events.cpp`](https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/sys_events.cpp)

Curiously `^` key in my keyboard requires shift+6 to reach, and then an additional space after, since it's a dead key, so it's curious it's appearing above (keycode 195).

ivan-mogilko:
Fixes #1561. (Also may be eventually useful for #1550)

Added some utf-8 utilities (encoding, decoding), based on the code from https://c9x.me/git/irc.git/tree/irc.c with minor adjustments.
Moved previously experimental "convert_utf8_to_ascii" function from translation.cpp to StrUtil module. Don't use allegro4 function uconvert, and use utf-8 utilities instead, to avoid depending on allegro.

Upon SDL_TEXTINPUT convert received text from utf-8 to ascii before assigning to the keycode. This fixes extended ANSI keycodes passed into on_key_press script callback.

NOTE: I had to use custom utf-8 decoding code, because apparently `setlocale` fails at setting "utf-8" on some older Windows systems (prior to Windows 10, judging by documentation, unless some patches are installed). Because of that I could not use standard `mbstowcs` to convert from utf-8 to widechar.

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -18,6 +18,7 @@
 #include "core/platform.h"
 #include "util/math.h"
 #include "util/stream.h"
+#include "util/utf8.h"
 
 using namespace AGS::Common;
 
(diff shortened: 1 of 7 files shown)


ericoporto:
This indeed fixed the issue on my end! While we want to be backwards compatible, if you think this is an edge case we don't want to accommodate more, the proposal in #1563 is a good one too.

I could not detect regressions caused by the changes in this PR.

PR #1564, opened 2022-03-07
Replaced allegro's findfile with our own (refactor)

ivan-mogilko:
Replaced remaining uses of allegro4's findfile functions with our own implementation.

Added some utf-8 utilities (encoding, decoding), based on the code from https://c9x.me/git/irc.git/tree/irc.c with minor adjustments, and two functions for converting utf8<->widestring. NOTE: this duplicates #1562, so either pr should be adjusted depending on which is merged first.

Removed unused allegro functions from the lib sources.


ivan-mogilko:
Tested, fixed, and reorganized the code.

PR #1568, opened 2022-03-08
Script API: support working with unicode characters

### LINKED ISSUE #1550: Unicode support: in script API String class should work with unicode chars

ivan-mogilko:
Resolves #1550.

1. Adds support for unicode character literals. Technically, script compiler already converts character literals to 32-bit integer, so the only thing remaining was to let it parse utf-8 multibyte sequence properly. For compatibility with ascii/ansi scripts this is done only if the new SCOPT_UTF8 compiler option is set: in ascii mode this will keep extended ansi (>127) codes parsed as single byte chars.
EDIT: i realized that the previous compilers did not support "extended ansi" character literals, as i wrongly assumed, because they resulted in a negative char code. Still this option is necessary to prevent attempts to parse these as utf-8.
2. `String.AppendChar` and `ReplaceChar` now accept "character" parameter as `int`.
3. `String.Chars[]` indexed attribute now works correctly in unicode mode, treats index as character index (not byte index) and returns correct character codes.

### Testing

Used following script for testing. Note that this will work correctly only in unicode mode, and if proper font is used.

<pre>
	char c = 'a';
	int uc = '步';

	String s = "unicode test: ";
	s = s.AppendChar(uc);
	s = s.Append(" x <--");
	s = s.ReplaceCharAt(16, uc);
	Display("%s \n %c", s, s.Chars[14]);
	Display("ascii char: %c\nunicode char: %c", 'a', '步');
</pre>

### Backward compatibility

Created a game in AGS 3.5.1 with the following script:

<pre>
	String s = "";
	s = s.AppendChar('a');
	s = s.AppendChar('b');
	s = s.AppendChar('c');
	s = s.AppendChar('d');
	s = s.AppendChar(-24);
	s = s.AppendChar('f');
	s = s.ReplaceCharAt(3, -31);
	s = s.AppendChar(s.Chars[4]);
	Display(s);

	int uc = 27493;
	s = "unicode test: ";
	s = s.AppendChar(uc);
	s = s.Append(" x <--");
	s = s.ReplaceCharAt(16, uc);
	Display("%s \n %c", s, s.Chars[14]);
	Display("ascii char: %c\nunicode char: %c", -24, 27493);
</pre>

Confirmed that the result looks precisely the same when running compiled 3.5.1 game with the new engine from this branch.

### DIFF
--- a/Common/script/cc_options.h
+++ b/Common/script/cc_options.h
@@ -27,6 +27,7 @@
 #define SCOPT_NOIMPORTOVERRIDE 0x20 // do not allow an import to be re-declared
 #define SCOPT_LEFTTORIGHT 0x40   // left-to-right operator precedance
 #define SCOPT_OLDSTRINGS  0x80   // allow old-style strings
+#define SCOPT_UTF8        0x100  // UTF-8 text mode
 
 extern void ccSetOption(int, int);
 extern int ccGetOption(int);
--- a/Compiler/script/cs_parser.cpp
+++ b/Compiler/script/cs_parser.cpp
@@ -13,8 +13,8 @@
 #include "script/script_common.h"
 #include "script/cc_error.h"
 #include "cc_variablesymlist.h"
-
 #include "fmem.h"
+#include "util/utf8.h"
 
 extern int currentline;
 
@@ -218,9 +218,14 @@ int cc_tokenize(const char*inpl, ccInternalList*targ, ccCompiledScript*scrip) {
             thissymbol.push_back(fmem_getc(iii));
             symlen++;
         }
-        if ((thissymbol[0] == '\'') && (thissymbol[2] == '\'')) {
-            // convert the character to its ASCII equivalent
-            thissymbol = std::to_string(thissymbol[1]);
+        if ((thissymbol[0] == '\'') && (thissymbol.back() == '\'')) {
+            int chr = 0;
+            if (ccGetOption(SCOPT_UTF8)) {
+                Utf8::GetChar(&thissymbol[1], thissymbol.size() - 2, &chr);
+            } else {
+                chr = thissymbol[1];
+            }
+            thissymbol = std::to_string(chr);
         }
         else if (thissymbol[0] == '\'') {
             cc_error("incorrectly terminated character constant");
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -446,7 +446,7 @@ internalstring autoptr builtin managed struct String {
   /// Returns a new string with the specified string appended to this string.
   import String  Append(const string appendText);
   /// Returns a new string that has the extra character appended.
-  import String  AppendChar(char extraChar);
+  import String  AppendChar(int extraChar);
   import int     Contains(const string needle);   // $AUTOCOMPLETEIGNORE$
   /// Creates a copy of the string.
   import String  Copy();
@@ -455,7 +455,7 @@ internalstring autoptr builtin managed struct String {
   /// Returns a lower-cased version of this string.
   import String  LowerCase();
   /// Returns a new string, with the specified character changed.
-  import String  ReplaceCharAt(int index, char newChar);
+  import String  ReplaceCharAt(int index, int newChar);
   /// Returns a portion of the string.
   import String  Substring(int index, int length);
   /// Truncates the string down to the specified length by removing characters from the end.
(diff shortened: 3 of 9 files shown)


ericoporto:
Tested the strings and also ReadRawChar with AppendChar, and ReadRawLine along too. Tested some old things of mine and they worked. And also tried some experiments with UTF-8 and it worked provided the fonts were up to the task too!

ivan-mogilko:
File.ReadRawChar is reading bytes. It will be impossible to read and add utf-8 chars by combining File.ReadRawChar and String.AppendChar. Unfortunately, there will be a discrepancy in names there. We may of course rename ReadRawChar to ReadRawByte, idk.

For reading utf-8 from the file there will be two methods:
* either read full string (ReadRawLine, or ReadStringBack)
* use ReadRawChar to read bytes into a `char` array, and then use that char array to initialize String object.

ericoporto:
Erh, `File.ReadRawChar` is working for me, try it on:

```AGS Script
  File * F = File.Open("my.txt",eFileRead);
  String st = ""; 
  while(!F.EOF) {
    st = st.AppendChar(F.ReadRawChar());
  } 
  
  Display(st);
  F.Close();
```

Load the `my.txt` in notepad++ and write any utf8 stuff you want there. You need to manually set the encoding there. If it should not work than the `ReadRawChar` api is wrong, since it's return type is an int.

ivan-mogilko:
I see. I think the reason why above works is because String does not make any checks on whether the added char code is a valid utf-8.

Regarding ReadRawChar, looking into it now made me realize that it works incorrectly for many years. Its original variant which returned `char` was already incorrect, as it returned -1 for end of file, but used [fgetc](https://en.cppreference.com/w/c/io/fgetc) for reading byte, meaning that there was no way to distinguish -1 return value between "end of file" and real byte value.
Looks like i tried to fix it at some point, but still done it wrong either by oversight, or trying to keep the original incorrect behavior (i'm puzzled of what was done there tbh).

rofl0r:
i'd argue that some games may have used ReadRawChar to read binary data that's actually not valid utf-8 when checked, therefore changing this behaviour might cause issues.
admittedly with the bug pointed out by ivan this would have caused issues if the binary data contains 0xff, but maybe game authors found a way to work around that.

fernewelten:
> Regarding the new "unicode" mode, I hate to reiterate myself, but imo it's important to make a clear decision, what is the String object in ags script:

Question is, why are we so fond of that `String` implementation in the first place. I've yet so see a place in code where its immutable-ness translates to efficiency. If there are any efficiency gains to be had from making strings read-only, we don't seem to actually  use them.  To the contrary, as Ivan rightly points out, building a long string incrementally means lots of temporary objects that are almost immediately discarded afterwards. And as far as I know, our memory routines aren't that bad but don't make an _exceptionally_ good job, either (for instance, they seem to rely on reference counting exclusively for garbage collection, which is said to be slower amortized than other concepts). 

This might have been decided at a time where AGS was simplified C and C didn't have any efficient string library yet.

So why aren't our Strings simple wrappers around `std::string``? I'm not at all sure that we can __actually__ improve on optimized compiler library code.

ericoporto:
String is a text format. If we need an additional array of bytes or whatever it can be figured out later. I would not do additional changes in ags3 and leave things as they are - I prefer the C# like immutable Strings also.

If we want to change lets do in ags4.

> So why aren't our Strings simple wrappers around std::string

This is a long discussion, essentially if we go that way we would just use std::string, no need for wrappers. But also, you are mixing the engine code String and the Script String, which are different things.

@ivan-mogilko I am sorry, I wasn't trying to do any discussion here, I think the code is fine. I was just reporting what were the results of experimenting with the resulting Editor from the branch. I won't comment anymore.

ivan-mogilko:
> This is a long discussion, essentially if we go that way we would just use std::string, no need for wrappers. But also, you are mixing the engine code String and the Script String, which are different things.

I understand this as @fernewelten speaks strictly about script String.

ivan-mogilko:
Adjusted the backward compat test script (first post) to include characters with diacritics. Since before 3.6.0 compiler did not support literal chars > 127, i had to use literal numbers instead (-24, -31).

Confirmed that this pr produces same result when running 3.5.1 game.

PR #1569, opened 2022-03-08
CI: speedup MacOS builds

ericoporto:
- use ninja
- cache ninja and cmake

This at least halved the time it takes for the current MacOS builds in my tests.


morganwillcock:
Do you know how much of this was just that ninja runs with parallel jobs by default?
We could just tell make to run in parallel using all cores if speed is a problem.

PR #1571, opened 2022-03-10
Provide "on_text_input" callback for handling unicode characters

### LINKED ISSUE #1563: Unicode: provide a script callback for handling unicode characters
#1561 made me remember another problem. In ags script we have callback `on_key_pressed(eKeyCode)`. This callback's purpose is to handle not the actual printed character, but the code of the pressed key. For instance, it does not distinct small and capital letters (users would have to use IsKeyPressed to test for shift pressed if they want to do so).

With the unicode support, this callback's limitation becomes even more apparent. Currently there's no way to handle any unicode character properly. This may become a problem if, for example, the script implements a custom text input.

I think we need another callback for that; and have something similar to SDL2 event division: key event and text event.
As a draft example, we could have `on_text_input(int ch);`. where `ch` is a unicode character (codepoint).

If we had that, then on_key_press could be called for each actual key press event, again following SDL2 logic (since 3.6.0, or in unicode mode).

ivan-mogilko:
Resolves #1563.

Existing `on_key_press` callback is not enough for handling keyboard input in unicode mode, and has other limitations even in ascii mode. The keycodes it receives do not represent the actual character typed, but rather key pressed, where textual representation depends on various factors. While this callback is enough for handling key presses, it may complicate coding of custom input fields in the user script.

This pr proposes a new `on_text_input (int chr)` callback, which receives a text character code in the current game mode (ascii or unicode).
Additionally `dialog_options_text_input(int chr)` callback is meant for doing the same in the "custom dialog options render" state (it may be seen inconvenient that we have such separation between 2 game states, but nothing may be improved here without a larger api change).

---

There's still a pending question of whether to change how `on_key_press` is run in case of multi-key input sequences. For example, when using "International English" layout some letters are typed by pressing `~` or `'` and then a letter key, in a sequence. Similar situation may take place in case of e.g. Chinese layouts where user has to press multiple keys in order to create 1 character (at least from what I heard).

While SDL handles each of these key presses, AGS engine does not let the event through until a final letter is formed. This may be changed for unicode mode. Then `on_key_press` will be called for any real key down, while `on_text_input` will be used for text input.

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -79,7 +79,8 @@
 #define OPT_WALKSPEEDABSOLUTE 47 // if movement speeds are independent of walkable mask resolution
 #define OPT_CLIPGUICONTROLS 48 // clip drawn gui control contents to the control's rectangle
 #define OPT_GAMETEXTENCODING 49 // how the text in the game data should be interpreted
-#define OPT_HIGHESTOPTION   OPT_GAMETEXTENCODING
+#define OPT_KEYHANDLEAPI    50 // key handling mode (old/new)
+#define OPT_HIGHESTOPTION   OPT_KEYHANDLEAPI
 #define OPT_NOMODMUSIC      98
 #define OPT_LIPSYNCTEXT     99
 #define PORTRAIT_LEFT       0
--- a/Common/ac/keycode.h
+++ b/Common/ac/keycode.h
@@ -254,7 +254,9 @@ enum eAGSKeyMod
     eAGSModLCtrl  = 0x0004,
     eAGSModRCtrl  = 0x0008,
     eAGSModLAlt   = 0x0010,
-    eAGSModRAlt   = 0x0020
+    eAGSModRAlt   = 0x0020,
+    eAGSModNum    = 0x0040,
+    eAGSModCaps   = 0x0080,
 };
 
 
--- a/Common/ac/keycode.h
+++ b/Common/ac/keycode.h
@@ -252,8 +252,9 @@ struct KeyInput
 {
     const static size_t UTF8_ARR_SIZE = 5;
 
-    eAGSKeyCode Key = eAGSKeyCodeNone;
-    char        Text[UTF8_ARR_SIZE] = { 0 };
+    eAGSKeyCode Key = eAGSKeyCodeNone; // actual key code
+    int         UChar = 0; // full character value (supports unicode)
+    char        Text[UTF8_ARR_SIZE]{}; // character in a string format
 
     KeyInput() = default;
 };
--- a/Common/ac/keycode.h
+++ b/Common/ac/keycode.h
@@ -246,6 +246,17 @@ enum eAGSKeyCode
     */
 };
 
+// AGS key modifiers
+enum eAGSKeyMod
+{
+    eAGSModLShift = 0x0001,
+    eAGSModRShift = 0x0002,
+    eAGSModLCtrl  = 0x0004,
+    eAGSModRCtrl  = 0x0008,
+    eAGSModLAlt   = 0x0010,
+    eAGSModRAlt   = 0x0020
+};
+
 
 // Combined key code and a textual representation in UTF-8
 struct KeyInput
@@ -253,6 +264,7 @@ struct KeyInput
     const static size_t UTF8_ARR_SIZE = 5;
 
     eAGSKeyCode Key = eAGSKeyCodeNone; // actual key code
+    int         Mod = 0; // key modifiers
     int         UChar = 0; // full character value (supports unicode)
     char        Text[UTF8_ARR_SIZE]{}; // character in a string format
 
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -98,6 +98,7 @@ namespace AGS.Editor
          *                - Cursor.AnimationDelay
          *                - Room.BackgroundAnimationEnabled
          *                - RuntimeSetup.FullscreenDesktop
+         * 3.6.0.20       - Settings.GameTextEncoding, Settings.UseOldKeyboardHandling;
         */
         public const int    LATEST_XML_VERSION_INDEX = 3060020;
         /*
@@ -925,6 +926,10 @@ namespace AGS.Editor
             if (!_game.Settings.UseOldCustomDialogOptionsAPI)
             {
                 preprocessor.DefineMacro("NEW_DIALOGOPTS_API", "1");
+            }
+            if (!_game.Settings.UseOldKeyboardHandling)
+            {
+                preprocessor.DefineMacro("NEW_KEYINPUT_API", "1");
             }
             // Define Script API level macros
             foreach (ScriptAPIVersion v in Enum.GetValues(typeof(ScriptAPIVersion)))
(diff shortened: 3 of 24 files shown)


messengerbag:
In developing the [TextField module](https://github.com/messengerbag/TextField/blob/master/TextField.asc), I've noticed a couple of shortcomings of `on_key_press()` that may perhaps be addressed when doing this API change.

First, because it does not capture the modifier key state (Shift, Caps Lock, Control, Alt, etc), it's necessary to check these separately using `IsKeyPressed()`. However, the problem is that there is a delay between the actual keyboard keypress and when the AGS script gets to the point of checking IsKeyPressed(), which means that the modifier key state may not reflect what it was when the button was actually pressed. In practice, the most noticeable effect is that it won't register a Shift press when typing fast, so that a letter that should be capitalized is instead registered as lowercase.

If the modifier key state _at the time the button was pressed_ was passed along to the function (e.g. as a bit field enum), that would ensure it was always in sync.

Second, some keypresses are apparently occasionally lost. I suspect this is due to a limited size of the `on_key_pressed()` buffer, so that if more buttons are pressed within one game loop, some of them are not registered or are dropped. It would be nice if this buffer could be increased to ensure it can accommodate many rapid keypresses.

ivan-mogilko:
@messengerbag ,
In regards to passing key mods to `on_key_press`, that would mean adding a second parameter. Currently the script vm errors on wrong number of parameters in functions. If a new parameter is added, then one of two things will happen:
* everyone will be forced to learn to write this extra parameter in on_key_press;
* script vm has to be somehow support "optional" parameters in callbacks that are skipped if user did not declare them in the function definition.

In regards to missing key presses, I can no longer tell about older versions which used allegro4, but this is unlikely a case with SDL2. There *is* a queued event limit in the engine itself, but going above that limit will cause game error rather than skipping one.

ivan-mogilko:
> Ctrl+C, Ctrl+X, Ctrl+V, Del, Arrow; keys that affect cursor position, selection, and other things but are not text, these can be retrieved from on_key_press

I think the @messengerbag's point is that some modifiers retrieved from isKeyPressed may be inconsistent in on_key_press. That  is a general problem, not only related to text input. The reason for this is that if on_key_press may be scheduled, and called at the time when the modifiers have already changed: basically, on_key_press's parameter is from one situation, but IsKeyPressed return value is from another.

Single Ctrl and Alt modifiers may stay correct, because they are made part of the keycode in ags script. Ctrl+Alt combinations are already impossible to get from the key code alone. Same for Shift - it is never part of the keycode.

ivan-mogilko:
Ah. So, there's another mistake in the engine right now, or rather "unfinished business". While in 3.6.0 engine honestly saves all the key events coming from SDL2, but it still *processes only one key event at a game frame*.

So, probably with allegro that would mean that some events will be missed. In 3.6.0 this means that some events will be *delayed*.

This may be already noticed with the test game that @ericoporto posted in another ticket #1561. If you press and hold a key for a while, the input field will continue adding keys even after you release it. It's a separate bug that has to be fixed.

ivan-mogilko:
> Ah. So, there's another mistake in the engine right now, or rather "unfinished business". While in 3.6.0 engine honestly saves all the key events coming from SDL2, but it still _processes only one key event at a game frame_.

Pushed fix for this: e489e187e88071c15fa64351852b8ede45f9f7ac

messengerbag:
> If a new parameter is added, then one of two things will happen:
> 
> - everyone will be forced to learn to write this extra parameter in on_key_press;

Well, I think ultimately this would be best, but it might be better to keep to the current API for now, for backwards compatibility. So I propose two alternative options:

- Make a new version of `on_key_press()` that does take the other parameter. (The functions would otherwise be equivalent, and would work off the same event queue.) Perhaps name it `OnKeyPress()` as a move towards more consistent function naming. At some future date, the old function could then be deprecated and removed.
- Or add a method to query the modifier key state at the time the currently-processing `on_key_press()` (that is, the keypress most recently retrieved from the buffer) was registered. Effectively, this would be a method to retrieve the missing function parameter.

ivan-mogilko:
So, I made an experiment, where engine supports script callbacks with incomplete argument list. For example, if engine wants to pass 2 arguments to a callback, but the function in script is defined having only 1, then only 1 will be passed and remaining discarded.

This allows to have `on_key_press` with two arguments in the new version, while still being able to run the old version with 1 arg.

I added modifiers flag set for a test, which seems to work in general, but it does not work for shift + key combinations at the moment, because of how the handling of sdl events is coded. I need to address that.

> Perhaps name it OnKeyPress() as a move towards more consistent function naming

All the standard callbacks are currently named in lowcase, if i use camel case in callback's name that will be rather inconsistent.

ericoporto:
I tested linking to an on_key_press with a single parameter and it worked. I also used the `on_text_input` and it's working great, tried several key combinations and got the final character fine. Holding a key in the keyboard also behaved correctly. I think `on_text_input` is a great addition, I am sure is going to update ImGi to use it. :)

ivan-mogilko:
@messengerbag @ericoporto 

So... i decided to skip half-measures, and go all the way (still as an experiment).
Idea is to have a Key Handling api switch in the General Settings, similar to existing switches: old mode and new mode. It's called "Use old-style keyboard handling" and found in "Backwards Compatibility" group.

Old mode is a backward compatible input handling mode, where
 - lone mod keys (Ctrl, Alt, Shift) are not passed further into the engine nor script;
 - key + mod combos are merged into one key code for the `on_key_press` callback: e.g. eKeyCodeCtrlA key codes;
 - `on_text_input` script callback is **NOT** called.

New mode is:
 - **all keys** are passed into the engine and `on_key_press` script callback as-is;
 - `on_text_input` script callback receives printable characters;

Added new compiler macro `NEW_KEYINPUT_API` to allow distinct handling mode. E.g. if you are writing a script module, you may process all keys in `on_key_press` when `NEW_KEYINPUT_API` is not defined, and process text input in `on_text_input` when `NEW_KEYINPUT_API` is defined.

ivan-mogilko:
> https://github.com/ivan-mogilko/ags-refactoring/blob/3aca2aa6b9efaec622f28f52f6024487c378b605/Engine/ac/event.h#L60-L61
> In the line above, what is the `-1000` value there for?

That's just local for "unused argument", i may change to something distinct, like INT_MIN later

ericoporto:
I have tested and it appears to work, I modified some games and they seem to work - I also get a feeling that keys respond faster but I haven't measured, it may be just a feeling.

Let's say I was rewriting the templates for loading the debug room teleport ui, is the below correct?

```AGS Script
function on_key_press(eKeyCode keycode, int mod) 
{
  if (keycode == eKeyX && (mod & eKeyModCtrl) )
  {
    // Ctrl-X will let the player teleport to any room
    Debug(3, 0);
  }
}
```

> That's just local for "unused argument", i may change to something distinct, like INT_MIN later

Uhm, I think it's better a macro or const with a name like UNUSED_ARG, and then it can have any value (-1000 or INT_MIN...).

ivan-mogilko:
>  I also get a feeling that keys respond faster but I haven't measured, it may be just a feeling.

This may be due to the previous fix in master (e489e187e88071c15fa64351852b8ede45f9f7ac).

> Let's say I was rewriting the templates for loading the debug room teleport ui, is the below correct?

Yes, that looks about right.

ericoporto:
I liked the proposed approach a lot, by using the option people can opt in or not to this new behavior. This will probably be a thing that will end up in the "Upgrading to AGS 3.6" topic once we get to it.

messengerbag:
Brilliant! Thanks so much for this, @ivan-mogilko!

Just one suggestion/request: add the CapsLock state to the modifier info, so that the correct capitalization of a character can be known with certainty. (I'll have to think carefully about whether on_text_input() is better suited for the TextField module. In some ways it probably is, but it also has to handle things like the arrow keys and Ctrl+C/Ctrl+V, so it will need to listen to both in any case.)

> All the standard callbacks are currently named in lowcase, if i use camel case in callback's name that will be rather inconsistent.

True, but I assumed that at some future date they would all be turned to CamelCase for consistency with the rest of the script API. (Probably at the same time that Character.x/y/z are finally made uppercase.)

ivan-mogilko:
Right, added eKeyModNum and eKeyModCaps

ericoporto:
Hey, after this is merged, it would be useful to generate a new version so we can update the templates with this change.

Issue #1566, opened 2022-03-07
Remove or properly support Non-threaded audio runtime option


ericoporto:
Currently AGS can only run with threaded audio. Alternatively, it can be built for non-threaded audio too - single thread build is a requirement for Emscripten port.

We can either

- remove the multithread audio flag from command line, OPTIONS.md and WinSetup 
- or reimplement some way to turn threaded audio off at runtime too

I think it's easier to just remove the config options but I also thought it was better to open an issue.

--- 

**context for emscripten**

- I already talked a bit on this in https://github.com/adventuregamestudio/ags/pull/1346 and also on the [`Emscripten/README.md`](https://github.com/adventuregamestudio/ags/tree/master/Emscripten#changes-from-other-ports)

Current pthread implementation in Emscripten comes with multiple requirements both on the browser but also on the server used to host the game:

> Browsers [that have implemented and enabled](https://webassembly.org/roadmap/) SharedArrayBuffer, are gating it behind Cross Origin Opener Policy (COOP) and Cross Origin Embedder Policy (COEP) headers. Pthreads code will not work in deployed environment unless these headers are correctly set.

see: https://emscripten.org/docs/porting/pthreads.html

ivan-mogilko:
Resolved by #1573

PR #1573, opened 2022-03-13
Update sound config to 3.6.0

### LINKED ISSUE #1565: Adjust Sound Config for SDL2 AGS
The Sound and MIDI configuration in AGS in Config Tree, and the interfaces (command line, acsetup.cfg, and WinSetup) currently expects the behavior from the previous allegro4 based AGS. 

We need to either 
- remove these configurations and rely in the automatic behavior from SDL2 or 
- update the configuration so it can modify what SDL2 uses.

**Additional info**

Retrieving the list of available audio drivers can be achieved by using `SDL_GetNumAudioDrivers()` and `SDL_GetAudioDriver(int i)`, at runtime.

Setting the audio driver can be done through  `SDL_setenv("SDL_AUDIODRIVER", "directsound", 1);` to specify a specific audio driver. Since SDL2 2.0.16 it's possible to also pass a list of comma separated values (ordered) so that if the device fails the list is used to provide fallback options. 

Alternatively, using `SDL_AudioInit(driver_name)` can directly specify the driver name.

---

Currently it's possible to influence AGS through OS environment variable `SDL_AUDIODRIVER`, similar to `SDL_setenv`, following https://wiki.libsdl.org/FAQUsingSDL#how_do_i_choose_a_specific_audio_driver

---

Interfaces must also be adjusted:

- [ ] Adjust acsetup.cfg documentation in OPTIONS.md
- [ ] Adjust Command Line parameters
- [ ] Adjust WinSetup

ivan-mogilko:
Resolves #1565, #1566

Implemented config options:
```
[sound]
enabled = [1,0]
driver = id
```

"driver" value is system specific; empty value means use defaults.

Removed config option for threaded audio.

Updated winsetup, and Editor's Default Setup pane.


ericoporto:
This works, and the winsetup has the options removed. I noticed the tell command still mentions `midiid` and `digiid`:

```
.\audio_test.exe --console-attach --tell-config
...
[config@sound]
digiid = -1
driver =
enabled = 1
midiid = -1
threaded = 0
usespeech = 1
```

ivan-mogilko:
--tell-config prints everything found in config files.

ericoporto:
ah, that's correct! Very curious, I rebuilt the game but it still had the removed options in the acsetup.cfg file, once I manually removed them with notepad from the file they were going forever and did not appear in any tell flag.

ivan-mogilko:
Changed to delete the config file before writing.

ivan-mogilko:
Ah! i forgot to update the readme, will to that separately.

PR #1574, opened 2022-03-13
Engine: support option for software renderer's output driver

### LINKED ISSUE #1567: Enhancement: SDL Renderer output hinting Config
The current SDL2 based AGS engine has a new Software driver that uses the SDL Renderer to provide the final on-screen rendering. The SDL Renderer automatically figures out something like OpenGL, Direct3D, or other, and use it when drawing on the screen.

Instead of relying only on SDL2 current automatically behavior, this issue is for an additional manual config so one can specify what the SDL Renderer should use.

There's no equivalent of such config from previous AGS version, so while this could be an interesting config, it's not a priority right now.

ivan-mogilko:
Resolves #1567.

Implemented config option for SDL2's driver selection, this is currently used only for software renderer.

```
[graphics]
software_driver = name
```
where driver's name could be one of: direct3d, opengl, opengles, opengles2, metal, software.
Not all of the drivers are supported on all systems.
Empty name, or missing option means use defaults.

Note that AGS prints SDL's driver name in the log, when using software renderer. This looks like:
> Created SDL Renderer: opengl

### DIFF
--- a/Engine/ac/gamesetup.h
+++ b/Engine/ac/gamesetup.h
@@ -90,6 +90,7 @@ struct GameSetup {
     ScreenRotation rotation;
 
     DisplayModeSetup Screen;
+    String software_render_driver;
 
     GameSetup();
 };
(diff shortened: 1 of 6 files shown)


ericoporto:
It works. I think this implementation suffices for the issue at hand.

PR #1575, opened 2022-03-13
Feature scintilla/script editor: toggle line comments

ericoporto:
- allow toggling line comments in the script editor with Ctrl+Shift+Q

toggle_line_comment

If it's not something for now I can close this PR for now and later reopen to ags4 at an appropriate time.

When experimenting with this I noticed Ctrl+Shift+letter keys were inserting non-printable characters in the text buffer of the script editor.

### DIFF
--- a/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
+++ b/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
@@ -424,6 +424,86 @@ namespace AGS.Editor
             }
         }
 
+        public void ToggleLineComment()
+        {
+            string commentLineSymbol = "//";
+            string aSpace = " ";
+            string comment = commentLineSymbol + aSpace;            
+            bool isUncomment = false;
+            bool setSelectionAtEnd = false;
+            int selStart, selEnd;
+            
+            if (string.IsNullOrEmpty(scintillaControl1.SelectedText))
+            {
+                int tmpline = scintillaControl1.LineFromPosition(scintillaControl1.CurrentPosition);
+                selStart = scintillaControl1.Lines[tmpline].Position;
+                selEnd = scintillaControl1.Lines[tmpline].EndPosition;
+            }
+            else
+            {
+                int tmplineStart = scintillaControl1.LineFromPosition(scintillaControl1.SelectionStart);
+                int tmplineEnd = scintillaControl1.LineFromPosition(scintillaControl1.SelectionEnd);
+
+                selStart = scintillaControl1.Lines[tmplineStart].Position;
+                selEnd = scintillaControl1.Lines[tmplineEnd].EndPosition;
+                setSelectionAtEnd = true;
+            }
+
+            int lineSelStart = scintillaControl1.LineFromPosition(selStart);
+            int lineSelEnd = scintillaControl1.LineFromPosition(selEnd);
+            int indentPos = scintillaControl1.GetLineIndentationPosition(lineSelStart);
+
+            string firstLine = scintillaControl1.Lines[lineSelStart].Text;
+            isUncomment = firstLine.Trim().StartsWith(commentLineSymbol);
+
+            scintillaControl1.BeginUndoAction();
+
+            for(int i = lineSelStart; i < lineSelEnd; i++)
+            {
+                Line line = scintillaControl1.Lines[i];
+                int lineStartPos = line.Position;
+                int lineEndPos = line.EndPosition;
+                int lineIndentPos = scintillaControl1.GetLineIndentationPosition(i);
+
+                string lineText = line.Text;
+
+                if (isUncomment)
+                {
+                    bool canUncommentLine = lineText.Trim().StartsWith(commentLineSymbol);
+                    if (!canUncommentLine) continue;
+
+                    lineText = scintillaControl1.GetTextRange(lineIndentPos, lineEndPos - lineIndentPos);
+                    if (lineText.Length <= (commentLineSymbol.Length + 1)) continue;
+
+                    int len = lineText[comment.Length - 1] == aSpace[0] ? comment.Length : comment.Length - 1;
+
+                    scintillaControl1.SetSel(lineIndentPos, lineIndentPos + len);
+                    scintillaControl1.ReplaceSelection("");
+                }
+                else
+                {
+                    if (string.IsNullOrEmpty(lineText.Trim())) continue; // skip empty lines
+
+                    this.scintillaControl1.InsertText(lineIndentPos, comment);
+                }
+            }
+
+            if (setSelectionAtEnd)
+            {
+                selStart = scintillaControl1.Lines[lineSelStart].Position;
+                selEnd = scintillaControl1.Lines[lineSelEnd-1].EndPosition;
+                scintillaControl1.SetSel(selStart, selEnd - 1);
+            }
+            else
+            {
+                scintillaControl1.GotoPosition(selEnd);
+                scintillaControl1.CurrentPosition = selStart;
+                scintillaControl1.SetEmptySelection(scintillaControl1.CurrentPosition);
+            }
+
+            scintillaControl1.EndUndoAction();
+        }
+
         public void GoToPosition(int newPos)
         {
             int lineNum = scintillaControl1.LineFromPosition(newPos);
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
I think that the rule to add a space after comment symbol may lead to unreliable uncommenting of manually commented sections.

E.g.:
```
//  if (System.SupportsGammaControl)
//  {
//    System.Gamma = 100;
//    sldGamma.Value = 100;
//  }
```
uncommenting leads to
```
 if (System.SupportsGammaControl)
 {
   System.Gamma = 100;
   sldGamma.Value = 100;
 }
```
with only 1-space indentation.

ericoporto:
I could make that if there's more than 1 space on the first line it understand it as it should preserve the spacing for all of the following.

Overall I like the space after the line comment symbol because it makes it easier on the eyes to read the line.

ivan-mogilko:
> I could make that if there's more than 1 space on the first line it understand it as it should preserve the spacing for all of the following.

That won't fix the problem, as the editor won't know whether the space was intended by user, or added by commenting command.

morganwillcock:
I think it is fine to just consider the current lines, unless there is actual structured parsing of the code that could be used to reconstruct the "correct" indentation.

So given this:
```
// if (System.SupportGammaControl)
/////{
///  System.Gamma = 100;
//  sldGamma.Value = 100;
////}
```

...I think it is fine to uncomment to this:
```
if (System.SupportGammaControl)
{
  System.Gamma = 100;
  sldGamma.Value = 100;
}
```

...and then recomment to this:
```
// if (System.SupportGammaControl)
// {
//  System.Gamma = 100;
//  sldGamma.Value = 100;
// }
```

Given something that is indented:
```
    if (System.SupportGammaControl)
    {
      System.Gamma = 100;
      sldGamma.Value = 100;
    }
```

...just preserve the indentation before the comment:
```
    // if (System.SupportGammaControl)
    // {
    //   System.Gamma = 100;
    //   sldGamma.Value = 100;
    // }
```

...and then that should uncomment correctly:
```
    if (System.SupportGammaControl)
    {
      System.Gamma = 100;
      sldGamma.Value = 100;
    }
```

morganwillcock:
> I could make that if there's more than 1 space on the first line it understand it as it should preserve the spacing for all of the following.
> 
> Overall I like the space after the line comment symbol because it makes it easier on the eyes to read the line.

The space is pretty critical when double commenting because you will need to delimit the original comment from the new one.

So this:
```
// FIXME Do more tests
if (System.SupportGammaControl)
{
  // FIXME Lower Gamma value
  System.Gamma = 100;
  sldGamma.Value = 100;
}
```

...comments to this:
```
// // FIXME Do more tests
// if (System.SupportGammaControl)
// {
//   // FIXME Lower Gamma value
//   System.Gamma = 100;
//   sldGamma.Value = 100;
// }
```
...and uncomments to this:
```
// FIXME Do more tests
if (System.SupportGammaControl)
{
  // FIXME Lower Gamma value
  System.Gamma = 100;
  sldGamma.Value = 100;
}
```

> > I could make that if there's more than 1 space on the first line it understand it as it should preserve the spacing for all of the following.
> 
> That won't fix the problem fully, as the editor won't know whether the space was intended by user, or added by commenting command. This inevitably leads to a potential case where this command will remove space which user did not want to remove.

I don't think it can be perfect and cover every case. For anything more involved involved the Editor needs to be more customisable by the user not by the Editor developer.

ivan-mogilko:
@morganwillcock if i understood correctly, your suggestion is to keep the auto-inserted comments on one column?

ericoporto:
Two things I need to comment (other things I am still thinking about).

- the second commit, do we all agree it's needed? After the Scintilla update we can now insert non-printable characters there with Ctrl+shift+[A-Z], if the person use a font that doesn't have anything to represent those things can get a bit hard to debug. I think this was not possible in the old Editor - I don't remember inserting any of those.
- the comment toggle code is made using Scintilla features, this means I am pulling the indent from it, I don't think inserting the comments in a single column is a good idea - what if the column is shorter in the following line? Also the edits have to be done using Scintilla commands so it can be in the undo pile so the person can Ctrl+Z if needed. And I need to figure out if I am commenting or uncommenting, currently this is picked from the state of the first line, I think this is an easy enough algorithm for the user to remember.

ivan-mogilko:
> the second commit, do we all agree it's needed? After the Scintilla update we can now insert non-printable characters there with Ctrl+shift+[A-Z], if the person use a font that doesn't have anything to represent those things can get a bit hard to debug. I think this was not possible in the old Editor - I don't remember inserting any of those.

Yes, this seem to be a necessary bug fix, regardless of the feature.

> I don't think inserting the comments in a single column is a good idea - what if the column is shorter in the following line?

Not sure what do you mean here by "shorter column"?

morganwillcock:
> @morganwillcock if i understood correctly, your suggestion is to keep the auto-inserted comments on one column?

Relative to the indentation of the entire block, yes, otherwise it potentially becomes difficult to read the original code and read the intention of the comment.

i.e. The original indentation is easier to read when commented like this:

```
// // FIXME Do more tests
// if (System.SupportGammaControl)
// {
//   // FIXME Lower Gamma value
//   System.Gamma = 100;
//   sldGamma.Value = 100;
// }
```
...than if it was commented like this:
```
// // FIXME Do more tests
// if (System.SupportGammaControl)
// {
     // // FIXME Lower Gamma value
     // System.Gamma = 100;
     // sldGamma.Value = 100;
// }
```
...because the visual alignment is lost. It also helps to communicate the reasoning for the comment. i.e. it marks the code as being disabled all at once rather than being a combination of commenting the entire block of code and commenting of individual lines.

> the comment toggle code is made using Scintilla features, this means I am pulling the indent from it

For operating on an indented region you can effectively move what is considered the beginning of the line based on the indentation level. So commenting the assignments in this:

```
// FIXME Do more tests
if (System.SupportGammaControl)
{
  // FIXME Lower Gamma value
  System.Gamma = 100;
  sldGamma.Value = 100;
}
```

...would become this:
```
// FIXME Do more tests
if (System.SupportGammaControl)
{
  // FIXME Lower Gamma value
  // System.Gamma = 100;
  // sldGamma.Value = 100;
}
```

...and from there indenting the entire block would give you:
```
// // FIXME Do more tests
// if (System.SupportGammaControl)
// {
//     // FIXME Lower Gamma value
//     // System.Gamma = 100;
//     // sldGamma.Value = 100;
// }
```

It is then possible to reverse in either order, as well as reversing the original steps you can of operate on the assignments again to remove one level of commenting:
```
// // FIXME Do more tests
// if (System.SupportGammaControl)
// {
//     // FIXME Lower Gamma value
    // System.Gamma = 100;
    // sldGamma.Value = 100;
// }
```

...and removing one level of commenting from the entire block then gives you the original back:
```
// FIXME Do more tests
if (System.SupportGammaControl)
{
    // FIXME Lower Gamma value
    System.Gamma = 100;
    sldGamma.Value = 100;
}
```

Probably the hardest part is how to decided whether the toggle is adding or removing. For a multiline region you could do that by checking for any existing comments in any line in the selection (would be true if anything that isn't a comment opener or whitespace is present at the beginning of any line in the selected region, where the beginning of the line is defined by the outer indentation level of the selected region).

ericoporto:
I won't do any immensely complex code in a place that is essentially untestable code, better to close this.

ivan-mogilko:
For information, I found there probably were LineComment and LineUncomment functions in older versions of ScintillaNET, judging by this conversation:
https://github.com/jacobslusser/ScintillaNET/issues/286

ericoporto:
My code is pulled from the notepad++ behavior for line comments

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/10ec944800567e0905a898a63e737e0c0c7d6158/PowerEditor/src/Notepad_plus.cpp#L4660

ericoporto:
> For information, I found there probably were LineComment and LineUncomment functions in older versions of ScintillaNET, judging by this conversation:
> https://github.com/jacobslusser/ScintillaNET/issues/286

It was done as this

https://github.com/transistor1/ScintillaNET-2.6/blob/1e2a9e0d476abc18043acd90033db5a9aee7cc25/ScintillaNET/Lexing.cs#L123-L192

PR #1576, opened 2022-03-14
Script API: expanded Button.Animate() to feature all common params + some refactor

ivan-mogilko:
This was mentioned on forums recently, and made me realize I missed Buttons when expanding Character and Object animation with "starting frame".

Added all parameters available for Character and Object's Animate function:
* blocking;
* direction;
* starting frame

So it looks like
```
import void Button.Animate(int view, int loop, int delay, RepeatStyle=eOnce, BlockingStyle=eBlock, Direction=eForwards, int frame=0);
```

NOTE: same as with Character and Object's starting frame, the reverse animation starts with *frame - 1*, hence passing `frame = 0` will begin reverse animation with the last frame.

Refactored the animation handling (frame, loop increment/decrement) algorithm, use common code among Character, Object and Button.

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -11,11 +11,11 @@
 // http://www.opensource.org/licenses/artistic-license-2.0.php
 //
 //=============================================================================
-
 #ifndef __AC_CHARACTERINFO_H
 #define __AC_CHARACTERINFO_H
 
 #include <vector>
+#include "core/types.h"
 #include "ac/common_defines.h" // constants
 
 namespace AGS { namespace Common { class Stream; } }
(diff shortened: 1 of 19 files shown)


ivan-mogilko:
Added animation refactor, sharing the code between room object and button. Wanted to also do same for character, but their code is more entangled, so probably should not put that into this very pr...

ivan-mogilko:
Also refactored animation cycling algorithm, sharing common code between buttons, room objects and characters.

ericoporto:
I tested that the button Animate is really working, as are the new blocking, forward and backwards and initial frame. I also experimented with the character and object to check regressions and did not find anything. I only tested with building from the pr, did not run previously made games on the resulting engine from the pr.

ivan-mogilko:
I think better merge this after first beta release, to avoid new random bugs in it. (Although i also tested this with both old and new games, but sometimes you can miss things)

PR #1578, opened 2022-03-16
Android: fix can't open vox packages in single game launcher

### LINKED ISSUE #1577: Android: can't play any sound in an external vox package in single game launcher
**Describe the bug**
Sounds not in the game main .ags package can't be played. Because speech is always packaged as speech.vox, we can't play any speech. First reported here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59772.msg636644240#msg636644240

**AGS Version**
AGS 3.6.0.20 but also current master https://github.com/adventuregamestudio/ags/tree/b8a811f9e939328987af3adbbad7c72617dfbafd

**To Reproduce**
In the editor create a new game
1. add any sound and make sure that it's marked to be in a .vox package instead of being packaged inside the game
2. build for Android in the Editor
3. verify the sound doesn't play

**Expected behavior**
The sound should play

**Desktop:**
 - OS: Windows
 - Version: 10

**Smartphone:**
 - Device: Pixel 2 A
 - OS: Android 11
 - Emulated

**Additional context**
When listing the libraries in AssetManager `_activeLibs`, I can only see the root dir and the game.ags file - the vox package doesn't appear as a library. I can see the game tries to look for the `"/au000004.mp3"` clip but it can't find it in either the ags game or in the root dir. The vox packages apparently can't be found.

In the single game package the files are not in a regular directory, and instead are accessible through the AAssetManager from Android. It appears some logic is missing for the vox packages there, but I am not sure what it's yet.

ericoporto:
android asset manager can't find files when they are preceded by `/` or `./` inside it's packaged asset.

fix #1577 

I placed the path stuff in the path utilities but if you think it's better to be in the android file then I can move too. I have a feeling that there must be some other proprietary ROM format that would have some similar need but I don't know any at the time.

I also named the function that simplifies the path as `GetPathInForeignAsset` to illustrate that the asset in question is not an AGS asset but a ROM type of information that another system also calls as "asset".

### DIFF
--- a/Common/util/path.cpp
+++ b/Common/util/path.cpp
@@ -173,6 +173,24 @@ String FixupSharedFilename(const String &filename)
     return fixed_name;
 }
 
+#if AGS_PLATFORM_OS_ANDROID
+    String GetPathInForeignAsset(const String &filename)
+    {
+        if(filename.IsEmpty()) return filename;
+
+        if(filename[0] == '/')
+        {
+            return filename.Mid(1);
+        }
+        else if(filename[0] == '.' && filename[1] == '/')
+        {
+            return filename.Mid(2);
+        }
+
+        return filename;
+    }
+#endif
+
 #if AGS_PLATFORM_OS_WINDOWS
 String WidePathToUTF8(const wchar_t *ws)
 {
--- a/Common/util/aasset_stream.cpp
+++ b/Common/util/aasset_stream.cpp
@@ -20,6 +20,7 @@
 #include <android/asset_manager.h>
 #include <android/asset_manager_jni.h>
 #include "util/android_file.h"
+#include "util/path.h"
 #include "util/string.h"
 
 namespace AGS
@@ -176,15 +177,9 @@ namespace AGS
                 throw std::runtime_error("Couldn't get AAssetManager, SDL not initialized yet.");
             _ownHandle = true;
 
-            if(!asset_name.IsNullOrSpace() && asset_name[0] == '/') {
-                String aname = asset_name;
-                aname.ClipLeft(1);
-                _aAsset = AAssetManager_open(aAssetManager, aname.GetCStr(), asset_mode);
-            }
-            else
-            {
-                _aAsset = AAssetManager_open(aAssetManager, asset_name.GetCStr(), asset_mode);
-            }
+            String a_asset_name = Path::GetPathInForeignAsset(asset_name);
+
+            _aAsset = AAssetManager_open(aAssetManager, a_asset_name.GetCStr(), asset_mode);
 
             if (_aAsset == nullptr)
                 throw std::runtime_error("Error opening file.");
--- a/Common/util/android_file.cpp
+++ b/Common/util/android_file.cpp
@@ -17,6 +17,7 @@
 #include <android/asset_manager.h>
 #include <android/asset_manager_jni.h>
 #include <SDL.h>
+#include "util/path.h"
 
 namespace AGS
 {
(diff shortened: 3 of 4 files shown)


Common/util/path.cpp
```diff
@@ -173,6 +173,28 @@ String FixupSharedFilename(const String &filename)
...
+            String adjusted_filename = filename;
+            adjusted_filename.ClipLeft(1);
+            return adjusted_filename;
+        }
+        else if(filename[0] == '.' && filename[1] == '/')
+        {
+            String adjusted_filename = filename;
+            adjusted_filename.ClipLeft(2);
+            return adjusted_filename;
+        }
```
ivan-mogilko, line 192: The shorter code could be
```
if (filename[0] == '/' || (filename.GetLength() > 1 && filename[0] == '.' && filename[1] == '/))
{
    return filename.Section('/', 1, -1);
}
```

Common/util/path.cpp
```diff
@@ -173,6 +173,28 @@ String FixupSharedFilename(const String &filename)
     return fixed_name;
 }
 
+#if AGS_PLATFORM_OS_ANDROID
+    String GetPathInForeignAsset(const String &filename)
+    {
+        if(filename.IsNullOrSpace()) return filename;
```
ivan-mogilko, line 179: This is not significant, but in general it's best to use String.IsEmpty(), because it tests for empty string (uninitialized or zero length), while NullOrSpace parses whole string to see if it only contains whitespace chars.

Common/util/path.cpp
ericoporto: Curiosity, is the length check needed? If the string is 0 terminated, then it has at least two characters if it's not null or empty. 🤔

Common/util/path.cpp
ericoporto: Ah, I mistook what the function did, I did intent for IsEmpty then, I will adjust.

Common/util/path.cpp
ivan-mogilko: > Curiosity, is the length check needed? If the string is 0 terminated, then it has at least two characters if it's not null or empty.

That's a good point, the String does do regular test in operator[], but it has `assert` which will trigger in debug config.

ivan-mogilko: That is, right now the String class has this:
```
    inline char operator[](size_t index) const
    {
        assert(index < _len);
        return _cstr[index];
    }
```
May be changed to
```
    inline char operator[](size_t index) const
    {
        assert(index <= _len);
        return _cstr[index];
    }
```

IIRC String class guarantees _cstr is valid at all times, pointing to "" literal if was not initialized.

ivan-mogilko: Actually, this is not way too important, there are various ways to simplify it, one is to use same conditions but avoid creating local variables, example:
```
        if(filename[0] == '/')
        {
            return filename.Left(1);
        }
        else if(filename[0] == '.' && filename[1] == '/')
        {
            return filename.Left(2);
        }
```

ivan-mogilko: Actually, there are various ways to simplify this, one way to keep same conditions but only avoid creating local variables:
```
        if(filename[0] == '/')
        {
            return filename.Mid(1);
        }
        else if(filename[0] == '.' && filename[1] == '/')
        {
            return filename.Mid(2);
        }
```

Maybe it's the best one.
`Mid` acts like normal "substring".

ericoporto: I went with this last suggestion! Pushed below!

PR #1583, opened 2022-03-19
Adjust Slider's elements to not exceed the controls' rect

### LINKED ISSUE #1581: Sliders are clipped incorrectly
**Describe the bug**
Recently there has been a change that allows clipping GUI controls (#1495), but unfortunately sliders are a bit tricky regarding their sizes, and their handles are being clipped.

**AGS Version**
3.6.0.20, current master https://github.com/adventuregamestudio/ags/tree/ed971d2c5d289db6df31e2a2214771bcb162e5ca

**Game**
[SliderTest.zip](https://github.com/adventuregamestudio/ags/files/8308060/SliderTest.zip)

**To Reproduce**
1. Run the attached game and see the behavior of the sliders in the GUI that is opened
2. Close, now in general settings, visual, disable "GUI controls clip their contents"
3. Run again and noticed now the sliders are correctly presented (see screenshots below)

**Expected behavior**
Sliders should work even when clipping is turned on.

**Screenshots**
SliderTest sliders with clipping turned on
image

SliderTest sliders with clipping turned off
image

**Desktop:**
 - OS: Windows
 - Version: 10

ivan-mogilko:
Fixes #1581 ; I kept the math for element sizes, but added simple rect clamping to control's bounds in the end, as it seems simplier (clipping mistakes are usually 1-2 pixels, and a result of imprecise use of fractions rather than intent), and also simplier to keep the code for backwards compatibility mode, where it won't do such clamping and keep same sizes.

This fixup is only active in 3.6.0 games and later. Old <3.6.0 games did not have an option to clip GUI controls.

The handle sprites will not be affected by this recalculation, it will be up to the user to setup correct control's size and provide a matching graphic.

### DIFF
--- a/Common/gui/guislider.cpp
+++ b/Common/gui/guislider.cpp
@@ -38,6 +38,7 @@ GUISlider::GUISlider()
     _scEventCount = 1;
     _scEventNames[0] = "Change";
     _scEventArgs[0] = "GUIControl *control";
+    _valueField = 0;
 }
 
 bool GUISlider::IsHorizontal() const
@@ -66,6 +67,9 @@ void GUISlider::Draw(Common::Bitmap *ds)
     if (spriteset[HandleImage] == nullptr)
         HandleImage = 0;
 
+    // New (3.6.0+) drawing calculation method
+    const bool new_draw = loaded_game_file_version >= kGameVersion_360_21;
+
     // Depending on slider's orientation, thickness is either Height or Width
     const int thickness = IsHorizontal() ? Height : Width;
     // "thick_f" is the factor for calculating relative element positions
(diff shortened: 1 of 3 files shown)


ericoporto:
When the user provides the graphics, won't it always clip the handle at the minimum/maximum position with this PR?

ivan-mogilko:
> When the user provides the graphics, won't it always clip the handle at the minimum/maximum position with this PR?

it will clip the handle, problem is that there's no guarantee that the image will be fully visible.

I've got a feeling that we might have to add more properties to the slider eventually, for example, a starting and ending position of a slider's bar, to allow larger handles not getting clipped.

ericoporto:
I was thinking just consider it's width when calculating the offset if it's horizontal or height when calculating the offset if it's vertical.

ivan-mogilko:
> If it was possible to support controls made of more than one bitmap piece than the clipping would not be a problem at all - it would be just two textures instead of two bitmaps that get drawn to a sprite...

But it's not the point. If the clipping alone was a problem, we could simply resize the clipping area.
Regardless of whether there are 1 texture or multiple textures, - the primary is the question of design (rule): do we allow an image to go outside of the control's rectangle (x,y,width,height), or not. The design choice may be then solved using various methods, depending on drawing implementation.

ericoporto:
slider

hey, the last commit fixed the clipping at left, but it's still clipping at right.

ivan-mogilko:
Hm, it was supposed to keep things unchanged, at least for 3.5.1

I'm currently trying to implement your suggestion about offsetting bar by the half of the handle.

ivan-mogilko:
So, here's another option. I may keep the slider's bar same width, but make the interactive field (i call it "value field") smaller.
This way there won't be a need to fix any background images, and the handle will be confined within the bar, instead of going little off.

ivan-mogilko:
Hmm, no. Upon further testing and comparison, it seems that this also breaks the vertical handle's position on horizontal sliders in these templates: they either get clipped or moved down (depending on enabling or disabling certain piece of code), changing their looks.

It's apparent that in regards to sliders no one really cared about their actual control bounds, so graphics could easily go outside of them.

If these restrictions are implemented, we'd also have to redesign the slider's look using some good rule, and maybe add some customization, such as let users select where the bar is located (middle, bottom or top of the control), etc.

I'd close this pr for now, and instead *turn clipping off exclusively for the sliders* to keep their looks.

Some bits of this pr may be still applied, as a code refactor.

Then we may return to this problem later when the 3.6.0 is in Beta, better tested and more serious problems fixed.

ericoporto:
the important part of solving control area or clipping at some point is to enable a transition to the implementation in #1393 , which in turn I think is useful for ags4 where there is GUI Rotation and possibly will have some form of scaling at some point. But as you mention, it may not be as important for 3.6.0.

ivan-mogilko:
> the important part of solving control area or clipping at some point is to enable a transition to the implementation in https://github.com/adventuregamestudio/ags/pull/1393 ,

Like i mentioned in one comment above, it's not the clipping that is a problem, but a design rule. The #1393 is a different way to implement drawing, but it must follow same design rules.

PR #1584, opened 2022-03-20
Exclude Sliders from the GUI clipping rule + refactor slider drawing

### LINKED ISSUE #1581: Sliders are clipped incorrectly

ivan-mogilko:
Fixes #1581 + refactors its drawing function. Latter is not necessary, but I wrote this code as a part of previous pr attempt (#1583), and imho this may be useful, as it reduces code duplication and clarifies certain settings.

The idea is that Sliders are currently too complicated to simply clip them to the rectangle defined by their position properties. As AGS historically did not restrict the sizes of slider handle and background images, clipping it would perhaps require to redesign its default looks and settings.

### DIFF
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -281,8 +281,10 @@ void GUIMain::DrawAt(Bitmap *ds, int x, int y)
         if (!objToDraw->IsVisible())
             continue;
 
-        if (GUI::Options.ClipControls)
+        if (GUI::Options.ClipControls && objToDraw->IsContentClipped())
             subbmp.SetClip(RectWH(objToDraw->X, objToDraw->Y, objToDraw->Width, objToDraw->Height));
+        else
+            subbmp.ResetClip();
         objToDraw->Draw(&subbmp);
 
         int selectedColour = 14;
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
@ericoporto I think existing 3.6.0 templates work as intended with these changes even if GUI clipping is on; but could you double check in case i missed anything?

ericoporto:
I built locally here and I did not notice any difference regarding the sliders, they appear good! 
- The slider test is like this
image
- The sierra slider appears unchanged
image
- All looks good in tumbleweed
image

Basically figure out what to write in the Beta topic in the forums and revise if anything more is needed from the milestone stuff for the beta and that is it as far as I can tell.

PR #1586, opened 2022-03-24
Resolving (most) engine compilation warnings

### LINKED ISSUE #46: tons of warnings with -Wall -Wextra
http://sprunge.us/hVVL

most of them are because of unused args
if you don't use args, you can make the warning go away like this

```
void func(int arg) {
  (void) arg; // removes the unused warning
}
```

ivan-mogilko:
Resolves #46 .

In the past couple of months i've been working on resolving compilation warnings. Some of these ended up real bugs and oversights, so this work was definitely not in vain. For the real mistakes and bigger refactors I've been making commits directly into master. This PR addresses too numerous but mostly trivial warnings; i made it to be sure that it builds everywhere, as there are several platform-specific ifdefs in the code which I might have missed.

If necessary i might further split commits into smaller ones, and add more fixes if something interesting is found in CI logs.

For warnings lookup I used:
* MSVS 2015 C++ compiler with /W4 option; I found /Wall to be *way* too verbose with this compiler, reporting obscure warning types which i was not interested in (and even reported lots of warnings for standard headers).
* gcc 5.4.0 with -Wall -Wextra (run using our plain makefile on a Ubuntu vm).

I focused on warnings strictly in the engine code, but decided to fix few annoying ones in the allegro headers, as they are included in lots of places. But did not touch:
* library sources (allegro, alfont, freetype, and so on);
* plugins

Two most common warning types were "unused argument/variable" and type cast issues. The scripting api was perhaps the primary source of "unused argument" because every script function is linked to vm using same uniform function prototype.

I did not aim for 100% removal of warnings; for instance left some typecast and comparison ones for the latter, to look into them more carefully. Some warnings should probably be left as-is for the time being, as they don't cause trouble but indicate a need to rework the code.

Did not count for exact numbers, but judging by gcc output this pr makes the number of warnings go from many hungreds to maybe couple of dozens (counting only engine code, in Common and Engine dirs).

### DIFF
--- a/Common/ac/dynobj/scriptaudioclip.cpp
+++ b/Common/ac/dynobj/scriptaudioclip.cpp
@@ -22,9 +22,9 @@ void ScriptAudioClip::ReadFromFile(Stream *in)
     id = in->ReadInt32();
     scriptName.ReadCount(in, SCRIPTAUDIOCLIP_SCRIPTNAMELENGTH);
     fileName.ReadCount(in, SCRIPTAUDIOCLIP_FILENAMELENGTH);
-    bundlingType = in->ReadInt8();
-    type = in->ReadInt8();
-    fileType = in->ReadInt8();
+    bundlingType = static_cast<uint8_t>(in->ReadInt8());
+    type = static_cast<uint8_t>(in->ReadInt8());
+    fileType = static_cast<AudioFileType>(in->ReadInt8());
     defaultRepeat = in->ReadInt8();
     defaultPriority = in->ReadInt16();
     defaultVolume = in->ReadInt16();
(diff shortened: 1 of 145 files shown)


ericoporto:
I only read the code, so I haven't tested, but it looks alright. 

I am curious about the choice of commenting the parameters instead of removing them, are because they are in some virtual table/inheritance/or need to implement some interface where the parameter exist on other implementation?

There are some opengl/direct3d code changes here that are the correct implementation and previous ifdef usage/behavior was plain wrong.

Overall it looks pretty safe to merge for me.

ivan-mogilko:
> I am curious about the choice of commenting the parameters instead of removing them, are because they are in some virtual table/inheritance/or need to implement some interface where the parameter exist on other implementation?

I tried to remove parameters completely where it was possible, but most of the cases are either implementations of a virtual inheritance, or, in case of script api, have to follow exact function prototype to be linked to script vm.

ivan-mogilko:
Split into more commits; removed some func args instead of hiding where possible; even removed few dummy functions (remains of some refactor maybe).

PR #1591, opened 2022-03-29
Use more common code for config reading and writing among platforms

### LINKED ISSUE #104: Unified configuration-reading routines

ivan-mogilko:
Resolves #104.

* Reorganized config reading and writing functions, added read parameters used by other platforms (such as min/max value range).
* Moved common config management from Android and iOS into a file called "mobile_base.cpp". I noticed that both platforms host and read exactly same options, so there's no reason to have duplicate code. Made sure this code uses common cfg read/write functions.
* Reorganized platform-dependent file lists in CMakeList.txt, to have all non-common code under os condition. I am not well versed in cmake yet, so if this is wrong, i may return things back.
* Not related to config, but i noticed that selectLatestSavegame may be replaced by common functions too. If wanted, this could be done as a separate pr.

Potential todo:
* remove unused options (i noticed both Android and iOS were still using deprecated options, some refering Allegro backend). But this will require also adjusting them in corresponding java/objective-c parts.
* there's also getAvailableTranslations function that may be reimplemented using available engine's utilities, but maybe it's too much for this pr.

### DIFF
--- a/Common/util/ini_util.cpp
+++ b/Common/util/ini_util.cpp
@@ -84,6 +84,13 @@ String CfgReadString(const ConfigTree &cfg, const String &sectn, const String &i
     return str;
 }
 
+void CfgReadString(char *cbuf, size_t buf_sz,
+    const ConfigTree &cfg, const String &sectn, const String &item, const String &def)
+{
+    String str = CfgReadString(cfg, sectn, item, def);
+    snprintf(cbuf, buf_sz, "%s", str.GetCStr());
+}
+
 //-----------------------------------------------------------------------------
 // ConfigWriter
 //-----------------------------------------------------------------------------
--- a/Common/util/ini_util.cpp
+++ b/Common/util/ini_util.cpp
@@ -11,12 +11,13 @@
 // http://www.opensource.org/licenses/artistic-license-2.0.php
 //
 //=============================================================================
-
+#include <cstdio>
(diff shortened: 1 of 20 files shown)


ericoporto:
I don't understand the platform stuff moved in CMake, in theory the unwanted file won't be built because of the ifdefs in the code itself, did it not work for you?

ivan-mogilko:
> I don't understand the platform stuff moved in CMake, in theory the unwanted file won't be built because of the ifdefs in the code itself, did it not work for you?

What was the reason to keep these files for platforms that do not need them?

> I also think https://github.com/adventuregamestudio/ags/issues/104 means to use acsetup.cfg everywhere.

Originally it did, probably, but since it looks like this cannot be done now, we may postpone the config file merge.
Right now i'm interested in reducing the code duplication to make this easier to maintain.

ericoporto:
> Was there a reason to keep these files in the "common" list for all platforms?

Helps people not forget things...

ericoporto:
Looks all good now. After this is in we should probably figure a issue that describes getting rid of the `android.cfg` file. I made a issue for the problem of the CI not failing for Android, I will look at it soon.

Issue #702, opened 2019-04-04
Editor crashes when importing a large number of sprites


morganwillcock:
Both the existing build of the Editor (tested on 3.4.3.1) and the work in progress one, will crash when too many sprites are imported and the project data is not saved. This will be an issue for bulk sprite re-import, to recreate a missing spritecache.

As a test, I imported the same image (720x450 32bit PNG) 500 times, and memory usage went from around 30 MB to around 700 MB. When saving the project, usage returns to 30MB. A second import of the same 500 images (for a total of 1000) will sometimes crash or sometimes complete. Typically the memory usage at the time of crashing is between 1.2 GB and 1.5 GB.

I've tried setting `<gcAllowVeryLargeObjects enabled="true"/>` in app.config, but this didn't make a noticeable difference to the memory usage at time of crashing.

A memory snapshot shows that the usage is down to AGS Native (you have to enabled mixed-mode in the memory profiler to see it at all).

![mem2](https://user-images.githubusercontent.com/14958747/55591048-7a16c180-572c-11e9-97a1-e9e5ca0c98b7.png)

ivan-mogilko:
As a quick thought, guess the only proper way to handle this is to introduce periodical saving and freeing the cache or part of the cache. But assuming default sprite file is a part of the project and by design should not be saved unless user hits Save, then we need another temporary sprite file that is being appended and then copied into (or replaces) default file upon saving the project.

morganwillcock:
I think it might be better to postpone the task of trying to recreate the spritecache from scratch and look at moving its build process into the game compiler (so project/design-time data is always separate from game data). So far (for version 3.5 of the editor) I don't think we have removed any functionality, only fixed or improved what was there. For this (and fixing edge cases for clipboard imports and room palettes, and concurrent editing of room data) I think it probably aligns better with a change in approach that gives a more straightforward solution for all of these issues.

I keep meaning to re-write my notes that were on the YouTrack server here, but would I just open that in a new issue or should I add it as a project?

ivan-mogilko:
If I knew this is going to be that complicated I'd work on this too instead of wasting time on not so important things! This feature was one of the reasons I kept working on AGS in this year.

ivan-mogilko:
Alright, so, we call it a failure and wrap the version up. We need to restructure the milestone list to know what else left to be fixed.

ivan-mogilko:
I am not certain, but I think I am going to look into this when I'm done with my current tasks.
[...]

PR #1594, opened 2022-03-30
Editor: feature recreating full spritefile from sources

### LINKED ISSUE #1404: Recreate Spritecache when no spritecache is found by Editor
**Describe the problem**
Spritecache is a big binary file that is hard to store in source control systems like git, requiring all sorts of workarounds. I would like to be able to restore the spritecache from the source files (images) in the project directory and the `Game.agf` file information (which holds things like the sprite ID, sourcefile and import configs.

**Suggested change**
I have often been thinking that #702 is a requirement for this but maybe there's some other way, so I decided to open a separated issue for just that. Being frank, if it solves at least just this case, it would bring a huge quality of life improvement.

The idea is, when the Editor loads a project, if no spritecache is found, it would import all sprites on load and recreate the missing spritecache.

**Relevant forum topics**
- https://www.adventuregamestudio.co.uk/forums/index.php?topic=57647.0
- https://www.adventuregamestudio.co.uk/forums/index.php?topic=48685.0

ivan-mogilko:
Meant to resolve #1404, but currently is a draft; still need to try this out and figure out the best approach to running the recreation tasks.

Implemented managed `SpriteFileWriter`, which is a C++ CLI wrapper around native SpriteFileWriter, also performing necessary conversions between .NET and native data.

Implemented `SpriteTools.WriteSpriteFileFromSources` method that writes a spritefile to the requested destination by reimporting images from sprite sources, or resorting to getting them from the current spritecache, if available.
If image is not available, then simply writes a dummy bitmap of the known size.
This process gets one sprite at a time, and disposes it after writing, thus the memory use supposed to stay low.
This function uses game's sprite compression settings.

Added a menu command "File -> Recreate Sprite File from sources", that writes a new spritefile using aforementioned method, and replaces game's file with it.

In case a spritefile is missing in the game, Editor no longer refuses to load a game, but instead generates a dummy spritefile with one sprite at index 0 (it is required as placeholder to prevent crashes). This allows user to get into the project, and try reimporting sprites either using "File -> Recreate Sprite File from sources", or commands from the sprite manager.
NOTE: all the sprite properties remain, as they are part of the Game.agf rather than a spritefile. Only images are lost in this situation.

### DIFF
--- a/Common/ac/spritefile.h
+++ b/Common/ac/spritefile.h
@@ -186,7 +186,9 @@ public:
     // Get the sprite index, accumulated after write
     const SpriteFileIndex &GetIndex() const { return _index; }
 
-    // Initializes new sprite file format
+    // Initializes new sprite file format;
+    // store_flags are SpriteStorage;
+    // optionally hint how many sprites will be written.
     void Begin(int store_flags, SpriteCompression compress, sprkey_t last_slot = -1);
     // Writes a bitmap into file, compressing if necessary
     void WriteBitmap(Bitmap *image);
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -986,7 +986,7 @@ namespace AGS.Editor
             return dialogScripts;
         }
 
-        private object CompileScripts(object parameter)
+        private object CompileScripts(IWorkProgress progress, object parameter)
         {
             CompileScriptsParameters parameters = (CompileScriptsParameters)parameter;
             CompileMessages errors = parameters.Errors;
@@ -1069,7 +1069,7 @@ namespace AGS.Editor
             }
         }
 
-        private object CreateCompiledFiles(object parameter)
+        private object CreateCompiledFiles(IWorkProgress progress, object parameter)
         {
             CompileScriptsParameters parameters = (CompileScriptsParameters)parameter;
             CompileMessages errors = parameters.Errors;
@@ -1651,7 +1651,7 @@ namespace AGS.Editor
             }
         }
 
-        private object SaveGameFilesProcess(object parameter)
+        private object SaveGameFilesProcess(IWorkProgress progress, object parameter)
         {
 			WriteConfigFile(Path.Combine(OUTPUT_DIRECTORY, DATA_OUTPUT_DIRECTORY));
 
--- a/Editor/AGS.Editor/BuildTargets/BuildTargetDebug.cs
+++ b/Editor/AGS.Editor/BuildTargets/BuildTargetDebug.cs
@@ -40,7 +40,7 @@ namespace AGS.Editor
             Utilities.TryDeleteFile(filename);
         }
 
-        private object CreateDebugFiles(object parameter)
+        private object CreateDebugFiles(IWorkProgress progress, object parameter)
         {
             Factory.AGSEditor.SetMODMusicFlag();
             CompileMessages errors = (parameter as CompileMessages);
--- a/Editor/AGS.Editor/Components/FileCommandsComponent.cs
+++ b/Editor/AGS.Editor/Components/FileCommandsComponent.cs
@@ -198,7 +198,7 @@ namespace AGS.Editor.Components
 
         private object RecreateSpriteFileProcess(IWorkProgress progress, object parameter)
         {
-            Utils.SpriteTools.WriteSpriteFileFromSources((string)parameter);
+            Utils.SpriteTools.WriteSpriteFileFromSources((string)parameter, progress);
             return null;
         }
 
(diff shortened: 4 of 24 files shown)


ericoporto:
Hey, so I tried to make it work with a game that really had a ton of sprites - acsprset.spr has around 4.2 GB, no compression on it.

Monitoring on the task manager I can see the Editor RAM going up, and at some point I get this 
image

My gues is it still hits #702 . I am not sure on the solution for when someone clicks on the recreate sprite source on the menu, but if a solution is made that does so right when the Editor loads the project, it should be ok to save it to disk directly.

ivan-mogilko:
> Hey, so I tried to make it work with a game that really had a ton of sprites - acsprset.spr has around 4.2 GB, no compression on it.

> Monitoring on the task manager I can see the Editor RAM going up, and at some point I get this

Please clarify, when was the RAM going up, is it during adding the sprites or during recreation?
Did the sprites have the source files, or no?
Have you called recreation right after adding big sprites, or during the next launch?

Because this method still allows to refer to spritecache, it may try loading some sprites into the cache memory from the old sprite file - if they dont have sources. I'd imagine if the memory had been already taken by unsaved sprites, then such process could get over the memory cap.

ericoporto:
> Did the sprites have the source files, or no?

It had the source files but not the acsprset.spr file, because I had deleted it and any backup. 

> Please clarify, when was the RAM going up, is it during adding the sprites or during recreation?

After clicking the recreation options, if you load task manager you can see it starts going up and then at around 1.2GB the error happens.

> Have you called recreation right after adding big sprites, or during the next launch?

After adding the sprites and saving once per time, I saved and closed the AGS Editor. Then, I opened Windows Explorer and deleted `acsprset.spr` and backup. Then I reopened AGS Editor, selected the project and just after it opened I went to File->Recreate Sprite File from Sources.

Here's an example project that reproduces the problem, it's made by reimporting the same image over and over:  [TestSpriteRecreate.zip](https://github.com/ericoporto/ags/releases/download/v.3.6.0.20.afix/TestSpriteRecreate.zip) ***(warning: when this actually works you will have a 4GB acsprset.spr file in your harddrive!!!)***

> Sorry, but i don't understand what you are saying, could you give more clear steps to observe these situations?

This is the editor after loading a game that has no acsprset.spr file:
<details>
</details>

Now this is an example of a project where I loaded Game.agf and intentionally added two more entries on the sprite list that are invalid (I also deleted one entrie just for fun), this will make the editor miss a sprite in the acsprset.spr file, but it appears as a bluecup in it (last two sprites)

<details>

</details>

ivan-mogilko:
> After clicking the recreation options, if you load task manager you can see it starts going up and then at around 1.2GB the error happens.

Well, this does not sound right, this means there's something wrong with the current code.

ivan-mogilko:
Ah! There's a stupid memory leak in one function. I'll rebase the branch with fixes.

ericoporto:
@ivan-mogilko after the fix I don't have the previous mentioned problem! It works!

I think think this is already good enough and we could use it. Further improvements in the future:

- the progression shows a bar so you know how far the process is (I have a really fast SSD and it still took a long time, it would be nice to see how far things are to know it is actually doing something)
- when there's no acsprset.spr at load, it would be recreated automatically if the user builds/rebuilds their game, during the compilation process.

ivan-mogilko:
> Now this is an example of a project where I loaded Game.agf and intentionally added two more entries on the sprite list that are invalid (I also deleted one entrie just for fun), this will make the editor miss a sprite in the acsprset.spr file, but it appears as a bluecup in it (last two sprites)

Please tell, what did you add into Game.agf, what properties do these sprites have?

ivan-mogilko:
> This is the editor after loading a game that has no acsprset.spr file:
> Now this is an example of a project where I loaded Game.agf and intentionally added two more entries on the sprite list that are invalid (I also deleted one entrie just for fun), this will make the editor miss a sprite in the acsprset.spr file, but it appears as a bluecup in it (last two sprites)

So, this turns to be simple, the missing sprites are displaying placeholder from sprite 0. In the first game sprite 0 is recreated as a 1x1 clear image, and in the second game it's blue cup sprite.

ivan-mogilko:
> when there's no acsprset.spr at load, it would be recreated automatically if the user builds/rebuilds their game, during the compilation process, the first time it's needed.

Added a suggestion to recreate a missing sprite file on game load.

I'm not certain whether doing this automatically on compilation is a good idea or not, because this may be a lengthy process, and become a surprise of a user. Also, not sure how to deduce that it must be done. User may have their own plans about sprites.

ericoporto:
> there's also automatic game compilation. I think, it may be done for that without asking, or maybe require an additional command option?

You mean when one passes `/compile` and a game.agf to the Editor? That sounds nice, yeah, I think it's alright to do that!

I think the PR is good, this will be a really nice addition!

PR #1601, opened 2022-04-05
3.6.0: remove Game Explorer integration from the engine and editor

ivan-mogilko:
Game Explorer seem to be discontinued in Windows itself since one of the Win 10 updates, and it's not clear whether this was used much by AGS game creators. Today there are other popular game platforms and libraries (steam, gog, and so on) which won't benefit from this engine function. Even if wanted, same functionality may be recreated as a stand-alone tool or script, as it is required only during game's installation, not run.

1. Removed Game Explorer resource integration from the Editor. Deprecated few game properties, and left few others as temporarily unused, as some may be utilized as game's meta-settings in the future: they include fields like "description", "release date" and "version". The deprecated properties are:
  * GameExplorerEnabled
  * WindowsExperienceIndex
  * EnhancedSaveGames

2. Removed Game Explorer integration feature from the engine. It was normally used by running game.exe with --registergame and --unregistergame commands. This function acted as a part of the game's installation only, and was not required to run the game. Therefore, imo, does not have to be a part of the engine itself.

3. Removed "Rich Media Header" from the game saves. This had been used by Windows Game explorer feature to display custom thumbnails and other meta data for the save files. According to ms docs it is no longer supported even on Windows: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ee417684(v=vs.85)
Not sure why I did not remove this years ago, or when rewriting the save format for 3.4.0, probably just kept forgetting.
The engine still accounts for older saves having this data, so tries to skip its size if the AGS own save header is not found at the start of the file.


ericoporto:
Hey, the release date property, I know it won't be used anywhere, but can it be kept editable and available in the general settings? I use it for my games/modules for my own checking.

ivan-mogilko:
> Hey, the release date property, I know it won't be used anywhere, but can it be kept editable and available in the general settings? I use it for my games/modules for my own checking.

It's not removed, the deprecated properties are:
* GameExplorerEnabled 
* WindowsExperienceIndex 
* EnhancedSaveGames 

> Edit: can an obsolete property be brought back if an use appears? If they can, I would suggest merging this and later figuring out the Editor part if someone else asks a property back.

Yes, it's a matter of removing [obsolete] attribute.

Issue #1590, opened 2022-03-28
New compiler: static variables inside a struct and doubt about pointer types in conditions


AlanDrake:
## Static variable inside struct
@fernewelten I noticed a case with `static` where the compiler doesn't complain, but things don't work as one would expect:

```
struct MyStruct {
	static bool fake_static_bool;
}
```

Normally, only `static attribute` would be allowed. From what I tested, static (not attribute) variables inside structs  don't seem to be actually implemented, so it may be cause of undefined behaviors.

It wouldn't be bad to have actual static variables without all the attribute boilerplate, though.

## About pointer types

Previously it was possible to test pointer types directly like this
```
if (pointerType)
```
Without explicitly testing againt null, but now it throws the error `'if' condition: Cannot convert a pointer to 'int'`.
I don't know if this was intended, so I'm reporting it.
I personally don't mind either way, but sometimes it's nice to omit the `!= null`.

fernewelten:
Hi @AlanDrake , sorry for the delay. My mother has died, and everything is in turmoil right now. I'll issue fixes pronto, I'm starting right now.

AlanDrake:
I'm sorry to hear that, fernewelten. Take all the time you need, familiy should always come first.

fernewelten:
Addressed in #1602 and #1603.

ivan-mogilko:
@AlanDrake could you double check that the above PRs cover this issue?

AlanDrake:
Looks good to me.

PR #1602, opened 2022-04-05
AGS4: New compiler: Disallow static variables in `struct`s

fernewelten:
This partially addresses #1590.

Fix bug: AGS doesn't have static variables in `struct`s although there __are__ static functions and static attributes. The new compiler didn't catch this.

Typical code:
```
struct MyStruct {
	static bool foo; // ← Compiler should complain at ';': 
                         //   Only function decls. may start with 'static' and no 'attribute'
}
```


ivan-mogilko:
I've compared error messages in the old and new compiler, -

Old compiler: `static variables not supported`
New compiler: `Expected '(', found ';' instead`

Would it be possible to have similar error in the new compiler? Saying that compiler expected bracket after variable declaration may be misleading imo.

fernewelten:
> Would it be possible to have similar error in the new compiler?

Here you go :-)

PR #1608, opened 2022-04-09
Script API: allow to scale overlays by setting Width & Height

### LINKED ISSUE #1585: Support Setting Width and Height of Overlays
**Describe the problem**
Overlays got a readable Width and Height. That is great. But being able to actually set those values would enable actually manipulating a texture through AGS Script, which would allow a great amount of effects - plus you could distort things by making width different than height, which could be really fun!

**Suggested change**
Make Width and Height a writable property, at least in graphical overlays!

Use case: wanted to replace my scripted software renderer in [mana match](https://eri0o.itch.io/mana-match) to leverage the overlays.

ivan-mogilko:
Resolves #1585.

1. Makes `Overlay.Width` and `Height` settable (previously were read-only). Changing them will scale overlay's image.
2. Added readonly `Overlay.GraphicWidth` and `GraphicHeight`, which return original image's size. This is necessary, because overlays may have autogenerated image, which sizes cannot be found using `Game.SpriteWidth/Height`.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1292,6 +1292,10 @@ builtin managed struct Overlay {
   import attribute int Width;
   /// Gets/sets the height of this overlay. Resizing overlay will scale its image.
   import attribute int Height;
+  /// Gets the original width of this overlay's graphic.
+  import readonly attribute int GraphicWidth;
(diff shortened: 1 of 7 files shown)


ericoporto:
I verified this is working!

keys_and_more_cups

[keysAndCups.zip](https://github.com/adventuregamestudio/ags/files/8457948/keysAndCups.zip)

PR #1611, opened 2022-04-10
Editor: letterbox mode moved to compatibility category

ericoporto:
It looks like the `Old-style letterbox mode` doesn't belong in the `(Basic properties)` category anymore, and may be more well suited to the Backward Compatibility category.

This change moves it.


ivan-mogilko:
The reason why I haven't moved it there is that it is a part of the Resolution setting.

ericoporto:
As far as I can tell, the categories are only used by the Editor rendering them in the General Settings tab:

image

In Game.agf, nothing changes, it's still a `LetterboxMode` entry under `Settings`.

PR #1613, opened 2022-04-14
Script API: implemented GUIControl.Transparency

ivan-mogilko:
As #1393 makes it possible to accelerate gui controls, there's a next step to improve gui perfomance: support GUIControl.Transparency, implemented as a texture's render mode. Before that if one would like to have a translucent button graphic, they would have to either put that button on a standalone GUI, or use raw drawing (DrawImage with transparency arg) which is slower, especially in high-res games.

Adding transparency property itself does not require much changes, as all the logic is there. But since controls are part of the GUI sprite batch, this also means supporting batch's own alpha.

### DIFF
--- a/Common/gfx/gfx_def.h
+++ b/Common/gfx/gfx_def.h
@@ -35,11 +35,12 @@ enum BlendMode
 
 namespace GfxDef
 {
+    // Converts percentage of transparency into alpha
     inline int Trans100ToAlpha255(int transparency)
     {
         return ((100 - transparency) * 255) / 100;
     }
-
+    // Converts alpha into percentage of transparency
     inline int Alpha255ToTrans100(int alpha)
     {
         return 100 - ((alpha * 100) / 255);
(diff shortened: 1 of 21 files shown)


ericoporto:
The Script API appears to work!

Would it be possible to also add a property for GUI Controls in the Editor itself? (even if it's not shown in there)

ivan-mogilko:
Not a single object has transparency exposed in the editor. If I add to the controls, then I should perhaps add for everything?

ivan-mogilko:
Hmm, this is strange, I thought GUI does not have Transparency in the editor, but apparently it does. Maybe I missed where it is declared.

ericoporto:
Ah, right, the class has a weird name... For GUIs, it's here: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Types/NormalGUI.cs

ivan-mogilko:
According to the past decision, all the new fields to the game data are serialized using extensions:
https://github.com/adventuregamestudio/ags/blob/v.3.6.0.22/Editor/AGS.Editor/DataFileWriter.cs#L1762-L1771

PR #1619, opened 2022-04-17
Installer: Editor shortcut in start menu includes version

### LINKED ISSUE #1617: Installer: add version numbers to the created links
When AGS Editor is installed on a system, it has 3 first version numbers in the title, however these are not added to the link files. As a result it may be difficult to recognize a version when running AGS from a Start menu (or desktop).

The suggestion is to also add these 3 numbers to all the links created by the installer.
In other words change:
* AGS Editor
to
* AGS Editor X.X.X

ericoporto:
fixes #1617

image


ivan-mogilko:
Thank you.

PR #1632, opened 2022-04-23
Editor: fixed "Select area" tool issues

### LINKED ISSUE #1625: AGS 4: using "area picker" tool selects wrong area index (index -1), may cause unhandled exception
**Describe the bug**
Using a "picker" tool in the room editor would select not the chosen area, but the area of (index-1). If you try to pick an area 1 editor throws an unhandled exception.

EDIT: may be related, so did not open a separate ticket: using the picker outside of the room background also causes exception.

**AGS Version**
3.99.99

**To Reproduce**
1. Make a new room.
2. Paint few walkable areas, say, 1,2,3...
3. Switch to area picker tool and try to select area 3. Notice how area 2 is selected.
4. Try to select area 1. Observe the exception.

**Expected behavior**
Proper area selected, no unhandled exceptions should be caused by this.

AlanDrake:
Corrected missing adjustment for mask resolution and handled out of bounds coordinates.

Closes #1625 

Is it fine like this or should the x/y traslation be moved elsewhere?

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1738,7 +1738,15 @@ namespace AGS.Editor.Components
                 throw new InvalidOperationException("No room is currently loaded");
             }
 
+            // adjust coordinates to mask resolution
+            double scale = _loadedRoom.GetMaskScale(maskType);
+            x = (int)(x * scale);
+            y = (int)(y * scale);
+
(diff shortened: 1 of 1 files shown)


persn:
I was already certain it already did this scaling. I distinctly remember writing code for it. I'll have to check later when I'm by a computer

persn:
I was thinking about somewhere else, this scale multiplication is done lots of places

persn (approved):

PR #1633, opened 2022-04-23
AGS 4 Editor: Fixes importing room template and existing room

AlanDrake:
One problem was that the open format conversion function expects files with extension, while the room template unpacking saves them as "room3asc" and similar. 

The other issue with importing an existing room was that the conversion was not happening.

I tested game template, room templates, adding existing room, and creating a blank room. I hope I covered all the cases with this last fix.


ivan-mogilko:
> One problem was that the open format conversion function expects files with extension, while the room template unpacking saves them as "room3asc" and similar.

That's actually a bug in 3.6.0, something got broken.

PR #1635, opened 2022-04-26
CMake: fix missing change to script/cc_internal.h

ericoporto:
Edit: feel free to include this directly in the master branch, adding script/cc_internal.h there is not needed to build but is needed for editing `script/cc_internal.h` in CMake based IDEs.

### DIFF
--- a/Compiler/Makefile
+++ b/Compiler/Makefile
@@ -23,8 +23,7 @@ LDFLAGS  += -rdynamic -Wl,--as-needed $(addprefix -L,$(LIBDIR))
 CFLAGS   += -Werror=implicit-function-declaration
 
 COMMON_OBJS = \
-	../Common/script/cc_error.cpp \
-	../Common/script/cc_options.cpp \
+	../Common/script/cc_common.cpp \
 	../Common/script/cc_script.cpp \
 	../Common/util/bufferedstream.cpp \
 	../Common/util/cmdlineopts.cpp \


ericoporto:
Checking test... just a sec

ericoporto:
I found that the compiler is the thing with problems

https://github.com/adventuregamestudio/ags/blob/e33b79399d6109ff749933d112ea301a5248b08c/Compiler/compiler.cpp#L18-L19

ivan-mogilko:
There's a number of things that has to be changed in compiler, and compiler tests, including changing how the error is formatted.

ericoporto:
I am making this a draft to avoid accidental merge, this will probably be closed once building is figured out.

ivan-mogilko:
> It will be really boring to figure lines in all tests... 

Tests are supposed to use separate error formatting, there were specific functions for that, they need to be adjusted to the new code.

ericoporto:
I think `cc_get_callstack` also need to be defined as empty in the compiler like in the agsnative case has one, not sure where to put it.

ivan-mogilko:
I pushed another commit for the tests. Few tasks have built successfully, so maybe it's ready; or almost ready.

ivan-mogilko:
Oh, did not realize you are speaking of Makefile and not CMake lists.

ericoporto:
ok, I think the Makefile was the last thing missing so I am going to close this now... I will recheck the compiler and the tests later to verify they are alright when running them.

PR #1639, opened 2022-04-28
3.6.0: Room Overlays

### LINKED ISSUE #1044: Feature request: Room Overlays
Room Overlays work exactly like Overlays in AGS, the difference is they exist in Room Space. Their transformations (coordinates, scaling, ...) are relative to the room, meaning, they are affected by Camera/Viewport. Overlays are the most thin and simple and direct elements in AGS, that are closer to a texture on the screen, so at the same time, there's a balance to not add more things to them to keep them simple, which is relevant to whether this should or should not be implemented.

Previously @ivan-mogilko has made [an experiment for Room Overlays](https://github.com/ivan-mogilko/ags-refactoring/commit/e91dbe5ad74861a9d39589e0d01433169d49a687) by using a property that when set would make the Overlay be in the room. I linked the code above because there are unanswered questions in the code comments that I believe are relevant. 

My usage of Room Overlays would be use them for particles and objects that have no animation but are created dynamically with the agsbox2d plugin.

I wonder if it's better to solve #967 before introducing these.

### LINKED ISSUE #1209: In-Room Overlays sorted among other objects
Originally supposed to be a part of #1202, but it's more complicated and has more potential problems, so split into this ticket.

**To recap**, Overlay is a primitive object that has a sprite (may be one of the game sprites, or generated by printing a text on bitmap). At the moment it is only drawn upon a GUI layer, above all rooms.<br>
In special case its position may be given in room coordinates, but that simply means that it's placed over certain room location; it will still be drawn as a part of GUI layer. The use of this is that it moves to remain over certain spot in room when room camera moves. This is used for lucas-arts speech style.<br>
Since #1202 completed Overlay has ZOrder property like GUI and is sorted among other GUIs on screen.

### Proposal

* Implement in-room overlays. They have their position in room coordinates and drawn in room layer.
* Room overlays are created either by passing another argument into one of the existing Overlay.Create* functions or by adding new pair of functions for this purpose.
* Add readonly boolean property telling if this is a room overlay named IsRoomLayer, IsRoomOverlay etc.
* Room overlays are sorted among the room objects (characters, walk-behinds), where it's compared to their Baseline property, which is practically same thing (and has sorting rules similar to GUI & Overlay.ZOrder).

**Questions**

1. The biggest conceptual question is, should room overlay presence be tied to the room or not. That is, if a room overlay is created in room A, should it continue to display when player switches to room B or remain in room A (remain in memory, but not show up)?
**EDIT:** overlays are currently forcefully removed whenever player changes a room. Perhaps we could just leave this behavior as-is, equally for screen and room overlays.
2. As noted there are screen overlays whose position is given in room coords. These are not really room overlays, so they have a special state not covered by properties; this is done using internal flag. Is it possible to improve this somehow, in terms of design, keeping same behavior? Might this position mode be useful to game developers?
**EDIT:** i made a mistake here, although the overlay is positioning itself related to some point in room, but it still reports its position in screen coordinates in such case. So nothing extra is necessary to do here.

**Potential complications**

* Software renderer is made to not draw walk-behinds as separate sprites all the time; instead it has its own sprite cache for room object/character images which have parts of walk-behinds painted over. This supposedly improves perfomance when there's lots of walk-behinds and little moving objects. In opposite case it may be not worth it or maybe even have worse perfomance, so it's kind of a questionable function at the moment.<br>
If we keep this method for now, then room overlays have to be sharing same sprite cache and be subject to same procedure. If we don't then it's a matter of disabling this cache (code refactor might be necessary).

ivan-mogilko:
Resolves #1044.
Resolves #1209.

Implements room overlays, that is Overlays that are drawn in the room "layer", inside the room viewport and sorted among characters, room objects and walk-behinds. This opens large possibilities, as overlays are created at runtime and are unlimited since 3.6.0. They may be used both as a one time visual effects, or to build a custom room object from ground up.

Technically, not a lot of changes were necessary, we just use same overlays with a flag; depending on the flag they are drawn either as a part of room or part of UI layer. For software renderer room overlays are also processed for walk-behind cropping.

In script, I've been in doubts how to organize this, so ended up with two extra factory functions duplicating two existing ones:
```
  /// Creates an overlay that displays a sprite inside the room.
  import static Overlay* CreateRoomGraphical(int x, int y, int slot, bool transparent);
  /// Creates an overlay that displays some text inside the room.
  import static Overlay* CreateRoomTextual(int x, int y, int width, FontType, int colour, const string text, ...);
```

In regards to the sprite sorting, Overlays have ZOrder property since 3.6.0, and it is compared to other room elements baselines.
Overlays are **not** changing their z-order automatically though, so moving them along the Y axis don't affect their sorting order. IMHO that is proper, as overlays are the basic "sprite containers" without specific behavior.
If user wants to have overlay sorted depending on its Y pos, simply doing `over.ZOrder = over.Y;` or `over.ZOrder = roomover.Y + roomover.Height;` in rep-exec is enough.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1292,6 +1292,8 @@ builtin managed struct Overlay {
   import static Overlay* CreateRoomGraphical(int x, int y, int slot, bool transparent);  // $AUTOCOMPLETESTATICONLY$
   /// Creates an overlay that displays some text inside the room.
   import static Overlay* CreateRoomTextual(int x, int y, int width, FontType, int colour, const string text, ...);  // $AUTOCOMPLETESTATICONLY$
+  /// Gets whether this overlay is located inside the room, as opposed to the screen layer.
+  import readonly attribute bool InRoom;
   /// Gets/sets the width of this overlay. Resizing overlay will scale its image.
   import attribute int Width;
   /// Gets/sets the height of this overlay. Resizing overlay will scale its image.
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1288,6 +1288,10 @@ builtin managed struct Overlay {
   /// Gets/sets the Y position on the screen where this overlay is displayed.
   import attribute int Y;
 #ifdef SCRIPT_API_v360
+  /// Creates an overlay that displays a sprite inside the room.
+  import static Overlay* CreateRoomGraphical(int x, int y, int slot, bool transparent);  // $AUTOCOMPLETESTATICONLY$
+  /// Creates an overlay that displays some text inside the room.
+  import static Overlay* CreateRoomTextual(int x, int y, int width, FontType, int colour, const string text, ...);  // $AUTOCOMPLETESTATICONLY$
   /// Gets/sets the width of this overlay. Resizing overlay will scale its image.
   import attribute int Width;
   /// Gets/sets the height of this overlay. Resizing overlay will scale its image.
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -2728,7 +2728,7 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
             }
             if (game.options[OPT_SPEECHTYPE] == 3)
                 overlay_x = 0;
-            face_talking=add_screen_overlay(overlay_x,ovr_yp,ovr_type,closeupface, closeupface_has_alpha);
+            face_talking=add_screen_overlay(false,overlay_x,ovr_yp,ovr_type,closeupface, closeupface_has_alpha);
             facetalkframe = 0;
             facetalkwait = viptr->loops[0].frames[0].speed + GetCharacterSpeechAnimationDelay(speakingChar);
             facetalkloop = 0;
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -343,7 +343,7 @@ ScreenOverlay *_display_main(int xx, int yy, int wii, const char *text, int disp
 
         if (!overlayPositionFixed)
         {
-            screenover[nse].positionRelativeToScreen = false;
+            screenover[nse].SetRoomRelative(true);
             VpPoint vpt = play.GetRoomViewport(0)->ScreenToRoom(screenover[nse].x, screenover[nse].y, false);
             screenover[nse].x = vpt.first.X;
             screenover[nse].y = vpt.first.Y;
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -71,7 +71,8 @@ struct DisplayVars
 // Pass yy = -1 to find Y co-ord automatically
 // allowShrink = 0 for none, 1 for leftwards, 2 for rightwards
 // pass blocking=2 to create permanent overlay
-ScreenOverlay *_display_main(int xx, int yy, int wii, const char *text, int disp_type, int usingfont, int asspch, int isThought, int allowShrink, bool overlayPositionFixed)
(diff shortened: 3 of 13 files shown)


ericoporto:
https://user-images.githubusercontent.com/2244442/166164652-cc4cf143-d5a9-407d-a132-36a2f8323685.mp4

Took the opportunity to play around with my old project and update to the version from this PR. It does work and it's pretty fun!

ivan-mogilko:
Fixed overlays not saving their bitmaps (was broken by a cherry-picked commit).

ivan-mogilko:
Fixed speech incorrectly registered as a "room overlay".

ericoporto:
Hey, in the serialization, I noticed you switched two bools for a flag, does this implies any modifications in the saveformat version or is something that won't be impacted - by say, loading a previous savefile?

ivan-mogilko:
> Hey, in the serialization, I noticed you switched two bools for a flag, does this implies any modifications in the saveformat version or is something that won't be impacted - by say, loading a previous savefile?

The previous saves should load correctly because their layout on disk will match the flag values layout, but I probably still need to increase the component version because new version may have more values there.

ericoporto:
What's the difference between `kOver_PositionAtRoomXY` and `kOver_RoomLayer`?

ivan-mogilko:
> What's the difference between `kOver_PositionAtRoomXY` and `kOver_RoomLayer`? (they look like the same but I am probably not understanding the difference...)

There are overlays that are aligned to certain room position while being drawn on gui layer; example: Lucas-Arts style speech, background speech.

ivan-mogilko:
And I just realized that I set wrong value to this flag...

ivan-mogilko:
I found my older ticket on this: #1209. It has at least two remaining subtasks for this feature:
* *Add readonly boolean property telling if this is a room overlay named IsRoomLayer, IsRoomOverlay etc.*
* *The biggest conceptual question is, should room overlay presence be tied to the room or not. That is, if a room overlay is created in room A, should it continue to display when player switches to room B or remain in room A (remain in memory, but not show up)?*

> What happens if I switch rooms and have the room overlay pointer in a specific room script?

This is the same case as with any other pointer. The room script data is stored in memory at all times, but is not accessible. Thus the reference to the overlay remains, but you cannot control it unless you return to that room; or unless you somehow pass the pointer copy to a global script.

ericoporto:
>This is the same case as with any other pointer. The room script data is stored in memory at all times, but is not accessible if the room has changed. Thus the reference to the overlay remains, but you cannot control it unless you return to that room; or unless you somehow pass the pointer copy to a global script (before changing the room).

So, check this video.

https://user-images.githubusercontent.com/2244442/166340492-3d9e5f94-7f46-4e32-8255-8bd30c2f032e.mp4

ivan-mogilko:
Now I remembered that all overlays are removed when changing rooms.

ericoporto:
This is alright, but why it is not recreated?

ivan-mogilko:
> This is alright, but why it is not recreated? (check my room_Load function)

Might be some bug.

ivan-mogilko:
Opened a bug ticket: #1647

ivan-mogilko:
> *Add readonly boolean property telling if this is a room overlay named IsRoomLayer, IsRoomOverlay etc.*

Added `readonly bool Overlay.InRoom`

> *The biggest conceptual question is, should room overlay presence be tied to the room or not. That is, if a room overlay is created in room A, should it continue to display when player switches to room B or remain in room A (remain in memory, but not show up)?*

So, since all the overlays are traditionally just removed on room change, I guess we may leave it like that for now.

ericoporto:
InRoom sounds like it returns an integer (the room number), it would be better something  like InGroup or Layer, which would return an enum eLayerRoom or eLayerGUI.

ivan-mogilko:
I can rename to IsInRoom, or IsRoomOverlay.

We do not have layer enum, no other API uses that, and it's unclear whether will.

ivan-mogilko:
Well, I opened a PR: #1652, personally I have doubts about this.

PR #1640, opened 2022-04-29
AGS 4: compile rooms without using global native room object

### LINKED ISSUE #1629: AGS 4: when *.crm output file does not exist, on room save editor still tries to read "blank.crm" from resources
**Describe the bug**
After the change to open room format roomN.crm files are no longer sources, but output files.
However, if the file roomN.crm is not present in the project folder, and current room is being saved, the Editor still tries to load "blank.crm" template file from the embedded resources. As this resource no longer exists, it fails, throwing unhandled exception.

**AGS Version**
3.99.99

**To Reproduce**
1. Create a new game in ags4.
2. Delete all *.crm files in the project folder.
3. Open a room in the editor and try saving it.
4. Observe the exception

**Workaround**
Copy any random crm file into the game folder and rename it to match the missing one(s).

**Expected behavior**
As *.crm files are output files now in ags4, them missing is a normal situation. They should be recreated without any error or warning, and preferably not using embedded resources. If a native object is required for saving crm, a blank one should be created in memory programmatically.

ivan-mogilko:
Resolves #1629 in the sense that the process of saving CRM file no longer requires a file to exist.

Makes converting and saving room.crm file a single operation that does not require using a global native room object.
All necessary data is passed into NativeProxy.SaveRoom, where it is converted to a temporary room object, which is saved and disposed right after.

### DIFF
--- a/Common/game/roomstruct.h
+++ b/Common/game/roomstruct.h
@@ -59,7 +59,10 @@ enum RoomAreaMask
     kRoomAreaHotspot,
     kRoomAreaWalkBehind,
     kRoomAreaWalkable,
-    kRoomAreaRegion
+    kRoomAreaRegion,
+
+    kRoomArea_First = kRoomAreaHotspot,
+    kRoomArea_Last  = kRoomAreaRegion
 };
 
 // Room's audio volume modifier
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
I decided to try a different approach for some reason.

PR #1641, opened 2022-04-30
AGS 4: reimplemented native room operations as NativeRoom class

ivan-mogilko:
NativeRoom class wraps a native RoomStruct object and supports following operations:
* construct from a managed AGS.Types.Room;
* construct by loading from a CRM file (compiled room format);
* convert from native to managed AGS.Types.Room;
* retrieve backgrounds and masks, converting to System.Drawing.Bitmap;
* assign backgrounds and masks, converting from System.Drawing.Bitmap;
* saving native object to CRM file (compiled room format).

Removed `thisroom` global native object, because it's no longer usable anyway. Had to mark couple of places as broken (with FIXME) for the future.

Removed `_roomStructPtr` from the managed Room.

Removed precreating and loading a dummy CRM file when only saving the compiled room.

Removed unnecessary encoding conversion of compiled room files (since they are no longer project files but output artifacts).

### DIFF
--- a/Common/game/roomstruct.cpp
+++ b/Common/game/roomstruct.cpp
@@ -198,7 +198,19 @@ Bitmap *RoomStruct::GetMask(RoomAreaMask mask) const
     case kRoomAreaWalkBehind: return WalkBehindMask.get();
     case kRoomAreaWalkable: return WalkAreaMask.get();
     case kRoomAreaRegion: return RegionMask.get();
-    default: return nullptr;
+    default: assert(0); return nullptr;
+    }
+}
+
+void RoomStruct::SetMask(RoomAreaMask mask, Bitmap *bmp)
+{
+    switch (mask)
+    {
+    case kRoomAreaHotspot: HotspotMask.reset(bmp); break;
+    case kRoomAreaWalkBehind: WalkBehindMask.reset(bmp); break;
+    case kRoomAreaWalkable: WalkAreaMask.reset(bmp); break;
+    case kRoomAreaRegion: RegionMask.reset(bmp); break;
+    default: assert(0); break;
     }
 }
 
--- a/Common/game/roomstruct.h
+++ b/Common/game/roomstruct.h
@@ -59,7 +59,10 @@ enum RoomAreaMask
     kRoomAreaHotspot,
     kRoomAreaWalkBehind,
     kRoomAreaWalkable,
-    kRoomAreaRegion
+    kRoomAreaRegion,
+
+    kRoomArea_First = kRoomAreaHotspot,
+    kRoomArea_Last  = kRoomAreaRegion
 };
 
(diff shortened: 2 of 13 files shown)


ivan-mogilko:
@AlanDrake, @persn, could you also check this out if you have time?

AlanDrake:
I noticed that saving the room after editing a mask crashes to desktop without an error.

ivan-mogilko:
Updated, was doing an opposite conversion instead.

persn:
I'll try and check it out sometime soon

AlanDrake:
Now it doesn't always always crash on save, but still happening while drawing with different colors.

ivan-mogilko:
> but still happening while drawing with different colors.

Sorry, what do you mean by that, do you have steps to reproduce this?

AlanDrake:
Yes, draw with more than one color on a mask, then save. It doesn't crash if you draw just one color.
Unofrtunately it crashes to desktop without any error. I have some time now, so I'll take a look with the debugger.

ivan-mogilko:
> Yes, draw with more than one color on a mask, then save.

I tried, and it did not reproduce for me.

AlanDrake:
error

I ran it through the debugger, looks like some memory access violation triggering in the native code.

## Test project 
https://www.dropbox.com/s/ekssu0ywfzm48dx/nativeroom_ctd.zip?dl=0

## Steps
1. Open a room
2. Edit hotspots
3. Draw a blue rectangle
4. Draw a green rectangle
5. Save room
6. Crash

ivan-mogilko:
This does not reproduce to me...

Are you rebuilding full project, or only C# part?

Common/game/roomstruct.cpp
```diff
@@ -202,6 +202,18 @@ Bitmap *RoomStruct::GetMask(RoomAreaMask mask) const
     }
 }
 
+void RoomStruct::SetMask(RoomAreaMask mask, Bitmap *bmp)
+{
+    switch (mask)
+    {
+    case kRoomAreaHotspot: HotspotMask.reset(bmp);
```
AlanDrake, line 209: `return` is missing here, which was the cause of the last issue

AlanDrake:
I tested some more and it appears to be working fine now.

persn (approved):

ivan-mogilko:
> Off-topic but if these things are fixed does that maybe mean that it's close to possible to have multiple open tabs of the room editor? One tab for individual room. When I was coding the open rooms it seemed like to me that the only significant obstacle for that was the thisroom object in the native proxy.

From the look of it, the only case when it was still needed were palette remapping for imported sprites. I don't think it should be fixed it native, instead it's best to reimplement that in the managed code as soon as it is possible: now, or even later when the open sprite format is in place.

PR #1643, opened 2022-05-01
Tidy up mobile config once more and get rid of psp_* variables

### LINKED ISSUE #106: Rename psp_* variables

ivan-mogilko:
Resolves #106.

* Gathered mobile platform config options into MobileSetup struct.
* Removed all psp_* variables from the core engine code.
* Instead the engine receives port's special config using platform->ReadConfiguration().
* Mobile ports pass respective parts of the MobileSetup into the engine's own config.
* Added "load_latest_save" to the engine config.

More options could be transferred later (i noticed that some may still make sense but are not part of engine's config anymore).

### DIFF
--- a/Engine/ac/gamesetup.cpp
+++ b/Engine/ac/gamesetup.cpp
@@ -33,6 +33,7 @@ GameSetup::GameSetup()
     clear_cache_on_room_change = false;
     load_latest_save = false;
     rotation = kScreenRotation_Unlocked;
+    show_fps = false;
 
     Screen.Params.RefreshRate = 0;
     Screen.Params.VSync = false;
--- a/Engine/ac/gamesetup.cpp
+++ b/Engine/ac/gamesetup.cpp
@@ -31,6 +31,7 @@ GameSetup::GameSetup()
     RenderAtScreenRes = false;
     Supersampling = 1;
     clear_cache_on_room_change = false;
+    load_latest_save = false;
     rotation = kScreenRotation_Unlocked;
 
     Screen.Params.RefreshRate = 0;
--- a/Engine/ac/gamesetup.h
+++ b/Engine/ac/gamesetup.h
@@ -89,6 +89,7 @@ struct GameSetup {
     bool  clear_cache_on_room_change; // for low-end devices: clear resource caches on room change
     bool  load_latest_save; // load latest saved game on launch
     ScreenRotation rotation;
+    bool  show_fps;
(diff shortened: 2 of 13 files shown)


ivan-mogilko:
@ericoporto do you know where does the AGS_PLATFORM_DEFINES_PSP_VARS flag comes from? The psp_ variables were implemented in engine if this flag is not set, but I could not find it defined anywhere.

ericoporto:
@ivan-mogilko  it's defined in this line here:

https://github.com/adventuregamestudio/ags/blob/b482474a2d8b366d6b80dc9ccee8443d580138ba/Engine/main/main.cpp#L26

You can of course get rid of it if desired.

ericoporto:
Hey, so I tested this today, it mostly works, there's just a small thing I found, in `ApplyEngineConfiguration`, it's not complete, it's missing 
- `display_fps`, 
- `debug_write_to_logcat` (but this one I don't think it's needed there, but still, thought to mention), 
- sound stuff, but I think only `audio_enabled` is used.
- `mouse_control_mode`, which is used for relative mouse movements. (I think in the future we need to have a better way to handle this also)

ivan-mogilko:
I did not realize that `display_fps` is an engine variable exported from the engine. It's not present in config yet, but is present in the command line, so should be made a common config option.

ericoporto:
In this case then it's alright and can be addressed in a later PR.

ivan-mogilko:
Added audio_enabled (existing [sound][enabled], added show_fps to config ([misc][show_fps]).

The rest of the options seem to be either not handled by the engine, or not applicable at the time. They may be reimplemented later (such as sound cache, which had been removed during transition to SDL2).

ericoporto:
Hey, show fps in the command line used to use `kFPS_Forced` and now it's using 1. The difference is only `kFPS_Forced` allows to trigger the max fps mode. Everything else looks fine.

ivan-mogilko:
> Hey, show fps in the command line used to use kFPS_Forced and now it's using 1. The difference is only kFPS_Forced allows to trigger the max fps mode. 

The config option is a boolean. The internal display_fps variable takes different values, where kFPS_Forced is set after applying config, and kFPS_Display by a script command.

PR #1646, opened 2022-05-02
Support graphical Overlays not cloning the assigned sprite

ivan-mogilko:
I think this is the last addition to Overlays, complementing new features added in the last version.
There's a historical issue that Overlays make a full copy of a sprite when created, instead of saving a sprite number like all other objects do. While overlay total number was limited (to 20 afaik) this could be tolerated, but as this restriction is now removed, and overlays may be created in any numbers, this issue becomes quite noteable. It increases the memory usage, often unnecessarily, which becomes silly if a large number of overlays are created using same sprites.

This change adds support for storing a sprite's number in overlays instead of making a bitmap copy, when possible. By default it does so if overlay is assigned a "normal" sprite from the game's resources, as regular sprites cannot be edited or deleted.

To also cover the dynamic sprites, an extra `clone` parameter is added to Overlay.CreateGraphical which tells whether to clone the sprite for overlay's use, or use one from sprite set by the number as all other objects do.
* the default value of `clone` is false, to avoid unnecessary copying when users don't realize it's happening.
* old games will keep using old variant of this function, which translates into `clone = true`.
* regardless of the `clone` parameter, the engine internally still does not clone if the assigned sprite is a regular (non dynamic) one, because there's no point of doing that.

I believe that in the future this parameter may be removed, as well as image cloning itself. In my opinion this should not be an object's feature, but a part of the sprite management.

How to test this: create a large number of overlays using same sprite. In previous version the memory usage should increase roughly by (memory size of sprite x number of overlays). In this branch it should increase by (1 memory size of sprite + negligible amount for overlay structs).

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1274,7 +1274,11 @@ builtin managed struct InventoryItem {
 
 builtin managed struct Overlay {
   /// Creates an overlay that displays a sprite.
-  import static Overlay* CreateGraphical(int x, int y, int slot, bool transparent);  // $AUTOCOMPLETESTATICONLY$
+  import static Overlay* CreateGraphical(int x, int y, int slot, bool transparent = true
+#ifdef SCRIPT_API_v360
+	, bool clone = false
+#endif
+  );  // $AUTOCOMPLETESTATICONLY$
   /// Creates an overlay that displays some text.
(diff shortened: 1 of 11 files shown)


ivan-mogilko:
Alright, this seems working as expected; and the changes are minimal.

> How to test this: create a large number of overlays using same sprite. In previous version the memory usage should increase roughly by (memory size of sprite x number of overlays). In this branch it should increase by (1 memory size of sprite + negligible amount for overlay structs).

What remains in this sense is sharing *textures*, but that a) not related to the script api and b) requires implementing a general method in the engine.

PR #1648, opened 2022-05-03
Fix overlay script ref remains after overlay is removed

### LINKED ISSUE #1647: Strange case of overlays not getting recreated in Room_Load
**Describe the bug**
If there's an overlay created in "room before fade-in" event (aka "Room_Load") and saved in a room script's pointer variable, such overlay will be forcefully removed when changing the rooms, [in accordance to AGS game behavior](https://github.com/adventuregamestudio/ags/blob/1d321947974e7a94aafd2cafd0283af0b09c487d/Engine/ac/room.cpp#L296), but it won't get recreated when the player returns to the same room once more.

**AGS Version**
3.2.1 - 3.6.0

**To Reproduce**
1. Make a game with at least 2 rooms.
2. In room 1 create a "room before fade-in" event handler, and have following script:
```
Overlay *over;
function room_Load()
{
    over = Overlay.CreateGraphical(50, 50, 2000, true);
}
```
(the position and sprite number may be different)
3. Run the game, observe the created overlay.
4. Change to another room, then back to room 1.
5. Notice that overlay was not recreated.

**Expected behavior**
As there's a command to create overlay in room_Load, the overlay should be created an displayed on screen whenever player returns to this room.

ivan-mogilko:
Fixes #1647.

This seem to be years old oversight in the AGS engine.

Overlays are represented by a object in the engine (`ScreenOverlay`), but also by a simple "overlay reference" in the managed memory (`ScriptOverlay`) that allows to address them in script (this is similar to other game objects). ScriptOverlay is connected to the real thing using an id, and ScreenOverlay may save a "handle" to the script object.

When the overlay itself is removed, the script reference remains so long as it's stored in at least one variable. But because it needs to determine whether the real object is valid or not, it should be "invalidated" when that object is removed.

The problem is that while some overlay - script ref pairs are properly connected and engine does the invalidation part for them, others not. This seem to be a trivial oversight.
As a result, there could be a dangling script reference that still keeps some ID, and if by occasion there's a new overlay with matching ID, then all kinds of weird stuff may happen. In the above #1647 it led to immediate removal of the newly created overlay with same ID, because the pointer got overwritten and overlay reference disposed.

This PR fixes this by making sure that overlays created with script functions CreateGraphical and CreateTextual are properly connected to their script refs, and invalidate them when removed.
Additionally did some code refactor.


ericoporto:
I have an error: 
image
```AGS Script
// room script file

Overlay* ovr;

function room_AfterFadeIn()
{
  ovr = Overlay.CreateGraphical(64, 64, 2041, true);
  player.Say("Overlay should stay on screen");
  ovr = null;
  player.Say("Overlay should disappear from screen");
}
```

ivan-mogilko:
> Curiously the line the error occurs reference is wrong.

Tried to fix unnecessary line number with 538e9072b03b0441853ac0022d17f8990065f8de in master.

ivan-mogilko:
Possibly fixed the Say command.

ericoporto:
Say is fixed. The original bug reported is also fixed. I also could not cause problems playing with `Speech.TextOverlay`. Save appears to be working. A bit unsure of how to test more...

ivan-mogilko:
Tested all cases listed in the first post, found and fixed 1 bug that happened when Display message is removed.

PR #1651, opened 2022-05-03
Expand the sprite export options and add a quick sourceless sprite fixup command

### LINKED ISSUE #542: AGS 3: support recreating sprites that were imported from clipboard

ivan-mogilko:
Possibly resolves #542.

**Expanded sprite export dialog and function.**

I had to redesign the options a bit, changing their logical meaning from "Skip if ..." to "Export only if ...", because it was difficult to combine multiple options otherwise. So old "Skip sprites where source file exists" became "Export if source file is missing".

Added "Export if source file is outside the project folder" option.

The export function exports everything if none of the "export only if" flags is set. Otherwise it exports only those which match one of the conditions.

spriteexport-suggest1

~**Fixed the update source bug**~

~There has been a bug all this time: when the sprite export updates the source link, it does not add extension to the path, which makes it impossible to re-import from source (editor reports missing file). So fixed this.~
EDIT: this bug fix is now commited separately in 3.6.0 and 3.5.1

**The menu command for quick-fix sourceless sprites**

Added a new command right under "Export all...", called "Create source files for all sprites with missing / external sources..."
The name may be not perfect, but I could not come with anything better at this time.

What it does is simply runs export all with following options:
* Export if source is missing, or is outside of the project folder;
* Update source with the exported file.

I may remove it if you think it's not necessary.

### DIFF
--- a/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
+++ b/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
@@ -29,11 +29,12 @@
         private void InitializeComponent()
         {
             this.components = new System.ComponentModel.Container();
-            this.groupBoxExport = new System.Windows.Forms.GroupBox();
-            this.chkSkipValidSpriteSource = new System.Windows.Forms.CheckBox();
+            this.groupBoxExportFrom = new System.Windows.Forms.GroupBox();
             this.chkRecurse = new System.Windows.Forms.CheckBox();
             this.radRootFolder = new System.Windows.Forms.RadioButton();
             this.radThisFolder = new System.Windows.Forms.RadioButton();
+            this.chkIfSpriteSourceLocal = new System.Windows.Forms.CheckBox();
+            this.chkIfSpriteSourceValid = new System.Windows.Forms.CheckBox();
             this.groupBoxSaveAs = new System.Windows.Forms.GroupBox();
             this.btnBrowse = new System.Windows.Forms.Button();
             this.txtFolder = new System.Windows.Forms.TextBox();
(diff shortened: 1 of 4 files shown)


messengerbag:
Neat!

messengerbag:
1Neat

ivan-mogilko:
@morganwillcock do you have any opinion on this?

morganwillcock:
To me it feels a little confusing to have these options as flags and presented in this style because (to me) both are implying that the underlying selection is done with a logical 'and' but in the implementation it is treated as a logical 'or'.

These two states are mutually exclusive:
* source file is missing
* source file is outside the project folder

i.e. you cannot test whether something is outside of the project directory when it does not exist so (to me) it also does not make sense to have both checkboxes checked. I'd prefer it to be labelled as a filter rather than a selector, and the logic of the wording be flipped to indicate that the criteria is for excluding rather than including.

ivan-mogilko:
So, I changed this back to "Skip if".
spriteexport-suggest2

> i.e. you cannot test whether something is outside of the project directory when it does not exist so (to me) it also does not make sense to have both checkboxes checked. 

But it seems there's still a problem: it does not make sense to check both separately, because checking that the source is local is impossible without checking that it's valid.

messengerbag:
The objection doesn't really make sense to me either way around. To me, this is conceptually a question about the source path property stored in the project, regardless of whether it points to a file that exists or not. For example, `./Sprites/guybrush01.png` is inside the project folder whether or not that file can be found.

In fact, in theory you could have an absolute path pointing to a file inside the project folder, and I think it should still trigger this condition, because the important thing is whether it will still be valid if the whole project folder is moved (e.g. copied to another computer). Although in that case it should arguably just rewrite the property with the local version of the path.

messengerbag:
Oh and as for this:

> To me it feels a little confusing to have these options as flags and presented in this style because (to me) both are implying that the underlying selection is done with a logical 'and' but in the implementation it is treated as a logical 'or'.

To me, a dialog with "Do this if…" with multiple conditions that can be individually checkboxed implies a logical OR, so that it will be done in any case that matches any of the checked conditions.

ivan-mogilko:
> To me, a dialog with "Do this if…" with multiple conditions that can be individually checkboxed implies a logical OR, so that it will be done in any case that matches any of the checked conditions.

Hmm, i had a thought to propose simply renaming the group's name, but was not sure about the phrasing.

> In fact, in theory you could have an absolute path pointing to a file inside the project folder, and I think it should still trigger this condition, because the important thing is whether it will still be valid if the whole project folder is moved (e.g. copied to another computer). Although in that case it should arguably just rewrite the property with the local version of the path.

I believe editor may already do that when setting a source link.

ivan-mogilko:
> To me, this is conceptually a question about the source path property stored in the project, regardless of whether it points to a file that exists or not.

Does this mean that "skip if sprite source is relative" should not require source to exist? Should it only test for the path relation then and ignore missing sources too?

ivan-mogilko:
Well, I can do both ways, because technically both work the same, it's a matter of naming things and test for conditions correctly.

Personally, I like "Export if" approach more than "Skip if" because for me it seems more convenient to select what I'd _want_ to export rather than what I _don't want_ to export, but guess this may be a case of different perspective.

ivan-mogilko:
Okay, I am not sure what the consensus was, but the latest variant works, and if something is wrong then it is always possible to adjust, change wording etc.

PR #1652, opened 2022-05-04
Script API: remade Overlay.InRoom to Overlay.Layer, ObjectLayer enum

ivan-mogilko:
In script added ObjectLayer enum:
```
enum ObjectLayer
{
  eLayerRoom,
  eLayerUI
};
```

Remade recently added `bool Overlay.InRoom` to `ObjectLayer Overlay.Layer` property.

EDIT: There's also an option to use this layer as a parameter to CreateGraphical/CreateTextual instead of adding new functions, but this may make using them annoying. Also in case of CreateTextual it's difficult to add new parameters, as it's a variadic function with text formatting, so parameters cannot be added to the end of the argument list. If I add one, then all existing scripts with CreateTextual will have to be fixed.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -377,6 +377,12 @@ enum eKeyMod
   eKeyModNum        = 0x0040,
   eKeyModCaps       = 0x0080,
 };
+
+enum ObjectLayer
+{
+  eLayerRoom,
+  eLayerUI
+};
 #endif
 
 #ifdef SCRIPT_API_v3507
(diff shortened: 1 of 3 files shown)


ericoporto:
I thought about this a lot, I think you were right - sorry, I didn't thought you would have made a PR so soon. 

I like the usage of the word layer here, but perhaps the word `layer` will be useful for some concept in the future. I also like that, if the property is not only readable one day, assigning to it has more meaning with the enum language.

But in the end I believe the original API you made in #1639 is correct.

ivan-mogilko:
Hmm, alright. I guess there will be more consistent changes to ags4.

PR #1655, opened 2022-05-08
Minor tidy up

ericoporto:
really tiny code maintenance changes I picked up when I was experimenting with the codebase - nothing that changes how the engine behaves.


ivan-mogilko:
My only comment is regarding copying PfnWriteExtBlock; it's basically just a pointer, so copying is simple, but when you pass an argument as a reference or a pointer then there's an extra dereferencing takes place, so sometimes it's cheaper to pass as a copy.

PR #1657, opened 2022-05-10
Perfomance: implement short-term texture cache

ivan-mogilko:
Implements a short-term texture (DDB) cache in the graphics renderer(s).

**Has two purposes:**
* shares same texture data among multiple sprites on screen, somewhat reducing amount of texture recreations and memory requirements in case same sprite is used on a number of objects.
* potentially a foundation for implementing a long-term cache similar to the (bitmap) sprite cache in some future (may be seen as a practical experiment).

**Limits:**
* only short-term caching, the texture data is stored only so long as there's at least 1 object using this sprite.
* does not affect GUI and gui objects: since these may contain generated images, they would require additional changes to use this method. So, currently affects: characters, room objects and overlays.

**Implementation:**
* IDriverDependentBitmap (aka DDB) implementations have the texture data itself split out into a separate object. DDB will use shared_ptr to reference that object. DDB will still represent an individual texture entity on screen, with draw settings (transform, coloring, etc), while the TextureData parent class represents a shareable texture tile.
* Graphics renderer contains a map of TextureData objects, identified by an arbitrary uint32 key (renderer itself does not care what this id means). Engine can use new GetSharedDDB method: it generates a DDB which either wraps an existing texture data or generates a new texture, if one with a given id is not present.
* DestroyDDB now checks if the destroyed texture is a part of the map, and if there are no more active "users", then the data is removed from the cache.

**Testing:**

The best way to test this is to have a large number of objects with same sprite assigned to them, the higher the resolution the easier is to notice the effect. Using overlays is easiest as you may create them in script in thousands now.
Without this cache each object will have its own texture data allocated for it, with this PR there should be only one texture data per sprite number, + minimal amount of memory per object struct and DDB struct (in dozens of int32s).

---

**In regards to the long-term cache.**

I was thinking this over for a while, but came to realization that properly adding it would require bigger changes to the engine code, in several places. Just making one in the renderer might be almost trivial, but that won't be fully efficient and will cause number of issues. I'm not ready to work on that right now, but perhaps will be able to write a task for that for the future.

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -102,6 +102,8 @@ IDriverDependantBitmap *debugConsole = nullptr;
 // a renderer's texture and an optional position
 struct ObjTexture
 {
+    // Sprite ID
+    uint32_t SpriteID = UINT32_MAX;
     // Raw bitmap
     std::unique_ptr<Bitmap> Bmp;
     // Corresponding texture, created by renderer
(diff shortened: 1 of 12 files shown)


ericoporto:
Hey, so in I Rented a Boat I do the following, where I create a [`DynamicSprite and assign it's Graphic property to an object`](https://github.com/ericoporto/i_rented_a_boat/blob/main/I_rented_a_boat/room34.asc#L8496-L8499), then whenever the Dynamic Sprite updates, the object sprite is updated. But using the Engine from this PR, this doesn't work, the graphic doesn't update.

ivan-mogilko:
Oh right, I forgot to clear the cache item when the dynamic sprite is updated.

ericoporto:
hey, it still did not work for me with the game I added in the above comment. Did it work for you?

ivan-mogilko:
Ok this time it seems to work.

ericoporto:
This looks fine. Are there more things you would like to test? When you say it affects overlays, does it includes portraits?

ivan-mogilko:
> When you say it affects overlays, does it includes portraits?

Hm, probably not. There are several cases when the sprites are prepared in a non-efficient way, and portraits are one of these iirc. There has to be a good systematic work to resolve these, either going case by case, or remaking how these graphical objects are stored and updated in general.

PR #1661, opened 2022-05-12
Use BufferedStream for writing files + added stream tests

ivan-mogilko:
BufferedStream has been excluded for writing operations for a while, because there were some problems with it. This PR fixes writing ops within BufferedStream, and adds a number of autotests for streams. Tests are cruicial, as mistakes in writing may lead to greater problems (e.g. damage user files).

Speed tests show that writing with BufferedStream is about 50% faster than with the FileStream (tested by writing long byte-by-byte sequences).

There's an annoying issue of BufferedStream being derived from FileStream rather than let choose what to work on, this is a problem with current streams hierarchy; meanwhile memory-backed FILE functions such as `fmemopen` and `open_memstream`, as these are not in standard C. For simplicity the tests now write and read the file, but just in case I added a new compiler flag AGS_PLATFORM_TEST_FILE_IO that lets to skip the parts of code if you build for running from a read-only location (tbh not sure how useful that would be).

### DIFF
--- a/Common/core/platform.h
+++ b/Common/core/platform.h
@@ -133,7 +133,11 @@
 
 // Certain methods or classes may be meant for the code testing purposes only
 #if !defined(AGS_PLATFORM_TEST)
-    #define AGS_PLATFORM_TEST (AGS_PLATFORM_DEBUG)
+    #define AGS_PLATFORM_TEST (0)
+#endif
+// This option surrounds the test code which requires write disk access
+#if !defined(AGS_PLATFORM_TEST_FILE_IO)
+    #define AGS_PLATFORM_TEST_FILE_IO (AGS_PLATFORM_TEST)
 #endif
 
 #define AGS_HAS_DIRECT3D (AGS_PLATFORM_OS_WINDOWS)
--- a/Common/test/stream_test.cpp
+++ b/Common/test/stream_test.cpp
@@ -197,7 +197,18 @@ TEST(Stream, VectorStream) {
 
 static const char *DummyFile = "dummy.dat";
 
-TEST(Stream, BufferedStreamRead) {
+class FileBasedTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        File::DeleteFile(DummyFile);
+    }
+
+    void TearDown() override {
+        File::DeleteFile(DummyFile);
+    }
+};
+
+TEST_F(FileBasedTest, BufferedStreamRead) {
     //-------------------------------------------------------------------------
     // Write data into the temp file
     FileStream out(DummyFile, kFile_CreateAlways, kFile_Write);
@@ -264,7 +275,7 @@ TEST(Stream, BufferedStreamRead) {
     File::DeleteFile(DummyFile);
 }
 
-TEST(Stream, BufferedStreamWrite) {
+TEST_F(FileBasedTest, BufferedStreamWrite) {
     //-------------------------------------------------------------------------
     // Write data into the temp file
     // fill in to ensure buffered stream reach buffer size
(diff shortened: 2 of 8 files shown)


ivan-mogilko:
Hmm there's a strange problem about certain tools not building, could be AGS_PLATFORM_TEST is not defined correctly in there?

ericoporto:
I will look into the tests, soon but not right now. 

It looks like some undefined references. The tools and compiler have their own list of cpp files from `Common/` instead of using the same list as the Engine, they use a specific subset, so possibly some file needs to be added there.

ivan-mogilko:
> so possibly some file needs to be added there - in Tools/ and Compiler/ CMakeLists.txt.

the question is, why would they need it: nothing should have changed in them.

ericoporto:
Just to clarify, `AGS_PLATFORM_TEST` will always be true in the platforms currently failing since they have tests - we currently build the files and then build the tests executables and link to the built libraries, they are not separate things, it's the same compiler library that was built and used for the `agscc` and the `libtools` used on each tool. If the tests need to use different libraries built specifically for test, we need to change this. 

Is there any reason for not letting the tests write to disk? I don't understand why it could not write to disk.... 

But if you need that, then `COMPILER_COMMON_SOURCES` in the `Compiler/CMakeLists.txt` need to include `memorystream.cpp/h` additionally - and later we need to separate the libraries built for test and the ones in fact used, but this can be adjusted later. I am against though, it's better to test with the things that are used.

ivan-mogilko:
> Is there any reason for not letting the tests write to disk? I don't understand why it could not write to disk.... Note the CI is indifferent, we can write to disk there fine.

I'd prefer to use real FILE operations there, because it's of course preferrable to have as close functionality as possible. IIRC my concerns were that it may be a problem to clean things up after running auto tests: e.g. if test fails or something else goes wrong then some dummy file could be left on disk. I did not know if it's fine to write files on CI servers either.

ivan-mogilko:
> Is there any reason for not letting the tests write to disk? I don't understand why it could not write to disk.... Note the CI is indifferent, we can write to disk there fine.

I'm in doubts here; but I thought that unit tests should not write to disk, firstly because that makes them dependent on a filesystem, and secondly because it makes it harder to cleanup after a test.

ericoporto:
About cleanup you can use the Google test `TearDown` in a test fixture for cleanup if needed. I don't see a problem of the test writing to disk in this case though - I think it's better than having the class operate differently during tests. Perhaps an additional test could verify that the file operation is alright can be added if you want to make sure those are working.

ivan-mogilko:
> I don't see a problem of the test writing to disk in this case though - I think it's better than having the class operate differently during tests

Well, I do see a problem in depending on filesystem, but the concern of class differences is also valid, so guess that would be honest to test it as it is until the code is improved to let test certain operations separately.
Something that i'd like to try though is to introduce another flag which defines that the test program "depends on file writing" - the code under such flag cannot be run in read-only location.

> But if you need that, then COMPILER_COMMON_SOURCES in the Compiler/CMakeLists.txt need to include memorystream.cpp/h additionally - and later we need to separate the libraries built for test and the ones in fact used, but this can be adjusted later.

IMO the separation should be done regardless, because the current situation means that the tools are built with wrong parameters (AGS_PLATFORM_TEST is on in them). This has not been a problem until now by coincidence.

ericoporto:
hey, does al file writing in AGS now uses this? I can attest that things like saving the game still works.

ivan-mogilko:
> hey, does all file writing in AGS now uses this? I can attest that things like saving the game is working!

Most important is to ensure that compiling game and especially saving rooms works, as rooms are the only part of the project that is saved in binary in 3.6.0.

PR #1663, opened 2022-05-19
Add WaitInput in the Script API

ericoporto:
we are going to support more input types in the near future (Gamepad/Joystick/Touch Screen), at least for Gamepad we are probably going to need to some way of dealing with blocking actions like Speak that can be skipped. Instead of adding new `WaitGamepadMouseKey` functions, it would be interesting to have a more generic Wait function that could take additional parameters for the which inputs should be accepted when skipping this specific function.

I don't know yet how to make an `fInputAny` that would mean any and if when gamepad is added, I need it to include this in the same any or it's alright to then update the value of `fInputAny` on the script API.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -454,6 +454,16 @@ enum LogLevel
 };
 #endif
 
+#ifdef SCRIPT_API_v360
+enum InputType
+{
+	eInputNone     = 0x00000000,
+	eInputTime     = 0x00010000,
+	eInputKeyboard = 0x00020000,
+	eInputMouse    = 0x00040000,
+	eInputAny      = 0xFFFF0000
+};
+#endif
 
 internalstring autoptr builtin managed struct String {
   /// Creates a formatted string using the supplied parameters.
@@ -1489,6 +1499,8 @@ import int  WaitKey(int waitLoops = -1);
 /// Blocks the script for the specified number of game loops, unless a key is pressed or the mouse is clicked.
 import int  WaitMouseKey(int waitLoops = -1);
 #ifdef SCRIPT_API_v360
+/// Blocks the script for the specified number of game loops, unless a input is issued. Input are flags, and can be combined using bitwise operators.
+import int  WaitInput(InputFlag input_flag, int waitLoops = -1);
 /// Blocks the script for the specified number of game loops, unless the mouse is clicked.
 import int  WaitMouse(int waitLoops = -1);
 /// Cancels current Wait function, regardless of its type, if one was active at the moment.
--- a/Engine/ac/gamestate.cpp
+++ b/Engine/ac/gamestate.cpp
@@ -421,12 +421,7 @@ void GameState::SetWaitSkipResult(int how, int data)
 
 int GameState::GetWaitSkipResult() const
 {
-    switch (wait_skipped_by)
-    {
-    case SKIP_KEYPRESS: return wait_skipped_by_data;
-    case SKIP_MOUSECLICK: return -(wait_skipped_by_data + 1); // convert to 1-based code and negate
-    default: return 0;
-    }
+    return wait_skipped_by << 16 | (wait_skipped_by_data & 0x0000FFFF);
 }
 
 bool GameState::IsBlockingVoiceSpeech() const
--- a/Engine/ac/global_api.cpp
+++ b/Engine/ac/global_api.cpp
@@ -2215,6 +2215,12 @@ RuntimeScriptValue Sc_WaitMouseKey(const RuntimeScriptValue *params, int32_t par
     API_SCALL_INT_PINT(WaitMouseKey);
 }
 
+// int (int input_flags, int nloops)
+RuntimeScriptValue Sc_WaitInput(const RuntimeScriptValue *params, int32_t param_count)
+{
+    API_SCALL_INT_PINT2(WaitInput);
+}
+
 RuntimeScriptValue Sc_SkipWait(const RuntimeScriptValue *params, int32_t param_count)
 {
     API_SCALL_VOID(SkipWait);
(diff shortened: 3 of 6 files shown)


ivan-mogilko:
Could you explain a resoning behind adding "flag_" to the existing names? I don't think there other similar styled names in the engine right now. Do you suggest to adopt this new naming style where each flag is called with "flag_" prefix? Because then we'd also have to change others too.

Same with script api names, there are few enums with flags already, yet none have "f" in them, they all have "e" prefix.

> I don't know yet how to make an fInputAny that would mean any 

That would be simply 0xFFFFFFFF (UINT32_MAX value) for a int32 integer.

ericoporto:
> "any" or "all" flag value is done by having all the bits set. That would be 0xFFFFFFFF (UINT32_MAX value) for a int32 integer (or corresponding decimal value).

Thanks!

> Formally, according to the code style, the enums are called like kGroup_Name, perhaps for flags we could use fGroup_Name? Then this rule should be applied to all flag constants.

So I started doing this, but thought maybe `kf` worked better. Decided to stop right now before going forward, but added three separate refactors to showcase. 

Also some header have specific `defines` in their names, so when possible I moved out the enum declaration to outside of it, since it sound like a name specific for macros.

Anyway, still work in progress - and haven't yet correct the `FLAG_` prefix to enum `kf` as proposed...

ivan-mogilko:
> So I started doing this, but thought maybe kf worked better.

Why?

ericoporto:
> Why?

Since every enum is already preceded by `k`, renaming then to just `f` precedence felt like it reduced the meaning, f can be float, but k for enum, kf makes it related. I just felt weird when I refactored the already existing Overlay flags enum from `k` to `f`

> On bringing all the things to the chosen code style, this is perhaps better to do as a separate pr, as it's not related to the new functionality.

I will break this away then to a different PR.

ivan-mogilko:
One important part is missing: it's the return value that would allow input type extension. Right now it just returns previous Wait values.

I think that if we want to support extendable input types list there, the return value should be composed of type flags and "key" code value. If we keep value as integer, key codes could occupy first 16-bits (that's enough even for unicode chars, even though the ags keycodes are all below 500 anyway), while last 16-bits are left for flags.

Now, there are two options here, you may keep flags as they currently are (low bits) and shift them up, but that will require users to shift these down when analysing return value. That is extra work, difficult to understand for beginners, and also inconsistent with the function argument, where no shift is done. So, there's a second option: make these input type flags already have values in the upper 16-bits. That is, declare them as:
```
	eInputNone     = 0x00000000,
	eInputKeyboard = 0x00020000,
	eInputMouse    = 0x00040000,
	eInputAny      = 0xFFFFFFFF
```
both in script and internally. As far as I recall, these were not serialized anywhere, so no conversion is necessary for the old games.

What's remaning is to convert new WaitImpl return value to WaitMouse/Key return values. This is doable, but may not actually be necessary: this return values were implemented only for 3.6.0, and this is a rather obscure feature. So there's a chance they were not used in any game yet. Therefore all the Wait functions may return same values. _If_ later it turns out there has been already a game that used old variant, then the engine may be amended to return it checking the version of ags it was made with (or script api version).

ericoporto:
Hey, I like the flag approach and also think it's much easier to filter out the specific keycode or whatever is the data part when using bitwise operations, so I implemented this. If you think I should squash the commits, just tell me and I will do it, for now things are a separated.

ericoporto:
ok, found two problems

1. it doesn't work (!)
2. `key_skip_wait` which stores the skip type is an 8-bit only when serializing -> https://github.com/adventuregamestudio/ags/blob/8137918c209a479edf5127bb7c21d29470b8c40e/Engine/ac/gamestate.cpp#L584

ivan-mogilko:
> Edit: temporarily adjusted in the API level, but probably ideally would be to update the saveformat so we could use int and simplify this.

You could go opposite way and serialize shifted to 8-bit.

ivan-mogilko:
Actually, your temp fix might be a good solution in this case.

ericoporto:
ok, in this case I just reworded the commit message to remove the TEMP message. 

If you would like me to squash the commits or additional changes, just ask me. :)

ivan-mogilko:
There's one tiny mistake: the new script function is declared not under SCRIPT_API_v360, it needs to be moved just couple lines lower.

ericoporto:
> There's one tiny mistake: the new script function is declared not under SCRIPT_API_v360, it needs to be moved just couple lines lower.

Good catch! I adjusted and push a fix!

ericoporto:
Hey, I noticed now there's an enum called `SkipSpeechStyle` which has these 

```
enum SkipSpeechStyle {
#ifdef SCRIPT_API_v360
  eSkipNone         = -1,
#endif
  eSkipKeyMouseTime = 0,
  eSkipKeyTime      = 1,
  eSkipTime         = 2,
  eSkipKeyMouse     = 3,
  eSkipMouseTime    = 4,
  eSkipKey          = 5,
  eSkipMouse        = 6
};
```

should this new `InputFlag` enum have any relationship with this SkipSpeechStyle?

ivan-mogilko:
Yes, and there's also a separate enum for skipping cutscenes, and a separate one for skipping videos.

Ideally they all should be using the same one (flags).

If you change them now, there will have to be conversions for old games and loading data from old saves.

ivan-mogilko:
By the way, i just noticed that the enum is called `InputFlag`, would not `InputType` be better?

ericoporto:
> By the way, did not realize before that the enum is called InputFlag, would not InputType be better?

It could be, yeah. I can change.

I don't plan on looking into the other CutsceneSkipType, SkipSpeechStyle and VideoSkipStyle right now, but I think in the future I may look into simplifying these at the gamepad time. I don't want to hold this, as I said, it's mostly about changing the return of the wait stuff to use flag + data approach.

PR #1665, opened 2022-05-23
Editor: DialogEditor selects line when required

ericoporto:
fixes what was reported by Dave here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59842.msg636646209#msg636646209

I noticed this was different in Script Editor: https://github.com/adventuregamestudio/ags/blob/874fbbc7a729866dd8fa85c0cf1b0e2c75fce25c/Editor/AGS.Editor/Panes/ScriptEditor.cs#L515-L518

but also, it wasn't respecting what was asked in the `ZoomToFileEventArgs`, since `SelectLine` was true.

### DIFF
--- a/Editor/AGS.Editor/Components/DialogsComponent.cs
+++ b/Editor/AGS.Editor/Components/DialogsComponent.cs
@@ -283,7 +283,27 @@ namespace AGS.Editor.Components
             Dialog dialog = GetDialog(evArgs.FileName);
             if (dialog != null)
             {
-                ShowPaneForDialog(dialog).GoToScriptLine(evArgs);
+                DialogEditor dialogEditor = ShowPaneForDialog(dialog);
+
+                // has dialogEditor has already taken the area available?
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Had you chance to look into the window not scrolling to the selected position? I've been trying to reproduce that, and it happens sometimes. I'm not yet fully certain, but it seems this happens only the first time the dialog editor is opened by clicking on found item.

ericoporto:
Ended up using the Paint event to make sure the layout was already adjusted. I ended up using it through an elimination process, since other events I tried failed. Since the code is self contained there, I think it should be fine.

ivan-mogilko:
The latest commit seem to fix the reported problem, but it broke the "Find next" command in the dialog script: now it does not scroll the window nor highlights the text.

ericoporto:
ok, I switched so I only delay the find in the next paint event if the dialog editor is not yet shown on screen. This appears to fix the find next issue.

PR #1666, opened 2022-05-23
Restore support for sound streaming if file is big enough + sound cache for small sounds

ivan-mogilko:
After we changed to SDL2 and the new audio code, every sound clip is loaded into the memory at once now. The old code allowed to stream sound clips under some condition. In addition it had a sound cache for most frequently used sounds. This pr reimplements streaming within the new audio code, which is done if the sound asset is large enough, and a sound cache, this time coded in similarity with sprite cache (i believe in that in the future it might be possible to pick out shared code into a more universal resource cache class).

1. Engine decides whether to stream sound data from file, or load all at once, based on an asset size threshold.
2. For streaming, it passes opened ags stream into audio core, where the stream is wrapped into custom rwops for SDL sound. The reason why this is necessary is that assets are read from the package, and have to be limited by certain offsets. Unless i missed something, default rwops cannot do the same thing, which may result in the incorrect detection of the sound size.
3. Sound cache only stores sounds loaded at once; so only those smaller than the threshold setting. NOTE: in theory it's possible to code streaming into the shared cached buffer, but I decided to leave this for later (such streaming would also require remembering how much of the sound data has been read, and probably have a mutex, as same buffer may be read and written two from main and audio threads).
4. Added config options for "sound cache size" and "stream threshold". The latter may be renamed if the current name seem difficult to understand.

### DIFF
--- a/Engine/media/audio/sound.cpp
+++ b/Engine/media/audio/sound.cpp
@@ -53,21 +53,33 @@ static int GuessSoundTypeFromExt(const String &extension)
     return 0;
 }
 
+// Maximal sound asset size which is allowed to be loaded at once;
+// anything larger will be streamed
+// TODO: make configureable?
+static const size_t MaxLoadAtOnce = 1024u * 1024;
+
 static SOUNDCLIP *my_load_clip(const AssetPath &apath, const char *extension_hint, bool loop)
 {
-    auto *s = AssetMgr->OpenAsset(apath);
+    std::unique_ptr<Stream> s(AssetMgr->OpenAsset(apath));
     if (!s)
         return nullptr;
-
-    const size_t asset_size = static_cast<size_t>(s->GetLength());
-    std::vector<uint8_t> data(asset_size);
-    s->Read(data.data(), asset_size);
-    delete s;
 
     const auto asset_ext = AGS::Common::Path::GetFileExtension(apath.Name);
     const auto ext_hint = asset_ext.IsEmpty() ? String(extension_hint) : asset_ext;
 
-    auto slot = audio_core_slot_init(data, ext_hint, loop);
+    int slot{};
+    const size_t asset_size = static_cast<size_t>(s->GetLength());
+    if (asset_size > MaxLoadAtOnce)
+    {
+        slot = audio_core_slot_init(std::move(s), ext_hint, loop);
+    }
+    else
+    {
+        std::vector<uint8_t> data(asset_size);
+        s->Read(data.data(), asset_size);
+        slot = audio_core_slot_init(data, ext_hint, loop);
+    }
+
     if (slot < 0) { return nullptr; }
 
     const auto sound_type = GuessSoundTypeFromExt(ext_hint);
--- a/Engine/media/audio/sound.cpp
+++ b/Engine/media/audio/sound.cpp
@@ -13,11 +13,14 @@
 //=============================================================================
 #include "media/audio/sound.h"
 #include <cmath>
+#include <list>
+#include <unordered_map>
 #include "core/assetmanager.h"
 #include "media/audio/audio_core.h"
 #include "media/audio/audiodefines.h"
 #include "util/path.h"
 #include "util/stream.h"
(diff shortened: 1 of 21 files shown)


Engine/media/audio/sound.cpp
```diff
@@ -13,11 +13,13 @@
 //=============================================================================
 #include "media/audio/sound.h"
 #include <cmath>
+#include <unordered_map>
```
ericoporto, line 16: it needs `#include <list>` too

Engine/media/audio/sound.cpp
```diff
@@ -53,31 +56,153 @@ static int GuessSoundTypeFromExt(const String &extension)
...
+        const auto id = *it;
+        {
+            _cacheSize -= _map[id].Data->size();
+            _map[id].Data.reset();
+        }
+        // Remove from the mru list
+        _mru.erase(it);
+    }
+    // Keep disposing oldest elements until cache size is reduced to the given threshold
+    void FreeMem(size_t threshold)
```
ericoporto, line 124: threshold is never used. Maybe `_cacheSize + threshold >= _maxSize` was supposed to be used later?

Engine/media/audio/sound.cpp
```diff
@@ -53,31 +56,153 @@ static int GuessSoundTypeFromExt(const String &extension)
...
+        assert(ref);
+        if (!ref)
+            return; // safety precaution
+        if (Get(name))
+            return; // already in cache
+        if (_maxSize == 0)
+            return; // cache is disabled
+        // Clear up space before adding
+        if (_cacheSize + ref->size() > _maxSize)
+            FreeMem(_maxSize);
```
ericoporto, line 94: This reads like it Frees all cache. But FreeMem doesn't take an amount, it takes a threshold, so perhaps it doesn't. I am finding this line hard to read.

Engine/media/audio/sound.cpp
```diff
@@ -53,31 +56,153 @@ static int GuessSoundTypeFromExt(const String &extension)
...
+    }
+
+    struct SoundEntry
+    {
+        String Name;
+        DataRef Data;
+        // MRU list reference
+        std::list<String>::const_iterator MruIt;
+        SoundEntry() = default;
+        SoundEntry(const String &name, DataRef data)
```
ericoporto, line 139: Since you are using std::move in other places, you can just use here in this case too:
```
SoundEntry(String name, DataRef data)
    : Name(std::move(name)), Data(std::move(data)) {}
```

ivan-mogilko: Can I? DataRef is a shared_ptr.

Engine/media/audio/sound.cpp
ivan-mogilko: This function may be also used to resize the cache to certain limit in case cache size setting changes. But indeed the argument should take ref->size() into account to, so it should be `_maxSize - ref->size()`.

I may rename FreeMem to something else to emphasize that it frees not for the exact amount but to keep at certain limit.

Engine/media/audio/sound.cpp
ivan-mogilko: I don't know if std::move is important here; I could pass DataRef also as a `const &` and then it won't make unnecessary reference increment and decrement.

ivan-mogilko:
Fixed couple of mistakes, and also changed how FreeMem works to taking `space` argument instead, meaning how much free space is required.

Engine/media/audio/sound.cpp
ivan-mogilko: I changed this to `SoundEntry(const String &name, DataRef &data)` for consistency to prevent extra reference increment/decrement. Perhaps it could be revised later, to set a consistent code style throughout the program.

ericoporto:
I verified it's working. :) I like the change to `FreeMem` you did. It works, maybe in the future adding a new line in Ctrl+V debug menu to show sound cache usage or some debug command, but right now it's alright.

Issue #1674, opened 2022-06-05
AGS 4: New compiler does not correctly parse integer literals at the negative limit


ivan-mogilko:
@fernewelten 

The new compiler has some general mistake in the [`Scanner::ReadInNumberLit`](https://github.com/adventuregamestudio/ags/blob/4a1452a4700535ca7c41c666e6238e46d566ccfa/Compiler/script2/cs_scanner.cpp#L310-L323), where it would incorrectly determine the int32 value range. Possibly this is because it parses sign separately from the numeric value itself, so when testing for overflow it uses the positive value instead.

As a result a valid int32 value `-2147483648` could not be parsed correctly. The code
```
int a = -2147483648;
```
would result in a compiler's error:
`Literal integer '2147483648' is out of bounds (maximum is '2147483647')`
(Notice how it drops the sign in the error message too, and clearly only considers the positive value limit).

This seem to be a know issue, as all the new compiler's tests that test for integer limits use *not* INT32_MIN value, but INT32_MIN + 1 value with comments. For example: https://github.com/adventuregamestudio/ags/blob/4a1452a4700535ca7c41c666e6238e46d566ccfa/Compiler/test2/cc_parser_test_0.cpp#L142
At the same time the scanner test considers valid INT32_MIN to be parsed with error: https://github.com/adventuregamestudio/ags/blob/4a1452a4700535ca7c41c666e6238e46d566ccfa/Compiler/test2/cc_scanner_test.cpp#L322-L329

Another consequence of this mistake in scanner is that it could not correctly parse the large hexadecimal numbers, e.g. `0xFFFFFFFF`, thinking that they exceed the integer limit.

Please note that @AlanDrake made a fix for this latter problem recently (53538101712f81359a4b4162e1142bb4f0c875bb), as it was necessary to let compile the recent additions to the script API (the built-in header). But this fix is not working well, as it created an opposite problem. Now the compiler cannot detect the value overflow correctly. The code
```
    int a = -2147483648;
    int b = -2147483649;
    Display("%d, %d", a, b);
```
will compile, but result in displaying numbers -2147483648 and 2147483647, as internally the variables are still int32.

ivan-mogilko:
Please note that I opened a PR with adjusted tests for this: #1675

fernewelten:
[...]

PR #1675, opened 2022-06-05
AGS 4: new compiler: fixed int32 limit tests

ivan-mogilko:
There has been some kind of a mistake in the new compiler's Scanner class that resulted in the incorrect parsing and/or testing of int32 negative limit (see #1674). Unfortunately, the tests were adjusted to comply for this incorrect behavior, which is quite a wrong thing to do, in my opinion. A recent commit (53538101712f81359a4b4162e1142bb4f0c875bb) fixed one problem in practice, but introduced new ones, and some tests no longer work.

In order to introduce some clarity to this, I suggest correcting the tests first. Note that "Scan.LiteralIntOverflow" test does not work correctly after (53538101712f81359a4b4162e1142bb4f0c875bb), while "Scan.LiteralIntLimits" won't work if that commit would be reverted.

In addition this adds hexadecimal test for the scanner.

### DIFF
--- a/Compiler/test2/cc_parser_test_0.cpp
+++ b/Compiler/test2/cc_parser_test_0.cpp
@@ -138,13 +138,11 @@ TEST_F(Compile0, ParsingIntSuccess) {
 
 TEST_F(Compile0, ParsingIntLimits) {    
 
-    // Note, 2147483648 will result in an overflow found by the scanner
-
     const char *inpl = "\
-        import int int_limits(int param_min = -2147483647, int param_max = 2147483647); \n\
+        import int int_limits(int param_min = -2147483648, int param_max = 2147483647); \n\
         int int_limits(int param_min, int param_max)    \n\
         {                                               \n\
-            int var_min = - 2147483647;                 \n\
+            int var_min = -2147483648;                  \n\
             int var_max = 2147483647;                   \n\
         }\
         ";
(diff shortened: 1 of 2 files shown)


AlanDrake:
I'd suggest to also add test for hex and octal literals.

ericoporto:
@AlanDrake wait, do we support octals? How do I write it?

fernewelten:
@ericoporto Yes, we can do octals in the new compiler. In C style, with a leading `0`. `012345670` is 2739128. Or hex in C style, `0x1234ABCD` is 305441741.

ericoporto:
With a naked 0 prefix? I am sorry, but this is a terrible idea. If I type 015 I expect it to be the same as 15...

AlanDrake:
@ericoporto Octals have a zero prefix in pretty much all the major C-like languages. Even Java and javascript.

ericoporto:
I fully expect this to silently break a bunch of scripts for 0 gain. I think it's terrible design.

fernewelten:
Well, fight it out amongst you and I'll implement what the consensus will turn out to be.

ivan-mogilko:
Hm, I sort of understand @ericoporto 's concern about this. Octals are way too obscure, and i ever only saw them in use when passing file permissions in C code. Cannot really imagine why would I use them in AGS script. So I think the more explicit their syntax is - the better, to help avoid occasional mistakes.

I think this should be a separate change, and perhaps a ticket. I would delay adding tests until the syntax question is resolved.

> I'd suggest to also add tests for hex

I added hex test for scanner, but may add one for parser too.

ivan-mogilko:
Merged the tests, but they will keep failing at least until #1674 is resolved.

fernewelten:
> Merged the tests, but they will keep failing at least until #1674 is resolved.

Am working on it. Sorry for the delay, I had been stalled in a game jam.

Issue #1681, opened 2022-06-07
Incorrect Blank Lines when saving Dynamic Sprite


ericoporto:
**Describe the bug**
When saving a dynamic sprite, a white line or area appears mysteriously in the dynamic sprite.

**AGS Version**
Current master (c7bd448dfffc197c4398a38395b444457c958d24).

**Game**
[TestSpriteSave.zip](https://github.com/adventuregamestudio/ags/files/8857205/TestSpriteSave.zip)

**To Reproduce**
1. Create a new game with the sierra template
2. Place following code in the room
```AGS Script
// room script file

Overlay* ovr;
DynamicSprite* dynspr;

function room_AfterFadeIn()
{ 
  dynspr = DynamicSprite.Create(320, 80, true);
  DrawingSurface* surf = dynspr.GetDrawingSurface();
 
  surf.DrawImage(20, 20, 2168);
  surf.DrawImage(0, 0, 2042);
  surf.DrawImage(40, 40, 2041);
  
  surf.Release();
  
  dynspr.SaveToFile("$SAVEGAMEDIR$/save_dynspr.bmp");
  
  ovr.CreateGraphical(40, 40, dynspr.Graphic, false);  
}
```
3. Now go into your saved games dir an see the bitmap is defective and has a white area in it.

Alternatively, run the attached game and look into the savegame dir.

**Expected behavior**
The dynamic sprite should be full pink with or alpha with a blue cup, a key and a blueish rounded rectangle, **but no white areas**!!

**Screenshots**
![image](https://user-images.githubusercontent.com/2244442/172502484-348ab5ed-396a-4286-b465-91a620c1166e.png)

The white line doesn't exist in script

**Desktop:**
 - OS: Windows
 - Version 10

**Additional context**
Reported in the forums here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59842.msg636646806#msg636646806

ivan-mogilko:
This is an old bug, supposedly since after the switch to SDL2 backend. I now suspect something was broken in allegro4 after its code got stripped.

ivan-mogilko:
Pushed a fix: 623db88c135304bc3c462edffb742633ec50d608

There was a mistake in allegro's packfile implementation, it did not initialize or ended writing correctly.
I guess that ideally we should change allegro image saving and loading to just use our streams instead, but this is a hotfix for now.

ericoporto:
Thanks!

[...]

PR #1682, opened 2022-06-08
Add bitmap streamer

ericoporto:
Following discussion from #1681

The idea is to create a method to read `.bmp` images from Stream (and store in bitmap objects), and write a `.bmp` image to a Stream (based from a bitmap object).

This is because current allegro methods in use work in either File or PackFile, but not the in use AGS Streams.

- [x] method to write `.bmp` to a Stream from a bitmap object
- [x] method to read a `.bmp` from a Stream to a bitmap object
- ~~method to convert a bmp color depth to the game color depth~~
- [x] color conversion in read `.bmp` itself
- make sure game palette is not overridden by the palette on bmp file. Not sure how to do this?

We can also decide we don't need this right now, since the packfile bmp writing bug has been fixed (#1681), since it's really close for 3.6.0 to be stable.

### DIFF
--- a/Common/gfx/allegrobitmap.cpp
+++ b/Common/gfx/allegrobitmap.cpp
@@ -12,10 +12,12 @@
 //
 //=============================================================================
 
+#include <memory>
 #include <stdexcept>
 #include <string.h> // memcpy
 #include <aastr.h>
 #include "gfx/allegrobitmap.h"
+#include "util/filestream.h"
 #include "debug/assert.h"
 
 extern void __my_setcolor(int *ctset, int newcol, int wantColDep);
@@ -123,6 +125,12 @@ bool Bitmap::WrapAllegroBitmap(BITMAP *al_bmp, bool shared_data)
     return _alBitmap != nullptr;
 }
 
+void Bitmap::ForgetAllegroBitmap()
+{
+    _alBitmap = nullptr;
+    _isDataOwner = false;
+}
+
 void Bitmap::Destroy()
 {
     if (_isDataOwner && _alBitmap)
@@ -133,35 +141,9 @@ void Bitmap::Destroy()
     _isDataOwner = false;
 }
 
-bool Bitmap::LoadFromFile(const char *filename)
-{
-    Destroy();
-
-	BITMAP *al_bmp = load_bitmap(filename, nullptr);
-	if (al_bmp)
-	{
-		_alBitmap = al_bmp;
-        _isDataOwner = true;
-	}
-	return _alBitmap != nullptr;
-}
-
-bool Bitmap::LoadFromFile(PACKFILE *pf)
-{
-    Destroy();
-
-    BITMAP *al_bmp = load_bmp_pf(pf, nullptr);
-    if (al_bmp)
-    {
-        _alBitmap = al_bmp;
-        _isDataOwner = true;
-    }
-    return _alBitmap != nullptr;
-}
-
 bool Bitmap::SaveToFile(const char *filename, const void *palette)
 {
-	return save_bitmap(filename, _alBitmap, (const RGB*)palette) == 0;
+	return BitmapHelper::SaveToFile(this, filename, (const RGB*) palette);
 }
 
 color_t Bitmap::GetCompatibleColor(color_t color)
--- a/Common/gfx/bitmap.cpp
+++ b/Common/gfx/bitmap.cpp
@@ -12,8 +12,14 @@
 //
 //=============================================================================
 
+#include <array>
+#include <memory>
 #include "gfx/bitmap.h"
 #include "util/memory.h"
+#include "util/memorystream.h"
+#include "util/file.h"
+#include "util/path.h"
+#include "util/string_utils.h"
 
 namespace AGS
 {
@@ -82,24 +88,17 @@ Bitmap *CreateBitmapCopy(Bitmap *src, int color_depth)
 
 Bitmap *LoadFromFile(const char *filename)
 {
-	Bitmap *bitmap = new Bitmap();
-	if (!bitmap->LoadFromFile(filename))
-	{
-		delete bitmap;
-		bitmap = nullptr;
-	}
-	return bitmap;
-}
+    std::unique_ptr<Stream> in (
+            File::OpenFile(filename, FileOpenMode::kFile_Open, FileWorkMode::kFile_Read));
 
-Bitmap *LoadFromFile(PACKFILE *pf)
-{
-    Bitmap *bitmap = new Bitmap();
-    if (!bitmap->LoadFromFile(pf))
-    {
-        delete bitmap;
-        bitmap = nullptr;
-    }
-    return bitmap;
+    RGB in_palette[256];
+
+    if(in == nullptr) return nullptr;
+
+    String ext = Path::GetFileExtension(filename).Lower();
+    Bitmap* bmp = BitmapHelper::LoadBitmap(ext, in.get(), in_palette);
+
+    return bmp;
 }
 
 Bitmap *AdjustBitmapSize(Bitmap *src, int width, int height)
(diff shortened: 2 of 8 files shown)


ivan-mogilko:
My point was that you could add these functions into BitmapHelper namespace, where LoadBitmap already exist, instead of creating a separate code file. Bitmap's member functions SaveToFile and LoadFromFile may also be removed.

ericoporto:
> I mean, there's already a [BitmapHelper namespace](https://github.com/adventuregamestudio/ags/blob/0fd748eada90e1f07a8cb47430703744bc4613b1/Common/gfx/bitmap.h#L70-L72), where LoadBitmap entry already exists, so you could put these functions there instead of creating a separate code file, replacing existing functions.

I placed the functions there in BitmapHelper, but haven't finished the function for loading bitmaps yet!

> member functions SaveToFile and LoadFromFile may be removed from the Bitmap class altogether, and BitmapHelper functions used simply on their own instead to do the loading/saving

Right, this would mean in DynamicSprite and SaveScreenShot, the functions from BitmapHelper would be used directly instead of using the methods in the Bitmap class. I think these two are the only places that we write to a bitmap in the engine, need to check if there's any case in other places - agsnative or a tool.

ericoporto:
So, **reading the bitmap file** I am a bit stuck on something... When the bitmap is read, turns out there's a depth conversion that happens, and this is dependent on a global variable. In AGS we occasionally use `set_color_depth(int bpp)` to change this globally. Now, from what I can tell, reading full blown bitmaps always happen when bpp is `game.GetColorDepth()`. I have three ways to go here:

1. use internally `game.GetColorDepth()` or even full allegro `get_color_depth()` to get game color depth
2. have `target_bpp` as a parameter that the function can receive
3. always assume the bmp file has the same bpp as the game, and thus do not do this bpp conversion

For backwards compatibility, I am tending for 2, and externalizing the method mentioned in 1 to retrieve the game color depth.

ivan-mogilko:
There are two variants that seem reasonable to me here, first is p2 mentioned in your comment, and another is read from file as-is, and then convert externally if needed.

ericoporto:
> read from file as-is, and then convert the returned bitmap if needed (outside of the function)

This is probably a nice way to approach this, but the original allegro4 version was done using an internal function. From what I can understand, it's mostly done so when an indexed bmp is read, the palette is applied and we get a 32bpp image and then adjust to whatever is the bpp of allegro itself (or in this case, the game).

I also have a feeling the overall matching of bpps was done for the sake of the software renderer.

Anyway, about reading the bitmaps, I found two approaches:

- one is the one I did, where I extracted the allegro4 original methods and replaced where need for stream functions. I think we already were able to use these through some hoops by using the packfiles.
- The other way I could think would be using some library like stb_image, which has places to plug something like our streams. Though I had a hard time figuring out if the results would be similar to what we get from allegro4 bmp reading, since specially in the case of 8-bit images where allegro4 really expects a unified palette for the game.

ivan-mogilko:
> I also have a feeling the overall matching of bpps was done for the sake of the software renderer.

BTW, this may not be wanted in all cases. For instance, a person might want to have a 8-bit image in their game to be used as a mask (especially since we have a way to draw upon room masks now). Unfortunately, current editor logic is to convert everything to the game bpp on import, and engine seem to do the same when creating or loading sprites. Perhaps it may be in todo for the future to add extra options there.

> Though I had a hard time figuring out if the results would be similar to what we get from allegro4 bmp reading, since specially in the case of 8-bit images where allegro4 really expects a unified palette for the game.

It's always possible to apply any changes after the data was loaded, if it has to match something.

ericoporto:
ok, just need to provide some way to implement the functionality from both `_color_load_depth` and `_fixup_loaded_bitmap` which are internal allegro methods, using parameterized bpp - instead of using a global one. Then I will be able to do the depth conversion.

> PS. C/C++ library like stb_image may be come useful in later development 

Yeah, once I can figure how to mimic the bpp conversion, I can split this in it's own method, so in the future if one wants to replace with stb_image methods, at least this will be adjusted. For now I think repurposing allegro4 methods is best in terms of getting it working and keeping it compatible.

ericoporto:
Found something that looks like what I want but doesn't actually do what one think it does (and it's in the Engine instead of Common)
https://github.com/adventuregamestudio/ags/blob/e1777f8c3bb92a668a10833ce6351d8c2fad58ad/Engine/gfx/gfx_util.cpp#L29-L42
The blit strategy IS what allegro does, but one also needs to fix up the palette
https://github.com/adventuregamestudio/ags/blob/e1777f8c3bb92a668a10833ce6351d8c2fad58ad/libsrc/allegro/src/readbmp.c#L116-L125

ericoporto:
This should be working, but I could not figure out the last steps of refactoring to move the color conversion to outside. Mostly, allegro4 functions are moved to BitmapHelper, but these could be moved elsewhere if needed. The write and read functions though are redone to use AGS Stream methods.

Common/gfx/allegrobitmap.cpp
```diff
@@ -137,31 +139,18 @@ bool Bitmap::LoadFromFile(const char *filename)
...
-    Destroy();
+    std::unique_ptr<Bitmap> bmp (BitmapHelper::LoadFromFile(filename));
 
-    BITMAP *al_bmp = load_bmp_pf(pf, nullptr);
-    if (al_bmp)
+    if (bmp)
     {
-        _alBitmap = al_bmp;
-        _isDataOwner = true;
+        this->CreateCopy(bmp.get());
```
ivan-mogilko, line 146: This makes an completely unnecessary copy operation of the bitmap.

My earlier proposal was to remove Bitmap members LoadFromFile and SaveToFile completely, as they are no longer needed.

Optionally, you could separate a function that loads an array of pixels, and let create a Bitmap object from that array, knowing parameters.

Common/gfx/bitmap.h
```diff
@@ -77,6 +77,12 @@ namespace BitmapHelper
     // Pitch is given in bytes and defines the length of the source scan line.
     // Offset is optional and defines horizontal offset, in pixels.
     void    ReadPixelsFromMemory(Bitmap *dst, const uint8_t *src_buffer, const size_t src_pitch, const size_t src_px_offset = 0);
+
+    // Write to the stream a bitmap, optionally along with the palette
+    void SaveBitmapStream(Stream *out, const Bitmap *bmp, const RGB* pal = nullptr);
```
ivan-mogilko, line 82: I'd propose to call the functions simply LoadBitmap and SaveBitmap, having them differentiate with parameter types (if loading from file should still be kept).

Common/gfx/allegrobitmap.cpp
ericoporto: erh, we never use this method, so I simply deleted it.

Common/gfx/bitmap.h
ericoporto: renamed!

ericoporto:
please test this before merging it. I haven't yet deleted the original allegro bmp reading and saving methods in case we find a problem later and need to revert this.

Common/gfx/bitmap.cpp
```diff
@@ -221,6 +214,738 @@ void ReadPixelsFromMemory(Bitmap *dst, const uint8_t *src_buffer, const size_t s
...
+        /* restore original palette except if it comes from the bitmap */
+        if ((bpp != 8) && (!want_palette))
+            pal = NULL;
+
+        if (bmp) {
+            Bitmap* tmp_bmp = BitmapHelper::CreateRawBitmapOwner(
+                    _fixup_loaded_bitmap(bmp->GetAllegroBitmap(), pal, dest_depth));
+            Bitmap* tmp_bmp2 = BitmapHelper::CreateBitmapCopy(tmp_bmp);
+
+            bmp->Destroy();
```
ivan-mogilko, line 926: Calling Destroy here is not enough, as this method only deletes the internal data, but not the `bmp` object itself. You need to do `delete bmp` in order to completely deallocate Bitmap object.

Common/gfx/bitmap.cpp
```diff
@@ -221,6 +214,738 @@ void ReadPixelsFromMemory(Bitmap *dst, const uint8_t *src_buffer, const size_t s
...
+
+    if (dest_depth != bpp) {
+        /* restore original palette except if it comes from the bitmap */
+        if ((bpp != 8) && (!want_palette))
+            pal = NULL;
+
+        if (bmp) {
+            Bitmap* tmp_bmp = BitmapHelper::CreateRawBitmapOwner(
+                    _fixup_loaded_bitmap(bmp->GetAllegroBitmap(), pal, dest_depth));
+            Bitmap* tmp_bmp2 = BitmapHelper::CreateBitmapCopy(tmp_bmp);
```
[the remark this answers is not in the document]
ericoporto: you are right, `_fixup_loaded_bitmap` returns a new bitmap, removed this addition

Common/gfx/bitmap.cpp
ericoporto: good catch! I adjusted this too.

ericoporto: thanks! Fixes added, [comparison here](https://github.com/adventuregamestudio/ags/compare/c16711bae808d5145c8561d98e3ea35a835fd0fd..42b9759d880f15bc4103ed9e6d62915c51f70916)

ivan-mogilko:
I think this works in general. I will make few extra tests with different color depths in a short while.

ivan-mogilko:
> We can also decide we don't need this right now, since the packfile bmp writing bug has been fixed (https://github.com/adventuregamestudio/ags/issues/1681), since it's really close for 3.6.0 to be stable.

Actually, I'm asking this question to myself now.

This change is not enough to get rid of PACKFILE, because there's also `load_pcx` function, so one would also have to reimplement that with ags streams. And there's also APEG library that still uses PACKFILE to stream video.
So even if we merge this, we will still have to keep PACKFILE wrapper over ags streams.

In the following versions we might use another library for loading images which would support wider variety of formats (also necessary for standalone tools), which means there will be another change to this code.

ivan-mogilko:
> We can also decide we don't need this right now, since the packfile bmp writing bug has been fixed > (https://github.com/adventuregamestudio/ags/issues/1681), since it's really close for 3.6.0 to be stable.

I'm sorry that I mention this so late, but I'm actually asking this question to myself now.

This change is not enough to get rid of PACKFILE, because there's also load_pcx function, so one would also have to reimplement that with ags streams. And there's also APEG library that still uses PACKFILE to stream video.
So even if we merge this, we will still have to keep PACKFILE wrapper over ags streams.

In the following versions we might use another library for loading images which would support wider variety of formats (also necessary for standalone tools), which means there will be another change to this code.

At the same time, like you said, 3.6.0 is almost out, and making big code changes for no serious reason is not a very good idea.

I may merge this, but this won't solve any actual problem at the moment.

ivan-mogilko:
> We can also decide we don't need this right now, since the packfile bmp writing bug has been fixed > (https://github.com/adventuregamestudio/ags/issues/1681), since it's really close for 3.6.0 to be stable.

I'm sorry that I mention this so late, but I'm actually asking this question to myself now.

This change is not enough to get rid of `PACKFILE`, because there's also `load_pcx` function, so one would also have to reimplement that with ags streams. And there's also APEG library that still uses `PACKFILE` to stream video.
So even if we merge this, we will still have to keep `PACKFILE` wrapper over ags streams.

I found there's a lost behavior with the bitmap loading and saving now: previously allegro's function `load_bitmap` chose which format to save depending on a file extension. For instance, currently it supports bmp and pcx. With the given change this is no longer going to work. Same happened with `save_bitmap`.

In the following version(s) we might use another library (i think you mentioned stb_image) for loading images which would support wider variety of formats (also necessary for standalone tools), which means that most of this new implementation will be removed.

At the same time, like you said, 3.6.0 is almost out, and making big code changes for no serious reason is not a very good idea.

ivan-mogilko:
Ah... I did not think about this at first, but this change currently breaks the file format selection that Allegro 4 was doing. If you look into its `load_bitmap` and `save_bitmap` functions, they select one of the supported image formats to load or save. I think currently it supports `bmp` or `pcx`. So `load_pcx` and `save_pcx` should also be reimplemented in similar way to fully recreate this behavior.

Also, to be honest, I miscalculated the size of the bmp loading code, when suggesting to place the whole new code right into the "bitmap helper" and bitmap.cpp. Now when I see how big it is, and considering that pcx support is also required, perhaps the resulting code that does format reading and writing itself could be placed in a separate file to make things cleaner. I might suggest "bitmap_file.cpp" or "image_file.cpp", but anything similar will work.

ivan-mogilko:
> The thing about stb_image or other library is, for bmps, the particular way it handles it, is different. So, backwards compatibility and stuff.

Unfortunately, I never had a chance to check how stb_image works, so cannot talk about details. But, like I mentioned previously, any loaded result is an array of pixels and may be adjusted and converted further if necessary. Things like converting 8-bit image to 32-bit image in a 32-bit game is possible to do on top of the image loading code (or library). This also may not always be wanted, if we would like to support e.g. using 8-bit masks in any game, so may be done as an option.

In overall, this requires some planning, to establish a) what is the desired behavior of loading an image to the game, b) what does the loading code / library does, c) what additional processing should our code do after the image is loaded.

> I think I would merge this after 3.6.0 and refactor the color depth conversion out of it.

> About apeg, I just checked it and it looks doable rewrite it similarly - if we slash out the audio parts since we have the decoder and stuff running in AGS, the remaining video stuff is not that much

Possibly, but, that is not necessarily a good way to follow. I definitely made a mistake with that remark about replacing everything with ags stream, it was not thought through very well. Because every "loading" or "saving" library which we use or going to use will have its own reader interface anyway. Like, SDL_Sound uses SDL RWops, [which we customize to use our streams](https://github.com/adventuregamestudio/ags/blob/0af788ef8a35580fde95eead30d48e7fbc30884e/Engine/util/sdl2_util.h#L32). So we will have to use their interfaces anyway.

I might suggest to make a full experiment with existing image loading libs first and see how it will work if you use them with ags, because if it's possible to make one work with some additions on top (color conversion, and the like), then it's better to keep one library for all the file formats.

> I think these bmp format specific would work better in a separate bmp format loading class, like the original bitmap streamer.

Yes, sorry, I miscalculated the size of the code which loads bitmap, it definitely should have been a part of its own module.

ericoporto:
Added pcx support

ivan-mogilko:
>  I need to figure out a good test game that really verifies this works, and works similar to the previous implementation.

Some games used preload.pcx feature that shows pcx image shortly on startup. This may be tested with a random game. From the top of my head, "Heroine Quest" had it.

ivan-mogilko:
Speaking of the code, simply comparing BMP and PCX load functions, there are bits that are repeated in both:
* temp palette declaration (a subst if a user did not provide one)
* the image conversion afterwards

This means that these operations do not depend on format, and therefore should be picked out.

My proposal would be to instead have 1 function that works as a wrapper and does these two things, while calling the actual loading function in the middle.

E.g. something like
<pre>
Bitmap* LoadImage(Stream *in, RGB *pal, int type) {
    PALETTE tmppal;
    bool want_palette = true;
    /* we really need a palette */
    if (!pal) {
        want_palette = false;
        pal = tmppal;
    }

    // call a loadBMP/ loadPCX depending on a type

    if (dest_depth != bpp) {
        /* restore original palette except if it comes from the bitmap */
        etc......

     /* construct a fake palette if 8-bit mode is not involved */
    if ((bpp != 8) && (dest_depth != 8) && want_palette)
        generate_332_palette(pal);

    return bmp;
}
</pre>

Maybe there's something else shared that I did not notice yet.

The `type` check may instead be replaced with a function pointer, e.g.
```
typedef Bitmap* (*load_fmt)(Stream *in);
```

Then the main function becomes:
```
Bitmap *LoadImage(Stream *in, RGB *pal, Bitmap* (*load_fmt)(Stream *in, RGB *pal)) {
```

And you can have, say `Bitmap *LoadBMP(Stream *in, RGB *pal)` and `Bitmap *LoadPCX(Stream *in, RGB *pal)`, and call LoadImage like:
```
LoadImage(in, pal, LoadBMP);
LoadImage(in, pal, LoadPCX);
```

ericoporto:
@ivan-mogilko I did a minor refactor to remove the duplicated code, I added you in my repo if you want to do the change you mention, also it's alright to pick the commits and do your version. I realized I do not know how to use function pointers yet...

ericoporto:
[TestCreateFile.zip](https://github.com/adventuregamestudio/ags/files/9815435/TestCreateFile.zip)

Hey, went a bit different route using enums and tried to do something that matches the file/asset extension to the enum. Above is a small test project I made, I used your code but added an overlay that shows up for two seconds, just so I could verify it was working correctly.

ericoporto:
@ivan-mogilko  thanks, pulled directly from your commit!

These commits could all use merging everything in a single commit possibly, but I managed to figure out what was wrong with the palette, I had to use ReadByte to get the same behavior - the division by 4 from the int8_t was giving a different result from the int in ReadByte. Now it looks like it's working with all file types

ivan-mogilko:
> I had to use ReadByte to get the same behavior - the division by 4 from the int8_t was giving a different result from the int in ReadByte. Now it looks like it's working with all file types

I suspect that the point was to cast ReadInt8 to unsigned char (or uint8_t). ReadByte returns practically uint8_t stored in int32.

ericoporto:
Done! I really think this all should be merged in a single commit before merging, but I think it's working. I used the [three games in the comment above](https://github.com/adventuregamestudio/ags/pull/1682#issuecomment-1283074982) for testing.

Common/gfx/bitmap.cpp
```diff
@@ -256,9 +256,9 @@ static void read_palette(int bytes, RGB *pal, Stream *in, bool is_os2)
     int i, j;
 
     for (i=j=0; (i+3 <= bytes && j < PAL_SIZE); j++) {
-        pal[j].b = in->ReadByte() / 4;
-        pal[j].g = in->ReadByte() / 4;
-        pal[j].r = in->ReadByte() / 4;
+        pal[j].b = static_cast<uint8_t>(in->ReadInt8()) >> 2;
```
ivan-mogilko, line 259: I don't think it's proper or worth to replace `/4` with `>>2` here. Compilers today are smart enough to replace division on 2, 4 etc with bit shift, while with division here will be easier to understand the code intention.

Common/gfx/bitmap.cpp
```diff
@@ -221,6 +218,1003 @@ void ReadPixelsFromMemory(Bitmap *dst, const uint8_t *src_buffer, const size_t s
...
+    out->WriteInt32(hdr.mapSizeImage); /* image size of raw bitmap data; a dummy 0 can be given for BI_RGB bitmaps. */
+    out->WriteInt32(hdr.xPxPerMeter);  /* horizontal resolution. (pixel per metre, signed integer) (0xB12 = 72 dpi) */
+    out->WriteInt32(hdr.yPxPerMeter);  /* vertical resolution. (pixel per metre, signed integer)  */
+
+    out->WriteInt32(hdr.colorsCount);          /* number of colors in the color palette */
+    out->WriteInt32(hdr.importantColorsCount);          /* number of important colors used */
+
+    /* palette */
+    for (i=0; i<hdr.colorsCount; i++) {
+        out->WriteInt8(pal[i].b);
```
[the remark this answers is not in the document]
ericoporto: Thanks, I will adjust this!

Common/gfx/bitmap.cpp
ericoporto: I will adjust this to use the division!

ericoporto:
I also found in the preload code a place where a tab was inserted wrongly instead of space. I will adjust these and merge all this in a single commit.

ivan-mogilko:
I guess, the tests should include saving and reloading same image, in order to ensure that saving works correctly.

ericoporto:
Uhm, I have tried saving and loading in 32-bit game, which uses 24bpp for saving, and can load either 24bpp or 8bpp.

But for saving and loading 8pp, I think I need to make a full 8-bit game, I am trying to make one here for testing. I think I found another bug where UpdatePalette needs to also refresh the bitmap-texture cache.

ericoporto:
[TestSaveLoad8Bit.zip](https://github.com/adventuregamestudio/ags/files/9834433/TestSaveLoad8Bit.zip)

Made me an 8-bit test game, attached above! Conclusion: it works, but the saved sprite will have the game palette, instead of whatever was the original palette. It took me a little because I ended up trying to play around with 8-bit when I got it work and forgot what I was doing...

PR #1683, opened 2022-06-09
3.6.0: replaced hanging loop iterations check with a timeout, and handle closing window during hanging script

### LINKED ISSUE #1411: Engine is not able to close the game with the "hanging" script
**Describe the bug**
If the game features a script that "hangs", e.g. with infinite loop, the game window cannot be closed. Before 3.6.0 engine would simply not respond to a app close event, in 3.6.0 it seem to crash in this situation.

**AGS Version**
Probably any recent version, tested in 3.5.1 and 3.6.0.

**To Reproduce**
Have following script in game:
<pre>
void noloopcheck Hang()
{
   while(true) {
      int i; //arbitrary no-op
   }
}
 
void repeatedly_execute_always()
{
    Hang();
}
</pre>

**Expected behavior**
Game script should not prevent the engine from closing the game. At the minimum closing application when receiving a system event should work (for example, player closing the game window). If possible, AGS abort key (alt+x by default) should also work, but I'm not sure if it's easy to do without larger code refactor.

ivan-mogilko:
Resolves #1411.

There's a "hanging script" check which was done based on number of loops passed in script (basically, execution jumps) without any engine update in between them. This PR replaces that with a timeout instead (based on real time). There are two timeouts in fact: a minimal timeout, after which engine will do a system events poll, and a "critical" timeout, after which the engine will either forcefully kill the script if there's no `noloopcheck` keyword used with a function (as before) or merely log a warning.

Doing events poll allows player to close the window by hitting an "x" button or by any other regular method supported by the system, and not having to kill the process.


ericoporto:
Just to check if the behavior is as expected. I added the following code in the room script:

```AGS Script
function noloopcheck room_AfterFadeIn()
{
  while(true) {
    String a = "nothing";
  }
}
```

I can't quit using alt+x but I now can exit the game by hitting alt+f4 (this wasn't possible before this PR).

If I remove the `noloopcheck` keyword, I can now quit using both alt+x and also using alt+f4 (again, wasn't possible before this PR).

ivan-mogilko:
> If I remove the `noloopcheck` keyword, I can now quit using both alt+x

I don't see how that's possible, alt+x is not supposed to work, because key input is not handled while the game script is in the loop.

Without `noloopcheck` the game is supposed to abort after "critical timeout".

> Also, weird thing, the code referenced in the issue https://github.com/adventuregamestudio/ags/issues/1411 still hangs for me.

I see, this solution is not ready yet then.

ivan-mogilko:
> Also, weird thing, the code referenced in the issue https://github.com/adventuregamestudio/ags/issues/1411 still hangs for me.

Pushed with a small change, which fixed this for me.

ericoporto:
@ivan-mogilko I confirm this is working. Curiously, if you do alt+x and then hit alt+f4, it will open a small windows saying that abort key was pressed.

PR #1693, opened 2022-06-18
Editor: added Help command for pane's tab context menu

ivan-mogilko:
Just a small suggestion, added "Help" command to the document's tab context menu, in case user does not know about F1, also if there are multiple subpanes with different keywords, this will ensure displaying help for the pane itself:

panectxhelp


ericoporto:
I paired this with the fixes in the chm generated [here](https://github.com/adventuregamestudio/ags-manual/actions/runs/2522385869) and it was possible to click and go to a page in the manual for all of the right click Help! This is a very interesting additional feature. I think it is a good one and motivates to improve each separate Editor Pane page.

PR #1697, opened 2022-06-19
Tidy Library classes

ivan-mogilko:
This is not related to any open issue; but when debugging one problem recently I realized that Library class does not save library's name etc. So I decided to add this information, and tidy the code along.

* No significant functional changes.
* Tidy code and apply the preferred code style.
* Hid class members that should not be public.
* Fixed implementation that did not have safe checks when disposing.
* Save name and path in the Library class.
* Added GetName(), GetFilePath() and IsLoaded() members.

In addition use wide-char winapi function in the WindowsLibrary (because we are supposed to support utf-8 paths now).

TODO: need to be tested on Linux and Android.

### DIFF
--- a/Engine/plugin/agsplugin.cpp
+++ b/Engine/plugin/agsplugin.cpp
@@ -1015,10 +1015,9 @@ Engine::GameInitError pl_register_plugins(const std::vector<Common::PluginInfo> 
             apl->filename = "ags_snowrain";
         }
 
-        String expect_filename = apl->library.GetFilenameForLib(apl->filename);
         if (apl->library.Load(apl->filename))
         {
-          AGS::Common::Debug::Printf(kDbgMsg_Info, "Plugin '%s' loaded as '%s', resolving imports...", apl->filename.GetCStr(), expect_filename.GetCStr());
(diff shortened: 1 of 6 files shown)


Engine/plugin/agsplugin.cpp
```diff
@@ -1016,10 +1016,9 @@ Engine::GameInitError pl_register_plugins(const std::vector<Common::PluginInfo>
             apl->filename = "ags_snowrain";
         }
 
-        String expect_filename = apl->library.GetFilenameForLib(apl->filename);
         if (apl->library.Load(apl->filename))
         {
-          AGS::Common::Debug::Printf(kDbgMsg_Info, "Plugin '%s' loaded as '%s', resolving imports...", apl->filename.GetCStr(), expect_filename.GetCStr());
+          AGS::Common::Debug::Printf(kDbgMsg_Info, "Plugin '%s' loaded as '%s', resolving imports...", apl->filename.GetCStr(), apl->library.GetFilePath().GetCStr());
```
ericoporto, line 1021: In the string, the meaning changed from "loaded as {expect_filename}" to "loaded from {file path}", so better use the word from there.

PR #834, opened 2019-06-12
[WIP] AGSSpriteFont modifications from KrusAGS and some fixes

onitake:
:warning: **DO NOT MERGE** This is a work in progress.

It needs some more work and testing (with older and newer games) to be mergeable.

This PR contains all modifications to the AGSSpriteFont plugin from the KrusAGS fork, plus further fixes to incorrect data types and bugs in the modified code.

List of changes:
* Added a new script function, `void SetLineHeightAdjust(int fontNum, int LineHeight, int SpacingHeight, int SpacingOverride)` that allows changing text layout properties. It is used by (at least) KR and WoaM
* Added adhoc colour changing in the `Draw()` function. `RenderText()` already has a a `color` argument, but this was never used previously.
* Fixed a number of incorrect data types in the `Draw()` functions. The framebuffer was previously accessed using aliased pointers of generic C types. On certain CPU architectures, this may lead to incorrect data offsets (`long` is usually 64 bits on 64-bit architectures), unaligned memory access or even crashes. These types were replaced by fixed-size `stdint.h` types.
* Fixed some more instances where `int` was used instead of a more appropriate `size_t`

The KrusAGS fork is found here: https://github.com/josthas/KrusAGS/
It was forked off AGS 3.3.0 to implement engine feature additions for several commercial games, but never merged back into the main AGS code.

In particular, Kathy Rain and Whispers of a Machine make use of this fork.

See #751 for the (lenghty) discussion on this fork.

@ivan-mogilko @rofl0r @i30817 Please provide feaddback, if you can. Thanks!

Issue #868, opened 2019-07-13
Whispers of a Machine is run in wrong native resolution


ghost:
When run with our AGS engine "WoaM" appears to have higher native resolution than when run with the original executable. This is easily noticeable during intro sequence because the title splashscreens do not cover whole screen, and later in the main menu the game version is printed not in the left-bottom corner but somewhat higher.

If you run winsetup in the original game it indicates custom resolution 640x360, but the game data defines standard size of 640x400. Looking into [their engine code](https://github.com/josthas/WoaMAGS/blob/master/Engine/main/graphics_mode.cpp#L212) it's apparent they modified the meaning of standard resolution constants.

I think in this situation the only way to force it run correctly (besides guessing by game GUID) is to support overriding options in config which tell wanted native resolution (and players will have to type this in config by hand).

morganwillcock:
Could they not patch their game to specify a custom resolution instead of modifying constants?

ivan-mogilko:
> Could they not patch their game to specify a custom resolution instead of modifying constants?

Their custom engine is made on top of AGS 3.3.0, while custom resolutions were implemented in 3.4.0.
So the question is rather, may they upgrade to 3.4.0 (at least). It's also still unknown what other changes they had there (I did not compare their code fully to our 3.3.0 yet). And apparently their Android port works better too (https://github.com/adventuregamestudio/ags/issues/863#issuecomment-511150292) (or maybe it was our old Android port that worked better).

AlanDrake:
I don't think tyhis should be counted as an actual issue.
Making guesses about custom engine versions is not something AGS should do.

fernewelten:
Analyzing C++ code and fiddling with it is a lot of very time-consuming hassle. So the game authors must have **felt a lot of pain** before resorting to this.

**Why** would they have thought that they cannot live without the 600x400 resolution constant, even to the point of perusing the code for it and then patching the code? 

That would be the **real** issue behind the issue. If there is a bad “deficiency” in the AGS code that annoys other game writers, too, then we can improve the engine for everyone by fixing the “deficiency” in the main code path.

Apart from that, AGS isn't SCUMM; we are under no obligation to maintain compatibility with every obscure engine modification hack or copy protection that has built into Adventure games anytime anywhere. 

[...]

PR #1705, opened 2022-06-25
Implement AGSSpriteFont plugin for Clifftop Games

criezy:
This pull request does mainly two things:
* Register the AGSSpriteFont plugin (since currently it is not used and stub functions are used instead).
* Implement the variant of the plugin used for Clifftop Games (Kathy Rain and Whispers of a Machine).

It also fixes some issues with the AGSSpriteFont plugin, such as memory leaks.

I am aware of PR #834, but I used a different approach in the implementation here: I added two new font renderer classes that derive from the unmodified ones of the original plugin and implement the modified functions only. The advantage of this approach is that we can either instantiate the base renderers or the modified ones depending on the game. However that is something I did not actually know how to do (see f89a2f5ff14c111efd7ed52a57ad309cf491f7cb) so any guidance on that point would be welcome.

This is the same approach I used previously in ScummVM to implement the Clifftop Games variant of the plugin and most of the changes in this pull request is borrowed from my previous work. Sadly the approach we used in ScummVM to select the renderers is probably not applicable here since it relies on the game detection system specific to ScummVM.

Another impact of this design decision is that I had to make the `IAGSFontRenderer` destructor virtual so that the destruction of the renderers in the plugin works properly.

I updated the Visual Studio and Xcode files, but this was not tested and I don't know if this is correct.
The changes were only tested on macOS using cmake.

And finally here is a comparison with Kathy Rain:

Current AGS code:
image

With the changes from this pull request:
image

### DIFF
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -507,10 +507,14 @@ if (AGS_BUILTIN_PLUGINS)
         ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFont.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFontRenderer.cpp
         ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFontRenderer.h
+        ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFontRendererClifftopGames.cpp
+        ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFontRendererClifftopGames.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthFont.cpp
         ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthFont.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthSpriteFont.cpp
         ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthSpriteFont.h
+        ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthSpriteFontClifftopGames.cpp
+        ../Plugins/AGSSpriteFont/AGSSpriteFont/VariableWidthSpriteFontClifftopGames.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/color.cpp
         ../Plugins/AGSSpriteFont/AGSSpriteFont/color.h
 
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -500,6 +500,7 @@ if (AGS_BUILTIN_PLUGINS)
         ../Plugins/ags_snowrain/ags_snowrain.h
 
         ../Plugins/AGSSpriteFont/AGSSpriteFont/AGSSpriteFont.cpp
+        ../Plugins/AGSSpriteFont/AGSSpriteFont/AGSSpriteFont.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/CharacterEntry.cpp
         ../Plugins/AGSSpriteFont/AGSSpriteFont/CharacterEntry.h
         ../Plugins/AGSSpriteFont/AGSSpriteFont/SpriteFont.cpp
--- a/Engine/Makefile-objs
+++ b/Engine/Makefile-objs
@@ -55,7 +55,7 @@ else
 COMMON := $(wildcard $(addsuffix /*.cpp,$(COMMON_subdirs))) $(wildcard $(addsuffix /*.c,$(COMMON_subdirs)))
 endif
 
-PLUGINS = ../Plugins/AGSflashlight/agsflashlight.cpp ../Plugins/agsblend/AGSBlend.cpp ../Plugins/ags_snowrain/ags_snowrain.cpp ../Plugins/ags_parallax/ags_parallax.cpp ../Plugins/agspalrender/ags_palrender.cpp ../Plugins/agspalrender/raycast.cpp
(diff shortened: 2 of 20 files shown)


ericoporto:
@criezy I think the build errors is the missing objects in this file: https://github.com/adventuregamestudio/ags/blob/master/Plugins/AGSSpriteFont/Makefile-objs

i30817:
Time to delete the hacked up plugin i had to make this 'work'.

BTW, there is another place than dialog font where this plugin effect is apparent (iirc, it's been some years). In the ending zone of the game there is a puzzle with a refrigerator magnets. It used letters, which were invisible (iirc) with the old plugin.

criezy:
To make sur this is not missed: **this pull request is currently missing one crucial peace of the puzzle to work with Kathy Rain and Whispers or a Machine**: there are two variants of the font renderer in the plugin and the plugin needs to know which variant to instantiate. Currently it always instantiates the original version. For the screenshot I included I just changed the code to always instantiate the Clifftop Games variant of the renderers, but that is obviously not a good solution.

criezy:
I have added one more commit to properly use the Clifftop Games variants of the renderers for Kathy Rain and Whispers of a Machine.

The renderer selection is based on the game name and uses the `Game::get_Name` plugin function. But because the game name is not yet set in the `play` GameState struct when `AGS_EngineStartup` is called for the plugins, I had to delay the renderers instantiation.

I don't find this solution very clean, but I could not think of a better one.

ivan-mogilko:
> The renderer selection is based on the game name and uses the Game::get_Name plugin function. But because the game name is not yet set in the play GameState struct when AGS_EngineStartup is called for the plugins

I could see if it's okay to move the "play" struct initialization earlier; but I wonder if it's a good idea to call script functions in AGS_EngineStartup, as the scripts nor game loop itself is run at the time. Also, Game.Name is a non-unique settable property and may be freely set or even changed in script at runtime.

Since this is about porting/enhancing plugin anyway, an alternate solution could be to expand the plugin API, either pass some info on startup, or add an API method to retrieve an extendable info struct from a game. For instance we added a new `GetRenderStageDesc` method recently for the purpose of better supporting 3D plugins.

criezy:
> Since this is about porting/enhancing plugin anyway, an alternate solution could be to expand the plugin API, either pass some info on startup (game's GUID, etc), or add an API method to retrieve an extendable info struct from a game.

That sounds like a better solution. I will take a look at that in the next few days.

ivan-mogilko:
> Another impact of this design decision is that I had to make the IAGSFontRenderer destructor virtual so that the destruction of the renderers in the plugin works properly.

Regarding this... I might need to double check this, but if I recall right, we intentionally kept the interface classes of the plugin API without virtual destructors, as these change virtual tables of such classes, which in turn could break compatibility with some plugins. The decision was to *not* use pointers of the interface class's type when deleting an object. (Engine cannot do that anyway, as it does not know what kind of memory the pointer points to: an object on heap, an object on global stack in plugin, etc.)

Perhaps the alternate solution could be to change the class hierarchy within the plugin, and make its own parent class with virtual destructor.

criezy:
> > Another impact of this design decision is that I had to make the IAGSFontRenderer destructor virtual so that the destruction of the renderers in the plugin works properly.
> 
> Regarding this... I might need to double check this, but if I recall right, we intentionally kept the interface classes of the plugin API without virtual destructors, as making them virtual would also change virtual tables of such classes, which in turn could break compatibility with older plugins. 

I reviewed that change, and I did not actually need to make the destructor virtual in `IAGSFontRenderer`. It was sufficient to make it virtual in the standard font renderers of the AGSSpriteFont plugin. The custom renderers for Kathy Rain and WOAM derive from those already. I have now forced-pushed a change to remove the changes to the `IAGSFontRenderer` class.

I have also pushed another two commits to add a function in the plugin interface to get game info, and to use that in the AGSSpriteFont plugin to determine which renderers to instantiate. If this looks like a good approach I can squash commits 885a3d7 (the old approach) and 6814b60 (the new approach).

ivan-mogilko:
@criezy I think it looks fine now. Could you also add `guid` and `uniqueid` fields to the struct? In my opinion, it is safer to test the game by guid if it's available rather than the title.

ivan-mogilko:
Hmm, ah, don't know why I did not think about this earlier... probably was distracted and not thinking this through. But maybe this plugin could expose a switch between renderers into the script. Because right now this switch will be hardcoded. So it's good for running with pre-existing games, but becomes useless for making new ones.

I guess this is something for the future TODO list though.

criezy:
> could you also add `guid` and `uniqueid` fields to the struct (`uniqueid` is an older integer game id, used before `guid` was added)?

Done

> In my opinion, it is safer to test the game by guid if it's available rather than the title.

I also made that change.

ivan-mogilko:
I tried this with "Kathy Rain" on Windows, built plugin using MSVS solution***** (don't know if that matters), and although the font itself seem to load up fine, but the text colors are not applied correctly making speech unreadable. Does this work for you on other platforms?

This is what I see:
spritefontplugintest-001



***** as we don't have a configuration for builtin plugins in MSVS project, I had to build engine and plugin separately.

ivan-mogilko:
Ah, I just realized that the engine has not been updated yet, so it does not include the new plugin API.

ivan-mogilko:
Alright, I've built tested this with "Kathy Rain" and "Woam" on Windows, building from MSVS solution.

"Kathy Rain" is working fine, from what I see.

"WoaM" has some problem with the text height calculation, apparently. In the main menu the option texts are cut from below, and the speech text is missing outlines (I think?):

spritefontplugintest-002

I don't know if this is related, but with "WoaM" there's another problem. We had a ticket about it, but in the end decided to not fix this, as it's related to their custom engine: #868.

criezy:
> "WoaM" has some problem with the text height calculation, apparently. In the main menu the option texts are cut from below, the line spacing is too small, and the speech text is missing bottom parts of the outlines (I think?):
> 
> spritefontplugintest-002
> 
> Do you also have same problem? what about scummvm?

I indeed see the same problem. I also checked with ScummVM and this appears to be a regression in the engine code.
ScummVM 2.5.1 is displaying the text correctly:
image

But with current code from ScummVM master I see the same issue you saw:
image

Between those two versions dreammaster and myself have merged a lot of changes from the AGS engine in ScummVM. I will do a bisection to find what is the culprit (it will be easier than to do it on upstream AGS since at least in ScummVM we have a AGSSpriteFont plugin that worked).

However the issue with the menu is specific to AGS:
image

And in current ScummVM:
image

The first image seems familiar though, so this is likely an issue that we had in the past in ScummVM as well, but have fixed in the engine since.

On the same screen I also see the version text is cut-off in AGS but not in ScummVM.

However there is a regression in ScummVM on the load screen:
ScummVM 2.5.1:
image

Current ScummVM (and AGS):
image

> I don't know if this is related, but with "WoaM" there's another problem where engine does not handle its screen size correctly. We had a ticket about it, but in the end decided to not fix this, as it's related to their custom engine: #868.

This is not related. We have this issue in ScummVM as well (and have not fixed it), but as mentioned above the text display used to work.

> In regards to MSVS solution, it has to be upgraded, and is missing couple of include paths, but it's not a big deal and I guess we may fix this separately.

Since I don't have Windows I did the updates blindly in a text editor for it. So I am not overly surprised I missed some things.

ivan-mogilko:
Well, I guess we may merge this then, since plugin itself seem to work. Then fixing these remaining problems may be done separately in the engine.

i30817:
That game also has a problem in the 'abilities' that you can see in the intro scenario. If you click on for instance, the analyze ability the text appears in a completely different place from where the popup menu is. It's a bit awkward.

i30817:
@criezy , this was recently fixed here (the previous fix had some problems), as well as the custom resolution. Could you downstream the fixes to scummvm, or are you waiting to see if it all settles down a bit, since there are still some minor changes happening?

PR #1710, opened 2022-06-25
Android: add credits

### LINKED ISSUE #1470: Android Launcher: Credits do not work
**Version of AGS**
3.5.1, 3.6.0

**Problem**

1. Run the Android AGS Launcher (now called "AGS Player");
2. Open the settings menu ("..." in the top-right corner);
3. Select "Credits".
4. A completely blank page is displayed. Pressing BACK will bring you back to the games list.

**Expected**

A valid credits text is displayed.

Note that 3.5.1 does not display anything either, so this got broken a while ago. Need to research what was the original behavior, restore it or replace with something new but valid.

ericoporto:
android_credits

fix #1470

### DIFF
--- a/Android/library/runtime/src/main/res/layout/credits.xml
+++ b/Android/library/runtime/src/main/res/layout/credits.xml
@@ -1,7 +1,37 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
-  xmlns:android="http://schemas.android.com/apk/res/android"
-  android:orientation="vertical"
-  android:layout_width="fill_parent"
-  android:layout_height="fill_parent">
-</LinearLayout>
+<ScrollView
+    xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
So, this just is simply a manual copy for now?

BTW I noticed there are few things that should not be in credits anymore:
* Hq2x and Hq3x scalers were discontinued in 3.6.0 (maybe temporarily);
* alogg is no longer used;
* "Windows engine uses libcda CD player by Peter Wang." perhaps should not be mentioned in Android engine.

ericoporto:
> So, this just is simply a manual copy for now?

I delayed this for a long time but I could not figure any easy way to do this. It's possible to vaguely automate this in gradle, but it puts one more remote dependency to grab this info from the manual... In the end, I think we better customize and have one here, hopefully we won't have to update this as much from now on - the allegro stuff pretty much survived like ten years.

PR #1718, opened 2022-06-27
CMake: debug defines only set per external config

ericoporto:
We used to set macro DEBUG_SPRITECACHE=1 in Debug builds and DEBUG_SPRITECACHE=0 in Release builds,
and also set DEBUG_MANAGED_OBJECTS=0 or 1 depending on the command line config of AGS_DEBUG_MANAGED_OBJECTS

But in the code, we actually test using `#ifdef` instead of `#if`.

This commit

- only define DEBUG_MANAGED_OBJECTS if AGS_DEBUG_MANAGED_OBJECTS is true
- adds a new command line flag AGS_DEBUG_SPRITECACHE, which will only define DEBUG_SPRITECACHE if true

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(AGS_NO_MP3_PLAYER "Disable MP3" OFF)
 option(AGS_NO_VIDEO_PLAYER "Disable Video" OFF)
 option(AGS_BUILTIN_PLUGINS "Built in plugins" ON)
 option(AGS_DEBUG_MANAGED_OBJECTS "Managed Objects Log" OFF)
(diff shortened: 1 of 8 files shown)


ericoporto:
But should `DEBUG_SPRITECACHE` be tied to debug build or should it be changed to be enabled through a switch as done here?

ivan-mogilko:
`//#define DEBUG_MANAGED_OBJECTS` is not a correct way to do this anymore, because these log messages belong to a certain log group now, which is only registered if the macro is defined globally.

ericoporto:
Redid the PR going the `#if` route instead.

PR #1723, opened 2022-06-29
Editor: fix for autocomplete not selecting proper line

### LINKED ISSUE #1599: Editor: Auto Complete not always selecting proper line from the list
**Describe the bug**
Auto complete doesn't always complete the typed name e.x. AGS_MAX_REGION doesn't work, but AGS_MAX_OBJECT works just fine

**AGS Version**
4.0

**To Reproduce**
1. Type AGS_MAX_R and noticed that auto complete doesn't select the proper name from the list.
2. Type AGS_MAX_O and it will select AGS_MAX_OBJECTS correctly

**Expected behavior**
It should select proper name from the list

**Screenshots**
Built in AGS variable
https://i.gyazo.com/40993ad3bce2ff8e00ed80ab1b79c85e.mp4

Custom variable
https://i.gyazo.com/050dc2ea83d4bd1f1fe373f12800d0dc.mp4

 - OS: Windows
 - Version 10
 
 **Additional Context**
 I have another projected opened on the side in 3.5.1 version of AGS(No idea if having 2 projects opened at the same time, even in 2 versions of the editor could be responsible for it, but who knows...)

ivan-mogilko:
Fixes #1599, at least partially. Would need to check all mentioned cases in that ticket, and ammend with more fixes if necessary.

The root of the "underscore" problem was that Scintilla library used an uncommon "case insensitive comparison" implementation. While many libraries, including .NET, seem to use lowercase values for that, Scintilla converts to uppercase. Uppercase letters have different relational position to "underscore" character. Changing it to use lowercase comparison instead makes it match the Editor's sort.

### DIFF
--- a/Editor/scintilla/lexlib/CharacterSet.cxx
+++ b/Editor/scintilla/lexlib/CharacterSet.cxx
@@ -18,10 +18,10 @@ namespace Scintilla {
 int CompareCaseInsensitive(const char *a, const char *b) noexcept {
 	while (*a && *b) {
 		if (*a != *b) {
-			const char upperA = MakeUpperCase(*a);
-			const char upperB = MakeUpperCase(*b);
-			if (upperA != upperB)
-				return upperA - upperB;
+			const char lowerA = MakeLowerCase(*a);
+			const char lowerB = MakeLowerCase(*b);
+			if (lowerA != lowerB)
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Closed in favor of #1726

PR #1725, opened 2022-06-29
Editor: fix for autocomplete not working for extenders if base type is in another script

### LINKED ISSUE #1623: Editor: Auto complete list does not appear for extended structs
**Describe the bug**
Auto complete does not appear in various cases(one case that can be tested are extended structs)
As if the ags doesn't recognize the struct, yet the game compiles and struct properties can be read and used in code.
Similar issue was in 3.5.1 but I haven't tested with extended structs.

**AGS Version**
4.0

**Game**
[AutoCompleteTest.zip](https://github.com/adventuregamestudio/ags/files/8529711/AutoCompleteTest.zip)

**To Reproduce**
1) Open a script in Test > Test Auto Complete
2) In game_start function(or any other) type `dialogConfig.` to see if any properties auto complete

**Expected behavior**
Auto Complete should show all properties and functions from a base struct + extended struct, currently it doesn't show anything.

 - Windows 10

**Additional context**
Scripts in Config folder are used to setup the extended structs.
The bug also happens in the config script itself in `Initialize` function which means that export/import should not be causing it.
Struct type used in script is not colored green even after restarting the editor.

ivan-mogilko:
Fixes #1623

Local script's autocomplete failed to register an extended struct or extender method properly if the base type is declared in another script. This change makes it so that whenever `AutoComplete.ConstructCache()` is called, it also accepts a list of imported headers, from which it's possible to retrieve their own autocomplete cache(s). This is used to know which other types exist.

Note that autocomplete cache is also used for a) "Go to definition" and b) keyword highlighting. So these are also fixed for the extended structs.

NOTE: based on top of #1723 for convenience.


ivan-mogilko:
Closed in favor of #1726.

Issue #1454, opened 2021-11-22
Editor bug: Autocomplete fails for Room objects after game is run in editor


vga-256:
Problem: When the game is run via the editor, autocomplete stops working for all Room objects in the currently open room script. This is the same behaviour first [noticed here by Silverspook and Khris in 2015](https://www.adventuregamestudio.co.uk/forums/index.php?topic=52313.msg636515895#msg636515895).

Editor version: 3.5.1.11

Workaround: Closing the room script and re-opening it after running the game re-enables autocomplete for room Objects.

ivan-mogilko:
Not 100% certain, but I think this may be related to room itself not being loaded into the memory. As the editor gets object names from room data, if the room is not loaded, then there's no way to get this information from.

So, workaround is actually to open the room in the editor.

On fixing, I don't have good easy ideas at the moment. Maybe this could be fixed by preloading certain information from the room and keeping it in the cache.

ericoporto:
This bug is super annoying.

> So, workaround is actually to open the room in the editor

I think the main issue is the room looks open, but apparently it's not. I am curious if there is a way to have the generated header or autocomplete objects in a way that is independent of the room being open, so one could edit two or more room scripts at the same time, with the autocomplete in each.

ivan-mogilko:
> I am curious if there is a way to have the generated header or autocomplete objects in a way that is independent of the room being open, so one could edit two or more room scripts at the same time, with the autocomplete in each.

This should be doable, both in ags3 and ags4, although would require different approach.
All is needed is hotspot and object script names, which may be loaded from either room crm or xml. See crm2ash tool for loading this from crm.
The names may be cached in editor memory, as that's not a lot of data. Then updated whenever room is modified or saved after being modified.

ivan-mogilko:
So, I guess, it should be like this:

1. Have a dictionary of autocomplete cache objects per room somewhere. For example, in a RoomComponent (but need to figure out if that will be convenient).
2. On room itself loaded: generate autocomplete for this room and put/update into this dictionary.
3. On modified room saved: regenerate and update an element in this dictionary.
4. On room script opened: check if autocomplete for this room exists, if not then load partial (minimal necessary) room data (either from crm in ags3 or room's data.xml in ags4) and update an element in this dictionary.
5. When generating autocomplete for the room script: take an element from this dictionary and append to gathered symbols.

ericoporto:
[...]

Issue #364, opened 2016-10-02
Autocomplete not ignoring preprocessor directives in function header


monkey0506:
If a function header includes a `#ifdef` or `#ifndef` block that is not preceded by an optional argument, then autocomplete includes the directive as part of the previous or current argument's definition:

```
int myFunc(
  #ifdef SCRIPT_API_v340
  int param=0
  #endif
  )
{
}
```

Will autocomplete as:

```
int myFunc(optional #ifdef SCRIPT_API_v340
int param)
```

And

```
int myFunc2(int a
  #ifdef SCRIPT_API_v340
  , int param=0
  #endif
  )
{
}
```

Will autocomplete as:

```
int myFunc2(int a
#ifdef SCRIPT_API_v340, optional int param)
```

This behavior does not happen if the preprocessor directive follows an optional argument (e.g., it doesn't happen to `Character.ChangeRoom`).

Additionally, it includes the ROL, so even autocomplete tags like `$AUTOCOMPLETEIGNORE$` are not utilized or seen as comments, but included as part of the function header.

rofl0r:
just a thought... when seeing this: `#ifdef SCRIPT_API_v340` it appears to me that `#if SCRIPT_API >= 340` would be a way more powerful idiom...

monkey0506:
@rofl0r you are correct, but the AGS preprocessor doesn't have an `#if` directive (currently).

ivan-mogilko:
Speaking of autocomplete, I stumbled upon it when doing some changes to the script, and I found that it actually duplicates precompiler instead of reusing it.
I wrote a quick note in the code: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Panes/ScintillaWrapper.cs#L1742
Maybe if someone is going to work on this could take this into consideration.

gurok:
Why was this closed? As far as I know, it's still an issue.

ivan-mogilko:
The culprit for this bug is [`AutoComplete.ConstructCache()`](https://github.com/adventuregamestudio/ags/blob/ea4e6762236a13621ba0771b9d5d8ac4bcf33ee8/Editor/AGS.Editor/AutoComplete.cs#L142), which tries to handle preprocessor options in parallel with parsing script constructs, and thus potentially missing number of cases.

The solution could be to preprocess first, and then parse resulting script.

ericoporto:
> The solution could be to preprocess first, and then parse resulting script

[...]

PR #1726, opened 2022-06-30
Editor: some fixes for the script Autocomplete

### LINKED ISSUE #1481: Autocomplete issue with strings with commas in the Editor
**Describe the bug**
Sooo I'm having a weird autocomplete issue and I can't make heads or tails of it. I have a custom say function called "`GSay`", which takes two enums I declared called "`eGesture`" and "`eExpression`". It looks like this:

```AGS Script
import function GSay(this Character *, String sayThis, eExpression myExp=ex_none, eGesture myGest=eGestureNone);
```
and works like this:

```AGS Script
cJoe.GSay("Hello!",exSmile, eGestureWave);
```

It works just fine, but if I write a line a dialog that includes a comma, the autocomplete skips the "Expression" parameter and goes right to the "Gesture."

For example, the line below (which has no comma) behaves properly when autocompleting
image

But this one (which includes a comma in the string) does not.
image

**AGS Version**
3.5.1.14

 - OS: Windows
 - Version 10

**Additional context**
[**SierraTest.zip**](https://github.com/adventuregamestudio/ags/files/7831022/SierraTest.zip)

Check the room1.asc file, from the project above, in the Editor to see the autocomplete bug manifest.

Originally reported here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59644.0

ivan-mogilko:
Fixes #1481, #1599, #1623; see fix notes below.

NOTE: honestly I don't want to spend much time on autocomplete feature, because in my opinion it's not done well, and the biggest mistake is that it basically reimplements script parsing, including parsing of preprocessor commands (check [this my older comment](https://github.com/adventuregamestudio/ags/blob/0af788ef8a35580fde95eead30d48e7fbc30884e/Editor/AGS.Editor/Panes/ScintillaWrapper.cs#L1972), for example). So i think we need a task ticket for replacing it with something else in the future versions.

Anyway, this PR fixes following:

1. for #1481: fixes autocomplete for the function parameter list, not distinguishing commas in string literals and local comments (`/* */`).

2. for #1599: fixes incorrect order of some words in autocomplete list, if some of them include underscore (`_`), and possibly camelcase style (not fully certain if it was a related problem).

The root of the "underscore" problem was that Scintilla library used an uncommon "case insensitive comparison" implementation. While many libraries, including .NET, seem to use lowercase values for that, Scintilla converts to uppercase. Uppercase letters have different relational position to "underscore" character. Changing it to use lowercase comparison instead makes it match the Editor's sort.

3. for #1623: fixes extender structs not properly register for autocomplete, the struct properties did not appear in the list.

Local script's autocomplete failed to register an extended struct or extender method properly if the base type is declared in another script. This change makes it so that whenever AutoComplete.ConstructCache() is called, it also accepts a list of imported headers, from which it's possible to retrieve their own autocomplete cache(s). This is used to know which other types exist.

NOTE: autocomplete cache is also used for a) "Go to definition" and b) keyword highlighting. So these are also fixed for the above cases.

4. Fixed call tip fails to show up if you put a comma *right* after closing quote or inline comment (this seem like a regression in 3.6.0 after upgading Scintilla).
5. Fixed Enum names (not values) not included in autocomplete list.
6. Fixed local structs and enums were not highlighted in script.


ivan-mogilko:
Hmmmmm.... something went wrong after I merged all the fixes together. One fix works alone, but not in this PR. Will have to investigate...

ivan-mogilko:
Also fixed call tip fails to show up if you put a comma *right* after closing quote or inline comment (this seem like a regression in 3.6.0 after upgading Scintilla).

ivan-mogilko:
Also fixed:
* Enum names (not values) not included in autocomplete list;
* Local structs and enums were not highlighted in script;

ivan-mogilko:
Of other autocomplete issues I found in the tracker, these two:
- #364
- #1454 
would require bigger changes, so I won't be fixing them for this PR.

PR #1730, opened 2022-07-08
README: adjust forum threads and styling

ericoporto:


ivan-mogilko:
Ah, did not notice that it changes the alignment. TBH it was easier to read when it's aligned to the left, at least for me personally.

What do other people think?

PR #1736, opened 2022-07-15
Editor: zoom View's preview frame and Character preview

ivan-mogilko:
Completes #710.

The View editor's zoom was implemented by #1415, but preview frame was still unscaled.
While working on this I also noticed that the Character preview pane does not have a zoom control, unlike Inventory and Cursor preview panes which recently received a zoom control.

This PR:
1. ViewPreview control now displays a frame inside a panel with automatic scrollbar, which is displayed if the image is larger than the panel. The frame resizes if the (scaled) sprite cannot fit. It keeps the "default" size in memory, and sets it back if the sprite is small enough: this prevents small sprites from being drawn too low at the window bottom.
2. ViewPreview control now has AutoResize property, that tells it to resize itself (whole control) if the image frame is too big or small. AutoResize is *false* by default.
3. On View editor: adds a splitter between preview and the loop editor.
4. On View editor: zoom slider also affects the preview frame.
5. Character pane now has same kind of layout as Inventory Editor panes (preview panels inside a Layout Table). The two ViewPreviews have AutoResize property set to true, so they grow in size if the scaled image is too big to fit.
6. Extra: replaced FlowLayout with TableLayout on Inventory Editor (and use same on Character editor), to force a horizontal layout and scroll (may be argued and adjusted later).

### DIFF
--- a/Editor/AGS.Editor/NativeProxy.cs
+++ b/Editor/AGS.Editor/NativeProxy.cs
@@ -264,6 +264,22 @@ namespace AGS.Editor
             }
         }
 
+        public Size GetMaxViewFrameSize(View view, out bool hasLowResSprites)
+        {
+            lock (_spriteSetLock)
+            {
+                List<int> sprites = new List<int>();
+                foreach (ViewLoop loop in view.Loops)
+                {
+                    foreach (ViewFrame frame in loop.Frames)
+                    {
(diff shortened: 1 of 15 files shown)


ivan-mogilko:
Also changed the Character pane layour to keep consistent with Inventory and Cursor panes (see updated description in the first post).

Issue #1711, opened 2022-06-25
AGS 4: New compiler: remove or change octal literals syntax


ivan-mogilko:
Opened after conversation in #1675.

CC @fernewelten 

The new compiler currently supports octal literals, which syntax is `0number` (starting with `0`). There's a valid concern that such syntax may be confusing and may cause script mistakes which will be difficult to find. In overall, octal literals are rather obscure thing today, and less useful in a scripting language (personally I cannot imagine any practical use for them in AGS script); so it would be wiser to make them more explicit and less prone to typos. My suggestion is to do one of the following:

* Change octal prefix to `0o`/`0O` instead (complementing hexadecimal `0x`/`0X`), similar to javascript's strict mode, as per [ericoporto's suggestion](https://github.com/adventuregamestudio/ags/pull/1675#issuecomment-1146806667);
* Remove octal literal support altogether, to simplify the case; unless there's a real interest in having them in script.

fernewelten:
Addressed in #1738.

I've currently removed octals for the time being, but I'm considering taking up eric0o's suggestion in a later, separate PR.

ivan-mogilko:
Resolved by #1738 (disabled for now).

PR #1738, opened 2022-07-16
AGS4: New compiler: Accept LONG_MIN in scanner + parser; don't treat leading '0' as octal

fernewelten:
This PR addresses #1711 and #1674.

- Accept '2147483648' in the scanner but balk in the parser unless a unary '-' is in front.
- Accept hexadecimals between '0x80000000' and '0xFFFFFFFF' in the scanner, interpret them as negative integers. (In particular, '0x80000000' is '-2147483648'.)
- Reject hexadecimals beyond '0xFFFFFFFF'.
- Compile-time calculation: The negative of '-2147483648' yields an overflow at compile time.
- Treat integers that start with '0' and no following 'x' as decimal instead of octal.
- Add more googletests for scanner + parser. (I've slightly extended existing tests or moved them into the parser googletests but not reduced them AFAICS.)
- If error messages feature exceptionally long numbers, only show the first 20 digits.

### DIFF
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -336,6 +336,7 @@ AGS::SymbolTable::SymbolTable()
     MakeEntryLiteral(kKW_Null);
     entries[kKW_Null].LiteralD->Vartype = kKW_Null;
     entries[kKW_Null].LiteralD->Value = 0u;
+    AddKeyword(kKW_OnePastLongMax, "2147483648");
     AddKeyword(kKW_Protected, "protected");
     AddKeyword(kKW_Readonly, "readonly");
     AddKeyword(kKW_Return, "return");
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
@fernewelten is there a reason why do you store all the string literals in `char*` in tests? Like here:
https://github.com/adventuregamestudio/ags/pull/1738/commits/08b47772b2448f75699449090563270529f8e821#diff-6156fd2083abb1e22c52606bd4b549cd87676e7aa0e4dc2f7f3e422659382f64R335-R338
(and in several other places)
I've been fixing these to `const char*` [earlier](30775e563a42ee0abad710479a50dfd7dc3478ce), but this PR reintroduces this case.

fernewelten:
> is there a reason why do you store all the string literals in `char*` in tests?

No, none at all. 

> I've been fixing these to const char* [earlier](https://github.com/adventuregamestudio/ags/commit/30775e563a42ee0abad710479a50dfd7dc3478ce), but this PR reintroduces this case.

I didn't do this knowingly. I'd started my change before yours.  After finishing and before pushing, I rebased my work onto the newest AGS4 version, and it seems that some unwanted modifications have crept in when git integrated automatically. This is a very repetitive file with lots of very similar tests, and git seems to have a hard time sometimes in deciding just what to change just where.

ivan-mogilko:
It looks like tests succeed now, so I'll trust these changes.

PR #1739, opened 2022-07-16
New compiler: Handle C strings as `const char *` or `const char[]` in Googletests

fernewelten:
Perused the files and found some cases that haven't been fixed yet.

In all googletest files for the compiler:
- Change `char *` and `const const char *` to `const char *`
- Change `char inpl[]` and similar to `const char inpl[]`

### DIFF
--- a/Compiler/test2/cc_bytecode_test_0.cpp
+++ b/Compiler/test2/cc_bytecode_test_0.cpp
@@ -44,7 +44,7 @@
 TEST_F(Bytecode0, P_r_o_t_o_t_y_p_e) {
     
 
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Speaking of `const char inpl[]`, may not these be just `const char*` as well?

fernewelten:
I didn't give it much thought when creating the tests. We might as well have them consistent. 

There are some tests where the C string that makes up the AGS snippet is built up incrementally, but I think that I used `std::string` for these cases. Apart from that, I never try nor need to patch into the snippets, so yes, why not declare them `const`.

PR #1741, opened 2022-07-18
Engine: VSync toggle support

AlanDrake:
Experiment to support togglable VSync at runtime.

Currently implemented: D3D, OGL, and SW mode.
Software mode require SDL2 >= 2.0.18 (gracefully degrades depending on available version at compile)

## Notes
- D3D requires a reset(), while OGL can be switched directly.
- I removed the outdated restriction about windowed modes not supporting VSync.
- I had to force the current mode's Vsync value to `last_opposite_mode.Vsync` and `usetup.Screen.Params.VSync` so it doesn't get lost while switching from windowed<->fullscreen
- SW and D3D mode appears to keel over whenever reapplying the same vsync value at fullscreen, therefore I added a check to avoid doing so

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -811,8 +811,7 @@ void render_to_screen()
     // Stage: engine overlay
     construct_engine_overlay();
 
-    // only vsync in full screen mode, it makes things worse in a window
-    gfxDriver->SetVsync((scsystem.vsync > 0) && (!scsystem.windowed));
+    gfxDriver->SetVsync(scsystem.vsync > 0);
 
     bool succeeded = false;
     while (!succeeded && !want_exit && !abort_engine)
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -1924,7 +1924,11 @@ void OGLGraphicsDriver::SetScreenTint(int red, int green, int blue)
 
 bool OGLGraphicsDriver::SetVsync(bool enabled)
 {
-    _mode.Vsync = SDL_GL_SetSwapInterval(_mode.Vsync ? 1 : 0) == 0;
+    if (SDL_GL_SetSwapInterval(enabled) == 0)
+    {
+        _mode.Vsync = enabled;
+    }
+
     return _mode.Vsync;
 }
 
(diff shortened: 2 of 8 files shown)


AlanDrake:
Alright, looks like I'm not good enough to make changes to our build system.
But the VSync toggle appears to work with all three renderers on Windows.

What else do I need to do to fix the build system?

ivan-mogilko:
Updating SDL should be a separate PR of its own, because this is a much bigger change than the vsync toggle.

AlanDrake:
Alright, I removed the commit that attempted the SDL upgrade, then I'll rebase once we manage to get these CI changes done proper.

ivan-mogilko:
BTW, I think you may add a stub for the missing SDL function. There is a way to test if SDL version is defined:
https://wiki.libsdl.org/SDL_VERSION_ATLEAST

Something like:
```
#if !(SDL_VERSION_ATLEAST(2,0,18))
extern "C" {
int SDL_RenderSetVSync(SDL_Renderer* renderer, int vsync) { }
}
#endif
```

ericoporto:
I think 2.0.24 is out next week, if you all are alright I can make a PR to upgrade to it once it's out.

AlanDrake:
I only test on windows, so if anyone could try out what happens on other systems it would be appreciated.

ivan-mogilko:
Tested a game set to 80 fps. Theoretically vsync should restrict fps to 60 when enabled. For easier testing run the engine with "--fps" option (fps may also be shown using `Debug(4,1)` script command).

I tried this on Linux, with software and opengl renderers. Although both report that Vsync is set, neither seem to apply the fps restriction  (it stays at 80fps). This may or not be because of the virtual machine. So I'd ask someone who has a real linux machine to try this out.

On Windows, I noticed that OpenGL is always doing vsync (fps always limited to 60), regardless of the setting.

AlanDrake:
Looks like I had managed to break the OGL implementation after an ill thought cleanup... fixed.

ivan-mogilko:
The OpenGL bug is fixed.

Trying the linux vm again, it looks like OpenGL never vsyncs in fullscreen and always in windowed, regardless of the setting (at least judging by the fps).

Software mode actually works, but it's almost unnoticeable in fullscreen, because it refreshes faster, apparently. But it's clearly works in windowed.

PR #1743, opened 2022-07-23
Editor: share base menu functionality between ScriptEditor and DialogEditor

### LINKED ISSUE #1480: "Go to definition" feature in dialog editor
**Describe the problem**
Often when I'm coding, I want to zap myself to the definition of a function I created. To do this, I right-click the function name and select "Go to Definition of XXX".

In most circumstances, this works fine. Like from this room script here:
image

But if I want to do the same thing from the dialog editor, I see this instead:
image

Would it be possible to enable this feature from the dialog editor?

**Suggested change**
Add the Go to Definition of XXX left click option from the Script Editor on the Dialog Editor.

Originally from the forums here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59680.0

ivan-mogilko:
Resolves #1480

The problem with DialogEditor is that it's a stripped copy-paste version of the ScriptEditor (when it comes to working with script itself). So many menu commands that work in the ScriptEditor control are not available in DialogEditor.

The solution seem to be to create a parent class that shares a chosen functionality among these two.

In this PR, the parent class ScriptEditorBase is created, which has following functionality picked out from ScriptEditor and DialogEditor and merged:
* Menu commands and their functions;
* Context menu commands for Scintilla control and their functions;
* Various operations on a script: find, replace, goto line, goto definition, etc.

The menu commands appear to work per se, but some of them have exceptions where they don't work. This is not related to command itself, but to internal functionality of the editor class. For instance, DialogEditor does not support AutoCompleteData for some reason (maybe because it's more complicated to parse dialog scripts). Because of that neither autocomplete nor "go to definition" will work on *variables declared locally in dialog script*. I tried to investigate this, but it seems like implementing this for dialogs would require picking out much more code from the ScriptEditor, and it's not clear how well that will work. So I'd rather leave that for future consideration.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -221,6 +221,9 @@
     <Compile Include="Panes\Room\RoomEditFilters\CharactersEditorFilter.cs" />
     <Compile Include="Panes\Room\RoomEditFilters\DesignTimeProperties.cs" />
     <Compile Include="Panes\Room\RoomEditFilters\RoomFilterContextMenuArgs.cs" />
+    <Compile Include="Panes\ScriptEditorBase.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
     <Compile Include="Panes\ZoomTrackbar.cs">
       <SubType>UserControl</SubType>
     </Compile>
--- a/Editor/AGS.Editor/Panes/DialogEditor.cs
+++ b/Editor/AGS.Editor/Panes/DialogEditor.cs
@@ -2,50 +2,31 @@ using AGS.Editor.TextProcessing;
 using AGS.Types;
 using System;
 using System.Collections.Generic;
-using System.Drawing;
 using System.Text;
 using System.Windows.Forms;
 
 namespace AGS.Editor
 {
-    public partial class DialogEditor : EditorContentPanel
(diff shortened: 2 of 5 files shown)


ericoporto:
I wonder if the autocomplete issues would be fixed after a rebase, since #1726 was made and merged shortly after this.

ivan-mogilko:
> I wonder if the autocomplete issues would be fixed after a rebase, since #1726 was made and merged shortly after this.

I had plans to rebase this for 3.6.1, but autocomplete problems mentioned in this ticket will not be fixed, because #1726 did not address them in particular.

ivan-mogilko:
Recreated for 3.6.1, using old pr as a reference, because some ScriptEditor code could have changed since.

ericoporto:
Testing with simple scripts, go-to definition looks like it's working. The text above of the PR mentions the following

> Because of that neither autocomplete nor "go to definition" will work on variables declared locally in dialog script.

Yes, this is true, but I had to read the issue above to figure this was a thing because ultimately I rarely declare variables locally in a dialog script, but instead mostly call functions that are either global or in objects defined elsewhere (like a character).

ivan-mogilko:
Yes, it's a quite straightforward change, should have added this to 3.6.0, but i did not expect it to be released about 6 months later.

PR #1744, opened 2022-07-27
Introduced extended FontRenderer plugin API, more fixes for "Clifftop's" SpriteFont

### LINKED ISSUE #1742: Regression in text display in Whispers of a Machine
**Describe the bug**
The text in Whispers of a machine is not displayed correctly in some places. This was discussed in PR #1705.

**AGS Version**
Current code from master (commit c31e55c69db5356d6d22e80f05f2e5a9d6425f67).

**Game**
Whispers of a Machine
https://store.steampowered.com/app/631570/Whispers_of_a_Machine/

**To Reproduce**
1. Start the game
2. Watch the intro
3. See error

**Expected behavior**
Correct text display (see images below).

**Screenshots**

Correct:

Incorrect:

 - OS: macOS
 - Version: 11.6

**Additional context**
The screenshot above are from ScummVM where the text used to be displayed correctly. A git bisect points to https://github.com/scummvm/scummvm/commit/df29806aec718c81dfbae08e664e6a47ad76d545.

This was a backport of commit d9c4a0306a6718cb03ccd6e84c1e63d093d9a64a
Engine: replaced the remaining wgettextheight() with get_font_height()

The AGSSpriteFont plugin was not working for WOAM before PR #1705 was merged, so I did not try to check in AGS if it indeed worked before that commit (it probably wasn't). But since the behaviour we get is the same in AGS and in ScummVM I am guessing the issue in AGS has the same root cause as in ScummVM and is related to the change to the font height.

ivan-mogilko:
Fixes #1742

CC: @criezy 

This is an experimental change and related to plugin API, so opening as a draft for now. But it actually fixes all the "Whispers of the machine" font issues.

**The reasoning for changes:**

The old `IAGSFontRenderer` interface unfortunately did not provide proper means to return neither font height nor custom linespacing. Instead it had GetTextHeight(text), which in my honest opinion was a mistake to add, as fonts normally should have a specific height property even if individual characters (glyphs) have varied height.<br>
But anyway, the "Clifftop Games" abused GetTextHeight to also return linespacing depending on which text is passed as an argument, using "secret knowledge" about which lines engine uses for a font height test. This mechanic was doomed to be unreliable and eventually got broken after certain changes in the engine (see #1742 for details).

The proposed solution is to implement an extended FontRenderer interface, that lets to explicitly query for necessary font properties. In addition, engine plugin API needs a function for notifiying if the custom font have changed - because the SpriteFont allows to modify fonts from script in any given moment.

**Proposed changes**

Added `IAGSFontRenderer2` interface that inherits `IAGSFontRenderer` and adds following new members:
Member function | Description
-- | --
int GetVersion () | Returns engine API version this font renderer complies to. Must not be lower than 26 (this interface was added at API v26). If the interface is going to be expanded further in the future, the engine will be able to use this to find out whether it's safe to use newer methods.
const char* GetRendererName () | Returns an arbitrary renderer name; this is for informational purposes only (logging, etc)
const char* GetFontName (int fontNumber) | Returns given font's name (if available); again for informational purposes
int GetFontHeight (int fontNumber) | Returns the given font's height: that is the maximal vertical size that the font glyphs may occupy.
int GetLineSpacing (int fontNumber) | Returns the given font's linespacing; is allowed to return 0, telling that no specific linespacing is assigned for this font.

Added new function to the engine API called `ReplaceFontRenderer2(int fontNumber, IAGSFontRenderer2* newRenderer)`, for setting new font renderer type. The reason why it needs a separate function is mainly that the old interface IAGSFontRenderer does not have any means to find out its version, therefore engine won't be able to find out if it's an extended interface or not. But since IAGSFontRenderer2 includes GetVersion function, it will be possible from now on.

Added new function to the engine API called `NotifyFontUpdated(int fontNumber)`. This may be used by plugins to tell that the custom font has changed, so that engine could react to this (if it has cached font properties or has to redraw stuff, or else).

In **SpriteFont** plugin: all font renderers implement IAGSFontRenderer2. Fixed VariableWidthSpriteFontRendererClifftopGames class to not return text height based on secret argument handling, but instead provide font height and linespacing in corresponding new functions.

### DIFF
--- a/Common/font/agsfontrenderer.h
+++ b/Common/font/agsfontrenderer.h
@@ -80,7 +80,7 @@ struct FontMetrics
 
 // The strictly internal font renderer interface, not to use in plugin API.
 // Contains methods necessary for built-in font renderers.
-class IAGSFontRendererInternal
+class IAGSFontRendererInternal : public IAGSFontRenderer2
 {
 public:
     // Tells if this is a bitmap font (otherwise it's a vector font)
@@ -88,8 +88,11 @@ public:
     // Load font, applying extended font rendering parameters
     virtual bool LoadFromDiskEx(int fontNumber, int fontSize, const FontRenderParams *params,
         FontMetrics *metrics) = 0;
+    // Fill FontMetrics struct; note that it may be left cleared if this is not supported
+    virtual void GetFontMetrics(int fontNumber, FontMetrics *metrics) = 0;
     // Perform any necessary adjustments when the AA mode is toggled
     virtual void AdjustFontForAntiAlias(int fontNumber, bool aa_mode) = 0;
+
 protected:
     IAGSFontRendererInternal() = default;
     ~IAGSFontRendererInternal() = default;
--- a/Common/font/agsfontrenderer.h
+++ b/Common/font/agsfontrenderer.h
@@ -17,22 +17,49 @@
 
 struct BITMAP;
 
+// Basic font renderer interface.
 // WARNING: this interface is exposed for plugins and declared for the second time in agsplugin.h
 class IAGSFontRenderer
 {
 public:
-  virtual bool LoadFromDisk(int fontNumber, int fontSize) = 0;
-  virtual void FreeMemory(int fontNumber) = 0;
-  virtual bool SupportsExtendedCharacters(int fontNumber) = 0;
-  virtual int GetTextWidth(const char *text, int fontNumber) = 0;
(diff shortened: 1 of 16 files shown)


ivan-mogilko:
Fixed few bugs, added 1 potential optimization, confirmed that it's working with both "Kathy Rain" and "WoaM".

PR #1745, opened 2022-07-29
Editor: Implement "Go to Definition" for Define and Enum

AlanDrake:
Enabled the "Go to DEFINITION" contextual menu option to work with `#define`s, enum types, and enum values.

Possible improvements:
- ~track enum values to the exact line~ DONE
- throw message for builtin enums, like when trying to go to views/audios script names (unless we plan to show builtin headers in the future)

### DIFF
--- a/Editor/AGS.Editor/AutoComplete.cs
+++ b/Editor/AGS.Editor/AutoComplete.cs
@@ -209,7 +209,7 @@ namespace AGS.Editor
                 {
                     if (state.WordBeforeLast == "enum")
                     {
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
This seem to work.

> track enum values to the exact line
I suppose for this you'd need to make a separate class for Enum Value, derived from ScriptToken, and store these in ScriptEnum instead of list of strings.

AlanDrake:
Alright, ScriptEnumValue has been implemented.

PR #1747, opened 2022-08-02
3.6.0: Initialize SDL audio using env variables, don't SDL_AudioInit

### LINKED ISSUE #1692: Invalid SDL_AUDIODRIVER environment variable may prevent audio to initialize
**Describe the bug**
If you set an environment variable `SDL_AUDIODRIVER` to an invalid value, AGS may not initialize, due to SDL complaining the system does not exist.

**AGS Version**
Current master (5c3a5eabc1daf331acc2d2e9568f4f4dda4c7b0c).

**Game**
It can happen in any game. 

**To Reproduce**
1. On Windows, click on System Properties
2. Then on Environment Variables and then create a new one name `SDL_AUDIODRIVER` and set it to an invalid value (say, `dsound2`).
3. Start AGS, it will pop-up an error message

**Expected behavior**
The expected behavior is for AGS to figure a good known driver and start anyway.

**Desktop:**
 - OS: Windows
 - Version 10

**Additional context**
I am not sure if this is an AGS bug, so I opened it upstream too https://github.com/libsdl-org/SDL/issues/5818

ivan-mogilko:
Fixes #1692.

CC: @ericoporto 

Following the discussion in #1692, we must use a combination of SDL_setenv(SDL_AUDIODRIVER) and SDL_InitSubSystem here, and *not* use SDL_AudioInit, because SDL_AudioInit does not increment subsystem's reference count. Which in turn may cause problems down the way when initializing any additional SDL-based audio lib or plugin; at the very least - the mojoAl (OpenAL's implementation we're using).

If we specifically requested a driver name, and that failed, we fallback to automatic drivers list by resetting environment variable SDL_AUDIODRIVER and calling SDL_InitSubSystem(SDL_INIT_AUDIO) again. That forces SDL to try every available driver in an order.

**IMPORTANT:** For the second step, the point is to make SDL_getenv to return NULL, so that SDL_AudioInit would iterate the default drivers list. It has to be NULL, not a empty string (at least according to the SDL2 source code).
Now, I found it's impossible to pass NULL to SDL_setenv, SDL refuses to apply that, therefore I pass empty string when trying to reset its value. This nevertheless worked on Windows, but I'm unsure if this is not a system-specific behavior. Must test if this works on Linux too. If not, then there's still a workaround: use SDL_GetNumAudioDrivers / SDL_GetAudioDriver(index) to iterate through the default driver list and try each one manually.

### DIFF
--- a/Engine/platform/base/sys_main.cpp
+++ b/Engine/platform/base/sys_main.cpp
@@ -99,22 +99,34 @@ void sys_renderer_set_output(const String &name)
 
 bool sys_audio_init(const String &driver_name)
 {
+    // IMPORTANT: we must use a combination of SDL_setenv and SDL_InitSubSystem
+    // here, and NOT use SDL_AudioInit, because SDL_AudioInit does not increment
+    // subsystem's reference count. Which in turn may cause problems down the
+    // way when initializing any additional SDL-based audio lib or plugin;
+    // at the very least - the mojoAl (OpenAL's implementation we're using).
     bool res = false;
+    // If user config contained a driver request, then apply one for a try
     if (!driver_name.IsEmpty())
+        SDL_setenv("SDL_AUDIODRIVER", driver_name.GetCStr(), 1);
+    const char *env_drv = SDL_getenv("SDL_AUDIODRIVER");
+    Debug::Printf("Requested audio driver: %s", env_drv ? env_drv : "default");
+    res = SDL_InitSubSystem(SDL_INIT_AUDIO) == 0;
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Tested on Linux, confirmed working if
- game config has an invalid audio driver name set;
- enviroment variable SDL_AUDIODRIVER is set to an invalid audio driver name;

in both cases SDL2 switches to default driver from its list, and mojoAL also initializes successfully.

PR #1749, opened 2022-08-06
Ags4 modern command line

jeancallisti:


ivan-mogilko:
Since this is related to the larger engine update, I may suggest to first of all clarify command line syntax rules. In the past the commands were added without much thinking about this, except we moved to having double dashes ("--") before most multicharacter option names following the unix standard. But some things may be inconsistent, both among each other or compared to the widely used conventions. This I think it may be a good idea to clarify that, especially for commands which have arguments, and multiple arguments.

ericoporto:
I have been thinking about this, and still don't have a good idea for the interface of passing the number of expected parameters for a flag, in theory I think nice flags should have just one parameter but unfortunately we do have some flags that take multiple parameters in ags. It could either use something like `^3` in the string or use a set of pair string,int.

Since this is for ags4, we could alternatively redesign the command line interface where it takes multiple parameters for one `--key`? Like either separate them without using spaces or use multiple flags or have a comma.

ivan-mogilko:
@jeancallisti hello, it's been almost 3 years as you posted this as a draft, but you have never explained what's the final goal, and whether you consider this change complete, or been planning to work more on this.
Unfortunately, I was too busy to investigate this draft thoroughly back in the time and, given the lack of updates, it fell out of my mind later.
Are you still interested in this work?

Perhaps some things may be cherry picked from here. If the changes can be separated into:
* change to the interface
* fixes to parsing
* using this cmd opts class to parse cmd line in the engine (right now it is only used in tools).

ivan-mogilko:
Given there's no response from the author, I'll close this.

PR #1759, opened 2022-09-02
Feature joystick

### LINKED ISSUE #1364: Add a gamepad API in the Engine
**Describe the problem**
AGS is a Game Engine, and most engines have it's own Joystick API. In AGS, using Joysticks requires using a plugin. There's at least 3 different plugins floating around for this, and things aren't great for crossplatform with them. 

If AGS had it's own gamepad API, then people would be able to make script modules with gamepad in mind.

**Suggested change**
I would like to propose adding a gamepad API directly in AGS.

Now, there are thousands of different joysticks and SDL2 joystick API has a lot of surface, so there's obvious different ways to do this. I would like to proppose supporting [SDL2 GameController AP](https://wiki.libsdl.org/CategoryGameController)I first and later figure out joystick

gamepad
- I have an example implementation here (code has leftovers, still POC): [github.com/ericoporto/ags/tree/feature-gamepad](https://github.com/ericoporto/ags/tree/feature-gamepad)
- Editor and Engine Build: [AGS-3.6.0.5-Gamepad.zip](https://github.com/adventuregamestudio/ags/files/6826907/AGS-3.6.0.5-Gamepad.zip)
- Example Project: [GamepadExample.zip](https://github.com/adventuregamestudio/ags/files/6826905/GamepadExample.zip)

<details>
  <summary>Here's my current header!</summary>
  
```AGS Script
enum eGamepad_Axis
{
    eGamepad_AxisInvalid=0,
    eGamepad_AxisLeftX,
    eGamepad_AxisLeftY,
    eGamepad_AxisRightX,
    eGamepad_AxisRightY,
    eGamepad_AxisTriggerLeft,
    eGamepad_AxisTriggerRight,
    eGamepad_AxisCount
};

enum eGamepad_Button
{
    eGamepad_ButtonInvalid=0,
    eGamepad_ButtonA,
    eGamepad_ButtonB,
    eGamepad_ButtonX,
    eGamepad_ButtonY,
    eGamepad_ButtonBack,
    eGamepad_ButtonGuide,
    eGamepad_ButtonStart,
    eGamepad_ButtonLeftStick,
    eGamepad_ButtonRightStick,
    eGamepad_ButtonLeftShoulder,
    eGamepad_ButtonRightShoulder,
    eGamepad_ButtonDpadUp,
    eGamepad_ButtonDpadDown,
    eGamepad_ButtonDpadLeft,
    eGamepad_ButtonDpadRight,
    eGamepad_ButtonCount
};

builtin managed struct Gamepad{  
  /// get number of connected joysticks
  import static int GetCount();  
  /// attempt to open a gamepad, returns null if it's not a valid gamepad
  import static Gamepad* Open(int index);  
  /// if joystick is a valid gamepad, get it's name
  import static String GetName(int index);  
  /// gamepad name
  import readonly attribute String Name;
  /// checks if gamepad is really connected
  import bool IsConnected();    
  /// checks if a gamepad button is pressed, including dpad.
  import bool IsButtonDown(eGamepad_Button button);    
  /// get gamepad axis or trigger, trigger only has positive values.
  import int GetAxis(eGamepad_Axis axis);
};
```
</details>

You can just copy this header in an empty script header in a regular Editor if you just wish to rebuild the Engine.

**References**
- https://love2d.org/wiki/love.joystick
- https://github.com/love2d/love/tree/master/src/modules/joystick
- https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.Gamepad.html
- https://github.com/gabomdq/SDL_GameControllerDB
- https://docs.godotengine.org/en/stable/tutorials/inputs/controllers_gamepads_joysticks.html
- https://manual.yoyogames.com/GameMaker_Language/GML_Reference/Game_Input/GamePad_Input/Gamepad_Input.htm

**Notes** 
- Which branch? Well on one hand ags4 is the branch for new features, oth I can imagine it on ags3 one day replacing agsjoy...
- if you don't want to rebuild the editor to test script APIs, you can just place this header in an empty script header, and just build the engine
- one probably never want to save anything about the joystick itself since the game may be loaded with no device later, or the device may have changed index, there's still a mystery in how to communicate this to developer using this API
- mapping to physical device, say button 3 is Triangle in PS4 gamepad, but it's Y in Xbox One gamepad, and then Square in PS5, just hypothetically, if the developer wants to present an icon with the button name, do we want to have this physical mapping solved in the Engine or rely on it being scripted?
- mapping buttons to actions is probably a separate problem that could be addressed later, perhaps with some input struct which could then be linked to functions, but this is a different problem that is not going to be solved here.
- which aspects of the engine need to be adapted? Do I need to provide a way for characters to walk by pressing arrows instead of pathfinding?

Also, thanks for the [wiki entry](https://github.com/adventuregamestudio/ags/wiki/A-quick-guide-to-Script-API), it helped immensely figuring out what was needed to do in my code.

**Post in the forums**
https://www.adventuregamestudio.co.uk/forums/index.php?topic=59945.0

ericoporto:
fix #1364  | [Forum thread](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59945.0)

Adds a Joystick API with support to SDL Game Controller too (named Gamepad in Script API here).

- Joysticks are managed internally in the engine: you can't open or close connection to joysticks from the script api.
- Missing a better integration with the engine and a way to set the buttons for things like skipping text dialogs with the joystick. 
- Haptics (Rumble) is not added here.
- Navigating default AGS menus with joystick/gamepad is not intended to be supported because it would require a new concept of selection.
- Also I have no idea what to do regarding save files.

### Joystick Static Methods

**`Joystick.JoystickCount`**
```AGS Script
static readonly int Joystick.JoystickCount
```
Get the number of connected joysticks. No joysticks should return 0!

---

**`Joystick.Get`**
```AGS Script
static readonly Joystick* Joystick.Joysticks[int index]
```

Get a joystick from it's index in the internal engine list of joysticks.

---

### Joystick Instance Attributes and Methods

**`Joystick.IsConnected`**
```AGS Script
readonly bool Joystick.IsConnected
```
True if joystick is really connected.

---

**`Joystick.Name`**
```AGS Script
readonly String Joystick.Name
```
joystick name.

---

**`Joystick.IsButtonDown`**
```AGS Script
bool Joystick.IsButtonDown(int button)
```
checks if a joystick button is pressed, by index. DPad is usually mapped separately as a hat.

---

**`Joystick.GetAxis`**
```AGS Script
bool Joystick.GetAxis(int axis, optional float deadzone)
```
get a joystick axis or trigger, trigger only has positive values, by axis number. Values varies from -1.0 to 1.0 for axis, and 0.0 to 1.0 for triggers.

---

**`Joystick.GetHat`**
```AGS Script
eJoystick_Hat Joystick.GetHat(int hat)
```
returns hat position as enum.

---

**`Joystick.AxisCount`**
```AGS Script
readonly int Joystick.AxisCount
```
get the raw number of axis in the joystick.

---

**`Joystick.ButtonCount`**
```AGS Script
readonly int Joystick.ButtonCount
```
get the raw number of buttons in the joystick.

---

**`Joystick.HatCount`**
```AGS Script
readonly int Joystick.HatCount
```
get the raw number of hats in the joystick.

---

**`Joystick.IsGamepad`**
```AGS Script
readonly bool Joystick.IsGamepad
```
True if joystick is a valid gamepad connected - this means SDL2 recognized it as a valid GameController and has successfully mapped it's buttons to an Xbox360 gamepad.

---

**`Joystick.IsGamepadButtonDown`**
```AGS Script
bool Joystick.IsGamepadButtonDown(eGamepad_Button button)
```
checks if a gamepad button is pressed, including dpad.

Possible buttons:

- `eGamepad_ButtonA`
- `eGamepad_ButtonB`
- `eGamepad_ButtonX`
- `eGamepad_ButtonY`
- `eGamepad_ButtonBack`
- `eGamepad_ButtonGuide`
- `eGamepad_ButtonStart`
- `eGamepad_ButtonLeftStick`
- `eGamepad_ButtonRightStick`
- `eGamepad_ButtonLeftShoulder`
- `eGamepad_ButtonRightShoulder`
- `eGamepad_ButtonDpadUp`
- `eGamepad_ButtonDpadDown`
- `eGamepad_ButtonDpadLeft`
- `eGamepad_ButtonDpadRight`

---

**`Joystick.GetGamepadAxis`**
```AGS Script
float Joystick.GetGamepadAxis(eGamepad_Axis axis, optional float deadzone)
```
get gamepad axis or trigger, trigger only has positive values. Values varies from -1.0 to 1.0 for axis, and 0.0 to 1.0 for triggers.

You can optionally pass an additional parameter to use as deadzone. If an axis absolute value is smaller than the value of deadzone, it will return 0.0. Default value is AXIS_DEFAULT_DEADZONE, which is for now 0.125, use the name if you need a number.

Possible axis and triggers:

- `eGamepad_AxisLeftX`
- `eGamepad_AxisLeftY`
- `eGamepad_AxisRightX`
- `eGamepad_AxisRightY`
- `eGamepad_AxisTriggerLeft`
- `eGamepad_AxisTriggerRight`

---

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -64,6 +64,10 @@
 
 #define SCR_NO_VALUE   31998   // $AUTOCOMPLETEIGNORE$
 
+#ifdef SCRIPT_API_v399
+  #define AXIS_DEFAULT_DEADZONE 0.125
+#endif
+
 enum bool {
   false = 0,
   true = 1
@@ -435,7 +439,61 @@ enum InputType
   // 0x0100... is used internally to define Timeout (may be worked around in the future)
   eInputKeyboard = 0x02000000,
   eInputMouse    = 0x04000000,
+#endif
+#ifdef SCRIPT_API_v399
+  eInputGamepad  = 0x08000000,
+#endif
+#ifdef SCRIPT_API_v36026
   eInputAny      = 0xFF000000
+};
+#endif
+
+#ifdef SCRIPT_API_v399
+enum eGamepad_Axis
+{
+    eGamepad_AxisInvalid=0,
+    eGamepad_AxisLeftX,
+    eGamepad_AxisLeftY,
+    eGamepad_AxisRightX,
+    eGamepad_AxisRightY,
+    eGamepad_AxisTriggerLeft,
+    eGamepad_AxisTriggerRight,
+    eGamepad_AxisCount
+};
+
+enum eGamepad_Button
+{
+    eGamepad_ButtonInvalid=0,
+    eGamepad_ButtonA,
+    eGamepad_ButtonB,
+    eGamepad_ButtonX,
+    eGamepad_ButtonY,
+    eGamepad_ButtonBack,
+    eGamepad_ButtonGuide,
+    eGamepad_ButtonStart,
+    eGamepad_ButtonLeftStick,
+    eGamepad_ButtonRightStick,
+    eGamepad_ButtonLeftShoulder,
+    eGamepad_ButtonRightShoulder,
+    eGamepad_ButtonDpadUp,
+    eGamepad_ButtonDpadDown,
+    eGamepad_ButtonDpadLeft,
+    eGamepad_ButtonDpadRight,
+    eGamepad_ButtonCount
+};
+
+enum eJoystick_Hat
+{
+    eJoystick_HatInvalid = -1,
+    eJoystick_HatCentered = 0,
+    eJoystick_HatUp = 0x01,
+    eJoystick_HatRight = 0x02,
+    eJoystick_HatDown = 0x04,
+    eJoystick_HatLeft = 0x08,
+    eJoystick_HatRightUp = 0x03,
+    eJoystick_HatRightDown = 0x06,
+    eJoystick_HatLeftUp = 0x09,
+    eJoystick_HatLeftDown = 0x0C,
 };
 #endif
 
@@ -2471,6 +2529,36 @@ builtin struct Screen {
 };
 #endif
 
+#ifdef SCRIPT_API_v399
+builtin managed struct Joystick {
+  /// get number of connected joysticks.
+  import static readonly attribute int JoystickCount; // $AUTOCOMPLETESTATICONLY$
+  /// get a connected joystick by index or null on invalid index.
+  import static readonly attribute Joystick *Joysticks[]; // $AUTOCOMPLETESTATICONLY$
+  /// gets joystick name
+  import readonly attribute String Name;
+  /// checks if joystick is really connected
+  import readonly attribute bool IsConnected;
+  /// checks if joystick is a valid gamepad
+  import readonly attribute bool IsGamepad;
+  /// checks if a gamepad button is pressed, including dpad.
+  import bool IsGamepadButtonDown(eGamepad_Button button);
+  /// get gamepad axis or trigger, trigger only has positive values.
+  import float GetGamepadAxis(eGamepad_Axis axis, float dead_zone = AXIS_DEFAULT_DEADZONE); 
+  /// get joystick axis or trigger, by index
+  import float GetAxis(int axis, float dead_zone = AXIS_DEFAULT_DEADZONE);
+  /// checks if a joystick button is pressed.
+  import bool IsButtonDown(int button);
+  /// gets the direction a dpad from the joystick is pointing to.
+  import eJoystick_Hat GetHat(int hat);
+  /// get axis count from the joystick
+  import readonly attribute int AxisCount;
+  /// get button count from the joystick
+  import readonly attribute int ButtonCount;
+  /// get hat count from the joystick
+  import readonly attribute int HatCount;
+};
+#endif
 
 
 import readonly Character *player;
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -98,6 +98,8 @@ target_sources(engine
     ac/dynobj/scriptdynamicsprite.h
     ac/dynobj/scriptfile.cpp
     ac/dynobj/scriptfile.h
+    ac/dynobj/scriptjoystick.cpp
+    ac/dynobj/scriptjoystick.h
     ac/dynobj/scriptgui.h
     ac/dynobj/scripthotspot.h
     ac/dynobj/scriptinvitem.h
@@ -127,6 +129,8 @@ target_sources(engine
     ac/file.h
     ac/game.cpp
     ac/game.h
+    ac/joystick.cpp
+    ac/joystick.h
     ac/gamesetup.cpp
     ac/gamesetup.h
     ac/gamestate.cpp
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -47,6 +47,7 @@
 #include "ac/mouse.h"
 #include "media/audio/audio_system.h"
 #include "ac/timer.h"
+#include "joystick.h"
 
 using namespace AGS::Common;
 using namespace AGS::Common::BitmapHelper;
@@ -241,6 +242,21 @@ bool display_check_user_input(int skip)
             if (skip & SKIP_MOUSECLICK && !play.IsIgnoringInput())
             {
                 play.SetWaitSkipResult(SKIP_MOUSECLICK, mbut);
+                return true; // stop display
+            }
+        }
+        else if (type == kInputGamepad)
+        {
+            GamepadInput gbut;
+            if (!run_service_gamepad_controls(gbut) || play.fast_forward)
+                continue;
+            eAGSGamepad_Button gbn = gbut.Button;
+            if (check_skip_cutscene_gamepad(gbn))
+                return true;
+            if (skip & SKIP_GAMEPAD && !play.IsIgnoringInput() &&
+                    is_default_gamepad_skip_button_pressed(gbn))
+            {
+                play.SetWaitSkipResult(SKIP_GAMEPAD, gbn);
                 return true; // stop display
             }
         }
--- a/Engine/ac/dynobj/cc_serializer.cpp
+++ b/Engine/ac/dynobj/cc_serializer.cpp
@@ -26,6 +26,7 @@
 #include "debug/debug_log.h"
 #include "plugin/plugin_engine.h"
 #include "util/memorystream.h"
+#include "scriptjoystick.h"
 
 using namespace AGS::Common;
 
@@ -158,6 +159,12 @@ void AGSDeSerializer::Unserialize(int index, const char *objectType, const char 
     else if (strcmp(objectType, "AudioClip") == 0)
     {
         ccDynamicAudioClip.Unserialize(index, &mems, data_sz);
+    }
+    else if (strcmp(objectType, "Joystick") == 0)
+    {
+        // joysticks cannot be restored properly, any operation will fail
+        auto *scj = new ScriptJoystick();
+        ccRegisterUnserializedObject(index, scj, scj);
     }
     else
     {
--- a/Engine/ac/dynobj/scriptjoystick.cpp
+++ b/Engine/ac/dynobj/scriptjoystick.cpp
@@ -1,0 +1,44 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+
+#include "ac/dynobj/scriptjoystick.h"
+#include "ac/joystick.h"
+#include "util/stream.h"
+#include "dynobj_manager.h"
+
+using namespace AGS::Common;
+
+ScriptJoystick::ScriptJoystick(int id) : _id(id) {}
+
+const char *ScriptJoystick::GetType() {
+    return "Joystick";
+}
+
+int ScriptJoystick::Dispose(void *address, bool force) {
+    delete this;
+    return 1;
+}
+
+size_t ScriptJoystick::CalcSerializeSize(void* /*address*/){
+    return 0;
+}
+
+void ScriptJoystick::Serialize(void *address, AGS::Common::Stream *out) {
+
+}
+
+void ScriptJoystick::Unserialize(int index, AGS::Common::Stream *in, size_t data_sz) {
+    this->Invalidate();
+    ccRegisterUnserializedObject(index, this, this);
+}
--- a/Engine/ac/dynobj/scriptjoystick.h
+++ b/Engine/ac/dynobj/scriptjoystick.h
@@ -1,0 +1,46 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+
+#ifndef __AC_SCRIPTJOYSTICK_H
+#define __AC_SCRIPTJOYSTICK_H
+
+#include "ac/dynobj/cc_agsdynamicobject.h"
+#include "SDL_gamecontroller.h"
+
+struct ScriptJoystick final : AGSCCDynamicObject {
+public:
+    ScriptJoystick(int id = -1);
+    // Get gamepad index; negative means the gamepad was deleted
+    int GetID() const { return _id; }
+    void SetID(int id) { _id = id; }
+    // Reset gamepad index to indicate that this reference is no longer valid
+    void Invalidate() { _id = -1; }
+    inline bool IsInvalid() const {return _id == -1; }
+
+    int Dispose(void *address, bool force) override;
+    const char *GetType() override;
+    void Unserialize(int index, AGS::Common::Stream *in, size_t data_sz) override;
+
+
+protected:
+    // Calculate and return required space for serialization, in bytes
+    size_t CalcSerializeSize(void *address) override;
+    // Write object data into the provided stream
+    void Serialize(void *address, AGS::Common::Stream *out) override;
+
+private:
+    int _id = -1;
+};
+
+#endif // __AC_SCRIPTJOYSTICK_H
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1112,6 +1112,17 @@ bool check_skip_cutscene_mclick(int mbut)
     return false;
 }
 
+bool check_skip_cutscene_gamepad(int gbut)
+{
+    CutsceneSkipStyle skip = get_cutscene_skipstyle();
+    if (gbut>0 && skip == eSkipSceneAnyKey)
+    {
+        start_skipping_cutscene();
+        return true;
+    }
+    return false;
+}
+
 // Helper functions used by StartCutscene/EndCutscene, but also
 // by SkipUntilCharacterStops
 void initialize_skippable_cutscene() {
--- a/Engine/ac/game.h
+++ b/Engine/ac/game.h
@@ -154,6 +154,7 @@ CutsceneSkipStyle get_cutscene_skipstyle();
 void start_skipping_cutscene ();
 bool check_skip_cutscene_keypress(int kgn);
 bool check_skip_cutscene_mclick(int mbut);
+bool check_skip_cutscene_gamepad(int gbut);
 void initialize_skippable_cutscene();
 void stop_fast_forwarding();
 
--- a/Engine/ac/joystick.cpp
+++ b/Engine/ac/joystick.cpp
@@ -1,0 +1,450 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+#include <algorithm>
+#include <vector>
+#include "ac/string.h"
+#include "ac/gamestructdefines.h"
+#include "ac/joystick.h"
+#include "SDL_joystick.h"
+#include "SDL_gamecontroller.h"
+#include "debug/debug_log.h"
+#include "script/script_api.h"
+#include "ac/dynobj/dynobj_manager.h"
+
+// clamps to -1.0 to 1.0, taking into account a dead-zone from 0.0
+float gamepad_clamp_val(float val, float dead_zone)
+{
+    dead_zone = std::max(dead_zone, 0.01f); // ignore negative values, but force minimal deadzone
+
+    if (val < 0.0f) {
+        val = val < -dead_zone ? val : 0.0f;
+    } else {
+        val = val > dead_zone ? val : 0.0f;
+    }
+
+    if (val < -0.99f) return -1.0f;
+    if (val > 0.99f) return 1.0f;
+
+    return val;
+}
+
+float axis_to_float_with_deadzone(int axis_val_int, float dead_zone) {
+    float axis_val = axis_val_int < 0 ? static_cast<float>(axis_val_int) / 32768.0f : static_cast<float>(axis_val_int) / 32767.0f;
+    return gamepad_clamp_val(axis_val, dead_zone);
+}
+
+SDL_GameControllerAxis Gamepad_Axis_AGStoSDL( eAGSGamepad_Axis ags_axis) {
+    switch (ags_axis)
+    {
+        case eAGSGamepad_AxisInvalid: return SDL_CONTROLLER_AXIS_INVALID;
+        case eAGSGamepad_AxisLeftX: return SDL_CONTROLLER_AXIS_LEFTX;
+        case eAGSGamepad_AxisLeftY: return SDL_CONTROLLER_AXIS_LEFTY;
+        case eAGSGamepad_AxisRightX: return SDL_CONTROLLER_AXIS_RIGHTX;
+        case eAGSGamepad_AxisRightY: return SDL_CONTROLLER_AXIS_RIGHTY;
+        case eAGSGamepad_AxisTriggerLeft: return SDL_CONTROLLER_AXIS_TRIGGERLEFT;
+        case eAGSGamepad_AxisTriggerRight: return SDL_CONTROLLER_AXIS_TRIGGERRIGHT;
+    }
+    return SDL_CONTROLLER_AXIS_INVALID;
+}
+
+SDL_GameControllerButton Gamepad_Button_AGStoSDL(eAGSGamepad_Button ags_button) {
+    switch (ags_button)
+    {
+        case eAGSGamepad_ButtonInvalid: return SDL_CONTROLLER_BUTTON_INVALID;
+        case eAGSGamepad_ButtonA: return SDL_CONTROLLER_BUTTON_A;
+        case eAGSGamepad_ButtonB: return SDL_CONTROLLER_BUTTON_B;
+        case eAGSGamepad_ButtonX: return SDL_CONTROLLER_BUTTON_X;
+        case eAGSGamepad_ButtonY: return SDL_CONTROLLER_BUTTON_Y;
+        case eAGSGamepad_ButtonBack: return SDL_CONTROLLER_BUTTON_BACK;
+        case eAGSGamepad_ButtonGuide: return SDL_CONTROLLER_BUTTON_GUIDE;
+        case eAGSGamepad_ButtonStart: return SDL_CONTROLLER_BUTTON_START;
+        case eAGSGamepad_ButtonLeftStick: return  SDL_CONTROLLER_BUTTON_LEFTSTICK;
+        case eAGSGamepad_ButtonRightStick: return  SDL_CONTROLLER_BUTTON_RIGHTSTICK;
+        case eAGSGamepad_ButtonLeftShoulder: return SDL_CONTROLLER_BUTTON_LEFTSHOULDER;
+        case eAGSGamepad_ButtonRightShoulder: return SDL_CONTROLLER_BUTTON_RIGHTSHOULDER;
+        case eAGSGamepad_ButtonDpadUp: return  SDL_CONTROLLER_BUTTON_DPAD_UP;
+        case eAGSGamepad_ButtonDpadDown: return  SDL_CONTROLLER_BUTTON_DPAD_DOWN;
+        case eAGSGamepad_ButtonDpadLeft: return  SDL_CONTROLLER_BUTTON_DPAD_LEFT;
+        case eAGSGamepad_ButtonDpadRight: return  SDL_CONTROLLER_BUTTON_DPAD_RIGHT;
+    }
+    return SDL_CONTROLLER_BUTTON_INVALID;
+}
+
+eAGSJoystick_Hat Joystick_Hat_SDLtoAGS(uint8_t sdl_hat) {
+    switch (sdl_hat)
+    {
+        case SDL_HAT_CENTERED: return eAGSJoystick_HatCentered;
+        case SDL_HAT_UP: return eAGSJoystick_HatUp;
+        case SDL_HAT_RIGHT: return eAGSJoystick_HatRight;
+        case SDL_HAT_DOWN: return eAGSJoystick_HatDown;
+        case SDL_HAT_LEFT: return eAGSJoystick_HatLeft;
+        case SDL_HAT_RIGHTUP: return eAGSJoystick_HatRightUp;
+        case SDL_HAT_RIGHTDOWN: return eAGSJoystick_HatRightDown;
+        case SDL_HAT_LEFTUP: return eAGSJoystick_HatLeftUp;
+        case SDL_HAT_LEFTDOWN: return eAGSJoystick_HatLeftDown;
+    }
+    return eAGSJoystick_HatCentered;
+}
+
+uint8_t Joystick_Hat_AGStoSDL(eAGSJoystick_Hat ags_hat)
+{
+    switch (ags_hat)
+    {
+        case eAGSJoystick_HatCentered: return SDL_HAT_CENTERED;
+        case eAGSJoystick_HatUp: return SDL_HAT_UP;
+        case eAGSJoystick_HatRight: return SDL_HAT_RIGHT;
+        case eAGSJoystick_HatDown: return SDL_HAT_DOWN;
+        case eAGSJoystick_HatLeft: return  SDL_HAT_LEFT;
+        case eAGSJoystick_HatRightUp: return  SDL_HAT_RIGHTUP;
+        case eAGSJoystick_HatRightDown: return SDL_HAT_RIGHTDOWN;
+        case eAGSJoystick_HatLeftUp: return SDL_HAT_LEFTUP;
+        case eAGSJoystick_HatLeftDown: return SDL_HAT_LEFTDOWN;
+    }
+    return SDL_HAT_CENTERED;
+}
+
+eAGSGamepad_Button Gamepad_Button_SDLtoAGS(SDL_GameControllerButton sdl_button) {
+    switch (sdl_button)
+    {
+        case SDL_CONTROLLER_BUTTON_INVALID: return eAGSGamepad_ButtonInvalid;
+        case SDL_CONTROLLER_BUTTON_A: return eAGSGamepad_ButtonA;
+        case SDL_CONTROLLER_BUTTON_B: return eAGSGamepad_ButtonB;
+        case SDL_CONTROLLER_BUTTON_X: return eAGSGamepad_ButtonX;
+        case SDL_CONTROLLER_BUTTON_Y: return eAGSGamepad_ButtonY;
+        case SDL_CONTROLLER_BUTTON_BACK: return eAGSGamepad_ButtonBack;
+        case SDL_CONTROLLER_BUTTON_GUIDE: return eAGSGamepad_ButtonGuide;
+        case SDL_CONTROLLER_BUTTON_START: return eAGSGamepad_ButtonStart;
+        case SDL_CONTROLLER_BUTTON_LEFTSTICK: return eAGSGamepad_ButtonLeftStick;
+        case SDL_CONTROLLER_BUTTON_RIGHTSTICK: return eAGSGamepad_ButtonRightStick;
+        case SDL_CONTROLLER_BUTTON_LEFTSHOULDER: return eAGSGamepad_ButtonLeftShoulder;
+        case SDL_CONTROLLER_BUTTON_RIGHTSHOULDER: return eAGSGamepad_ButtonRightShoulder;
+        case SDL_CONTROLLER_BUTTON_DPAD_UP: return eAGSGamepad_ButtonDpadUp;
+        case SDL_CONTROLLER_BUTTON_DPAD_DOWN: return eAGSGamepad_ButtonDpadDown;
+        case SDL_CONTROLLER_BUTTON_DPAD_LEFT: return eAGSGamepad_ButtonDpadLeft;
+        case SDL_CONTROLLER_BUTTON_DPAD_RIGHT: return eAGSGamepad_ButtonDpadRight;
+    }
+    return eAGSGamepad_ButtonInvalid;
+}
+
+struct JoystickImpl {
+    SDL_GameController* sdlGameController = nullptr;
+    SDL_Joystick* sdlJoystick = nullptr;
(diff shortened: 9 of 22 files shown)


ivan-mogilko:
> Missing a better integration with the engine and a way to set the buttons for things like skipping text dialogs with the joystick.

By the way, this is a very interesting question in regards to script API. At first I thought that maybe you could "pack" any input type in existing variables, but then realized that a user may want to have separate skip variables per each device too. For example, right now you can tell which key to skip with, but not which mouse button to skip with. And even with keys, you can only tell one key (even though that might be enough).

So, the most straightforward approach is to keep adding "skip key" kind of option per each new input device. But I am currently wondering, if there's a better way to design this.

ericoporto:
Yeah, there's also a problem that a device may be added and then be gone - disconnected or battery ended when wireless. But I think we can just use: joystick index 1 -> player input 1 mapping.

There is an idea I am looking into for an "action object", and you add inputs that trigger such action. So, in an action game you can wire things to say an action called "Jump", and later you can assign inputs either in a pre-defined way or remap in-game. There would be a new entry in the project tree for actions where you could create new ones, and use the properties to assign inputs/change name and event to create a triggered function in global script.

ericoporto:
> So, the most straightforward approach is to keep adding "skip key" kind of option per each new input device. But I am currently wondering, if there's a better way to design this.

I think this is the best idea to add joystick/gamepad input for now. Other designs I could think end up being a work of their own.

AlanDrake:
Tomorrow I'm gonna give this a try.

Does it catch all inputs if I initialize the Joystick with a zero index?
Can I use both XInput and DInput devices?

Scripting a VirtualController module is pretty trivial, I already did so to emulate mouse with keyboard, so interfacing with a simple API is preferable to any other implementation, in my opinion.

ericoporto:
> Does it catch all inputs if I initialize the Joystick with a zero index?

I have no idea what this means.

> Can I use both XInput and DInput devices?

It should not matter at all.

Right now, if the joystick is a gamepad (GameController in SDL2 backend, means can be mapped to Xbox360 like controller), it has hardcoded buttons to skip dialog. If it's not, there is no buttons at all for skipping dialog, yet. Need to figure out an API so you can set these.

AlanDrake:
I see eGamepad_ButtonInvalid=0 and eGamepad_AxisInvalid=0, but all buttons and axes start from zero, thus all enums are incorrectly offset by 1.

What I meant earlier was if there was a way to not connect a specific gamepad and just pretend any pad input was from a single unified controller.

ericoporto:
> What I meant earlier was if there was a way to not connect a specific gamepad and just pretend any pad input was from a single unified controller.

I still doesn't understand much of what that means, but all the most popular joysticks have a mapping to a xbox360 like controller in SDL2, called a GameController, I find this naming confusing and used Gamepad instead. You can use an [external program](https://generalarcade.com/gamepadtool/) to change this if you want, but I did not want to deal with this, and instead I wanted to enlarge this mapping concept later, to include other inputs, using an [Actions Map](https://www.adventuregamestudio.co.uk/forums/index.php?topic=60235.0), which is my interpretation of [Valve's Action Set](https://partner.steamgames.com/doc/features/steam_controller/action_set_layers) transplanted to AGS.

> I see eGamepad_ButtonInvalid=0 and eGamepad_AxisInvalid=0, but all buttons and axes start from zero, thus all enums are incorrectly offset by 1

I still am not sure here which way to go, but the idea is 0 is false, and true means some button or axe in Gamepad land, so it's easy to test skip reason (`if(button)`). 

This all actually lingers on figuring what to do with `SkipSpeechStyle` , `VideoSkipStyle` and `CutsceneSkipType`, which need to accommodate:

- Somehow not break existing stuff, probably with a new API and internally mapping the old stuff to the new API, but haven't figured what the new API would be
- Joystick and Gamepad + Keyboard + Mouse , ...
- Some way to differentiate what is Joystick 1, 2, ...
- Which key to use in Joystick and Gamepad

AlanDrake:
In regard to skip styles, I think AGS should just provide a DIY way to let developers handle the decision themselves. I much enjoyed the introduction of eSkipScriptOnly because with that I could implement a hold-to-skip system for cutscenes.
It could be AGS calling a custom function, or some other way to trigger the skip types via script. Much simpler and more flexible all around.

> * Somehow not break existing stuff, probably with a new API and internally mapping the old stuff to the new API, but haven't figured what the new API would be
> * Joystick and Gamepad + Keyboard + Mouse , ...
> * Some way to differentiate what is Joystick 1, 2, ... - or not, always assume these only matters in single player context.
> * Which key to use in Joystick and Gamepad

All of this should be handled via script by the game developer. You never know who wants to remap which button and how, XInput and DInput have different mappings, differing glyphs... it's better to have a simple API like this without too many assumptions and let a module do the rest.

Anyway, I wonder if it's a good idea to define a eGamepad_Button enum like that... I bet our users will come complain the moment eGamepad_ButtonDpadUp is not working because on XInput the dpad is read from the Hat.
.

ericoporto:
On steam you can do the remapping in steam directly, so for games released on steam it's very useful - the mappings are passed through an environment variable behind the scenes.

> eGamepad_ButtonDpadUp is not working

Wait, can you pass the exact steps to reproduce this? I need the exact controller, if this is true, this has to be fixed upstream in SDL2.

ivan-mogilko:
@ericoporto could you please post a full list of added API in the first post? Which functions, properties and callbacks were added. That would make it easier to understand what this pr is providing.

> This all actually lingers on figuring what to do with SkipSpeechStyle , VideoSkipStyle and CutsceneSkipType

All of these old skip styles should be deprecated, because they are not consistent and are difficult to customize and expand.

There may be two approaches here. These approaches do not conflict with each other, but are rather "quick and high-level" and "advanced and low-level" ones:

1) A variable or set of variables that let you customize any combination of inputs. In ags4 this may be done using the Action object that you proposed on forums earlier. Since that would take time to design and implement, this may be preceded by a more limited solution, such as a variable storing InputType flags, and maybe a set of variables per input type, defining exact keys or buttons.
2) Provide means for user to precisely customize the preferred behavior in script callbacks, as @AlanDrake mentioned. If there's not enough callbacks - then add missing ones. For example, right away there's no callbacks to do continuous input tests during display box, and video playback.

AlanDrake:
> Wait, can you pass the exact steps to reproduce this?

Ah sorry, I started from the older joystick project, so I hadn't noticed I was supposed to use the "Gamepad" variant of the functions!

Yes, now everything makes more sense and works as expected.

Anyway, I'm pleasantly suprised that my DS4 works through bluetooth right off the bat as a DirectInput device. It's only of recent that I've seen it work on games, usually was ignored unless connected through USB or thorugh DS4Windows.

ericoporto:
> could you please post a full list of added API in the first post?

Added above! No callbacks are added yet - I don't know how the event callback should be yet.

> There may be two approaches here.

The 2 approach you propose seems fine, but it requires the additional callbacks and I really would like to deal somehow with the Display box. Also reminder that any callback won't be able to skip a blocking regular Say textbox, so you need to poll using `repeatedly_execute_always`, it would be nice if a Say textbox was dealt the same way as a Display box... I am dealing with them in hardcoded way, but missing a config.

And yes, in AGS4 I would like to pursuit my mentioned Actions proposal.

> DS4Windows

If you see a duplicated controller with a certain config for this, there's [a fix for it](https://github.com/libsdl-org/SDL/issues/6167#issuecomment-1240291644) in the next SDL2 release.

ivan-mogilko:
>  it would be nice if a Say textbox was dealt the same way as a Display box.

could you clarify this, please? I'm not sure what is the "Say textbox" you are refering to.

ericoporto:
The one that appears when using a `player.Say("This text box")`, in Sierra-style. May not be a textbox in other speech modes. :0

ivan-mogilko:
One thought I have is, possibly the most straightforward initial approach is to skip speech and display boxes whenever they are set to be skipped by either keyboard or mouse. The reasoning behind this is that when using gamepad/joystick, it's highly unlikely that player will be expected to use anything else (with mouse/keyboard there are still options of how the game is played). So if skip is set to e.g. skip by mouse, then it's likely okay to also skip by a gamepad.

Lately we may look into adding customization options, if that's necessary.

ericoporto:
The API of the joystick alone I think it's almost there, but the skipping of stuff is something I don't know yet. Also I originally had thought about having the Gamepad/Joystick as separate thing in the Skipping because the button numbers are different, but I am not sure on this.

AlanDrake:
I think we should retarget this to ags4 branch (once you're happy with the api)
We can merge it quicker there without having to sweat about uncertainties about skipping and other potential backcompat issues.

Also, the skipping coud come as a separate PR, I think. Those behaviors are not stricly part of the API, it's inner game logic from ags.

ericoporto:
About the API, the way I attempted here is similar to what I saw people do (and what I did) in plugins.

One thing about it in general that I am not sure is if the joystick/gamepad should be opened manually like I did, or if AGS itself should have some polling of it's own that simply opens and maintains it's own joystick/gamepad objects internally and then we expose this to the game in some way.

About retargeting to ags4, maybe the logging PR could be there, but this one I don't think it's that close to be done because of the skipping.

ivan-mogilko:
I'm sorry that I did not commented earlier, I was planning to, but then got distracted by couple of new serious bugs in 3.6.0; and because of these the development is sort of "stuck" in 3.6.0 longer than anticipated (again).

I agree that the skipping, and other built-in engine behavior, may be added separately. There's [my comment](https://github.com/adventuregamestudio/ags/pull/1759#issuecomment-1242999455) about existing issues above; and there's definitely a number of things to consider and possibly redesign.

I also recall my [proposal was](https://github.com/adventuregamestudio/ags/pull/1759#issuecomment-1244001004), as a starting support, for simplicity sake, to add a rule that if a skipping is set to *any* player's input (key or mouse button), then make it work for any joystick / gamepad button as well. This is not very configurable, but at least will let players to advance the game in case they use solely the joystick device.

Anyway, my question right now is this. I recall @ericoporto telling me, that one game developer (or maybe there were more?) was requesting this new device support for a 3.6.0 game. How critical this feature request is? Is there a possibility to support these devices with plugins, and are there any things that cannot be done with plugins?

I see there's a possibility to add the new device support as a special 3.6.X update, with only the minimal built-in behavior necessary to make the game playable.

If this is not critical, and may be worked around for now (e.g. with plugins), then perhaps it's best to target ags4 and focus on a proper support there.

So, I'd like to know now, because I was ready to create a separate branch for developing and testing a 3.6.X update (since 3.6.0 is still in fixing in the master branch), and I've created a milestone for 3.6.X in the past (https://github.com/adventuregamestudio/ags/milestone/13). If that's not necessary, then I won't.

---

Unfortunately, I was not able to check the presented API itself, so cannot make much comments on it now. But I peeked an interesting question about whether to let user to open device manually or do automatic polling. My personal opinion at the moment is that engine should provide user with a list of available devices, **unless** "opening" them will have an impact on performance or require significant resources (in which case they should only be "opened" when user really needs them). I also think that it would be better to make all devices handled similarly. Which reminds me, that there's no way, for example, to know if mouse or keyboard was connected or disconnected at runtime.

ivan-mogilko:
> I am curious if there's any way to change the target branch from admin controls or I need to close this and open a new PR.

There's, but your own branch is based on master, which may cause problems, so it's best to remake it based on top of ags4.

ericoporto:
Uhm, there's a SDL3 in the works and they also renamed game controller as gamepad - see [migration guide](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md). This can probably be done in sdl3 time 😬

ivan-mogilko:
I would suggest to split the big commit "add joystick and gamepad support" into at least 2 commits:
1) The one that adds joystick API implementation;
2) The one that does additional built-in behavior, such as wait skipping.

(similar was suggested by @AlanDrake earlier, I think)

Other than that, this PR is still targetting master branch, should it be retargeted to ags4 now?

ericoporto:
@ivan-mogilko thanks for the comments, I will adjust the code to be like you mentioned.

About ags4, I plan to open a new PR and close this one.

You asked me to make the Joystick management in engine though, so I am experimenting with this idea here: https://github.com/ericoporto/ags/tree/add-joystick

You can see Joystick.Open is removed. As soon as I remember how to add a new event for joystick connection I will do so, so that one can have `on_joystick_event(Joystick* joy, JoystickEventType event, int data)` or something like it. The connection event though needs to be non-blocking, so not sure yet if I will attempt to smoosh together.

For now there is a Joystick.Joysticks[] and a Joystick.Length - don't know if there is a way I could do Joystick.Joysticks.Length in ags4.

ivan-mogilko:
> For now there is a Joystick.Joysticks[] and a Joystick.Length - don't know if there is a way I could do Joystick.Joysticks.Length in ags4.

That is not possible, because indexed attributes (`Joysticks[]`) return not the array, but individual elements. So you would have to add another attribute for `JoystickCount`, similar to how there's Game.CharacterCount, and so forth.

ericoporto:
> I also wanted to add notes on how to separate Joystick struct into the script object

It's fine to add in the original issue. :)

ericoporto:
> no space between `if` and the condition (`if(condition)` vs `if (condition)`);

@ivan-mogilko  I don't understand this, what is the one that I should use?

ivan-mogilko:
> > no space between `if` and the condition (`if(condition)` vs `if (condition)`);
> 
> @ivan-mogilko I don't understand this, what is the one that I should use?
> 

To to have a space, this is how we write all the new code.

ericoporto:
@ivan-mogilko I am trying to figure it out [what you wrote here](https://github.com/adventuregamestudio/ags/issues/1364#issuecomment-1507432268), but the viewports or the overlays case seem to have lots of specifics to their case, I can't quite get much of the idea mentioned there... But I don't like the idea of invalidating the references that you mention there, a disconnection event can happen at any moment, this would make the joystick script code be full of null checks, and even then it could fail! I think a better approach is to never invalidate the reference and instead just let things fail silently if needed and have a `Joystick.Valid()` or `Joystick.Connected()` be used for this check if needed.

ivan-mogilko:
> But I don't like the idea of invalidating the references that you mention there, a disconnection event can happen at any moment, this would make the joystick script code be full of null checks, and even then it could fail! I think a better approach is to never invalidate the reference and instead just let things fail silently if needed and have a Joystick.Valid() or Joystick.Connected() be used for this check if needed.

Invalidation does not make references null, there's no way to do this anyway (engine cannot set a user variable to null). Invalidation is for setting an internal state of a script object, telling that it's invalid. This is exactly for failing silently.

ivan-mogilko:
> I am trying to figure it out https://github.com/adventuregamestudio/ags/issues/1364#issuecomment-1507432268, but the viewports or the overlays case seem to have lots of specifics to their case, I can't quite get much of the idea mentioned there...

Which part needs clarification? I may try to find better words.

But viewport or overlay examples do not have to be taken fully of course, only generic bits. They are mentioned in the step-by-step list.

ericoporto:
It's only that it looks very specific and I think in general the JoystickImpl can be added only

1. On initialization (for joysticks that are already connected before the engine starts)
2. When a connection event happens

And it can only be removed

1. When a disconnection event happens

So it looks it should be rather simple to interact with the vector. I don't understand much of vectors and references, but I think if a joystick disconnects the references from the elements in the vector after it would change, right? This means we should always get a reference when handling the Script API calls, and never store this reference to it, and because of how events are handled in sys_events, so not in parallel, the reference would be valid for the entirety of acting on the script call.

Now on Script API, there's no open or close there anymore, and there has to be a way to count the number of joysticks and to get them by number.

ivan-mogilko:
> So it looks it should be rather simple to interact with the vector. I don't understand much of vectors and references, but I think if a joystick disconnects the references from the elements in the vector after it would change, right? This means we should always get a reference when handling the Script API calls, and never store this reference to it, and because of how events are handled in sys_events, so not in parallel, the reference would be valid for the entirety of acting on the script call.

First of all, you do not have to store JoystickImpl in a vector, you may use a map. Actually, the map may be more suitable, this way the order in memory does not matter, so when an object get removed nothing has to be fixed.
For instance, with Overlays engine has to keep empty slots in the vector if something is removed in the middle (after recent optimization). I kept vector for overlays because it proved to work slightly faster than a map when overlays are added/removed alot. But joysticks probably won't be added/removed by hundreds every game tick, so that is not necessary for them.

Secondly, I was not able to clearly understand what you say about references in this paragraph, so cannot tell whether we speak about same thing or not.
I should have probably draw charts along with text, I will try to do this next time...
The idea, in simple terms, was this:

```
[ ScriptJoystick ]
 holds JoystickImpl's ID  ----------> [ container of JoystickImpls ]
                                       can give JoystickImpl by ID

[ JoystickImpl ]
 holds a handle to ScriptJoystick  ---> [ Managed Pool ]
                                       can give ScriptJoystick by handle
````

When a new joystick connects:
* add JoystickImpl to a internal container;
* create and register ScriptJoystick in the dynamic objects system;
* save Joystick's ID in ScriptJoystick;
* save handle to ScriptJoystick in JoystickImpl;
* also dont forget to add 1 ref to this ScriptJoystick so that GC won't clear it up;
So we have a mutual connection between these two.

When a user asks for a joystick of certain ID (static function Joystick.GetJoystick or whatever):
* get JoystickImpl by that ID;
* use saved handle to retrieve connected ScriptJoystick and return to the script.

When a user uses some joystick function or property:
* Script function accepts ScriptJoystick object;
* ScriptJoystick stores joystick ID, use it to get JoystickImpl and do something with it;

When a joystick disconnects:
* get related JoystickImpl;
* use handle saved in it to get related ScriptJoystick;
* invalidate ScriptJoystick by setting an ID stored in it to some invalid value, say -1. This will let the joystick's script functions know that it should not be used, and let just silently ignore user commands (for example).
* remove 1 ref to this ScriptJoystick in managed system (we added 1 extra ref when creating it to prevent it from disposal).
* finally remove JoystickImpl from its container.

---

Now, if I understand correctly, your concern is about real joysticks connecting and disconnecting in parallel with the script run. This is a valid concern, but actually same concern would be about anything in the engine that uses joystick, such as testing for a speech skip, for example (a joystick may disconnect while we're testing for a skip key/button press).

Unfortunately I haven't yet looked in how the SDL handles this, and what does the engine do in events, so I'll have to do this to understand the problem better. But overall, situations like this may be solved either
* by use of mutexes, if some data may be invalidated from another thread;
* by keeping a record of last device state, and update only when it's safe to update.
That's generally speaking.

**Comment by ivan-mogilko** [comment]
I looked into the latest commits, did not really check everything thoroughly, but from the first glance the overall idea is correct.

**Comment by ericoporto** [comment]
I did it in a computer without joysticks 😬, but this night and through the weekend my plan is to work with actual hardware to see if things work!

Assuming they do, I will merge them and then the work has to be done in internal skipping things: video skipping, and messages (display and say), and some form of configuring this (no idea how yet).

**Comment by ericoporto** [comment]
[JoystickTestAgsProj.zip](https://github.com/adventuregamestudio/ags/files/12205717/JoystickTestAgsProj.zip)
Made a project to test the buttons and it looks like things are working so far! I left the history of commits to explain how things are the way they are so far.

**Comment by ivan-mogilko** [comment]
Something I could not understand is, there's a commit that changes map back to vector, and it has this comment:

> Getting by index doesn't work if it's not a vector

What is this index, and why do you need to be getting by it, if there's already a map's key?

**Comment by ivan-mogilko** [comment]
Small hint: the default value of a handle should be 0, this means "no object" in managed pool.

**Comment by ericoporto** [comment]
For `Joystick.Get(int I)` and `Joystick.GetCount()` to work it needs a sequential list of joysticks in somewhere that persists for each Get call. The only container that works for this is a vector - the other IDs aren't guaranteed sequential numbers. :/

**Comment by ivan-mogilko** [comment]
> For `Joystick.Get(int i)` and `Joystick.GetCount()` to work it needs a sequential list of joysticks in somewhere that persists for each Get call. The only container that works for this is a vector - the other IDs aren't guaranteed sequential numbers. :/

This sounds very strange, as if the interface depends on implementation, which would be a wrong thing.
IDs assignment should not depend on which container you use underneath, this should work independently from that.
If you need to guarantee sequential numbers, then you may have a ID counter somewhere, this is similar to how overlays or managed objects work, for example.

What is the meaning of "i" in `Joystick.Get(int i)`?

**Comment by ivan-mogilko** [comment]
Ah. Sorry, I got confused again.

So `Joystick.Get()` actually is supposed to return different results if a joystick in the middle is removed. It accepts not some arbitrary fixed IDs, but the index in a sequential list (0 -> Joystick.Count - 1).

**Comment by ericoporto** [comment]
I think I understood what you mentioned about the IDs, I made so once a Joystick is removed, it iterates through the vector and fixes the IDs of all of the script joysticks that are connected to some implementation and the one removed has the ID set to -1, so now for each script check the ID is directly the index, so it's really fast, and removal of joysticks should not happen all the time and also there will be usually a very low number of joysticks, so overall this should be fast too.

The IDs now for the script joysticks are the index of the joystick impl objects.

**Comment by ivan-mogilko** [comment]
> Back at trying to figure it out what to do with Display, Say and video skipping.

Previously we found that it's too complicated to do using current skip enums. The only exception is WaitInputType that accepts multiple input types as flags. Therefore, to solve this cleanly, we would have to fully redesign skipping configuration in API.
I believe that this lies out of the scope of this PR, and should be carefully designed in a separate task.

(don't remember where I posted this but there was def a comment somewhere)

For a quick first solution I proposed to *skip by joystick whenever there's a skip by any other input*, because it's unlikely that devs or players will want to let skip with mouse but not joystick, for instance.

Besides that, there's another question: this PR still targets the master branch. I think we decided that it should be retargeted to ags4?

**Comment by ericoporto** [comment]
I will retarget to ags4 and rebase.

I will adjust to skip on any button press then!

**Comment by ericoporto** [comment]
ok, if the joystick is also a valid gamepad, then it will generate the internal events for skipping in some way (we know how their buttons map!), but if the joystick is not a valid gamepad, then it should be handled in script. At least this is how things are done currently, maybe later once there is an API to configure this, we can adjust this.

In case it's easier for review, I can break the engine commit in two and have one for the general joystick stuff and other for the skipping, just let me know.

**Comment by ivan-mogilko** [comment]
> In case it's easier for review, I can break the engine commit in two and have one for the general joystick stuff and other for the skipping, just let me know.

Yes, please, that was the idea: first change adds a API implementation, next change implements default game behavior.

**Comment by ivan-mogilko** [comment]
I don't have a real joystick, but I recall there were emulators, I will try testing this in the following days.

**Comment by ericoporto** [comment]
I separated the commits, they both should build fine, but I only tested on Windows.

**Comment by AlanDrake** [comment]
So far so good.
Only my non-standard directinput device failed to bind LB/LT RB/RT properly, but it's an ancient piece of junk.

I've made and updated test project that can show both xbox or ds4 glyphs, and fixes the imgi button not working when there are 2 or more devices. 
[Gamepadv4.zip](https://github.com/adventuregamestudio/ags/files/12242292/Gamepadv4.zip)

**Comment by AlanDrake** [comment]
I also finished testing with the Steam Deck. Works both in windows and linux versions.
The only weird thing is that there are three devices available, though only the first actually does anything. Must be a Deck quirk.

**Comment by ericoporto** [comment]
Holy shit this is beautiful!!! 

image
(I am adding the executable here [AgsGameGamepadV4.zip](https://github.com/adventuregamestudio/ags/files/12245665/AgsGameGamepadV4.zip), to make it easier for me testing in other OSes)

File: Editor/AGS.Editor/Resources/agsdefns.sh [hunk 1]
```diff
@@ -2471,6 +2529,36 @@ builtin struct Screen {
 };
 #endif
 
+#ifdef SCRIPT_API_v360
```
AlanDrake, line 2532: I think you missed to update this to "SCRIPT_API_v399"

ericoporto: ok, fixed this!

ivan-mogilko:
I did not think things should have changed much, you could probably just modify `run_service_gamepad_controls` like this:
```
bool run_service_gamepad_controls(GamepadInput &out_key)
{
    out_key = no key code; // clear the output
    if (ags_inputevent_ready() != kInputGamepad)
        return false; // there was no gamepad event

    const SDL_Event mb_evt = ags_get_next_inputevent();
    if (key_evt.type == SDL_CONTROLLERBUTTONDOWN)
    {
        out_key.Button = Gamepad_Button_SDLtoAGS(static_cast<SDL_GameControllerButton>(key_evt.button));
        break;
    }
    return out_key.Button != no code;
}
```

ericoporto:
Uhm right. Well, I can not worry about this now and leave the problem for later - the queue dropping. The issue I think is once there's some sort of mapping for which is the skipping button this will be lost (I think) once the controller is disconnected - at least I imagine this map would be in the same place where the SDL controller current is.

I also noticed I am doing something wrong with the save game - as I can't load saves with the joystick without crashing the engine.

ivan-mogilko:
Hm, I tried testing this using one virtual joystick program I found (https://sourceforge.net/projects/vjoystick/), but while it detects the joystick itself, it fails to detect the button and axes changes. Possibly this is a problem in SDL2.

ericoporto:
> while it detects the joystick itself, it fails to detect the button and axes changes

I think vJoy may not have a Gamepad string, are you checking regular buttons and axis or the gamepad variant? Does it work at least in the browser?

ivan-mogilko:
I found the one that works, although it's quite an ancient program from 2009, but it at least allows to map buttons and axes from keyboard directly: https://www.dropbox.com/scl/fi/a5m08rxvgmq94uuf2k4ix/vjoysetup.zip?rlkey=t224awe1hvlusoqykdzk3xvs3&dl=0

ivan-mogilko:
> I also noticed I am doing something wrong with the save game - as I can't load saves with the joystick without crashing the engine.

When you say "crashing", do you mean a program exception or engine reporting its own internal error?
I got error which tells that unserializer does not know how to deal with this type. You need to add "Joystick" to this list:
https://github.com/adventuregamestudio/ags/blob/f600979d73243d9321cd5c8fe393e9ab4c6b2435/Engine/ac/dynobj/cc_serializer.cpp#L75

ivan-mogilko:
In regards to script API, I noticed a certain discrepancy: a number of elements is being read using a property (ButtonCount, AxisCount, etc), but the number of joysticks is read using a function GetCount().

ericoporto:
Oh, right, I could read the joysticks through a property, I just don't know how to name it - like, it looks like it wants a global joystick [] array with a Length property (but this only exists in the new compiler)

So I went for simply "Get" name - instead of Joystick.Joystick[] and Joystick.JoystickCount, it felt a bit weird to repeat the name. We traditionally use `Create` for factory methods, but nothing is being created... So overall I am mostly in doubts about the right way to name this factory static method that gets internal references.

ivan-mogilko:
Hmm, what about `Device[]` and `DeviceCount`?

ericoporto:
I used Joysticks and JoystickCount because I think it's one less concept than adding the word `Device`. I also added the change so that a save won't break. I can squash these in the joystick creation commit.

ivan-mogilko:
> I used Joysticks and JoystickCount because I think it's one less concept than adding the word Device. I added these in the API description above in the post.

But if we have more device arrays in the future, this term would be used there too. For example, Mouse.Devices, Touch.Devices, and whatnot.

ericoporto:
For some reason `Joysticks[]` is not added to the autocomplete, even though calling it works and compiles fine.

ivan-mogilko:
I suppose this may be merged now, unless you want to have more people to test this.

ivan-mogilko:
I suppose that this may be merged now, unless you want to get more people in testing this first.

ivan-mogilko:
I have a question: what is the difference between gamepad buttons and joystick buttons? I notice that you check only gamepad buttons but not joystick when skipping.

Editor/AGS.Editor/Resources/agsdefns.sh
```diff
@@ -2471,6 +2529,36 @@ builtin struct Screen {
...
+  /// checks if a gamepad button is pressed, including dpad.
+  import bool IsGamepadButtonDown(eGamepad_Button button);
+  /// get gamepad axis or trigger, trigger only has positive values.
+  import float GetGamepadAxis(eGamepad_Axis axis, float dead_zone = AXIS_DEFAULT_DEADZONE); 
+  /// get joystick axis or trigger, by index
+  import float GetAxis(int axis, float dead_zone = AXIS_DEFAULT_DEADZONE);
+  /// checks if a gamepad button is pressed, including dpad.
+  import bool IsButtonDown(int button);
+  /// checks if a gamepad button is pressed, including dpad.
+  import eJoystick_Hat GetHat(int hat);
```
ivan-mogilko, line 2553: You seem to copy/pasted the function description for IsGamepadButtonDown, IsButtonDown and GetHat, these descriptions all say the same. Is this intended or a mistake?

In terms of style, all the existing comments to functions and properties in the script header begin with a capital letter, so you might change that for the style consistency too.

ericoporto: Need to fix the comments, it's a mistake

ericoporto:
> have a question: what is the difference between gamepad buttons and joystick buttons?

Gamepad buttons are mapped in a big list to predictable buttons, so I can get specific ones through similarity with the xbox360 controller.

https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_gamepad_db.h (this is SDL3 but sdl2 has a similar internal db)

Joystick buttons can be anything - in a PlayStation joystick, if you press the top of am axis, it's a button, so ideally it's best to set specific buttons for skipping, but this would require some function to set this button and this is best in a different pr - it also needs a logic to skip events from joysticks with open gamepads that you want to prefer the gamepad mapping.

ivan-mogilko:
Why cannot it skip from any button from a joystick, is there a reason for that?

But this made me concerned about something else in API. I think you mentioned this earlier, but other frameworks put all buttons in one list (e.g. Unity has all keyboard keys, mouse buttons and joystick buttons in [a single list of constants](https://docs.unity3d.com/ScriptReference/KeyCode.html)).
In our current API the only way to do the same is to combine a InputType flag with a key. This may be not as convenient, but still possible.

But here comes the trouble: judging by this PR, there's a single InputType for both joysticks and gamepads called `eInputGamepad`. Gamepad buttons begin with 0 (actually 1, as 0 is `eGamepad_ButtonInvalid`). But joystick's buttons also begin with 0, which means that their indexes and gamepad indexes will conflict.

How may we resolve this, if we want to let write both gamepad buttons and joystick buttons somewhere as a "skip button"?

ivan-mogilko:
Or do I misunderstand the whole thing, and these gamepad and joystick buttons are essentially same thing, the only difference is that on gamepad they have fixed "name/meaning", and not on joystick?

ivan-mogilko:
Oh, I suppose I understand what the problem is now.

Because non-gamepad joysticks don't have a fixed button meaning, we cannot make global properties that say "do this on joystick button N press", as each connected joystick may have their own meanings for buttons. We may only say "do this on joystick ID button N press", that is have both joystick AND button indexes together.

If we do not specify joystick's ID, then there are only 2 sort-of valid variants remaing:
- any button;
- particular gamepad button, which should only trigger if the device is a gamepad.
or something like that...

ericoporto:
> there's a single InputType for both joysticks and gamepads called eInputGamepad. 

I think we should have eInputJoystick too I just didn't add because I wasn't sure how to process the events. And I imagine no joystick has more than 30 buttons so we have space to pass something that identifies the joystick (the index in the vector we have perhaps? I don't know what to pass as identifier...)

ivan-mogilko:
> I think we should have eInputJoystick too I just didn't add because I wasn't sure how to process the events.

Which events?

> And I imagine no joystick has more than 30 buttons so we have space to pass something that identifies the joystick (the index in the vector we have perhaps? I don't know what to pass as identifier...)

If this is necessary to combine in a single integer, I think that some simple method like ((Joystick ID * 100) + Button Index) might work?

ivan-mogilko:
Maybe separate input type is not necessary? After all joysticks and gamepads are mostly similar, they have buttons and axes, which are identified by a ID in about same range, the only difference seems to be that gamepads have a fixed meaning for these and joysticks don't, as you said.

ericoporto:
> If this is necessary to combine in a single integer, I think that some simple method like ((Joystick ID * 100) + Button Index) might work

Uhm right, that is a good idea, I will think a little bit about this!

ivan-mogilko:
But what is left for this PR now? If a separate type is necessary, what exactly prevents from adding it?

Editor/AGS.Editor/Resources/agsdefns.sh
ericoporto: adjusted these comments and a few others

ericoporto:
I am trying to figure it out if Joystick that is a valid Gamepad also produces joystick events or not. I am almost just dropping the skip checks from this PR and adding them separately because I don't have a clear idea yet on how to assign joystick buttons to do skip checks.

I am thinking about something like the first joystick is assumed to be joystick 0 (or 1 if we want it to be 1 indexed), and the second 1 (or 2), the third 2 (or 3)... And then give some way to reorder (no idea on the API yet) - so you can say this is the joystick for player 1, this is the one for player 2, ... Because then, the buttons could be in an unified integer

> If this is necessary to combine in a single integer, I think that some simple method like ((Joystick ID * 100) + Button Index) 

And then if doing something like this, a future skip api maybe could take a dynamic array of integers with the buttons that are accepted for it - it would probably nice to be able to declare a new dynamic array with it's values in the same line.... So it would be like

```AGS Script
Video.SetSkipInput(new Keys[]{eKeyEscape, eKeySpace, eKeyJoy1Button2});
```

ivan-mogilko:
> I am thinking about something like the first joystick is assumed to be joystick 0 (or 1 if we want it to be 1 indexed), and the second 1 (or 2), the third 2 (or 3)... And then give some way to reorder (no idea on the API yet) - so you can say this is the joystick for player 1, this is the one for player 2

I see a serious problem here, which I did not notice earlier: the Joysticks are lacking a fixed ID. This will make it impossible to configure control mapping. Because if you remember mapping using the volatile index in the Joystick's array, then removing a disconnected joystick from that array will break the mapping.

Therefore, there **must** be a fixed ID of some sort, even if it's an arbitrary number. Like, for example, a counter that is continuously incremented for every connected joystick.

ivan-mogilko:
> I am almost just dropping the skip checks from this PR and adding them separately because I don't have a clear idea yet on how to assign joystick buttons to do skip checks.

These skip checks now are a temporary workaround, they do not have to be fixed for the future. It's simply something that lets player progress the game if they are playing using a joystick.
For that reason, you could make it skip literally on any joystick button, for now. And look for the better solution later.

> And then if doing something like this, a future skip api maybe could take a dynamic array of integers with the buttons that are accepted for it - it would probably nice to be able to declare a new dynamic array with it's values in the same line.... So it would be like
> 
> ```
> Video.SetSkipInput(new Keys[]{eKeyEscape, eKeySpace, eKeyJoy1Button2});
> ```

I remember you mentioned an idea of Actions, and IMO it's much better to work towards these instead of passing arrays into a function. In that case the Video would have a Skip action as a property (e.g. `Video.SkipAction`) that returns a `Action` (or `InputAction` etc) struct. I see a number of benefits here, starting with being able to read your configuration back.

Ermmm... deja vu, I wonder if I said something like this before...

ericoporto:
> I remember you mentioned an idea of Actions,

Yeah, my idea is to be able to set these either through script or in the editor, and then you could link events to the actions. There are two things that I need before doing that which is, some way to pass a function (or pointer downcast of an object) and some way to have events happen in the non-blocking thread (rep_exec_always).

In my test game the approach I have currently is a [virtual controller](https://github.com/ericoporto/sandwalker/blob/ags4-test/SandWalker/ControlMapper.asc) that maps joystick/gamepad and mouse+keyboard to it, and then I map from this virtual controller to the [Input](https://github.com/ericoporto/sandwalker/blob/ags4-test/SandWalker/Input.asc) of the things (actors and the GUIs).

> I see a serious problem here, which I did not notice earlier: the Joysticks are lacking a fixed ID. This will make it impossible to configure control mapping. Because if you make the mapping using a volatile index in the Joystick's array, then removing a disconnected joystick from that array will break the mapping.

I am unsure if I need to figure this out in this PR as it doesn't have mapping in it yet. I mentioned the approach of assigning a number (usually something from 1-4) because some joysticks also can show a led and sometimes even remember which led they had as an indicator of their number.

ivan-mogilko:
> I am unsure if I need to figure this out in this PR as it doesn't have mapping in it yet

But mapping can be done by user in the script. I've been wondering, how would they be able to distinguish joysticks?

I suppose we need to hold for a moment, and clarify: what are the requirements for this PR? Which abilities it supposes to give to users? And think whether these are covered correctly.

ericoporto:
> For example: how do you write a script that maps 2 joysticks to 2 player actions in a game, assuming that any joystick may get disconnected and reconnected again?

For now, the only way is to prompt on reconnection and ask who is it or guess - say if it there was a disconnection, store who was disconnected and reload later. The biggest problem is still save loading, where afaict there is no easy way to recover and the best is to always let the developer handle how they want - say assume only one joystick in one player game to auto-recover and other configs you prompt the user after load.

ivan-mogilko:
Well, the way I see this PR works for now, besides few narrow cases, like the Display box that you mentioned.

I believe this may be merged; but later we need to gather some plan on how to progress from here, and group future tasks, because there's too many things came up in this ticket, that should rather be addressed separately:
problem of super-blocking Display box, problem of identifying joysticks,
and there's also a question of having separate button down/up events in script (I thought to mention this, but then thought this pr is already taking too long).

ericoporto:
> and there's also a question of having separate button down/up events in script 

The problem is the events are kinda useless because they happen in blocking context - they can't call SkipWait for example.

ivan-mogilko:
> The problem is the events are kinda useless because they happen in blocking context

That might mean that they are currently useless for skipping, but that does not mean that they are ultimately useless. There are actions that might happen on a regular "thread", similar to the ones that are commonly dealt with in on_key_press and on_mouse_click.

Issue #1123, opened 2020-09-27
Character's explicit animation resets to frame 0 if char stops moving


ivan-mogilko:
Most recent report: https://www.adventuregamestudio.co.uk/forums/index.php?topic=58446.0

**Describe the bug**
If a non-blocking character animation is started, and then character's Move command is called (either blocking or non-blocking), then, as soon as character arrives to destination and stops:
* if animation was already complete, then character's frame resets to frame 0 of the animation loop and keeps that;
* if animation was still playing, then it resets to frame 0 and continues playing to the end again.

**AGS Version**
At least in AGS 3.4.3 - 3.5.0, but could be an older issue.

**To Reproduce**
1. Create a separate view with 1 loop and few frames;
2. Use similar script:
<pre>
player.LockView(VMOVEANIM);
player.Animate(0, 2, eOnce, eNoBlock);
player.Move(50, 100, eBlock, eAnywhere);
</pre>
3. Run the game and trigger above script.
4. Observe the problem.
Note you may experiment with the movement and animation speeds to see both potential outcomes described above.

**Expected behavior**
Character stopping after movement should not reset (or otherwise affect) ongoing animation if the view is locked.

**Known workarounds**
The simpliest known is to change coordinates by hand instead of calling Move, but this is only useful if you are moving without walkable areas, as you won't be able to utilise pathfinding, for example:
<pre>
player.LockView(VMOVEANIM);
player.Animate(0, 2, eOnce, eNoBlock);
int current_x = player.x;
while (player.x > current_x - 100)
{
    player.x -= 5; // change 5 to the wanted speed
    Wait(1);
}
</pre>

May be or not relevant to #935.

Issue #935, opened 2019-10-01
Walk animation keeps going after character arrived at destination


ericoporto:
![walkinplace](https://github.com/adventuregamestudio/ags/assets/2244442/6f5288bf-c679-4da6-a52c-136155e6017c)

Observed by Francisco in higher resolution games, when character walks all the way from right to left in some of his rooms, with movement linked to animation, once the character has arrived at destination, the animation still goes for a few frames before stopping. Not sure if relevant but he also reported using 1:2 mask resolution. He also noted he has observed the glitch before at AGS 3.4.

ivan-mogilko:
Francisco's gif: https://cdn.discordapp.com/attachments/221048075207180288/628636698913538049/walkinplace.gif

ivan-mogilko:
Not sure if relevant, but there's a confirmed bug when on-going animation may reset to frame 0 and play again as soon as character stops: https://github.com/adventuregamestudio/ags/issues/1123

ivan-mogilko:
By the way, I did a fix where a character was stuck in either a "walking" state or "between" states for a short while. I wonder if this is related somehow:

1084cf5207cf3e05af6e6b8eeb3a3345ffbe13bc

ivan-mogilko:
@AlanDrake do you know if we are allowed to attach the test game from Franscisco, the one that you sent to me, as a test case which reproduces 100%?

I wonder if it will still reproduce if we replace game gfx with placeholders, as this likely is not related to sprites.

ivan-mogilko:
This is for the record, because there has to be a reproducible test case in the ticket.

ivan-mogilko:
So, testing this under debugger, what happens in the code is this.

The algorithm in question is `do_movelist_move` function:
https://github.com/adventuregamestudio/ags/blob/f4bd80df1d7b487fd6ff1d69636659ab67b2ecea/Engine/main/update.cpp#L64
Specifically, what happens when the character finished walking along one axis, but did not finish walking along another axis. For instance, it reached X destination, but not Y:
https://github.com/adventuregamestudio/ags/blob/f4bd80df1d7b487fd6ff1d69636659ab67b2ecea/Engine/main/update.cpp#L73-L76

The the engine does the following:
1. Sets flag "x done", locking x position (won't be updated further);
[...]

PR #1761, opened 2022-09-02
Engine: rewrote MoveList to avoid using packed shorts for coords

AlanDrake:
I wanted to make this part more readable.
Not sure to what extent I should drag this PR but I really want to rename some stuff because last time I looked at this code I had a bad time.

Also, those `fixed` variables are some legacy from DOS era, aren't they? I could convert them to floats maybe.

Things to note:
- there are `memcpy`s in route_finder_impl.cpp and route_finder_impl_legacy.cpp
- `class Navigation` is still packing shorts into int, will look at it later
- I found a comment about implementing the movelist as a `vector`, could be the next step

### DIFF
--- a/Engine/ac/movelist.cpp
+++ b/Engine/ac/movelist.cpp
@@ -48,8 +48,8 @@ HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
         pos[i].X = in->ReadInt32();
         pos[i].Y = in->ReadInt32();
     }
-    in->ReadArrayOfInt32(xpermove, numstage);
-    in->ReadArrayOfInt32(ypermove, numstage);
+    in->ReadArrayOfFloat32(xpermove, numstage);
+    in->ReadArrayOfFloat32(ypermove, numstage);
     return HSaveError::None();
 }
 
@@ -70,6 +70,6 @@ void MoveList::WriteToFile(Stream *out)
         out->WriteInt32(pos[i].X);
         out->WriteInt32(pos[i].Y);
     }
-    out->WriteArrayOfInt32(xpermove, numstage);
-    out->WriteArrayOfInt32(ypermove, numstage);
+    out->WriteArrayOfFloat32(xpermove, numstage);
+    out->WriteArrayOfFloat32(ypermove, numstage);
 }
--- a/Engine/ac/movelist.cpp
+++ b/Engine/ac/movelist.cpp
@@ -21,11 +21,6 @@ using namespace AGS::Engine;
 
 HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
 {
-    if (cmp_ver < 1)
-    {
-        return new SavegameError(kSvgErr_IncompatibleEngine, "Outdated movelist format.");
-    }
-
     numstage = in->ReadInt32();
     // TODO: reimplement MoveList stages as vector to avoid these limits
     if (numstage > MAXNEEDSTAGES)
@@ -53,7 +48,7 @@ HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
     return HSaveError::None();
 }
 
-void MoveList::WriteToFile(Stream *out)
+void MoveList::WriteToFile(Stream *out) const
 {
     out->WriteInt32(numstage);
     out->WriteInt32(fromx);
--- a/Engine/ac/movelist.cpp
+++ b/Engine/ac/movelist.cpp
@@ -19,28 +19,11 @@
 using namespace AGS::Common;
 using namespace AGS::Engine;
 
-void MoveList::ReadFromFile_Legacy(Stream *in)
-{
-    in->ReadArrayOfInt32(pos, MAXNEEDSTAGES_LEGACY);
-    numstage = in->ReadInt32();
-    in->ReadArrayOfInt32(xpermove, MAXNEEDSTAGES_LEGACY);
-    in->ReadArrayOfInt32(ypermove, MAXNEEDSTAGES_LEGACY);
-    fromx = in->ReadInt32();
-    fromy = in->ReadInt32();
-    onstage = in->ReadInt32();
-    onpart = in->ReadInt32();
-    lastx = in->ReadInt32();
-    lasty = in->ReadInt32();
-    doneflag = in->ReadInt8();
-    direct = in->ReadInt8();
-}
-
 HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
 {
     if (cmp_ver < 1)
     {
-        ReadFromFile_Legacy(in);
-        return HSaveError::None();
+        return new SavegameError(kSvgErr_IncompatibleEngine, "Outdated movelist format.");
     }
 
     numstage = in->ReadInt32();
@@ -60,7 +43,11 @@ HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
     doneflag = in->ReadInt8();
     direct = in->ReadInt8();
 
-    in->ReadArrayOfInt32(pos, numstage);
+    for (int i = 0; i < numstage; ++i)
+    {
+        pos[i].X = in->ReadInt32();
+        pos[i].Y = in->ReadInt32();
+    }
     in->ReadArrayOfInt32(xpermove, numstage);
     in->ReadArrayOfInt32(ypermove, numstage);
     return HSaveError::None();
@@ -78,7 +65,11 @@ void MoveList::WriteToFile(Stream *out)
     out->WriteInt8(doneflag);
     out->WriteInt8(direct);
 
-    out->WriteArrayOfInt32(pos, numstage);
+    for (int i = 0; i < numstage; ++i)
+    {
+        out->WriteInt32(pos[i].X);
+        out->WriteInt32(pos[i].Y);
+    }
     out->WriteArrayOfInt32(xpermove, numstage);
     out->WriteArrayOfInt32(ypermove, numstage);
 }
--- a/Engine/ac/movelist.h
+++ b/Engine/ac/movelist.h
@@ -13,7 +13,6 @@
 //=============================================================================
 #ifndef __AC_MOVE_H
 #define __AC_MOVE_H
-#include <allegro.h> // fixed math
 #include "game/savegame.h"
 #include "util/geometry.h"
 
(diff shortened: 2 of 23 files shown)


Engine/ac/movelist.h
```diff
@@ -23,8 +23,17 @@ using namespace AGS; // FIXME later
 #define MAXNEEDSTAGES 256
 #define MAXNEEDSTAGES_LEGACY 40
 
+struct MoveListPoint
```
ivan-mogilko, line 26: Unless you need extra variables, there's already a Point struct in geometry.h; in theory it may be extended later if more parameters will be needed.

AlanDrake: Made the switch to Point

AlanDrake:
So, I tried converting those `fixed` variabled to `float` and for some reason this has fixed the issue with the "sliding in place" (see #935 and maybe also #1123).

Because I didn't use epsilon for some float comparison (see update.cpp), and the cognitive dissonance in some code vs comment is confusing me, I don't know if this is a good or bad sign...

ivan-mogilko:
> Because I didn't use epsilon for some float comparison (see update.cpp), and the cognitive dissonance in some code vs comment is confusing me, I don't know if this is a good or bad sign...

Please elaborate?

AlanDrake:
> > Because I didn't use epsilon for some float comparison (see update.cpp), and the cognitive dissonance in some code vs comment is confusing me, I don't know if this is a good or bad sign...
> 
> Please elaborate?

1) In update CPP we had those completely unreadable ifs that were comparing the integer part of the fixed number using bitwise ops. But some comments are departed from reality, like `// X per move is > -1, so finish the move` but then it just checks for == -1.

2) Then in the router finders (both legacy and new) we have:
```
    if (move_speed_x > move_speed_y) {
      // speed = y + ((1 - xproportion) * (x - y))
      useMoveSpeed = move_speed_y + (xproportion * (move_speed_x - move_speed_y));
    }
    else {
      // speed = x + (xproportion * (y - x))
      useMoveSpeed = move_speed_x + ((1 - xproportion) * (move_speed_y - move_speed_x));
    }
  }
```
The comments are switched. I'm inclined to believe it's just the comment being wrong but without further investigation I don't actually know for sure.

3) Because float equality for small values should actually be using the epsilon value, I'm not sure my changes in updat.cpp are entirely correct, though it probably doesn't matter where I'm truncating the decimals.

AlanDrake:
I managed to once trigger a bug where the character turned to walk downward halfway the path, and the walking animation didn't play while it moved.
So, even if this fixes the walk-in-place bug, there are still issues to iron out...

AlanDrake:
I still don't know why the slide-in-place bug is not happening like in the original code, but since my last fixes I haven't seen bad behaviors from this rewrite. 
But I could just have been lucky, so I'm not sure if this PR is worthy of being merged yet.

ivan-mogilko:
Few notes:

1. Please, squash (merge together) first two commits, to avoid confusion (you add MovelistPoint first, and then replace with Point).

2. You have actually changed the save format for MoveList: it saved 1 int per step (with 2 shorts in it), now it saves 2 ints EDIT: 2 floats. Alternatively, you could save them converting to unsigned short, if you think that max coordinate of 65535 is fine (e.g. ```out->WriteInt16(static_cast<uint16_t>(pos[i].X));```). But if you prefer to save them as int32/float32, then you need to increment save format for anything that uses movelists: characters and room objects. If you don't want to support older save formats, then you also need to rise the "least supported component version" as well.

ivan-mogilko:
By the way, speaking of "walking in place" bug, have you noticed this comment in code?
https://github.com/adventuregamestudio/ags/blob/2b03d989c8ac5de3dac49b483bd216aaeb807354/Engine/main/update.cpp#L73-L75

Engine/main/update.cpp
```diff
@@ -81,48 +82,48 @@ int do_movelist_move(short*mlnum,int*xx,int*yy) {
...
     // 2.61 RC1: correct this to work with > -1 as well as < 1
-    if (ypermove == 0) { }
+    if (fabs(ypermove) < delta) { }
     // Y per move is < 1, so finish the move
-    else if ((ypermove & 0xffff0000) == 0)
+    else if (trunc(ypermove) == 0)
       targety -= adjAmnt;
     // Y per move is -1 exactly, don't snap to finish
-    else if (ypermove == 0xffff0000) { }
+    else if (fabs(ypermove - 1.0) < delta) { }
```
ivan-mogilko, line 95: This is incorrect, you are subtracting from a still signed value here. Should be: ```fabs(ypermove) < (1.0 + delta)```

Engine/main/update.cpp
```diff
@@ -81,48 +82,48 @@ int do_movelist_move(short*mlnum,int*xx,int*yy) {
...
 
-    if (xpermove == 0) { }
+    if (fabs(xpermove) < delta) { }
     // Y per move is < 1, so finish the move
-    else if ((xpermove & 0xffff0000) == 0)
+    else if (trunc(xpermove) == 0)
       targetx -= adjAmnt;
     // X per move is -1 exactly, don't snap to finish
-    else if (xpermove == 0xffff0000) { }
+    else if (fabs(xpermove -1) < delta) { }
```
ivan-mogilko, line 117: Same as above, should be: ```fabs(xpermove) < (1.0 + delta)```

Engine/main/update.cpp
ivan-mogilko, line 95: This is not correct, the condition will succeed for values in range (+1.0 - delta) to (+1.0 + delta), but not for -1.0, as original code assumes.

Perhaps fabs is not even necessary here, you may do ```if ((ypermove >= -1.0 - delta) && (ypermove <= -1.0 + delta))```

Engine/main/update.cpp
ivan-mogilko, line 117: Same as above, should have double negation:
```fabs(xpermove - (-1.0)) < delta```

**Comment by ivan-mogilko** [comment]
Sorry, another thing, your pr just made me realize there's this extra iagsstream class in otherwise empty folder; I wanted to get rid of it as some point, but maybe will just move the file into "util" dir for now.

**Comment by AlanDrake** [comment]
> Few notes:
> 
>     1. Please, squash (merge together) first two commits, where you add MovelistPoint first, and then replace with Point; as this is basically editing same code twice.

Ok

>     2. You have actually changed the save format for MoveList: it saved 1 `pos` int per step (with 2 shorts in it), now it saves 2 `pos.X,Y` ints. Alternatively, you could save them converting to unsigned short, if you think that max coordinate of 65535 is fine (e.g. `out->WriteInt16(static_cast<uint16_t>(pos[i].X));`). But if you prefer to save them as int32, then you need to increment save format for anything that uses movelists: characters and room objects. If you don't want to support older save formats, then you also need to rise the "least supported component version" as well.

I'll drop support for the older format in this case, there's no reason to have conversion logic for older structures in AGS4 savegames.


> By the way, speaking of "walking in place" bug, have you noticed this comment in code?

Yes, and I have no idea why it's commented or if it was superseded by something other change.
As you saw, the comments in this section don't match the code, the way it's supposed to work is obscure.


> Regarding changing x/ypermove to float, the changes in `do_movelist_move` look a bit strange, because some of the comments mismatch what the code does, at least what the new code does. I'll try to look more closely again... PS. I have no opinion on changing to floats per se, as I've never researched how the pathfinding works there.


`do_movelist_move()` is really a mess, and the places where it's used have messy named variables as well, which makes it harder to understand the context.
It manipulates all the parameters in input.
The return value has 0, 1, and  2 as possible values, but the only one that matters is whether it's `== 2`, in which case it appears to signal that it should change sprite direction. The value of 1 is not used anywhere, which could be a mistake. Or maybe it was deprecated. Nobody knows.
It's unclear what the ranges of xpermove/ypermove are meant to be, maybe it's doing some things according to angle.
`adjAmnt` is used for some obscure hack (?).
There's a `doneFlag` with magic numbers.

Even without the unreadable comparisons with the fixed variables, it's still largely unclear what exactly it's doing in a human readable way. I get the feeling that even CJ had lost track of what was happening there.

To document this algorithm I need to go around at random and rename the related misnamed variables and replace magic numbers. I didn't do that before because I wanted to avoid causing merge issues, but maybe it's time.

File: Engine/main/update.cpp (hunk 3 above)
Comment (AlanDrake) [reply Engine/main/update.cpp]: I'm tempted to use the almost_equal function, but I don't know what value of ULPs we'd need for equality, or even what "equality" means in this specific context.

Comment (ivan-mogilko) [reply Engine/main/update.cpp]: Hmm maybe make a local function here for now, with the chosen delta constant. The point is rather to reduce possible mistakes when comparing, since you will be using a function instead.

Comment (AlanDrake) [reply Engine/main/update.cpp]: Redone with local function

File: Engine/ac/movelist.cpp [hunk 4]
```diff
@@ -19,13 +19,13 @@
 using namespace AGS::Common;
 using namespace AGS::Engine;
 
-HSaveError MoveList::ReadFromFile(Stream *in, int32_t cmp_ver)
+HSaveError MoveList::ReadFromFile(Stream *in)
 {
-    if (cmp_ver < 1)
+    int version = in->ReadInt32();
```
ivan-mogilko, line 24: Um, this contradicts to every other save format, where the format version is only mentioned as "component version" once; here it will be mentioned per each MoveList object. What was the reason to do so?

AlanDrake:
Because I had already spent the morning fixing the PR, and as usual there came the annoying AGS thing that made me lose patience.
Anyway there are pros and cons, maybe the approach of saving version directly in the component could be better for having those standalone structures you mentioned in the draft. It's certainly more convenient to write, as it's all self-contained.

I'll try to redo with your other approach once I have some time.

ivan-mogilko:
> Anyway there are pros and cons, maybe the approach of saving version directly in the component could be better for having those standalone structures you mentioned in the draft. It's certainly more convenient to write, as it's all self-contained.

The save system currently is written to make automatic checks on component version, so you don't even have to do these "if (version < N) return error" inside the read functions: this check is done beforehand automatically; you just need to keep the table of supported versions updated.

With the latest change the save format is still not correct, as the movelists have a different format, but they are written inside the character & "current room" components, which results in changing their format as well. You would still have to adjust their component versions as well.

ivan-mogilko:
> I've increased the savegame version

There's just one thing, the `GameStateSvgVersion` is not the overall save version, but a "Game State" component version. Perhaps the comment is misleading there.
The global format version is `enum SavegameVersion`.

ivan-mogilko:
Ok, maybe it's not a big deal atm, i could just make minor edits after.

PR #1769, opened 2022-09-10
Mask handling improvements

### LINKED ISSUE #1768: Support importing indexed png, with fewer than 256 colors, as mask in room editor
**Describe the problem**
When editing masks in my image editor (Aseprite), I would like to work with indexed png images since they are much smaller than bitmap. 

Problem is I really would prefer to just have as many colors as I need in my palette before importing, instead of having to hold a 256 color palette - also because by default it loads with a 32 color palette. 

**Suggested change**
I would like to 
- support importing png images as rooms masks in the editor
- allow these png images to contain less than 256 colors
  - possibly ags should replace whatever palette the png had before with it's own right after import

Allow import of png files can be done by replacing the `MASK_IMAGE_FILE_FILTER ` in `Editor\AGS.Types\Constants.cs` with the one below

```C#
public const string MASK_IMAGE_FILE_FILTER = "All supported images(*.bmp; *.png)|*.bmp;*.png|Windows bitmap files (*.bmp)|*.bmp|Portable Network Graphics (*.png)|*.png";
```

The palette change is a bit trickier and I don't know how to implement

AlanDrake:
- Normalizes palettes also handling partial palettes
- Safer handling of invalid masks
- Added PNG to the formats available for mask export/import
- Restored the original red color for mask brushes

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1740,6 +1740,11 @@ namespace AGS.Editor.Components
                 _maskCache[mask] = bmp.Clone() as Bitmap;
                 _loadedRoom.Modified = true;
             }
+            else if (!_maskCache.ContainsKey(mask) || _maskCache[mask] == null)
+            {
+                // we can get here while opening a room the first time, and after closing it without saving then reopening
+                _maskCache[mask] = CreateMaskBitmap(mask, _loadedRoom.Width, _loadedRoom.Height);
+            }
         }
 
         int IRoomController.GetAreaMaskPixel(RoomAreaMaskType maskType, int x, int y)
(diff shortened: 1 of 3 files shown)


ericoporto:
Hey, I still get the "This is not a valid mask bitmap" error when using an indexed png with fewer than 256 colors (tried with 32 colors). [mask_test_3.zip](https://github.com/adventuregamestudio/ags/files/9541552/mask_test_3.zip)

AlanDrake:
It seems the .NET facilities are convertig the image to 32bit when they find a `tRNS` chunk (palette transparency data).
Removing said chunk allowed me to import the mask.

If you want to use Aseprite make sure you're not using a transparent palette, (editing a png without tRNS chunk should work) because I can't see any convenient way to easily change this behavior

ericoporto:
I found a non-convenient way [here](https://stackoverflow.com/questions/24074641/how-to-read-8-bit-png-image-as-8-bit-png-image-only). Without it, I essentially can't use Aseprite with AGS honestly. But this can close that issue if you want, and then I create a new issue for this specifically. Curiously, I think the wine implementation doesn't have this bug, because I distinctly remember using Aseprite with AGS in Linux without this problem.

ericoporto:
If you merge this in ags4, I can PR doing the transparency stuff.

AlanDrake:
Alright, I'll merge the PR since it does what it was meant to, except working around the tRNS thing that requires an hack.
I don't have an opinion yet about the workaround, but I see it revolves around removing the tRNS chunk on the fly.

I might do some more work on masks in the future, I saw there was a ticket about the mask brush palettes, and since more hotspots have been added maybe it's time to redo the color part in order to have a single source of colors, unlike now that masks inherit the game palette and the brushes colors are hardcoded.

PR #1772, opened 2022-09-11
Editor: allow loading Room Masks with transparent chunks

### LINKED ISSUE #1771: Room Masks: Importing an indexed PNG with alpha as mask is not supported
**Describe the bug**
When trying to import an indexed PNG image that has transparency, AGS will error, saying the image is not valid and doesn't have 256 colors.

The problem is transparency in png works by adding an optional "tRNS" chunk in the header, to specify the alpha of each palette entry, but after the .Net classes read and apply this correctly, it converts the image to 32bit for some reason. The result is any indexed png generated by Aseprite can't be imported as masks in AGS.

**AGS Version**
AGS 4, but if you enable the filter you can reproduce in other versions.

**Game**
Here's an example mask you can import [mask_test_3.zip](https://github.com/adventuregamestudio/ags/files/9541552/mask_test_3.zip).

**To Reproduce**
1. In the editor, create a new game from the default sierra template
2. Open the room, select hotspots from the list of layers 
3. Click in the import mask button and select the above .png file.

**Expected behavior**
It should import the png, and ignore the palette sent.

**Desktop:**
 - OS: Windows
 - Version 10

ericoporto:
Fix #1771 

From the discussion in [here](https://github.com/adventuregamestudio/ags/pull/1769#issuecomment-1242970015).

Still need to review the code and test it with more images.

### DIFF
--- a/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomEditFilters/BaseAreasEditorFilter.cs
@@ -464,7 +464,7 @@ namespace AGS.Editor
         {
             try
             {
-                using (Bitmap bmp = new Bitmap(fileName))
+                using (Bitmap bmp = BitmapExtensions.LoadNonLockedBitmap(fileName))
                 {
                     if (!(((bmp.Width == _room.Width) && (bmp.Height == _room.Height))))
                     {
--- a/Editor/AGS.Editor/Utils/BitmapExtensions.cs
+++ b/Editor/AGS.Editor/Utils/BitmapExtensions.cs
@@ -6,9 +6,176 @@ using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 
 namespace AGS.Editor
 {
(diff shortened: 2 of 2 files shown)


ericoporto:
Does anyone knows if there's any use in loading images from disk and locking them, or if everytime we load an image from disk, we could do it unlocked?

AlanDrake:
I don't think we have any case where locking is preferable, but I could be wrong.

ivan-mogilko:
I confirm this lets the 8-bit pngs with transparency through as a mask, but I have 2 questions:
1. I see that this changes how BitmapExtensions.LoadNonLockedBitmap, and that is also used for Room bgs. Which consequences does this have?
2. Is there any known rule for how the "transparency color" should be interpreted when loading a mask, like which index it's assigned to, or it depends simply on transparency's order in the generated palette?

ericoporto:
That is a good question, when I looked around I didn't see any reason why there is a LoadNonLockedBitmap, I believe there should be only a LoadBitmap and that it should always be unlocked.

Now about room bgs, I am not sure what would change, the difference is before indexed bitmaps with transparency in their palette would be silently converted to 32-bit, so for things that aren't that it should behave the same as before.

Issue #1716, opened 2022-06-26
Android: improve the touch-to-mouse input control scheme


ivan-mogilko:
There has been a number of suggestions to change the current control scheme in the past. I've posted one below under spoiler as an example, and more may probably be found around the issue tracker or ags forums (unfortunately they were never gathered into a dedicated ticket).

The point however is that it's desired to have configurable controls for an Android port (or generally - mobile / touch screen controls).

This may be approached from two directions:
1. Design a script API for touch controls and let game authors script their own control scheme. EDIT: there's a ticket for this: #1538.
2. Make default "touch screen - to mouse" controls configurable through a config file (and AGS Player's preferences). This will cover games that don't have a scripted touch controls. For this we'd need to _gather a list of things_ that may require an option (left mouse click, right mouse click, and so on).

This ticket is apparently dedicated to p2.

<details>
  <summary>Following was split out of the #275</summary>

_I was also wondering how I could change the controls, specifically, the two-finger double tap for right clicking. It's a little strange to pull off to be honest and sometimes the object is too small to get 2 fingers on it. Right now I'm struggling to right click these tiny objects in Resonance._

_I'd prefer every mouse control to be able to be executed with a single finger or stylus. SCUMMVM has a very good control system that AGS should consider adopting. Its:_

* _Tap + movement: Mouse movement_
* _Tap without movement: Left mouse button click_
* _Tap held for >0.5s without movement: Right mouse button click_
* _Tap held for >1s without movement: Middle mouse button click_
* _Double Tap + movement: Drag and drop_

_On devices supporting multitouch:_

* _Hold down one finger, tap with second: Right mouse button click_
* _Hold down two fingers, tap with third: Middle mouse button click_

</details>

ericoporto:
Please, let's prioritize #1538 before going this route.

i30817:
What about OS definitions of what is a mouse 'gesture'? For instance, i can configure in GNOME if a two finger click is right click and a 3 finger click a middle click, or setup the bottom zones of the touchpad for that (except middle click anyway).

What happens then, if you introduce highly divergent gestures apis?

ivan-mogilko:
> What about OS definitions of what is a mouse 'gesture'? For instance, i can configure in GNOME if a two finger click is right click and a 3 finger click a middle click, or setup the bottom zones of the touchpad for that (except middle click anyway).

> What happens then, if you introduce highly divergent gestures and touch apis?

[...]

Issue #1538, opened 2022-02-11
Enhancement: Proper touch input support


ericoporto:
AGS assumes that the player uses a mouse to play the game. In today times, there is a myriad of devices that support touch input that are used quite frequently, famously our mobile ports should be able to run in such devices. As developers mature they would need to be able to cater specifically for the constraints of such devices. The assumption of a mouse that can click can make some control schemes challenging to support in touch first devices.

- I propose we implement either touch specific events or pointer events (mixed touch and mouse) that could be used in the script API in multitouch control schemes.
  - since there is only `on_mouse_click` but not `on_mouse_down`, `on_mouse_move` and `on_mouse_release`, it may be better to use the pointer concept and create `on_pointer_start`, `on_pointer_move`, `on_pointer_end` to cover mouse and touch on the same elements. 
  - html also has an additional on_touch_cancel, need to see if this is needed/possible with sdl2.
  - ~~AGS currently has a limit of 5 events in the queue, it may be needed to allow a bigger queue or some touched points may be missed in a frame.~~

- Additionally, I would like to propose additional event bindings for the Button GUI Control, so it has additionally a touch down and touch release event (or pointer down, pointer release), and also a property that can be checked (IsTouched / IsPointed)
  - We can postpone doing the GUI button events

These are drawn from my observations of other engines, that I wrote about it here: https://ericonotes.blogspot.com/2020/11/a-quick-look-at-touch-handling-apis-in.html

Such API additions would allow for multitouch in GUIs, which can be used to better handle mobile device usage and respond more quickly in the interface (hitting two fingers at screen today will result in a finger being dismissed), and also this allows support for on screen joysticks.

*Forum topic*

- https://www.adventuregamestudio.co.uk/forums/index.php?topic=59710.0

**On Screen Joystick example**

Using AGS GUIs, it should be possible to construct something like below. Right now, it's not possible since using two or more fingers at the same time is not possible.

![touch joystick](https://user-images.githubusercontent.com/2244442/154060037-c44656be-e731-4fa7-81b1-6701bc436e36.gif)

**Note**

https://youtu.be/B_IqYy4T_AA?si=xOASxzLCrI0F1lV8&t=916

In this talk, the Broken Sword dev talks about the recent mobile port and it's interesting how their interface got adapted.

ivan-mogilko:
Just for a quick reference:

SDL2 touch-related events:
https://wiki.libsdl.org/SDL_TouchFingerEvent
[...]

Issue #275, opened 2015-09-13
Occasional crash to desktop on Android & controls question. 


vile80r:
Apologies if this is not the right area to post this. I'm not a frequent user of GitHub and I DO have trouble navigating around.  

I recently got AGS for Android.  I'm running it on an Nvidia Shield Tablet.  I've been playing Resonance and I have to say,  it looks and plays fantastic.  You've done a great job with this port. The only issue I've encountered is at least once in a game play AGS will just close with no warning or message. Just straight to desktop but, it fires right back up when I open it and I continue to play. 

I was also wondering how I could change the controls, specifically,  the two-finger double tap for right clicking. It's a little strange to pull off to be honest and sometimes the object is too small to get 2 fingers on it. Right now I'm struggling to right click these tiny objects in Resonance. 

I'd prefer every mouse control to be able to be executed with a single finger or stylus. SCUMMVM has a very good control system that AGS should consider adopting. Its:

Tap + movement: Mouse movement
Tap without movement: Left mouse button click 
Tap held for >0.5s without movement: Right mouse button click
Tap held for >1s without movement: Middle mouse button click
Double Tap + movement: Drag and drop 

On devices supporting multitouch: 

Hold down one finger, tap with second: Right mouse button click 
Hold down two fingers, tap with third: Middle mouse button click

One last feature request is to have the screen shrink when the keyboard is toggled.  Currently the keyboard pops up and covers half the screen.  I'm unable to see what I'm typing.  So,  if the game window could compress in height in relation to the keyboard's space I think that would make a more seamless gameplaying environment. 

Thanks for taking a look.  Again,  excellent work on this. 

T

sonneveld:
Thank you for reporting your feature request.. first up, do you might logging separate issues for 
1) the crashing (possibly any hints to reproduce)
2) the touch controls suggestions
3) keyboard toggle/screen shrink?

Second up, not sure if anyone is currently working on android port. One of the things on my list todo after my holiday is to build android releases via the build server.  I think that's a good first step before I can start to look at these features.

ericoporto:
The idea of this issue is to do all three things? Like the crash I can tell I can't reproduce. :/

ivan-mogilko:
> The idea of this issue is to do all three things?

First of all, this should be separated into 3 tickets each for one problem. The crash is likely to not be relevant anymore, as we have a different backend and different java app.

Then, of course, it's not necessary to do this exactly as suggested.

[...]

PR #1773, opened 2022-09-12
Android: reimplement touch-to-mouse controls in the engine

ivan-mogilko:
This reimplements touch-to-mouse controls inside the engine, similar to the control scheme of 3.5.1 (before moving to SDL2 and new Android port), with minor differences. Touch-to-mouse emulates mouse buttons and motion, letting players to control AGS games with Android port. The 3.6.0 currently uses default SDL2 emulation, which only supports LMB, but that's not enough for games which require RMB too.

Mouse emulation is enabled using a config setting ("[touch]", "emulate_mouse"). This parameter may have several values, acting as a choice between the control schemes. Right now it supports 2 schemes: a default SDL2-like scheme and reimplementation of an old AGS scheme (see below). The touch-to-mouse may also be disabled completely. In theory, it may be disabled also by the game's own setting, which is useful, for example, if AGS will have a proper touch API in the future version (see #1538).

The SDL2's internal touch-to-mouse is disabled using hints (SDL_HINT_TOUCH_MOUSE_EVENTS and SDL_HINT_MOUSE_TOUCH_EVENTS).

Control schemes:

**1. SDL2-style. LMB drag only.**

* Only the first finger (id = 0) is tracked, others are ignored;
* Touch up and down events directly correspond to mouse up and down events; so you put finger down to press mouse button, and release a finger to release a mouse button.
* Touch motion directly corresponds to the mouse motion.

Pros: simple;
Cons: no RMB emulation, cannot move cursor without holding a button down (for example: to scan around the room without issuing any commands).

This style is more or less suitable for one-click game control scheme, and especially games with drag-n-drop controls.

**2. Classic AGS-style.**

* One finger tap is LMB (mouse click issued when you *release* the finger);
* Two finger tap is RMB; you may hold 1st finger and keep tapping 2nd finger to keep "clicking" RMB, - this may be used for instance to cycle verbs in Sierra-style games.
* Dragging a finger means mouse movement, but without a button down.
* Double tap + drag: means mouse movement with LMB down. This means you have to quickly tap once and then press finger again to start a drag. This is implemented after another user's suggestion in the old ticket #275, also mentioned in #1716. The original method to start a drag was long-click, but I had concerns that it's not obvious when the dragging starts. With double tap it's more explicit, in my opinion.

**Potential TODO:**
* Add control scheme selection to the AGS Player's preferences and save to android.cfg.
* Tidy code by gathering variables into structs.

### DIFF
--- a/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/AGSRuntimeActivity.java
+++ b/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/AGSRuntimeActivity.java
@@ -279,9 +279,4 @@ public class AGSRuntimeActivity extends SDLActivity {
         return super.dispatchKeyEvent(ev);
     }
 
-    @Keep
-    protected  void AgsEnableLongclick() {
-
-    }
-
 }
--- a/Android/library/runtime/src/main/res/values/arrays.xml
+++ b/Android/library/runtime/src/main/res/values/arrays.xml
@@ -13,6 +13,18 @@
         <item>2</item>
     </string-array>
 
+    <string-array name="mouse_emulation">
+        <item>Off</item>
+        <item>One Finger (touch means button down)</item>
+        <item>Two Finger (tap to click)</item>
+    </string-array>
+
+    <string-array name="mouse_emulation_values">
+        <item>0</item>
+        <item>1</item>
+        <item>2</item>
+    </string-array>
+
 
     <string-array name="game_language">
         <item>Default</item>
--- a/Android/library/runtime/src/main/res/values/strings.xml
+++ b/Android/library/runtime/src/main/res/values/strings.xml
@@ -17,8 +17,8 @@
     <string name="CONFIG_GFX_SMOOTH_SPRITES">16</string>
     <string name="CONFIG_TRANSLATION">17</string>
     <string name="CONFIG_DEBUG_LOGCAT">18</string>
-    <string name="CONFIG_MOUSE_METHOD">19</string>
-    <string name="CONFIG_MOUSE_LONGCLICK">20</string>
+    <string name="CONFIG_MOUSE_EMULATION">19</string>
+    <string name="CONFIG_MOUSE_METHOD">20</string>
     <string name="toggle_keyboard">Toggle keyboard</string>
     <string name="exit_game">Exit game</string>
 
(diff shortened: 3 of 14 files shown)


ericoporto:
Hey, this looks really good! Can you explain how do I test these modes on my device? Is there any of them selected by default?

ivan-mogilko:
> Hey, this looks really good! Can you explain how do I test these modes on my device? Is there any of them selected by default?

This may be done simply through the default config file (acsetup.cfg)
<pre>
[touch]
emulate_mouse = 1 // 0,1,2 - disabled, first scheme, second scheme
</pre>

Default value is 1 (sdl-style).

ericoporto:
So the scheme 2 appears to work great.

But, if I enable relative mouse control, the cursor doesn't move at all. Relative is the one if you move your finger  deltaX,deltaY, it moves the same amount of the cursor, accounting for stopping at screen borders. It's useful because you can see the cursor instead of having it hidden by your finger.

ivan-mogilko:
Hm, I made relative controls working, but when it's disabled the cursor now has very strange movement, looks like either it reports dx/dy incorrectly, or I'm using them incorrectly. What happens is that the relative movement seem to be doubled when moving left/up and halved when moving right/down.

Is there a distinct way to tell whether engine is working in relative or absolute touch mode?

ericoporto:
I believe it's this line that enables the Relative mouse mode:
https://github.com/adventuregamestudio/ags/blob/9f091926e393f75c90c951f9597710bf17395332/Engine/main/engine_setup.cpp#L248

But I think for your thing to work, the SDL own relative mode has to be disabled.

ivan-mogilko:
> Question: Does your relative mode imply that the SDL own relative mode has to be disabled? (because it isn't and I was wondering if that could cause the movement being doubled in some directions)

That appeared to be float rounding errors, they caused delta movement to be larger towards zero and smaller when moving away from zero.

This error was not with relative mode, but with non-relative mode; anyway, in the end I changed the code so now it's not a problem.

ericoporto:
So, the relative mode while the finger is on screen works, as the finger moves, the cursor moves with the same delta, all alright.

But if I lift the finger, next time the finger touches the screen, the cursor teleports to somewhere. I _think_ it's moving the delta between the last position the finger was on screen and the new one. Ideally nothing would happen, and as the finger motioned the cursor would move from the position it was.

ericoporto:
Hey, the relative movement works perfectly in Android now! Apparently something broke on compiling on Linux, I can take a look at that when I am back home.

ericoporto:
Looks good now! :) Tested again and it's working.

ivan-mogilko:
Alright, I think what's left is a control scheme selection in the player's Preferences;
also "Long click" option should be removed.

ivan-mogilko:
Changed preferences, now has "touch-to-mouse emulation" selection.

Issue #1022, opened 2019-12-26
Editor: Updating translation for a recently translated file may result in loss of work


ghost:
**Describe the bug**
If you have a game opened in the editor, edited translated file in your text editor, **saved it**, then updated translation from the editor with e.g. some new speech lines, your latest work in TRS file may get lost.

**AGS Version**
3.5.0.22, but the bug may be present since early AGS 3.*.

**To Reproduce**
1. Open any game in AGS Editor.
2. Create a new translation by right-clicking on "Translations" and choosing "New translation".
3. Do "Update translation", this will fill the new translation file.
4. Don't close the editor, go to game folder and open newly created TRS file in a simple text editor.
5. Add one translation under any original line. **Save TRS**.
6. Switch to the editor, and add any kind of text to the script, e.g. type `player.Say("hello");` somewhere.
7. **Don't do anything else**, like saving, compiling etc, - nothing.
8. Update translation by right-clicking on its name in the project tree and choosing "Update". Wait until process finishes.
9. Reopen/reload from disk your TRS file.
10. Observe that your last added translation has dissapeared.

**Expected behavior**
If the TRS file was properly saved prior to doing "Update" (translation) command in the editor, all the translations you added should remain, so long as original lines remain at least (tbh I dont remember rules about lines that you remove from the game).

**Additional info**
From what I found, the nature of the problem is this.
Editor keeps a dictionary of translations in memory at all times, and saves it into TRS file on Update, which seem to work correctly. But it only reloads this dictionary at certain points, such as - game compilation or individual translation compilation. However, it does not reload it just before doing Update, which means at that moment the dictionary may not contain newest translations yet (not sync with TRS).

There's definitely a operation mistake here, and it was not noticed (alot) before probably because no one usually does translation simultaneously with updating the game.

Possibly the order of updating TRS should work about like:
1) Generate lines from the game
2) Load existing TRS and intersect with new generated lines, adding translations
3) Save resulting dictionary

AlanDrake:
This is supposedly fixed by #1779, where translations are reloaded before any operation.

PR #1779, opened 2022-09-15
Handle unterminated strings to avoid data corruption

### LINKED ISSUE #1777: Preprocessor: Uncaught unterminated strings
**Describe the bug**
Unterminated strings are compiled without errors.
The fact that they aren't caught will cause unpredictable outcomes when creating or updating translations, and numbering speeches.

**AGS Version**
All versions, including new compiler.

**To Reproduce**
```
player.Say("This line should throw an error.");"
// starting from here code will be inserted into translations instead of actual string
```

**Expected behavior**
An error about an unterminated literal should be thrown instead of being swept under the rug.

**Notes**
I remember we had discussed this very issue ages ago, about translation breaking due to similar circumstances, but I can't find a ticket.

AlanDrake:
The main goal of this PR is to prevent data loss and corruption from dealing with unclosed strings,
- Fixed the preprocessor not reporting unterminated literals
- Abort translation update on error
- Abort auto-numbering speech line on error
- Reload translation froms disk before any update operation to prevent loss of user changes

With these commits all causes of corruption and data loss, when dealing with translations and autonumbering speech, should be covered.
When something fails, the operation is aborted without applying changes.

Might be worth backporting to 3.5.1

fixes #1777,  #1022

### DIFF
--- a/Compiler/preproc/preprocessor.cpp
+++ b/Compiler/preproc/preprocessor.cpp
@@ -190,6 +190,7 @@ namespace Preprocessor {
                     size_t endOfString = text.FindChar(text[i],i);
                     if (endOfString == NOT_FOUND) //size_t is unsigned but it's alright
                     {
+                        LogError(ErrorCode::UnterminatedString, "Unterminated string");
                         break;
                     }
                     endOfString++;
--- a/Editor/AGS.CScript.Compiler/Preprocessor.cs
+++ b/Editor/AGS.CScript.Compiler/Preprocessor.cs
@@ -356,6 +356,7 @@ namespace AGS.CScript.Compiler
 						int endOfString = FindIndexOfMatchingCharacter(text, i, text[i]);
 						if (endOfString < 0)
 						{
+							RecordError(ErrorCode.UnterminatedString, "Unterminated string");
 							break;
 						}
 						endOfString++;
--- a/Editor/AGS.Editor/Components/TranslationsComponent.cs
+++ b/Editor/AGS.Editor/Components/TranslationsComponent.cs
@@ -218,6 +218,11 @@ namespace AGS.Editor.Components
         private void DoTranslationUpdate(IList<Translation> translations)
         {
             _agsEditor.SaveGameFiles();
+            
+            // reload current translations, in case the trs files were modified externally
+            foreach (Translation translation in translations)
+                translation.LoadData();
+
             CompileMessages messages = (CompileMessages)BusyDialog.Show("Please wait while the translation(s) are updated...", new BusyDialog.ProcessingHandler(UpdateTranslationsProcess), translations);
             _guiController.ShowOutputPanel(messages);
             if (!messages.HasErrors)
--- a/Editor/AGS.Editor/Components/TranslationsComponent.cs
+++ b/Editor/AGS.Editor/Components/TranslationsComponent.cs
@@ -186,7 +186,11 @@ namespace AGS.Editor.Components
         {
             List<Translation> translations = (List<Translation>)translationList;
             TranslationGenerator generator = new TranslationGenerator();
-            CompileMessages errors = generator.CreateTranslationList(_agsEditor.CurrentGame);
+            CompileMessages messages = generator.CreateTranslationList(_agsEditor.CurrentGame);
+
+            if (messages.HasErrors)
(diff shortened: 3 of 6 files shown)


ericoporto:
Can you add the line here too:
https://github.com/adventuregamestudio/ags/blob/f4b33d29bdff8b08880596ef8c069b1223c98f19/Compiler/preproc/preprocessor.cpp#L193

I believe it's same in the C# version of the preprocessor.

AlanDrake:
Added the same fix for the cpp preprocessor

Editor/AGS.CScript.Compiler/Preprocessor.cs
```diff
@@ -356,6 +356,7 @@ private string RemoveComments(string text)
 						int endOfString = FindIndexOfMatchingCharacter(text, i, text[i]);
 						if (endOfString < 0)
 						{
+							RecordError(ErrorCode.UserDefinedError, "Unterminated string");
```
ivan-mogilko, line 359: UserDefinedError is an incorrect error type for this; it's for errors "created" by scripters with #error command, not actual mistakes made by user.

Editor/AGS.Editor/Components/TranslationsComponent.cs
```diff
@@ -187,6 +187,10 @@ private object UpdateTranslationsProcess(IWorkProgress progress, object translat
             List<Translation> translations = (List<Translation>)translationList;
             TranslationGenerator generator = new TranslationGenerator();
             CompileMessages errors = generator.CreateTranslationList(_agsEditor.CurrentGame);
+
+            if (errors.Count > 0)
```
ivan-mogilko, line 191: `errors` here is CompileMessages object, which may or not contain actual errors (it may be also warnings or other notifications). Looking at the CompiledMessages class, it provides `HasErrors` property to tell if there were any errors.
Same `HasErrors` property should be checked below in `DoTranslationUpdate`.

Editor/AGS.Editor/Components/TranslationsComponent.cs
```diff
@@ -216,7 +220,10 @@ private void DoTranslationUpdate(IList<Translation> translations)
             _agsEditor.SaveGameFiles();
             CompileMessages errors = (CompileMessages)BusyDialog.Show("Please wait while the translation(s) are updated...", new BusyDialog.ProcessingHandler(UpdateTranslationsProcess), translations);
             _guiController.ShowOutputPanel(errors);
-            _guiController.ShowMessage("Translation(s) updated.", MessageBoxIcon.Information);
+            if (errors.Count == 0)
+                _guiController.ShowMessage("Translation(s) updated.", MessageBoxIcon.Information);
+            else
+                _guiController.ShowMessage("Translation(s) update failed.", MessageBoxIcon.Error);
```
ivan-mogilko, line 231: Maybe add something like "Check output panel for more information" here? because if output panel is not shown, then the user will wonder why the process has failed at all.

Editor/AGS.Editor/TextProcessing/SpeechLinesNumbering.cs
```diff
@@ -72,6 +72,18 @@ private void PerformNumbering(Game game)
 				game.GlobalMessages[i] = processor.ProcessText(game.GlobalMessages[i], GameTextType.Message, Character.NARRATOR_CHARACTER_ID);
 			}
 
+            if (_errors.Count > 0)
```
ivan-mogilko, line 75: Similarily as above, it's better to use `_errors.HasErrors` here, as the list may contain other messages.

Compiler/preproc/preprocessor.cpp
```diff
@@ -190,6 +190,7 @@ namespace Preprocessor {
                     size_t endOfString = text.FindChar(text[i],i);
                     if (endOfString == NOT_FOUND) //size_t is unsigned but it's alright
                     {
+                        LogError(ErrorCode::UserDefinedError, "Unterminated string");
```
ivan-mogilko, line 193: Same here, `UserDefinedError` is only for #error preprocessor commands, not for actual user's mistakes.

AlanDrake:
All suggestions implemented.

ivan-mogilko:
Both issues seem to be fixed.

PR #1781, opened 2022-09-17
Android: remove deprecated/unused preferences

ivan-mogilko:
This removes following preferences that no longer have implementation in the engine:
* Sound - Use multithreading (cannot be toggled now, depends on compilation flags)
* Sound - samplerate (was not exposed to menu)
* Midi options group

### DIFF
--- a/Android/README.md
+++ b/Android/README.md
@@ -97,24 +97,15 @@ The preferences modifies the `android.cfg` file, either the local or global one.
 
 #### Controls `[controls]`
 
+- **Touch-to-mouse emulation** : Off, One Finger, Two Fingers
+  `mouse_emulation = 0, 1, 2`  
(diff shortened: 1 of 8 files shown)


ericoporto:
Not sure if you have to do it here, but other places that have these preferences too
- [Android README here](https://github.com/adventuregamestudio/ags/tree/master/Android#preferences).
- [Editor generated android.cfg](https://github.com/adventuregamestudio/ags/blob/318045de7aef415daf7084891b829ba00956c338/Editor/AGS.Editor/BuildTargets/BuildTargetAndroid.cs#L36)
- The sample [`android.cfg`](https://github.com/adventuregamestudio/ags/blob/master/Android/misc/android.cfg)

ericoporto:
All looks good. Tried the APK and works too.

PR #1782, opened 2022-09-17
Editor: scintilla automatically highlights braces in current position

ericoporto:
I am a bit confused, according to [rongel in the forums](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59842.msg636649791#msg636649791) we had this feature previously. But I just checked and we didn't. I did found some code that is meant to provide some brace highlighting while typing, but it doesn't work for me in the 3.5.1 branch. 

I honestly don't remember this feature, but in case we actually had it, this works and is a small change to merge.

### DIFF
--- a/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
+++ b/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
@@ -959,7 +959,7 @@ namespace AGS.Editor
             }
         }
 
-        private Tuple<int, int> GetBraceAndMatchingBracePositions(bool beforeCursor, bool afterCursor)
+        private Tuple<int, int> GetBraceAndMatchingBracePositions()
         {
             if (InsideStringOrComment(false))
                 return Tuple.Create(INVALID_POSITION, INVALID_POSITION);
@@ -969,8 +969,8 @@ namespace AGS.Editor
             int currentPos = scintillaControl1.CurrentPosition;
             int prevChar = scintillaControl1.GetCharAt(currentPos - 1);
             int curChar = scintillaControl1.GetCharAt(currentPos);
-            bool isBraceBefore = beforeCursor && (prevChar == '{' || prevChar == '}' || prevChar == '(' || prevChar == ')');
-            bool isBraceAfter = afterCursor && (curChar == '{' || curChar == '}' || curChar == '(' || curChar == ')');
+            bool isBraceBefore = (prevChar == '{' || prevChar == '}' || prevChar == '(' || prevChar == ')');
+            bool isBraceAfter = (curChar == '{' || curChar == '}' || curChar == '(' || curChar == ')');
             
             if (isBraceBefore)
             {
@@ -985,7 +985,7 @@ namespace AGS.Editor
 
         public void DoIdentationAlignAfterBrace()
         {
-            Tuple<int, int> pos = GetBraceAndMatchingBracePositions(true, true);
+            Tuple<int, int> pos = GetBraceAndMatchingBracePositions();
             int currentPos = pos.Item1;
             int matchPos = pos.Item2;
             if (matchPos >= 0)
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
> I am a bit confused, according to [rongel in the forums](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59842.msg636649791#msg636649791) we had this feature previously. But I just checked and we didn't. I did found some code that is meant to provide some brace highlighting while typing, but it doesn't work for me in the 3.5.1 branch.

It works for me in 3.5.1.

ivan-mogilko:
Testing the latest commit, I noticed following:
- Highlighting works for both curved and round brackets, but only when you put cursor at the already existing, previously typed bracket. It does not highlight immediately while you are typing the expression (as it did in 3.5.1).
- There's a very strange regression: whenever I type a curved bracket, the whole line is moved to the left, to the zero indentation level.

ericoporto:
I separated the indentation alignment from the brace matching, because they use really similar pattern matching, they were glued together.

I think it's working. I also made so it doesn't trigger in the middle of a dialog string.

ericoporto:
moved the brace matching graphic update to happen after everything, this fixes any differences between dialog and the script editor.

ericoporto:
@ivan-mogilko is the highlighting different for you after the last commit, or is it working alright now?

ivan-mogilko:
With the last commit, the highlighting seem to work correctly in _dialog scripts_: it works both while typing the brackets and when placing cursor at the already existing brackets.

In the regular scripts highlighting while typing still does not work.

ivan-mogilko:
Hmm, I found something else wrong. The red highlight of a extra closing bracket does not work correctly in either scripts. Instead of highlighting a bracket itself it highlights the symbol right after the bracket.

ericoporto:
> In the regular scripts highlighting while typing the closing braces still does not work, neither for round nor curved ones. It works when moving cursor around though, placing the cursor at both opening or closing braces.

I think there's a race condition at hand somehow. Because it works when I debug here...

> Hmm, I found something else wrong. The red highlight of a extra closing bracket does not work correctly in either scripts. Instead of highlighting a bracket itself it highlights the symbol right after the bracket.

This one is because I need to pass false [here](https://github.com/adventuregamestudio/ags/pull/1782/commits/4039491f097b684960d896027a15a449a1b0e8d4#diff-52ab1073717926111082758762c5cc6cfe18c3e8cdb6a61c8b616cea0a037bc8R1005).

ivan-mogilko:
So, this is quite strange, but after rebuilding this branch and testing again, I no longer experience the missing highlight. Also tested the editor built by CI, and it also is working correctly.

The only problem remaining is incorrect red highlight.

ericoporto:
I have the missing highlight with the build from CI here, so it has to be a race condition somewhere, but I haven't yet had idea how/time to investigate it further.

ericoporto:
For some reason I don't understand yet, UpdateUI can be sometimes called twice, constantly, so I removed my previous check that would only check for braces when position changed, because when position did not changed, if UpdateUI was called twice, then the highlight would be removed by the logic that resets the brace position from `_braceMatchVisible`.

ivan-mogilko:
Well, judging by the CI build, the yellow highlight of correct braces is working, but the red highlight still has errors.

ivan-mogilko:
I tried looking into this; and apparently the problem with red highlight was that the `GetBraceAndMatchingBracePositions` function did not know which way there's a brace under cursor (before or after). Since it was testing both ways and failing to find a match every time, it always highlighted the second case (after cursor) regardless of where the brace actually is.

I tried to refactor this further by making this function test for braces itself, and it seems to work fine: https://github.com/ivan-mogilko/ags-refactoring/commit/310c9de0733f4bb4c43b11c2654f1d406d21afb4

ericoporto:
applied the commit and it works!

---

While playing around I found a scintilla bug where brace matching internal scintilla code can consider characters in a string, managed to reproduce in notepad++ : https://github.com/notepad-plus-plus/notepad-plus-plus/issues/362 , I could not curiously reproduce in a simple or small code, but I can reproduce in this module:

https://github.com/dkrey/ags_customdialog/blob/0d88e9741f086b7df43bd559cb868b76106caf5c/CustomDialogGui.asc#L252

Loading this module in the editor, and then typing a single `(` in line 184, matches it to the one at the right of the string above, had to use the VS debugger for this. Because I could not find any other way to cause this, I don't think this is significant.

PR #1784, opened 2022-09-20
Android: support relative speed factor for touch-to-mouse emulation

### LINKED ISSUE #862: Android: mouse sensitivity
Would it be possible to add a RELATIVE MOUSE sensitivity setting? I only play games with relative mouse, to the point of actually choosing running games through AGS rather than their native android ports (I hate point and click on a the phone). That's one thing that is severely missing from an otherwise great port.

ivan-mogilko:
Resolves #862.

Use the mouse speed config option when generating emulated touch-to-mouse events.

@ericoporto added your commit with the option in AGS Player's preferences.
image

### DIFF
--- a/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/PreferencesActivity.java
+++ b/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/PreferencesActivity.java
@@ -13,6 +13,7 @@ import androidx.preference.Preference;
 import androidx.preference.PreferenceCategory;
 import androidx.preference.PreferenceFragmentCompat;
 import androidx.preference.PreferenceGroup;
+import androidx.preference.SeekBarPreference;
 import androidx.preference.SwitchPreference;
 
 import androidx.preference.PreferenceManager;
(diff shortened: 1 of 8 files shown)


ericoporto:
Hey, after the fix1 it works correctly in my phone. :)

ivan-mogilko:
Yes, looks like this is working.

The only problem I see now is that the preference is reset to 1.0 every time I open the menu...

ericoporto:
Ah, that summary should not be visible, I actually wanted to trigger the update of the summary once before any changes happens, so that it shows the actual value from the start, but don't know yet how to do that.

Also, I won't be on computer all day today, so can't fix this right now.

ericoporto:
Ah, that summary should not be visible, I actually wanted to trigger the update of the summary once before any changes happens, so that it shows the actual value from the start. 

Fixed it here: https://github.com/ericoporto/ags/commit/bfb35375e40026e64a52b2d98f2ec3909224d8a4

ivan-mogilko:
>  I did not notice that distance change from the last commit, meaning using it, it feels the same and it was alright. But if this is something that needs adjustment in the future we may make it a config entry.

I was just experimenting, seeing what is more logical. The initial code was outright wrong anyway, changed it to something else, but it should work same overall.

PR #1785, opened 2022-09-23
AGS3: Editor UI refresh

AlanDrake:
Alright, this is the editor refresh stuff for AGS3.
I fixed a few things and improved theme compatibility with the old format, though old themes will need to add the property for the sprite import button for it to work proper.

My new themes from the other PR will need some additions to play well with the old VS2005 skin, since they offload some stuff to the dockpanelsuite base theme.

Once this is merged I'll collect all themes available out there and upgrade them. I have a converter to turn colors into hex strings.
Until then, old themes should continue working as expected.

EDIT: Refreshed old themes - [Download](https://www.dropbox.com/s/xghlzr5ic99wlh1/AGS%20Refreshed%20Current%20Themes.zip?dl=0)


ericoporto:
This is working, but it needs updated themes - it will still load both old and new ones, but there's some details to adjust. I suggest after this is in, we create a new repository in the organization to hold AGS Editor Themes.

ivan-mogilko:
Is there any information on what does "new theme format" mean? What has changed?

Does it still load existing themes? This is important for 3.6.0, as I've seen few users created their custom themes for AGS in the past (posted on forums).

ivan-mogilko:
@ericoporto 
> This is working, but it needs updated themes - it will still load both old and new ones, but there's some details to adjust.

Could you please elaborate this, does this mean that we cannot distribute this change without new themes, some ui elements are missing (not colored)?

ericoporto:
When using an old theme, it requires you have specified the individual controls, so even though the commit 
[`Theme: Missing sprite import button`](https://github.com/adventuregamestudio/ags/pull/1785/commits/3707497a3decdc0989d31c35b6492260e478058a) is there, it won't work unless the theme actually specify it's colors. Below is **VisualStudioDark** (a theme for the current ags3 editor).

image

Alternatively, you can use the new themes, which use globals, which can then be used as fallback. But the new themes were made with a different VS base theme (from AGS 4 branch of Alan's PR), so they need to be adjusted to show the colors. Below is the **Dark** theme from the other PR.

image

AlanDrake:
Well, I already pointed those issues in the first post.
After I prepare the updated themes I'll also write some documentation on the available properties.

ericoporto:
I think theoretically we just need the VisualStudioDark theme updated to avoid regressions, as that is the only theme made available with the release, currently, in the forums.

AlanDrake:
I added on top a link to the refreshed old themes.
Another change I had forgotten about was that inconsistent "category-fore" which was "category" elsewhere, so I kept just "category". All already accounted for in the refreshed themes.

Anyway, are you guys aware of the dead theme link in the 3.6 thread?

ericoporto:
The refreshed themes fixes all mentioned issues in my comment.

> are you guys aware of the dead theme link in the 3.6 thread?

I was not aware. If there's a theme repository I can make releases there or create a gh-page to make they easy to check/download.

ivan-mogilko:
Fixed the theme link, it had a typo apparently.

ivan-mogilko:
I finally gave this a test.

There's a serious problem with the theming perfomance in debug mode, probably due to the theming code throwing too many exceptions; at least this is what I experienced with the older theme. Every time I open a pane or even a menu the program freezes for a few seconds.

Question: is theming format change necessary for ags3, can we leave it for ags4? I guess the code and perfomance could be improved in the future, but as 3.6.0 is practically complete, this may be annoying to have to add another potential issues in it.

AlanDrake:
Technically speaking, adding the sprite import button will also require a format change, and therefore an updated theme.
If the debug perf is that bad I must try optimizing the code,

ivan-mogilko:
> Technically speaking, adding the sprite import button will also require a format change, and therefore an updated theme.

It will require just an additional entry, but not any other bigger changes.

AlanDrake:
This last commit should address the performance issues in debug mode. I reduced the likelihood of triggering exceptions without changing too much.
I tested it by reducing my cpu clock by half, so I hope it's good enough in your machine as well.

ivan-mogilko:
Yes, it seems to work about as fast as before now.

ivan-mogilko:
Something I have missed; there's at least one different theme key, for the GeneralSetting's category (foreground) color.

Previously it was called `category-fore`, now it's called `category`.

Don't know if there are any other differences which I did not notice yet.

This breaks existing themes.
While the theme file itself may be adjusted, this brings a more serious question: how to distinguish old themes and the new ones? Is there any "tag" inside the json file that tells the "version"? The complication is that all editor versions are reading from the same folder. Was this issue discussed previously?

For now the only workaround I see is to make themes that have both keys, so that both 3.5.1 and 3.6.0 could use same theme file. Otherwise we'd have to make theme files have distinction in their names, and tell users to only select proper themes, as AGS cannot filter them itself.

AlanDrake:
I already bumped up the versions number in the updated old themes.
Do we really need to add the "category-fore" as fallback? As I said previously this seemed a mistake in the original format because it was called just "category" elsewhere so it only affects the category name in the general settings. 
Adding it to the string array is trivial but people need to upgrade anyways for the sprite import button.

ivan-mogilko:
> Do we really need to add the "category-fore" as fallback?

There are options, one of the option is to have this fallback inside the theme file, rather than a program.

Suppose you have 3.5.1 and 3.6.0 installed at the same time. They both read themes from the same directory. If you put old theme there it won't work for 3.6.0. If you put an updated theme then it won't work for 3.5.1.

This means that either theme files need to include the version in their names to be able to separate them, or there should be one universal theme file meant for both 3.5.1 and 3.6.0.

Also optionally AGS might distinguish theme versions and don't show obsolete/unsupported ones in the selection list.

AlanDrake:
You're right, the themes are shared globally.... sometimes I forget people still sue 3.5.1.
So the easiest is adding the version in the filename.
Not showing unsupported themes could require using a different extension, or implementing some system on 3.5.1, the latter might not be worth it.

Alright, I'll add "category-fore" as a fallback in the meanwhile.

Maybe using a different extension could be the cleaner option to keep things separated. What do you say?

ivan-mogilko:
Imho adding fallbacks is the cheapest solution at the moment, since this seem to be only 1 entry (at least from what I noticed).
This may imply:
1. a fallback in the program: this will let 3.6.0 keep working with the older themes properly;
2. a fallback in the theme file: this will let to distribute 1 file for both older and newer editors. Since the files are jsons and may contain optional entries, that should not cause any trouble.

> Maybe using a different extension could be the cleaner option to keep things separated. What do you say?

I don't think the changes are too big to justify the change of the extension?

Overall, I believe that theme management and compatibility is a whole separate topic, and it's best to not do any rushed decisions here. It also best to be left for ags4, or at least later 3.6.* updates (if there will be any), in my opinion.

AlanDrake:
Fallback added. I also spotted and fixed a small regression in the fallback logic because I had forgotten to adjust the returns not to rely on exceptions after the last commit.

PR #1792, opened 2022-09-27
Editor: fixed how dialog script is styling the comments and strings

### LINKED ISSUE #1783: Incorrect syntax highlighting for commented AGS script in dialog
**Describe the bug**
Within the dialog script editor it looks like the ordering of syntax highlighting that is being applied to embedded AGS script is incorrect, leading to portions of the script being incorrectly highlighted.

**AGS Version**
3.5.1.19

**Screenshot**
dialog_syntax

**To Reproduce**
Create a dialog script with the following content:
```
Roger: I'm looking for my key.
  //if (player.HasInventory(iKey)) {
  //    player.Say("Actually, I'd better go.");
  //    return RUN_DIALOG_STOP_DIALOG;
  //}
```
The commented out code will not be fully highlighted as being commented out. The highlighting of the comment opening `//` on the 2nd line will also be incorrect.

**Expected behavior**
Comment opening syntax and commented out code should both be highlighted as a comment.

ivan-mogilko:
Fixes #1783

There have been a number of inconsistencies in how the dialog scripts are styling comments and strings; or rather in how they detect when to style.

Firstly, there have been this silly mistake in condition here, which made the condition make no sense (it will always be true for letter or digits, as style `!=` are ORed)
https://github.com/adventuregamestudio/ags/blob/084f3f281359560a94ed819ff649edbf8bf9c4a0/Editor/AGS.Editor/Panes/ScintillaWrapper.cs#L2517-L2518

Secondly a number of clauses below never tested for being inside a comment.

~Introduced couple of booleans for more tidy code, and made sure that all necessary cases test for comments.
I guess the code could be reorganized to simplify these checks, but I decided to avoid further refactor at the moment.~

EDIT: Done a larger refactor of this code, because there's no actual need to test 10 cases when we are inside a string or comment. So instead I moved tests for being inside comment and string (testing for the end) up, and restructured the tests.

The older variant, with a simplier fix, may be found here: https://github.com/ivan-mogilko/ags-refactoring/commit/9934c3716b527f151cb078dd6e496ad4522dd28b

### DIFF
--- a/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
+++ b/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
@@ -2514,8 +2514,84 @@ namespace AGS.Editor
             {
                 char c = (char)scintillaControl1.GetCharAt(i);
 
-                if (!Char.IsLetterOrDigit(c) && (stylingMode != Style.Cpp.Comment ||
-                    stylingMode != Style.Cpp.CommentLine || stylingMode != Style.Cpp.String))
+                // Handle end of line
+                if (c == '\n')
+                {
+                    onNewLine = true;
+                    atNewWord = false;
+                    onScriptLine = false;
+                    // The multiline comments and strings should keep their style;
+                    // everything else is cut of at the end of the line
+                    if (stylingMode != Style.Cpp.Comment && stylingMode != Style.Cpp.String)
+                    {
+                        if (laststyle < i)
+                        {
+                            scintillaControl1.SetStyling(i - laststyle, stylingMode);
+                            laststyle = i;
+                        }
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Changed my mind and did a somewhat larger refactor of the function `StyleDialogScript`. Because there's no actual need to test 10 cases when we are inside a string or comment. So instead I moved tests for being inside comment and string (testing for the end) up, and restructured the tests.

The older variant, with a simplier fix, may be found here: https://github.com/ivan-mogilko/ags-refactoring/commit/9934c3716b527f151cb078dd6e496ad4522dd28b

AlanDrake:
I tried it and it seemed to work fine.

PR #1798, opened 2022-10-03
Common: fixed ReadEncInt32() on big-endian systems

dwatteau:
Hi,

This small fix comes from ScummVM testing on a big-endian system. Games using `MFLUtil::ReadV21()` would try to allocate gigantic amounts of memory, because some values were not read properly.

`MFLUtil::ReadEncInt32()` just didn't use the return value of `SwapBytesInt32()`, and so this value was still represented with the wrong endianness.


ivan-mogilko:
Right, thank you; I will also backport this to release-3.5.1 branch.

Issue #1797, opened 2022-10-03
Game suspicious mind crashes on start menu


i30817:
**Describe the bug**

```
Adventure Game Studio v3.6 Interpreter
Copyright (c) 1999-2011 Chris Jones and 2011-2022 others
ACI version 3.6.0.35

Initializing backend libs
SDL Version: 2.0.20
Initializing game data
Located game data pak: /media/i3/Mordred/Games/Adventure Game Studio/Suspicious Mind/MAGS March 2019.exe
Opened game data file: game28.dta
Game data version: 49
Compiled with: 3.4.1.14
Startup directory: /media/i3/Mordred/Games/Adventure Game Studio/Suspicious Mind
Data directory: /media/i3/Mordred/Games/Adventure Game Studio/Suspicious Mind/
Setting up game configuration
Unable to init voice pack 'speech.vox', file not found or of unknown format.
audio.vox found and initialized.
Initializing TTF renderer
Initializing mouse: number of buttons reported is 3
Install timer
Audio driver: pulseaudio
AudioCore: opened device "Default OpenAL playback device"
Supported sound decoders:
 - Play modules through ModPlug : 669,AMF,AMS,DBM,DMF,DSM,FAR,GDM,IT,MDL,MED,MOD,MT2,MTM,OKT,PTM,PSM,S3M,STM,ULT,UMX,XM,
 - MPEG-1 Audio Layer I-III : MP3,MP2,MP1,
 - Microsoft WAVE audio format : WAV,
 - Audio Interchange File Format : AIFF,AIF,
 - Sun/NeXT audio file format : AU,
 - Ogg Vorbis audio : OGG,
 - Creative Labs Voice format : VOC,
 - Raw audio : RAW,
 - Shorten-compressed audio data : SHN,
 - Free Lossless Audio Codec : FLAC,FLA,
Install exit handler
Initialize legacy path finder library
Game title: 'Suspicious Mind'
Game GUI version: 118
Requested script API: v3.4.1 (5), compat level: v3.2.1 (0)
Checking for disk space
Game native resolution: 320 x 240 (32 bit)
Graphic settings: driver: OGL, windowed: no, screen size: 0 x 0, game scale: round
Running OpenGL: 4.6
Graphics mode set: 1920 x 1080 (32-bit) fullscreen desktop
Mouse speed control: disabled, unit: 1.000000, user value: 2.000000
Multitasking mode set: 0
Setting up window
Multitasking mode set: 0
Initialize sprites
[...]

PR #1802, opened 2022-10-04
Don't recreate overlay object in call to Overlay.SetText()

ivan-mogilko:
This addresses a problem which I found while was fixing #1797 (it was not directly related to 1797 though).

**The problem**

Previously `Overlay.SetText()` was passing into `CreateTextOverlay()` -> `Overlay_CreateTextCore()` -> `_display_main()`.
This effectively removed and recreated an internal overlay object with different image but the same ID.

Also, previously, this method could have been only called for custom overlays in practice. For these the script object was not touched, and kept the correct overlay reference.

The situation has changed when we let user script to access some of the internally created overlays, such as Text and Portrait overlays created by a blocking Say command (`Speech.TextOverlay` and `Speech.PortraitOverlay`). When these overlays are recreated, they also must be completely destroyed, invalidating any existing script references. Hence calling `_display_main()` from `Overlay.SetText()` is no longer acceptable, as it leads to a full overlay recreation, and possible duplication of a script reference with two (or more) separate managed handles (this is almost like, two shared pointers owning same object). An alternative was to invalidate the old script reference, which might greatly surprise a user... and makes no sense from the scripting perspective.

```
function repeatedly_execute_always()
{
    if (Speech.TextOverlay) {
        Overlay* o = Speech.TextOverlay;
        o.SetText(100, eFontNormal, 123456, "a new text");
        // overlay will suddenly disappear here in the latest master,
        // because `o` pointer is detached from internal reference count,
        // and engine will destroy overlay as soon as pointer is lost
    }
}
```

To fix this problem, here we replace a call to `CreateTextOverlay()` with manual recreation of a textual image, and assigning a new image to the same overlay. This **does not change anything from the script's POV**, as SetText method already suggests that the overlay reference remains valid (previously only the internal object was secretly replaced).

With this change Overlay.SetText method should correctly work with overlay pointers copied from Speech.TextOverlay (and maybe even Speech.PortraitOverlay, lol). E.g.:

```
int counter;
function repeatedly_execute_always()
{
    if (IsKeyPressed(eKeySpace)) {
        if (Speech.TextOverlay) {
            Overlay* o = Speech.TextOverlay;
            o.SetText(100, eFontNormal, 123456, "text: %d", counter);
            counter++;
        }
    }
}
```

**Implementation notes**

1. Had to refactor `_display_main()` by picking out a function I called `create_textual_image()`. This function generates a bitmap with text on it and adjusts the position. `_display_main` calls it internally, and then creates overlay and optionally waits for a skipping player input, as usual.
2. `Overlay_SetText()` now does not call `CreateTextOverlay()` anymore, but instead calls `create_textual_image()` directly, skipping the whole chain of functions, and lots of unnecessary code. For this I also had to copy two small pieces from `CreateTextOverlay` and `Overlay_CreateTextCore`, which adjust some parameters, because I could not quickly figure out a good way to share this code; so I'd leave this for future refactor.

Normally I don't like refactoring important functions like `display_main` right before the final version release. Here i tried to keep changes to minimum, and they are mostly moved code (and some lines were auto-styled by VS, but I left that).

### DIFF
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -60,7 +60,6 @@ extern SpriteCache spriteset;
 
 int display_message_aschar=0;
 
-
 TopBarSettings topBar;
 struct DisplayVars
 {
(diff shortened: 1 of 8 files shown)


ericoporto:
https://user-images.githubusercontent.com/2244442/193946508-4b228aeb-f3d6-4553-9f70-aa20fd1c4eef.mp4

<details><summary>tried this</summary>


    // room script file

    function hGlowingOrb_AnyClick()
    {
      player.Say("Hi rock!");
      player.Say("Looking like a nice rock today.");
      player.Say("Could say you are glowing...");
    }

    int counter;

    #define PORTRAIT_YMIN 5
    #define PORTRAIT_YMAX 30
    int plast = -1;
    int pmove = 1;
    function repeatedly_execute_always() {
      if (Speech.PortraitOverlay != null) {
        Overlay* p = Speech.PortraitOverlay;
        
        if (plast >= 0) p.Y = plast;
        p.Y += pmove;
        if (p.Y < PORTRAIT_YMIN) pmove = 1;
        if (p.Y > PORTRAIT_YMAX) pmove = -1;
        plast = p.Y;
      }
      
      if (mouse.x > 120) {
        if (Speech.TextOverlay) {
            Overlay* o = Speech.TextOverlay;
            o.SetText(100, eFontSpeech, 123456, "text: %d", counter);
            counter++;
        }
      }
    }

    #define MAX_OVRS 1024

    int k=0;
    Overlay* ovr[MAX_OVRS];
    int ovr_ori_x[MAX_OVRS];
    int ovr_ori_y[MAX_OVRS];
    float ovr_ori_d[MAX_OVRS];

    float ang;

    function room_RepExec()
    {
      ang += Maths.Pi/128.0;
      
      if(ang == 2.0 * Maths.Pi)
      {
        ang = 0.0;
      }
      
      for(int i=0; i < k; i++)
      {
        int x = ovr_ori_x[i] + FloatToInt(Maths.Cos(ang)*32.0);
        int y = ovr_ori_y[i] + FloatToInt(Maths.Sin(ang)*32.0);
        
        Overlay *ov = ovr[i];
        
        float ori_w = IntToFloat(ov.GraphicWidth);
        float ori_h = IntToFloat(ov.GraphicHeight);
        
        int w = FloatToInt( ori_w * ( Maths.Cos(ovr_ori_d[i] + ang) + 1.0) );
        int h = FloatToInt( ori_h * ( Maths.Sin(ovr_ori_d[i] + ang) + 1.0) );
        
        ov.X = x;
        ov.Y = y;
        ov.Width = w;
        ov.Height = h;   
      }
      
      
      if(k >= MAX_OVRS) return;
      
        
      ovr_ori_x[k] = Random(320);
      ovr_ori_y[k] = Random(200);
      ovr[k] = Overlay.CreateGraphical(ovr_ori_x[k], ovr_ori_y[k], 2041 + Random(1), true);
      ovr_ori_d[k] = ang;
      
      k += 1;
    }
</details>

Looks working. Not sure how to test this overall.

PR #1806, opened 2022-10-08
AGS4: Additional UI fixes and experiments

AlanDrake:
A set of experiments meant to fix or improve some stuff about the UI:
* Editor: anti-flicker for ComboBoxCustom
* Editor: CustomPropertyGrid with exposed toolrenderer (**so it can be themed**)
* Editor: Remove extra borders (**that were ugly for theming**)
* Editor: better default layout (**less horizontal space for output/find/callstack**)
* ~~Editor: restore proper panel dock fill behavior (**fixes dock fill and removes padding hacks**)~~ already merged on master
* Editor: word wrap toggle
* ~~Editor: Show BusyDialog without focus (**prevents titlebar flashing during compilation**)~~ postponed for the future
* Editor: Change Show Stats shortcut from F2 to CTRL+F2
* Editor: enable F2 rename for treeview labels
* Editor: fix double click triggering the wrong nodes in treeview

So far they seem to work without bad side effects, so I might merge them after some more testing.
Regarding extra borders, the scintilla control didn't use to have 3d borders, there must have been a regression.


AlanDrake:
Okay, I think I fixed all the things that were annoying me.
Some stuff might be worth backporting.

AlanDrake:
Note to self: not giving the busy-window focus makes it possible to start another compile while it's already compiling.

ivan-mogilko:
@AlanDrake you've mentioned on the forums, that some of these fixes may fix a problem reported for 3.6.0:
https://www.adventuregamestudio.co.uk/forums/ags-engine-editor-releases/ags-3-6-0-release-candidate-rc1/msg636651138/#msg636651138

Can you confirm this? (I was not able to reproduce the problem myself) If so, could you open a separate PR with only necessary fixes for 3.6.0?

AlanDrake:
> Can you confirm this?

Yes, sorry. I could swear I had seen the commit having been picked already, but I see my brain had lied to me.
I created the PR #1831 and it does indeed also fix the extra space on top when the incriminated panels are floating.

Another commit that should be safe is the one where I remove unnecessary borders, but if you only want to address regressions then the border should be only removed from the scintilla control (I think the 3d border was introduced after the scintilla upgrade).

I've been quite busy so I couldn't give AGS enough love.
The only commit here appears to cause bad side effects is the busy dialog focus one... I originally wanted to keep and update the current busydialog but it wasn't as easy as it looked because it spawns on different threads.
Then not focusing seemed a good alternative to prevent the title bar flashing from repeated focus switching, but I had overlooked that AGS does not have guards against relaunching recompile again while it's already compiling.

Maybe I'll remove that one and rework it in the future...

ivan-mogilko:
Started testing this...

Don't know how to test:
* anti-flicker for ComboBoxCustom
* CustomPropertyGrid with exposed toolrenderer
* Remove extra borders
* restore proper panel dock fill behavior
* fix double click triggering the wrong nodes in treeview (can't remember if I ever had any problem with this)

Tested and works:
* word wrap toggle
* Change Show Stats shortcut from F2 to CTRL+F2
* enable F2 rename for treeview labels

Tested, but have issues:
* "better default layout (less horizontal space for output/find/callstack)"
The menu option "Layout - Reset to Defaults" seem to keep the old behavior, or maybe does not affect this particular composition at all. The only way to get the "new" default layout is again to delete the Layout.xml.

AlanDrake:
> Don't know how to test:
> 
>     * anti-flicker for ComboBoxCustom
Not sure if this was easily visible on windows 7, but on 10 the rendering competed with the system causing light gray flashing on the main control, if you hovered over and out quickly.
That was solved by that SetStyle block which tells the system that WM_PAINT is meant to do all the work.
The only flickering left now is on the expanded menu, while moving the mouse cursor up/down the options, but it's barely noticeable so I didn't bother.

>     * CustomPropertyGrid with exposed toolrenderer
This is just to allow theming on the tool buttons of the property grid, which are normally unaccessible.
If the theme colors are being applied, it's working.

>     * Remove extra borders
Can be easily seen with a dark theme, if you compare before/after on photoshop or similar. I just got rid of a few unaesthethic unthemed borders, nothing major.
The only thing to note is that the thick border around scintilla's control was a regression.

>     * restore proper panel dock fill behavior
Undocked output/callstack/results/propertygrid panel used to have ugly padding on top.
Stacked output/callstack/results panel bar on bottom used to cover the lower part of the list.
The padding I removed was a hack that was only there because the dock fill property wasn't working proper, and it wasn't working because frmMain was redefining the Dock property externally, causing side effects.
The order in which Dock is set is important for the components within the control to size correctly in the layout. 

>     * fix double click triggering the wrong nodes in treeview (can't remember if I ever had any problem with this)
Easy to test by draggin the cursor before releasing the mouse button.
The implementation was incomplete because it didn't account that the target on release could be different from where the one the double-click was initiated on.

> Tested, but have issues:
> 
>     * "better default layout (less horizontal space for output/find/callstack)"
>       The menu option "Layout - Reset to Defaults" seem to keep the old behavior, or maybe does not affect this particular composition at all. The only way to get the "new" default layout is again to delete the Layout.xml.

Here I only changed the order of composition. But you're right, the reset seems to keep memory of the current disposition (be it old or new).
Maybe the reset was always missing a step to actually reset completely, because now that I know what to try, even 3.6 appears to show the same behavior when clicking "Reset to Defaults".

AlanDrake:
Removed the dock fill commit, which has been merged to master by #1831

ivan-mogilko:
> Removed the dock fill commit, which has been merged to master by https://github.com/adventuregamestudio/ags/pull/1831

I wanted to mention, that you really did not have to do this. These are two separate versions, there's no problem in applying same change to both separately.

AlanDrake:
I thought I'd avoid merge conflicts, since it's going to be merged soon anyway

ivan-mogilko:
> I thought I'd avoid merge conflicts, since it's going to be merged soon anyway

I think it's possible that there would not be any merge conflicts in this case. Git usually detects identical changes in two branches; at least that was my experience previously. But it is possible to test locally to double check.

Issue #1505, opened 2022-01-17
Port AGS 3.6.0 to iOS (SDL2 backend)


ericoporto:
The previous Allegro based AGS engine was ported to iOS, and the port is maintained in the release-3.5.1 branch. But the newer SDL2 based AGS that is on the current master branch has not been ported to iOS yet, so the README in iOS directory is not updated.

Porting to iOS should require following the porting guide in SDL2 repository:

[github.com/libsdl-org/SDL/blob/main/docs/README-ios.md](https://github.com/libsdl-org/SDL/blob/main/docs/README-ios.md)

Current MacOS builds use CMake, there are two possible approaches here

- Devise a way to use CMake for iOS AGS project (see [github.com/libsdl-org/SDL/blob/main/docs/README-cmake.md](https://github.com/libsdl-org/SDL/blob/main/docs/README-cmake.md) for iOS info)
- Author Xcode projects for MacOS too (and also iOS)

Authoring the Xcode projects may require additional steps to build the dependencies (these are currently automated when using CMake build system). This was previously done by some scripts and makefiles that have not been updated.

It's recommended that the person doing the port owns at least one iOS device so they can test in real hardware.

Until this is resolved, consider the information in [iOS/](https://github.com/adventuregamestudio/ags/tree/master/iOS) directory not updated.

---

Current roadmap:

- [x] update MacOS Xcode project #1807
- [x] update scripts for libraries
- [x] update iOS Xcode project (#2253 )
- [x] update readme (#2253)
- [x] delete unused files in OSX and iOS directories (#2253)
- [ ] additional fixes to platform and agstouch plugin

ericoporto:
I am currently looking into updating the Xcode projects, this will take a bit of time, but thought of mentioning it here in case someone else is looking into this too.

ericoporto:
Hey, I [started](https://github.com/adventuregamestudio/ags/compare/master...ericoporto:ags:add-new-ios-xcode) this, the branch is here:
https://github.com/ericoporto/ags/tree/add-new-ios-xcode

The new readme is here: [add-new-ios-xcode/iOS/README.md](https://github.com/ericoporto/ags/blob/add-new-ios-xcode/iOS/README.md)

If someone would like to try things out. It's basically you add your game files in the Resource directory in Xcode and then you build and your game should run. The code has lots of things to be cleaned up and adjusted still.

[...]

PR #1807, opened 2022-10-09
MacOS: update xcode project

ericoporto:
This should make the Xcode project be useable again in macOS. This is useful to figure out the iOS port (#1505), as it will require using Xcode.

It requires some manual steps which I detailed in the README.

I can split the contents of this commit in different ones if necessary.

---

When trying to use Xcode to debug AGS, I could not make AGS work for nothing... Turns out Xcode by default adds some parameters to a debugged executable and AGS was picking up these. 

I configured the debug schemes here to not add these parameters, so things are working. But I think maybe we should print the full argv if there's an unused parameter that is preceded by `-`. I will probably open an issue on this later.


ericoporto:
Also updated the macOS makefile instructions, should someone want to build and run it this way. :)

PR #1812, opened 2022-10-12
Editor: do not repopulate the tree when adding new folders

### LINKED ISSUE #1811: Nested folders creation issue
**Describe the bug**
In the Explore Project panel, creating new subfolders can cause unwanted behaviors.

https://user-images.githubusercontent.com/1430815/195325663-b8a8e78f-88a1-41e0-bc50-1ae4ec5d901a.mp4

The target tree gets collapsed and the wrong item is selected for renaming.
It appears to happen when creating new folders deeper than one level.
After the creation, it seems like the editor is blindly trying to get an entry by the name "New folder" instead of just keeping the correct reference to the created one.

**AGS Version**
3.x and 4.x

**To Reproduce**
Create several subfolders deeper than two levels.

**Expected behavior**
Creating a new subfolder should correctly place the selection on the created item without collapsing the current location.

AlanDrake:
The tree view was entirely repopulated in order to sort folders before other entries, but doing so could break focusing on the newly created item due to keys being recreated from scratch.

I added a check to directly insert folders where appropriate, along with an exception for the Audio where the speech icon stays above.

Fixes #1811

### DIFF
--- a/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
+++ b/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
@@ -161,7 +161,6 @@ namespace AGS.Editor.Components
             FolderType newFolder = CreateFolderObject("New folder", parentFolder);
             parentFolder.SubFolders.Add(newFolder);
             string newNodeID = AddNodeForFolder(newFolder);
-            RePopulateTreeView();
             _guiController.ProjectTree.BeginLabelEdit(this, newNodeID);
         }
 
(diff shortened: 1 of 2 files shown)


AlanDrake:
Or the speech icon could stay below so there's no need to have the special case when we are in the main folder of Audio.

ivan-mogilko:
Sorry for delay, I forgot about this pr.

This seem working, so will merge it now.

ivan-mogilko:
Hmm, @AlanDrake , could you also check a similar issue which happens when you drag folders around? It also causes folders to collapse unnecessarily.

AlanDrake:
Sure, I'll take a look at it when I can.

Issue #1514, opened 2022-01-23
AGS4: Attribute `DrawingSurface::DrawingTransparency`


fernewelten:
**Describe the problem**
True-colour space provides  a full-fledged alpha channel (RGBA). This alpha channel isn't easily accessible in AGS games.

In AGS, when you want to draw onto a `DrawingSurface` in true-colour games, you can set the pen __colour__ with `DrawingSurface::DrawingColor`. However, there's no way to set the pen __transparency__. So convoluted workarounds are needed in order to draw semi-transparent shapes onto a surface.

**Suggested change**
Introduce `int DrawingSurface::DrawingTransparency` to set the alpha channel value of the drawing pen.

The range of this attribute would need to be discussed: 
AFAIK, the real range of the alpha channel in true-colour space is `[0 .. 255]`. However, traditionally in AGS, alpha values are given in percent. This must by needs be inexact from the get-go, and 255 isn't an exact multiple of 100, so some percentages must by needs be “more inexact” than others. 
- `[0 .. 100]` (100 means completely transparent, so nothing drawn is visible)
- `[0 .. 255]` (255 means completely transparent)
- `[0 .. 99]` (99 translates to 252 in the scale of [0 .. 255])
- `float [0.0 .. 1.0}`

We also need to discuss how this parameter should be handled in non-true-colour games, i.e., games where pixels can only be completely transparent or completely opaque.
- Ignore the value
- Don't draw anything whenever the value is non-zero
- Have some other cut-off, e.g., don't draw anything whenever the value is above 50 %.

ivan-mogilko:
I mentioned this elsewhere, so will restate here. In this feature finding a good script API could be _the least_ of the issue.

This feature will be blocked by two major problems:

1. Currently the draw colours in AGS are represented in 16-bit range (at least ones settable by script API).  Even though they are stored in 32-bit integer, they are remapped on assignment to match the range available for 16-bit graphics. Having `DrawingTransparency` as a separate field may potentially mitigate this issue (completely or to some extent), but otoh if you like `DrawingColor` to have alpha channel component instead, then solving it becomes a higher priority. Moving to full 32-bit color representation in all "color" fields is a must.

[...]

PR #1813, opened 2022-10-14
AGS 4: Remove "has alpha channel" sprite flag, and related parameters

### LINKED ISSUE #1808: AGS 4: Remove "has alpha channel" flag from sprites, and drawable objects
In AGS sprites have a "has alpha channel" flag which tells whether the image's alpha channel should be used or not. Certain drawing operations depend on it; for instance a blending op or conversion from the sprite to texture may be performed in one or another way depending on the combination of source and destination surfaces' alpha flags.

My proposal is to remove this flag and those checks altogether, and imply that 32-bit images always have a valid alpha channel. This should somewhat simplify all the drawing operations.

If necessary, the flag may remain in sprite import settings; in such case removal of an alpha channel from a 32-bit source image would correspond to filling alpha bytes with 0xff (fully opaque).

ivan-mogilko:
Resolves #1808.

In AGS sprites have a "has alpha channel" flag which tells whether the image's alpha channel should be used or not. Certain drawing operations depend on it; for instance a blending op or conversion from the sprite to texture may be performed in one or another way depending on the combination of source and destination surfaces' alpha flags.

The purpose of this change is to remove this flag and those checks altogether, and imply that 32-bit images always have a valid alpha channel. This will simplify a lot of things internally, and probably will make life easier to user who uses true color graphics by default. The only downside I see is that forcing an opaque image will become a responsibility of game author, either when preparing an image for import, or when generating a dynamic sprite in script (in which case one would have to clear the sprite with some color then blit the image on it to ensure there's no transparency).

**What is done**

1. Removed "has_alpha" and "opaque" parameters internally from textures (DDB) and drawing functions.
2. Removed SPF_ALPHACHANNEL flag from sprites. All 32-bit sprites now assume alpha channel.
3. In Script API removed "preserve alpha" parameter from DynamicSprite.Create and CreateFromExistingSprite functions; removed HasAlphaChannel from DialogOptionsRenderingInfo. **NOTE:** I don't know if we want to keep old-style functions in ags4 to ease the project migration. If yes, then I'd have to put old function variants back.

**What is kept**

Kept "Import alpha channel" in the Sprite Import dialog. For now it works simply: if set then do nothing, if unset then the sprite will be made fully opaque by filling alpha channel with 0xFF.
If not wanted, this setting may be removed later as well.

The "magic pink" color constant is still kept internally even in 32-bit. This is unfortunate, but may be necessary for DrawingSurfaces, because currently there's no way to set alpha for a DrawingColor, or in SetPixel/GetPixel, so people will have to use COLOR_TRANSPARENT (see #1514). Unless I'm missing or misunderstanding something here... I may look afterwards if it's okay to completely remove it and any related conversions for 32-bit.

---

TODO:
- [x] fix Editor compilation;
- [x] blank room template generates with RGB background, it seems, which turns into fully-transparent. Besides that, need to ensure that RGB backgrounds are converted into RGBA with opaque alpha?
- [x] any "has alpha" options in the script API.
- [x] double check that in the sprite import disabled "use alpha channel" should result in filling bitmap with opaque alpha;
- [x] ensure that all special opaque sprites in the engine have proper alpha channel values when created.
- [x] ensure that all the primitive drawing operations in the engine and script use color with opaque alpha.
~- [ ] perhaps merge all the `draw_xxx_sprite` functions, because they seem to do almost same thing now (may be done separately later).~
- [x] investigate what happens when upgrading older project with both "has alpha" and "no alpha" 32-bit sprites.
- [x] investigate what happens to room backgrounds when upgrading older project (from both open rooms and closed rooms 3.6.0).
- [x] ensure that 16-bit / 8-bit games display sprites properly (direct3d/opengl runs always in 32-bit display mode).

> ensure that all special opaque sprites in the engine have proper alpha channel values when created.

This includes:
- room background (currently it must be opaque) [WORKS - but only ags4 editor keeps opaque alpha]
- fadeout surface
- fx (tint) sprite
- default dialog options

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -216,12 +216,8 @@ enum GameGuiAlphaRenderingStyle
 
 
 // Sprite flags (serialized as 8-bit)
-#define SPF_HIRES           0x01  // sized for high native resolution (legacy option)
-#define SPF_HICOLOR         0x02  // is 16-bit
 #define SPF_DYNAMICALLOC    0x04  // created by runtime script
-#define SPF_TRUECOLOR       0x08  // is 32-bit
 #define SPF_ALPHACHANNEL    0x10  // has alpha-channel
-#define SPF_VAR_RESOLUTION  0x20  // variable resolution (refer to SPF_HIRES)
 #define SPF_HADALPHACHANNEL 0x80  // the saved sprite on disk has one
 
 // General information about sprite (properties, size)
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -217,8 +217,6 @@ enum GameGuiAlphaRenderingStyle
 
 // Sprite flags (serialized as 8-bit)
 #define SPF_DYNAMICALLOC    0x04  // created by runtime script
-#define SPF_ALPHACHANNEL    0x10  // has alpha-channel
-#define SPF_HADALPHACHANNEL 0x80  // the saved sprite on disk has one
 
 // General information about sprite (properties, size)
 struct SpriteInfo
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -500,6 +500,13 @@ HRoomFileError UpdateRoomData(RoomStruct *room, RoomFileVersion data_ver, const 
         }
     }
 
+    // Prior to ags4 room backgrounds were ignoring alpha channel
+    if (data_ver < kRoomVersion_399)
+    {
+        for (int i = 0; i < room->BgFrameCount; ++i)
+            BitmapHelper::MakeOpaque(room->BgFrames[i].Graphic.get());
+    }
+
     // sync bpalettes[0] with room.pal
     memcpy(room->BgFrames[0].Palette, room->Palette, sizeof(RGB) * 256);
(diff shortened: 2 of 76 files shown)


ivan-mogilko:
A note for memory: AGS 3.* editors *seem* to save 32-bit room backgrounds without alpha channel (so after this change these would be considered fully transparent). AGS 4 does not, and its backgrounds keep proper alpha there. I did not test what would happen if someone would make a half-translucent room background though, and not particularly certain if that would need any fix.

ivan-mogilko:
Note 2: in regards to drawing primitives (using Bitmap class and allegro functions), apparently the alpha channel is secure there and always set to 0xFF if color set by `__my_setcolor`, `MakeColor` or `Bitmap::GetCompatibleColor` (latter two call `__my_setcolor`).

ivan-mogilko:
This is ready for review and testing. Updated the first post with details on what was done.

ivan-mogilko:
Had to remake something & rebase: reverted an "opaque" flag in textures, because it's not related to the "has alpha" flag, and is used primarily for optimizing bitmap->texture conversions for images that are known to never have transparent pixels: room backgrounds, debug surfaces, built-in gui, fx effects.

This flag may be reviewed and refactored later, as a separate task.

PR #1814, opened 2022-10-15
Fix macos icon

ericoporto:
Minor fix for macOS port

- README includes CMake main usage link
- Xcode links the app statically to the engine library and doesn't embed the AGSKit framework, reducing 17 MB in `.app` size (an ags macOS app now starts at around 7MB)
- Xcode project correctly shows icon in the app when using the archive build scheme

### DIFF
--- a/OSX/README.md
+++ b/OSX/README.md
@@ -27,7 +27,11 @@ After you build your game with the Editor, your own game `.ags` and additional f
 In this port, all game resources are taken from `OSX/Resources`, remove the stub files that exist there and replace them with your game files taken from `Compiled/Data/`.
 
 
-## Building your game with CMake and Xcode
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
By the way, I was wondering recently if we should still do b88a0b571863a04ee115415c1f333e90e251d0fe (remove file refs from project.pbxproj), as it was not clear whether these lists are used for anything.

ericoporto:
> By the way, I was wondering recently if we should still do b88a0b5 (remove file refs from project.pbxproj), as it was not clear whether these lists are used for anything

Uhm, I simply removed the files from the project using Xcode itself, I am not sure why they appear in different places there, the file format is not clear for me, but it was building and running after removing them.

PR #1816, opened 2022-10-22
mismatched free in cc_dynamicarray

ericoporto:
A minor mismatched free picked in `cc_dynamicarray.cpp`, it uses new char[...] but frees it using a delete instead of delete[]. I am trying to learn how to use valgrind and picked this.

I am leaving it as draft because I want to run a simple game that allocates and deallocates a dynamic array to verify this is correct. There isn't many things in AGS code in valgrind, it generates a bunch of false positives when using my intel video card with the opengl driver.

### DIFF
--- a/Engine/ac/dynobj/cc_dynamicarray.cpp
+++ b/Engine/ac/dynobj/cc_dynamicarray.cpp
@@ -42,7 +42,7 @@ int CCDynamicArray::Dispose(const char *address, bool force) {
         }
     }
 
-    delete address;
+    delete[] address;
     return 1;
 }
 


ivan-mogilko:
This is a correct fix.

ericoporto:
I just wanted to make sure, but looks alright indeed!

ericoporto:
Valgrind also pointed at these as MismatchedFree, but I am not sure what is the relationship between these.

https://github.com/adventuregamestudio/ags/blob/869bfd248929fc76dca506d73f05bcd5e063fd86/Common/game/main_game_file.cpp#L711

https://github.com/adventuregamestudio/ags/blob/869bfd248929fc76dca506d73f05bcd5e063fd86/Common/ac/gamesetupstructbase.cpp#L71

ivan-mogilko:
> Valgrind also pointed at these as MismatchedFree, but I am not sure what is the relationship between these. My guess is it's a false positive because it allocates using a malloc.

This is not a false positive, that's an actual bug, because message[i] is allocated with malloc, but freed with delete[]. It must be either freed with `free`, or not use `strdup` to create a string.

ivan-mogilko:
> From following along, it's because as ags starts out, it will load a room, and when it loads, it runs the FadeOut (my_fade_out in global_screen), and then it will do a render BEFORE one render_graphics or something that calls construct_game_scene which sets the smoothScaling.

Hm, technically this may be fixed by assigning a value in constructor (which it should do anyway), but this also points out a logical mistake in the use of the gfx driver (maybe not overly critical, but still).

ivan-mogilko:
Fixed the `messages` problem with b47a086

ericoporto:
> Fixed the messages problem with b47a086

Awesome! Thanks for that!

About the other thing

> Hm, technically this may be fixed by assigning a value in constructor (which it should do anyway), but this also points out a logical mistake in the use of the gfx driver (maybe not overly critical, but still).

I think it's fine as is, if it was critical we would have noticed. I will later open an issue just to note this down.

PR #1821, opened 2022-10-27
Common: fix warnings in directory.cpp

ericoporto:
minor warnings and clang-tidy notices

### DIFF
--- a/Common/util/directory.cpp
+++ b/Common/util/directory.cpp
@@ -106,7 +106,7 @@ bool GetFilesImpl(const String &dir_path, std::vector<String> &files,
         return false;
     int dir_fd = dirfd(dir);
     struct dirent *ent;
-    struct stat f_stat;
+    struct stat f_stat{};
     while ((ent = readdir(dir)) != nullptr)
     {
         if (strcmp(ent->d_name, ".") == 0 ||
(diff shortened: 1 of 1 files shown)


Common/util/directory.cpp
```diff
@@ -106,15 +106,15 @@ bool GetFilesImpl(const String &dir_path, std::vector<String> &files,
...
+    struct stat f_stat{};
     while ((ent = readdir(dir)) != nullptr)
     {
         if (strcmp(ent->d_name, ".") == 0 ||
             strcmp(ent->d_name, "..") == 0) continue;
         if (fstatat(dir_fd, ent->d_name, &f_stat, 0) != 0) continue;
         if (S_ISREG(f_stat.st_mode) == is_reg &&
             S_ISDIR(f_stat.st_mode) == is_dir)
-            files.push_back(ent->d_name);
+            files.emplace_back(ent->d_name);
```
ivan-mogilko, line 117: There's another instance of `files.push_back` in this code (slightly below), does it also need to be replaced?

ericoporto: thanks! I was using Linux, now tried on Windows and saw the different code paths. I noticed SetCurrentDirectory, GetDirs and GetFiles are apparently not used.

ivan-mogilko: > I noticed SetCurrentDirectory, GetDirs and GetFiles are apparently not used.

GetDirs and GetFiles are likely used by tools.

ericoporto: Ahh, that's right! I forgot those! OK! :)

Issue #1819, opened 2022-10-26
Editor: Android config is not fully represented in game project, file is overwritten with defaults


ivan-mogilko:
**Problem**

When making an Android build from within the editor, the most of the `android.cfg` contents are always reset to defaults, without any means to change these values. Editor does not have them present anywhere in settings, and because it's overwritten each time when the game builds, user cannot even edit the file manually (all changes will be lost). This makes it practically impossible to edit all options in this file.

On an extra note, the new "mouse emulation" settings are not represented in Default Setup at all, and because acsetup.cfg is recreated from scratch each time, it's also not possible to edit it and add wanted setting.

NOTE: the latter, imo, is a sign of another potential issue: the user cannot have any custom settings in config. I think this was changed only in 3.6.0 (but I could be wrong).

**Expected behavior**

Each config option should have a corresponding property in the game project. If it's more convenient to take existing settings from Default Setup (also duplicated in acsetup.cfg), then this may be done. If not, then new properties should be added specifically for Android.

In regards to the config file being reset each time, I suppose we could also have a setting that tells whether to reset it or not, because otherwise users won't be able to add anything there if they like or need to (e.g. if the game wants to read custom options from the config). In theory this could also be solved by adding special field for adding config contents, but that's more work, and may be left for the future updates.

AlanDrake:
I think we made it reset some time ago because changes to default settings were not ending up being written in the cfg.
The real problem here are the missing defaults options, and maybe a comment added inside the cfg to warn that the file will be overwritten on compile.

ericoporto:
At least the touch options are relevant for the Web Build too, so I think I want everything that can be set in Android be possible to set in Default Setup too.

If representing everything from Android in Default Setup is an ok resolution of this, I can probably implement a PR for this soon - and this can be assigned to me.

ericoporto:
I think in the future we may want to have it be possible to configure either a global DefaultSetup, and then per platform either inherit whatever is the global or customize particular settings per platform. I am not sure how to do this UI wise in a way that is natural with the WindowsForms property grid we use.

ivan-mogilko:
> At least the touch options are relevant for the Web Build too, so I think I want everything that can be set in Android be possible to set in Default Setup too.

Could you clarify, which options are used in web build?

ivan-mogilko:
> I think we made it reset some time ago because changes to default settings were not ending up being written in the cfg.
[...]

PR #1824, opened 2022-10-28
Editor: Default Setup has new Android relevant options

ericoporto:
- Added mouse_emulation, mouse_control_mode and show_fps

partial of #1819

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1635,14 +1635,25 @@ namespace AGS.Editor
             NativeProxy.WritePrivateProfileString("sound", "driver", "", configFilePath); // always default
             NativeProxy.WritePrivateProfileString("sound", "usespeech", _game.DefaultSetup.UseVoicePack ? "1" : "0", configFilePath);
 
+            int control_enabled = (int)_game.DefaultSetup.ControlEnabled;
             NativeProxy.WritePrivateProfileString("language", "translation", _game.DefaultSetup.Translation, configFilePath);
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
The "control enabled" option won't work like that, because it has different meanings on mobile and desktop. It has to be separated on settings pane.

ivan-mogilko:
Hm, so, this also brought me to realization that mouse speed no longer works in 3.6.0... or maybe it was not restored properly.

ericoporto:
About the hints, it should be fine to apply both for speed independently of what's set - If I remember correctly, it's possible to set relative mode in-game at runtime.

ivan-mogilko:
So, my suggestion for this is:
* Rename "ControlEnabled" (enum and property) to something like "MotionMode/Method" which is either Direct or Relative, and place it into the "Touch" category. Write only into android.cfg for now (this may be adjusted later if more platforms require this).
* A less important proposal, but imho it's better to have EmulateMouseType have a name which references "Touch", as in theory there may be several different mouse emulations (with other input devices). Could be TouchToMouseEmulation, or similar.

I also noticed that the mouse speed is saved as integer multiplied by 10 in android.cfg. I don't know all the reasons for this, but apparently this is because engine must return `int` values when passing preferences? I think this is a mistake of data format following GUI and/or internal program limitation, and the setting should be stored as it is stored in acsetup.cfg, while the conversion rather be done inside the program itself.

ivan-mogilko:
Added your commit, with additional changes:
197e1b2b6dc614043be521f18747ebf71ee4a2b1
9f50b653d5984f7d52b31a1dc570abf6360a0919

PR #1831, opened 2022-11-23
Editor: restore proper panel dock fill behavior

AlanDrake:
Assigning to the Dock property has side effect. It should only be specified on the components themselves, never from frmMain.


ivan-mogilko:
This seem to have fixed the user's problem.

AlanDrake:
Alright, I'll remove this commit from the other PR.

Issue #1832, opened 2022-11-24
New compiler: code lines that contain pre-evaluated math are missed during the step-by-step execution


ivan-mogilko:
**Describe the bug**
The new compiler in ags4 branch can precalculate simple arithmetical operations. So, if you have a variable declaration which assignment includes simple math (e.g. `int a = 2 + 2;`), the resulting value will be evaluated and used when writing a bytecode, instead of the sequence of operators.
It seems that when such line is written to the bytecode, it omits some information, like a line start opcode. Because of that, when running the game under debugger, then breakpoint on this line will never trigger, and when doing step-by-step execution, this line will be missed. This is inconsistent with the simpler assignments, such as `int b = 4`;

**AGS Version**
AGS 4

**To Reproduce**
Use following script:
```
function game_start()
{
   int a = 1;
   int b = 1 + 1;
   int c = a + b;
   c += 1;
   c += 1 + 1;
   return 0;
}
```

In General Settings make sure that the "Use extended script compiler" is enabled.
Place a breakpoint at the first line (`int a = 1;`), then run the game under debugger. When the breakpoint is hit, press F11 to make a step. Notice that the debugger will only stop on following lines, jumping the ones between which contain pre-evaluated math:
```
   int a = 1;
   int b = 1 + 1; // <---- skipped
   int c = a + b;
   c += 1;
   c += 1 + 1; // <---- skipped
   return 0;
```

**Expected behavior**
Script lines should not be missed by the debugger when doing step-by-step execution.

ericoporto:
Uhm, do you know if for `int a = 1; int b = 1 + 1; int c = b + 1;` the compiler is optimizing for `int c = 3` ?

ivan-mogilko:
> Uhm, do you know if for `int a = 1; int b = 1 + 1; int c = b + 1;` the compiler is optimizing for `int c = 3` ?

It does not, like I mentioned, the debugger stops on first (`int a = 1`) and third (`int c = b + 1`) lines.

ivan-mogilko:
> Uhm, do you know if for `int a = 1; int b = 1 + 1; int c = b + 1;` the compiler is optimizing for `int c = 3` ?

It does not look like it does. For a better test I used this script:
```
  int a = 1;
  int b = 1 + 1;
  int c = a + b;
```

The debugger stops on the first and third line, skipping the second. If it was optimizing, it would only stop on the third.

fernewelten:
[...]

PR #1835, opened 2022-11-25
New compiler: Don't kill LINENUM pseudo commands when shortcutting code

fernewelten:
## Don't integrate yet
[Don't integrate yet: I've rerun all the googletests whilst enabling linenum directives and found a test that generated illegal code. Need to investigate this further.]

This addresses #1832.

The new compiler often generates code but then decides to use some shortcut instead. In these cases, it  replaces code that has already be generated. 

Whenever that replaced code started with a LINENUM pseudo-directive then that directive got overwritten. In consequence, the AGS system didn't recognize that the corresponding lines are visited. So the debugger skipped them; and when a runtime error happened on those lines, the runtime system wouldn't show the line of the error but the line in front of that. 

Changed the compiler so that it keeps leading LINENUM directives whenever it replaces code.

Typical AGS code:

```
function game_start()
{
   int a = 1;
   int b = 1 + 1; \\ ←
   int c = a + b;
   c += 1;
   c += 1 + 1; \\ ←
   return 0;
}
```

When tracing this code in the Editor, it should (also) stop on the lines marked `\\ ←`

### DIFF
--- a/Compiler/script2/cc_compiledscript.cpp
+++ b/Compiler/script2/cc_compiledscript.cpp
@@ -69,7 +69,6 @@ AGS::ccCompiledScript::ccCompiledScript(bool emit_line_numbers)
     sectionOffsets = NULL;
 
     LastEmittedLineno = INT_MAX;
-    EmitLineNumbers = emit_line_numbers;
     Functions = {};
     ImportIdx = {};
 }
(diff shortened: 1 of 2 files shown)


fernewelten:
I'm redoing this.

PR #1844, opened 2022-12-04
Engine: Fix malloc/delete mismatch with RoomStatus::tsdata

criezy:
There is a mismatch in the allocation and deallocation of `RoomStatus::tsdata`. In `roomstatus.cpp` `new []` and `delete []`, but in a couple of other files, the memory is allocated with `malloc`. This pull request changes all allocations of `RoomStatus::tsdata` to use `new []`.

Something that is strange, but that I didn't change, is the amount of memory allocated. In `roomstatus.cpp` it allocates `RoomStatus::tsdatasize` chars, which makes sense. But in the other places it allocates 10, 8, or 5 more bytes, depending where the allocation is done, and I see no reason to allocate those additional bytes. So that is also potentially something else that could be changed.

### DIFF
--- a/Engine/ac/room.cpp
+++ b/Engine/ac/room.cpp
@@ -245,7 +245,7 @@ void save_room_data_segment () {
     
     croom->tsdatasize = roominst->globaldatasize;
     if (croom->tsdatasize > 0) {
-        croom->tsdata=(char*)malloc(croom->tsdatasize+10);
+        croom->tsdata = new char[croom->tsdatasize+10];
         memcpy(croom->tsdata,&roominst->globaldata[0],croom->tsdatasize);
     }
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
>  But in the other places it allocates 10, 8, or 5 more bytes, depending where the allocation is done, and I see no reason to allocate those additional bytes. So that is also potentially something else that could be changed.

Yes, these extra bytes should be removed. This happened in few other places (and we were removing them when found), and our guess was that this is probably remains of a very old code, and these extra bytes were a naive attempt to prevent crashes in case of the occasional memory access error when iterating the array.

ivan-mogilko:
Actually, looking at this, I think that the best way is to replace tsdata with std::vector<char>.

criezy:
> Actually, looking at this, I think that the best way is to replace tsdata with `std::vector<char>`. This is long overdue, unfortunately.

Indeed, that makes sense. I will do that change.

PR #1845, opened 2022-12-04
Common: Fixed new[]/delete mistmatch

criezy:
By default `std::shared_ptr` uses `delete` to delete its data. When using it to store an array we need to pass a custom deleter that uses `delete []` instead. The easiest way to do this is to use the partial specialisation of `std::default_delete` for arrays.

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -301,7 +301,7 @@ void ReadDialogs(std::vector<DialogTopic> &dialog,
     for (int i = 0; i < dlg_count; ++i)
     {
         // NOTE: originally this was read into dialog[i].optionscripts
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I suspect it will also make sense to replace the `std::vector< std::shared_ptr<unsigned char> >` with `std::vector< std::vector<unsigned char> >` here. It uses shared_ptr probably because this was written before moving to a standard supporting unique_ptr. But in reality this unlikely needs shared ownership, and instead may be moved (with std::move) when reassigned.

ivan-mogilko:
@criezy I replaced these with std::vector too in the end: c1afae64fe348c64ee4a8c8888349ae267241395

PR #1848, opened 2022-12-11
Engine: avoid dirty rects outside software render surface

### LINKED ISSUE #1847: Software renderer renders bottom line when a GUI is out of screen
**Describe the bug**
When using software mode, if a GUI is Visible, but positioned with Y in excess of the screen height, the bottom of the screen will produce a black line. You can move the mouse over the line to view a small part of the line.

**AGS Version**
3.6.0.38

**Game**
Create a new game using the Sierra Template, and in the Project Explorer, select the gStatusline GUI, and in properties, set it's Y position to 201 or bigger. In default setup, chose the Software Renderer. Alternatively, download the project below.
[SoftwareRendererBug.zip](https://github.com/adventuregamestudio/ags/files/10200803/SoftwareRendererBug.zip)

**To Reproduce**
1. Move the mouse to the bottom of the screen, and notice a black line there
2. Where the mouse goes over, the black line is removed

**Expected behavior**
There should be no line at the bottom

**Screenshots**
Note the bottom line.
bottom_black_line

**Desktop:**
 - OS: Windows
 - Version 10

**Additional context**
Originally reported here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=60323.msg636651847#msg636651847

ericoporto:
Fix #1847

### DIFF
--- a/Engine/ac/draw_software.cpp
+++ b/Engine/ac/draw_software.cpp
@@ -266,17 +266,18 @@ void invalidate_rect_on_surf(int x1, int y1, int x2, int y2, DirtyRects &rects)
         return;
     }
 
+    if(x1 > x2 || y1 > y2) return;
+
     int a;
 
     const Size &surfsz = rects.SurfaceSize;
-    if (x1 >= surfsz.Width) x1 = surfsz.Width - 1;
-    if (y1 >= surfsz.Height) y1 = surfsz.Height - 1;
+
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Is swapping x1/x2 and y1/y2 necessary, does that fix certain case?

ericoporto:
> Is swapping x1/x2 and y1/y2 necessary, does that fix certain case?

So, now I actually looked into all calls and it's always passed as something like x1,y1,x1+w,y1+h, so it's not necessary, but I left the suggested return clause in case I missed something.

PR #1849, opened 2022-12-11
Minor performance improvements

ericoporto:
First, this is not meant to include all commits, they can be dropped or all dropped. These are all minor things I picked up using [AMDuProf](https://developer.amd.com/amd-uprof/#download) and [Very Sleepy](http://www.codersnotes.com/sleepy/) profilers, using a high resolution (1920x1080) game for testing.

Some commits make the code uglier or bigger - even thought they perform better, specially in high resolution bitmaps, mostly through switching for iteration and depth check order.

### DIFF
--- a/libsrc/allegro/src/c/cblit.h
+++ b/libsrc/allegro/src/c/cblit.h
@@ -50,7 +50,7 @@ void FUNC_LINEAR_CLEAR_TO_COLOR(BITMAP *dst, int color)
    bmp_select(dst);
 
    for (y = dst->ct; y < dst->cb; y++) {
-      PIXEL_PTR d = OFFSET_PIXEL_PTR(bmp_write_line(dst, y), dst->cl);
+      PIXEL_PTR d = OFFSET_PIXEL_PTR(dst->line[y], dst->cl);
 
       for (x = w - 1; x >= 0; INC_PIXEL_PTR(d), x--) {
 	 PUT_PIXEL(d, color);


ivan-mogilko:
I'd wish this was split into 3 separate PRs, one for each problem, for easier review and testing:

1. Allegro cleanup.
No comments here, if it does not break anything - it's good to be applied.

2. The renderer's stage screen clear.
This is a very good catch! But thinking about this further, I'm now in doubts whether engine should clear this stage screen at all. Perhaps it should be made a plugin's responsibility to clear if it needs to. Because now engine will still do certain action without knowing whether plugin needs it at all (plugin may be performing a 3D rendering without using raw stage surface).

3. An issue of GetScanLine and iterating lines of pixels.
This is definitely a mistake that GetScanLine was still in use; it's a remnant of times when Bitmap class was introduced trying to fully mimic Allegro's BITMAP behavior. Allegro's BITMAP could contain different kind of data, starting from simple linear bitmaps, and following with video textures, planar bitmaps, etc. But the engine was mostly using linear memory bitmaps (except for the final render to screen, afaik), and now when Allegro's source is stripped, there are no other bitmap types supported at all.

So it's correct to assume that Bitmap is simply a linear array of pixels with fixed stride calculated as w*h*bpp, do not use GetScanLine at all, and use GetData instead for getting the beginning of array once.

ericoporto:
OK, I am separating this in some PRs. :)
- #1851 : the allegro bank removal
- #1850 : the plugin rawdraw adjustment
- #1852 : iterating pixels (this is a draft for now)

Left here is the ugly stuff which may be less relevant.

ericoporto:
Additionally moved #1858 out of this to clarify what those two specific commit does.

ericoporto:
I am going to close this to favor a new allegro clean up PR

PR #1850, opened 2022-12-12
Engine: only clear _stageScreen after it's dirty

ericoporto:
> Perhaps it should be made a plugin's responsibility to clear if it needs to. Because now engine will still do certain action without knowing whether plugin needs it at all (plugin may be performing a 3D rendering without using raw stage surface).

For the last point, in this case the plugin can simply return false, right?

https://github.com/monkey0506/ags2client/blob/7db4df8baccf955b0ff9fb496d9779a2fb90bf25/main.cpp#L136

The above is used in agsteam plugin, so it currently affects all AGS games on Steam. 

---

I think this is safe to be applied, but perhaps the responsibility change has to be something for a next version change.

TO-DO: figure out if there is any plugin besides SnowRain that uses this call for **drawing** edit: ags_parallax and palrender in this repository also use the `AGSE_PREGUIDRAW` which I believe is for RawDraw too. I don't understand how Snow Rain plugin works in this case as in the code it's returning 0, but it does manage to trigger the clear?

https://github.com/adventuregamestudio/ags/blob/78aa08036b9f245aef1a078d7ed71cb322edc1bd/Plugins/ags_snowrain/ags_snowrain.cpp#L826

From what I understand the int x, int y is actually int event, int data from `pl_run_plugin_hooks`.

### DIFF
--- a/Engine/gfx/gfxdriverbase.cpp
+++ b/Engine/gfx/gfxdriverbase.cpp
@@ -290,7 +290,6 @@ bool VideoMemoryGraphicsDriver::DoNullSpriteCallback(int x, int y)
     if (!_nullSpriteCallback)
         throw Ali3DException("Unhandled attempt to draw null sprite");
     _stageScreenDirty = false;
-    _stageScreen.Bitmap->ClearTransparent();
     // NOTE: this is not clear whether return value of callback may be
     // relied on. Existing plugins do not seem to return anything but 0,
     // even if they handle this event.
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I did some tests with both old and new (ours) SnowRain plugins, and found two things:
1. Clearing this surface in the engine has to be done (unfortunately?), because neither plugin does that.
2. _stageScreenDirty is set not by return value of the event callback, but when the GetVirtualScreen is requested. This was the workaround I found, as they do not return a proper return value from event. An alternative is to react to MarkRegionDirty instead (and both plugins call this function), but I guess that's a separate issue.

ericoporto:
> In case of these plugins _stageScreenDirty is set not by return value of the event callback, but when the GetVirtualScreen is requested

So this is the magic? Pretty interesting! Now I understand, in the SnowRain, on the `Update()` function called in the event, the plugin it blits to the screen and this does set `_stageScreenDirty` as I understand it does GetVirtualScreen. I had verified this worked but at the time I could not tell why, but this explains. :)

PR #1853, opened 2022-12-15
Fix snowrain plugin

### LINKED ISSUE #1843: SnowRain plugin draws particles only on part of the screen
Our reimplementation of the SnowRain plugin seem to draw particles only on the part of the screen. This was mentioned on forums at least once in the past, but I did not pay attention back then.

It's not clear whether this is a general issue, or depends on the game's resolution or other factors (I've never looked into the plugin's code before). One example where this happens is ["I Can't Sleep in Silence"](https://www.adventuregamestudio.co.uk/site/games/game/2031/) game (you have to either use an engine with embedded plugins or replace plugin dll with our own). The problem is mentioned in comments to #1840.

ericoporto:
fix #1843 

image

### DIFF
--- a/Engine/ac/screen.cpp
+++ b/Engine/ac/screen.cpp
@@ -233,4 +233,15 @@ void RegisterScreenAPI()
     ccAddExternalStaticFunction("Screen::ScreenToRoomPoint^2", Sc_Screen_ScreenToRoomPoint2);
     ccAddExternalStaticFunction("Screen::ScreenToRoomPoint^3", Sc_Screen_ScreenToRoomPoint3);
     ccAddExternalStaticFunction("Screen::RoomToScreenPoint", Sc_Screen_RoomToScreenPoint);
+
+    /* ----------------------- Registering unsafe exports for plugins -----------------------*/
+
+    ccAddExternalFunctionForPlugin("Screen::get_Height", (void*)Screen_GetScreenHeight);
+    ccAddExternalFunctionForPlugin("Screen::get_Width", (void*)Screen_GetScreenWidth);
+    ccAddExternalFunctionForPlugin("Screen::get_AutoSizeViewportOnRoomLoad", (void*)Screen_GetAutoSizeViewport);
+    ccAddExternalFunctionForPlugin("Screen::set_AutoSizeViewportOnRoomLoad", (void*)Screen_SetAutoSizeViewport);
+    ccAddExternalFunctionForPlugin("Screen::get_Viewport", (void*)Screen_GetViewport);
+    ccAddExternalFunctionForPlugin("Screen::get_ViewportCount", (void*)Screen_GetViewportCount);
+    ccAddExternalFunctionForPlugin("Screen::geti_Viewports", (void*) Screen_GetAnyViewport);
+    ccAddExternalFunctionForPlugin("Screen::RoomToScreenPoint", (void*)Screen_RoomToScreenPoint);
 }
--- a/Plugins/ags_snowrain/ags_snowrain.cpp
+++ b/Plugins/ags_snowrain/ags_snowrain.cpp
@@ -776,7 +776,6 @@ void AGS_EngineStartup(IAGSEngine *lpEngine)
   engine->RegisterScriptFunction("srSetBaseline", (void*)&srSetBaseline);
 
   engine->RequestEventHook(AGSE_PREGUIDRAW);
-  engine->RequestEventHook(AGSE_PRESCREENDRAW);
   engine->RequestEventHook(AGSE_ENTERROOM);
   engine->RequestEventHook(AGSE_SAVEGAME);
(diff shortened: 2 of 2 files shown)


ivan-mogilko:
Could you please give some explanation? For instance, what was wrong with using GetScreenDimensions?

ericoporto:
The timing, it needs the screen size to be used at the initialization which happens at the end of AGS_EngineStartup, but GetScreenDimensions will return 0 for width and height at this point - I am not sure why. Using Screen from script though, works at that point, so I went that way.

ivan-mogilko:
Perhaps its best to fix GetScreenDimensions then?

ericoporto:
> Perhaps it's best to fix GetScreenDimensions then?

This is possible. GetScreenDimensions curiously, doesn't return the screen dimensions, it gets the main viewport width and height!

The Screen Width and Height (from script), matches game.GetGameRes().Width and game.GetGameRes().Height.

ivan-mogilko:
I would also like to find out how come the viewport data is not initialized before plugins, because that sounds weird... but maybe that's a separate problem.

ericoporto:
Made the change using Screen Width and Height in GetScreenDimensions. I know there are legacy games that had some weird resolution (I think they were 640x400 but with 320x200 coordinates), I wonder if this is why it used the viewport previously.

ivan-mogilko:
> I know there are legacy games that had some weird resolution (I think they were 640x400 but with 320x200 coordinates), I wonder if this is why it used the viewport previously.

In that case the game resolution was initialized accordingly. But what makes a difference is the legacy letterbox resolution.
https://github.com/adventuregamestudio/ags/blob/dc620b2cbd89d47bf5dd3e9a9e57558744e3fddc/Engine/ac/room.cpp#L380-L392

ericoporto:
> Because of that this "main" viewport may change after loading a room.

But is it initialized before? Maybe it's why the main viewport is not created before plugins?

This change is small enough that it's easy to revert if it breaks anything. 

An alternative is to keep the previous main viewport, but return screen in case main viewport size is 0.

ericoporto:
> The viewport parameter should be initialized immediately after game is loaded

The letterbox code above depends on room information, is it alright if it's initialized to game screen dimensions instead of zero?

> On another hand, theoretically plugin should update the screen size upon each room change

I don't want to make this change here as I would need more games to test this.

ivan-mogilko:
> The letterbox code above depends on room information, is it alright if it's initialized to game screen dimensions instead of zero?

Yes, this is what happens in `engine_init_resolution_settings`, but I don't know why does not that work. I haven't looked into this yet, but I guess that may be either a problem with order of initialization, or something else breaks it later on.

ericoporto:
ok, I am going to leave this as draft until I can
- verify `engine_init_resolution_settings`
- see if updating the plugin screen size on enter room works

ericoporto:
Made a change that forces the baseline to reset on each room enter. I have no idea if there is a game that ever called this function, but this should make it behave like you mentioned.

ivan-mogilko:
I believe this should work, and is more logical than getting screen dimensions at AGSE_PRESCREENDRAW once and then unsubscribing.

ivan-mogilko:
So, testing the original plugin more, I found that calling `srSetBaseline` upon entering a new room is wrong, because that overrides the baseline parameter set in user script. E.g. if script has this in room 1:
```
srSetBaseline(0, 50);
```
then changing into any other room must keep the baseline 50. But our plugin version breaks this now.

ivan-mogilko:
[test--351snowrainplugin.zip](https://github.com/adventuregamestudio/ags/files/10366986/test--351snowrainplugin.zip)

This test game is made with 3.5.1 (for other, unrelated reasons).
Specifics:
* Resolution is 640x400 with letterboxed mode, which means that the game is actually 640x480, but when rooms are smaller it will display them with black borders and report 640x400 screen size.
* Contains 2 scrolling rooms: first is 960x400, second is 960x480. Room 2 has blue borders to make it easier to notice that it's bigger.
* Rain is configured to start in room 1 *without setting baseline yet*, so it should cover whole room.
* Pressing space key will issue `srSetBaseline(0, 50);` command.

Observation of the original plugin:
* without explicit baseline, the "drops" are falling across the whole room.
* when changing rooms, the "drops" correctly occupy whole room space: either 400 or 480 height respectively.
* after pressing space, and issuing custom baseline, the "drops" will only fall within 50 pixels from the top, respectively. Switching rooms will keep the custom baseline setting.

ivan-mogilko:
Not willing to spend too much time on this now, but I think the quick fix is to remember if baseline was previously set by a user command or not. If it was not, then adjust baseline upon a new room entry, otherwise - don't.

ivan-mogilko:
Alright, I added this change: a01f58a826e56746b3fc5e503bebb3f3dcbcf884

PR #1855, opened 2022-12-16
Warning cleanup

ericoporto:
very minimal warnings that I picked up when analyzing the code


Common/util/bufferedstream.cpp
```diff
@@ -190,7 +190,6 @@ bool BufferedStream::Seek(soff_t offset, StreamSeek origin)
         case StreamSeek::kSeekCurrent:  want_pos = _position   + offset; break;
         case StreamSeek::kSeekBegin:    want_pos = _start      + offset; break;
         case StreamSeek::kSeekEnd:      want_pos = _end        + offset; break;
-        break;
```
ivan-mogilko, line 193: Regarding this, I think it's missing "default:" line actually. In `FileStream::Seek` that would return false.

ericoporto: adjusted! :)

PR #1859, opened 2022-12-18
Allegro cleanup bmp related methods

ericoporto:
- remove bmp_unwrite_line, this was related to the removed bank stuff, so we don't need it anymore
- minor adjustments to cblit so it's a bit more lean, but without logic change


ivan-mogilko:
Regarding "move pointers out" commit, is this *really* produces a notable effect? These are mere stack variables, and I thought compilers should optimize this kind of things anyway.

ericoporto:
Testing with godbolt it looks like it may have no effect, so I dropped that commit, but the rest is still valid.

PR #1860, opened 2022-12-19
Engine: draw fix, true means texture don't change

ericoporto:
https://github.com/adventuregamestudio/ags/blob/b327011f0cd07a2d02f7fcafaee967b024f8c2f7/Engine/ac/draw.cpp#L1620

the `construct_object_gfx` should return `true` if the texture doesn't change, and `false` if it doesn't, but the logic was inverted in the commit https://github.com/adventuregamestudio/ags/commit/cca997fe83ec5f7014677a69095a7c9ec1494cd9 , only for the hardware accelerated renderer.

For some reason this doesn't break any small test game I made, but it does break some animations in a bigger game.

Edit: ok, I think I understand why it doesn't break my test games, the delay has to be 1 frame.

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -1508,7 +1508,7 @@ bool construct_object_gfx(int aa, int *drawnWidth, int *drawnHeight, bool always
     if ((hardwareAccelerated) && (walkBehindMethod != DrawOverCharSprite))
     {
         // HW acceleration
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Hmm weird, I remember specifically testing this...

> Edit: ok, I think I understand why it doesn't break my test games, the delay has to be 1 frame.

Could you elaborate?

Issue #1838, opened 2022-11-28
Transparent GUI's controls are incorrectly blending with the parent GUI


AlanDrake:
**Describe the bug**
In hardware accelerated rendering, semi-transparent GUIs are being rendered incorrectly because controls are being blended with the underlying GUI graphics.

**AGS Version**
AGS 3.6 and AGS4 (D3D9 and OpenGL)

**To Reproduce**
1. Create a GUI with a black background and colored border
2. Create a control, like a text box with some colors and borders so it overlaps with other visible elements
3. Make the GUI 50% transparent or so
4. Controls are drawn on top of the semi-transparent GUI mixing with the underlying content

**Expected behavior**
![immagine](https://user-images.githubusercontent.com/1430815/204387156-a7cd2356-bc2f-4137-89b3-ab48dcb0d394.png)
(correct, software rendering)

**Screenshots**
![immagine](https://user-images.githubusercontent.com/1430815/204385696-0c00f8f6-74e4-4878-b4df-ef7828352f8e.png)
(wrong, D3D/OGL)
The controls shouldn't be mixing with what's underneath that is part fo the GUI.

ivan-mogilko:
Also reproduced in 3.6.0, so it's a 3.6.0 bug.

ivan-mogilko:
@AlanDrake 
Looks like I made a successful experiment:
https://github.com/ivan-mogilko/ags-refactoring/tree/360--guitexturerenderfix2

Basically, this lets to pass a texture when you call BeginSpriteBatch, and the whole batch and all the nested sprites will be rendered on that texture target instead.
The nice thing is that then this texture may be added just like a regular sprite, with any transforms and effects.
Unfortunate thing is that the code organization is clumsy.

Made it work with Direct3D for now, next will do for OpenGL, and try to tidy the code.

AlanDrake:
I tried it, but I see the GUI transparency is not being applied somehow, though it's being set to the new texture.

ivan-mogilko:
> The 3.6 version of the Test Wretcher project, if you still have it.

Apparently the 3.5 (or 3.6?) version of the Wretcher had "Classic" gui render style set, which makes GUI draw in software mode. I haven't tried the ags4 version yet.
Pushed a fix that broke the software mode now.

AlanDrake:
Ah, that explains why I had never noticed the blending working differently on 3.x... I just wasn't using proper alpha blending.
[...]

PR #1862, opened 2022-12-19
Make sprite batch parameters more consistent

### LINKED ISSUE #1708: Make sprite batch parameters more consistent
This is rather a "note to self" atm.

After #967 sprite batches act like a node tree, but [their settings](https://github.com/adventuregamestudio/ags/blob/dc8ca8631c31dcaca3778c804fad120e457ea071/Engine/gfx/gfxdriverbase.h#L38) are not done well. For certain reason they initially used a "viewport" parameter both as a way to make offset on screen as well as clip a sprite group by a bounding rectangle. This is not convenient anymore, especially in ags4 as it fully supports object rotation.

There are also complications with room camera-viewport relations, that is - when inner objects may have offset/transform relative to camera's "eye" and then also an offset/transform relative to the screen coordinates. Right now this is done as one "node" (batch), which complicates the settings, as it has to have both "inner" and "outer" transforms.

Need to recode the sprite batch to have strictly one parent-child transform, that is easy to set and clear. The camera-viewport should then apparently done not as 1 but as 2 nodes: parent determining room viewport's location on screen and its child sub-batch determining room's location relative to camera.

EDIT:
Another problem is a global Offset + Flip, which is also assigned to each sprite batch. These settings should be also defined in a parent node, while child nodes should use the resulting parent transform. Global offset paremeter should likely be deprecated as redundant, and its values merged with the offset (coordinate translation) in SpriteTransform struct.

ivan-mogilko:
Resolves #1708, but also useful while resolving #1838 (which was the main reason to do this now).

1. Use glm in Direct3D renderer as well, to ease sharing the matrix code between OpenGL and Direct3D.
2. Global offset and flip are only set once in the parent sprite batch, instead of applying to every batch on screen.
3. For hw renderers - separate Room Viewport and Room Camera into two nested batches.
4. No longer use sprite batch's `Viewport` field as a source of coordinate translation (also removed extra `Offset` field); use strictly `Transform` field for that. This makes setting up a sprite batch more clear and consistent. `Viewport` will only define a clipping rectangle. Nested batch's `Viewport` will be considered relative to the parent batch, and will be transformed using parent batch's transform too.

Also, this was not necessary, but moved mouse cursor update into a separate function `update_mouse_cursor`, out from `construct_game_screen_overlay`.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -179,6 +179,7 @@ target_sources(common
     util/bufferedstream.h
     util/string_compat.c
     util/string_compat.h
+    util/matrix.h
 
 	platform/windows/windows.h
 )
@@ -187,7 +188,8 @@ target_link_libraries(common PUBLIC
         ${SDL2_LIBRARY} ${SDL2_LIBRARIES}
         Allegro::Allegro
         AlFont::AlFont
-        AAStr::AAStr)
+        AAStr::AAStr
+        glm::glm)
 
 if (WIN32)
     target_link_libraries(common PUBLIC shlwapi)
--- a/Common/util/matrix.h
+++ b/Common/util/matrix.h
@@ -1,0 +1,106 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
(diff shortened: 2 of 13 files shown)


ericoporto:
Hey, I tested with a bunch of games and it worked, I tested D3D9, OGL, Software Renderer and OGL but on the Web Browser. I couldn't catch any differences. This looks good!

PR #1866, opened 2022-12-22
Debugger related commits from Editor Log PR

ericoporto:
**Note:** since the engine part of that PR that actually adds the logging functionality requires a `--log-debugger=` flag, I think we could actually add the Engine [commit](https://github.com/adventuregamestudio/ags/pull/1518/commits/b58b186b05457787e45aafd415b264b1675b0a96) that adds it.

I am not 100% sure, but I think it may be possible to figure out an Editor plugin if the engine parts are taken care. It would require some different set of changes in the Editor to make this possible though.

---

Just picking up some of the debugger related commits from the Editor Log PR. I added a new one to tidy things a bit.

### DIFF
--- a/Engine/debug/debug.cpp
+++ b/Engine/debug/debug.cpp
@@ -89,6 +89,25 @@ int first_debug_line = 0, last_debug_line = 0, display_console = 0;
 float fps = std::numeric_limits<float>::quiet_NaN();
 FPSDisplayMode display_fps = kFPS_Hide;
 
+void send_message_to_debugger(const std::vector<std::pair<String, String>>& tag_values, const String& command)
+{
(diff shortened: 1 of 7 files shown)


Engine/debug/debug.cpp
```diff
@@ -638,10 +636,10 @@ void scriptDebugHook (ccInstance *ccinst, int linenum) {
 
     const char *scriptName = ccinst->runningInst->instanceof->GetSectionName(ccinst->pc);
 
-    for (size_t i = 0; i < breakpoints.size(); ++i)
+    for (auto & breakpoint : breakpoints)
```
ivan-mogilko, line 639: This also should be `const auto &breakpoint`.

ericoporto: added! :)

PR #1867, opened 2022-12-22
CI: Prefer RPATH over RUNPATH for Linux binaries

### LINKED ISSUE #1864: 3.6.0 linux "bundled" engine fails to load AGSSteam plugin (rpath issue)
When we distribute a linux engine along with the Editor, this engine is built not simply using a makefile, but using [debian scripts](https://github.com/adventuregamestudio/ags/tree/master/debian). What is also special about these scripts, is that they configure "rpath" for the engine, hardcoding "lib32" or "lib64" subdirectories. This lets an engine built like that "magically" know to look for any dependencies in `./lib32` or `/lib64`.
https://github.com/adventuregamestudio/ags/blob/24fdb20eadf56f9a6d9a432d15491f8741540623/debian/make_ags%2Blibraries.sh#L77

Such engine binary was intended to be used with the file structure deployed for the linux game ports by the Editor, where all libraries, including necessary game plugins, are placed into lib32/64 dirs relative to the engine binary.

This seem to be broken in the latest 3.6.0, where it cannot find a library in lib32/64 dir.

This can be noticed if a game is using plugins that also have extra custom dependencies, not installed in the system paths. (Like, another user-made library, or simply a so file that is distributed as-is) One good example is AGSSteam plugin (`libagsteam.so`) which also requires `libsteam-api.so` (part of the Steam SDK, afaik). Engine can find a plugin `libagsteam.so` itself if it's placed along with the engine, or the link to the plugin is created in the engine's dir (generated launch script does that, supposedly), because that is how it is coded. But agssteam's dependency is being searched by `dlopen` automatically, and it fails to find it in turn.

I previously spoke with @ericoporto about this, and he proposed to also set `LD_LIBRARY_PATH` in the game's [launch script](https://github.com/adventuregamestudio/ags/blob/master/debian/ags%2Blibraries/startgame). (NOTE: this script is also generated by the Editor [here](https://github.com/adventuregamestudio/ags/blob/24fdb20eadf56f9a6d9a432d15491f8741540623/Editor/AGS.Editor/BuildTargets/BuildTargetLinux.cs#L97))
I confirmed that this method fixes the problem.

But @morganwillcock claims that setting `LD_LIBRARY_PATH` is not recommended, and suggests to fix the problem with the bundle's build instead, where `rpath` is set, and is supposed to work. So I opened this ticket instead, for discussion and search of a reason why it stopped working in 3.6.0.

**NOTE:** I confirmed that pre-3.6.0 "bundle" engine builds work well (3.5.1, 3.5.0).

**TEST GAME:** 
[test--350steamplugin.zip](https://github.com/adventuregamestudio/ags/files/10282525/test--350steamplugin.zip)

**PS.** There's still an issue of a "common" engine (built with makefile or CMake) not being able to do the same. I shall open a separate ticket for that. EDIT: #1865

morganwillcock:
This restores the previous behaviour for library search paths.

I think the problem is the change in the build environment rather than any change in the engine source code, but build artifacts from this PR need to be tested against everything that is known to be failing to be sure.

When testing, be sure to use the binaries that are packaged to go with the Editor rather than the regular make or cmake builds.

P.S. This just restores the previous behaviour. Whether the previous behaviour and directory layout for game files was a good idea is probably another discussion.

Fixes #1864

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -35,8 +35,8 @@ build_linux_cmake_task:
     dockerfile: ci/linux/Dockerfile
     docker_arguments:
       matrix:
-        - FROM_DEBIAN: debian:jessie
-        - FROM_DEBIAN: i386/debian:jessie
+        - FROM_DEBIAN: debian:buster
+        - FROM_DEBIAN: i386/debian:buster
   env:
     matrix:
       - BUILD_TYPE: release
@@ -68,8 +68,8 @@ build_linux_debian_task:
     dockerfile: ci/linux/Dockerfile
     docker_arguments:
       matrix:
-        - FROM_DEBIAN: debian:jessie
-        - FROM_DEBIAN: i386/debian:jessie
+        - FROM_DEBIAN: debian:buster
+        - FROM_DEBIAN: i386/debian:buster
   env:
     matrix:
       - RPATH_PREFIX: lib
@@ -119,8 +119,8 @@ build_linux_make_task:
     dockerfile: ci/linux/Dockerfile
     docker_arguments:
       matrix:
-        - FROM_DEBIAN: debian:jessie
-        - FROM_DEBIAN: i386/debian:jessie
+        - FROM_DEBIAN: debian:buster
+        - FROM_DEBIAN: i386/debian:buster
   setup_destdir_script: |
     mkdir destdir
     arch=$(dpkg --print-architecture)
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -91,7 +91,7 @@ build_linux_debian_task:
           exit 1
           ;;
       esac
-      DEB_BUILD_OPTIONS="rpath=$RPATH_PREFIX$bit" fakeroot debian/rules binary
+      DEB_BUILD_OPTIONS="rpath=$RPATH_PREFIX$bit" DEB_LDFLAGS_MAINT_APPEND=-Wl,--disable-new-dtags fakeroot debian/rules binary
       sed -E "/^BI(NDMOUNT|T)=/d" debian/ags+libraries/hooks/B00_copy_libs.sh | BIT=$bit BINDMOUNT=$(pwd) sh
       ar -p ../ags_${version}_$arch.deb data.tar.xz | unxz | tar -f - -xvC data --transform "s/.*ags/ags$bit/" ./usr/bin/ags
       cd data && \
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -35,8 +35,9 @@ build_linux_cmake_task:
     dockerfile: ci/linux/Dockerfile
     docker_arguments:
       matrix:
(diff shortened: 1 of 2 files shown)


morganwillcock:
Something else is turning the previous settings back on with `--enable-new-dtags`:
```
LDFLAGS = -Wl,-z,relro -Wl,--disable-new-dtags -rdynamic -Wl,--as-needed -Wl,-rpath,$ORIGIN/lib64 
LIBS = -logg -ltheora -logg -lvorbis -lvorbisfile -L/usr/local/lib -Wl,-rpath,/usr/local/lib -Wl,--enable-new-dtags -lSDL2 -lSDL2_sound -ldl -lpthread -lm 
```
I don't get this when testing locally. I'll try to track it down.

morganwillcock:
It should be ready to test now:

```
$ readelf -d ags64 | grep PATH
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/lib64]
```

Unfortunately there was a second problem that the cached Docker images were not able to be rebuilt. There are separate EOL images for the same Debian version but they are delivered as multi-architecture images and the CI system doesn't seem to support the syntax needed to use them directly. To get it building again I've switched to the oldest Debian version which is listed with a separate i386 image but this is 2 versions newer and raises the minimum GLIBC version:
```
$ objdump -T ags64 | grep --only-matching "GLIBC_[^ ]*" | sort -V | uniq
GLIBC_2.2.5
GLIBC_2.3
GLIBC_2.3.2
GLIBC_2.4
GLIBC_2.7
GLIBC_2.14
GLIBC_2.28
```
2.28 might be too high to work with Steam runtime.

I guess that the current builds can be used to test the library issue, but the build environment might have to change again if the GLIBC requirement is a problem.

morganwillcock:
It should be downgraded to the original OS version now. The binaries look OK but I haven't tried to use them. The highest GLIBC requirement matches previous builds which are being distributed (I checked against the downloads for `3.5.1.23-P16`).

```sh
for a in 32 64
do
    echo $a bit
    readelf -d data/ags$a | grep PATH
    objdump -T data/ags$a | grep --only-matching "GLIBC_[^ ]*" | sort -V | tail -n1
done
```
```
32 bit
 0x0000000f (RPATH)                      Library rpath: [$ORIGIN/lib32]
GLIBC_2.7
64 bit
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/lib64]
GLIBC_2.14
```

ericoporto:
I tested this on Linux (ubuntu 16.04!!) and it's working, I can use agsteam with a build from this PR! Any reason why it's stil a draft?

morganwillcock:
> Any reason why it's stil a draft?

I haven't had a chance to test it yet, but given the Steam plug-in was the more problematic one it is likely OK to merge.

ivan-mogilko:
Works on my side too.

PR #1870, opened 2022-12-25
Allow Windows 64-bit Engine Builds using Visual Studio Solutions

ericoporto:
**context:** when testing a high definition, demanding, game, it may be the case that the memory usage may grow enough to cause OoM errors in some specific hard to reproduce scenarios, with 32-bit Windows app (limited to <2GB RAM). It may be wise to explore building Windows engine for 64-bit.

**what this pr does:** allows using VS to build Windows engine to 64-bit - currently other desktop ports and mobile ports of the engine are already 64-bit capable.

image

---

**Warning:** _this changes impact some things in the VS Workflow, please read below_

This change builds a new [**WinDevDependenciesVS.zip**](https://github.com/adventuregamestudio/ags/files/11069235/WinDevDependenciesVS.zip), which include pre-builts of 64-bit of SDL_sound, Xiph libraries and SDL2!

The paths for the new ones in your environment variables may change, specially SDL and SDL_SOUND which will point to a `x86` directory. This allows us to go back in git commit history without changing these environment variables, the prop files use `\..\x64` like paths.

image

When _using_ Visual Studio, builds for the Engine will work the same. The intermediary files though, they are built in Platform specific dirs. The final output files will be built in the same dirs as previously though, this means that building different platforms will overwrite the final binary - but because intermediary files are built separately, this happens instantly for already built sources.

The Editor isn't changed by this, and no 64-bit target is added to the editor.

No work is done in this PR to ensure 64-bit Windows builds actually behave correctly at runtime, this is intended to enable Windows developers using Visual Studio to explore and debug the engine for 64-bit more easily.

Changes are broken in multiple commits to ease readability.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,13 +6,13 @@ build_windows_task:
     matrix:
       - BUILD_CONFIG: Release
       - BUILD_CONFIG: Debug
-    AGS_LIBOGG_LIB: C:\Lib\Xiph
-    AGS_LIBTHEORA_LIB: C:\Lib\Xiph
-    AGS_LIBVORBIS_LIB: C:\Lib\Xiph
+    AGS_LIBOGG_LIB: C:\Lib\Xiph\x86
+    AGS_LIBTHEORA_LIB: C:\Lib\Xiph\x86
+    AGS_LIBVORBIS_LIB: C:\Lib\Xiph\x86
     AGS_SDL_INCLUDE: C:\Lib\SDL2\include
     AGS_SDL_SOUND_INCLUDE: C:\Lib\SDL_sound\src
     AGS_SDL_LIB: C:\Lib\SDL2\lib\x86
-    AGS_SDL_SOUND_LIB: C:\Lib\SDL_sound\build\Release
+    AGS_SDL_SOUND_LIB: C:\Lib\SDL_sound\lib\x86
   build_script: >
     "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
     cd Solutions &&
@@ -344,15 +344,11 @@ windows_packaging_task:
   archive_artifacts:
     path: AGS-*.zip
   make_windevdependencies_script: >
-    pushd "C:\Lib\SDL_sound\build\Release"  &&
-    move SDL_sound.lib ..\..  &&
-    popd  &&
-    pushd "C:\Lib\SDL_sound" &&
-    del /Q /S build\*  &&
-    rd /Q /S build  &&
-    mkdir build  &&
-    mkdir build\Release  &&
-    move SDL_sound.lib build\Release &&
+    pushd "C:\Lib\SDL_sound"  &&
+    del /Q /S build_x86\*  &&
+    rd /Q /S build_x86  &&
+    del /Q /S build_x64\*  &&
+    rd /Q /S build_x64  &&
     popd  &&
     tar -f WinDevDependenciesVS.zip -acv --strip-components 2 C:\Lib  
   windevdependenciesvs_artifacts:
--- a/Editor/AGS.Native/NativeDLL.vcxproj
+++ b/Editor/AGS.Native/NativeDLL.vcxproj
@@ -44,21 +44,24 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
     <_ProjectFileVersion>14.0.25431.1</_ProjectFileVersion>
+    <OutDir>$(Configuration)\</OutDir>
+    <IntDir>$(Configuration)\$(Platform)\</IntDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <OutDir>$(Configuration)\</OutDir>
-    <IntDir>$(Configuration)\</IntDir>
     <IgnoreImportLibrary>true</IgnoreImportLibrary>
     <LinkIncremental>true</LinkIncremental>
     <LinkKeyFile>AGS.Native.snk</LinkKeyFile>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
-    <OutDir>$(Configuration)\</OutDir>
-    <IntDir>$(Configuration)\</IntDir>
     <IgnoreImportLibrary>true</IgnoreImportLibrary>
     <LinkIncremental>false</LinkIncremental>
     <LinkKeyFile>AGS.Native.snk</LinkKeyFile>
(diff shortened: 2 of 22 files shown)


ivan-mogilko:
> When using Visual Studio, builds for the Engine will go to a Configuratio/Platform directories, so things that you would find before now on a Release directory will instead be in a Release/Win32 directory.

> The Editor isn't changed by this, and no 64-bit target is added to the editor. NOTE: You will need to copy acwin.exe from either Win32/acwin.exe or x64/acwin.exe manually and place in the Editor directory when building locally, for the time bein_.

This is going to make working and testing the Editor along with the Engine quite inconvenient.

Since x64 Windows build is currently experimental, and not added to the AGS Editor installation either, I propose to keep the final output location for 32-bit engine where it is (Release dir), and only add another dir for x64 builds (Release/Win64 and whatnot).

An alternate approach would be to give x64 build different name (e.g. `acwin64.exe`) and put both final outputs into the same dir. But this may be complicated by SDL2.dll, which also would need to have two variants. In theory, we may link to a library of a different name too (e.g. `SDL2x64.dll`); I leave this as something to consider.

ericoporto:
About having both 32bit and 64bit engines available in the editor, they would be in a Windows/x86 and Windows/x64 directories (in the future), it's not only the additional SDL2 dll, but also user plugins.

About the destination dirs, I think I will try adding 64-bit in the editor too, in this way the engine and editor would both be in a Win32 or x64 directory.

ivan-mogilko:
> About the destination dirs, I think I will try adding 64-bit in the editor too, in this way the engine and editor would both be in a Win32 or x64 directory.

Are you saying you will add to this PR, or you want to make a separate PR for that?

ericoporto:
I will experiment separately first and if it works I will add here. :)

ivan-mogilko:
> I will experiment separately first and if it works I will add here. :)

Perhaps make the Editor as a separate PR?

ericoporto:
> This is going to make working on and testing the Editor along with the Engine quite inconvenient.

Just curious, how do you currently deal with the empty acsprset.spr that is along with the Editor? I always need to manually move it out or my ags games don't start because their acsprset.spr is overriden by that.

ivan-mogilko:
> Just curious, how do you currently deal with the empty acsprset.spr that is along with the Editor? I always need to manually move it out or my ags games don't start because their acsprset.spr is overriden by that.

To clarify, you are asking about running the engine, not the editor?
I always set both path to game in a command line args, and working directory before running anything. The latter is necessary also in case of plugins, and few other things.

ericoporto:
Did a change, updated top post

- Environment `_LIB` variables point to specific `x86` library directories, and require VS Dependencies that have these directories (built in this PR). But once these are set, you can go back in git history and still build fine, as previous `_LIB` vars already expected to point to the `x86` dir, we just didn't name them specifically.
- The final output is now generated in the same dir as before this PR, the difference is intermediary files are generated in platform specific dirs, so the workflow doesn't change but the internal build used by VS is kept separated.
- This means that if you build for 32-bit and then build for 64-bit in the same config (Release/Debug) the final binaries are overwritten, but if you go back and build for 32-bit it should be fast, because the intermediary files are preserved in platform specific dirs (like `ags\Solutions\Engine.App\Debug\Win32`).

ivan-mogilko:
One thing I do not like is how the MSVS props are using partially hardcoded path for 64-bit libs, such as
`$(AGS_SDL_LIB)\..\x64`. This gives these env vars a "special" meaning, with an assumption that `AGS_SDL_LIB` must point exactly to something like "x86" dir, while "x64" is located inside the same parent dir nearby. This also forces user to create exactly such dir structure.
(For instance, I am building SDL2 from their repo, and the lib output has different structure by default:
- Win32 / Debug
- Win32 / Release <-- my `AGS_SDL_LIB` points here
- x64 / Debug
- x64 / Release
)

I noticed that in MSVS property manager the user macros are actually separate for x86 and x64 configurations: their lists do not match. This means that a macro of same name may be declared in both with *different values (paths)*. So likely this hardcoded addition of `\..\x64` is not even necessary.

If that's not the case for some reason, I would propose to have distinct macro names instead, like `AGS_SDL_LIB64` and similar.

ericoporto:
> If that cannot be used on some systems for some reason, I would propose to have distinct macro names instead, like AGS_SDL_LIB64 and similar.

I can add those but as an additional search path that is search first, if those are not set, then it searches using the hardcoded `\..\x64` path that just makes much more sense to me. It's one less thing to setup and works beautifully with our dependencies package.

ivan-mogilko:
Well, this works overall.

> This means that if you build for 32-bit and then build for 64-bit in the same config (Release/Debug) the final binaries are overwritten, but if you go back and build for 32-bit it should be fast, because the intermediary files are preserved in platform specific dirs

I was only able to do this reliably if I delete final binaries by hand; otherwise MSVS always picks up wrong engine and/or common lib.

On a separate note, it looks like 32-bit engine can work with 64-bit SDL.dll.

ericoporto:
> otherwise MSVS always picks up wrong engine and/or common lib.

This is weird, which VS version exactly you have? This didn't happen to me, it rebuilds all vcxproj files correctly. Is there any step in special you are making?

ivan-mogilko:
> > otherwise MSVS always picks up wrong engine and/or common lib.
> 
> This is weird, which VS version exactly you have? This didn't happen to me, it rebuilds all vcxproj files correctly. Is there any step in special you are making?

I use vs 2015.
I switch the platform on a toolbar and hit "start debugging" (f5), or optionally "build solution" (f6) then run.

ericoporto:
> What is the future plan for this, when the x64 editor will be complete, will they put final output to the separate dirs or still the same?

My guess is separate dirs, but the Editor requires figuring out some things
- irklang is picked from a fixed location, so either have a `References_x64` dir or rethink this (I would like to just get rid of irklang and redo using SDL_Sound).
- the "any CPU" c# projects have x86 specific references, so they either need to be made actually "any CPU" or have variants for x64 and x86
- use a different image-magick package so we can get both x86 and x64 packages - also there's a slight glue that needs to be made independent of being x86 or x64

I need help figuring specifics of the Editor but the Engine here the idea is to experiment (unfortunately Windows 64 has a tiny difference with the other 64 bit because of the long size difference) and have it be a possibility in the future in a memory demanding game when the game causes OoM problems. So I think the engine is useful even if the Editor isn't 64-bit yet. Still, I also believe it would be important to move the Editor to 64-bit too.

ivan-mogilko:
> unfortunately Windows 64 has a tiny difference with the other 64 bit because of the long size difference

Yes, the `long` should be removed completely from the code, and replaced in accordance to the intent: either as an int or intptr_t where it means to store a pointer.

Issue #988, opened 2019-11-11
Make GUI drawing compatible with hardware-accelerated sprite stack


ghost:
GUI rendering is currently a slow part in AGS, because it is all done using software operations (controls drawing) and only final result (whole GUI's surface) is converted into texture. Thus there are textures per each GUI, but controls are drawn outside of the renderer (see `guibg` and `guibgbmp` arrays).
Naturally, a slightlest change to GUI or any control on that GUI will result in whole GUI getting redrawn (actually, the situation is even worse, see note below).

Suggestion is to instead create and maintain a DDB (texture) per each GUI Control similarily to how it's done for characters and objects. These textures should then be passed into renderer according to gui/control z-order.

~**NOTE:** This may be viewed as an issue on its own, but currently there's a global flag `guis_need_update`, which is set whenever *anything* on *any GUI or control* changes. When this happens *all of the GUI* gets redrawn.~ This had been fixed.
**NOTE:** This ticket is essential before supporting any kind of GUI transformation (like scaling).

ivan-mogilko:
Forgot to make note, there's a draft pr opened, with couple of issues left unsolved at the time of writing, see more information here: #1393.

PR #1874, opened 2022-12-28
Render GUIs on a texture, then render that texture on screen

### LINKED ISSUE #1837: AGS 4: rotated GUIs not clipping their controls properly
**Problem**

Since the gui controls are now rendered as individual textures (see #988, #1393 for details), they were clipped in GUI borders using Clip/Scissor method in 3D renderers. However this does not work when GUI is rotated, as the clip/scissor rect must be orthogonal on screen. Currently this results in rotated GUI being cut itself (partially invisible). If the clipping is disabled, the GUI will stay fully visible, but the controls might display outside of the GUI's borders (if they are positioned so).

**Potential solutions**
From what I've read, there are following existing solutions:
1. Render onto a texture buffer, then render the resulting texture on screen. This is similar to how we render whole game to the texture buffer in "native resolution" mode (this does **not** mean using software drawing on bitmap). This seem to be a simpliest path, and it keeps the fast 3D rendering. The downside to this is that for the time being we lock GUIs in native resolution; which may conflict to the plans of supporting rendering UI, and especially - texts, in screen resolution (hi-res texts and ui in a low-res game). Still there might be some workarounds for that.
2. Shaders, may be used to make certain pixels transparent.
3. Stencils: in this case, may be used to create a sort of a texture mask of any shape and position. I did a small test before, and this method seem to work well.

Regardless of the chosen solution, there's likely a prerequisite to this: #1708.

EDIT: assuming #1838 is resolved using rendering onto a texture, looks like that would also automatically resolve current ticket.

### LINKED ISSUE #1838: Transparent GUI's controls are incorrectly blending with the parent GUI

ivan-mogilko:
Resolves #1838 and will potentially help #1837 after merged/ported to ags4.

After implementing rendering GUI controls as separate textures (see #988 and #1393 for details), it turned out that the implementation has a serious mistake: any additional visual effect applied to a whole GUI, such as Transparency, would apply incorrectly, as controls would blend with GUI background using same effect, producing different visual result.

I guess there may be different approaches for resolving this problem. But we also assume that in the future GUI may be assigned more various properties, such as rotation (already exists in ags4), which makes it more difficult to clip controls to GUI borders on screen, or custom shaders, which would be expected to be applied to GUI & controls in whole, while each control might also have its own individual effect, making things more complicated.

Therefore the chosen solution is to render whole GUI on a texture and then render that texture on screen. This is kind of similar in logic to what is done in software rendering, except still done as a scene render by GPU, without bitmap operations.

Additionally, same technique may be used later for anything, if necessary (maybe for room viewports, if we decide to support translucent viewports sorted among GUIs, etc).

What this PR includes:
* Direct3D and OpenGL renderers support creating a DDB object (texture) with a purpose of being used as a render target.
* Both support rendering a Sprite Batch to a custom render target, instead of a backbuffer. This resets all parent transforms for that batch, starting clean on a new target, but all the nested batches will be also drawn on same render target, and have merged parent-child transforms.
* In "draw controls as textures" mode engine will now render GUIs each on its own texture buffer, then render resulting buffers as common sprites on screen. The GUI parent transparency will be applied to that final buffer (not the contents).
* For Direct3D we have to reset all these buffer textures when changing display mode, unfortunately. This is required for D3D to be able to reset its mode. They will be recreated on fly during the very next game redraw.

### DIFF
--- a/Common/util/matrix.h
+++ b/Common/util/matrix.h
@@ -70,6 +70,17 @@ namespace glmex
         return inv_transform2d(glmex::identity(), x, y, sx, sy, anglez, pivotx, pivoty);
     }
 
+    // Setup Direct3D-compatible orthographic projection
+    // CHECKME: glm seem to supply some ortho variants meant for Direct3D,
+    // but none of them worked right in the engine.
+    inline glm::mat4 ortho_d3d(float width, float height)
+    {
+        return glm::mat4((2.f / width), 0.f, 0.f, 0.f,
+                         0.f, (2.f / height), 0.f, 0.f,
+                         0.f, 0.f, 0.f, 0.f,
+                         0.f, 0.f, 0.f, 1.f);
+    }
+
 
     // Linearly transform a rectangle using the given matrix;
     // This is an optimized case where rotation is not expected.
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -167,6 +167,8 @@ std::vector<ObjTexture> actsps;
 std::vector<ObjTexture> walkbehindobj;
 // GUI surfaces
 std::vector<ObjTexture> guibg;
+// GUI render texture, for rendering all controls on same texture buffer
+std::vector<IDriverDependantBitmap*> gui_render_tex;
 // GUI control surfaces
 std::vector<ObjTexture> guiobjbg;
 // first control texture index of each GUI
@@ -565,6 +567,7 @@ void init_game_drawdata()
     actsps.resize(actsps_num);
 
     guibg.resize(game.numgui);
+    gui_render_tex.resize(game.numgui);
     size_t guio_num = 0;
     // Prepare GUI cache lists and build the quick reference for controls cache
     guiobjddbref.resize(game.numgui);
@@ -585,6 +588,7 @@ void dispose_game_drawdata()
     walkbehindobj.clear();
 
     guibg.clear();
+    gui_render_tex.clear();
     guiobjbg.clear();
     guiobjddbref.clear();
 }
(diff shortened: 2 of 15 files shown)


ivan-mogilko:
Update on this: currently there are at least 2 known errors in OpenGL renderer:
1. GUI contents are always drawn 1x1 pixels down-left.
2. GUI contents are missing (either not drawn, or draw offscreen) in "render sprites at screen resolution" mode (which is quite strange, since the gui contents are not drawn on screen directly anymore anyway).

ericoporto:
Hey, this works for me! Tested and both error 1 and 2 seem to be fixed indeed! :)

ivan-mogilko:
I need to tidy (squash) the commits again, and there's also one logical mistake about switching render targets that has to be fixed (it is not seen in the current setup, but may surface if there's more complex batch configuration).

> I think you fixed the caveat you mention in https://github.com/adventuregamestudio/ags/issues/1837 , on item 1 , about rendering at screen res.

No, I did nothing in regards to that.

ivan-mogilko:
I wanted to do some additional refactor, but met couple of difficulties. So, I'll merge this pr first and make another attempt on refactor bit later.

PR #1877, opened 2022-12-31
Fix logic mistakes

ericoporto:
Picked up some logical mistakes when reading the code and using a static analyzer

- Memory stream doesn't check position (soff_t) after offset being lower than 0 before assigning
- In the compiler frontend, I wrote an error for empty input filename, but it's never checked
- In the palrender, two checks for the stars y position were wrong, and a get method that is used, the function was written as void

### DIFF
--- a/Common/util/memorystream.cpp
+++ b/Common/util/memorystream.cpp
@@ -130,7 +130,7 @@ bool MemoryStream::Seek(soff_t offset, StreamSeek origin)
         return false;
     }
     _pos = static_cast<size_t>(std::max<soff_t>(0, pos));
-    _pos = static_cast<size_t>(std::min<soff_t>(_len, pos)); // clamp to EOS
+    _pos = std::min(_len, _pos); // clamp to EOS
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


Common/util/memorystream.cpp
```diff
@@ -130,7 +130,7 @@ bool MemoryStream::Seek(soff_t offset, StreamSeek origin)
         return false;
     }
     _pos = static_cast<size_t>(std::max<soff_t>(0, pos));
-    _pos = static_cast<size_t>(std::min<soff_t>(_len, pos)); // clamp to EOS
+    _pos = static_cast<size_t>(std::min<soff_t>(_len, _pos)); // clamp to EOS
```
ivan-mogilko, line 133: Since both `_len` and `_pos` are `size_t` here, specifying std::min as `soff_t` is no longer right, and casting to `size_t` is also not necessary.

ericoporto: removed the specifier and the casting!

https://github.com/adventuregamestudio/ags/compare/bac760d52b4ffddc7c410c73c04795821272d5cb..1c55df6a73bc1a1528c7f9b95304cb382ea11c3a

Issue #1836, opened 2022-11-25
New compiler: Check the `LINENUM` pseudo-directives, too


fernewelten:
**Describe the problem**
The new compiler features around 120 googletests that check exactly what bytes are generated for specific short example programs. But line number generation is switched off in these tests, so they __don't__ check what `LINENUM` pseudo-directives are generated. 

We've already encountered a situation where such directives were missing in real code, and the existing googletests hadn't flagged that (#1832).

**Suggested change**
Turn on line numbers for the bytecode googletests. Check whether all the `LINENUM` directives are generated in the tests we already have.

ivan-mogilko:
Resolved by #1886

PR #1886, opened 2023-01-07
New compiler: Don't omit LINENUM directive with CTC; improved Googletests

fernewelten:
This addresses #1832 and #1836.
This replaces #1835.

Bug 1: When using compile-time compilations, the compiler emitted some `Linenum` pseudo-directives but cut them out afterwards so that the generated code didn't know about the affected lines any longer. This meant that breakpoints couldn't be set on such lines. (`Linenum` directives link specific points of generated code to specific source lines.)

This bug had been able to creep in because the Bytecode tests of the compiler ran without emitting `Linenum` directives. (The Bytecode tests make the compiler emit actual code for small AGS snippets and check whether the code is as expected.) So in particular, the Bytecode tests did not check whether these directives were correctly set. In my opinion, I had made a bad design decision when I decided this.

Thus I enabled the directives and redid all the tests, checking manually whether the directives were now emitted correctly. 

Bug 2: It now turned out that in one Googletest, the compiler had emitted correct code when `Linenum` directives were turned __off__, but now that the directives had been turned __on__, the compiler suddenly emitted incorrect code. 

This  bug happened when the compiler was cutting out code that had already been generated and pasting it into a different place. This code could sometimes contain cells that needed to be patched later on, and the bug was that some specific cells weren't patched correctly when `Linenum` directives where __on__.

The compiler code that used that functionality and had the bug was messy. So I cleaned up: I isolated the cutting, pasting, and patching functionality into classes and rewrote the compiler to use these classes. This fixed the bug.

As part of these changes, `for` loops are now generated in a slightly more space efficient way.

Typical AGS code:  All kinds of `for` loops. And:
```
function game_start()
{
   int a = 1;
   int b = 1 + 1; \\ ←
   int c = a + b;
   c += 1;
   c += 1 + 1; \\ ←
   return 0;
}
```
When breakpoints are set on the lines marked `←` then execution should stop when reaching those lines.

### DIFF
--- a/Compiler/script2/cc_compiledscript.cpp
+++ b/Compiler/script2/cc_compiledscript.cpp
@@ -74,6 +74,29 @@ AGS::ccCompiledScript::ccCompiledScript(bool emit_line_numbers)
     ImportIdx = {};
 }
 
+void AGS::ccCompiledScript::ReplaceLabels()
+{
+    std::vector<CodeLoc> RemainingLabels;
+
+    for (size_t idx = 0; idx < Labels.size(); idx++)
+    {
+        CodeLoc const loc = Labels[idx];
+        if (loc >= codesize)
+            continue;
+        try
+        {
+            CodeCell value = Label2Value.at(code[loc]);
+            code[loc] = value;
+        }
+        catch (const std::exception)
+        {
+            // Can't resolve the label at 'loc', so hang on to this location
+            RemainingLabels.push_back(loc);
+        }
+    }
+    Labels.assign(RemainingLabels.begin(), RemainingLabels.end());
+}
+
 AGS::ccCompiledScript::~ccCompiledScript()
 {
     FreeExtra();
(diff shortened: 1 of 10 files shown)


fernewelten:
This PR fails some `build_macos` checks, but I don't think my code is at fault. The failure message is:
```
[Intel macOS instances are no longer supported!](https://cirrus-ci.org/blog/2022/11/08/sunsetting-intel-macos-instances/)
```

ericoporto:
@fernewelten the macOS errors is because the ags4 base you have is not the last one, if you rebase to the latest aga4 it will pickup the latest CI - although I haven't yet added the new compiler to macOS builds.

fernewelten:
> if you rebase to the latest ags4

Rebasing pulled in all kinds of code that didn't concern the compiler, and git couldn't integrate the changes automatically. So in the end I resorted to getting my local 'ags4' up-to-date, basing a fresh branch from there and cherry-picking my four changes into that branch. So that fixed the 'build_macos' errors. Let's hope I didn't cause any nuclear explosions.

PR #1891, opened 2023-01-09
Alfont: fix leak, release before freeing last reference

ericoporto:
In Alfont, `s_pointer` holds a copy of the string we are drawing, but in case it's completely clipped, we return. But in this case, we don't free `s_pointer`, which is incorrect. You can see that later on, incase the string is not completely clipped, the `s_pointer` is freed.

Because the templates use wfn fonts, made a small game that draws text with a truetype font that gets clipped. You can run it using the build from master and see the RAM usage grow in realtime! :)

[TestClip.zip](https://github.com/adventuregamestudio/ags/files/10369913/TestClip.zip)

### DIFF
--- a/Common/libsrc/alfont-2.0.9/alfont.c
+++ b/Common/libsrc/alfont-2.0.9/alfont.c
@@ -947,8 +947,11 @@ void alfont_textout_aa_ex(BITMAP *bmp, ALFONT_FONT *f, const char *s, int x, int
 
   /* is it under or over or too far to the right of the clipping rect then
      we can assume the string is clipped */
-  if ((y + f->face_h < bmp->ct) || (y > bmp->cb) || (x > bmp->cr))
-    return;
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
So, in practice, this happens if a string is drawn partially off bitmap?

ivan-mogilko:
So, in practice, this happens if a string is drawn partially off the DrawingSurface in AGS?

ericoporto:
> So, in practice, this happens if a string is drawn partially off the DrawingSurface in AGS?

If a string is drawn off the DrawingSurface, in these particular cases:

> /* is it under or over or too far to the right of the clipping rect then
     we can assume the string is clipped */
 > `if ((y + f->face_h < bmp->ct) || (y > bmp->cb) || (x > bmp->cr))
    return;`

It doesn't clip if the string is too much to the left, presumably because it doesn't know the string width yet.

PR #1903, opened 2023-01-24
4.0: Add description field to the Game Template selection page

### LINKED ISSUE #564: move template description to template selection
and not after selecting the template.

ivan-mogilko:
Resolves #564.

This adds a description field to the template selection page in the New Game wizard. The template descriptions are now preloaded, along with the icons. This lets user read about the template prior to creating a game.

The pr is made against ags4 branch for now. It's simple enough, thus if wanted, this may be backported to some 3.6.* update later.

Example of looks:
template-preview

Known issues:
1. There's something strange about how the linebreaks are managed by the text box. Double linebreaks seem to work fine (see the screenshot), but single linebreaks do not cause the line wrap for some reason... This might be a .NET bug, or something that I'm missing in text box setup.
2. I had to make the Wizard dialog to auto-resize based on the "minimal size" property of all the pages. This caused the picture on the "intro" page to not fill the full window height. Don't know what may be the best way to handle this. Is the wizard dialog supposed to stay same size all the time, or may it only increase in size when displaying particular pages? OTOH it may be adjusted later, as this is but a minor annoyance.

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -21,7 +21,6 @@ namespace AGS.Editor
         public const string FILE_MENU_ID = "fileToolStripMenuItem";
         public const string HELP_MENU_ID = "HelpMenu";
         private const string CONTROL_ID_SPLIT = "^!^";
-        private const string TEMPLATE_INTRO_FILE = "template.txt";
 		private const string ROOM_TEMPLATE_ID_FILE = "rtemplate.dat";
 		private const int ROOM_TEMPLATE_ID_FILE_SIGNATURE = 0x74673812;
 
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
Added a linebreak replacement.

> There's probably space to have something like a short and a long description of the templates - the current text is presented both before and after so a wording like "you are using the empty game template" sounds weird when presented before actually using it.

I don't know if that's necessary. Current wording was meant for descriptions appearing as popups right after you create and open a game project. But this is a matter of deciding how the description is presented and adjusting the text if necessary. Personally, I would remove popups too, maybe even as a part of this PR. If a user wants to read description again, it may be found in template.txt in the project folder.

ericoporto:
> I would propose to remove popups too

I really don't like popups so I agree with this.

PR #1904, opened 2023-01-26
Engine: font width of empty text should be zero

### LINKED ISSUE #1892: Text width of empty string with outline should be zero
**Describe the bug**
Text width of an empty string with automatic outline 1 value has changed in the specific test game attached from being 1 to now being 2, but it should be 0.

**AGS Version**
- AGS 3.6.0.7 has the previous **behavior (A)**
- Change happens in this commit https://github.com/adventuregamestudio/ags/commit/3a9b8f199e09b9919a201c1d2fa486a79761b339 
- AGS 3.6.0.8 and current 3.6.0.40 maintains the **behavior (B)**

**Game**
[TestWidth.zip](https://github.com/adventuregamestudio/ags/files/10377414/TestWidth.zip)

**To Reproduce**
1. Use GetTextWidth with a string that is empty `""` and use a font that has an outline set.
2. Alternatively, run the demo game above

**Expected behavior**
Text width for an empty string should be 0 regardless of outline size.

**Screenshots**
image
**behavior (A)** (AGS 3.6.0.7 and before, also current 3.5.1.23)

image
**behavior (B)** (AGS 3.6.0.8 and forward, including current 3.6.0.40)

**Desktop:**
 - OS: Windows
 - Version 10

**additional context**

Alternatively, we can also maintain previous behavior (compatibility) and fix this in a later version if it's too late to change this in 3.6.0.

ericoporto:
fix #1892

### DIFF
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -260,6 +260,9 @@ int get_text_width_outlined(const char *text, size_t font_number)
 {
     if (font_number >= fonts.size() || !fonts[font_number].Renderer)
         return 0;
+    if(text == nullptr || text[0] == 0) // we ignore outline width since the text is empty
+        return 0;
+
     int self_width = fonts[font_number].Renderer->GetTextWidth(text, font_number);
     int outline = fonts[font_number].Info.Outline;
     if (outline < 0 || static_cast<size_t>(outline) > fonts.size())


Engine/ac/global_gui.cpp
```diff
@@ -167,6 +167,12 @@ int GetTextWidth(const char *text, int fontnum) {
   if ((fontnum < 0) || (fontnum >= game.numfonts))
     quit("!GetTextWidth: invalid font number.");
 
+  if(loaded_game_file_version < kGameVersion_360 &&
+     String_IsNullOrEmpty(text) &&
+     get_font_outline(fontnum) == FONT_OUTLINE_AUTO)
+  {
+    return game_to_data_coord(1);
```
ivan-mogilko, line 174: I don't think this fallback is necessary here, this is such a niche case, and I have doubts it was relied upon in some game's design. So I'd propose to not do this unless someone finds a game which gets broken by this fix.

ericoporto: That's fine, I will remove it :)

ericoporto: done, removed!

Issue #1871, opened 2022-12-27
Android: AGS Player does not remember "Mouse-to-touch emulation" setting (in Global Preferences)


ivan-mogilko:
**Describe the bug**
Each time you restart AGS Player and run any game, it seems like "Mouse-to-touch emulation" setting from Global preferences is ignored, and default value is used instead. User has to go into Preferences, and explicitly apply the setting again. It's worth noting that Preferences table displays a correct value remembered from the last time.

**AGS Version**
3.6.0

**To Reproduce**
1. Run "AGS Player", open Global preferences by touching ("...") button at the top-right corner.
2. Set "Mouse-to-touch emulation" to anything but default value.
3. Run any game and confirm that the setting works correctly.
4. Quit the game and close "AGS Player".
5. Run "AGS Player" again, open Preferences, see that "Mouse-to-touch emulation" displays the correct remembered value, but do not touch anything.
6. Run any game and notice that the setting is instead reset to default ("One finger" mode).

**Expected behavior**
Game should be run using last set preferences values.

ivan-mogilko:
So, this also happens with the "Screen orientation" setting.

ericoporto:
Need to check how it works in the single app build, but in the AGS Player, the `android_app_directory` points inside the apk, and in the `android_base_directory` we get the top ags directory (like `"/storage/emulated/0/ags"`). In the `setup.game_file_name` we get the full path. 

It looks like this line always fails in the AGS Player
https://github.com/adventuregamestudio/ags/blob/b1611316632ae4193634d05d59444f273d0c4bb8/Engine/platform/android/acpland.cpp#L349

I think for it to work, we would need to either be in the `android_base_directory` or prefix the filename with it.

There needs to be a check on how does it work in the single game, because I think in that case it either reads the config in the first step or read it twice.

ivan-mogilko:
To clarify, I was talking about Global preferences in this report. I did not realize (or forgot) that there are also individual preferences for each game, because there's no visual indication that you may open them for a particular game (no preferences icon besides the game title).

ericoporto:
Just noting from my comment in #1908

> I am not 100% sure but I think the problem is this here: https://github.com/adventuregamestudio/ags/blob/e069c3f51c08c163df023157e99e14bbcefd317a/Engine/platform/android/acpland.cpp#L101
> 
> So it would be using different directories as base dir, need to think a bit on how to properly fix in a way that is good for the single game player too.
>
[...]

PR #1908, opened 2023-01-28
Android: AGS Player add per game button options

ericoporto:
add a button with three dots on each game entry in the ags player to invoke the context menu of the item.

image

following this [comment](https://github.com/adventuregamestudio/ags/issues/1871#issuecomment-1405353587)

> there's no visual indication that you may open them for a particular game (no button/icon besides the game title).

### DIFF
--- a/Android/agsplayer/app/build.gradle
+++ b/Android/agsplayer/app/build.gradle
@@ -46,14 +46,14 @@ dependencies {
     // implementation fileTree(include: ['*.aar'], dir: 'libs')
     // implementation files('libs/uk.co.adventuregamestudio.runtime.aar')
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation 'androidx.core:core-ktx:1.3.2'
-    implementation 'androidx.appcompat:appcompat:1.2.0'
-    implementation 'com.google.android.material:material:1.2.1'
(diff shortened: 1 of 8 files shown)


ericoporto:
@ivan-mogilko to reduce trouble testing this, I added here the fix for storage access in the AGS Player, for Android 10 and above

ivan-mogilko:
Okay, good news is that directory permissions work again now, so that is fixed; I will report the rest of test results bit later.

ivan-mogilko:
So, the custom game preferences seem to work fine, as far as I can tell.

Global preferences, on the other hand, work strangely. It looks like they work the first time after you opened them, and then run the game (even if you haven't changed anything). But if you launch AGS Player again, and run the game right away, then, it seems like all options are reset to defaults (software renderer, absolute cursor movement, and so on).

ericoporto:
I am not 100% sure but I think the problem is this here: https://github.com/adventuregamestudio/ags/blob/e069c3f51c08c163df023157e99e14bbcefd317a/Engine/platform/android/acpland.cpp#L101

So it would be using different directories as base dir, need to think a bit on how to properly fix in a way that is good for the single game player too.

But the changes here are not related to that, so if you want to merge it should be fine, if not then it's alright too, but I need to think a bit on how to properly fix this.

PR #1910, opened 2023-02-01
Fix android ags player global settings

### LINKED ISSUE #1871: Android: AGS Player does not remember "Mouse-to-touch emulation" setting (in Global Preferences)

ericoporto:
fix #1871

The fix is what I explained in the issue. I added some comments to clarify what is going on.

- Removed chdir when reading the config, which was confusing. 
- There was also a misleading chdir [here](https://github.com/adventuregamestudio/ags/compare/master...ericoporto:ags:fix-android-ags-player-global-settings#diff-52a453b9246fe11573220e59a7f04e2498bd6cbd37802cf98494faf83dd1d61bL357-L358), because the passed variable was always null it did nothing.

### DIFF
--- a/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/AgsSettingsFragment.java
+++ b/Android/library/runtime/src/main/java/uk/co/adventuregamestudio/runtime/AgsSettingsFragment.java
@@ -11,6 +11,9 @@ import androidx.preference.PreferenceManager;
 import java.io.File;
 
 public class AgsSettingsFragment extends PreferenceFragmentCompat {
+
+    String configDirectory;
+
     @Override
     public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
         setPreferencesFromResource(R.xml.preferences, rootKey);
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Yes, this seem to fix the problem.

PR #1913, opened 2023-02-08
Editor: Add option to reset tile settings on sprite export

morganwillcock:
This adds an option to reset the tile settings for a sprite during bulk export, which fixes the case where the exported sprite is later reloaded from the exported source file (i.e. what was imported from a spritesheet is not exported as a spritesheet and so cannot be reloaded as a spritesheet). I did consider defaulting this checked in the GUI but decided against it because the original spritesheet import data (tile offsets) would be deleted. It seems safer to default to keeping data rather than deleting it. It could also be [ab]used to bulk reset all tile settings independently of the souce path rewrite.

I also noticed that pressing the tab key wasn't moving between the GUI controls in an expected order so I've reset the order to run from to to bottom, and from left to right.

### DIFF
--- a/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
+++ b/Editor/AGS.Editor/GUI/SpriteExportDialog.Designer.cs
@@ -52,6 +52,7 @@
             this.btnExport = new System.Windows.Forms.Button();
             this.groupBoxSpriteProperties = new System.Windows.Forms.GroupBox();
             this.groupBoxExportIf = new System.Windows.Forms.GroupBox();
+            this.chkResetTileSettings = new System.Windows.Forms.CheckBox();
             this.groupBoxExportFrom.SuspendLayout();
             this.groupBoxSaveAs.SuspendLayout();
             this.contextMenuStripExport.SuspendLayout();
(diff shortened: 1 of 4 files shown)


ericoporto:
This is working as stated.

ivan-mogilko:
>  I did consider defaulting this checked in the GUI but decided against it because the original spritesheet import data (tile offsets) would be deleted. It seems safer to default to keeping data rather than deleting it. It could also be [ab]used to bulk reset all tile settings independently of the souce path rewrite.

Don't you think that this should not be an option at all, but something that is done always when "Set exported file as a sprite source" is set?
I may be missing some use-cases, but from the first glance this fixup seem to only have a meaning when switching a sprite source, and should always be done when switching a sprite source (for the reasons you mentioned).

morganwillcock:
> > I did consider defaulting this checked in the GUI but decided against it because the original spritesheet import data (tile offsets) would be deleted. It seems safer to default to keeping data rather than deleting it. It could also be [ab]used to bulk reset all tile settings independently of the souce path rewrite.
> 
> Don't you think that this should not be an option at all, but something that is done always when "Set exported file as a sprite source" is set? I may be missing some use-cases, but from the first glance this fixup seem to only have a meaning when switching the sprite source to a exported file, and this should always be done when switching a sprite source (for the reasons you mentioned).

It would remove the possiblity of using it just for the side effect. Two possibilties might be:
1. generating source references which are then manually replaced by spritesheets where the offsets make sense
2. resetting the tile settings without rewriting the source references (i.e. files that are exported can just be thrown away)

Both of those would be preferable to telling someone that they have to carefully edit the XML data to perform bulk changes. I was pretty convinced that I hadn't removed any viable workflows when switching the sprite importer, but I had. I'd prefer to leave the options open (although potentially both this and the option to update the sprite source reference can be checked by default, as using both together is likely the most common case).

PR #1914, opened 2023-02-08
Editor: Add sprite selector option "Replace sprite using previous files..."

### LINKED ISSUE #1911: Restore option to replace sprite using the most recent image file
**Describe the problem**

When the updated sprite importing was added it removed the option to replace a sprite using the most recently used image file.
Bildschirm-foto-2023-02-01-um-21-53-03
Now there is not an equivalent function that can be used to repeatedly replace the sprite images from a common spritesheet file.

There is not a workaround using image export because this produces individual images rather than restoring a spritesheet which can be edited as if it were the original spritesheet.

Discussed here in full here:
https://www.adventuregamestudio.co.uk/forums/index.php?msg=636653195

**Suggested change**

Restore the removed menu option, using the first (or last, it might depend on how the option is named) image from the previous set of images which were used.

The tracking of previously used files would probably need to be changed. I think at the moment it is just remembering the files used for a brand new import and doesn't consider files used to replace the image on an exsiting sprite.

morganwillcock:
This restores some functionaility that was lost by the removal of "Replace sprite using last sprite" and makes the file tracking more consistent.

Note: There appears to be a separate problem of no error handling when the "previous" files no longer exist, but this was a pre-existing condition which can be looked at separately.

Closes #1911

### DIFF
--- a/Editor/AGS.Editor/Panes/SpriteSelector.cs
+++ b/Editor/AGS.Editor/Panes/SpriteSelector.cs
@@ -510,6 +510,7 @@ namespace AGS.Editor
 
         private void ReplaceSprite(Sprite sprite, string[] filenames)
         {
+            _lastImportedFilenames = filenames;
             SpriteImportWindow impWin = new SpriteImportWindow(filenames, sprite);
 
             if (impWin.ShowDialog() == DialogResult.OK)
@@ -522,6 +523,7 @@ namespace AGS.Editor
 
         private void ReplaceSprite(Sprite sprite, string filename)
         {
+            _lastImportedFilenames = new string[] { filename };
             SpriteImportWindow impWin = new SpriteImportWindow(new string[] { filename }, sprite);
 
             if (impWin.ShowDialog() == DialogResult.OK)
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Does this action suppose to keep the import settings, and apply them to the Import dialog?

I noticed that some options are reapplied, but not Tiled Import checkbox.

morganwillcock:
> Does this action suppose to keep the import settings, and apply them in the Import dialog?
> 
> I noticed that some options are reapplied, but not Tiled Import checkbox.

If the existing "Replace sprite from file" option does the same thing, this should probably be looked at as a separate issue. I don't think there is any existing functionality to re-create the selection recentagle.

ivan-mogilko:
> If the existing "Replace sprite from file" option does the same thing, this should probably be looked at as a separate issue.

Yes, "Replace sprite from file" does the same thing. All the setting seem to be applied in the import dialog, *including tile coordinates*, but not the "tiled import" checkbox.

ivan-mogilko:
Well, I am not sure what the expected behavior should be, or what are peoples' workflow that involve replacing sprites, so I will leave this to those who understand this better.

PR #1917, opened 2023-02-11
4.0: declare script compiler extension strings as preprocessor macros

ivan-mogilko:
Wanted to do this long ago...

This lets compilers to return a list of extension names (arbitrary string IDs). These extensions are taken by the Editor and turned into preprocessor macros, in a form of `SCRIPT_EXT_<NAME>`.

The primary purpose is being able to write scripts meant for both new and old compilers, where some code may only work with the new compiler. This may be useful for module writers, but also if you make a test game and need to switch compilers during the test.

For starters, the new compiler declares one generic "AGS4" extension. But it's possible to add one extension name per new individual syntax feature, if necessary, later.

Example of script:
```
#ifdef SCRIPT_EXT_AGS4
    Display("AGS 4 compiler extensions are ENABLED");
#endif
#ifndef SCRIPT_EXT_AGS4
    Display("AGS 4 compiler extensions are DISABLED");
#endif
```

### DIFF
--- a/Compiler/script/cs_compiler.cpp
+++ b/Compiler/script/cs_compiler.cpp
@@ -17,6 +17,13 @@ std::vector<const char*> defaultHeaderNames;
 
 MacroTable predefinedMacros;
 
+void ccGetExtensions(std::vector<std::string> &exts)
+{
+    // TODO: we may consider creating a managed user object an extension, etc,
+    // although it was introduced years ago when extensions were not declared.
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Well, since there's no comments, I will merge this, because it's easy to change later before ags4 becomes a "proper" release.
And this is very useful when writing code for RTTI test.

Issue #1371, opened 2021-07-22
Version-independent save games


hobbesjaap:
**Describe the problem**
As mentioned in [this topic on the AGS Forums](https://www.adventuregamestudio.co.uk/forums/index.php?topic=59293.0) it would be great to improve the save system in AGS. Right now, when I update a game / change something in the game, the saved games don't work anymore. This makes bug-fixing or expanding a game rather tiresome for players, since they have to start from scratch.

**Suggested change**
First change: As Crimson Wizard mentions, the main issue is that AGS lacks a system "of reliable identification of game objects within the game data."

This has multiple underlying issues (which go way beyond my programming knowledge). By resolving these issues, saved games would continue to work even if a game is modified/changed/expanded.

Second change: In a more severe situation (where someone changes the game beyond recognition) we could also consider a system where AGS gives a warning "This save file differs too much from this game - are you sure you wish to proceed?" and then loads the saved game, ignoring all the errors present and only restoring "what it can".

ivan-mogilko:
For the reference: following minimal solution has been implemented in v3.6.2 of the engine:
- Support for restoring saves with *less* data, and running "validation" callback in user script: #2489
- A function that prescans and tests saves for compatibility without actually overwriting current data: #2579

Unfortunately, in this ticket the task is defined rather vaguely. But formally it is not fully covered by aforementioned changes.
We may keep it for the reference until there's a ticket with more thorough requirements.

PR #1922, opened 2023-02-15
AGS 4: Implement script RTTI

### LINKED ISSUE #1259: AGS 4: RTTI for AGS script
*NOTE: RTTI stands for "RunTime Type Information".*

### Overview

When compiled, AGS script currently looses all information about types. This is tolerable with regular structs, because their work is fully controlled by the compiler and interpreter may rely on its instructions. But when it comes to managed structs lack of knowledge on engine side restricts their functionality and use.
The related problems are, but may not be limited to:
* Managed pointers in managed structs. While engine handles managed struct disposal it cannot decrease refcount for other managed objects referenced from that struct, simply because it does not know that the struct has these pointers, nor at which offsets in data. This is why pointers in managed structs were disabled: to prevent memory leaks.
* Dynamic pointer casting. Currently only child-to-parent cast works, and parent-to-child cannot work because there's no way to tell what type the pointed object really is.
* C++/C# - style virtual struct methods, cannot be implemented because as there's no type info so there are no virtual tables and no place to store these. (C-style virtual tables are separate issue, as these would first of all require function pointer type in script, or other way to reference functions.)
* Debugging user objects: would become easier if we knew what they are.

While each of the above problems could be individually solved by its own workaround, my belief is that RTTI would provide more consistent and wholesome solution.

RTTI may come in a form of a table containing information about types. It's generated and written by compiler and/or accompanying tool. This information may be at first limited to our basic needs but later expanded (thus when serialized it also has to have format number).
Each type comes with convenient ID, and when created managed objects would also store this type ID. Thus when engine needs to know something about their type it would use this ID to retrieve the type from the table.

Several years ago we had a conversation about this with other developers and following are points we came to as a result.

---

### Building the RTTI table.

Before we know how to use it or what information to store there, the first question is how to build it.

When compiler compiles a script unit it gathers all the types visible to that script, and information on struct contents which let it calculate memory offsets for all operations. So we already have this info from the perspective of a single compiled unit. Let's say we append this information as an extra segment to the compiled script format. Let's say each table entry would have a numeric type ID which we could pass into "new" operator for user objects.
This sounds fine at first, but having just local tables per script is not going to be sufficient:
* Type duplication; if we rely on local tables we will have same types included so many times as there are script units - in separate tables of course.
* Engine will not be able to know which of those are same type. So if there's any operation that requires type comparison - that won't work between objects allocated by different scripts.
* There may be situations where the script that allocated an object is no longer in memory while the object still is. This may happen if there's a global pointer to base type, and a room script creates an instance of an extended type and assigns to that pointer.
This problem on its own could be worked around by keeping type tables separate from script data in the engine, so that when script is unloaded the type data remains in memory.

One could speculate that, because the order of how script headers are currently included in scripts is always fixed, then the order of types (and their numeric index) will also always be fixed. So it seems like we could use equal local type indexes found in multiple scripts as refering to same type.
In truth that is not correct, as scripts may have types declared inside their bodies - and then their indexes would conflict with types declared in script headers following them in the script list.
But even if above did work, in the long term this would not be a good solution, because that would impose limitations on how scripts are used. Consider if we will change the way script headers are included, especially with stand-alone compiler tools: that will break whole system.

This means that we cannot rely on local type table alone and need the *global type table*.
But how may it be constructed if compiler knows only about one script it compiles at a time?

If I remember correctly, we already have similar mechanism for function binding. Function binding works this way: every script has a list of functions it knows / uses, and when writing call ops it puts their local index in the byte code. Engine uses this local function index to look up in the local table and find function name, under which it is registered in global table.
Perhaps we could use same solution for types. Each compiled script module will have a local table of types where it will map local type index to something that defines its global key. So we would use local numeric id to map to a global (e.g. string) id from scripts's table of types, then this global id is used further to find a type description from global table.

What could be the unique global type ID? For instance it could be a string constructed as a pair of a "scope name" (header/script name?) and a type name.

Now, how a global table is built? I think there may be two approaches here.
1. First is simply building one as we load individual script tables. Engine is capable of doing this. Obviously objects may be allocated only in loaded scripts, so we always will have their type infos ready.
The downside is that each compiled unit will contain full info of all the included types. I don't think it will be large in practice, but still worth noting.
2. An alternative is to have a separate tool that would do same thing as a post-compilation step, and write this global table into special file, called e.g. `script.rtti`. This would provide a full type table for whole game ready to be read at once.
In such case individual compiled units will only have ID mapping in their tables, and it's the global table file that would contain type descriptions.
The upside of this is that we have a type table for the whole project at once, and no data duplication.
The downside though is that we would need extra handling for this file inside a project, need to know when and how to update it, and so forth.

So even if method 2 may be beneficial as an option, guess we'd rather first try method 1.
It may be also possible to set up compiler switch to toggle between saving full type info and ID mapping only, if we'd like to support method 2 in the future.

---

### Using Type IDs in script (bytecode)

Assuming we have a local type table, where entries are identified with numeric IDs, we may pass these IDs as an argument when creating a user object. Current `SCMD_NEWUSEROBJECT` command accepts 1 arg meaning object size. Because object size may also be a part of type info, there are various ways we may go from here:
* Keep `SCMD_NEWUSEROBJECT` but interpret its argument as either size or type ID depending on a "script format" or something.
* Introduce new updated command, e.g. `SCMD_NEWUSEROBJECT2` which accepts type ID. This will work faster (no if switch), but has an extra opcode.

Reading old discussion on this topic I also found a curious proposal to try to merge `SCMD_NEWUSEROBJECT` and `SCMD_NEWARRAY` in one new opcode. Because "dynamic array of T" may also be considered a type on its own, and written into type table, then type info could distinguish "plain" types and "array" types. OR array may be indicated as a flag in an object itself.
If either of this would seem feasible and convenient, then we'll only need one command for allocating anything managed in script.

---

### Type Info in the engine

Like described above, engine would either construct global type table adding entries as it loads various scripts, or by reading a single file.

Supposing `SCMD_NEWUSEROBJECT` or an additional command would have local type ID as an argument, engine creates a user object (currently `ScriptUserObject` struct) by storing three members: global type info ID / index (or pointer to type info), size, and a buffer for instance data.

The necessary contents of type info are dictated by what we want to use it for. For example, if we'd like to be able to release managed pointers we'll need at least a list of offsets of those pointers (handles).
How the engine will deal with recursive object release, or with circular dependencies - I believe this is another topic entirely and may be discussed separately.

If I missed or forgot any potential problems - these may be added to this ticket as we realize them.

ivan-mogilko:
Resolves #1259.

CC @fernewelten , because he is working on a new script compiler, and I added few functions to it.

The detailed description of the purpose is given in the task ticket #1259, here I will give a brief overview.
There are several potential operations within the script interpreter that are impossible to achieve without having an indication and description of object's type and its inner contents:
* Having managed pointers in managed structs and being able to dereference them at runtime when disposing the parent object (thus avoiding memory leaks).
* Dynamic pointer casting (at runtime), parent-to-child type casting specifically.
* Being able to identify the script data in game saves (see #1371).

Following actions would be also potentially easier to implement if we had type indication and content description:
* Virtual methods (would need a runtime type reference mostly, full RTTI tables are probably not needed for this).
* Runtime debugging, such as watching the struct's fields in real-time, for example.

This is where RTTI feature comes from. The idea is to have a table of types and their contents, generated by the script compiler as an extra step, and written either as an extra (optional) part of a script data, or a separate file along with the game project.

This PR does the following:
1. Implements RTTI generation in both supported script compilers (classic ags3 and new ags4 one).
2. Serialize generated RTTI per script, as an extension to the script data.
3. The runtime interpreter creates a joint RTTI collection, updating it after each loaded script.
4. "--print-rtti" command arg, which makes engine print resulting joint collection on the each room load into the log (this seemed to be the most logical place to have this, as room script is the last loaded script).

`--print-rtti` command arg is useful for testing the resulting RTTI table.

Additionally, engine also creates "quick references" within the gathered RTTI, which is basically pointers between type and field structs. This allows for easier traversing of types and respective fields under the debugger.

---

### Generation

Currently the type information consists of an ID, a name, a location (where this type was declared), a parent's ID (if it has a parent), a set of flags which define type's kind, and a collection of fields.
The fully qualified type's name may be generated by combining a location's name (usually header or script name) and a type's name. This is required to be able to distinguish potential unrelated types of same names declared in different scripts.
Fields info contains: relative offset, name, typeid, and a set of flags which define field's kind and qualifiers.
The gathered data may be easily expanded in the future (also see notes to the serialization).

Compilers generate RTTI per each compiled script unit. Because of that the type's ID is a **local ID**, which is relative to this compiled unit only. The engine will have to use fully qualified names to map script's local ID with a global ID in a joint table.

---

### Serialization

The RTTI serialization format is designed having *ease of expansion* in mind. For that purpose it is not done as a single table with types and nested field items inside, but as a number of separate tables, where each table's entry has a *fixed size*. The RTTI header describes tables' offsets, and the fixed sizes of their items. The tables are connected using index references (meaning, an item in table 1 may refer to an item in table 2 using its index).

The advantages of such structure are:
* faster reading;
* faster and easier parsing, if one is e.g. writing a tool that does not read full data, but only wants to find particular item in a file stream;
* **much** easier extension: if you want to expand an existing item then you increase the fixed item's size in header, and even older program will still be able to read new data (although it will only understand old parts of data); if you want to add completely new kind of information then you add a new table, which may be easily skipped by a parser if it does not need it or does not know it.

Following is a format description.

**RTTI header**

field | type / size | comment
--|--|--
format | uint32 | for expanding the rtti format
header size | uint32 | size in bytes of a header (counting from "format" field, until header ends)
full rtti size | uint32 | size in bytes of a rtti data (counting from "format" field, until data ends)
location entry size | uint32 | fixed size of a location's description in bytes (may depend on format)
locations table offset | uint32 | a relative position of a locations table (in file)
num locations | uint32 | total number of location entries
type entry size | uint32 | fixed size of a type's description in bytes (may depend on format)
types table offset | uint32 | a relative position of a types table (in file)
num types | uint32 | total number of type entries
field entry size | uint32 | fixed size of a type field's description in bytes (may depend on format)
fields table offset | uint32 | a relative position of a type fields table (in file)
num type fields | uint32 | total number of all fields in all types
string table offset | uint32 | a relative position of a strings table (in file)
string table size | uint32 | total size of a string table, in bytes

**RTTI tables**
field | type / size | comment
--|--|--
location table | num locs * loc size | see "Location description" below
type table | num types * type size | see "Type description" below
fields table | num type fields * field entry size | see "Field description" below
string table | string table size | all RTTI null-terminated strings packed in a single array (separated by 0s)

**Location description**
field | type / size | comment
--|--|--
local id | uint32 | local ID of this location
name | uint32 | an offset of a name in a string table

**Type description**
field | type / size | comment
--|--|--
local id | uint32 | local ID of this type
name | uint32 | an offset of a name in a string table
location id | uint32 | ID of location this type was declared at
parent type | uint32 | local type ID; 0 if no parent
type flags | uint32 | may contains helper flags which simplify analyzing this type
size | uint32 | in bytes
num fields | uint32 | number of member fields this type has, 0 if none
field table index | uint32 | index of the first field in the fields table

**Field description**
field | type / size | comment
--|--|--
offset | uint32 | relative offset of this field, in bytes
name | uint32 | an offset of a name in a string table
type | uint32 | this field's local type ID
type flags | uint32 | may contains helper flags which simplify analyzing this member
num elements | uint32 | number of (array's) elements

---

### TODO

- [x] Make sure there's type with typeid 0 with a meaning "No Type" generated by each compiler (I think old compiler does not do this atm).
- [x] There's a nasty problem with fully qualified names taking extra space in a file and mem because of the repeated location (script) name. I'm investigating options, but possibly the type desc may contain a "location id" instead, and then either 1) location names are stored separately, 2) location names taken from section names in basic script data, 3) create a new table of "locations" inside RTTI (which may have some extra uses in the future).
- [x] Compiler should only put member name for fields, currently one or both of the compilers put Type::Member there instead, which is redundant.
- [x] Perhaps better format the `--print-rtti` output.
- [x] Might add explicit compiler option for generating & saving rtti.
- [x] I made Fields that are not array save "num elems" as 1, but maybe that's redundant (need to think this over).

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -109,6 +109,8 @@ target_sources(common
     script/cc_common.cpp
     script/cc_common.h
     script/cc_internal.h
+    script/cc_reflect.cpp
+    script/cc_reflect.h
     script/cc_script.cpp
     script/cc_script.h
     util/aasset_stream.cpp
--- a/Common/script/cc_common.h
+++ b/Common/script/cc_common.h
@@ -29,8 +29,9 @@
 #define SCOPT_NOIMPORTOVERRIDE 0x0020 // do not allow an import to be re-declared
 //#define SCOPT_LEFTTORIGHT 0x40   // left-to-right operator precedance (DEPRECATED)
 #define SCOPT_OLDSTRINGS    0x0080   // allow old-style strings
-#define SCOPT_UTF8          0x0100  // UTF-8 text mode
-#define SCOPT_HIGHEST       SCOPT_UTF8
+#define SCOPT_UTF8          0x0100   // UTF-8 text mode
+#define SCOPT_RTTI          0x0200   // generate and export RTTI
+#define SCOPT_HIGHEST       SCOPT_RTTI
 
 extern void ccSetOption(int, int);
 extern int ccGetOption(int);
--- a/Common/script/cc_internal.h
+++ b/Common/script/cc_internal.h
@@ -18,8 +18,10 @@
 #ifndef __CC_INTERNAL_H
 #define __CC_INTERNAL_H
 
-#define SCOM_VERSION 90
-#define SCOM_VERSIONSTR "0.90"
+#define SCOM_VERSION_OLD  90
+#define SCOM_VERSION_EXT  95
+#define SCOM_VERSION      SCOM_VERSION_EXT
+#define SCOM_VERSIONSTR   "0.95"
 
 // virtual CPU registers
 #define SREG_SP           1     // stack pointer
--- a/Common/script/cc_reflect.cpp
+++ b/Common/script/cc_reflect.cpp
@@ -1,0 +1,277 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// http://www.opensource.org/licenses/artistic-license-2.0.php
+//
+//=============================================================================
+#include "script/cc_reflect.h"
+#include <string> // std::string
+#include <string.h> // memcpy etc
+#include "util/stream.h"
+
+namespace AGS
+{
+
+using namespace Common;
+
+// Adds a string into the string table; saves associated offset, at which
+// this sting will be packed, and increases packed length.
+// Returns last saved offset if the string is already in the table.
+static uint32_t StrTableAdd(std::map<std::string, uint32_t> &table,
+    const std::string &str, uint32_t &packed_len)
+{
+    const auto str_it = table.find(str);
+    if (str_it != table.end())
+        return str_it->second;
+    
+    table[str] = packed_len;
+    uint32_t last_len = packed_len;
+    packed_len += str.size() + 1; // count null-terminator
+    return last_len;
+}
+
+//*****************************************************************************
+// RTTI serialization
+//
+// RTTI header:
+// ----------------
+//  uint32 | format                 | for expanding the rtti format
+//  uint32 | header size            | size in bytes (counting from "format")
+//  uint32 | full rtti data size    | size in bytes (counting from "format")
+//  uint32 | type entry size        | fixed size of a type info in bytes
+//  uint32 | types table offset     | a relative pos of a types table
+//  uint32 | num types              | number of types in table
+//  uint32 | field entry size       | fixed size of a type field info in bytes
+//  uint32 | fields table offset    | a relative pos of a type fields table
+//  uint32 | num type fields        | number of fields in table
+//  uint32 | string table offset    | a relative pos of a strings table
+//  uint32 | string table size      | size of a string table, in bytes
+//
+// Type Info:
+// ----------------
+//  uint32 | fully qualified name   | an offset in a string table
+//  uint32 | local id               | local type ID
+//  uint32 | parent type (local id) | local type ID; 0? if no parent
+//  uint32 | type flags             |
+//  uint32 | type size              | in bytes
+//  uint32 | num fields             |
+//  uint32 | field table index      | an index of a first field
+//
+// Type Field Info:
+// ----------------
+//  uint32 | offset                 | relative offset, in bytes
+//  uint32 | name                   | an offset in a string table
+//  uint32 | type (local id)        | local type ID of this field
+//  uint32 | flags                  |
+//  uint32 | number of elements     | for arrays, 0 = single var
+//
+//*****************************************************************************
+
+void RTTI::Read(Stream *in)
+{
+    // RTTI Header
+    const soff_t rtti_soff = in->GetPosition();
+    const uint32_t format = in->ReadInt32();
+    const size_t head_sz = (uint32_t)in->ReadInt32();
+    const size_t full_sz = (uint32_t)in->ReadInt32();
+    const size_t typei_sz = (uint32_t)in->ReadInt32();
+    const uint32_t typei_table_off = (uint32_t)in->ReadInt32();
+    const uint32_t typei_table_len = (uint32_t)in->ReadInt32();
+    const size_t fieldi_sz = (uint32_t)in->ReadInt32();
+    const uint32_t fieldi_table_off = (uint32_t)in->ReadInt32();
+    const uint32_t fieldi_table_len = (uint32_t)in->ReadInt32();
+    const uint32_t str_table_off = (uint32_t)in->ReadInt32();
+    const size_t str_table_sz = (uint32_t)in->ReadInt32();
+
+    const soff_t typei_soff = rtti_soff + typei_table_off;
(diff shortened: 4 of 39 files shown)


ivan-mogilko:
I decided to do some refactor on this pr, because in the current variant the RTTI struct has too many internal "mechanics" exposed, so it's too easy to produce an invalid state.

ivan-mogilko:
Added a separate "locations" table. Now the fully qualified type's name is generated at runtime by combining location's name and type's name.

AlanDrake:
Does this PR have any impact on script performance?

ivan-mogilko:
> Does this PR have any impact on script performance?

A little longer script loading.

I was planning to make one last refactor tbh in the nearest days (I got distracted by the latest 3.6.0 bugs), because I put everything in script.h/cpp, and one class may be split into two for more consistent behavior, but this may also be done later too.

ivan-mogilko:
Okay, I think this is done. I had doubts about how the structs are organized, but I am maybe overthinking again, and it should not be impossible to refactor into something more convenient (if a need arises). The serialization format is something I am certain about though.

As been mentioned, it's possible to completely disable rtti generation by toggling a compiler option.

Issue #1818, opened 2022-10-23
AGS 4 / new compiler: support dynamic arrays of normal structs


ivan-mogilko:
I've been wondering, if it's possible to support dynamic arrays of non-managed structs?
For example:

```
struct Cell
{
    int data;
};

Cell cells[];
```
Above results with error in the extended compiler:
> Can only have dynamic arrays of integer types, 'float', or managed structs. 'Cell' isn't any of this.

(old compiler does not error on declaration, but on the `new` operator instead)

Right now this is restricted in both old and new compilers. Although, theoretically, you should be able to store an element of any size in dynamic arrays. Perhaps there have been a problem in coding accessing these elements, or anything else I cannot think of right now?

fernewelten:
At first sight I'd think that you're right, this _should_ work. I don't see any problems that could come up if this limitation were removed. We'd need to make sure, however, that the struct doesn't contain any dynamic entities, either directly or indirectly.

fernewelten:
Yes, I could implement this, by eliminating the error message and adding a check to make sure that `Cell` doesn't contain any dynamic arrays or pointers. Do we want to have it? 

From the vantage point of the game coders, the code would be very similar to the code below that both the compilers can compile:
```
managed struct Cell
{
    int data;
};

Cell *cells[];
```

In the example above, each cell would be `null` and would need to be initialized individually with a ` cells[…]= new Cell;`. In Ivan's code, above, all the cells would be ready for use at the get-go.

ivan-mogilko:
Note that compiler will still have to deny dynamic arrays of structs that have nested managed pointers, until #1923 is completed.

ericoporto:
A note, that #1923 was finished and is working, so perhaps this could be done now with less checks.

fernewelten:
Addressed in #2106

ivan-mogilko:
Implemented by #2106.

Example of use:
```
struct Warrior
{
    int Strength;
    int Health;
};

function room_FirstLoad()
{
    Warrior Lemmings[] = new Warrior[10];
    player.Say("That one's strength is %d", Lemmings[1].Strength);
}
```

PR #1923, opened 2023-02-15
AGS 4: properly support managed pointers in managed structs

ivan-mogilko:
**WARNING:** this requires #1922 to work.
(pr's branch is based on top of that pr for easier testing)

This resolves a long existing problem of AGS script not supporting managed pointers inside user's managed structs.
Such syntax feature was explicitly disabled, because the runtime interpreter did not have any information on the struct's contents, only about it's size. When the user's managed object is disposed, engine cannot unreference the pointers contained in such struct, which leads to nested objects never disposed, occupying program memory (sort of "memory leak").

With addition of RTTI it is now possible to gather information about pointers contained within a type, and correctly unref all them on type's instance's disposal.

This PR introduces new script op code: `SCMD_NEWUSEROBJECT2`, which has 2 args: typeid and size.
The typeid is a local type id from the script's RTTI table.
Runtime interpreter converts local type id to a global one when creating a user object.

After loading scripts, engine generates a "flat" table of managed pointer offsets per struct. This speeds up user object disposal, as we don't have to analyze all the nested types all the time. Instead we do that only once the script is loaded, and then use a plain list of offsets.

Managed pointers in managed structs are enabled in both compilers, old and new one.

Both compilers declare extension "NESTEDPOINTERS", and a script author can use SCRIPT_EXT_NESTEDPOINTERS macro to test whether this feature is supported.

**TESTING**

The idea is that a structure of any complexity should be correctly disposed and removed from the memory when all references to it are gone. For example, you may have a managed struct that contains a particularly large dynamic array, and continuously create instances of that struct, observing program's memory usage.

```
#ifdef SCRIPT_EXT_NESTEDPOINTERS
managed struct ManagedStruct
{
    String Name;
    int BigArray[];
};

managed struct NestedPointers
{
    ManagedStruct* sp;
    ManagedStruct* sparr[];
};

{
    if (k == eKeySpace) {
        NestedPointers *t = new NestedPointers;
        t.sp = new ManagedStruct;
        t.sp.BigArray = new int[1000000];
        t.sp.Name = "some name";
        t.sparr = new ManagedStruct[10];
        for (int i = 0; i < 10; i++)
        {
            t.sparr[i] = new ManagedStruct;
            t.sparr[i].BigArray = new int[1000000];
            t.sparr[i].Name = String.Format("name %d", i);
        }
        // NestedPointers should be fully disposed here
    }
}
#endif // SCRIPT_EXT_NESTEDPOINTERS
```

**ISSUES / TODO:**
- [x] Also might need a new op code for creating a dynamic array with typeid, for knowing which type does they store. This may potentially allow dynamic arrays of *regular structs*, as they may contain nested managed pointers with more pointers inside, but neither compiler nor engine won't know how to properly unref these.
- [x] Must update the savegame format of a "user object", right now these will loose the type id when restoring a save.
- [x] Must update new compiler's tests.
- [x] Might add explicit compiler option for generating & saving rtti / allowing pointers in man structs.
- [x] Save game problem (see explanation [in comments below](https://github.com/adventuregamestudio/ags/pull/1923#issuecomment-1444370882)).

### DIFF
--- a/Common/script/cc_common.h
+++ b/Common/script/cc_common.h
@@ -31,7 +31,8 @@
 #define SCOPT_OLDSTRINGS    0x0080   // allow old-style strings
 #define SCOPT_UTF8          0x0100   // UTF-8 text mode
 #define SCOPT_RTTI          0x0200   // generate and export RTTI
-#define SCOPT_HIGHEST       SCOPT_RTTI
+#define SCOPT_RTTIOPS       0x0400   // enable syntax & opcodes that require RTTI to work
+#define SCOPT_HIGHEST       SCOPT_RTTIOPS
 
 extern void ccSetOption(int, int);
 extern int ccGetOption(int);
--- a/Common/script/cc_internal.h
+++ b/Common/script/cc_internal.h
@@ -107,6 +107,7 @@
 #define SCMD_DYNAMICBOUNDS 71   // check reg1 is between 0 and m[MAR-4]
 #define SCMD_NEWARRAY     72    // reg1 = new array of reg1 elements, each of size arg2 (arg3=managed type?)
 #define SCMD_NEWUSEROBJECT 73   // reg1 = new user object of arg1 size
+#define SCMD_NEWUSEROBJECT2 74  // reg1, reg2 = new user object of arg1 type and arg2 size
 
 #define CC_NUM_SCCMDS     74
 #define MAX_SCMD_ARGS     3     // maximal possible number of arguments
--- a/Common/script/cc_internal.h
+++ b/Common/script/cc_internal.h
@@ -106,10 +106,11 @@
 #define SCMD_JNZ          70    // jump to arg1 if ax!=0
 #define SCMD_DYNAMICBOUNDS 71   // check reg1 is between 0 and m[MAR-4]
 #define SCMD_NEWARRAY     72    // reg1 = new array of reg1 elements, each of size arg2 (arg3=managed type?)
-#define SCMD_NEWUSEROBJECT 73   // reg1 = new user object of arg1 size
-#define SCMD_NEWUSEROBJECT2 74  // reg1, reg2 = new user object of arg1 type and arg2 size
+#define SCMD_NEWUSEROBJECT 73   // reg1 = new user object of arg2 size
+#define SCMD_NEWUSEROBJECT2 74  // reg1 = new user object of arg2 type and arg3 size
+#define SCMD_NEWARRAY2    75    // reg1 = new array of reg1 elements, arg2 type and arg3 size
 
-#define CC_NUM_SCCMDS     75
+#define CC_NUM_SCCMDS     76
 #define MAX_SCMD_ARGS     3     // maximal possible number of arguments
 
 #define EXPORT_FUNCTION   1
--- a/Common/script/cc_internal.h
+++ b/Common/script/cc_internal.h
@@ -109,7 +109,7 @@
 #define SCMD_NEWUSEROBJECT 73   // reg1 = new user object of arg1 size
 #define SCMD_NEWUSEROBJECT2 74  // reg1, reg2 = new user object of arg1 type and arg2 size
 
-#define CC_NUM_SCCMDS     74
+#define CC_NUM_SCCMDS     75
 #define MAX_SCMD_ARGS     3     // maximal possible number of arguments
 
 #define EXPORT_FUNCTION   1
--- a/Common/script/cc_reflect.cpp
+++ b/Common/script/cc_reflect.cpp
@@ -38,15 +38,21 @@ static uint32_t StrTableAdd(std::map<std::string, uint32_t> &table,
     return last_len;
 }
 
-// Copies a string from one string table to another at a new location.
-// Returns the string's offset in the new_table.
-static uint32_t StrTableCopy(std::vector<char> &new_table,
-    const std::vector<char> &old_table, uint32_t old_pos)
+// Copies a string to a string table at a new location.
+// Returns the string's offset in the str_table.
+static uint32_t StrTableCopy(std::vector<char> &str_table, const char *string)
 {
-    const size_t old_packsz = new_table.size();
-    const size_t new_strsz = strlen(&old_table[old_pos]) + 1; // count null-terminator
-    new_table.resize(new_table.size() + new_strsz);
-    memcpy(&new_table.front() + old_packsz, &old_table[old_pos], new_strsz);
+    if (!string || !string[0])
+        return 0u; // assume string table starts with "null-terminator" slot
+    // If the string belongs to our string table already, then return existing offset;
+    // otherwise - copy this string over to our strings
+    if (!str_table.empty() && (string >= &str_table.front() && string <= &str_table.back()))
+        return string - &str_table.front();
+
+    const size_t old_packsz = str_table.size();
+    const size_t new_strsz = strlen(string) + 1; // count null-terminator
+    str_table.resize(str_table.size() + new_strsz);
+    memcpy(&str_table.front() + old_packsz, string, new_strsz);
     return static_cast<uint32_t>(old_packsz);
 }
 
@@ -317,7 +323,7 @@ void RTTIBuilder::AddField(uint32_t owner_id, const std::string &name,
     _fieldIdx.insert(std::make_pair(owner_id, fi)); // match field to owner type
 }
 
-RTTI &&RTTIBuilder::Finalize()
+RTTI RTTIBuilder::Finalize()
 {
     // Save complete string data
     _rtti._strings.resize(_strpackedLen);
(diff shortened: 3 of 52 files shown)


ivan-mogilko:
So... in overall this is ready for testing. The only mandatory thing missing is compiler tests update, where the bytecode tests are failing because of the new opcodes (and different number of args). I will look into this soon.

I put it back to draft, because I suddenly realized there's a serious loophole with game saves. I will explain this later, in the second part of the comment.

For now, the manual tests might include following:
* A script that includes managed user structs and dynamic arrays.
* The structs now can be composed of **any** combination of nested regular and managed structs.
* Dynamic arrays cannot contain regular structs at the moment (see #1818), so there's no way to test these, but the engine has (supposedly) a necessary code to dispose these correctly (will have to test that separately when #1818 is resolved).
* The managed structs containing nested pointers should be correctly disposed as soon as the last reference to them is removed, leaving no memory leaks.
* Serialization: storing and restoring game saves should work correctly with managed structs in the script mem. (But see a note  below.)
* Structs should still dispose correctly after restoring a save, this would mean that they keep type ids.

**NOTE**
Old saves won't work, because the UserObject and DynamicArray serialization has changed. But it's possible to add compatibility fix and allow to restore these too. This may be done because I assigned new slightly different type names to these script objects, so it's possible to distinguish from the previous ones (unfortunately there's no other reliable way now).

---

**The game save issue**

While I've been writing this, it suddenly came to me that the (un)serialized typeids may get broken if RTTI changes. And more...
There are following major cases here that I may think of:

1. New types added or their order changes, in which case the numeric IDs may shift. To counter this we may write a simple typename-typeid (string to number) map into the save. That should not take too much extra space. When restoring a save, we first use typenames to find matches in the current game's RTTI, and then use this map of matches to remap loaded user objects and dynamic arrays to the new RTTI. This is similar to how individual script's RTTI is remapped to a global table on script load.

2. Existing type names changed in script. That will break the object matching completely, because there will be no way to find out if it's the same type or not. I don't know if anything may be done with this, or should be... Given problem 1 is resolved, I might throw an error if no match was found. But this is somewhat annoying...

3. The struct's contents change... Now, this could be an issue before as well. In such case a restore object data may be less or larger than the type's, or contain fields in an old order. After this addition, it will also may cause memory leaks, if pointers are added, removed, or reordered, as the RTTI will contain the new information. *Hypothetically*, this may be resolved by storing RTTI, or necessary bits (like managed field offsets per type) in game's save. But I don't know if all this is necessary at this point.

ivan-mogilko:
Eh, thinking more about game saves problem; there's another major issue: the types declared in room scripts.

Because no room scripts are loaded at the time when the save is restored (and only 1 particular room will be loaded later), we won't be able to "match" the objects created from room types.

So, the only solution I see at the moment is to have either full RTTI written in saves, or at least a limited pieces of data necessary for supporting the features (like nested pointers in our case). Then, upon restoring a save, create "temporary types", and rematch these as more scripts are loaded.

ivan-mogilko:
Hmm, well, with the latest changes the "saves problem" seems to be formally resolved (in a minimal necessary way); but the code is quite ugly, so I'll take couple more days to tidy it, and then write an explanation of what it does.

ivan-mogilko:
Alright, it took me unexpected extra time to figure out something about restoring saves (because I did not realize this problem early enough), but I *suppose* that current implementation has a "minimal necessary" solution for it. By that I mean that it will formally work, except for a few niche cases. It will definitely work for a unchanged (finalized) game script. There are options for enhancing this in the future, both resolving some of the niche cases, and improving performance.

But, general words aside, following is a quick recap of problems with saves, and how the engine will deal with that now.

1. As the script compiler passes typeid as a number in the `new` opcode, this number is local (relative) to the current script. That's why it's called a local typeid. Local typeids are resolved into global ones at runtime, because this is when we have a reliable joint RTTI table (this is similar to how global imports are resolved). This is done using their actual type names.
2. All the managed structs, and arrays, are created by storing these global numeric ids. They are also saved and restored with these global ids.
3. The problem arises if the game script was changed between save and restoring that save. Both local and global numeric ids are tied up to the *order of registration* (declaration at compile time). This means that if you move the types around, or add more types somewhere, the order of previously existing types registration would change. Well, that's normal, this is why we don't rely on their order anyway, and the type's name is the only unique identifier.
4. The second problem comes from types *declared in room scripts*. As room scripts are loaded only when player is in the room, there may be objects in memory which have such "room type" loaded before the room was loaded, so we basically do not know this type yet. This problem is specifically important if the room's type is a child of a parent global type. In such case the object of room's type may be assigned to a pointer of parent's type in a global variable, addressed by the global script, and even deleted in the global script. And we *must* know some info about the actual time for deletion (so that we dispose all the nested managed pointers properly).

This is how I tried to solve this.

1. Write a "stripped" global RTTI into the save. In theory we could dump whole RTTI there if necessary, but I went with a partial for now, to save disk space. For instance, it does not write full struct fields info, but only the precalculated list of managed pointer offsets per types which have these.
2. When restoring the save, this RTTI will be restored, and joined with the current global RTTI. This does two things: if a type with the same name already exists (normally that would be every type declared in the regular script modules), then it only adds the restored -> real typeid conversion to the map. This map is then used to remap typeids in all the loaded managed objects (that need it), so they will now contain global typeids corresponding to the current game.
3. If a type is not known yet (will usually happen for types declared inside a room script), then it adds a *placeholder* type (I called them "generated") marked by a special flag. If this placeholder type contained any managed pointers, it will also generated pseudo-fields for this type, just enough for the engine to subref these on object's disposal.
4. As the game progresses, and rooms are loaded, the room's own RTTI will be further joined, and once we have a real type from the room script, they will replace the "placeholders", while keeping the global typeids (so no more typeid remapping is needed).

This way
* the managed objects always have typeids corresponding to the current global RTTI table
* the objects of types declared in unloaded (room) scripts may still be safely disposed.

ivan-mogilko:
Well, I think this is working, overall.

I made certain that the new opcodes and RTTI generation itself could be disabled completely from compilation, in which case compilers will write old opcodes and also error on nested managed pointers (see compiler options SCOPT_RTTI and SCOPT_RTTIOPS. So the whole feature may be turned off if something bad happens.

The object disposal is fine, at least according to my tests. My only doubts at the moment are related to saves problem, as mentioned above, but this is maybe because the whole situation about AGS saves relying on exact game data state is ugly. The implemented solutions results in save files being several tens of Kilobytes larger, and having extra processing after the load. It might be possible to optimize probably, and also possible to replace with some other solution at the cost of changing the save format in particular component.

I think the biggest TODO after this will be to implement some kind of a managed pool statistics, which are gathered and optionally printed somewhere. Testing proper disposal manually may be a tedious task, and only suitable if you have few big objects in memory - then you can clearly notice the memory change using trivial system tools. But this kind of test may not be suitable for a regular game run.

In regards to this PR, the only thing remaining is fixing compiler tests.

ivan-mogilko:
Eh, had to fix couple of bugs in dynamic array, happened during some rebase probably.

ivan-mogilko:
So, there's another problem which I forgot about: it's circular dependencies. The objects not referenced from script may still reference each other, which keeps them in memory until the program ends.
I never researched how this is done in other managed languages, but from a quick check they somehow are able to distinguish whether reference is from a script or internally from an object, and then consider every object that does not have a "script" reference a garbage.

Secondly, the current subref/dispose algorithm causes the infinite recursion.

ivan-mogilko:
Implemented a GC, capable of dealing with the circular dependencies and "detached" objects.
Used the [python dev docs](https://devguide.python.org/internals/garbage-collector/) for the reference on a concept.
Ours is not as efficient, of course; that would take a good rewrite of a object storage structure to make it better.

There are now 2 subref methods:
* SubRefAndCheck will subref and try if disposal is necessary; this is to use from a script interpreter, and engine code.
* SubRefNoCheck will subref only and leave a "dangling" object for GC to collect; this is to call when cleaning internal references inside a managed object. It's used to avoid deep recursions, and unwanted side-effects.

ivan-mogilko:
To speak honestly of performance effect, I've been testing this branch with couple of "script heavy" games for the last couple of days, and I *do* see a change, best visible when running in "infinite fps" mode, although perhaps not a critical one.

In one game (cannot link it atm, because it's been a private prototype, but maybe will open later since it was never finished anyway), which is generally pretty slow in AGS (it's a 3D FPS space flight sim prototype with hundreds of "particles") the fps went down from **66** to about **63**. This games has a lot of dynamic array manipulations and matrix math.

Another example is the [racing game](https://github.com/ivan-mogilko/ags-lastfurious), which in turn has lots of small managed structs used for vector calculations, shows a drop from around **156** to around **144** in average on my PC.

In the recent commits I added few notes in the code about possible optimization. I haven't tested these suggestions yet, so cannot be 100% certain, but believe that they may improve performance to some degree.

ericoporto:
Hey, did something change in this PR for the rest of AGS? I can't call `Overlay.CreateGraphical` anymore in the AGS built from it.

ivan-mogilko:
After the #1990 the script performance got improved significantly, and it's time to compare once again:

Testing the [3D space sim prototype](https://github.com/ivan-mogilko/ags-wcproto)
Latest master branch: 88-90 fps
Latest ags4 branch, old compiler: 87-88 fps
This PR, old compiler: 85-86 fps
Latest ags4 branch, new compiler (somewhat optimized bytecode): **110-111 fps**
This PR, new compiler: 107-109 fps

fernewelten:
From originally 65 fps up to 110 fps: that's great improvement!

So the combination seems to do it best – a more efficient bytecode interpreter **and** more efficient bytecode.

fernewelten:
I'm not quite up-to-date yet: Do we do __both__ reference counting and trace-and-sweep garbage collection to reclaim unused memory? If so, can we perhaps do away with the reference counting and only rely on the collector? Because I __think__ I read somewhere some time ago that reference counting can potentially be slower than garbage collecting, amortized over bytecode instructions executed. I don't know whether that could be relevant in our system.

ivan-mogilko:
> I'm not quite up-to-date yet: Do we do both reference counting and trace-and-sweep garbage collection to reclaim unused memory? If so, can we perhaps do away with the reference counting and only rely on the collector?

The GC I wrote uses reference counter to find out which objects are still attached to pointers in script (or from internal storage in the engine). It uses the idea described here: https://devguide.python.org/internals/garbage-collector/

In order to do this without ref counter, we'd have to keep records of all the pointer variables in global script memory at least, and then trace managed object starting from those variables that have a valid handle. From what I understood this uses concept of a "gcroot" object. For that to work we'd need to generate a list of pointers in script data, similarly to how the RTTI is generated now. This was my first thought, but I decided to use gc refcount method described in the linked article, as it did not require any additional script data.

ivan-mogilko:
@fernewelten oh, by the way, there are not explicit bytecode instructions that do ref count change. Ref count is changed on pointer assignment, that would be MEMWRITEPTR, MEMINITPTR and MEMZEROPTR.

fernewelten:
When we have structs that contain pointers, we can do this:
```
struct Foo;

struct Foo {
    int Payload;
    Foo *Pointer;
};

function game_start()
{
    Foo *local = new Foo;
    local.Payload = 77;
    local.Pointer = local;
}
```

Now, immediately before the function ends, there is a dynamic object that has 2 references: One reference is from `local` and the other is from within the object itself, namely from `local.Pointer`. 
image

Now the function ends and `local` is freed. The reference from `local` goes away, and the object isn't reachable any longer, but the dynamic object has 1 reference remaining. This will never change, the number of references will never become 0. 

So the memory for the object is __lost__ if we only do reference counting.  (Any more elaborate circles will have the same effect.)
image

As far as I understand, a trace-and-sweep gc finds these cases by following all the pointers on the stack recursively, taking care not to get into infinite loops (e.g., by marking each object it has visited so that it can skip processing it when reaching it the second time). That's the “trace”. Then those nodes that the gc has __not__ traced are known to be unreachable and can be freed. That's the “sweep”. 

I'm not a specialist, but that's the gist of gc as far as I have heard of it.

ivan-mogilko:
Well, this is generally working, the only issue I see here is testing. Manual testing will only show that the pointers themselves work.

What could be a convenient way to add some automated aid here? For example, I could add a compilation flag for logging managed pool stats, and then print number of allocated objects, once per N object allocations.

ericoporto:
It all depends on the problems you want to detect, so what kind of problems the GC can have?
- an object is never collected, but should
- time taken to run GC methods (maybe they could eat a significant time at some cliff of objects/relationships?)
- something memory related (I think this would be hard to measure without special allocators so we can skip this)
In general my head can't think many possible things that could go wrong there - beyond a case where the behavior itself is wrong, but the n I don't know how logging something could help.

There could also be something similar to Ctrl+G to print some stats in a Display (maybe add to the Debug things).

ericoporto:
In theory, if we could run it on a script and get some picture of it's state, say before and after some script point, we could write a test to verify it's not incorrect.

If there was some way to dump the GC state to an String at runtime these tests could be written in [AGS Script](https://github.com/ericoporto/tap/blob/01d6fbc367aaf31703bdb0b4935f1682fed1347e/tap_demo/room2.asc#L14) itself with regular Editor and Engine.

Problem is, I think the different states even on an empty game would have a lot of elements you don't care I think, so one may need to figure something to parse out the parts of the state that don't matter. Alternatively these could only spit these states to a text file.

Of course this testing, for now, is something we would make and run at one time, and not something to continuously verify games in general, I don't have many ideas on that.

ivan-mogilko:
Added some trivial collected stats to the managed pool, printed every now and then after N GC runs, on room change and game exit.

Issue #1618, opened 2022-04-17
Support asynchronously saving the game


ericoporto:
**Describe the problem**
Saving the game in AGS may produce a stutter, because it saves in a way that blocks the game loop. This is usually fine when the player is manually saving the game, but may produce visible stutter when done in a Autosave like way, where the game is saved after a room fades in or some in-game event happens. This is less perceptible in Windows since the OS is able to alleviate, but may be more perceptible in other systems.

**Suggested change**
There are different ways to go here:

- it could have some script function so that you schedule a save and then either through polling or an engine event you find out that the save went alright (enabling the developer to show and fade an icon on screen indicating the saving went ok).
- Alternatively, some append only buffered writing stream of sort, where you write to ram and the engine figures a way to write to the disk at a later time. Perhaps this writing mode could even be available to File API one day - just to explain a why to this alternative.
- some other way I haven't considered...

Essentially this means that until the game is saved, the state of the game at that point has to be duplicated in RAM to allow the game to proceed forward without waiting a write to disk.

Whatever the way, we don't want this in the Emscripten port since it will write to ram anyway for now, and also because we don't want threads there (in case this is enabled through threads).

ivan-mogilko:
To clarify: is the goal to keep game fully running, or prevent the game window/screen "freezing" because the system events don't poll?

ivan-mogilko:
EDIT:
> Essentially this means that until the game is saved, the state of the game at that point has to be duplicated in RAM to allow the game to proceed forward without waiting a write to disk.

Literally duplicating the game state may be quite complicated with the current state of the engine code.

Writing into the ram is already supported using a memory stream class, so that may be an alternative. That is virtually a matter of providing a temporary buffer and substituting one stream object with another for the "save game" function call. Then test and see how much faster that is.

One extra concern I have, if the game has _alot_ of dynamic objects in use, that might push it closer to user's system limits.

ericoporto:
> is the goal to keep game fully running, or prevent the game window/screen "freezing" because the system events don't poll?

I don't understand what is the difference between the two options. I observe in Linux when an autosave happens in a game, the game tends to lock until the write to disk finishes, usually the screen and everything seems to lock by some time.

> Writing into the ram is already supported using a memory stream class, so that may be an alternative.

[...]

Issue #1349, opened 2021-07-08
Refactor AGS for 1 game loop and game state managing


ericoporto:
AGS has many little internal loops, I proppose we refactor to having one loop and a way to manage game states. This should clarify in which game state it's and make it easier to work with things like updating the screen. There are weird things going on, like game responsabilities in graphics drivers to support fading!

List of all loops:
- **Main Game Loop**
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/main/game_run.cpp#L754
- **Inner Game Loop** (ShakeScreen)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/global_screen.cpp#L63
  - Has two inner loops, one for software and other for hardware gfx rendering of shakescreen
- **Inner Game Loop** (Software CrossFade)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/event.cpp#L308
- **Inner Game Loop** (Software Dissolve Fade)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/event.cpp#L338
- **Inner Game Loop** (Inventory Screen)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/invwindow.cpp#L354
-  **Inner Game Loop** (Dialog Processing)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/dialog.cpp#L853
- **Inner Game Loop** (Display blocking)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/ac/display.cpp#L273
- **Inner Game Loop** (OpenGL Fade)
https://github.com/adventuregamestudio/ags/blob/6a2a2c14b45a12b2b4a38d3feefd8c0574fda67b/Engine/gfx/ali3dogl.cpp#L1714
- **Inner Game Loop** (OpenGL Boxout)
[...]

PR #1928, opened 2023-02-18
Add emscripten improvements

ericoporto:
- AGS Script errors and messages use Alert boxes on browsers too
- A new directory is used by the engine, `/home/web_user/saved_games`, files saved there are persisted (config and save slots)
- A new modal ui is added to be able to clear these files
- We leverage this modal ui to add an extra button for toggling fullscreen. Exiting fullscreen is accomplished by using Esc key (in mobile browsers, default exiting of fullscreen using back gesture also works). Alt+Enter still works the same.

I already updated here https://ericoporto.github.io/agsjs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,9 +178,10 @@ if(EMSCRIPTEN)
 
     set(EMSDK_FLAGS_LINKER "${EMSDK_FLAGS_LINKER} \
         -s FULL_ES2=1 \
-        -s FORCE_FILESYSTEM=1 -lidbfs.js  \
-        -s EXPORTED_FUNCTIONS=['_main'] \
-        -s EXTRA_EXPORTED_RUNTIME_METHODS=['ccall','callMain']")
+        -s FORCE_FILESYSTEM=1 -lidbfs.js \
+        --post-js ${CMAKE_SOURCE_DIR}/Emscripten/post.js \
+        -s EXPORTED_FUNCTIONS=['_main','_ext_syncfs_done'] \
+        -s EXPORTED_RUNTIME_METHODS=['ccall','callMain']")
 
     set(EMSDK_FLAGS_LINKER "${EMSDK_FLAGS_LINKER} ${EMSDK_DEBUG_LINKER_FLAGS} --shell-file ${CMAKE_SOURCE_DIR}/Emscripten/launcher_index.html ")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ if(EMSCRIPTEN)
         -s FULL_ES2=1 \
         -s FORCE_FILESYSTEM=1 -lidbfs.js \
         --post-js ${CMAKE_SOURCE_DIR}/Emscripten/post.js \
-        -s EXPORTED_FUNCTIONS=['_main','_ext_syncfs_done'] \
+        -s EXPORTED_FUNCTIONS=['_main','_ext_syncfs_done','_ext_toggle_fullscreen','_ext_get_windowed'] \
         -s EXPORTED_RUNTIME_METHODS=['ccall','callMain']")
 
     set(EMSDK_FLAGS_LINKER "${EMSDK_FLAGS_LINKER} ${EMSDK_DEBUG_LINKER_FLAGS} --shell-file ${CMAKE_SOURCE_DIR}/Emscripten/launcher_index.html ")
--- a/Common/util/filestream.cpp
+++ b/Common/util/filestream.cpp
@@ -57,6 +57,13 @@ void FileStream::Close()
     }
     _file = nullptr;
     _ownHandle = false;
+    if (FileCloseNotify)
+    {
+        CloseNotifyArgs args;
+        args.Filepath = _fileName;
+        args.WorkMode = _workMode;
+        FileCloseNotify(args);
+    }
 }
 
 bool FileStream::Flush()
@@ -148,7 +155,10 @@ void FileStream::Open(const String &file_name, FileOpenMode open_mode, FileWorkM
     if (_file == nullptr)
         throw std::runtime_error("Error opening file.");
     _ownHandle = true;
+    _fileName = file_name;
 }
+
+FileStream::FFileCloseNotify FileStream::FileCloseNotify = nullptr;
 
 } // namespace Common
 } // namespace AGS
--- a/Common/util/filestream.h
+++ b/Common/util/filestream.h
@@ -19,6 +19,7 @@
 #define __AGS_CN_UTIL__FILESTREAM_H
 
 #include <stdio.h>
+#include <functional>
 
 #include "util/datastream.h"
 #include "util/file.h" // TODO: extract filestream mode constants
@@ -31,6 +32,16 @@ namespace Common
 class FileStream : public DataStream
 {
 public:
+    struct CloseNotifyArgs
+    {
+        String Filepath;
+        FileWorkMode WorkMode;
+    };
+
+    // definition of function called when file closes
+    typedef std::function<void(const CloseNotifyArgs &args)> FFileCloseNotify;
+
+    static FFileCloseNotify FileCloseNotify;
 
     // Represents an open file object
     // The constructor may raise std::runtime_error if 
@@ -81,6 +92,7 @@ private:
     bool                 _ownHandle;
     const FileOpenMode  _openMode;
     const FileWorkMode  _workMode;
+    String              _fileName;
 };
 
 } // namespace Common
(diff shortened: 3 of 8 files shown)


ivan-mogilko:
Question: do all file writing operations need the "syncing with fs", or strictly save game and config file?

If all of them need this, perhaps it would be better to implement a function call when closing a writeable File object instead.

ericoporto:
First, let me explain what FS.syncfs, this operation can be used to save all files that in directories with a mounted IDBFS to disk. While this happens, the browser tab does a slight hang, but the operation can be asynchronous in theory - we could have a worker that does this if we want to minimize this hang. We can't pass a specific file, so it always sync everything that can be synced.

So in theory, if we knew when the user would stop playing the game, but before closing, we could sync just there, instead of everytime, but when I tried this, this didn't work for all cases - say the browser is closed suddenly.

> If all the file writes need this, then it would be better to implement a function call when closing a writeable FileStream object instead. This will ensure that whenever engine writes and closes a file this action will be performed.

So there are two things here

- Adding a platform specific thing in the FileStream, is this ok?
- It would be nice to know we actually need to sync, knowing this is a write operation in a "permanent storage" location, but at the time of closing the file, but I couldn't quite figure this out.

> Question: do all file writing operations need the "syncing with fs", or strictly save game and config file?

I think most of the time people write to a config file some stuff and savegame, so it felt a way to smash at least two file syncs in one. Most of the time these save games happen in places (menus, quiet areas) where it's alright to hang the game a bit while it saves to disk.

> Why "post-savegame" and "post-config" actions should be separate platform functions?

Oh, I tried to follow the existing "Post" stuff, and it looks like Config and Savegame can be different directories. If it's acceptable they could be some single platform method.

ivan-mogilko:
> So in theory, if we knew when the user would stop playing the game, we could sync just there, but this doesn't work when the browser is closed suddenly.

Hmm, looks like I misunderstood. Then I'll ask a different question: what is the purpose of "syncing", or rather, which problems might occur if we do not sync?

If this operation is required, for example, to be able to open a file for reading after it's being written, then it has to be enforced after writing.

But, if this operation is moving files from some "temporary" location to a "permanent" one, then perhaps a worker thread with some "please sync now" signal call would work better.

The worker thread that syncs periodically will be useful, because AGS game may have custom saves, or other custom files that should be saved while user may not save the game itself.

ericoporto:
The virtual filesystem we use for AGS is in ram (MEMFS). We can FS.mount a IDBFS filesystem in any directories - currently the saved_games dir.

FS.syncfs is an operation that you can either populate from IndexedDB, or store back to IndexedDB (disk), files in mounted directories - it has a parameter "populate" that can be true or false. The operation we are discussing is writing to disk (populate=false).

> Hmm, looks like I misunderstood. Then I'll ask a different question: what is the purpose of "syncing", and which problems might occur if we do not sync?

Not syncing means the game doesn't store the files it changed back into disk.

> The worker thread that syncs periodically will be useful

There's a small catch that I couldn't yet figure how to do this without pthreads - which require an inconvenient header.

> so, to clarify more, there are files that should be in permanent storage, and those that should not; how is it decided which are which?

So only files in the `/home/web_user/saved_games` dir are stored in the IndexedDB permanent storage.

---

This is not the ONLY permanent storage in the browser, say we want to install the game and have it behave like an app, the game files (game.ags, speech.vox, ...) they have a different browser API they go through - I haven't implemented yet, but once this is created a new button will appear in the address bar for installation.

ivan-mogilko:
Hm, so, if I understand correctly, the platform code explicitly defines that certain *directory* is going to be "synced". In other words, it's not certain "objects" or files of certain purpose (save, config) that are synced, but files that are written to particular directory (user's save directory).

In such case, in my opinion, the new platform functions that are introduced here are going to be misleading, because they hide the intent: what important is not that "save" or "config" is made, but that something is written to the particular dir. There are  few issues with this:
* These files may in theory be written by the script File API (at least the config file, there's even a $CONFIGFILE$ tag introduced in recent version);
* This assumes that the save and config are always written to the "save dir" defined by the platform, which may not be the case (even if only hypothetically speaking).
* This ignores a situation when any custom user file is written to "save dir".
* I think (judging by the code in your pr), that it currently ignores any file or serialization errors; so it may ask for a sync when writing failed completely.

I think that in such circumstance a platform code should react to a signal telling that *something is written to the particular directory*. Then it could check the path and decide *for itself* whether this is a permanent location or not; because it's "platform" that knows these nuances, not the engine.

This brings us back to the previous suggestion about FileStream.

For example, FileStream could record the filepath (something that would be useful in any case, for debugging).
(Whether "what" was written is necessary to tell or not is a separate question, this may be also added to Stream class, or passed somehow else, but I think this may be postponed until actually needed.)
The problem would then be to invent how to pass the signal from the closing FileStream into the platform code. This is mostly a problem of organizing a code, since "platform driver" is currently in engine, while FileStream is the utility class in the Common lib.

The second part, since you mentioned sync/async issue, is making sure that this operation won't block the game too much.

Now, I don't know or don't remember what is this problem with threads in Emscripten (I believe it should be resolved the earlier the better, because it will complicate things every time we'll decide to utilize a thread for background work).

Unless there's a way to use some emscripten's library utility to run this action async, normally, I'd say, there should be a working thread started but sleeping most of the time, until there's a signal about written file closed, in which case it wakes up and run file sync. Or maybe two kinds of signals: one that only tells to write when possible (to avoid too many syncs often), and another that asks to do it immediately for important files. But I am theorizing now.

If we cannot use threads, neither run this async no matter what now, I might suggest to design a workaround in such way that would *pretend* that it runs async. This is to make our intent clear, and also prepare a "foundation" for the proper solution.
Perhaps it could be done like this: a FileStream tells what it writes on close, a platform decides if that should be synced to a persistent location, but does not sync immediately and increments some counter. Finally, there may be introduced a platform driver's method, which is explicitly meant to ask for immediate file syncing (might need to find a good name). This method will see if a counter > 0, and if it's then sync and reset the counter.

ericoporto:
> I don't know or don't remember what is this problem with threads in Emscripten

I need Coop+Coep headers for SharedBuffer Arrays ([see more here](https://web.dev/coop-coep/)), which are used to implement threads, but most free webhosts doesn't serve them. After [I started nagging](https://itch.io/t/1028526/cross-origin-policies-for-webassembly), there is now [some experimental support at least in itch](https://itch.io/t/2025776/experimental-sharedarraybuffer-support), but not in GH pages, or gamejolt or elsewhere.

> If we cannot use threads,

We can't - maybe in the future, but not right now.

For the asynchronous part, ignoring Emscripten, the issue is #1618 - that issue is a specialization of the Commit problem from Switch Homebrew I mentioned earlier which is similar to here, file operations happen in a fast but ram location and pushing to a permanent storage can hang a bit.

> The problem would then be to invent how to pass the signal from the closing FileStream into the platform code.

If there was say some default callback on Common that we could override in the Engine by a platform one, then I could use this on the platform driver.

It's good to see if this callback would work for your suggested implementation. #1349 would make implementing this really easy, I think right now such code goes in as some custom even in the SDL event queue?

> I think (judging by the code in your pr), that it currently ignores any file or serialization errors; so it may ask for a sync when writing failed completely.

The write of the FileStream is to RAM, and yes it can fail if we are low on RAM.

ivan-mogilko:
> If there was say some default callback on Common that we could override in the Engine by a platform one, then I could use this on the platform driver. (A hacky way would be to use the extern to link to something in Engine)

Yes, callback seem to be a better option than having an extern, because then we'd have to implement that extern function in every program that links the FileStream.

Let's say, if there were a static FileStream's member, a function pointer that accepts some args, or maybe 1 arg which is a struct with necessary fields (latter will be easier to expand when needed). And called in `FileStream::Close()`:
```
if (_fileCloseNotify)
{
    _fileCloseNotify(args);
}
```

I don't think this will add any noteable overhead, as closing a file is not something that happens alot and often.

I doubt that SDL event is a way to go here though, as theoretically a reaction to this callback may have to be instantaneous.

ericoporto:
Hey, I need to just check how to organize things and fix building in other platforms, but it appears to work well. I added the build in the [agsjs](https://ericoporto.github.io/agsjs/)

ericoporto:
Hey, it appears to work well. I added the build in the [agsjs](https://ericoporto.github.io/agsjs/).

I haven't yet rebased to replace the previous commits with the new way as I am curious if this implementation is acceptable.

ericoporto:
Hey, it appears to work well. I added the build in the [agsjs](https://ericoporto.github.io/agsjs/). (did a quick update there now)

I ended going for `std::function` because I wanted to capture the context of the current platform driver in the lambda I used. In this way I used less globals as possible and kept the methods in the Emscripten platform driver.

I reorganized the commits a bit. I erased from the commits of this PR the original syncfs attempt using the PostSaveGame and PostSaveConfig.

I placed the old commits I did here, in case you need them to help review: https://github.com/ericoporto/ags/tree/add-emscripten-improvements-old

ivan-mogilko:
Yes, this will work.

PR #1931, opened 2023-02-24
Emscripten: fix launcher keep gear btn visible when windowed

ericoporto:
Minor fix to guarantee the gear icon ⚙️ is always visible when the screen is "windowed".


ericoporto:
Actually, I think I need to try one more thing before this.

PR #1934, opened 2023-02-26
Editor: Native use %zd compatible with size_t

ericoporto:
When I was building the Editor for 64-bit, build errors appeared in VS on these lines. Looking into wikipedia I found this

> z | For integer types, causes printf to expect a size_t-sized integer argument.
> -- | --

</body>
</html>

https://en.wikipedia.org/wiki/Printf_format_string

The fix is still compatible with win32.

### DIFF
--- a/Editor/AGS.Native/agsnative.cpp
+++ b/Editor/AGS.Native/agsnative.cpp
@@ -3608,7 +3608,7 @@ void convert_room_from_native(const RoomStruct &rs, Room ^room, System::Text::En
 		if (rs.DataVersion < kRoomVersion_300a)
 		{
 			char scriptFuncPrefix[100];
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I think it should be `%zu` for `size_t`:
https://en.cppreference.com/w/c/io/fprintf

According to the table, `%zd` is for the "signed size_t" (I read somewhere else that there may be specified a special `ssize_t` type, which stands for "signed size_t"), but `%zu` is for "unsigned size_t" (which is a regular `size_t`).

We use strictly `%zu` in the engine, afaik.

ericoporto:
Ah, that is a good point, I didn't notice this. Thanks for the reference from cppreference, that is something always useful to have around.

PR #1935, opened 2023-02-27
Engine: add plugin stub for agsappopenurl

ericoporto:
I know it's a bit soon, but the idea of this plugin is to be able to remove it when no external url loading is desired. Unfortunately the way for this is stubs.

My hope is to switch out people from `agsshell` as I hope to reduce surface of problems through this url specific plugin. I know this is a bit soon, but I have got an increased number of asks about agsshell in the last days.

### DIFF
--- a/Engine/plugin/global_plugin.cpp
+++ b/Engine/plugin/global_plugin.cpp
@@ -70,6 +70,12 @@ bool RegisterPluginStubs(const char* name)
     ccAddExternalStaticFunction("ShellExecute",                 Sc_PluginStub_Void);
     return true;
   }
+  else if (ags_stricmp(name, "agsappopenurl") == 0)
+  {
+    // agsappopenurl.dll
+    ccAddExternalStaticFunction("AppOpenURL",                 Sc_PluginStub_Int0);
+   return true;
+  }
   else if (ags_stricmp(name, "ags_snowrain") == 0)
   {
     // ags_snowrain.dll


ivan-mogilko:
Okay, we may add this. But we really need to investigate options that would help us to avoid hardcoding anything plugin-related in the engine, because this makes engine somewhat subservient to plugins that other people write.

ericoporto:
The stubs are using for linking.

The problem is the plugin header once used by the game, the functions are used, but there's no distinction of what functions used are scripted or from a plugin/which plugin.

So after the plugin is requested, if the plugin is not available, linking fails at runtime, before actually using the functions. If instead, the plugin wrote in the game it's own failsafe stubs, then at least the linking wouldn't fail and the game can test for the plugin, because we could have some device that if the plugin fails to be found we switch to reading some information that would help me build the stubs.

The other alternative is adding some attribute automatically to all functions and things from a plugin that would cause linking to delay at first usage or something - unfortunately I think this would still cause runtime linking errors when stubs won't.

ivan-mogilko:
> The problem is the plugin header once used by the game, the functions are used, but there's no distinction of what functions used are scripted or from a plugin/which plugin.

By the way, this is theoretically possible to do, if compiler or preprocessor would add this info, as it's known where the headers are coming from.

ericoporto:
Erh, now I was thinking, the stubs could even be a script module, built to a .o binary, and then we only load it if the plugin failed to be found?

By using the standalone compiler this could be done?

ivan-mogilko:
> Erh, now I was thinking, the stubs could even be a script module, built to a .o binary, and then we only load it if the plugin failed to be found?
> 
> By using the standalone compiler this could be done?

Perhaps theoretically; right now engine does not let to link unknown modules, only ones which names are written into the game data, so you also have to provide a "ScriptModules.lst" file.
<https://github.com/adventuregamestudio/ags/blob/master/Engine/main/game_file.cpp#L135>

ivan-mogilko:
> Erh, now I was thinking, the stubs could even be a script module, built to a .o binary, and then we only load it if the plugin failed to be found?
> 
> By using the standalone compiler this could be done?

I think this may be done by any compiler, from inside the Editor too.

PR #1937, opened 2023-02-28
3.6.0: separate absolute/relative switch for Touch-to-mouse emulation

### LINKED ISSUE #1930: Different uses of Relative vs Direct in Touch vs Mouse input
Currently (AGS 3.6.0.43 8dce503b05728edabe6205e7fbf3f660ed31e957), we can set the cursor movement to be direct or relative using `[mouse] control_enabled=0` or `1` (it also requires `[mouse] control_when=always` and defaults to `[mouse] control_when=fullscreen`). 

- When using **direct** and **touch**, the input is _usually_ for a game with hidden cursor, where the user tap directly where it's intended in the screen, in a game that has been properly designed for this - bigger hotspot areas and such to take into account finger size.
- When using **relative** and **touch**, the input is _usually_ for a game with visible cursor, where the user controls this cursor and keeps it near but not under the finger, so it can be used with more precision, either because the mobile port was rushed or because a vocal part of the userbases asks for this mode. Usually this works better also with `[touch] emulate_mouse=2`
- Using **direct** and **mouse**, is the expected default for most AGS games.
- When using **relative** and **mouse**, this is _usually_ intended for some infinite mouse implementation, that can be useful for FPS-like games and other non-point and click adventure like games. This also can be used simply for a game controlled mouse speed.

The platforms available are mostly used with one type of input device - most desktop platforms (Windows, Linux, macOS) are used with Mouse input and most mobile platforms (Android, iOS) are used with Touch input. But the Web platform exists in both desktop and mobile platforms, so you can use a game built for web with either mouse or touch input.

This is more a consequence of the existing logic and how it manifests on the two different types of input, not sure what can or should be done about this, but it looks like a game can be intended to use `relative` cursor control in touch input and `direct` cursor control in mouse input.

**suggested change**

I would suggest ignore `control_enabled` and `control_when` for input _originated_ from touch and use a new `[touch] emulate_mouse_relative=0` or `1` to decide if that would be direct or not. Not sure how possible this is in code though and also what this means for `Mouse.ControlEnabled` in the game.

ivan-mogilko:
Resolves #1930.

Support interpreting mouse motion events as absolute or relative, individually per "mouse" device (real or emulated). The inner engine code that assigns a game cursor position no longer relies on config setting, but instead relies on the accumulated mouse position, provided by the event handling unit. Relative motion (movement delta) is currently unused but kept in the code, as it may become useful later for certain purposes.

Event handling assumes that the received mouse position and deltas are generated with all the settings already applied by the sending device (absolute/relative mode, speed, and so forth).

Touch2mouse code is adjusted to have its own absolute/relative option, and calculate new position in the relative mode correspondingly.

Added a separate config setting for touch controls: `[touch] motion_relative` (name may be adjusted, if it looks bad). This setting is applied to event handler "component", which affects how the touch-to-mouse emulated device is handled.
Changed mobile code to write this setting, instead of `[mouse] control_enabled`, which is really unrelated to this.

Additionally (although this was not required) changed how `[touch] emulate_mouse` is written: now it has a string value instead of a number (i'm not sure why I left it like this the last time): "off", "one_finger", "two_fingers".

### DIFF
--- a/Common/util/geometry.h
+++ b/Common/util/geometry.h
@@ -93,6 +93,12 @@ struct Point
         Y = y;
     }
 
+    inline static Point Clamp(const Point &p, const Point &floor, const Point &ceil)
+    {
+        return Point(AGSMath::Clamp(p.X, floor.X, ceil.X),
+                    AGSMath::Clamp(p.Y, floor.Y, ceil.Y));
+    }
+
     inline bool operator ==(const Point &p) const
     {
         return X == p.X && Y == p.Y;
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1639,8 +1639,12 @@ namespace AGS.Editor
             NativeProxy.WritePrivateProfileString("mouse", "auto_lock", _game.DefaultSetup.AutoLockMouse ? "1" : "0", configFilePath);
             NativeProxy.WritePrivateProfileString("mouse", "speed", _game.DefaultSetup.MouseSpeed.ToString(CultureInfo.InvariantCulture), configFilePath);
 
-            int emulate_mouse = (int) _game.DefaultSetup.TouchToMouseEmulation;
-            NativeProxy.WritePrivateProfileString("touch", "emulate_mouse", emulate_mouse.ToString(), configFilePath);
+            // Touch input
+            string[] emulate_mouse_str = new string[] { "off", "one_finger", "two_fingers" };
+            NativeProxy.WritePrivateProfileString("touch", "emul_mouse_mode",
+                emulate_mouse_str[(int)_game.DefaultSetup.TouchToMouseEmulation], configFilePath);
+            NativeProxy.WritePrivateProfileString("touch", "emul_mouse_relative",
+                ((int)_game.DefaultSetup.TouchToMouseMotionMode).ToString(), configFilePath);
 
             // Note: sprite cache size is written in KB (while we have it in MB on the editor pane)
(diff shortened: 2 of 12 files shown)


ericoporto:
`[touch] motion_relative` ... Uhm perhaps `[touch] relative_cursor` ? I thought about `[touch] relative_mouse_emulation`, but that is too big.

ivan-mogilko:
Hmm, after some thought, I changed to "emul_mouse_mode" and "emul_mouse_relative" correspondingly...

ericoporto:
hey, the relative mode, if the mouse cursor moves beyond the borders, on mobile, using touch input, it clamps the cursor to the borders, but it doesn't clamp any internal x,y position of the cursor, so if you move it beyond the border and keep pushing it, then you need to push it back the same amount to be able to snap it out of the border.

ivan-mogilko:
> hey, the relative mode, if the mouse cursor moves beyond the borders, on mobile, using touch input, it clamps the cursor to the borders, but it doesn't clamp any internal x,y position of the cursor, so if you move it beyond the border and keep pushing it, then you need to push it back the same amount to be able to snap it out of the border.

Oh right, I noticed that too, but forgot to fix.

ivan-mogilko:
Pushed a fix for clamping to the window size, also sync to mouse coords in case
* there are mouse bounds enabled by a script;
* there's a real mouse device acting in parallel.

ericoporto:
The relative mouse mode is now fixed, I tested in both Android and Web builds on a phone. I also tested on the desktop and it looks like it kept the direct mode there, as intended, so great work!

PR #1944, opened 2023-03-08
AGS4 Flipped frames animation for gui buttons

AlanDrake:
Enables animations with flipped frames to work properly on GUI buttons.

### DIFF
--- a/Common/gui/guibutton.cpp
+++ b/Common/gui/guibutton.cpp
@@ -69,6 +69,7 @@ GUIButton::GUIButton()
 
     IsPushed = false;
     IsMouseOver = false;
+    IsFlipped = false;
     _placeholder = kButtonPlace_None;
     _unnamed = true;
 
@@ -367,8 +368,9 @@ void GUIButton::DrawImageButton(Bitmap *ds, int x, int y, bool draw_disabled)
     // NOTE: the CLIP flag only clips the image, not the text
     if (IsClippingImage() && !GUI::Options.ClipControls)
         ds->SetClip(RectWH(x, y, Width, Height));
+
     if (spriteset[CurrentImage] != nullptr)
-        draw_gui_sprite(ds, CurrentImage, x, y, true);
+        draw_gui_sprite_flipped(ds, CurrentImage, x, y, true, kBlend_Normal, IsFlipped);
 
     // Draw active inventory item
     if (_placeholder != kButtonPlace_None && gui_inv_pic >= 0)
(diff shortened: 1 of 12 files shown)


Engine/ac/draw.cpp
```diff
@@ -1087,6 +1087,34 @@ void draw_gui_sprite(Bitmap *ds, bool use_alpha, int x, int y, Bitmap *sprite, b
...
+}
+
+void draw_gui_sprite_flipped(Bitmap *ds, bool use_alpha, int x, int y, Bitmap *sprite, bool src_has_alpha,
+    BlendMode blend_mode, int alpha, bool is_flipped)
+{
+    if (alpha <= 0)
+        return;
+
+    if (is_flipped) {
+        Bitmap *tempspr = BitmapHelper::CreateTransparentBitmap(sprite->GetWidth(), sprite->GetHeight(), sprite->GetColorDepth());
```
ivan-mogilko, line 1103: You do not delete tempspr below, causing a memory leak, I propose using `std::unique_ptr<Bitmap>` to make this automatic.

Common/gui/guibutton.cpp
```diff
@@ -58,7 +58,7 @@ GUIButton::GUIButton()
     Image = -1;
     MouseOverImage = -1;
     PushedImage = -1;
-    CurrentImage = -1;
+    _currentImage = -1;
```
ivan-mogilko, line 61: Please make _currentImage and flipped private/protected to enforce that they cannot be set without each other, but only using SetCurrentImage method.

Common/gui/guibutton.cpp
```diff
@@ -69,7 +69,7 @@ GUIButton::GUIButton()
 
     IsPushed = false;
     IsMouseOver = false;
-    IsFlipped = false;
+    IsImageFlipped = false;
```
ivan-mogilko, line 72: My suggestion would actually to make "ImageFlags" here, to ease any possible expansion (and expansion of the save format), since flip may have only limited number of options (H, V, H+V). There are already Flip flags somewhere in the engine (I think you could use ones that Bitmap class uses too).

Engine/ac/draw.cpp
AlanDrake: I admit I'm always at loss on how to use these internal APIs

ivan-mogilko: > I admit I'm always at loss on how to use these internal APIs

Well, I must add more comments to these functions.

AlanDrake:
Alright, I applied the suggested corrections.
I left draw_gui_sprite_flipped to only check for the horizontal flip flag, no point overplanning until we actually decide to support vertical flip as well.

Engine/ac/draw.cpp
AlanDrake: I just found out this solution can cause garbage data. 
Maybe the unique_ptr assignment `sprite = tempspr.get();` , which triggers a delete, is causing this?

PR #1948, opened 2023-03-13
3.6.0: an attempt to fix a window centering

### LINKED ISSUE #1895: 3.6.0: on Windows, window placement disregards the task bar
**Problem**

In the previous versions, at least on Windows, AGS took taskbar into account, and clamped the maximal window size to the remaining space. 3.6.0 no longer seem to do this. With the "stretch to fit screen" the window is created too large, so it may partially hide behind a taskbar. Even if window's size is enough to fit the free space on desktop (e.g. with "max round multiplier" scaling), it's positioned without accounting for a taskbar, so I have to move it aside if I want to see the whole game screen.

**Expected behavior**

The game window should account task bar into account, clamp the maximal size, and position itself so that it is not covered by taskbar on start.

ivan-mogilko:
Fixes #1895

This is an attempt to fix a window not centering correctly on Windows; also may improve situation on other desktop systems too, although there may be caveats that this does not address yet (but to be fair they were not addressed by the previous versions either).

1. Adds a generic version for ValidateWindowSize that uses SDL's API to know the usable desktop space. Unfortunately, SDL does not allow to query future window's border size without having window created first, so I omit that for now.
2. Provides a workaround for SDL2's mistakes in centering the created window on screen; see comments to the commit and the code; also see https://github.com/libsdl-org/SDL/issues/6875


ericoporto:
I think we should only use `sys_window_center` if the window is "windowed".

ivan-mogilko:
Damn, because my commit's description has a link to libsdl's ticket, they get a new notification every time I make a force-push... I hope they won't get annoyed by this :).

ericoporto:
Sorry for commenting more on the matter, but 

- I would remove the comment outside of sys_window_center and let it only inside of it, 
- In the code of sys_window_center, make it be the default behavior of `SDL_SetWindowPosition(window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);` or do nothing (just a `return;`), in mobile+Emscripten (either is fine) and then use the other behavior in the desktop systems (Windows, Linux, macOS and FreeBSD).

In this way there is no need for the `#if` macro in the outside call to sys_window_center, since it's called twice.

ericoporto:
Looks good for me

ivan-mogilko:
Got an answer from user who noticed this first; he also confirms this works for him.

Issue #1949, opened 2023-03-13
Going from Windowed to Fullscreen hangs in OpenGL driver with VSync


ericoporto:
**Describe the bug**
If the game is using OpenGL video driver, when going from Windowed to Fullscreen, if VSync is set, it hangs.

I have only tested on Linux, haven't tried on Windows or different linuxes.

**AGS Version**
3.6.0.44 and since this https://github.com/adventuregamestudio/ags/commit/bcf90697ac93080d0edac3e99727041a4e2b461a

**Game**
The game I tried had vsync set.

**To Reproduce**
1. Start the game in Windowed mode
2. Press Alt+Enter to go into Fullscreen
3. The game should hang

**Expected behavior**
It should just go into fullscreen and the game should proceed

**Desktop:**
 - OS: Ubuntu
 - Version 16.04 (it's old, I know)

**Additional context**
If I make the support check return false, the game no longer hangs

https://github.com/adventuregamestudio/ags/blob/f19043d902ff9a7e7d6478f51093ef3c807dd258/Engine/gfx/ali3dogl.h#L230

I tried to redo SetVsync like this 

```Cpp
bool OGLGraphicsDriver::SetVsync(bool enabled)
{
    if (_mode.Vsync == enabled)
    {
        return _mode.Vsync;
    }

    if (SDL_GL_SetSwapInterval(enabled) == 0)
    {
        _mode.Vsync = enabled;
    }

    return _mode.Vsync;
}
```

But the game still hanged.

I saw some weird comments on `SDL_GL_SetSwapInterval` from icculus, so now I am a bit suspicious of it - and I think we call it on everyframe, which not sure is a good idea.

- https://github.com/libsdl-org/SDL/issues/4471#issuecomment-886069233
- https://github.com/libsdl-org/SDL/issues/5122#issuecomment-1010569143

<details>
<summary>$ glxinfo | grep version</summary>
server glx version string: 1.4 <br>
client glx version string: 1.4 <br>
GLX version: 1.4 <br>
    Max core profile version: 4.5 <br>
    Max compat profile version: 3.0 <br>
    Max GLES1 profile version: 1.1 <br>
    Max GLES[23] profile version: 3.2 <br>
OpenGL core profile version string: 4.5 (Core Profile) Mesa 18.0.5 <br>
OpenGL core profile shading language version string: 4.50 <br>
OpenGL version string: 3.0 Mesa 18.0.5 <br>
OpenGL shading language version string: 1.30 <br>
[...]

PR #1950, opened 2023-03-13
3.6.0: improve how vsync is changed at runtime

ivan-mogilko:
This is a reaction to #1949, but I cannot tell whether this is enough to fix it at the moment.

This does two things:
1. Small refactor that ensures that VSync is not necessarily reset when using a gfxdriver's interface if it's already set to the given value. This is to reduce code duplication and unify behavior a little.
2. Gfx driver will try to detect when vsync is not supported, and also remember when changing vsync at runtime fails. In this case it won't try to set it again, until the display mode changes.

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -932,7 +932,10 @@ void render_to_screen()
     // Stage: engine overlay
     construct_engine_overlay();
 
-    gfxDriver->SetVsync(scsystem.vsync > 0);
+    // Try set new vsync value, and remember the actual result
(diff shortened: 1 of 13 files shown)


ivan-mogilko:
Testing this out on my ubuntu vm, I found that software renderer has a weird issue, when switching the vsync at runtime, the SDL's renderer reports that vsync is still on, while in reality it's off (judging by the fps counter).

ivan-mogilko:
Well, this does not fix #1949, but imo these changes will be useful.

Also, added change 3) save real vsync value after trying to apply to gfxdriver, in case it fails. The only caveat here is that it happens not instantly, but only at the next draw pass. Might add this note to the manual.

PR #1953, opened 2023-03-18
Fix Windows Exception handling in 64-bit builds

ericoporto:
- size_t requires %zu in format (see #1934)
- on 64-bit , ExceptionAddress, an intptr_t requires a different format string to be readable
  - I looked the PRIxPTR from inttypes.h, but I think this custom format prints nicely as a counterpart of 0x%08X
  - we use 16 zeroes as the fixed length, because of 64 bits (from 8 for 32-bit variant) ((inttypes.h has nothing for this)
  - I64 is windows for unsigned 64-bit int (I think I could also use `ll` here)
  - X is for the hexadecimal, same as the 32-bit variant

### DIFF
--- a/Engine/platform/windows/win_ex_handling.cpp
+++ b/Engine/platform/windows/win_ex_handling.cpp
@@ -15,6 +15,7 @@
 
 #if AGS_PLATFORM_OS_WINDOWS
 #include <new.h>
+#include <cinttypes>
 #include <stdio.h>
 #include "ac/common.h" // quit
 #include "ac/common_defines.h"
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
> we use 16 zeroes as the fixed length, because of 64 bits (from 8 for 32-bit variant) ((inttypes.h has nothing for this)

I recalled that there should be a way to pass field width (number of zeroes in this case) as another argument.
From https://en.cppreference.com/w/c/io/fprintf:

> integer value or * that specifies minimum field width. The result is padded with space characters (by default), if required, on the left when right-justified, or on the right if left-justified. In the case when * is used, the width is specified by an additional argument of type int, which appears before the argument to be converted....

Example:
```
intptr_t value = 0xABCD;
printf("%0*X", (int)sizeof(intptr_t) * 2, value);
```
(because with hex you need 2 digits per byte; conversion to (int)needed to prevent a warning unfortunately)

So I guess something like this would work?
```
"An exception 0x%X occurred in ACWIN.EXE at EIP = 0x%0*" PRIXPTR "; ...the rest of string...",
       excinfo.ExceptionCode, (int)sizeof(intptr_t) * 2, (intptr_t)excinfo.ExceptionAddress, ...rest of args
```

ericoporto:
I did the change you suggested, but better wait the CI build this before anything

PR #1955, opened 2023-03-19
Engine: fix touch events not working in iOS Safari

### LINKED ISSUE #1954: Web Port in iOS Safari doesn't register clicks
**Describe the bug**
When running our Web port in the iOS Safari browser, clicks are not registered

The reason it appears to be that [Touch.identifier](https://developer.mozilla.org/en-US/docs/Web/API/Touch/identifier) works differently in it.

Here is a log after I added three log functions in sys_events.cpp, this is in a simulator, so the mouse is only one finger, I click, move and release I think six times and get this.

```
[Log] 20:31:36:062: on_sdl_touch_down: fingerId 998779017, xy (0.611224, 0.570836), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:31:36:096: on_sdl_touch_motion: fingerId 998779017, xy (0.611224, 0.570243), dxdy (0.000000, -0.000593); (tst, line 583)
[Log] 20:31:36:151: on_sdl_touch_up: fingerId 998779017, xy (0.611224, 0.570243), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:31:39:831: on_sdl_touch_down: fingerId 998779018, xy (0.586735, 0.581506), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:31:39:895: on_sdl_touch_motion: fingerId 998779018, xy (0.586735, 0.580913), dxdy (0.000000, -0.000593); (tst, line 583)
[Log] 20:31:39:948: on_sdl_touch_up: fingerId 998779018, xy (0.586735, 0.580913), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:31:51:444: on_sdl_touch_down: fingerId 998779019, xy (0.518367, 0.548903), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:31:51:485: on_sdl_touch_motion: fingerId 998779019, xy (0.518367, 0.548311), dxdy (0.000000, -0.000593); (tst, line 583)
[Log] 20:31:51:485: on_sdl_touch_up: fingerId 998779019, xy (0.518367, 0.548311), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:32:49:277: on_sdl_touch_down: fingerId 998779020, xy (0.271429, 0.595732), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:32:49:278: on_sdl_touch_motion: fingerId 998779020, xy (0.271429, 0.594547), dxdy (0.000000, -0.001186); (tst, line 583)
[Log] 20:32:49:278: on_sdl_touch_motion: fingerId 998779020, xy (0.275510, 0.594547), dxdy (0.004082, 0.000000); (tst, line 583)
[Log] 20:32:49:321: on_sdl_touch_motion: fingerId 998779020, xy (0.303061, 0.594547), dxdy (0.027551, 0.000000); (tst, line 583)
[Log] 20:32:49:323: on_sdl_touch_motion: fingerId 998779020, xy (0.488776, 0.598103), dxdy (0.185714, 0.003557); (tst, line 583)
[Log] 20:32:49:389: on_sdl_touch_motion: fingerId 998779020, xy (0.518367, 0.597510), dxdy (0.029592, -0.000593); (tst, line 583)
[Log] 20:32:49:389: on_sdl_touch_motion: fingerId 998779020, xy (0.544898, 0.596918), dxdy (0.026531, -0.000593); (tst, line 583)
[Log] 20:32:49:439: on_sdl_touch_motion: fingerId 998779020, xy (0.550000, 0.596918), dxdy (0.005102, 0.000000); (tst, line 583)
[Log] 20:32:49:822: on_sdl_touch_up: fingerId 998779020, xy (0.550000, 0.596918), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:32:54:715: on_sdl_touch_down: fingerId 998779021, xy (0.225510, 0.544754), dxdy (0.000000, 0.000000); (tst, line 583)
[Log] 20:32:54:773: on_sdl_touch_motion: fingerId 998779021, xy (0.225510, 0.543568), dxdy (0.000000, -0.001186); (tst, line 583)
[Log] 20:32:54:871: on_sdl_touch_motion: fingerId 998779021, xy (0.225510, 0.542976), dxdy (0.000000, -0.000593); (tst, line 583)
[Log] 20:32:55:072: on_sdl_touch_motion: fingerId 998779021, xy (0.253061, 0.553646), dxdy (0.027551, 0.010670); (tst, line 583)
[Log] 20:32:55:073: on_sdl_touch_motion: fingerId 998779021, xy (0.277551, 0.561944), dxdy (0.024490, 0.008299); (tst, line 583)
[Log] 20:32:55:126: on_sdl_touch_motion: fingerId 998779021, xy (0.318367, 0.573800), dxdy (0.040816, 0.011855); (tst, line 583)
[Log] 20:32:55:126: on_sdl_touch_motion: fingerId 998779021, xy (0.371429, 0.587433), dxdy (0.053061, 0.013634); (tst, line 583)
[Log] 20:32:55:169: on_sdl_touch_motion: fingerId 998779021, xy (0.404082, 0.593954), dxdy (0.032653, 0.006520); (tst, line 583)
[Log] 20:32:55:170: on_sdl_touch_motion: fingerId 998779021, xy (0.441837, 0.602845), dxdy (0.037755, 0.008892); (tst, line 583)
[Log] 20:32:55:253: on_sdl_touch_motion: fingerId 998779021, xy (0.468367, 0.606995), dxdy (0.026531, 0.004149); (tst, line 583)
[Log] 20:32:55:254: on_sdl_touch_motion: fingerId 998779021, xy (0.512245, 0.611144), dxdy (0.043878, 0.004149); (tst, line 583)
[Log] 20:32:55:307: on_sdl_touch_motion: fingerId 998779021, xy (0.525510, 0.611737), dxdy (0.013265, 0.000593); (tst, line 583)
[Log] 20:32:55:307: on_sdl_touch_motion: fingerId 998779021, xy (0.534694, 0.611737), dxdy (0.009184, 0.000000); (tst, line 583)
[Log] 20:32:55:580: on_sdl_touch_up: fingerId 998779021, xy (0.534694, 0.611737), dxdy (0.000000, 0.000000); (tst, line 583)
```

**AGS Version**
it happens on current 3.6.0.45, but curiously it did not happen in an old version.

**Game**
https://ericoporto.github.io/agsjs/tst/

**To Reproduce**
1. Load the URL and see that touching the screen does nothing.
2. If you have Safari Technology Preview installed on a macOS you can connect to the iOS system with Safari
3. the log will have the mentioned on_sdl_touch log messages

**Expected behavior**
It's expected to work

**Smartphone:**
 - Device: iOS Simulator with iPhone 14 Pro Max
 - OS: iOS
 - Version 16.2

**Additional context**
When SDL own touch emulation was used, there was no reliance in fingerId, so it worked with that.

The API from SDL only states that fingerId is unique, and nothing else, so apparently we can't rely on actual values from it - even though on Android it's consistent with both the App and the Web version on different browsers, it appears on other devices this is not guaranteed.

ericoporto:
fix #1954

FingerContainer is probably not a good name, and I guess it could be more efficient, but I don't expect the number of fingers to be more than one digit ever, so I think this should work fast enough.

### DIFF
--- a/Engine/ac/sys_events.cpp
+++ b/Engine/ac/sys_events.cpp
@@ -498,9 +498,67 @@ int ags_check_mouse_wheel() {
 // TOUCH INPUT
 // ----------------------------------------------------------------------------
 
+// SDL_FingerIDs are unique for down-motion-up events, but beyond this there is no cross-platform guarantee
+// this struct tracks currently active fingers on the touch screen, and matches a slot index for each
+// we will use this to implement the logic in this example
+// - first finger down (consider LMB, index 0);
+// - second finger down (consider RMB, index 1);
+// - first finger up (consider LMB up);
+// - second finger is still pressed and still considered RMB;
+// - third(?) finger down, takes the free role of LMB (index 0).
+struct Fingers
+{
+public:
+    static const int MAX_FINGERS = 2;
+    static const int NO_INDEX = -1;
+
+    // store fingerId, return given finger index
+    int push(SDL_FingerID fingerId)
+    {
+        if (contains(fingerId))
+            return NO_INDEX; // invalid, fingerId already present
+
+        auto it = std::find(_fingers.begin(), _fingers.end(), NO_ID);
+        if(it == _fingers.end())
+            return NO_INDEX; // no slot for new finger
+
+        *it = fingerId;
+        return it - _fingers.begin();
+    };
+
+    int get_index(SDL_FingerID fingerId) const
+    {
+        auto it = std::find(_fingers.begin(), _fingers.end(), fingerId);
+        if(it != _fingers.end())
+            return it - _fingers.begin();
+
+        return NO_INDEX;
+    };
+
+    void pop(SDL_FingerID fingerId)
+    {
+        int idx = get_index(fingerId);
+        assert(idx != NO_INDEX);
+        if (idx != NO_INDEX) {
+            _fingers[idx] = NO_ID;
+        }
+    };
+
+private:
+    const SDL_FingerID NO_ID = -1; // std::find reads by reference, can't be static
+    std::array<SDL_FingerID, MAX_FINGERS> _fingers{{NO_ID, NO_ID}};
+
+    bool contains(SDL_FingerID fingerId) const
+    {
+        return std::find(_fingers.begin(), _fingers.end(), fingerId) != _fingers.end();
+    }
+};
+
 // Touch input state
 struct TouchState
 {
+    // on-screen fingers, used to remember the finger's role until all of them are released
+    Fingers fingers;
     // Accumulated finger bits (a collection of bits shifted by finger index)
     int fingers_down = 0;
     // Double tap detection
@@ -560,9 +618,9 @@ void ags_touch_set_mouse_emulation(TouchMouseEmulation mode,
 }
 
 // Converts touch finger index to the emulated mouse button
-static int tfinger_to_mouse_but(int finger)
+static int tfinger_to_mouse_but(int finger_index)
 {
-    switch (finger)
+    switch (finger_index)
     {
     case 0: return SDL_BUTTON_LEFT;
     case 1: return SDL_BUTTON_RIGHT;
@@ -678,7 +736,10 @@ static void sync_sys_mouse_pos()
 
 static void on_sdl_touch_down(const SDL_TouchFingerEvent &event)
 {
-    touch.fingers_down |= 1 << event.fingerId;
+    int finger_index = touch.fingers.push(event.fingerId);
+    if(finger_index == Fingers::NO_INDEX) return;
+
+    touch.fingers_down |= 1 << finger_index;
     detect_double_tap(event, true);
 
     switch (t2m.mode)
@@ -686,7 +747,7 @@ static void on_sdl_touch_down(const SDL_TouchFingerEvent &event)
     case kTouchMouse_OneFingerDrag:
     {
         // Touch down means LMB down
-        int mouse_but = tfinger_to_mouse_but(event.fingerId);
+        int mouse_but = tfinger_to_mouse_but(finger_index);
         if (mouse_but == SDL_BUTTON_LEFT)
         {
             set_t2m_pos(event.x, event.y, 0.f, 0.f);
@@ -698,7 +759,7 @@ static void on_sdl_touch_down(const SDL_TouchFingerEvent &event)
     }
     case kTouchMouse_TwoFingersTap:
     {
-        int mouse_but = tfinger_to_mouse_but(event.fingerId);
+        int mouse_but = tfinger_to_mouse_but(finger_index);
         // Handle double tap for drag-n-drop movement
         if ((mouse_but == SDL_BUTTON_LEFT) && (touch.tap_count == 2))
         {
@@ -739,7 +800,10 @@ static void on_sdl_touch_down(const SDL_TouchFingerEvent &event)
 
 static void on_sdl_touch_up(const SDL_TouchFingerEvent &event)
 {
-    touch.fingers_down &= ~(1 << event.fingerId);
+    int finger_index = touch.fingers.get_index(event.fingerId);
+    if(finger_index == Fingers::NO_INDEX) return;
+
+    touch.fingers_down &= ~(1 << finger_index);
     detect_double_tap(event, false);
 
     // TODO: better way to get SDL's logical size? we cannot access sdl renderer here
@@ -751,7 +815,7 @@ static void on_sdl_touch_up(const SDL_TouchFingerEvent &event)
     case kTouchMouse_OneFingerDrag:
     {
         // Touch up means LMB up
-        int mouse_but = tfinger_to_mouse_but(event.fingerId);
+        int mouse_but = tfinger_to_mouse_but(finger_index);
         if (mouse_but == SDL_BUTTON_LEFT)
         {
             send_mouse_button_event(SDL_MOUSEBUTTONUP, mouse_but, t2m.emul_pos.X, t2m.emul_pos.Y);
@@ -762,7 +826,7 @@ static void on_sdl_touch_up(const SDL_TouchFingerEvent &event)
     }
     case kTouchMouse_TwoFingersTap:
     {
-        int mouse_but = tfinger_to_mouse_but(event.fingerId);
+        int mouse_but = tfinger_to_mouse_but(finger_index);
         // If there's a force button set, then click only if the current button matches it;
         // otherwise use whatever was released
         if ((mouse_but > 0) &&
@@ -794,10 +858,14 @@ static void on_sdl_touch_up(const SDL_TouchFingerEvent &event)
     }
     default: break; // do nothing
     }
+    touch.fingers.pop(event.fingerId);
 }
 
 static void on_sdl_touch_motion(const SDL_TouchFingerEvent &event)
 {
+    int finger_index = touch.fingers.get_index(event.fingerId);
+    if(finger_index == Fingers::NO_INDEX) return;
+
     switch (t2m.mode)
     {
     case kTouchMouse_OneFingerDrag:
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
> but I don't expect the number of fingers to be more than one digit ever

Actually, for this reason, I wanted to suggest to not use any containers, but use a plain array of fixed size.
In the current implementation, having std::map looks redundant, as you are searching it as a plain array/vector anyway, and use `find` only in one case.

Other than that, I'd like to ask to not sequence all the commands in one line like this:
```
for(; idx < _fingers.size(); idx++) if(_fingers.count(idx) == 0) break;
```
Imo that makes it difficult to read; and contradicts to our code style in general.
```
for (; idx < _fingers.size(); idx++)
    if (_fingers.count(idx) == 0)
        break;
```

ericoporto:
> The size is the number of simultaneous touches that we care about (two currently)

I can do 20, if I do 2 the logic will crumble once a person puts a third finger in the screen, people have ten fingers, 20 seems safe.

ivan-mogilko:
> if I do 2 the logic will crumble once a person puts a third finger in the screen

Could you explain? Maybe I cannot imagine the problem myself.

ericoporto:
I am not sure I understand how that fingers_down should work, but I thought it was like a bitfield where the first finger is first bit, second finger is second bit and so on, so if I place an "invalid" finger, I don't know what to return there, and also I need to now ignore this invalid finger in all the other two cases (motion/up). If I simply support more fingers (up to 31), things kinda just work.

ivan-mogilko:
Hmm, I thought `fingers_down` could be removed actually, or even made a part of this FingerContainer. But iirc at the moment this bit field is not used as bit field, but strictly to count number of fingers (0, 1, >1). So if you remove it and replace with "get finger count", that should work for now.

>  so if I place an "invalid" finger, I don't know what to return there, and also I need to now ignore this invalid finger in all the other two cases (motion/up).

Maybe it will work if you don't place any "invalid" fingers at all, and ignore them. So, let's say, you receive a touch up / motion event, check whether this finger is in FingerContainer, and if not, just bail.

ericoporto:
https://github.com/adventuregamestudio/ags/pull/1955#issuecomment-1475356629

I haven't added comments yet or tried to move fingers_down into the container, but overall, did a bit of refactor there

> Maybe it will work if you don't place any "invalid" fingers at all,

I am trying this idea in the commit: [SKETCH deal with invalid fingers](https://github.com/adventuregamestudio/ags/pull/1955/commits/c2e335fc4076177f4e2bda4703ca52099728f20c)

ericoporto:
I think I can squash all of these in a single commit, it is working in my tests, just let me know and I will do so.

ericoporto:
Hey, I squashed the commits. The previous individual commits are kept here: https://github.com/ericoporto/ags/tree/fix-web-ios-touch-old

ivan-mogilko:
Yes, I think that may work nicely, I just may have few nitpicks about the code...

Engine/ac/sys_events.cpp
```diff
@@ -498,9 +498,66 @@ int ags_check_mouse_wheel() {
...
+    static const int MAX_FINGERS = 2;
+    static const int NO_ID = -1;
+
+    void push(SDL_FingerID fingerId)
+    {
+        if (contains(fingerId))
+            return;
+
+        for (int i = 0; i < MAX_FINGERS; i++) {
+            if (_fingers[i] == NO_ID) {
```
ivan-mogilko, line 521: I think this may be replaced with `std::find(_fingers.begin(), _fingers.end(), NO_ID)`. If result is _finder.end(), then there's no slot for a new finger, otherwise `*it = fingerId;` will set a finger id in place.

Engine/ac/sys_events.cpp
```diff
@@ -498,9 +498,66 @@ int ags_check_mouse_wheel() {
...
+                _fingers[i] = fingerId;
+                return;
+            }
+        }
+    };
+
+    int get_index(SDL_FingerID fingerId) const
+    {
+        for(int i=0; i < MAX_FINGERS; i++) {
+            if(_fingers[i] == fingerId)
```
ivan-mogilko, line 531: Same as above, this may be replaced with std::find(_fingers.begin(), _fingers.end(), fingerId)

Engine/ac/sys_events.cpp
```diff
@@ -678,7 +737,11 @@ static void sync_sys_mouse_pos()
 
 static void on_sdl_touch_down(const SDL_TouchFingerEvent &event)
 {
-    touch.fingers_down |= 1 << event.fingerId;
+    touch.fingers.push(event.fingerId);
+    int finger_index = touch.fingers.get_index(event.fingerId);
```
ivan-mogilko, line 741: If you make `push` return a new index, then you can save on 1 function call here, that is
```int finger_index = touch.fingers.push(event.fingerId);```

File: Engine/ac/sys_events.cpp [hunk 4]
```diff
@@ -794,10 +860,14 @@ static void on_sdl_touch_up(const SDL_TouchFingerEvent &event)
     }
     default: break; // do nothing
     }
+    touch.fingers.pop(event.fingerId);
 }
 
 static void on_sdl_touch_motion(const SDL_TouchFingerEvent &event)
 {
+    int finger_index = touch.fingers.get_index(event.fingerId);
```
ericoporto, line 868: Do you have thoughts on `tfinger_to_mouse_but` using finger_index instead of fingerId? If it used finger_index then I don't need to ask for it again inside of it.

Engine/ac/sys_events.cpp
ericoporto: If I do this, then I don't think I should do the iterator thing you suggested inside the push call, because then I need to calculate a distance to extract the index from the iterator.

Engine/ac/sys_events.cpp
ericoporto: how do I return the index using this?

ivan-mogilko: > how do I return the index using this?

`it - _fingers.begin()`

Engine/ac/sys_events.cpp
ivan-mogilko: I was thinking about this:
```
    int push(SDL_FingerID fingerId)
    {
        if (contains(fingerId))
            return NO_ID;

        auto it = std::find(_fingers.begin(), _fingers.end(), NO_ID);
        if (it == _fingers.end()) return NO_ID;
        *it = fingerId;
        return it - _fingers.begin();
    };
```

Engine/ac/sys_events.cpp
ericoporto: ok, I tried to do this in the sketch commit below

Engine/ac/sys_events.cpp
ivan-mogilko: Oh, sure.

ericoporto: OK, I added this in the commit below

Engine/ac/sys_events.cpp
ericoporto: ok, so, imagining if two touch down occurred in sequence, the idea here is to dismiss one of them

ivan-mogilko: >  imagining if two touch down occurred in sequence, the idea here is to dismiss the second

Could you elaborate, I don't quite understand this?

ericoporto: well, theoretically, that contain clause is there as a safety, but we should not need it. But in case the system misbehaves and we get two touch down before an up, using the same fingerId, then the second touch down is ignored. I think this is the correct behavior btw, so I agree with your code change.

ericoporto:
I think I can squash the commits now, the std::find showed I was mixing types for index and SDL_FingerID, so I fixed that.

ivan-mogilko:
Yes, i think this turned out nice, as all changes are contained in "index conversion" step, and the rest of logic remains.

ericoporto:
Apparently std::find can read by reference and on some compilers  this fails when the value it's looking is a static const, I removed the static qualifier and it appears to work locally. Please wait the CI to make sure it all actually builds.

PR #1957, opened 2023-03-24
Fix touch controls requiring two touches to activate a gui control

ivan-mogilko:
The problem at hand is that when using touch-to-mouse emulation, the cursor position cannot be tracked (unlike with real mouse) . It's only updated upon a touch. Unfortunately, the gui reaction to cursor in AGS is a bit overcomplicated, where it must remember the last control the mouse was on in order to trigger an interaction correctly upon a mouse down/up event. At the same time this "mouse-over-control" state is updated too late (for unclear reasons), after the mouse button events were already handled. Because of that, following problems may occur:
* it takes 2 touches to press a button (or interact with some other controls): first sets a "focus" on a control, another triggers a push.
* if the cursor has been over a button, a touch on another control or on an empty place over same gui will cause that previous button activate one more time.

Above is resolved by moving 2 updates up the update order:
* mouse position & cursor update;
* mouse-over-gui update.

(Honestly, I could never understand why the mouse position is updated so late in the order of things; theoretically it should be updated right before the other control inputs such as keys and buttons. OTOH this may also be a missed regression happened at some point in the past, where it could have been updated several times...)

**Potential danger:**
1. Changing order of things may be dangerous, so would need to analyze potential consequences. This move places cursor position and gui "focus" updates *before* a) majority of other game updates, b) script callback `late_repeatedly_execute_always`. (Other script callbacks won't be affected by this, as their relative order won't change).
2. Both of these updates were located inside the game render routine. By moving them out, we likely now require them to be called explicitly whenever this render routine been called explicitly and separately from the rest of the game update. At least this should be true for the mouse cursor update.
EDIT: easiest to observe during Display command.

### DIFF
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -866,6 +866,7 @@ bool DialogOptions::Run()
       else
       {
         update_audio_system_on_game_loop();
+        update_cursor_and_dependent();
         render_graphics(ddb, dirtyx, dirtyy);
       }
 
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -316,6 +316,7 @@ ScreenOverlay *_display_main(int xx, int yy, int wii, const char *text, int disp
             sys_evt_process_pending();
 
             update_audio_system_on_game_loop();
+            update_cursor_and_dependent();
             render_graphics();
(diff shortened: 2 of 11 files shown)


ivan-mogilko:
Squashed commits and tidied up the code a bit.

PR #1959, opened 2023-03-27
Fix uses of `long` type in the code (for compatibility with MSVS)

ivan-mogilko:
Because `long` in VC compiler stands for int32 even in 64-bit, it should not be used in cross-platform code.

This replaces uses of `long` accounting for the intended use:
* if it's used to store a 32-bit value, then replace with `int32_t` or `uint32_t` (depends on context).
* if it's used to store a memory address, then replace with `uintptr_t`.


ericoporto:
this looks fine, but just to be safe I tested with some games, but they all worked fine too.

ivan-mogilko:
Fixed few cases where I converted `singed long` to `uintptr_t`, made it `intptr_t` instead to keep signness.

Tested actual cases where these modified functions are called. Most of allegro changes affect software drawing, such as blending, which we use for e.g. tint, and anti-aliased scaling (with "smooth sprites" option, and software renderer).

PR #1960, opened 2023-03-28
Fix a crash in unavowed when trying to use crossfade channel

criezy:
A user reported to us a crash in the AGS engine in ScummVM, and I have verified that it happens in this standalone ags as well.

In Unavowed, at the Staten Island stage, the game crashes with the following message:
`Error: System.AudioChannels: invalid sound channel index 8, supported 0 - 8`

(the error message is actually incorrect and should read `supported 0 - 7`, and I fixed that as well)

And it is indeed trying to use channel 8, which is `>= game.numGameChannels` (which is also 8). I suspect this is caused by 86564e983572b4ffb3cd12b00c116ae91c6fcbe4. In any case allowing using the crossfade channel  in `System_GetAudioChannels` does fix the issue and the game can continue. I am proposing the fix here first (before committing to scummvm) to hear from AGS experts in case there is a better fix.

For reference: https://bugs.scummvm.org/ticket/14353

AGS savegame to reproduce: 
[agssave.003.zip](https://github.com/adventuregamestudio/ags/files/11094736/agssave.003.zip)

From this savegame, use Eli on the ceiling to open the door, go through the dialogues, and you get to the screen where the crash occurs:
image

### DIFF
--- a/Engine/ac/system.cpp
+++ b/Engine/ac/system.cpp
@@ -177,7 +177,7 @@ int System_GetAudioChannelCount()
 ScriptAudioChannel* System_GetAudioChannels(int index)
 {
     if ((index < 0) || (index >= game.numGameChannels))
-        quitprintf("!System.AudioChannels: invalid sound channel index %d, supported %d - %d", index, 0, game.numGameChannels);
+        quitprintf("!System.AudioChannels: invalid sound channel index %d, supported %d - %d", index, 0, game.numGameChannels - 1);
 
     return &scrAudioChannel[index];
 }
--- a/Engine/media/audio/audio.cpp
+++ b/Engine/media/audio/audio.cpp
@@ -166,7 +166,7 @@ static int find_free_audio_channel(ScriptAudioClip *clip, int priority, bool int
 
     // NOTE: in backward compat mode we allow to place sound on a crossfade channel
     int startAtChannel = reserved_channel_count;
-    int endBeforeChannel = game.numCompatGameChannels;
+    int endBeforeChannel = game.numGameChannels;
 
     if (game.audioClipTypes[clip->type].reservedChannels > 0)
     {
@@ -175,6 +175,7 @@ static int find_free_audio_channel(ScriptAudioClip *clip, int priority, bool int
         {
             startAtChannel += game.audioClipTypes[i].reservedChannels;
         }
+        // NOTE: we allow to place sound on a crossfade channel for backward compatibility
         endBeforeChannel = std::min(game.numCompatGameChannels,
             startAtChannel + game.audioClipTypes[clip->type].reservedChannels);
     }


ivan-mogilko:
Hmm, prior to 3.6.0 System_GetAudioChannels compared to MAX_CHANNELS, which was 8. I'm curious to test how previous versions (3.5.1) run this game first.

ivan-mogilko:
I suddenly had a thought, that maybe the 86564e983572b4ffb3cd12b00c116ae91c6fcbe4 did one mistake where it applied `numCompatGameChannels` in two locations, while it should have applied it only in one (the second edit). Because this apparently is where the code was fixed originally in b954ad3

ivan-mogilko:
So, it's obvious that the previous fix had an excessive change in one place, which was not required, and which changed the logic of finding a free channel.

I had to push a fix to release-3.5.1 too (3d55ced41af151305a5fa2bfbe90bffcd3dab6fb), because unfortunately it was done there as well about same time. I will have to release a new 3.5.1 patch; and can only hope that no game made since is relying on this incorrect behavior, or this would have to be restored with maybe more compat fixes...

The above commit may be used as a reference to fix 3.6.0, I think, if it makes "Unavowed" work.

criezy:
Thank you for investigating it further. I indeed noticed that `System_GetAudioChannels` had the same limit to 8 channel in AGS 3.4, where it works, which is why I assumed the game uses a lower channel in that version and that the change of behaviour and crash might come from the change to `find_free_audio_channel` in 86564e983572b4ffb3cd12b00c116ae91c6fcbe4. I could have made that clearer in the PR description, but it's good to see we arrived to the same conclusion.

However not fully understanding the change to `find_free_audio_channel` as it depends on a series of other commits, I was afraid that changing it again might have unexpected consequences. Since you indicate such a change should be good, I will test a similar fix for 3.6 to the one you did in 3d55ced41af151305a5fa2bfbe90bffcd3dab6fb and update the PR if that works.

ivan-mogilko:
Ah, I should have made a more clear comment there in code. I'll explain here, perhaps a shorter version could be added to the comments.

In AGS these "channels" are logical slots. There may be "reserved channels" per "audio type" (the audio type is an abstract property, defined by the user). When searching for a free channel for the given clip, the engine must skip those reserved by audiotypes higher in a list of types. For example, assume you have audio type A with 2 reserved chans, then type B with 1 reserved chans, then type C; then script commands to play a clip with type C, the engine will first skip first 2 + 1 channels, and search for a free channel among the remaining ones.

Unfortunately, AGS Editor never made any warnings or restrictions for a number of reserved channels. This let users reserve virtually any number. The related problem that I found in the engine was that when too many channels are reserved, the engine code did not make any safety checks and tried to use a slot beyond the channels array, leading to a program crash. This was fixed in b954ad324b6d0bc6b981419093c53062fc11f1f5.

But later it turned out that there was at least 1 game, which relied on certain sound getting past the "normal" channels onto the special "crossfade" channel. This worked simply because crossfade channel is allocated in the same channels array too (and engine did not have any bounds check).

So, the backwards compatibility fix should have *only* been applied in case when the engine is skipping reserved channels and goes too far.

ivan-mogilko:
@criezy looks like the above fixes the problem. I reproduced it with the current master, but with this change the crash no longer occurs.

criezy:
I confirm that your proposed change fixed the issue for me in both AGS and ScummVM.
I have updated this pull request accordingly.

PR #1977, opened 2023-04-14
[in doubt] Implement a "Right-to-left (reversed)" text direction, as a workaround for Arabic and Persian translations

ivan-mogilko:
**THIS IS A DRAFT**; in theory this problem may be resolved using a scripting module. I'll postpone this PR until it becomes more clear whether the scripting solution works and convenient enough for users.

**Problem**

Currently AGS's font renderer does not support drawing ligatures correctly for languages like Arabic, Persian, and maybe few more. It draws the letters, but does not connect them, which makes the text look incorrect.

One of the known existing workarounds is to use a converter program, that converts text by joining multiple glyphs into one "special" glyph. The resulting text may be displayed using corresponding TTF fonts that support this.

A problem about this conversion is that the resulting text ends up reversed in memory. Normally the RTL texts which are supposed to be displayed right-to-left are stored in memory in the order of their *reading*. But with the above workaround it is stored in memory opposite to how it's read.

For this reason, the existing text wrapping algorithm in AGS does not work for them, as the syntactical beginning of the phrase will appear on the last line, and the end of the phrase on the first line.

**Proposed solution**

Add a new selection to the text direction game option, called "Right-to-left reversed". Allow both game and translations select this mode. When AGS is in this text mode, the text wrapping function accounts for this and scans & splits the text in reverse order.

1. Replaced "Right-to-left" (boolean) property in the Editor with "Text direction" property, having 3 options now ("Left-to-right", "Right-to-left" and "Right-to-left reversed"). When upgrading an older project "Text direction" will be correctly set from "Right-to-left".
2. Translation source (TRS) supports new tag for this direction: "RIGHTREV".
3. Both Editor and standalone translation compiler (trac) taught about this new tag.
4. Engine already had OPT_RIGHTLEFTWRITE option in game data, I renamed it to OPT_TEXTDIRECTION; as it was an integer, it's possible to store any new values there already.
5. For the time being I solved the reverse wrapping by completely reversing the text before doing line split. This could be optimized, but that would require refactoring split_lines() again, and I don't have spare time for that yet. I don't think that current implementation will cause any impact on performance, as majority of displayed texts in the games are not more than 2-3 lines long.

Runtime game and save formats was not changed.

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -56,7 +56,7 @@
 #define OPT_ANTIALIASFONTS  25
 #define OPT_THOUGHTGUI      26
 #define OPT_TURNTOFACELOC   27
-#define OPT_RIGHTLEFTWRITE  28  // right-to-left text writing
+#define OPT_TEXTDIRECTION   28  // text direction (LTR, RTL, extra modes)
 #define OPT_DUPLICATEINV    29  // if they have 2 of the item, draw it twice
 #define OPT_SAVESCREENSHOT  30
 #define OPT_PORTRAITSIDE    31
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -12,7 +12,7 @@
 //
 //=============================================================================
 //
-//
+// Various global game constants
 //
 //=============================================================================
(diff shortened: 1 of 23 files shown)


ivan-mogilko:
Apparently the same effect was achieved with the script, so I will cancel this pr.

PR #1984, opened 2023-04-22
AGS 4: completely remove legacy audio API support

ivan-mogilko:
Remove the old audio system, which has been deprecated after AGS 3.2.0

From the looks of it, the only API that does not have an equivalent in the new system is playing a sound file from disk. But this may be easily reimplemented as a new function, if wanted.


AlanDrake:
Looks good

ivan-mogilko:
Also removed 2 obsolete properties from the Room (PlayMusicOnRoomLoad and MusicVolumeAdjustment).

Issue #1780, opened 2022-09-17
Translations: Possible implementation of PO files


AlanDrake:
## What are PO files?
A text-based translation format widely used.
https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files

## Why PO files?
Because it's a format not too distant from TRS files, and we can incrementally implement its features.
It's also well supported by several third party tools, like Poedit and other online services, which are convenient to translators and developers alike.

## Format
A single record is composed of the following:
```
white-space
#  translator-comments
#. extracted-comments
#: reference…
#, flag…
#| msgid "previous-untranslated-string"
msgctxt "context"
msgid "untranslated-string"
msgstr "translated-string"
```

`#  translator-comments` (optional) is self explanatory, a comment added by a translator, probably with a third party tool
`#. extracted-comments` (optional) could be a comment in the same line of the string, these are generally used as translation notes for translators
`#: reference…` (optional) e.g: `#: module.asc:167`, can have multiple references separated by space and can appear multiple times (eg: xgettext splits it by 80 columns)
`#, flag…` (optional) can have `#, fuzzy, c-format`  fuzzy means the translator wasn-t sure of translation, c-format is a flag that can be optionally applied to string that have c-style syntax "like this %d"
`#| msgid` (optional) probably unnecessary for us
`msgctxt "context"` (optional) context, could be our way to declare same string and multiple translations, the question is how to apply the context? (script/module name? user-defined?)
`msgid "untranslated-string"` (required) the string to translate
`msgstr "translated-string"` (required) translated string, `""` if empty

Note: the empty msgid is reserved to declare headers, as follows.

## Headers
Metadata is usually recorded on a empty id "" as the first record on file, can be preceded by # comments
```
# Optional comment
msgid ""
msgstr ""
"Project-Id-Version: MyGame\n"
"POT-Creation-Date: 2022-09-17 18:04+0200\n"
"PO-Revision-Date: 2022-09-17 18:04+0200\n"
"Last-Translator: \n"
"Language-Team: \n"
"Language: it\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"
"X-Generator: AGS 3.99.99\n"

```
Additional headers could be added here, like NormalFont, SpeechFont, TextDirection, or they can stay as comments.

## Implementation roadmap

[...]

Issue #1226, opened 2021-02-28
Drop [ as line feed when displaying strings


ericoporto:
AGS supports using `[` as line feed when using `Display (my_string)` and also when rendering text in Labels and a couple more places. It also supports `\n`. I don't know what's the strategy to drop support but it would be nice if we could at least disable the parsing of `[` as a new line indicator - allowing to Display unexcaped `[`.

ivan-mogilko:
I'd say yes, because this is some workaround that I can't explain (other than CJ has some difficulties passing '\n' from script in early versions?).

AlanDrake:
Agreed

ivan-mogilko:
As pointed out by @AlanDrake , the text entered in the properties is not processed to convert escape sequences into corresponding symbols, e.g. "\n" into `\n`. This is only done for script text during compilation.

If we drop `[`, all the text properties need to be processed similarly to script at the time of compiling game data.

ivan-mogilko:
One important question remaining is a project upgrade. Supposing we drop support of runtime feature, but what should we do with the imported project?

I guess we'd require some utility which replaces "[" with "\n" everywhere in script, properties and translation sources...

AlanDrake:
The migration utility should be careful to differentiate `\\[` `\[` and `[` correctly.

ivan-mogilko:
'[' is dropped by #1988
"\n" is supported in editor's properties after #2055.

What is left now is to decide what migration process do we want when upgrading old projects. Potential issue is replacing '[' instances with '\n' in properties, script and translation sources.

messengerbag:
I think the project upgrade is complicated by the fact that before the latest version, there were differences in how different GUI Controls handled "[" and "\[". IIRC, on Buttons and in TextBoxes and ListBoxes they were just displayed without any conversion, and games may have relied on this behavior. For example, the AGS Awards Ceremony has code to format game names (which sometimes include brackets) correctly depending on what sort of Control they appear on.

My intuition is that an automatic conversion when upgrading old projects is too difficult/impossible, and it should be left to game developers to fix.

PR #1986, opened 2023-04-24
Editor: PO translation format implementation

AlanDrake:
This PR implements support for translation with PO file format, as discussed in  #1780

Todo list:
- [x] upgrade existing .trs files (without deleting them) 
- [x] preserve existing comments/metadata for the entries during updates
- [ ] add an option to declare the base game language (necessary to set the po file base language, the filename itself is the translation language)
- [ ] deal with obsolete entries (still to decide how, delete or flag as fuzzy and add comment)
Maybe to leave to another PR:
- store source and line number
- extract comments for translators from code

Compatibility issues:
\[ should be escaped as \\[, or that old syntax should be removed, because third-party tools may omit entries with invalid string escape sequences.


ivan-mogilko:
> \[ should be escaped as \\[, or that old syntax should be removed, because third-party tools may omit entries with invalid string escape sequences.

Perhaps move to '\n'? See #1226

ivan-mogilko:
>  deal with obsolete entries (still to decide how, delete or flag as fuzzy and add comment)

Perhaps this may also be left to the next update? AGS trs currently does not support this, so leaving this won't break any existing functionality.

AlanDrake:
> > deal with obsolete entries (still to decide how, delete or flag as fuzzy and add comment)
> 
> Perhaps this may also be left to the next update? AGS trs currently does not support this, so leaving this won't break any existing functionality. This PR will then be seen purely as a trs format migration.

Alright then.
Tracking the deleted records is easy, I'm just not sure about the best approach to mark them as such. Let's leave it for a future PR.

ivan-mogilko:
I tried this, and there's a number of problems.

The first to mention is related to changing Translation class. AGS.Types and everything in it is considered Editor's "API", apparently, and may be used by plugins. For instance, when you change Translation.TranslatedLines that breaks SpeechCenter plugin which relies on it.

Supposing this cannot be worked around, since we must have a complex translation entry now, that may be fine (although that means that SpeechCenter will have to be updated for ags4).
But I believe that TranslationEntry must be remade considering its use as a "interface". That is - it should be done as a distinct class with properties instead of a nested class inside Translation, with bare string fields. That would let extend the TranslationEntry without breaking plugins again. Also - potentially - would let view, edit and serialize TranslationEntry similar to other game data, if such need arises.

ivan-mogilko:
Speaking of performance now...

1. I don't know if this happens only in the new game (I created a new game from template for a test), but Updating a translation is causing an exception because `_baseLanguage` is null. Apparently it's never set. I think it should be either preset with some non-null value (e.g. empty string) for safety. But I was not able to understand how it's supposed to be set normally.
2. I tried to edit the file by hand instead, and found that the last entry in PO is not added to translation dict. This may be worked around by adding some extra line in the end. Later I found there's a comment in code:
``// note: if there's not an empty line at the end, the last record won't be stored``
Is it a requirement of PO specification? If not, why not simply add the current entry if it contains anything?
3. A bit of a random suggestion: I think it will simplify parsing algorithm if you test for comments first, as that will let skip multiple keyword checks.

AlanDrake:
> But I believe that TranslationEntry must be remade considering its use as a "interface". 

So you basically want to extend a Dictionary wrapper to hijack lookups? I never tried doing that.

Personally, I wish AGS code became simpler rather than more convoluted, but I lost hope about that.

> Speaking of performance now...
> 
>     1. I don't know if this happens only in the new game (I created a new game from template for a test), but Updating a translation is causing an exception because `_baseLanguage` is null. Apparently it's never set. I think it should be preset with some non-null value (e.g. empty string) for safety. But I was not able to understand how it's supposed to be set normally.

I was under the delusion that `private string _gameTextLanguage = "en_US";` would ensure there's always a default value, but I guess I should add an `[DefaultValue("xxx")]` attribute instead.

The base language string is changed from General Settings.
 
>     2. I tried to edit the file by hand instead, and found that the last entry in PO is not added to translation dict. This may be worked around by adding some extra line in the end. Later I found there's a comment in code:
>        `// note: if there's not an empty line at the end, the last record won't be stored`
>        Is it a requirement of PO specification? If not, why not simply add the current entry if it contains anything?

The empty line is explicitly required by the specification, so I didn't bother handling that case.

>     3. A bit of a random suggestion: I think it will simplify parsing algorithm if you test for comments first, as that will let skip multiple keyword checks.

I'll give it a try.

>     4. I think that the header should be written to PO file even if user chose to not update translation after creating it. But I don't remember if this always worked like this. If so, then I'd even fix that for the ags3 branch, for convenience.
>        In other words: create file, write header, then if user chose to "update now" then add the game texts.

I haven't changed how that works, but then again, if it's completely empty the headers aren't gonna do much anyway.

AlanDrake:
Hmm, for some reason multiline strings are no longer working, even though I see the value at index 1 in the matches... I think I can perhaps simply use match.Value directly in this case.
Also I spotted a mistake that ruins performances.

AlanDrake:
Damn, as usual my PRs have a way to "degrade" and stop working properly. I must be tired when I rewrite and "optimize" them.
I'll try moving up the metadata check tomorrow, it's too late now.

ivan-mogilko:
> > But I believe that TranslationEntry must be remade considering its use as a "interface".

> So you basically want to extend a Dictionary wrapper to hijack lookups? I never tried doing that.

I do not understand what you meant here.
What I proposed is to move TranslationEntry out as a standalone class and declare public properties to access the internal fields. This would allow to safely extend it if the editor plugins were accessing this object.

> Personally, I wish AGS code became simpler rather than more convoluted, but I lost hope about that.

What is "convoluted" and more convoluted compared to what?

ivan-mogilko:
> I was under the delusion that private string _gameTextLanguage = "en_US"; would ensure there's always a default value, but I guess I should add an [DefaultValue("xxx")] attribute instead.

It's not `_gameTextLanguage` that is used there, but the `_baseLanguage` field.

AlanDrake:
> > > But I believe that TranslationEntry must be remade considering its use as a "interface".
> 
> > So you basically want to extend a Dictionary wrapper to hijack lookups? I never tried doing that.
> 
> I do not understand what you meant here. What I proposed is to move TranslationEntry out as a standalone class and declare public properties to access the internal fields. This would allow to safely extend it if the editor plugins were accessing this object, as well as hide its implementation. This is a common approach when you have external programs accessing your classes.

TranslationEntry is another thing. I mean that I didn't know the getter/setter syntax to make TranslatedLines hijack the index key and wrap its behavior. Now my mind is fresher and I've found it.

> > Personally, I wish AGS code became simpler rather than more convoluted, but I lost hope about that.

Here not much, just a random rant because everytime I have to touch some new part I'm not acquainted with, it's always hours wasted merely trying understand how things are pieced together.
Or maybe it's just PTSD from the times I tried to add stuff and retrocompatibility made me waste thrice the time for what was meant to be a simple thing. So now when you ask me about make things compatible my brain lashes out.

> 's not `_gameTextLanguage` that is causing the exception, but the `_baseLanguage` field.

Same thing, `_baseLanguage` is instanced with `_gameTextLanguage`'s value.

AlanDrake:
I moved out the class, and turned props into acccessors, then renamed TranslatedLines into TranslatedEntries.
But I don't know how to code a compatbile wrapper for the original Dictionary<string, string> TranslatedLines.

ivan-mogilko:
> But I don't know how to code a compatbile wrapper for the original Dictionary<string, string> TranslatedLines.

I did not ask about this, I said that:

> TranslatedLines that breaks SpeechCenter plugin which relies on it.

Supposing this cannot be worked around, since we must have a complex translation entry now, that may be fine (although that means that SpeechCenter will have to be updated for ags4).

Editor/AGS.Types/TranslationEntry.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class TranslationEntry
+    {
+        public TranslationEntry()
+        {
+            metadata = new List<string>();
+        }
+
+        public string msgctxt { get; set; }
+        public string msgid { get; set; }
+        public string msgstr { get; set; }
```
ivan-mogilko, line 18: Please, make more comprehensible names, following the code style in AGS.Types.
These names do not need to follow the PO format keywords, as this is AGS translation entry, not PO's, we only save in PO.

E.g.:
```
string Context
string ID;
string String or Line;
etc
```

AlanDrake:
Renaming done.
What else would you do?

ericoporto:
The previous naming was direct from the PO file spec, I thought that renaming would be simply changing the case as MsgId and MsgStr. Naming that property only Id and String loses a bit of the key-value relationship these had - at least when one would check the PO file spec.

ivan-mogilko:
> The previous naming was direct from the PO file spec, I thought that renaming would be simply changing the case as MsgId and MsgStr. Naming that property only Id and String loses a bit of the key-value relationship these had - at least when one would check the PO file spec.

My thinking was that we don't have to use literal PO keywords in our classes, given this is practically an API.
Do you think they should?

AlanDrake:
I don't know, somebody toss a coin.

ericoporto:
I was just thinking how this works in terms of the relationship of an ID (key) and context specify a text (value). 

When String names a property, this relationship isn't clear - we don't have neither the Key-Value naming nor the PO naming. Line would be a better name for that property. Or using the PO concepts too.

Overall I mostly think String as a property means "give me this class instance serialized as a String".

AlanDrake:
Key/Value then and let's call it a day.

ivan-mogilko:
I came back at testing this, and the error I reported is not fully fixed.

1. Start a new game.
2. Create a new translation.
3. Editor will ask if you want to update the translation, say yes.
4. Observe exception in AGS.Types.Translation.Encode called from this line:

```
Encode(_baseLanguage)
```
because _baseLanguage is null.

Noteably, this error is now resolved if you close and then reopen the project. It looks like only affected translations created during current session.

ivan-mogilko:
In regards to the Translation.Encode and Translation.Decode methods. Do I understand correctly that the first does the special character escaping, and second unescapes them?
Decode does not seem to be used anywhere, and Encode is used only in 2 places when writing a meta header.

Overall I believe that the place for such methods is in Utilities (like [here](https://github.com/adventuregamestudio/ags/blob/ags4/Editor/AGS.Types/Utilities.cs)), otherwise there is a high chance that someone will write another version of the same thing.

But before that, i think it's worth to double check that there are not methods doing suitable escaping/unescaping in NET Framework. For instance, for unescaping I've been using Regex.Unescape. (Sadly Regex.Escape cannot be used for generic purposes, as it escapes regex-specific characters).

ivan-mogilko:
Testing migration from TRS, it seems working, but I have a question: what is the purpose of leaving old TRS files?
I have a concern that having them in the same folder like that may confuse people, and they may just keep them forever.

AlanDrake:
> I came back at testing this, and the error I reported is not fully fixed.

Damn, I made a test and I thought I had fixed it... I'll just force the value if null.

> In regards to the Translation.Encode and Translation.Decode methods. Do I understand correctly that the first does the special character escaping, and second unescapes them?

Sanity check. Also may be needed if we need parse the metadata in the first empty record, which I am currently regenerating.

> Testing migration from TRS, it seems working, but I have a question: what is the purpose of leaving old TRS files where they are? I have a concern that having them in the same folder like that may confuse people, and they may just keep them forever.

PR #1988, opened 2023-04-25
Removes special cases for [ as a newline character

AlanDrake:
This PR removes the ability for the `[` character to be used as a newline.

It also means that it won't be possible to add newlines to GUI Labels until another PR to unescape the escaped "\n" in their Text properties from the Editor (when set from script they will work as usual).

### DIFF
--- a/Common/font/fonts.cpp
+++ b/Common/font/fonts.cpp
@@ -370,40 +370,6 @@ int get_text_lines_surf_height(size_t fontNumber, size_t numlines)
 
 namespace AGS { namespace Common { SplitLines Lines; } }
 
-// Replaces AGS-specific linebreak tags with common '\n'
-static void unescape_script_string(const char *cstr, std::string &out)
-{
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
@AlanDrake 
> It also means that it won't be possible to add newlines to GUI Labels until another PR to unescape the escaped "\n" in their Text properties from the Editor (when set from script they will work as usual).

Please tell, was there actually a PR, or a ticket about this?

AlanDrake:
> Please tell, was there actually a PR, or a ticket about this?

Yes, sorry. I see you already found the old ticket (which I forgot existed).
It's also indirectly mentioned in #1986 about removing  the non-standard escape sequence "\[", which would cause problem with third party editors.

AlanDrake:
Migration could be a different task. Only thing to add here is point 1, so labels and properties convert \n to a newline.
I'll rebase asap.

PR #1989, opened 2023-04-26
Editor: feature console (log) panel

ericoporto:
no significant changes from #1518 , just reopening


ivan-mogilko:
From my memory, the things to improve to make this usable are only UI related:
1. Allow to scroll up while it's working (i recall it forced log to go onto the last line all the time). That was a primary problem with it when I last tested. Optionally: add a check tool button that "glues" to the last line.
2. Make sure there's a way to copy-paste the text: text should be selectable, and combinations like Ctrl+A and Ctrl+C work. Perhaps even have context menu with Select All / Copy commands.
3. Optimize copying the new entries to the text control (iirc there was a redundant copy of whole buffer instead of only the last log item).

Optional:
- Add a pause/resume control on toolbar, that would stop writing the log into the text control.

ericoporto:
> Make sure there's a way to copy-paste the text: text should be selectable, and combinations like Ctrl+A and Ctrl+C work. Perhaps even have context menu with Select All / Copy commands.

On the selectable note, I think allowing selection reduces performance of the richtextbox significantly, which is why it's disabled.

ivan-mogilko:
Started testing & experimenting with this.
It actually seem to work pretty nicely when there's no frequent log prints: then you may have a moment to scroll up and read or copy previous messages.
I'd like to try adding a toolbar with few commands, as noted above, something only to make this panel more convenient.

> On the selectable note, I think allowing selection reduces performance of the richtextbox significantly, which is why it's disabled.

Hmm, I can see that selection itself is working, I can select and press Ctrl+C to copy. Did not do any performance tests though yet.

Two more things for TODO list:
* There's a bug: for some reason the log settings only apply when you toggle them before starting the game for the first time.
* Perhaps need to add a font setting to Editor preferences, it starts way too small in size.

ericoporto:
One thing I could not figure a good approach is, if you have only one monitor, because AGS Engine runs in a new window and not in a panel in the Editor, the normal config of debug makes it occupy enough screen to be in front of the log - so you kinda want the debug screen to be a bit smaller (if the game resolution allows for it), so you can actually see the log panel.

ivan-mogilko:
Not sure how to proceed with this, but I began adding fixes to the branch in my repo:
https://github.com/ivan-mogilko/ags-refactoring/commits/feature-360-console-editor

Added so far:
* Fixed that user config was not applied to the log buffer right after loading a game;
* Added toolbar with Run/Pause buttons (helps to read the log history, and copy things);
* Added "glue" mode and "auto glue" tool button (checkable).

Warning: toolbuttons dont have any icons yet, I honestly did not want to bother at this time :). They may be distinguished by tooltips.

The "glue" works like this:
- in auto-glue mode the textbox always scrolls along with the log (sort of "glued" to it);
- in manual glue mode (button unchecked) it stops auto-scrolling if user scrolls up, and starts auto-scrolling again if user scrolls down to the very end.

Keeping the textbox at current position is done is a very ugly way, but that's as much as I could do in the situation where text is completely replaced all the time. Further improvement should involve optimization to how buffer->textbox communication works (i may look into this further).

ericoporto:
About the toolbar, any reason to not use the properties menu? I remember there was an area for buttons and stuff and then switched to use the properties bar, like the glue thing could just be there as a property, right?

ivan-mogilko:
> About the toolbar, any reason to not use the properties menu? I remember there was an area for buttons and stuff and then switched to use the properties panel, like the glue thing could just be there as a property, right?

Please elaborate, what is "properties menu"?

ericoporto:
When you click on the log panel, the properties panel (normally at bottom right I think), should switch for the properties of the log - where you can specify things like filtering for a specific message group or set the lowest level you accept on that.

About my repository, you never accepted my invite, so it had expired, I added you again, it should give you a new invite. But you can also close this and submit a PR from your branch too. :)

ivan-mogilko:
> When you click on the log panel, the properties panel (normally at bottom right I think), should switch for the properties of the log - where you can specify things like filtering for a specific message group or set the lowest level you accept on that.

Hm, I thought you meant something else.
I would not like to put things like "glue" switch there, because it's not a configuration, but a command meant to be used occasionally.

ivan-mogilko:
Speaking of properties pane, I may have forgotten this, but do I understand correctly that the DebugLogComponent creates ContentDocument for the purpose of linking to the properties pane?

I'm currently trying to find a way to let a Component to add menu items to the Window menu, and I found that in the Window menu manager there's a division between persistent panels that inherit DockContent, and temporary panels that inherit ContentDocument.
From what I understood, ContentDocument is meant for editing the game items (it has members like "tree node id" and IEditorComponent).
The LogPanel currently is somewhere in the middle, because it's supposed to be "persistent" but inherits ContentDocument.

There are two ways to go from here:
1. Allow Window menu to work with "persistent" ContentDocuments.
2. Make LogPanel to inherit DockContent, but then invent a different custom way to connect to the Properties Panel.

The p1 may work as a temporary method, perhaps.

ericoporto:
> do I understand correctly that the DebugLogComponent creates ContentDocument for the purpose of linking the LogPanel to the properties pane?

Yes. 

There's also the problem of where to put the config of the messages on either the game project or for the editor preferences (all projects), so this moves those as per game project.

ericoporto:
Sorry, I was trying to understand the three fields here: https://github.com/ivan-mogilko/ags-refactoring/commit/b41ed86be57056f250ba7c9a463aa2eb528ba623#diff-af54cb141d68b5f75b3119aec1ac8eb4cdd1d550bbfcafebf516b2580815a83cR22

I see one is a temporary state thing, maybe moving the two things that are mirrors for the checkboxes/button to the top.

> From what I understand, the LogComponent saves config to user data because it implements IPersistUserData interface. This does not depend on ContentDocument

That's correct, sorry, I confused with a previous version of the code. Yeah, that is fine.

ivan-mogilko:
```
        private bool _glue = true; // stick to the log's end
        private bool _autoGlue = true; // force glue, regardless of user's actions
        private bool _noScrollCheck = false; // temp disable checking scrolling event
```

So, `_glue` variable stores the current state: whether textbox scrolls down as log added or not.
`_autoGlue` is a property which tells whether to force `_glue` on all the time. If _autoGlue if off, then _glue may be turned off by user doing something, like scrolling up.
`_noScrollCheck` - tells to ignore scrolling events. This is necessary to temporarily disable reaction to scrolling when setting new text and reapplying position.

ericoporto:
A minor note on the panel, the [original version was just a console](https://github.com/ericoporto/ags/commit/adb5dbe2e5f15f66981ebff8a57de5fd9a4dc7c8), like a read-only terminal, because I was also prototyping with the command line tools in mind. This may have made me make bad decisions as I kept this in the back of my head when working on this, so if I made something that makes no sense, just change it.

ivan-mogilko:
Heh, I sort of made the Window menu work: https://github.com/ivan-mogilko/ags-refactoring/commit/8d65fd966fb65fcfe6c56da186203b834ccf9aa0

It's not perfect, as the LogPanel is wrapped into the ContentDocument, it also gets mentioned as a "active document" in the second part of the menu, under the splitter. But it works besides that, or so it seems.

ericoporto:
I pushed your commits here just to have them building.

I sketched some buttons, trying to match the original ags icons from the resources folder

image
image
It would be like: log running, log is silent, log glues to always scrolling, log is not autoscrolling.

ivan-mogilko:
Found a solution for textbox blinking when staying in the middle: https://github.com/ivan-mogilko/ags-refactoring/commit/96517ed650d0c37d82761c33e4c4f94635915b1c

ivan-mogilko:
Okay, fixed the log jumping to the caret in "unglued" mode: https://github.com/ivan-mogilko/ags-refactoring/commit/4bcc6e062270b9636a5d5761bcdf6c5526237603

ivan-mogilko:
Pushed an experimental change that reduces the amount of text copied into the rich text box each time, and also use control-specific messages to edit text, instead of replacing whole text each time.
(Commits: https://github.com/ivan-mogilko/ags-refactoring/commit/c9261a9ee60f56dd3b2b5617fc41fd8f9054305b, https://github.com/ivan-mogilko/ags-refactoring/commit/a9ab6845936da2587197eeb36f04386a38ce6ce6 and https://github.com/ivan-mogilko/ags-refactoring/commit/c14b0ca154d5b61a68ff781e591a98903619f61c)

The idea is following:
* LogBuffer has 2 dequeues now; one is "full" queue and another is "fresh" (dumb name but I ran out of imagination). When a new item is pushed, it's queued into the full queue, and also into the "fresh", applying a log filter.
* External user may request either full content, or "recent" content. If recent content is requested, the "fresh" queue is drained. If full content is requested, the "fresh" queue is discarded.
* LogBuffer event args now have 3 params: whether buffer was "reset" (this is when large changes were applied), whether a number of entries was pushed, and whether anything was popped out.
* When LogPanel receives the buffer event it choses what to do:
    - if buffer was reset, then it applies whole text anew;
    - if something was popped, it erases N characters from the start of the text;
    - if something was pushed, it appends new strings to the end of the text.

ivan-mogilko:
Pushed more changes to my branch, where remade LogPanel into DockContent, added support for adding new dock panes to the main form, which will be serialized in layout and added to the windows menu in the uniform way.
https://github.com/ivan-mogilko/ags-refactoring/commit/e78a8dceec8f17439a07584f469ce3825ce78920

For the property grid, I could not find other way but to create a new small interface, which i called IPropertyGridClient. The main form now detects when any docking pane becomes active, and tests whether this pane implements IPropertyGridClient. Otherwise it fallbacks to using property grid bindings from active ContentDocument.
https://github.com/ivan-mogilko/ags-refactoring/commit/969aa2c5a9f0df3eecab6da53f2131d8c215dc34

The tests were mostly successful, the only difference that I noticed is that log properties are hidden when the Editor window looses focus. Perhaps there is a way to fix this too...

If this works, I guess I would recreate the branch squashing at least 2 commits (the ones with EXPERIMENTAL note).

ivan-mogilko:
Rebased my branch, squashing few commits: https://github.com/ivan-mogilko/ags-refactoring/commits/feature-360-console-editor

Renamed few things, and added more comments.

ivan-mogilko:
Rebased my branch, squashing few commits: https://github.com/ivan-mogilko/ags-refactoring/commits/feature-360-console-editor

Renamed few things, and added more comments.

ivan-mogilko:
I rebased my branch, squashing few commits, renamed few things and added comments.
https://github.com/ivan-mogilko/ags-refactoring/commits/feature-360-console-editor

There's one irritating issue that I did not manage to solve at this moment. Because I tried to make LogPanel not a ContentDocument, but a DockContent, similar to other "persistent" panels (Output pane , and so on), I had to use a separate method to assign its config to the property grid panel. I was expecting the other ContentDocuments would switch the properties on activation, and they normally do. But the problem is that some of the editor panels have some kind of "inactive areas" on them, and clicks on these do not cause "activate" event, it seems. In these cases the only way to switch properties is to click on either the Tab caption, or any editable field/area inside the panel. So, this is not critical, but may be annoying to users.
I might keep searching for some good method of receiving the event.

ericoporto:
@ivan-mogilko hey, I think it makes more sense to close this PR and open a new one using your branch.

ivan-mogilko:
> @ivan-mogilko hey, I think it makes more sense to close this PR and open a new one using your branch.

Alright, I opened a new one: #1999

Issue #1987, opened 2023-04-25
Script performance drop between 3.4.3 and 3.6.0


ivan-mogilko:
This is rather a "note to self" ticket, to avoid information loss.

While doing a performance test on heavy script games (related to #1923) I noticed there have been a performance drop between 3.4.3 and 3.6.0 versions. I used 3.4.3 as a reference, because the test game was made in it.

Furthermore, this perf drop could be roughly divided in 2 parts: 3.4.3 -> 3.5.1, and 3.5.1 -> 3.6.0.
The actual values are of course subjective to a game & system; in my case I've been testing a 3D FP space sim prototype with lots of dynamic array manipulation and matrix math. Running in "infinite fps" mode, my values were:
3.4.3: 75 fps
3.5.0: 72-73 fps
3.5.1: 69 fps
3.6.0: 65 fps

I realize that finding a cause of such decrease may not be a easy task, because performance may differ after tiny changes in the code, and overall difference may accumulate over dozens of changes. But I am currently trying if I can pinpoint any "outstanding" changes.

---

In case of 3.5.1 -> 3.6.0 diff I found a changeset that drops from ~69 to ~65 fps. These are commits from f3443cb7a4ae5858e5d2d28131aaf6cf138bc0ec to 3b43353986ca91ecad9c5270e362c8716effaf37, that introduced a "stack" of running script instances for more detailed error reporting.

At first I suspected pushing and popping of a dequeue, but it's not as frequent. What *seem* to be causing a real impact is hiding of several global variables (containing last error info) behind the function calls: 48ef387226e4d420dc62d0a0b204cf4cd89a1139
I suspect that this is because cc_has_error() and cc_clear_error() are called way to frequently when executing the bytecode.
If so, inlining them might be fix this to some degree. Alternatively, the error data manipulation may be done differently (few related things there seem suspicious).
SIDE-NOTE: related, there's also #843, this may be addressed alongside, at least if some checks seem redundant or may be done better.

I will continue investigating this.

ivan-mogilko:
Fixed by #1990 (and improved significantly more than planned).

PR #1990, opened 2023-04-28
Performance fixes for script interpreter

### LINKED ISSUE #843: Disable script interpreter safe checks for regular build
As noted in previous discussion (#815) the checks in script interpreter may produce overhead, which is not good especially since it's not very fast and optimal in general.

The suggestion was to replace the checks with an assert-like function that definition depends on compilation flags. Quoting parts of the discussion:

> Still I would like to have some kind of "assert" function there that would trigger a break point automatically under debugger. Mere warning message may not be convenient because when debugging it may be essential to stop at exact moment something unexpected happen to observe the issue.

> Perhaps there could be a function like "script_assert" which definition depends on compiler flags. It could be a no-op for release build, a warning output for regular debug build and an actual assert if we have e.g. DEBUG_SCRIPT_VM flag defined.

> **PS.** maybe not literally "assert" though, since its implementation may abort the program afterwards, but something else that works as a break point.

> the only portable way to do this is to emit SIGTRAP signal, but that's only for POSIX compatible platforms. for windows, if x86 cpu is used, one can use __asm__("int3");. both of them will cause the program to abort if there's no debugger attached though. so one could make it dependant on whether a debugger is attached, which on POSIX can be found out via ptrace with PTRACE_TRACEME, iirc. but i don't know for windows.

ivan-mogilko:
Resolves #843 and #1987.

Contains a number of adjustments to the script interpreter function (ccInstance::Run), increases potential game fps and reduces CPU load for "script heavy" games.

Adds `DEBUG_CC_EXEC` compilation flag, and hides most error checks within the opcode implementations under this flag.
DEBUG_CC_EXEC is enabled if AGS_PLATFORM_DEBUG is enabled, but we may also make it completely separate later if wanted.

The changes were tested against two games with complex scripts in the "infinite fps" mode (where engine runs game loops as fast as it can):
1. First person space sim prototype: https://github.com/ivan-mogilko/ags-wcproto
2. Car racing arcade game: https://github.com/ivan-mogilko/ags-lastfurious

First has generally very slow script; showed a fps rise from ~64-65 to ~75 in idle state.
Second - from ~145-155 fps to 180-190 fps, when idling at the starting grid while 5 ai cars were racing the loop.
which gives an improvement by something like 15-20% depending on a case.
**UPDATE:** after few more commits made since:
First game: rise from ~65 to 80 fps
Second game: rise from 150 to 200 fps
**UPDATE:** after another [sudden idea](https://github.com/adventuregamestudio/ags/pull/1990/commits/719f67e831f7de3da831a957da0cc69a7a4a5026):
First game: rise from ~65 to 84 fps
Second game: rise from 150 to 210 fps
So, this is improvement by ~30% now, or more, depending on a game.

These numbers are of course subjective to an individual system (and game), but probably may give an idea of improvement.

### DIFF
--- a/Common/script/cc_internal.h
+++ b/Common/script/cc_internal.h
@@ -112,6 +112,7 @@
 #define EXPORT_FUNCTION   1
 #define EXPORT_DATA       2
 
+#define FIXUP_NOFIXUP     0     // no-op
 #define FIXUP_GLOBALDATA  1     // code[fixup] += &globaldata[0]
 #define FIXUP_FUNCTION    2     // code[fixup] += &code[0]
 #define FIXUP_STRING      3     // code[fixup] += &strings[0]
--- a/Engine/script/cc_instance.cpp
+++ b/Engine/script/cc_instance.cpp
@@ -623,18 +623,6 @@ int ccInstance::Run(int32_t curpc)
         /* End ReadOperation */
         //=====================================================================
 
-        // save the arguments for quick access
-        RuntimeScriptValue &arg1 = codeOp.Args[0];
-        RuntimeScriptValue &arg2 = codeOp.Args[1];
-        RuntimeScriptValue &arg3 = codeOp.Args[2];
-        RuntimeScriptValue &reg1 = 
-            registers[arg1.IValue >= 0 && arg1.IValue < CC_NUM_REGISTERS ? arg1.IValue : 0];
-        RuntimeScriptValue &reg2 = 
-            registers[arg2.IValue >= 0 && arg2.IValue < CC_NUM_REGISTERS ? arg2.IValue : 0];
-
-        const char *direct_ptr1;
-        const char *direct_ptr2;
-
 #if (DEBUG_CC_EXEC)
         if (dump_opcodes)
         {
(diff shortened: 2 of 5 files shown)


ericoporto:
> but we may also make it completely separate later if wanted.

In cc_instance.h, use

```
#ifndef DEBUG_CC_EXEC
#define DEBUG_CC_EXEC (AGS_PLATFORM_DEBUG)
#endif
```

ivan-mogilko:
> > but we may also make it completely separate later if wanted.
> 
> In cc_instance.h, use
> 
> ```
> #ifndef DEBUG_CC_EXEC
> #define DEBUG_CC_EXEC (AGS_PLATFORM_DEBUG)
> #endif
> ```
> 

I had an opposite thought of being able to turn it off in debug config too, as debug build is already quite slow.

ericoporto:
Along with the commit https://github.com/adventuregamestudio/ags/pull/1990/commits/d9284d8c1a265f0c4b329c27dde2997f465375d4

I suggest making the loopinteractions check use 1024 instead of 1000 too (https://github.com/ericoporto/ags/commit/513d18daf797b13492d4a11893f8ef8aed7790e3)

ivan-mogilko:
> Along with the commit [d9284d8](https://github.com/adventuregamestudio/ags/commit/d9284d8c1a265f0c4b329c27dde2997f465375d4)
> 
> I suggest making the loopinteractions check use 1024 instead of 1000 too ([ericoporto@513d18d](https://github.com/ericoporto/ags/commit/513d18daf797b13492d4a11893f8ef8aed7790e3))

This is storage vs arbitrary number of times.

ivan-mogilko:
So, after a curious @ericoporto 's idea to [replace a argument fixup loop with a switch](https://github.com/adventuregamestudio/ags/pull/1990/commits/a495a00f08c9aed395d6ad6513dab35d551fd613), which gave another good boost to performance at a cost of somewhat ugly code, I suddenly had a thought that we were doing this wrong all this time. Analysing the fixup uses shows that they potentially applied only to literal args, and only when arg can be a "address" (to a script, offset in stack, pointer to function, pointer to a object, etc). In the end this narrowed down to only *two* instructions: SCMD_WRITELIT and SCMD_LITTOREG.

*Unless I am making some stupid mistake*, these two instructions SCMD_WRITELIT and SCMD_LITTOREG are the only ones which argument may potentially be a result of fixup. So I pushed [another commit](https://github.com/adventuregamestudio/ags/pull/1990/commits/719f67e831f7de3da831a957da0cc69a7a4a5026) which simply deletes a fixup loop/switch in front, and only calls FixupArgument explicitly for these two instructions.

This tidied the code considerably, and also gave little more execution speed. I tested couple of games, and so far there seem to be no script errors, so maybe I'm right in my assumption.

ivan-mogilko:
Rebased last two commits, only to cleanup some unused code.

ericoporto:
Hey, my "Don't give up the cat" game already ran reasonably well, but with the newer improvements I felt more confident to make a web version of it! 
https://eri0o.itch.io/dont-give-up-the-cat

ivan-mogilko:
Merged!

This PR results were unexpected, I've compared the latest version with the wcproto ported to 3.2.1 (I might push that to my game's repository later), and the fps is pretty close to 3.2.1 engine (maybe 5-7% worse). This is a massive improvement compared to where we were at 3.6.0 release.

PR #1991, opened 2023-04-29
Added FLAC audio support

AlanDrake:
Since we already had underlying compatibility, I decided to add the missing bits.

FLAC is a lossless format that would mainly act as a replacement for WAV. It may cut sizes by just roughly half, but it boasts excellent decoding performances.
Also, it loops perfectly and supports positioning by milliseconds.

It could be used in the future as an automatic optimization for WAV files.


Engine/ac/global_audio.cpp
```diff
@@ -131,6 +131,11 @@ static bool play_voice_clip_on_channel(const String &voice_name)
         speechmp3 = my_load_mp3(get_voice_over_assetpath(asset_name), false);
     }
 
+    if (speechmp3 == nullptr) {
+        asset_name.ReplaceMid(asset_name.GetLength() - 3, 3, "flac");
+        speechmp3 = my_load_mp3(get_voice_over_assetpath(asset_name), false);
```
ivan-mogilko, line 136: Should be my_load_flac?

Engine/plugin/agsplugin.cpp
```diff
@@ -575,6 +575,9 @@ void IAGSEngine::PlaySoundChannel (int32 channel, int32 soundType, int32 volume,
         newcha = my_load_midi(asset_name, (loop != 0)); break;
     case PSND_MOD:
         newcha = my_load_mod(asset_name, (loop != 0)); break;
+    case PSND_FLACSTREAM:
+    case PSND_FLACSTATIC:
+        newcha = my_load_ogg(asset_name, (loop != 0)); break;
```
ivan-mogilko, line 580: Should be my_load_flac?

AlanDrake: Yes, and yes. I'm sure I had renamed these correctly, but somehow...

AlanDrake:
Ok, I fixed those two typos and tested speech. It's working as expected on my end.

Engine/plugin/agsplugin.h
```diff
@@ -154,6 +154,8 @@ class IAGSEditor {
 #define PSND_OGGSTATIC  5
 #define PSND_MIDI       6
 #define PSND_MOD        7
+#define PSND_FLACSTREAM 8
+#define PSND_FLACSTATIC 9
```
ivan-mogilko, line 158: There's no need to make two separate constants here. The Static/Stream separation is remnant of old internal implementation, and should not have been a "type" ever in plugin api (possibly it should have been a separate parameter, but it's too late now).

Overall this type should not be even used in command, as engine should be trying to decode the sound using the extension as a hint first, and if that does not work, then trying all the available decoders.

AlanDrake: Good thing this is AGS4. It's not too late to do anything.
I'll add just one type then and forcepush again.

ericoporto:
@AlanDrake does the Editor build from the CI actually works? I was a bit worried I saw irkflac and I don't remember if we already download and archive that or not.

AlanDrake:
> @AlanDrake does the Editor build from the CI actually works? I was a bit worried I saw ikpflac.dll and I don't remember if we already download and archive that or not - is it part of the VSDependencies already?

Yep, it works. It was one of the files irklang come bundled with by default, so I just add the references to editor and installer.

PR #1992, opened 2023-05-01
Script: Readable UseRegionTint property for Character/Object

AlanDrake:
@ericoporto suggested this may be of use.
I did not propose it before because I wasn't sure whether this missing script property should be writable or not.

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -28,7 +28,7 @@ using namespace AGS; // FIXME later
 #define CHF_NOINTERACT      4
 #define CHF_NODIAGONAL      8
 #define CHF_ALWAYSIDLE      0x10
-#define CHF_NOLIGHTING      0x20
+#define CHF_NOLIGHTING      0x20  // TODO: rename this CHF_USEREGIONTINTS with opposite meaning
 #define CHF_NOTURNING       0x40
 #define CHF_NOWALKBEHINDS   0x80   // deprecated, forbidden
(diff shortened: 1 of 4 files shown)


AlanDrake:
Ah, I had never noticed there was already property for that on characters. I started adding it for objects and then got sloppy.

Yeah the naming completely sucks, also the manual mentions some bogus "ignore lighting" checkbox that hasn't existed nobody knows how long. It's called "UseRoomAreaLighting" and it's what one should use as a search term to find this property.

What do we do? Make it `IgnoreLighting` for both? Or maybe use the same term as the editor "UseRoomAreaLighting" in both cases?

ivan-mogilko:
A while ago I replaced Object.IgnoreScaling with Object.ManualScaling, to comply with Character's property, but also because many years ago someone told me it's better to use "positive" names than "negative" names (Enable vs Disable, Use vs DontUse, and so forth).

There's still a question to whether default should be using area parameters, or not using them. Depending on the choice here the properties should be called e.g. UseExplicitTint/Light or UseRegionTint/Light.

For a reference there's already readonly HasExplicitTint/HasExplicitLight properties.

AlanDrake:
Currently, the default is false for objects, but true fro characters. I think it's fine, because in most cases these are the settings you'd expect.
`UseExplicit` creates the expectation I've applied a Tint directly to the entity, I would not use that wording for the ability to inherit from regions.
Maybe it's better to use `UseRegionTint` for both Characters/Objects, because it states the intent clearly.

PR #1995, opened 2023-05-02
Editor: turn dpi aware on

ericoporto:
Turns dpi-aware on, this makes ags not blurred in Windows 10 by default. I couldn't find any side effect, so I ask people to test this. The text becomes sharp and it's much nicer to code like this.

With dpi-aware on
image
image

Without dpi-aware on (commented)
image
image


ericoporto:
Unfortunately there were scaling issues I didn't catch at the time, so this PR was undone later, the scaling issues are listed in #2087 .

PR #1999, opened 2023-05-08
Editor: feature log panel

### LINKED ISSUE #1362: Read engine Log in the Editor

ivan-mogilko:
Resolves #1362
Based on previous PR attempts by @ericoporto  #1518 and #1989, with fixes and additions.

In short, this features an engine's log output inside the Editor. The log is recorded in a fixed-size buffer, which makes this suitable for a **short-term** log, not a full log.

### From the Engine's side

1. Implements a new log output DebuggerLogOutputTarget with id "debugger" that sends messages through the IDE debugger interface. Similarly to other output types, this may be started and configured using ""--log-debugger=" command line arg.

### From the Editor's side
1. Added DebugLogComponent that creates a log panel.
2. Log panel features a rich text box for printing log.
3. The log is recorded in a LogBuffer which internally implements a circular buffer of a fixed size. This means that when the buffer is filled, the first entries will be discarded to allow space for new ones.
4. The log panel is configured by log filter settings, and uses the property grid on Properties panel to display and let user modify these. These settings are serialized in **Game.agf.user**.
5. The log panel's config consists of two parts. The base config lets set the minimal log levels, while the filter config affects the dynamic filtering in the log panel itself.

### Changes to original PRs
1. Optimized the buffer->panel synchronization, and panel's drawing. Let user scroll the text up while the log is running.
2. Added toolbar with command buttons: Run, Pause, Glue (defines whether textbox should scroll automatically as the new entries are added), Copy selection, Copy all, Clear.
3. Added "Font" and "FontSize" settings for the Log Panel in the Editor's Preferences.
4. Remade LogPanel into the DockContent subclass, allowing it to be present in the Window menu and serialize in Layout.xml.

### Known issues
1. There's one irritating issue that I did not manage to solve at this moment. Because I tried to make LogPanel not a ContentDocument, but a DockContent, similar to other "persistent" panels (Output pane , and so on), I had to use a separate method to assign its config to the property grid panel. I was expecting the other ContentDocuments would switch the properties on activation, and they normally do. But the problem is that some of the editor panels have some kind of "inactive areas" on them, and clicks on these do not resend "activate" event, it seems. In these cases the only way to switch properties is to click on either the Tab caption, or any editable field/area inside the panel. So, this is not critical, but may be annoying to users.
What is curious, if you have 2 editor panes positioned side-by-side (e.g. 2 characters), then clicking on same areas switch them normally. It's only switching from log panel to e.g. Character pane does not send activate for the second time.
I might keep searching for some good method of receiving the event.
2. Need icons for the 3 last toolbar commands.


ivan-mogilko:
> There's one irritating issue that I did not manage to solve at this moment. Because I tried to make LogPanel not a ContentDocument, but a DockContent, similar to other "persistent" panels (Output pane , and so on), I had to use a separate method to assign its config to the property grid panel. I was expecting the other ContentDocuments would switch the properties on activation, and they normally do. But the problem is that some of the editor panels have some kind of "inactive areas" on them, and clicks on these do not resend "activate" event, it seems. In these cases the only way to switch properties is to click on either the Tab caption, or any editable field/area inside the panel. So, this is not critical, but may be annoying to users.

So, after some consideration, I decided to try out a different approach, and not use the common Properties Panel at all.
Instead, I added a property grid right onto the LogPanel, on a hidden splitter pane. A new toolbutton titled "Show settings" expands or collapses this pane.

The benefits of this are:
* It does not conflict with the Properties Panel.
* Log settings don't override the Properties Panel contents, which may be somewhat more convenient if user wants to continuously fix some object's values and the re-run game tests with active log window. Or even work with the game properties while the game is running.
* Presumably, log settings is not something used too often, so they may stay hidden for most times.

I left the previous variant in a separate branch just in case: 
https://github.com/ivan-mogilko/ags-refactoring/commits/feature-360-console-editor-ipropertyprovider

ivan-mogilko:
Added missing toolbar icons; i'm not an icon artist, so painted them like I could...

@ericoporto please tell, if the way logpanel works now is not too ugly to you?

ericoporto:
It works and it looks alright! Good work!

One thing I noticed is by default you get the game window on top of the editor and clicking the editor will raise it's focus, so any click on a button in the log panel brings the editor forward. This means that you can't have the editor below the game, it has to be side by side or in a different monitor. So something to think about, NOT in this PR, would be being able to load the game as a panel inside AGS Editor as a possible solution to this, some way to attach the log panel to the game window instead, or just some way the editor doesn't move up the game window - perhaps launching the game window with some always on top setting.

ivan-mogilko:
Rebased, with fixes:
* removed unneeded code left from experiments;
* fixed "Layout - Reset to defaults" was breaking things (this commit actually may be later applied to release-3.6.0, because it's a minor bug there too)
* safety checks to ignore detached panes when choosing them in menu: that at least should prevent exceptions.

ericoporto:
I confirm the issue of the Layout - Reset to defaults is fixed on my end! 

I have a weird issue here, that may be computer specific (I am on macOS, using AGS Editor on Wine!): if I place a System.Log in the room After Fade In, the generated log messages are written in the RichTextBox with a black font (so it looks like nothing is written, unless you select it), any other log messages that are generated later are correctly written with white color.

ivan-mogilko:
> I have a weird issue here, that may be computer specific (I am on macOS, using AGS Editor on Wine!): if I place a System.Log in the room After Fade In, the generated log messages are written in the RichTextBox with a black font (so it looks like nothing is written, unless you select it), any other log messages that are generated later are correctly written with white color.

I don't see why the font color in editor will depend on which place in script was the log called in.
Could be you put some RTE commands inside the text? Please post an actual script command with text?

ericoporto:
ok, I confirmed it works fine on Windows! It's indeed a weird issue of the wine on macOS - no idea why, but yeah, the first things called.

PR #2006, opened 2023-05-26
Xcode: fixes to Xcode project

ericoporto:
- remove unused lib makefile
- matches Xcode project with current files
- require SDL2.Framework in `libsrc` instead of Xcode project for ease of use.

I have the same updates for the master branch, where there are minor differences in the AGS files in the Engine/ac directory, not sure if open it already as other PR or after merge have this on master and then the other goes after.

The SDL2.framework place change is because previously I thought macOS and iOS would use different SDL2.frameworks. They are, but macOS uses a prebuilt one (the released SDL2 aren't building, as framework, for macOS on the latest Xcode and OS version combination). On iOS, because of the different architectures and platforms, instead SDL2.framework will be built from source, so there are no conflicts with having both the framework and the sources of SDL2 under `libsrc` dir.

Also, I considered adding `libsrc/vorbis`, `libsrc/ogg`, `libsrc/theora`, `libsrc/SDL` and `libsrc/SDL_sound` to gitignore.


ivan-mogilko:
@ericoporto do I need to cherry pick this commit to master, or wait for another pr?

ericoporto:
@ivan-mogilko if you cherry pick it to master I will just add a smaller PR on top to adjust for the removal of statobj, would it be ok or do you prefer a new PR of everything?

ivan-mogilko:
Okay, I will do soon.

PR #2011, opened 2023-05-30
Script API: implemented ScriptName and Type.GetByName() for all applicable types.

ivan-mogilko:
This implements readonly ScriptName property and a static GetByName() method for all the object types that have script names. The method looks up for the object of corresponding type with the given script name, and returns its pointer, or null if such object is not found.

NOTE: passing other type's script name will also return null, for example calling Character.GetByName() with a room object's name.

`scriptName` argument is case-sensitive, obviously.

```
static AudioClip* AudioClip.GetByName(const string scriptName);
static Character* Character.GetByName(const string scriptName);
static Dialog* Dialog.GetByName(const string scriptName);
static GUIControl* GUIControl.GetByName(const string scriptName);
static GUI* GUI.GetByName(const string scriptName);
static Hotspot* Hotspot.GetByName(const string scriptName);
static InventoryItem* InventoryItem.GetByName(const string scriptName);
static Object* Object.GetByName(const string scriptName);
```

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -913,12 +913,16 @@ builtin managed struct InventoryItem {
   import attribute int  Graphic;
   /// Gets the ID number of the inventory item.
   readonly import attribute int ID;
-  /// Gets/sets the name of the inventory item.
+  /// Gets/sets the human-readable name of the inventory item.
   import attribute String Name;
   /// Sets an integer custom property for this item.
   import bool SetProperty(const string property, int value);
   /// Sets a text custom property for this item.
   import bool SetTextProperty(const string property, const string value);
+#ifdef SCRIPT_API_v399
+  /// Gets the script name of the inventory item.
+  import readonly attribute String ScriptName;
+#endif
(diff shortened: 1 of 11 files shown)


ivan-mogilko:
Something for a distant future, but in a conversation we had with @AlanDrake , I realized this or similar methods may be used to initialize global variables in case we completely move to dynamic creation of everything.

Imagine the game data was loaded, and objects defined at design-time created, but how do you attach them to the variables to keep things consistent with objects created at runtime? Possibly that's how:
```
function game_start() { // or another "service init" function
    cEgo = Character.FindByName("cEgo");
    cNPC = Character.FindByName("cNPC");
}
```

ericoporto:
I think `GetByName` is a better name than `FindByName` for this behavior.

ivan-mogilko:
Renamed FindByName -> GetByName.

ivan-mogilko:
In regards to getting script names from these objects; I've been thinking of quickly adding a readable ScriptName property to all of them, but then thought to instead delay that, and experiment with creating the most basic parent class which would hold this property, thus avoiding duplicate API.

ivan-mogilko:
Note: removed Region.GetByName, realizing that regions don't have script names for some reason...

PR #2012, opened 2023-05-30
Script API: implement eEventGameSaved

### LINKED ISSUE #1295: eEventSavedGame to go with existing eEventRestoreGame
This is a feature request.
In the current editor, the command RestoreGameSlot() will not actually execute until the function enclosing it is finished; this makes it impossible to ever run a command *after* the game is restored, except by using a variety of true/false switches in repeatedly_execute_always(), etc. etc.
LUCKILY, there is an event trigged for this which solves the problem: eEventRestoreGame.
HOWEVER, there is no equivalent event for a SaveGameSlot() command, which has the same features and won't run until the enclosing function has finished, etc.

To Reproduce
Run RestoreGameSlot()
Set an event capture in on_event for eEventRestoreGame - this will run after the game restore is finished completely

Save a game with SaveGameSlot()
You can no longer figure out when this save function is done
hence:
  Set a switch for the savegame file name, and check for the file existing in repeatedly_execute_always()
  When the file exists, if ever, set and start a timer for something which seems reasonable to write the file completely, say 1sec
  Now you're as sure as you can be that it's saved and complete, set a 'JustSavedGame' switch to true, pick it up in repeatedly_execute_always()  
  This time, set JustSavedGame to false, do what you need to do following the SaveGameSlot() function
  You are still not 100% sure it all worked, but as long as the computer doesn't hang from some other process, or some other case, we're 99% sure we guess :/

ivan-mogilko:
Resolves #1295

By a user's request, this adds a `eEventSaveGame` event which is run after a game was saved.
Similar to other events, this is scheduled to run after the current script callback is completed.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1130,6 +1130,7 @@ enum EventType {
 #endif
 #ifdef SCRIPT_API_v399
   eEventLeaveRoomAfterFadeout = 11,
+  eEventGameSaved = 12,
 #endif
 };
 
--- a/Engine/ac/event.h
+++ b/Engine/ac/event.h
@@ -33,6 +33,7 @@
 #define GE_RESTORE_GAME  9
 #define GE_ENTER_ROOM_AFTERFADE 10
 #define GE_LEAVE_ROOM_AFTERFADE 11
+#define GE_SAVE_GAME     12
 
 // Game event types:
 // common script callback
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -919,6 +919,8 @@ void save_game(int slotn, const char*descript) {
 
     // Save dynamic game data
     SaveGameState(out.get());
+    // call "After Save" event callback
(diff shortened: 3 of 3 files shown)


ericoporto:
I think`eEventGameSaved` sounds better for this - I also think the other even would be better named as `eEventGameRestored`.

ivan-mogilko:
> I think`eEventGameSaved` sounds better for this - I also think the other event would be better named as `eEventGameRestored`.

I can rename it here in ags4, but i've also been thinking about backporting to 3.6.1, since it's a simple change, and I'll have to keep the names there to avoid annoying users.

ivan-mogilko:
> I think`eEventGameSaved` sounds better for this - I also think the other event would be better named as `eEventGameRestored`.

This follows the events naming, if you look at their list they all have this form:

```
eEventLeaveRoom*
eEventEnterRoom*
eEventGotScore
eEventAddInventory
eEventLoseInventory
```

so it's verb-noun. If we begin to fix this we'd have to change almost every event name for consistency.

ericoporto:
GotScore is definitely the better name. At least if adding one is better to be eEventGameSaved because it naturally sounds like it happens after. If one day #1618 is implemented we can use the same event.

PR #2014, opened 2023-05-31
Editor: reimplement SpriteSelector's multiple sprite properties editing

### LINKED ISSUE #2008: Editor: in Sprite Manager allow to modify properties of multiple selected sprites
Before the new sprite import was implemented in AGS 3.5.0, the sprite manager's property grid let users to select multiple sprites and the property grid would display only matching values for all of these sprites, and show others as a blank space. If a property's value was changed, it would apply the new value to all the selected sprites.

After the sprite import and import dialog were reimplemented in AGS 3.5.0, this is no longer supported.

As user proposed, this ability should be restored.

When multiple sprites selected:
* Property grid displays only matching values, and displays blank space in other fields where values don't match.
* Changing a value of a property would affect all the selected sprites at once.

SIDE NOTE: I recall similar thing was requested for the View Frames, but I could not find any open ticket here.
(also selecting multiple frames is not possible as of 3.6.0)

ivan-mogilko:
Resolves #2008

Originally the Sprite Manager allowed to edit properties of a multiple items at once. This was disabled (unintended) by a performance fix: ff02202

This commit re-enables multiple sprites set to the property grid, and additionally improve performance in case of a large selection range. The performance fix consists of two parts:
1) ~Remember selected sprites in a HashSet and only add/remove on each ItemSelectionChanged call.~
2) Don't invoke our OnSelectionChanged event on each item processing. Instead schedule it using a funny trick with Application.Idle subscription, and thus make it trigger only after whole range of (de)selected items was processed.

The Application.Idle trick was found here:
https://stackoverflow.com/questions/1191920/net-listview-event-after-changing-selection


ivan-mogilko:
After some tests, I removed the HashSet part of the performance fix, at least temporarily, because it seems unreliable: there are few cases where the selection is lost, or changed programmatically, which leaves the HashSet unsynced with the listview.

This dropped the performance again. I will look into other options now.

Also, I should probably retarget this to 3.6.1, as the change is somewhat risky, and I don't want to potentially break something in a 3.6.0 patch.

ivan-mogilko:
> This dropped the performance again. I will look into other options now.

Ah, no, that was because of my mistake, fixed now.

Also, I tried to make things slightly more faster by storing Sprite reference in the ListViewItem's tag. This saves the text parsing and a call to FindSpriteByNumber.

@morganwillcock , if you remember how the sprite selector works, can you tell if there are any risks there? The Sprite objects seem to be stored in SpriteFolder object, which is only updated on load from reading xml, and not strictly connected to the native sprite storage.

morganwillcock:
I don't remember very much, but I would guess it is safe as long as all of these actions are blocking and the idle callback is guaranteed to run immediately afterwards. Maybe there would be a problem if you could delay the idle callback by holding down or pressing keys (possibly repeat rate would be relevant) whilst dragging the selection, and modify the selection before the callback has run.

ivan-mogilko:
Well, I tested and it seems to work fine overall.

PR #2023, opened 2023-06-03
Adjust automatic layout of some panes (for 100% and 200% scaling)

ericoporto:
This should adjust the layout of the following panes so they work in both 100% and 200% scaling

- SpriteSelector
- InventoryEditor

### DIFF
--- a/Editor/AGS.Editor/Panes/InventoryEditor.Designer.cs
+++ b/Editor/AGS.Editor/Panes/InventoryEditor.Designer.cs
@@ -30,26 +30,30 @@ namespace AGS.Editor
         {
             this.currentItemGroupBox = new System.Windows.Forms.GroupBox();
             this.splitContainer1 = new System.Windows.Forms.SplitContainer();
+            this.flowLayoutPanel2 = new System.Windows.Forms.FlowLayoutPanel();
+            this.flowLayoutPanel1 = new System.Windows.Forms.FlowLayoutPanel();
             this.label1 = new System.Windows.Forms.Label();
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Testing this on Windows 7:
1. SpriteSelector and InventoryEditor has minor layout changes, but I had to compare 2 editors to notice them.
2. LipSync editor:
* the distance between number of text fields widened up noticeably, but maybe this is not a critical thing. (TBH I feels like the whole panel should rather be redesigned with some kind of a listview or table control)
* more importantly, the textboxes are now disregarding the "Type = Disabled" property setting, and are always enabled.
3. Palette/Color panel looks quite disfigured, everything is larger, and for some reason gained more width than height, so horizontal spaces between controls are too wide. The label location changed from being on top to being on bottom. See screenshots below.

Not sure what of this is intended.

editor-gui-scaling
editor-gui-scaling2

ericoporto:
The palette and the lipsync are generated on the fly, on creation. I think instead they need to be remade in some other way, but I am not sure how yet. I can drop these until further time and just let the sprite selector and inventory window in as is - their changes are rather minimal.

ivan-mogilko:
@ericoporto could you please open a task ticket for remaking the panes that look bad in scaled mode, and explain what has to be done with these? I don't think this is documented anywhere.

Another question: what about Editor plugins, are these left to be fixed by their own maintainers?

ericoporto:
I will revise the panes to see which ones specifically break and open an issue with the list.

ericoporto:
Added the revised list as promised in the above comment in #2087

Issue #1442, opened 2021-10-30
AGS4: Documentation for building the Engine : Can't reproduce steps to create SDL_Sound


fernewelten:
**Describe the bug**
I'm trying to follow the instructions in file `ags/Windows/README.md` to compile the Engine for Windows 32bit.
I can't reproduce the steps:
* I've downloaded SDL2 libraries from archive `SDL2-devel-2.0.16-VC.zip` as instructed in line 42, 
* I've downloaded and unpacked the sound source archive `997e90562b35.tar.gz` as instructed in line 58.
* So I open a VS developer prompt and `cd` into the source archive (the directory where CMakeLists.txt is located) and `cmake .`. This is how I interpret the first part of line 60 (quoted below), or am I supposed to do something fundamentally different here? 
> After you downloaded the source this way or another, use CMake to build MSVS solution from their provided CMakeList.txt, […]
* The cmake run fails with the following error:
```
CMake Error at CMakeLists.txt:32 (find_package):
  By not providing "FindSDL2.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SDL2", but
  CMake did not find one.

  Could not find a package configuration file provided by "SDL2" with any of
  the following names:

    SDL2Config.cmake
    sdl2-config.cmake

  Add the installation prefix of "SDL2" to CMAKE_PREFIX_PATH or set
  "SDL2_DIR" to a directory containing one of the above files.  If "SDL2"
  provides a separate development package or SDK, be sure it has been
  installed.
```

Files named `SDL2Config.cmake` or `sdl2-config.cmake` aren't provided anywhere in the archives that I've downloaded so far. So I can't make them available to `cmake`. 

Some step seems to be missing here. What am I supposed to do to get `cmake` to run successfully?

fernewelten:
So here's what I did to get at the built library `SDL_sound.lib`. 
The bold parts were unexpected for me.
This, or a shortened version, might help expand the instructions in file `ags/Windows/README.md` a little.

* Download and unpack SDL_sound-da0cecd33a39a16a84fb860e538e76e270589167.tar.gz 
* In the Windows Explorer, open the unpacked folder
* Within this unpacked folder, create the subfolders
	* include
	* lib
	* MSVC
* __Open the file `CMakeLists.txt` in the unpacked folder and insert a `#` at the start of line 32;__ it should now read:
> #find_package(SDL2 REQUIRED)
[...]

PR #2024, opened 2023-06-04
Windows: remove SDL2_sound-static.lib renaming

ericoporto:
Adjust for https://github.com/adventuregamestudio/ags/issues/1442#issuecomment-1575571683

Better wait the CI to build just to make sure


ivan-mogilko:
There's a potential issue, our readme points to the WinDevDependenciesVS.zip, which has libs with old name.
We would need to make another release to let new deps archive uploaded, and then edit readme to point there.
For simplicity, maybe merge this after a new beta release is out (I may make one right after merging #2023 (if it's working).

ericoporto:
It will only be "wrong" for a single release, I think it should be fine - I imagine if someone find themselves in this case they will simply ask through an issue or something.

PR #2028, opened 2023-06-07
Editor: option to display sprite's filenames in the SpriteSelector

### LINKED ISSUE #563: Sprites: show SourceFile name aligned to the right, so you can instantly see the name of the file and the extension
hovering over the SourceFileName shows the complete path, but mostly I'm interested in the file name and the bigger the project gets chances are the file is deep in some folder structure where the name is pushed very far to the right an invisible when quickly looking there. IT IS VISIBLE when hovering over the SourceFileName field, but when the displayed text (path) is aligned to the right inside that field the filename is always visible at once.
also aligning the hovering display to the right end of the field too should/would not display outside of the editor window when using it in windowed (and not fullscreen)

Same for audio files. Though not as relevant as the name is mostly the same as the ingame name.

ivan-mogilko:
Possibly fixes #563 (need clarification).

This adds a checkable context menu option (View -> Show filenames) into the Sprite Selector window. When checked, the sprites will have their filename (without preceding paths) displayed along with their numbers.

It looks roughly like this:
sprites-with-names

### DIFF
--- a/Editor/AGS.Editor/Panes/SpriteSelector.cs
+++ b/Editor/AGS.Editor/Panes/SpriteSelector.cs
@@ -58,6 +58,7 @@ namespace AGS.Editor
         private const string MENU_ITEM_PREVIEW_SIZE_2X = "PreviewSizeMedium";
         private const string MENU_ITEM_PREVIEW_SIZE_3X = "PreviewSizeLarge";
         private const string MENU_ITEM_PREVIEW_SIZE_4X = "PreviewSizeExtraLarge";
+        private const string MENU_ITEM_PREVIEW_TEXT_FILENAME = "PreviewTextFilename";
         
         private const int SPRITE_BASE_SIZE = 32;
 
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@selmiak , please tell, does this work as a solution for seeing sprite's filenames easier? (#563)

selmiak:
yes it does, plus see my remark about scaling down those images and my prefered behaviour there, but I see that at some really small scale compromises need to be made :)

ivan-mogilko:
> plus see my remark about scaling down those images and my prefered behaviour there

as I mentioned in another place, we don't have full control over how the text is drawn, so unless we code the drawing AND image arrangement ourselves, the text will be displayed as it is.

ivan-mogilko:
I will wait a little more in case anybody else would like to comment on this PR.

selmiak:
sure. some more activity by me posting here again will show the activity and maybe encourage someone else to give some input :)

AlanDrake:
it works and can be toggled. I think it can be merged.

PR #2029, opened 2023-06-07
Fixes non-building CI

ericoporto:
just minor updates to see if they are related to the CI outage. In draft until the support answers on the issues.

### DIFF
--- a/ci/android/Dockerfile
+++ b/ci/android/Dockerfile
@@ -1,4 +1,4 @@
-FROM cirrusci/android-sdk:33-ndk
+FROM ghcr.io/cirruslabs/android-sdk:33-ndk
 # CirrusCI Android NDK image uses Ubuntu 22.04
 # ANDROID_NDK_VERSION=25.1.8937393
 # ANDROID_BUILD_TOOLS_VERSION=33.0.1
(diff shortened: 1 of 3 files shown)


ericoporto:
Apparently the issue is a partial garbage collected image, see here: https://github.com/cirruslabs/cirrus-ci-docs/issues/1174#issuecomment-1580928673

I don't know yet exactly what action I have to do to fix the failing builds

ericoporto:
Force pushed a new commit with a comment added in the Emscripten Docker image.

As now we have build EMSDK to then be able to build things using the EMSDK (even in the docker image itself) - this makes the EMSDK a bit larger, so fetching it for use in our pipeline will take a bit more now.

We do this in the Linux Docker too, but mostly because we use a very outdated Linux distro for glibc compatibility ([mostly because of steam](https://gitlab.steamos.cloud/steamrt/scout/sdk/-/blob/steamrt/scout/README.md#developing-software-that-runs-in-scout)).

While it's somewhat convenient that using the Cirrus docker cache we don't have to worry about managing the docker packages, I think we should think about switching to use GitHub Container Registry for our CI build environments, as a lot of Cirrus issues comes from their docker cache. This would also make it easier to move to GitHub Actions, as it has no docker cache, and would require a container registry anyway.

Issue #2019, opened 2023-06-01
New compiler: parsing mistake involving incorrect increment/decrement operators


ivan-mogilko:
CC @fernewelten 
Consider script:

```
int a = 10;
int b = 5;
int c = a ++ b;
    
Display("a = %d, b = %d, c = %d", a, b, c);
```

The output is:
> a = 10, b = 5, c = 13 (???)

What is expected is probably a script error, as the above statement does not have a correct interpretation.

fernewelten:
That's a bug. `++` is mis-parsed as a binary operator. I'll look into it.

fernewelten:
Addressed in #2038.

ivan-mogilko:
Fixed by #2038

PR #2038, opened 2023-06-20
New Compiler: Fix bug: '++' and '--' can't be binary op

fernewelten:
This addresses #2019.

This PR makes the compiler throw an error message for expressions that use `++` or `--` as binary operators. These operators can only be used as unary-prefix or unary-postfix operators.

Typical code:
```
function game_start() 
{
            int a = 10;
            int b = 5;
            int c = a ++ b;  \\ ← Compiler will balk here
}
```


ivan-mogilko:
@fernewelten please rebase on the latest ags4 once more, there were problems with the build server, and last commit fixes it.

PR #2041, opened 2023-06-21
Engine: remember overlay index and try it before looking again

ericoporto:
Whenever we access an Overlay, we iterate through all overlays until we find the one with the corresponding ID. This has been this way since the origins of AGS, as you can see in [in ac_overlay.cpp in the initial commit](https://github.com/adventuregamestudio/ags/blob/bb1b708965a35da0853878ac4b670234ddeb5b17/Engine/acmain/ac_overlay.cpp).

Overlays touch a lot of things and I am sure there is some smarter way to redo things so we can just access directly what we mean when using the Script Overlays. Still, a minor easy to do optimization is to simply remember the last result and use it, if the result has changed, well then look for it again. This improves FPS in my example game below when running in infinite fps mode (I go from 66fps to 82fps in the Oasis area) and can also reduce hiccups in it. (I didn't prepared a special game, so you need to wait 811 game loops for it to finish loading before using it).

[ExampleGameOverlays.zip](https://github.com/adventuregamestudio/ags/files/11811311/ExampleGameOverlays.zip)

This is a naïve fix for #2042 , but performance wise it suffices, even though the code may not be pretty.

### DIFF
--- a/Engine/ac/overlay.cpp
+++ b/Engine/ac/overlay.cpp
@@ -44,6 +44,7 @@ extern IGraphicsDriver *gfxDriver;
 
 
 std::vector<ScreenOverlay> screenover;
+std::vector<int> overlookup(128);
 
 void Overlay_Remove(ScriptOverlay *sco) {
     sco->Remove();
@@ -404,9 +405,16 @@ void remove_screen_overlay(int type)
 
 int find_overlay_of_type(int type)
 {
-    for (size_t i = 0; i < screenover.size(); ++i)
+    int idx = overlookup[type];
+    if (idx >= 0 && idx < screenover.size() && screenover[idx].type == type)
+        return idx;
+    for (idx = 0; idx < screenover.size(); ++idx)
     {
-        if (screenover[i].type == type) return i;
+        if (screenover[idx].type == type)
+        {
+            overlookup[type] = idx;
+            return idx;
+        }
     }
     return -1;
 }
@@ -414,6 +422,9 @@ int find_overlay_of_type(int type)
 size_t add_screen_overlay_impl(bool roomlayer, int x, int y, int type, int sprnum, Bitmap *piccy,
     int pic_offx, int pic_offy, bool has_alpha)
 {
+    const auto new_size = screenover.size() + OVER_CUSTOM + 2;
+    if(overlookup.size() < new_size) overlookup.resize(new_size);
+
     if (type == OVER_CUSTOM) {
         // find an unused custom ID; TODO: find a better approach!
         for (int id = OVER_CUSTOM + 1; (size_t)id <= screenover.size() + OVER_CUSTOM + 1; ++id) {
@@ -461,6 +472,7 @@ size_t add_screen_overlay_impl(bool roomlayer, int x, int y, int type, int sprnu
         play.speech_face_schandle = over.associatedOverlayHandle;
     }
     over.MarkChanged();
+    overlookup[type] = screenover.size();
     screenover.push_back(std::move(over));
     return screenover.size() - 1;
 }
@@ -527,8 +539,11 @@ Point get_overlay_position(const ScreenOverlay &over)
 
 void recreate_overlay_ddbs()
 {
-    for (auto &over : screenover)
+    overlookup.resize(screenover.size() + OVER_CUSTOM + 2);
+    for (size_t i=0; i<screenover.size(); i++)
     {
+        auto &over = screenover[i];
+        overlookup[over.type] = i;
         if (over.ddb)
             gfxDriver->DestroyDDB(over.ddb);
         over.ddb = nullptr; // is generated during first draw pass


ivan-mogilko:
Have you considered a (hash)map instead? IIRC the overlay's index is one of the kind, so should work as a key.

ericoporto:
there is a lot of direct access to the `screenover` vector (in character, object and others), I was trying to decipher but I decided to test if the dumbest thing I could come up with would be alright and it gave me 24% FPS improvement already. I am sure a better approach exists, but the effort <-> performance tradeoff is much smaller.

ivan-mogilko:
In regards to current impl, are you sure valid ovrIdx begin with 1, and 0 is not a valid overlay's index?

ericoporto:
> find_overlay_of_type starts searching with 0, so it may return 0 as a valid index theoretically.

That was an oversight! Fixed!

ivan-mogilko:
> there is a lot of direct access to the `screenover` vector (in character, object and others), I was trying to decipher but I decided to test if the dumbest thing I could come up with would be alright and it gave me 24% FPS improvement already. I am sure a better approach exists, but the effort <-> performance tradeoff is much smaller.

There's still a potential performance problem with adding/erasing overlays dynamically; I posted a [comment explainng this](https://github.com/adventuregamestudio/ags/issues/2042#issuecomment-1601080753) in the problem ticket.
This may not be a problem in your game, if you're using a "overlay pool", but could be a problem if a game would create/delete lots of overlays often.

I may merge this pr as it is (assuming it works correct), but I am inclined to update overlays storage further afterwards.

Engine/ac/dynobj/scriptoverlay.h
```diff
@@ -18,6 +18,7 @@
 
 struct ScriptOverlay final : AGSCCDynamicObject {
     int overlayId = -1;
+    int ovrIdx = 0;
```
ivan-mogilko, line 21: Default value should be -1 here, as valid storage indexes begin with 0.

ericoporto: Adjusted!

ivan-mogilko:
I think it may be nice to also assign this index when the ScriptOverlay is created. Here:

https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/overlay.cpp#L316C5-L319

The problem is that the overlay's index is not a part of this function's args, and it's no longer returned from functions like `Overlay_CreateGraphicCore`; a ScreenOverlay ref is returned instead.

This may be worked around temporarily by storing an index inside ScreenOverlay too, when it's allocated.

Secondly, you might also reset the index when the ScriptOverlay gets invalidated here as well, just in case:
https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/overlay.cpp#L337-L338

ivan-mogilko:
Other places where the `find_overlay_of_type` may be replaced with your new function `find_overlay_of_type_for_script` are inside ScriptOverlay's Dispose and Remove methods:

https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/dynobj/scriptoverlay.cpp#L30-L34

https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/dynobj/scriptoverlay.cpp#L74-L80

ivan-mogilko:
One more thing, there's 1 overlay property that misses using `find_overlay_of_type_for_script` too:

https://github.com/ericoporto/ags/blob/b95a977a351cbe7affc5907532463c4885cd44dd/Engine/ac/overlay.cpp#L159-L164

ivan-mogilko:
In regards to game saves, currently the ScriptOverlays don't store this storage index; and that may be a right thing.
But this also means that they will loose the fast connection to overlays when restoring a save.

If you care about this, I think it may be possible to do a "reconnection" pass somewhere in [`DoAfterRestore` function](https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/game/savegame.cpp#L466).

The algorithm would be something like this:
```
for (int idx = 0; idx < screenover.size(); ++idx)
{
     if (screenover[idx].associatedOverlayHandle > 0)
     {
          ScriptOverlay *scover = (ScriptOverlay*)ccGetObjectAddressFromHandle(over.associatedOverlayHandle);
          if (scover)
              scover->ovrIdx= idx;
     }
}
```

(NOT TESTED!!)

ivan-mogilko:
I'm sorry, I did not figure out this earlier, but I might mention this now.

There's an alternate approach, it would be to not store index in ScriptOverlay, and instead have a lookup map along the screenover vector, that would have an overlay's id (aka type) as a key and vector index as a value. Then, instead of storing extra in ScriptOverlay, and, for example, try to find out overlay's index when Script object is created, or restored after a loading a save, you would keep this record in this lookup map.

The `find_overlay_of_type` then would look something like this:

```
int find_overlay_of_type(int type)
{
    int idx = overlookup[type];
    if (idx >= 0 && idx < screenover.size() && screenover[idx].type == type)
        return idx;
    for (idx = 0; idx < screenover.size(); ++idx)
    {
        if (screenover[idx].type == type)
        {
             overlookup[type] = idx;
             return idx;
        }
    }
    return -1;
}
```

ericoporto:
> There's an alternate approach, it would be to not store index in ScriptOverlay, and instead have a lookup map along the screenover vector

So assuming this is a size_t to size_t or int to int map, is this like a vector of some specific size? Like say screenover.size()+100. Just, trying to figure it out what container to use here.

ivan-mogilko:
> So assuming this is a size_t to size_t or int to int map, is this like a vector of some specific size? Like say screenover.size()+100. Just, trying to figure it out what container to use here.

Hm, I was thinking about unordered_map<int,int> or <int,size_t>, but given the indexes are sequential this may also be a vector, except you'd need to take care of that 100 offset.

> In the other way the ScriptOverlay is already sort of in memory, so accessing a property from it should be alright, in this other way we jump to look to a different memory. It's different, but I am not sure it's better.

It will be easier to manage, keeping everything close in one place, and also easier to replace later if the screenover vector is changed to something else.

ericoporto:
I need to think about this as this is a very different approach.

ivan-mogilko:
> I need to think about this as this is a very different approach.

It's not too different, it has practically same idea of storing index, the difference is in where it is stored, so to speak.

ivan-mogilko:
> I need to think about this as this is a very different approach.

I would not say that it's *very* different; this is similar in nutshell, remembering the lookup index (or "guess index" as you called it), except not in the script object, but in a lookup table stored alongside the overlays container.
On a conceptual side, it removes a need for ScriptOverlay to have a "knowledge" about how the screenover works, as a container.

ericoporto:
Sorry, it took me a long time to understand what you meant and wrote, I didn't understand that you explicit wrote working code and all, here it is.

I kept the previous branch under a different name here: https://github.com/ericoporto/ags/tree/overlay-dumb-opt-2

ivan-mogilko:
> I used a vector and placed a resize there in the only place I think can demand this index to be bigger.

Hmm, tbh this looks strange; I would instead propose to resize lookup along with the screenover.
This of course will require to adjust the check in `find_overlay_of_type`.

Besides that, you may sort of copy the initial lookup assignment from the old version, except it may be done easier now:
1. When overlay is created, it's pushed to vector, and so you may create a lookup for it
```
if (overlookup.size() <= type)
    overlookup.resize(type + 1);
overlookup[type] = screenover.size();
screenover.push_back(std::move(over));
```
2. When overlays are read from the save, e.g. in recreate_overlay_ddbs
```
overlookup.resize(screenover.size() + OVER_CUSTOM + 1);
for (size_t i=0; i<screenover.size(); i++)
{
     auto &over = screenover[i];
     overlookup[over.type] = i;
     ... etc....
}
```
I'm **not** absolutely certain about correctness of resize in this code though, just typed the idea of how it could be.

Engine/ac/overlay.cpp
```diff
@@ -417,6 +425,7 @@ size_t add_screen_overlay_impl(bool roomlayer, int x, int y, int type, int sprnu
     if (type == OVER_CUSTOM) {
         // find an unused custom ID; TODO: find a better approach!
         for (int id = OVER_CUSTOM + 1; (size_t)id <= screenover.size() + OVER_CUSTOM + 1; ++id) {
+            if(id >= overlookup.size()) overlookup.resize(overlookup.size()*2);
```
ivan-mogilko, line 428: Please disregard what I said earlier in the ticket thread, you having this *almost* in the right place.
But i think you may move it above the loop, and resize to `screenover.size() + OVER_CUSTOM + 2`, this should be the most it needs.

ericoporto: pushed a change for this. I kept, but adapted the if clause. It looks like it's working. I still don't know if I need to do something with `recreate_overlay_ddbs`, but I tested loading a save in don't give up the cat as is and it still worked, even after starting the game again.

ivan-mogilko: > but I tested loading a save in don't give up the cat as is and it still worked, even after starting the game again.

That is not obligatory, but might speed things up immediately after restoring a save, because the lookup will likely contain outdated indexes, so they all will have to be adjusted dynamically as script accesses the properties.

ericoporto: Hey, I gave it a shot [again in a new change](https://github.com/adventuregamestudio/ags/compare/7bc128f4798210884fa3dc1a247ce9cfd132d5ce..8fffaf0b6f844b946dc4bb1ad92c8dd14294fead), it works alright, do you believe it's correct?

ivan-mogilko: Yes it looks correct, but I will be able to test only later, after some time.

Also, I said it wrong again, that this is "not obligatory": you must do the "resize" thing after restore anyway, because the save could contain more overlays than before the save. It probably did not make any difference for your game because you are using a fixed size pool of overlays.

ericoporto: No rush, when you have time is alright!

PR #2044, opened 2023-06-22
Extend Right-to-left text render for GUI and non-wrapping DrawString

ivan-mogilko:
I do not know the reasons, but historically AGS only reverses messages (incl speech), labels and wrapped drawn strings in RTL mode.

The purpose of this PR is to expand RTL render to other controls and text drawing operations:
* [x] Buttons,
* [x] List Boxes
* [x] Text Boxes(?)
* [x] Single string drawing command.
* [ ] Anything else I forgot about(?).

Also to decide:
* [x] ~Do we need a per-control setting that tells whether to apply RTL render?~ Alternatively, all GUI controls have a hidden "translated" flag, which is set by default, although only ListBox has a visible property in the editor (because list boxes may be used to store saves and other filenames, which must not be translated). We could make this flag tell whether to apply RTL as well.
* [ ] How to deal with a situation when original text has one RTL mode, translation has another, but is missing an entry. Should the original text be drawn in original mode instead? For example: English game with Hebrew translation. Currently RTL is applied unconditionally. Should it be applied according to whether the text is original or got from translation?
* [x] How to deal with backwards compatibility. If any rtl translation was created earlier, it must have accounted for this original behavior. If the behavior is changed now, that might break these translations.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -112,6 +112,8 @@ New font load flags, control backward compatible font behavior
 Idle animation speed, modifiable hotspot names, fixed video frame
 3.6.0.21:
 Some adjustments to gui text alignment.
+3.6.1:
+In RTL mode all text is reversed, not only wrappable (labels etc).
 
 */
 
@@ -150,7 +152,8 @@ enum GameDataVersion
     kGameVersion_360_11         = 3060011,
     kGameVersion_360_16         = 3060016,
     kGameVersion_360_21         = 3060021,
-    kGameVersion_Current        = kGameVersion_360_21
+    kGameVersion_361            = 3060100,
+    kGameVersion_Current        = kGameVersion_361
 };
 
 // Data format version of the loaded game
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -23,6 +23,8 @@
 #include "core/assetmanager.h"
 #include "debug/out.h"
 #include "game/main_game_file.h"
+#include "gui/guibutton.h"
+#include "gui/guilabel.h"
 #include "gui/guimain.h"
 #include "script/cc_common.h"
 #include "util/alignedstream.h"
@@ -656,6 +658,18 @@ void UpgradeCharacters(GameSetupStruct &game, GameDataVersion data_ver)
     }
 }
 
+void UpgradeGUI(GameSetupStruct &game, GameDataVersion data_ver)
+{
+    // Previously, Buttons and Labels had a fixed Translated behavior
+    if (data_ver < kGameVersion_361)
+    {
+        for (auto &btn : guibuts)
+            btn.SetTranslated(true); // always translated
+        for (auto &lbl : guilabels)
+            lbl.SetTranslated(true); // always translated
+    }
+}
+
 void UpgradeMouseCursors(GameSetupStruct &game, GameDataVersion data_ver)
 {
     if (data_ver <= kGameVersion_272)
@@ -921,6 +935,7 @@ HGameFileError UpdateGameData(LoadedGameEntities &ents, GameDataVersion data_ver
     UpgradeFonts(game, data_ver);
     UpgradeAudio(game, ents, data_ver);
     UpgradeCharacters(game, data_ver);
+    UpgradeGUI(game, data_ver);
     UpgradeMouseCursors(game, data_ver);
     SetDefaultGlobalMessages(game);
     // Global talking animation speed
--- a/Common/gui/guibutton.cpp
+++ b/Common/gui/guibutton.cpp
@@ -316,8 +316,6 @@ void GUIButton::ReadFromFile(Stream *in, GuiVersion gui_version)
     if (TextColor == 0)
         TextColor = 16;
     CurrentImage = Image;
-    // All buttons are translated at the moment
-    Flags |= kGUICtrl_Translated;
 }
 
 void GUIButton::ReadFromSavegame(Stream *in, GuiSvgVersion svg_ver)
--- a/Common/gui/guidefines.h
+++ b/Common/gui/guidefines.h
@@ -140,7 +140,8 @@ enum GUIControlFlags
     kGUICtrl_Translated = 0x0080, // 3.3.0.1132
     kGUICtrl_Deleted    = 0x8000, // unused (probably remains from the old editor?)
 
-    kGUICtrl_DefFlags   = kGUICtrl_Enabled | kGUICtrl_Visible | kGUICtrl_Clickable,
+    kGUICtrl_DefFlags   = kGUICtrl_Enabled | kGUICtrl_Visible | kGUICtrl_Clickable |
+                          kGUICtrl_Translated,
     // flags that had inverse meaning in old formats
     kGUICtrl_OldFmtXorMask = kGUICtrl_Enabled | kGUICtrl_Visible | kGUICtrl_Clickable
 };
--- a/Common/gui/guilabel.cpp
+++ b/Common/gui/guilabel.cpp
@@ -145,8 +145,6 @@ void GUILabel::ReadFromFile(Stream *in, GuiVersion gui_version)
 
     if (TextColor == 0)
         TextColor = 16;
-    // All labels are translated at the moment
-    Flags |= kGUICtrl_Translated;
 
     _textMacro = GUI::FindLabelMacros(Text);
 }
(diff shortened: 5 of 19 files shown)


messengerbag:
For the other decisions, I still think it should be left as-is, and the focus should be to get full support for proper RTL (Unicode) text rendering instead of extending this text-reverse hack. If we had that, these issues would no longer be relevant, as displaying RTL text would no longer rely on the RTL setting in AGS, but would depend on the content of each String.

If I have it correctly, there are two parts to this that are currently missing:

1. Support ligatures, where particular combinations of characters appear differently than each of them separately. (Not strictly anything directly to do with RTL, but necessary for languages that use Arabic script because it requires ligatures in order to render correctly.)
2. Support full bidirectional layout (switching between LTR and RTL within a single String).

According to @ericoporto's post [here](https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/how-is-the-proper-right-to-left-text-written-and-wrapped/msg636654682/#msg636654682), the first can be achieved by integrating the Harfbutz library with FreeType, while the second would require a library switch.

(Even taking this approach, it might still be necessary to do something to support RTL input in TextBoxes.)

fernewelten:
I think that the very concept of buttons having a (right-to-left or left-to-right) direction, that concept is kludgy and flawed and wrong. It is __languages__ that have a text direction, not controls.

In right-to-left languages, it isn't even given that __all__ of a sentence has the __same__ direction. For instance, when a sentence contains a number then this number is commonly rendered left-to-right even when the rest of the sentence  must be rendered right-to-left. Same when right-to-left sentences quote English text. (The reason that filenames are written left-to-right even in right-to-left languages is that filenames are written in English, as it were,)

If we try to replicate directionality in self-written code, we're bound to get it wrong. Directionality is one of those concepts that are more complicated than it seems at first sight. The more so for developers such as us that don't use nor know themselves any right-to-left languages or their conventions. Ideally, we should make use of a library that does this for us and let the library worry about directionality.

IMHO the way to go is Unicode. We should find a Unicode library that supports directionality and stick to their rendering functions. 

For AGS4, I plead:
- __In the Engine,__ let's discontinue support for right-to-left __byte__ fonts and only deal with direction within Unicode rendering. Byte fonts would automatically be left-to-right.
- Cut out all the self-written code that deals with direction.
- Don't make a special case for controls that display filenames. These filename ought to be displayed with Latin letters, with the directionality that Latin letters have, namely, left-to-right. So this is already taken care of. If a programmer loads such a control with right-to-left Unicode all the same, well, that's on them: They or the operating system might perhaps have found a way to open files with these strings
- Engine code that accepts user input will probably need to provide a way for users to enter "Change the direction" Unicode characters. We should find a way to provide that.

- __In the Editor,__ for all the fields that don't receive identifiers, let the Windows Foms library deal with directionality and capture these strings as Unicode. There's probably some way to do this “officially” by now, and Microsoft is much more likely to get this right than we are. 
- For all the fields that receive identifiers such as Hotspot names, configure the fields within Windows Forms so that only ASCII can be entered.

ivan-mogilko:
I am looking to what to cover in 3.*.

ericoporto:
> while the second would require a library switch.

Just to correct the understanding from my post, it's not a library switch as there is no library beyond Cairo that can do so (and we probably don't want it, it's a fairly big and complex dependency). What I mention is bidi requires an additional library, which would help us to reason about it, but we still need to code things like line wrapping and all. 

SuperTux Kart had an approach for a little while that was similar to ours, but they currently are using a more complex approach from their engine themselves because they wrote some hardware accelerated text rendering.

PacketLauncher:
1. By per control do you mean that all GUI elements (apart from List Boxes that already have it) will have the hidden 'Translated' bool flag visible in the editor? If so then that could be great, as one may want to keep a certain button for example in the original language, without it being affected by the translation file.

Regarding List Boxes that serves for game saves: I'm not sure how AGS handles it if the list box is set to accept other languages input, like Hebrew. I do remember from some Hebrew version games that we could write our game saves description in Hebrew, while the filename created was always in English and with a general syntax of SaveXXX. But anyway, having a separated RTL bool option so the user could set whether the specific element should be switched to show text in RTL based on the translation file then that would be even better, as it gives more control on the appearance of the elements. I think that such bool should be for all GUI elements.

2. As for a missing entry - do you mean if the translation file has not been filled with a transation for a particular line yet but an empty line below the original text? If so then I think that the engine should always be based on the original; if the original is in English which is LTR then have the original text appear at the missing entry in LTR.

* I wouldn't want that any new feature will break backward compatibility when importing old games into the engine. Most people that replied here have much more knowledge than I do so I can't argue. :) The only thing I'm not sure about if I understood correctly is regarding the Unicode and TTF fonts mentioned by @fernewelten. If AGS go in that route, what would be the fate of pixel-art fonts that we need to show from RTL for Hebrew for example? I mean, most users probably don't have the knowledge of editing actual TTF fonts (which most of them are not free for commercial use) and so they are based on pixel-art drawn fonts that they make in simple paint programs or in SCI Studio / SCI Companion.

messengerbag:
@fernewelten 
> In right-to-left languages, it isn't even given that all of a sentence has the same direction. For instance, when a sentence contains a number then this number is commonly rendered left-to-right even when the rest of the sentence must be rendered right-to-left.

This is a digression, but I don't entirely agree with this analysis, because it assumes that numbers are inherently big-endian: starting with the most significant digit. However, the same number representation can equally well be understood as big-endian written left-to-right, or as little-endian written right-to-left. (And in fact, in school I was taught that the reason the positional notation increases in order of magnitude from right to left was because it came from a right-to-left writing system—Arabic, as I was taught, although they didn't in fact invent the system.)

Consider: When working with numbers by hand, doing arithmetic, you often start by writing the least significant digits (i.e. little-endian), and in spoken form, many languages express numbers in ways that are at least partially little-endian, e.g. German "einundzwanzig" ("one-and-twenty") for 21.

As a matter of implementation, Unicode if I understand correctly assumes a big-endian interpretation of Western Arabic numerals internally (i.e., they are stored in the string with the most significant digit first), and hence has to render them LTR, but only in a "weak" way: although laid out LTR, it's not considered a "real" direction switch within an otherwise RTL text.

ivan-mogilko:
This PR was meant as a fix to 3.* branch, and I really was not planning to discuss larger future changes here, but since this was brought up, then I *have* to mention this:

@fernewelten 
> Let's discontinue support for right-to-left byte fonts and only deal with direction within TTF font rendering. Byte fonts would automatically be left-to-right.

@PacketLauncher 
>  The only thing I'm not sure about if I understood correctly is regarding the Unicode and TTF fonts mentioned by @fernewelten. If AGS go in that route, what would be the fate of pixel-art fonts that we need to show from RTL for Hebrew for example

From what I understand, text direction control is not related to a font type, whether TTF or bitmap font. Text direction is defined by the control characters in text. Font is merely an execution of text. If we support control chars in text, then these may be used with any font type, I believe.

messengerbag:
To just address the 3.* branch, I think it makes sense to extend support to the other controls (TextBox presumably being the trickiest one), behind a version check in case older games have worked around the problem by reversing the text themselves.

I don't think it's worth it to have it as a game-dev-visible setting per control. Controls should take their text direction from the game-wide setting (which I assume is what is set by the translation). That presumably implies that if a translation is missing and it displays the default language text, it will show up reversed if the two languages have different text directions. Sure, that is not ideal, but it's easy to fix (just don't have any missing lines in the translation), and the alternative seems to me like too much of a change in behavior, which will require careful design to get right and a good deal more logic by game devs to handle. All in all, it seems like investing too much in a stopgap solution.

ivan-mogilko:
@messengerbag I decided to try applying text direction based on control's "Translated" flag. This flag already exists, since 3.3.0 iirc, but it was only exposed for ListBoxes in the editor. (Frankly, I suspect not many people realize it's there and use it)
My current thinking is, if a control is marked as "don't translate" that effectively means that it's intended to store some "technical" info, like a list of files, and therefore RTL should not apply as well.

On a side note, I suspect that leaving Translated property editable only for ListBoxes could've been an oversight on my side. As one may imagine a practical case where user would fill ListBox with "untranslated" items (like files), and then copy the item text to a label. In such situation, it would be preferable that the label does not translate its text as well.

I might move Translated property "up" from GUIListBox to GUIControl in the Editor, making it available in all controls. Apparently that might work without breaking anything, as game project loads properties by names, and name stays the same.

messengerbag:
Nice. I believe I can see edge-cases where you might want RTL text on untranslated controls, but it should be possible to work around that.

As for TextBox, remember that its contents can be set programmatically, so those strings may be translated like any others.

ericoporto:
https://faultlore.com/blah/text-hates-you/

Just interesting for explaining the little issues of rendering text

PR #2046, opened 2023-06-23
Implement a Video texture cache

### LINKED ISSUE #1869: Add a texture cache for sprites in Video memory
**Describe the problem**
When profiling AGS engine, specially in high resolution games, it's noticeable that there's a reasonable amount of time spent in [BitmapToVideoMem](https://github.com/adventuregamestudio/ags/blob/778c898f923ecdad3eb305f306015172df1aeb4d/Engine/gfx/gfxdriverbase.cpp#L397) and [BitmapToVideoMemOpaque](https://github.com/adventuregamestudio/ags/blob/778c898f923ecdad3eb305f306015172df1aeb4d/Engine/gfx/gfxdriverbase.cpp#L562) functions, from the VideoMemoryGraphicsDriver. These functions convert the bitmaps to a format friendly to the graphics driver color depth, with some specific adjustments (e.g., adjustment to texture tiles), and then push these bitmaps to the videocard memory as textures.

As I can understand, right now, when animating something repeatedly on screen, a previous frame that appeared recently must be pushed again, if it's not kept on screen somewhere.

**Suggested change**
Sprite that is loaded directly from the assets doesn't change dynamically in game, it would be nice if the textures created from these could be cached, maybe in a "first level" texture cache.

There may be some caveats

- Video memory varies greatly across videocards, we may need a limit to this cache
- When using dynamic sprites, or other generated sprites (texts?), it should possibly instead use the current method available, ignoring this new cache.
- Maybe if out of the cache memory limit (but required to show a sprite), also devolve to the current available method
- There may be some cases where the video driver requires we empty the cache and recreate all textures (maybe switching from Fullscreen to windowed mode?) Edit: apparently only in Direct3D and with sprites that change render target.
- ~~Caching room backgrounds is probably not very useful~~ Edit: forgot about animated backgrounds! Thanks 
- Not sure if the implementation should allow a case where a sprite is in Texture Cache but not in the spritecache anymore, although we don't need the data, the width and height information may be used elsewhere (need to check), but this mode may be something left for later.

**Additional context**

We already have a short term texture cache, see #1657

ivan-mogilko:
Resolves #1869.

1. Implements ResourceCache abstract template class, meant for storing items and tracking their recent use with MRU list. Cache has a size limit, and when the total size of items reaches this limit, it will free space automatically for the next item, disposing least used ones.
Supports Locking and Releasing an item (this temporarily prevents it from disposal, regardless of situation).
Supports storing External items. External items are permanently locked and not counted towards the cache limit, and the primarily purpose is to be able to keep externally managed items in the same storage and share same range of keys with them.
Example of "external item" is DynamicSprite put into the spritecache.
NOTE: I'm still bit conflicted about External items, whether to store them in the same cache class. They may be stored separately, but then getting them would require two key lookups instead of one.

2. SpriteCache inherits ResourceCache, but hides its methods, as it has additional layer of functionality, where sprites are loaded from assets if they are not found in cache. I think, in the future we may have another base class, something called ObjectManager (for example), which would contain both Cache and Stream for loading assets into cache in it.

3. SoundCache inherits ResourceCache.

4. Implemented TextureCache. This is the main purpose of this PR.
~TextureCache is placed inside VideoMemoryGraphicsDriver class, which is a parent for texture-based renderers (Direct3D/OpenGL).~
TextureCache is done as a global object; this makes it easier to manage by the engine, without having to go past the GraphicsDriver interface.
Stores Texture structs in it (previously known as TextureData), similar to the previously existing short-term cache.
Incorporates previously existing short-term cache: moved it out from the VideoMemoryGraphicsDriver.
Removed GetSharedDDB() methods from the GraphicsDriver interface, trying to keep it simpler, with only factory and rendering methods.
In the TextureCache, item size is currently approximated, using know texture resolution.
DDBs now can attach and detach Texture, this is useful to avoid redundant re-allocations of DDB struct.
Adjusted a function that syncs object texture, now the algorithm is this
    * ask gfx driver for a cached texture, on success use that;
    * if no texture, then request a raw bitmap from spritecache (so it won't waste time loading it if texture already exists);
    * having a raw bitmap, ask gfx driver to make a new cached texture from it.

5. Performance changes.
I only did few tests with "infinite fps" mode, paying most attention to available high-res game, which was Dave Gilbert's "Old Skies Demo" again.

The results were following:
| location | 3.6.1.1 | 361--texturecache |
| --- | --- | --- |
| main menu | ~525 | 670-680 |
| bedroom | 550-570 | 670-680 |
| travel room | 590-600 | 670-680 |

To summarize, the texture cache eliminates a redundant processing that caused most fps differences between the scenes. Now all game scenes run at more or less same fps (unless there are other specific issues).

6. KNOWN ISSUES.
    * ~[ ] I noticed a weird issue in Dave's game, during the intro the vertical scrolling scene slows down to a crawl. Not sure what was causing this, need to investigate.
    EDIT: not related to texturecache itself, but to a ResourceCaches internal logic. Somehow it does wrong things when either locking sprites, or freeing mem, and never frees enough, so images keep reloading.~
    EDIT2: not a bug, but a difference in how the cached size is calculated for "precached" sprites. The new cache is more "honest" about it, so user needs to give more cache size in config.

7. TODOs.
    * [x] Add a config option for the texture cache, basically wanted VRAM size to spend on it.
    * [x] Add this option to winsetup too.
    * [x] Investigate if it's possible to find actual VRAM the system has, and if it is, then clamp the config setting to a portion of it (say, 60-80%), as texture cache is not all that the game uses.
    * [~] Investigate if it's possible to find out exact size the texture has in vram. Although I've read that it may not be possible (at least not with OpenGL). In the worst case we may use current approximation.
    * [~] Investigate the GUI: right now they are using custom textures, as GUIs and Controls may have multiple elements to draw each. However, in theory, we could make them report whenever they are represented by a single sprite. For example: a button which has a graphic and no text, or a GUI which has a background image, and no colored border. In such cases they may reuse the texture cache as well. I suppose that this may be specifically advantageous for animated buttons.
    EDIT: better do this separately later.
    * [x] Investigate to optimize overlay -> texture, currently always loads a sprite bitmap first.
    * [x] I left some FIXME comments in the code, about further minor optimizations and improvements.


ivan-mogilko:
In regards to getting VRAM info, I only quickly found these examples for Direct3D 9:
https://github.com/walbourn/directx-sdk-samples/tree/main/VideoMemory

ericoporto:
I think you can make the defaults for texture cache a bit bigger. In case we think the texture is there, but it's not, what happens? I am mostly thinking about mobile where I expect the GPU to have a large memory to cache but also to aggressively remove things if the app is alternated with other app and get focus again.

ivan-mogilko:
>  In case we think the texture is there, but it's not, what happens?

The texture gets loaded from sprite again.

ericoporto:
> TBH this situation sounds absurd, and has to be clarified.

Disconsider what I said, I made some confusion with a different thing from shaders that is unrelated.

ivan-mogilko:
Made another variant of the texture cache, in a separate branch for now, because it does not reimplement all the needed functionalities: https://github.com/ivan-mogilko/ags-refactoring/tree/361--texturecache--variant2

The variant 2 is the texture cache stored as a separate object, as opposed to a member of the VideoMemoryGraphicsDriver class. The potential benefit of this is easier control over the cache by the engine, without a need to duplicate cache methods in gfxdriver interface (such as locking and clearing items, configuring max size, and so forth).

ericoporto:
That sounds reasonable. Not in this PR, but would moving the coordinates stuff and the software renderer stuff from draw to other place, make sense?

ivan-mogilko:
Remade the PR with the external TextureCache variant, updated the description.

ericoporto:
hey, in my tests of running games and have performance being good, this is working.

> not a bug, but a difference in how the cached size is calculated for "precached" sprites. The new cache is more "honest" about it, so user needs to give more cache size in config.

About the TODO about adding it in winsetup, I don't know what to write, but it looks like this user setting is important. Just curious, if the texture cache is set to 0 in config, does it revert to the previous behavior of the short term cache?

ivan-mogilko:
> Just curious, if the texture cache is set to 0 in config, does it revert to the previous behavior of the short term cache?

I wrote about this in the description.
Short-term cache always works, it keeps weak references to textures that are currently in use by the game objects. Whenever a texture is not found in the regular cache, the short term cache is checked first, in case there are objects with this texture already, and sprite is loaded only after.

ericoporto:
Hey, the TO-DO tasks don't look like they are needed for this PR to work, adding to winsetup would be nice but as the default is reasonably safe it should be good as is. 

The TO-DO about optimizing Overlay->Texture may be it's own ticket. The other TO-DOs look like they could be their own ticket too.

ivan-mogilko:
Did overlay update, it was a small change.
Another thing that I forgot to look into are mouse cursors, but probably it's something trivial too.

Mentioned GUI update may be done separately.

I'd look into "max vram" test, and winsetup update later today, or tomorrow.
Also, I think, software rendering may need small adjustment, as I could've made it slightly less efficient by these changes. It needs to skip the texture cache altogether.

ericoporto:
This all looks good, just two ideas (they can be dismissed): after a cache is created and size is set, maybe print in the debug log; maybe add texture cache information to the debug message in Ctrl+V where we mention sprite cache size too.

ivan-mogilko:
> This all looks good, just two ideas (they can be dismissed): after a cache is created and size is set, maybe print in the debug log; maybe add texture cache information to the debug message in Ctrl+V where we mention sprite cache size too.

Added info to log; the debug message already had texture cache in some earlier commits.

ivan-mogilko:
Found a bug which causes crash for non-shared textures when they change size (like textual overlays and gui).

Following this bug, found a bug in 3.6.0 which crashes the engine if an object is using DynamicSprite which is resized while the object was on screen.

PR #2048, opened 2023-06-24
3.6.1: improve ScreenOverlays storage performance

### LINKED ISSUE #2045: Improve ScreenOverlay storage and creation/deletion performance
3.6.0 removed the overlays limit, which previously was mere 30, but the way their creation is handled in the engine remained **very** suboptimal. Somehow this fell out of my mind during 3.6.0 development, so this has to be addressed now.

**What is the problem here?**

1. Overlay structs are stored in a vector.
2. They are added/removed using standard push_back/erase in vector, which means that vector contents get moved in memory every time. ScreenOverlay is not a big struct (few ints and a bitmap pointer), but regardless this creates an overhead.
3. Secondly, if you look at the [`add_screen_overlay_impl`](https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/overlay.cpp#L417-L422), you may notice that the engine is trying to find a free ID to assign for the new overlay. This is done in a **horribly inefficient way**, where it will iterate all possible ID values starting with OVER_CUSTOM, and *for each ID* it will iterate over the vector looking whether this ID is used. EDIT: this maybe slightly improved by recent changes in #2041, but not enough.

I made a test game in order to illustrate the problem:
[test--361-overlayperformance.zip](https://github.com/adventuregamestudio/ags/files/11851527/test--361-overlayperformance.zip)
[test--361-overlayperformance-project.zip](https://github.com/adventuregamestudio/ags/files/11851528/test--361-overlayperformance-project.zip)

On startup it creates 1000 overlays, then starts randomly moving them around, thus accessing X,Y properties.
On "Space" press it toggles the "recreation" mode, where it deletes 1 random overlay and creates new 1 overlay each game frame. (Another test was done adjusting this to 10 recreated overlays per frame)

Testing with 3.6.1.1 and a recent version with fixed performance (#2041), in the infinite fps mode:
Engine | Moving 100 per frame | Recreating 1 per frame | Recreating 10 per frame
---|---|---|---
3.6.1.1 | 600 | 560 | 340
new version | 650 | 600 | 380

---

**What may be a solution here?**

1. Store ScreenOverlays in a (hash)map, where ID is the key.
2. Keep them in vector, but at least don't erase the elements and replace with invalid (empty) overlay object. Added overlays will be placed in available empty slots first, and new slots pushed only if the vector is full.

The second variant will look similar to how the ManagedPool is [implemented currently](https://github.com/adventuregamestudio/ags/blob/169320f1b54b7f778b118f2cabd93d580286e1b0/Engine/ac/dynobj/managedobjectpool.cpp#L146-L161
). It records a `available_ids` stack, which contains a list of empty indexes, and reuses them first whenever a new object is added.

Similar "available_ids approach may be used for assigning an overlay ID. This is regardless of whether ScreenOverlays are stored in a map or kept in a vector.

On the other hand, we may actually "merge" overlay's ID (aka "type") and index, meaning that type == slot index. This will make a lookup simple. The only downside here would be that there's a large gap between "internal" engine's overlay types (there are 4 of them), and custom overlays which are assigned ID starting with 101 (see OVER_CUSTOM).
This problem may be reduced by making OVER_CUSTOM, say 20, to have some space "just in case".
Although, theoretically, we could remove the fixed overlay IDs completely, and instead save the new allocated IDs in variables when creating "special" overlays, such as blocking speech, etc. There's additional complication with this related to backwards compatibility, so this option has to be investigated carefully.

ivan-mogilko:
Fixes #2045.

Implemented variant 2, mentioned in the issue ticket.
I might also try variant 1 (hashmap) later to see if that makes things faster or more convenient.

In variant 2, we keep ScreenOverlays in std::vector, but do not erase elements in the middle and invalidate them instead. Record free indexes in a helper stack. When creating a new overlay, first check for the recorded free indexes.
Overlay's type now equals storage index, thus no need to search for overlay of certain type anymore.

This speeds up both creation/deletion of overlays, and their access.

Removed the gap between "fixed" overlay types and custom ones, to save space. If we ever need more special overlays, these could be created as custom and their ids saved in a variable (similar to how background speech works).
Previously present special overlay types are left for backwards compatibility (and to simplify the transition).

Was testing with this game:
[test--361-overlayperformance.zip](https://github.com/adventuregamestudio/ags/files/11851527/test--361-overlayperformance.zip)
[test--361-overlayperformance-project.zip](https://github.com/adventuregamestudio/ags/files/11851528/test--361-overlayperformance-project.zip)
On startup it creates 1000 overlays, then starts randomly moving them around, thus accessing X,Y properties.
On "Space" press it toggles the "recreation" mode, where it deletes 1 random overlay and creates new 1 overlay each game frame. (Another test was done adjusting this to 10 recreated overlays per frame)

FPS comparison (in the "infinite fps" mode):

Engine | Moving 100 per frame | Recreating 1 per frame | Recreating 10 per frame
---|---|---|---
3.6.1.1 | 600 | 560 | 340
recent master | 650 | 600 | 380
this PR | 650 | 645 | 615
hashmap | 640 | 630 | 610

Needs more testing, as certain algorithms have changed.

UPDATE: Made a hashmap variant: https://github.com/ivan-mogilko/ags-refactoring/tree/361--overlayperf3-hashmap
So far it looks like hashmap works slightly slower in this case. Also it possibly breaks backward-compatible z-order when drawing too, because the order of overlays in container is undefined.

### DIFF
--- a/Engine/ac/overlay.cpp
+++ b/Engine/ac/overlay.cpp
@@ -13,6 +13,7 @@
 //=============================================================================
 #include "ac/overlay.h"
 #include <algorithm>
+#include <queue>
 #include "ac/common.h"
 #include "ac/view.h"
 #include "ac/character.h"
@@ -42,9 +43,11 @@ extern int face_talking;
 extern std::vector<ViewStruct> views;
 extern IGraphicsDriver *gfxDriver;
 
-
+// TODO: consider some kind of a "object pool" template,
+// which handles this kind of storage; share with ManagedPool's handles?
 std::vector<ScreenOverlay> screenover;
-std::vector<int> overlookup(128);
+std::queue<int32_t> over_free_ids;
+
 
 void Overlay_Remove(ScriptOverlay *sco) {
     sco->Remove();
@@ -52,7 +55,7 @@ void Overlay_Remove(ScriptOverlay *sco) {
 
 void Overlay_SetText(ScriptOverlay *scover, int width, int fontid, int text_color, const char *text)
 {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!Overlay.SetText: invalid overlay ID specified");
     const int x = over->x;
@@ -83,7 +86,7 @@ void Overlay_SetText(ScriptOverlay *scover, int width, int fontid, int text_colo
 }
 
 int Overlay_GetX(ScriptOverlay *scover) {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!invalid overlay ID specified");
 
@@ -92,7 +95,7 @@ int Overlay_GetX(ScriptOverlay *scover) {
 }
 
 void Overlay_SetX(ScriptOverlay *scover, int newx) {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!invalid overlay ID specified");
 
@@ -100,7 +103,7 @@ void Overlay_SetX(ScriptOverlay *scover, int newx) {
 }
 
 int Overlay_GetY(ScriptOverlay *scover) {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!invalid overlay ID specified");
 
@@ -109,7 +112,7 @@ int Overlay_GetY(ScriptOverlay *scover) {
 }
 
 void Overlay_SetY(ScriptOverlay *scover, int newy) {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!invalid overlay ID specified");
 
@@ -117,7 +120,7 @@ void Overlay_SetY(ScriptOverlay *scover, int newy) {
 }
 
 int Overlay_GetGraphic(ScriptOverlay *scover) {
-    auto *over = find_overlay_of_type(scover->overlayId);
+    auto *over = get_overlay(scover->overlayId);
     if (!over)
         quit("!invalid overlay ID specified");
     return over->GetSpriteNum();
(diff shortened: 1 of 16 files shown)


ivan-mogilko:
@ericoporto could you please try and see if this does not make your game slower again?

ericoporto:
It looks like it doesn't make my SandWalker game slower or faster, it appears to be the same! Don't Give Up The Cat though gets a significant bump in my PC (from 200 fps to 240 fps with this PR!).

ivan-mogilko:
Fixed few things, also made it write only valid overlays into the game save again, empty slots are ignored. This made saving function a bit ugly, but maybe not much.

ivan-mogilko:
Made a hashmap variant: https://github.com/ivan-mogilko/ags-refactoring/tree/361--overlayperf3-hashmap

Updated comparison table:
Engine | Moving 100 per frame | Recreating 1 per frame | Recreating 10 per frame
---|---|---|---
3.6.1.1 | 600 | 560 | 340
recent master | 650 | 600 | 380
optimized vector | 650 | 645 | 615
hashmap | 640 | 630 | 600

Looks like hashmap works slightly slower in this case. Also, it has additional complications for saving/restoring and possibly drawing too, because the order of overlays in container is undefined.
So I guess I will stick to "optimized vector" variant for now.

ericoporto:
Hey, it's all working, performance is good, but just to be sure, savegames from 3.6.1 load alright, new savegames made with the latest from this pr load alright but savegames made with the initial build from this PR segfault

ivan-mogilko:
> but savegames made with the initial build from this PR segfault

Do you mean the earlier commits from this PR, before the recent force push?
There were couple of mistakes at first, although i was not certain if they are made during save or during restore.

ericoporto:
So yes, the current PR can save and the save can be loaded in 3.6.1 and the reverse is true too. Just when this PR was still only two commits, yesterday, that build makes a savegame that the current PR segfaults on load.

Engine/ac/overlay.cpp
```diff
@@ -390,24 +389,24 @@ void remove_screen_overlay(int type)
...
+    for (uint32_t i = 0; i < screenover.size(); ++i)
         remove_screen_overlay(i);
 }
 
-ScreenOverlay *get_overlay(int index)
+ScreenOverlay *get_overlay(uint32_t id)
 {
-    assert(index >= 0 && index < screenover.size());
-    return (index >= 0 && index < screenover.size()) ? &screenover[index] : nullptr;
+    return (id < screenover.size() && screenover[id].id != ScreenOverlay::NoID)
```
ericoporto, line 405: Hey, if id is lower than screenover.size(), isn't it automatically also not `ScreenOverlay::NoID` ? This is more a question really.

ericoporto:
https://github.com/adventuregamestudio/ags/pull/2048/commits/1146349dc65822ba8119665624e273163d85dc26#diff-94f963a99c10f8e1c3e82098d0b2e48bcdbcc3701231e67cf86e1bc70910c4ffR405

Question, with this new update, if id (or type?) is lower than screenover.size(), isn't it also different than NoID ? I thought it was a genius way to both leverage the type system and use a single check for two things.

ivan-mogilko:
> Question, with this new update, if id (or type?) is lower than screenover.size(), isn't it also different than NoID ? I thought it was a genius way to both leverage the type system and use a single check for two things.

Sorry, I reverted that.
I had a thought to try making overlay ID unsigned, but then I had doubts, because it's signed everywhere else (Chars, Objects, and so forth), and if exposed to script, script integers are only signed, so the benefit seems smaller.
So I redid in a much simpler fix with casts to unsigned only where necessary.

> This new version doesn't segfault with the save from the initial PR version!

I think this is fixed by adding extra check when reading an overlay
https://github.com/adventuregamestudio/ags/pull/2048/commits/ed520227cea80bd2503451b61780b4e6be962cc3

ericoporto:
I agree with making things consistent, if we use int everywhere else, lets keep it here! This will avoid type casting issues!

Anyway, I fetched from your branch and rebuilt locally it here and it seems the performance is still the same and it looks like things are working.

ivan-mogilko:
> Edit: erh, there is a save that is giving me

Please tell which version made this save, could you upload it?

ericoporto:
[dguptc_agssave.zip](https://github.com/adventuregamestudio/ags/files/11861023/dguptc_agssave.zip)
Here is the game and the savegame, I am not sure which version made it - there is a chance it's just some intermediary version from this PR and in this case it doesn't matter.

ivan-mogilko:
> So my guess is the above save doesn't matter and can be dismissed.

Yeah, I had some guess why, but could not actually prove that, so probably something broke in one of the older commits.
At least it does not crash.

ivan-mogilko:
> [1146349#diff-94f963a99c10f8e1c3e82098d0b2e48bcdbcc3701231e67cf86e1bc70910c4ffR405](https://github.com/adventuregamestudio/ags/commit/1146349dc65822ba8119665624e273163d85dc26#diff-94f963a99c10f8e1c3e82098d0b2e48bcdbcc3701231e67cf86e1bc70910c4ffR405)
> 
> Question, with this new update, if id (or type?) is lower than screenover.size(), isn't it also different than NoID ?

In regards to this, because similar condition is also in the latest version,
this condition checks for two things:
* if there's an allocated slot in the storage (vector)
* if the slot is actually valid (overlay's type is >= 0)

ericoporto:
Hey, this all looks good. It works on my few tests. I just don't know if I tested the other stuff (portraits, ...). I definitely didn't test the old savegame format.

PR #2053, opened 2023-07-02
Emscripten: Fix Safari can't Fullscreen

ericoporto:
opening as a draft to discuss the problem. I can't quite fix the complex layers of abstractions in Emscripten and SDL2 to make it work with Safari, but I can just ignore all that and just do the 4 lines of JS that are needed to make it work...

I added a new method in platform driver where I can tell if changing to fullscreen is supported or not, and if not, I just don't do anything in the engine.

This means that you can only enter fullscreen in Safari by using the "Toggle Fullscreen" button in the HTML launcher, and you can exit by press Esc (or Swiping down in iPhone). Additionally, in iPhone there is no way to remove this gesture so if your game has swipes (like my ManaMatch game), it's basically impossible to keep it fullscreen on iPhone...

(this all really gives me an urge to remove the fullscreen from the gear and add it separately in its own button!)

https://www.adventuregamestudio.co.uk/forums/engine-development/ags-engine-web-port/msg636656123/#msg636656123

### DIFF
--- a/Engine/platform/base/sys_main.cpp
+++ b/Engine/platform/base/sys_main.cpp
@@ -193,6 +193,7 @@ SDL_Window *sys_get_window() {
 
 void sys_window_set_style(WindowMode mode, Size size) {
     if (!window) return;
+    if(mode != kWnd_Windowed && !platform->FullscreenSupported()) mode = kWnd_Windowed;
     // NOTE: depending on which mode we are switching to, the order of
     // actions may be different; e.g. if we are going windowed mode, then
     // we first should disable fullscreen and set new size only after;
--- a/.gitignore
+++ b/.gitignore
@@ -28,8 +28,7 @@
 *.iml
 .idea/
 # CLion
-cmake-build-debug/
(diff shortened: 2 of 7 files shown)


Engine/platform/base/sys_main.cpp
```diff
@@ -193,6 +193,7 @@ SDL_Window *sys_get_window() {
 
 void sys_window_set_style(WindowMode mode, Size size) {
     if (!window) return;
+    if(mode != kWnd_Windowed && !platform->FullscreenSupported()) mode = kWnd_Windowed;
```
ivan-mogilko, line 196: Only one small nitpick: please use our codestyle, I noticed you are sometimes writing the new code using this different style:
* there has to be a space after `if`
* please place the statements under `if` on the next line; I know there have been several cases added where it's on the same line (either by nick sonneveld, or by me following his example), but these were the return statements, and here it's an assignment, also it becomes especially inconvenient to read the code when there's a long condition like this.

PR #2054, opened 2023-07-04
Make all script File commands treat paths as case insensitive

### LINKED ISSUE #2026: Make all file-related script API case-insensitive
As same AGS games may be run on both case-sensitive and insensitive filesystems, their asset loading at runtime is made case-insensitive. However, the File script commands were not, at least when they point to actual directories (paths with `$DATA$` token happen to be case-insensitive because they are resolved using AssetManager).

In order to make things consistent between various ways to load game resources, and also between systems, all these script commands should also be made case-insensitive.

This includes:
* File.Open;
* File.Exists;
* File.Delete;
* ListBox.FillDirList - treat subdirs in ci way; also should join multiple items which differ only by the letter case;
* DynamicSprite.CreateFromFile
* DynamicSprite.SaveToFile
* and corresponding legacy API which is not remapped to new funcs:
    * SaveScreenShot

In addition there's a potential problem with ResolveWritePathAndCreateDirs(), which is meant to support subdirs for files created by script. Right now it uses simply Directory::CreateAllDirectories() which ignores case sensitivity differences. Either we need to add a parameter that would instruct it to do CI tests for existing dirs while creating them, or run a test prior to calling this and find out which dirs already exist in CI way.
EDIT: i guess we could utilize FindFileCI for this as well, if we add an optional output parameter that would store the longest matching path found.

**Implementation issues**

We have a File::FindCI() function that helps finding a file in a case-insensitive way on cs-filesystems. But it misses one important bit: it does not process subdirectories, which is essential, as above script commands support relative subdirs in paths.

**Test game**

https://github.com/adventuregamestudio/ags/files/11992261/test--361fileci.zip
https://github.com/adventuregamestudio/ags/files/11992262/test--361fileci-proj.zip

Does a number of manipulations with files in script, using case mismatching filenames, and prints results to log.

ivan-mogilko:
Resolves #2026.

Makes all the script file commands simulate case-insensitive filesystem. This fixes compatibility issues in user scripts, and makes game data and file access from script work alike (as engine already uses case-insensitive file search for game assets).

TBH this is a bit annoying, as the problem in question is relatively rare, but I believe it has to be done to make engine consistent.
On another hand, *at least some* of the changes in this PR would be useful on their own.

The changes in this PR:

1. Implemented DirectoryIterator and DirectoryRecursiveIterator classes, extracting code from FindFile and FindFileRecursive. This lets to have an object with bare minimum functionality when you do not need a wildcard pattern match. Also, optimized stats reading on non-Windows systems (compared to old FindFile impl) by *not* calling stat() until user explicitly requests current entry's information.
**NOTE:** these classes will become redundant, if we upgrade to C++17, which has similar features.
~FindFile is converted into a template, which uses defined DirectoryIterator type to search some location in a loop, applying some pattern.~
FindFile and FindFileRecursive merged into a single FindFile class, which uses DirectoryRecursiveIterator, but may configure it to search only in the first dir.
2. Expanded FindFileCI() function:
    * Uses DirectoryIterator; this makes it fully generic, independent from system API;
    * Can search for nested subdirectories too, that is - subdirs that have different letter case in names;
    * Can search for a final directory, besides a file (don't expect to find a file but dir, and report success);
    * Optionally fills in the most matching path on failure (this lets to know how far it went).
3. Made all script File functions use a case-insensitive approach. Reading functions check for existing file using FindFileCI, writing functions use FindFileCI to find the most of the matching path before creating necessary subdirectories. Script functions affected are:
    * File.Delete
    * File.Exists
    * File.Open
    * DynamicSprite.CreateFromSprite, and LoadImageFile
    * DynamicSprite.SaveToFile
    * ListBox.FillDirList

More information about changes may be found in commit descriptions.

### DIFF
--- a/Common/util/file.cpp
+++ b/Common/util/file.cpp
@@ -14,16 +14,15 @@
 #include "util/file.h"
 #include <errno.h>
 #include <stdexcept>
-#if !AGS_PLATFORM_OS_WINDOWS
-#include <sys/stat.h>
-#include <dirent.h>
-#include <string.h> // strcasecmp
-#endif
 #include "core/platform.h"
 #include "util/bufferedstream.h"
 #include "util/filestream.h"
 #include "util/path.h"
 #include "util/stdio_compat.h"
+#include "util/string_compat.h"
+#if defined (AGS_CASE_SENSITIVE_FILESYSTEM)
+#include "util/directory.h"
+#endif
 #if AGS_PLATFORM_OS_ANDROID
 #include "util/aasset_stream.h"
 #include "util/android_file.h"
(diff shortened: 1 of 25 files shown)


ivan-mogilko:
Had to revert/remake new FindFile class design, because it turned out to be more difficult to configure in existing uses.
Will squash the commits later.

ivan-mogilko:
Still not ready, too many mistakes, so postponing again.

ivan-mogilko:
Seems working.

Used following game to test this.
[test--361fileci.zip](https://github.com/adventuregamestudio/ags/files/11992261/test--361fileci.zip)
[test--361fileci-proj.zip](https://github.com/adventuregamestudio/ags/files/11992262/test--361fileci-proj.zip)

It does a number manipulations with files in script, and prints results to log.

AlanDrake:
I think there's a bug, `File.Exists` is always returning true

Common/util/file.cpp
```diff
@@ -250,103 +249,130 @@ Stream *File::OpenStderr()
...
 }
 
-String File::FindFileCI(const String &dir_name, const String &file_name)
+String File::FindFileCI(const String &base_dir, const String &file_name,
+    bool is_dir, String *most_found, String *not_found)
 {
 #if !defined (AGS_CASE_SENSITIVE_FILESYSTEM)
     // Simply concat dir and filename paths
-    return Path::ConcatPaths(dir_name, file_name);
+    return Path::ConcatPaths(base_dir, file_name);
```
AlanDrake, line 257: Shouldn't we be checking if the file exists here instead of returning a path?

AlanDrake: Is this what you anted to do here?
```
    String path = Path::ConcatPaths(base_dir, file_name);
    if (ags_file_exists(path.GetCStr()) != 0)
        return path;
    return {};
```

ivan-mogilko: No, this function is not supposed to check whether file exists, unless there's no other way (for case insensitive filesystems).

ivan-mogilko: But yes, File_Exists is broken, because previously it was testing for file existence, now it incorrectly assumes that it's already tested by the call to ResolveScriptPathAndFindFile.

ivan-mogilko: It's possible that FindFileCI will need to be changed for case insensitive filesystems to also test for a file's presence, as you suggested, to make things consistent. But then all the uses of this function should be revisited, and possibly adjusted. From a quick glance, not all of them assume this tests for a file presence, or may fail.

ivan-mogilko: @AlanDrake pushed a fix: 8e67e325ebbac58a5cad4ba85ca620fce7fef4bf

PR #2055, opened 2023-07-10
Editor: unescape text properties when writing game data

ivan-mogilko:
Done for #1226, but is a valid change on its own (i suppose this could be also backported to 3.6.1).
(See [this comment](https://github.com/adventuregamestudio/ags/issues/1226#issuecomment-1522025191) for explanation)

This ensures that special sequences like "\n" are converted to characters '\n', and similar, in properties containing human-readable texts. Normally this is only useful for GUI controls, room messages, and perhaps Custom Properties, but I did it for all of them, for a formal consistency (Character "realname", and so forth).

### DIFF
--- a/Common/util/string_utils.h
+++ b/Common/util/string_utils.h
@@ -50,7 +50,7 @@ namespace StrUtil
     // returns def_val on failure
     float           StringToFloat(const String &s, float def_val = 0.f);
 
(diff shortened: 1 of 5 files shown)


AlanDrake:
Tested along with #1988 
Everything I tried has worked as expected. 👍

PR #2056, opened 2023-07-13
Script API: implemented File.ResolvePath()

### LINKED ISSUE #899: Script API: get resolved script path
https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/file.cpp#L277

This function enables the file open and write to have access to really useful tokens. Unfortunately, the resolved path is not available for a script be able to print the resulting path string from the game itself.

It would be useful to preprocess through script a file path and pass the resulting path to a plug-in API that implements their own file write and open operations.

ivan-mogilko:
Resolves #899

File.ResolvePath resolves a script path into the filesystem path, in exact same way as the engine does it when searching for a file.
 (like doing a case-insensitive search on case-sensitive filesystems).

As a bonus, added readonly File.Path property, that returns an opened File's path.

### DIFF
--- a/Common/util/file.cpp
+++ b/Common/util/file.cpp
@@ -252,23 +252,12 @@ Stream *File::OpenStderr()
 String File::FindFileCI(const String &base_dir, const String &file_name,
     bool is_dir, String *most_found, String *not_found)
 {
-#if !defined (AGS_CASE_SENSITIVE_FILESYSTEM)
-    // Simply try a match
-    String path = Path::ConcatPaths(base_dir, file_name);
-    if ((is_dir && ags_directory_exists(path.GetCStr())) ||
-        (!is_dir && ags_file_exists(path.GetCStr())))
-        return path; // success
-    // TODO: this is never used, but for a formal consistency might
-    // support searching for an existing part of the path, and fill
-    // most_found & not_found
-    return {}; // fail
-#else
+    // Case insensitive file find - on case sensitive filesystems
     if (most_found)
         *most_found = base_dir;
     if (not_found)
         *not_found = "";
 
-    // Case insensitive file find - on case sensitive filesystems
     if (file_name.IsEmpty())
         return {}; // fail, no filename provided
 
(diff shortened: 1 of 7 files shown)


ivan-mogilko:
@ericoporto please try if this is what you needed for #899?

ericoporto:
Hey, about plugin API, in ags4 we removed some things from it so people would instead call the script functions directly, if you don't think the plugin API addition is really needed, I can work with just the script without problems - I think, need to remember how script strings work from the plugin perspective.

Anyway the original ask was indeed for usage from a plugin though, because I started making a SQLite AGS plugin but stopped in the middle because I couldn't figure a nice way to tell where to place the DB from script. I will test this now and either edit or add a comment about it working.

ivan-mogilko:
> Hey, about plugin API, in ags4 we removed some things from it so people would instead call the script functions directly

This was to replace the direct access of structs from engine's memory: characters, objects and view frames.

ericoporto:
If I pass a filename like "myfile.txt" to this function it doesn't return anything. Shouldn't it return the `savedir/myfile.txt` ?

ivan-mogilko:
> Edit: this script has resolved_path always empty and I don't know why!

Because I started rewriting and broke it.

ericoporto:
@ivan-mogilko now it works perfectly!

ivan-mogilko:
I made a serious mistake with plugin API here, which unfortunately went unnoticed. Good thing is that function only exists in 3.6.1 wip version.

```
AGSIFUNC(const char*) ResolveFilePath(const char *script_path);
```

The problem with this function is that it returns a allocated object, and it allocates a "script string" for some reason, which is probably my another oversight. This has two consequences:
* plugin must know to delete this data,
* since it's implemented as a script string, it's reference counted and is present in the managed pool. But probably its ref count is 0, so will be removed from memory at the first GC.

I suppose that instead this has to either return a (char*) and instruct plugin to delete it, but then it should use malloc in case plugin is in C (i think that's formally supported).
Or it should receive a buffer with length as arguments, and return the written size for instance.

ericoporto:
Is it worth it to pass also a max length for the input buffer? So it would be, const char * + len in input and char * + max len in output.

ivan-mogilko:
> Is it worth it to pass also a max length for the input buffer?

No, it expects null terminated string. It has to be like any C-style string function with output buffer constraint, such as `snprintf` or `strncpy`.

ivan-mogilko:
Okay, I did this: 0b35d3135e86da37caa7e790642ed837736e86c8

PR #2057, opened 2023-07-15
Editor: cut out source control integration

### LINKED ISSUE #468: Add Git integration
As discussed in #463, it is very difficult to do parallel development in AGS. It requires a ridiculously complex workflow and merging parallel changes in binary room files is almost impossible.

### I suggest:

- finally getting rid of big binary files that function as a cache for other binary files. (See #469)
  - Merging changes in a binary file is hard enough. It is impossible when all the project's binary files are merged in a single file.
- adding a Git integration to the AGS editor. *(Plugin)*
- adding some kind of user interface to AGS that allows easy management of the project's git repository. *(Plugin)*

### Ideas:

**UI**: Something like Eclipse's Git integration or [gitg](https://wiki.gnome.org/action/show/Apps/Gitg)?
**Files**: A single XML file functioning as an index file and several `.agsasset` files containing preprocessed  assets (as far as I know, Unreal Engine uses this way).

## Required:

- [x] Change game project's Rooms to open format: #469

## Tasks:

- [ ] Implement Plugin
    - [ ] Explore current state of source control API

ivan-mogilko:
Closes #468, as one of the suggested alternatives.

Reasons to remove this feature are:
* it's not necessarily suitable for any arbitrary source control system's workflow;
* it's too much burden to maintain in the AGS codebase;
* it was frankly done in a naive and messy way, and there's a high chance it's been already broken by some changes.
* was likely never used in practice by anyone.


ivan-mogilko:
Nobody else commented on this, so I guess it's fine.

ivan-mogilko:
Something I missed is that there's now a warning about "SourceControl" component not found, because there's a xml entry in every Game.agf produced since AGS 3.0.

Need to see if it's possible to suppress this warning, by hardcoding the component name.

ivan-mogilko:
Apparently some plugins may be referring to the source control, SpeechCenter does this for some reason, I don't know why.

It was a bad idea to remove the interface method itself. I need to test if it will work if I provide a dummy implementation for IAGSEditor.SourceControl.

PR #2061, opened 2023-07-21
Scripting: add object pointer and verb parameters to object interaction functions

### LINKED ISSUE #1765: Script API: add object pointer argument to all "interact" callbacks
AGS allows users to attach same script function to several object events, or several objects (there's no difference). The problem is that all the script callbacks, except for GUI event callbacks, do not take object pointer as an argument, therefore there's no direct way to know which object was interacted with when the function is called.

The proposal is to add 2 arguments to all the interaction callbacks, in a style similar to GUI callbacks:
- the object pointer;
- the interaction verb (in case same function relates to multiple verbs).

The new callback prototype would be, for example:
`function Character_OnInteract(Character* c, CursorMode mode)`

For the reference, the GUI and GUI Control callbacks currently look like:
`function Button_OnClick(GUIControl *control, MouseButton button)`
`function GUI_OnClick(GUI *theGui, MouseButton button)`

Also note: since 3.6.0 engine was fixed to theoretically let user write callbacks that miss parameters, and not trigger any errors. This means that supporting these 2 extra args should not cause any trouble even if they are missing in the script's declaration. Hence old projects and template should continue to work too (needs to be tested).

ivan-mogilko:
Resolves #1765

I believe this is decade overdue...
This adds 2 optional parameters to all of the interaction event callbacks in AGS: object's pointer and a verb (aka cursor mode).
The purpose is to be able to distinguish object and interaction type when user links same function to multiple events. This is similar to how GUI events have GUI/Control and mouse button as parameters.
Right now the users have to resort to ugly workarounds, like calling GetAtScreenXY extra time (which is unreliable), or saving the interacted object in a global variable in on_mouse_click.

The event functions will have following prototype for Characters, Objects, Hotspots and InventoryItems:
```
function cEgo_Interact(Character *c, CursorMode mode)
```

Hotspots have 2 special events without CursorMode param, these are "Walk onto" and "Mouse over":
```
function hHotspot2_MouseMove(Hotspot *h)
```

Regions do not have CursorMode parameter, because their event types are different, not related to interaction mode:
```
function region0_WalksOnto(Region *r)
```

Code-wise the PR is ~50% code refactor.

**Backwards compatibility**

This change is fully backwards compatible, in the sense that script will work even if these parameters were omitted. This was achieved by some of the previous changes in 3.6.0 (related to `on_key_press`), which lets engine to ignore superfluous parameters not matching the function declaration in user script. But it's worth testing this of course.

### DIFF
--- a/Common/gui/guiobject.h
+++ b/Common/gui/guiobject.h
@@ -124,6 +124,7 @@ protected:
     bool     _hasChanged;
 
     // TODO: explicit event names & handlers for every event
+    // FIXME: these must be static!! per type
     int32_t  _scEventCount;                    // number of supported script events
     String   _scEventNames[MAX_GUIOBJ_EVENTS]; // script event names
     String   _scEventArgs[MAX_GUIOBJ_EVENTS];  // script handler params
--- a/Editor/AGS.Types/Character.cs
+++ b/Editor/AGS.Types/Character.cs
@@ -54,7 +54,8 @@ namespace AGS.Types
                 "$$02 character","$$03 character","Use inventory on character",
                 "Any click on character", "$$05 character","$$08 character", 
                 "$$09 character"}, 
-                new string[] { "Look", "Interact", "Talk", "UseInv", "AnyClick", "PickUp", "Mode8", "Mode9" });
+                new string[] { "Look", "Interact", "Talk", "UseInv", "AnyClick", "PickUp", "Mode8", "Mode9" },
+                "Character *c, CursorMode mode");
         }
 
         public Character()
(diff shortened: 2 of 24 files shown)


ivan-mogilko:
Two Hotspot events that do not work properly:
- Walk On
- Mouse Over

they are handled differently in the engine, and will need to be fixed. Their parameters should exclude CursorMode, and only have Hotspot pointer.

ericoporto:
Playing a bit with this, it does look the parameters are indeed optional when linking. After this is in, should the templates be updated? I mean for things like (perhaps there are more places...)

```
function cEgo_Interact(Character *c, CursorMode mode)
function cEgo_Look(Character *c, CursorMode mode)
function cEgo_Talk(Character *c, CursorMode mode)
```

ivan-mogilko:
> After this is in, should the templates be updated?

That would be nice to have a reference, yes.
If you think it's necessary to give an example of using this argument, one of these placeholder functions could refer Character.Name in displayed text somehow.

PR #2064, opened 2023-07-30
Engine: add support for hardware cursors

### LINKED ISSUE #1686: Support directly using Operating System (hardware) cursors
**Describe the problem**
Nowadays with monitors that uses different refresh rates, some with 144Hz, most still 60Hz, players may be used to their own computer refresh rates. While they may not care about the actual game refresh rate, the input system may imply a different feedback, in this case the mouse cursor may be perceived disconnected from their input if they are used to it.

**Suggested change**
The feature here is to provide the possibility of using OS cursor directly instead of rendering in engine. 

- probably leverage `SDL_CreateColorCursor`
  - This may fail with returning a nullptr, and SDL_error will message "Cursors are not currently supported", this may be a config error in the particular platform if the person builds from source and not turn this on, usually only in very old SDL2 in old Linux packages.
- possible optionally (game or winsetup or per platform? )
- applying custom cursors so it works indepently from the game
- mouse cursor though will probably render in between pixels
- needs to be scaled like the game
  - possibly using an AA aware method if the game has bilinear filtering set (but it seems we don't have this with alpha in bitmap form)
- make sure it's destroyed if game crashes so that regular cursor doesn't change, this looks like is taken care by SDL2, but just need to recheck

**Additional context**
- https://steamcommunity.com/app/1346360/discussions/0/3416557114763335804/?tscn=1655196272
- https://www.adventuregamestudio.co.uk/forums/index.php?topic=59842.msg636647042#msg636647042

ericoporto:
fix #1686

decided to leave it open here just so it's easier to find. There are design issues to solve and this will also probably actually be in ags4 branch - but for testing ideas, I find it easier to work in master, so while it's draft may be better to work in this way.


ericoporto:
for some reason using alt+enter is crashing any game now with an `Error: failed to restore graphics mode` message. :/

ivan-mogilko:
> for some reason using alt+enter is crashing any game now with an `Error: failed to restore graphics mode` message. Edit: uhm, it appears only when using DirectX driver, it works fine when using OpenGL.

This happens in master branch after aa0647dfbde7eb9103a515bd32b8ad8fb099d028, which was a part of #2223.
Might be some resource is not released before resetting device. I will look into this.

ivan-mogilko:
Okay, the D3D reset should be fixed: 5994b9022166f0ffa8e24497b852f86d1c5423e1

ericoporto:
> Okay, the D3D reset should be fixed: https://github.com/adventuregamestudio/ags/commit/5994b9022166f0ffa8e24497b852f86d1c5423e1

Thanks! This indeed fixed the error!

Issue #2062, opened 2023-07-24
Editor: implement IWorkProgress in Progress form, run task asynchronously


ivan-mogilko:
This is a revision of previously cancelled #2034

**Current situation**

In the Editor there are two dialogs used to display the progress of some action. They differ in looks, in what they may display, in how their progress indicator may change, etc.

One is an older [BusyDialog](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/BusyDialog.cs). It has a animating "busy" icon, and is designed to run on a separate thread, which prevents locking the application up.
BusyDialog is designed to run a procedure **asynchronously**, on its own thread, hence keeping the Editor UI refreshed (although blocked).
Later I added a small interface called [IWorkProgress](https://github.com/adventuregamestudio/ags/blob/941bb54f3859734fc1f287a4284b906c1193edb7/Editor/AGS.Editor/GUI/BusyDialog.cs#L14) that lets user change the current/total counters and assign a message dynamically. I used it in some procedures, such as [WriteSpriteFileFromSources](https://github.com/adventuregamestudio/ags/blob/941bb54f3859734fc1f287a4284b906c1193edb7/Editor/AGS.Editor/Utils/SpriteTools.cs#L617). (Technically it is now passed as an argument to any procedure run by BusyDialog, but it's not utilized in majority of these yet.)

Another is a newer [Progress](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/Progress.cs) form. 
Progress form is designed to be displayed as a child of another control, run **inside** a procedure, and let procedure change its state directly. The procedures using it seem to all run synchronously, not letting Editor to refresh itself (although this is not a peculiarity of the Progress control itself).

**Proposal**

It is clear that the choice of a project display form depends on a case. But at the same time the process itself should not depend on this choice. Therefore I propose to do 2 changes:
1. Implement IWorkProgress in Progress form (maybe rename this to a ProgressControl, or similar, for clarity), expanding this interface as necessary.
2. Change all procedures that currently want to access a Progress control, make them accept IWorkProgress interface instead and work with that.
3. Change the way procedures using Progress run: make them all run on a separate thread, to keep Editor refreshed. Note that it's possible to make Progress control own this thread, similarly to how BusyDialog does this, but this is *not necessary* (it's an open question).
[...]

PR #2074, opened 2023-08-05
Feature reimport audio clips

### LINKED ISSUE #778: Reimport audio from source
The sprite UI has a right click "reimport sprite from source", but there's nothing similar in the audio, currently you have to delete the audiocache folder and then press f7 to force an audio clip you updated to be reloaded.

### LINKED ISSUE #2051: Editor: allow to change the audioclip's source filename
Per user's request: it's possible to change the sprite's source either importing a new sprite over the existing sprite, keeping the same ID, or modifying the Source Filename property directly. But it's not possible to do the same for the Audio clips.
The use case is when you have a number of placeholder clips in your projects, and want to replace them with real sounds without losing their IDs, so not deleting existing item and creating a new one.

The proposal is then to make AudioClip's SourceFilename property editable, with custom editor type that opens a file dialog upon clicking a "..." button, in a similar way to how it's done with sprites.

Note, modifying property type is trivial, but the actual issue would be in forcing a audio file's recache.
Also see #778.

ericoporto:
I will start saying that I don't know if this is a good idea, but what this does is it adds a right click option to force reimport of one audio clip or force reimport all audio clips.

The way it works is it uses the clip source file name and copies the file to the clip cache file name, if it exists.

Maybe fix #778 and someone else can write a  better issue for the things that are in that issue that came up later on.

Also fix #2051

### DIFF
--- a/Editor/AGS.Editor/Components/AudioComponent.cs
+++ b/Editor/AGS.Editor/Components/AudioComponent.cs
@@ -11,6 +11,7 @@ namespace AGS.Editor.Components
         private const string COMMAND_ADD_AUDIO = "AddAudioClipCmd";
         private const string COMMAND_PROPERTIES = "PropertiesAudioClip";
         private const string COMMAND_RENAME = "RenameAudioClip";
+        private const string COMMAND_REIMPORT = "ReimportAudioClip";
         private const string COMMAND_DELETE = "DeleteAudioClip";
         private const string COMMAND_CHANGE_ID = "ChangeAudioID";
         private const string SPEECH_NODE_ID = "DummySpeechNode";
@@ -101,6 +102,11 @@ namespace AGS.Editor.Components
                 {
                     DeleteSingleItem(clipToDelete);                    
                 }
+            }
+            else if (controlID == COMMAND_REIMPORT)
+            {
+                AudioClip clipToReimport = _items[_rightClickedID];
+                CommandForceReimportOfAudioClip(clipToReimport);                
             }
             else if (controlID == COMMAND_CHANGE_ID)
             {
@@ -497,6 +503,28 @@ namespace AGS.Editor.Components
             }
         }
 
+        private void CommandForceReimportOfAudioClip(AudioClip clip)
+        {
+            if (!ForceReimportOfAudioClip(clip))
+            {
+                _guiController.ShowMessage("Failed to reimport Audio Clip " + clip.ScriptName + ".", MessageBoxIconType.Warning);
+            }
+        }
+
+        private bool ForceReimportOfAudioClip(string sourceFileName, string cacheFileName)
+        {
+            if (File.Exists(sourceFileName) && !string.IsNullOrEmpty(cacheFileName))
+            {
+                return Utilities.SafeCopyFileOverwrite(sourceFileName, cacheFileName, true);
+            }
+            return false;
+        }
+
+        private bool ForceReimportOfAudioClip(AudioClip clip)
+        {
+            return ForceReimportOfAudioClip(clip.SourceFileName, clip.CacheFileName);
+        }
+
         private void AddAudioClipToListIfFileNeedsToBeCopiedFromSource(AudioClip clip, PreCompileGameEventArgs evArgs, List<AudioClip> filesToCopy, List<string> fileNamesToUpdate)
         {
             string compiledFileName = clip.CacheFileName;
@@ -660,6 +688,8 @@ namespace AGS.Editor.Components
                 menu.Add(new MenuCommand(COMMAND_CHANGE_ID, "Change Clip ID", null));
                 menu.Add(new MenuCommand(COMMAND_RENAME, "Rename", null));
                 menu.Add(new MenuCommand(COMMAND_DELETE, "Delete", null));
+                menu.Add(MenuCommand.Separator);
+                menu.Add(new MenuCommand(COMMAND_REIMPORT, "Force Reimport", null));
                 menu.Add(MenuCommand.Separator);
                 menu.Add(new MenuCommand(COMMAND_PROPERTIES, "Properties", null));
             }
--- a/Editor/AGS.Editor/Components/AudioComponent.cs
+++ b/Editor/AGS.Editor/Components/AudioComponent.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
+using System.Linq;
 using AGS.Types;
 
 namespace AGS.Editor.Components
@@ -9,6 +10,7 @@ namespace AGS.Editor.Components
     class AudioComponent : BaseComponentWithFolders<AudioClip, AudioClipFolder>, IProjectTreeSingleClickHandler
     {
         private const string COMMAND_ADD_AUDIO = "AddAudioClipCmd";
+        private const string COMMAND_REIMPORT_ALL = "ReimportAllAudioClipCmd";
         private const string COMMAND_PROPERTIES = "PropertiesAudioClip";
         private const string COMMAND_RENAME = "RenameAudioClip";
         private const string COMMAND_REIMPORT = "ReimportAudioClip";
@@ -90,6 +92,10 @@ namespace AGS.Editor.Components
                 {
                     ImportAudioFiles(selectedFiles);
                 }
+            }
+            else if (controlID == COMMAND_REIMPORT_ALL)
+            {
+                CommandForceReimportOfAllAudioClips();
             }
             else if (controlID == COMMAND_RENAME)
             {
@@ -525,6 +531,54 @@ namespace AGS.Editor.Components
             return ForceReimportOfAudioClip(clip.SourceFileName, clip.CacheFileName);
         }
 
+        private void DoForceReimportOfAudioClips(IEnumerable<AudioClip> audioClips, IWorkProgress progress, out List<AudioClip> failedClips)
+        {
+            failedClips = new List<AudioClip>();
+            progress.Total = audioClips.Count();
+            progress.Current = 0;
+            foreach (AudioClip clip in audioClips)
+            {
+                if (!ForceReimportOfAudioClip(clip))
+                {
+                    failedClips.Add(clip);
+                }
+                progress.Current++;
+            }
+        }
+
+        private void CommandForceReimportOfAllAudioClips()
+        {
+            List<AudioClip> failedClips = new List<AudioClip>();
+            IEnumerable<AudioClip> allAudioClips = _agsEditor.CurrentGame.RootAudioClipFolder.AllItemsFlat;
+
+            try
+            {
+                BusyDialog.Show("Please wait while the Audio Clipes are reimported ...",
+                    new BusyDialog.ProcessingHandler(
+                        (IWorkProgress progress, object o) => {
+                            DoForceReimportOfAudioClips(allAudioClips, progress, out failedClips);
+                            return null;
+                        }), null);
+            }
+            catch (Exception e)
+            {
+                Factory.GUIController.ShowMessage(
+                    "The reimport of all audio clips was interrupted by error.\n\n" + e.Message,
+                    MessageBoxIconType.Error);
+                return;
+            }
+
+            Factory.GUIController.ClearOutputPanel();
+            if (!failedClips.Any()) return; // success!
+
+            CompileMessages errors = new CompileMessages();
+            errors.AddRange(failedClips.
+                Select(ac => new CompileWarning("Failed audio clip reimport of " + ac.ScriptName, ac.SourceFileName, 0)).
+                ToList<CompileMessage>());
+
+            Factory.GUIController.ShowOutputPanel(errors);
+        }
+
         private void AddAudioClipToListIfFileNeedsToBeCopiedFromSource(AudioClip clip, PreCompileGameEventArgs evArgs, List<AudioClip> filesToCopy, List<string> fileNamesToUpdate)
         {
             string compiledFileName = clip.CacheFileName;
@@ -730,6 +784,7 @@ namespace AGS.Editor.Components
         protected override void AddNewItemCommandsToFolderContextMenu(string controlID, IList<MenuCommand> menu)
         {
             menu.Add(new MenuCommand(COMMAND_ADD_AUDIO, "Add audio file(s)...", null));
+            menu.Add(new MenuCommand(COMMAND_REIMPORT_ALL, "Force reimport all file(s)", null));
         }
 
         protected override void AddExtraCommandsToFolderContextMenu(string controlID, IList<MenuCommand> menu)
--- a/Editor/AGS.Editor/Components/AudioComponent.cs
+++ b/Editor/AGS.Editor/Components/AudioComponent.cs
@@ -14,6 +14,7 @@ namespace AGS.Editor.Components
         private const string COMMAND_PROPERTIES = "PropertiesAudioClip";
         private const string COMMAND_RENAME = "RenameAudioClip";
         private const string COMMAND_REIMPORT = "ReimportAudioClip";
+        private const string COMMAND_REPLACE = "ReplaceAudioClip";
         private const string COMMAND_DELETE = "DeleteAudioClip";
         private const string COMMAND_CHANGE_ID = "ChangeAudioID";
         private const string SPEECH_NODE_ID = "DummySpeechNode";
@@ -75,7 +76,7 @@ namespace AGS.Editor.Components
         private void RecreateDocument()
         {
             _editor = new AudioEditor();
-            _document = new ContentDocument(_editor, "Audio", this, AUDIO_CLIP_TYPE_ICON);             
+            _document = new ContentDocument(_editor, "Audio", this, AUDIO_CLIP_TYPE_ICON);
         }
 
         public override string ComponentID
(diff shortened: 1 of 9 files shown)


ivan-mogilko:
Yes, this probably a better way for now, because the automatic reimport seems unreliable and confusing to users.

ericoporto:
In my computer it's pretty fast though because my storage is super fast.

But the version that acts on all clips, maybe requires a progress bar, but I didn't know how to add one. And also how fake it being slow for testing.

ivan-mogilko:
> But the version that acts on all clips, maybe requires a progress bar, but I didn't know how to add one. And also how fake it being slow for testing.

There's a very recent ticket which mentions existing progress bar options:
https://github.com/adventuregamestudio/ags/issues/2062

I believe for a blocking process like this using BusyDialog would be best.

ericoporto:
Hey, I made the progress bar change [here](https://github.com/adventuregamestudio/ags/compare/774c9be66fc26d78e51f5a0ad9c6df3aa5ef9427..81905965015e607eae051a4ecb82cbdc190c7dbf), I also made so that if many audio clips fail, they will be logged to a file (so it's easier to follow on them).

ivan-mogilko:
> I also made so that if many audio clips fail, they will be logged to a file (so it's easier to follow on them).

This is something that Editor never did before.

Have you considered printing to the output panel?

ericoporto:
> Have you considered printing to the output panel

I have, but I thought a file was easier to read. Plus, you can act later and not break workflow now.

If the output panel was a pure text panel that one could copy paste, I would place it there, but it's a weird table...

I can just remove this.

> It's better to return the list of failed clips to the caller, and let it do what is necessary to it.

I wanted to have the responsibility in the caller but I couldn't figure how to return the list of failed clips in this case.

ivan-mogilko:
> If the output panel was a pure text panel (like engine log) that one could copy paste, I would place it there, but it's a weird table... I don't know how to use it most of the time.

Output panel may be pasted a list of items, and you have a list files. User may then copy its contents to the clipboard.
We could add more options if necessary (like saving it to a file).

> I wanted to have the responsibility in the caller but I couldn't figure how to return the list of failed clips in this case.

AFAIK the BusyDialog's process is allowed to return `object` as a result. This object may be cast to any type, meaning it's possible to return a struct containing any data.

Alternatively, using a parameter with `out` modifier.

ericoporto:
I made it report failures to the Output panel as warnings and also moved the reporting responsibility to the caller. I could not figure how to return a thing for the process handler, so instead I used the out parameter that I find easier to use.

ericoporto:
Since it was in the same theme, took the opportunity to also add a fix to #2051

Editor/AGS.Editor/Components/AudioComponent.cs
```diff
@@ -497,6 +503,50 @@ private void SetActualAudioClipProperties(AudioClip clip, int inheritedVolume, b
...
+                _guiController.ShowMessage("Failed to reimport Audio Clip "  + clip.ScriptName + ".", MessageBoxIconType.Warning);
+            }
+        }
+
+        private bool ForceReimportOfAudioClip(AudioClip clip)
+        {
+            string sourceFileName = clip.SourceFileName;
+            string cacheFileName = clip.CacheFileName;
+            bool reimportOk = false;
+            if(File.Exists(sourceFileName) && !string.IsNullOrEmpty(cacheFileName))
```
ivan-mogilko, line 525: This whole code from here and until the end of the function looks like something that would be better as a generic method "try to replace a file and fallback to backup if failed". There's [Utilities class](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/Utils/Utilities.cs) that might be a good place to put such method (alternatively there could be FileUtilities class, but that's a separate refactor task).

Editor/AGS.Editor/Components/AudioComponent.cs
```diff
@@ -547,6 +553,55 @@ private bool ForceReimportOfAudioClip(AudioClip clip)
...
+            progress.Total = audioClips.Count();
+            progress.Current = 0;
+            int i = 0;
+            foreach (AudioClip clip in audioClips)
+            {
+                if (!ForceReimportOfAudioClip(clip))
+                {
+                    failedClips.Add(clip);
+                }
+                progress.Current = ++i;
```
ivan-mogilko, line 568: Would not `progress.Current++` work, without `i` variable?

Editor/AGS.Editor/Components/AudioComponent.cs
```diff
@@ -602,6 +602,35 @@ private void CommandForceReimportOfAllAudioClips()
...
+            if (string.IsNullOrEmpty(selectedFile) || !File.Exists(selectedFile)) return;
+
+            string fileExtension = Path.GetExtension(selectedFile).ToLower();
+            if (!_fileTypeMappings.ContainsKey(fileExtension)) return;
+
+            selectedFile = Utilities.GetRelativeToProjectPath(selectedFile);
+            clip.FileType = _fileTypeMappings[fileExtension];
+            clip.SourceFileName = selectedFile;
+            Path.ChangeExtension(clip.CacheFileName, fileExtension);
+
```
ivan-mogilko, line 619: I suggest to pick out a method that assigns a source to the clip, and share the same code from both ReplaceAudioClipSource and CreateAudioClipForFile. This is to ensure that whenever a source file is set, same operations are performed. For example, CreateAudioClipForFile also updates FileLastModifiedDate and game.FilesAddedOrRemoved, but you do not do that here.
Such shared method could also call ForceReimportOfAudioClip (or maybe you could rename that into something more suitable. like "recreate cache file").

Editor/AGS.Editor/Components/AudioComponent.cs
ericoporto: I think it would, but I saw this pattern in the other places with it and thought it was necessary for some reason.

Editor/AGS.Editor/Components/AudioComponent.cs
ericoporto: I think on creation there is some ID stuff being made along and some other things we don't want to change, to reuse code from there I need to think a bit on this.

ivan-mogilko: > I think on creation there is some ID stuff being made along and some other things we don't want to change, to reuse code from there I need to think a bit on this.

Yes of course, this refers only to a limited set of actions. Probably starting with this line:
https://github.com/adventuregamestudio/ags/blob/c03e186cdcb3f35c707d53301ff97bff6359d515/Editor/AGS.Editor/Components/AudioComponent.cs#L323

Editor/AGS.Editor/Components/AudioComponent.cs
```diff
@@ -602,6 +602,35 @@ private void CommandForceReimportOfAllAudioClips()
...
 
+        public void ReplaceAudioClipSource(AudioClip clip)
+        {
+            string dir = Path.GetDirectoryName(clip.SourceFileName);
+            if (!Directory.Exists(dir)) dir = Factory.AGSEditor.CurrentGame.DirectoryPath;
+            string selectedFile = _guiController.ShowOpenFileDialog("Select audio to add", AUDIO_FILES_FILTER, dir);
+            if (string.IsNullOrEmpty(selectedFile) || !File.Exists(selectedFile)) return;
+
+            string fileExtension = Path.GetExtension(selectedFile).ToLower();
+            if (!_fileTypeMappings.ContainsKey(fileExtension)) return;
```
ivan-mogilko, line 613: This and above failure cases should also report a visible error, preferably with a reason. Right now if I select an unsupported file, the operation fails without explanation.

ivan-mogilko:
I think this is generally good; I only posted few suggestion about sharing some code and missing error reporting.

Something I'd like to propose is to add a context menu command for replacing a source file, along with "force reimport". I think that will make this command more visible than if it would stay only in properties.

ericoporto:
ah, that is a good idea! I will add that!

Editor/AGS.Editor/Components/AudioComponent.cs
ericoporto: It looks like it works fine, just curious that it's type is `int?` instead of just `int`, probably means it can be null somehow?

ivan-mogilko: > It looks like it works fine, just curious that it's type is `int?` instead of just `int`, probably means it can be null somehow?

yes, `int?` means that it begins as null and may be assigned a null. But since you assign it with `0` earlier, this should not be a problem here.

iirc I made it nullable to indicate that the process does not feature a progress counter.

Editor/AGS.Editor/Components/AudioComponent.cs
ericoporto: I could not make such a big method and went for a smaller scope, due to how cache is generated from the index and not a string one can easily change.

ericoporto:
I think I added the mentioned things! If I forgot something just comment.

ivan-mogilko:
I'm afraid we had a misunderstanding about the ReplaceAudioClipSource and CreateAudioClipForFile functions. I realize that did not explain properly at that time. The changes do more than I expected, but at the same time don't share few things that I supposed should be shared. My meaning was to share a small piece of code that assigns properties to a clip.

Refering to the previous commit, ReplaceAudioClipSource:
```
clip.FileType = _fileTypeMappings[fileExtension];
clip.SourceFileName = selectedFile;
```
CreateAudioClipForFile:
```
newClip.SourceFileName = Utilities.GetRelativeToProjectPath(sourceFileName);
newClip.FileType = _fileTypeMappings[fileExtension];
newClip.FileLastModifiedDate = File.GetLastWriteTimeUtc(sourceFileName);
_agsEditor.CurrentGame.FilesAddedOrRemoved = true;
```

in addition there are two operations made when the new clip is created, which I suppose could also be done when source is replaced:
```
Utilities.CopyFileAndSetDestinationWritable(sourceFileName, newClip.CacheFileName);
Utilities.TryDeleteFile(Path.Combine(AGSEditor.OUTPUT_DIRECTORY,
                Path.Combine(AGSEditor.DATA_OUTPUT_DIRECTORY, AGSEditor.AUDIO_VOX_FILE_NAME)));
```


Your latest commit is puzzling to me, there are things that I do not understand from the first glance, like the recreation of AudioClip, and reading bytes from audio file... I must try reading it again.

Removing a constructor from AudioClip is not a good thing (removing things from AGS.Types overall should not be done so easily, because theoretically used by editor plugins and third-party tools), if you really need another constructor then it's better to add a new one, but leaving existing one(s).

ivan-mogilko:
Hmm, I need to review and rethink this again, because the way code got shared between ReplaceAudioClipSource and CreateAudioClipForFile was puzzling me at first. I did not realize that one of the properties is also a part of AudioClip constructor parameters.

One thing that I must mention though: removing a constructor from AudioClip is not a good thing; removing `public` things from AGS.Types overall should not be done so easily, because theoretically used by editor plugins and third-party tools. If you really need another constructor then it's better to add a new one, but leaving existing one(s).

ericoporto:
I force pushed added the constructor back in-place, but it's unused.

ivan-mogilko:
So, the operation of setting a source file to a clip may be viewed as having following steps:
- Test that the source file is valid;
- Assign a new source and other necessary properties to a AudioClip object;
- Perform any necessary post-step (update the cache, delete the audio.vox).

In your latest change these steps are reversed: you first assign new properties, then test if the file is valid. This is why you have to do this `RecoverFailClipWithMessage` thing. This results in a confusing code.

My suggestion is to perform these steps in a clear order, where first you validate the file, then assign the clip, and update the cache. This may require making 2 shared methods instead of one: one of them tests for file existence and supported extension, another updates clip properties.

Another thing, I do not think that testing that updating audiocache was a success is necessary at this point at all, because this error logically does not prevent from setting a new source.

ericoporto:
Just to explain, all that is because you may have like an mp3 file and replace with an ogg, which had a different extension and then will change the cache file name. If you would be alright with removing the "smartness" in how cache file name is generated by magic in AGS types, then this setting could be moved to something internal and things would be more sensible - I don't like how changing the source filename extension changes the cache file name by magic instead of a explicit Change Extension call.

ivan-mogilko:
Are you referring to the whole code, or to testing the cache update specifically?

I forgot about the extension, so copying a file is not enough to replace the cache, it has to first remove the previous file and then copy a new file over. That's alright.

My primary intent here is to keep similar things consistent. For instance, if adding a new clip based on a source file does not depend on whether placing file in a cache was a success or not, so should replacing the source file in a clip not depend on caching result. The caching problems may be still reported to the user, of course, but should not prevent from replacing a source, in my opinion. In the worst case the clip will be left without a cache entry, but this situation is similar to the case when a user clears the cache.

ivan-mogilko:
I honestly still do not understand...

> Because the cache file name rule is this magic here, I have to change the clip source filename to grasp the new cache filename and then I can try to do the file operations, and then if things failed I have to rechange the audio clip to revert things.

But why do you need to do all this?

I thought this is what is necessary:
- test if source file is valid
- save previous cache name
- update the audio clip
- remove the previous cache file using saved name.
- run the cache update. If it has problems, report to user, but keep the new source.

ericoporto:
I redid all the commits so the fixes in progress and the rest were in their own commits. 

I remade how the Replace Source File works by adding two extra static methods in the AudioClip so I could have raw access to the Cache File Name logic.

ivan-mogilko:
> > If it has problems, report to user, but keep the new source.

> I just didn't do this, because it fails (throws) before in my code, it will instead return before modifying the audio clip, so "similar to recovering to previous state" but without actually recovering. I also prefer it like this - to not set the source to an invalid value.

It seems we still have a misunderstanding here. I the proposed steps the source would not be set to an invalid value.
The cache is a separate stage in the project. Setting source to audio clip and updating a cache should not be related. Any errors during cache update should not invalidate audio clip.

ivan-mogilko:
>> If it has problems, report to user, but keep the new source.

> I just didn't do this, because it fails (throws) before in my code, it will instead return before modifying the audio clip, so "similar to recovering to previous state" but without actually recovering. I also prefer it like this - to not set the source to an invalid value.

It seems we still have a misunderstanding here, or maybe it's a disagreement on how the editor should work.

But I will stop here, because it takes more time than necessary.

Editor/AGS.Editor/Components/AudioComponent.cs
```diff
@@ -323,25 +325,67 @@ private void ImportAudioFiles(string[] selectedFiles)
...
+            }
+
+            if(!File.Exists(fullSourceFileName))
+            {
+                throw new FileNotFoundException("Source file doesn't exist.");
+            }
+            else
+            {
+                byte[] audioData = File.ReadAllBytes(fullSourceFileName);
+                if (audioData.Length == 0) throw new FileLoadException("New source file is empty.");
```
ivan-mogilko, line 342: Why do you think this test is necessary, and is there other way to test other than loading whole file into memory?

ivan-mogilko, line 342: I found there's a FileInfo class that can let know the length of the file without reading its data:
https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo?view=netframework-4.6

ericoporto: Thanks, I added this in [the push below](https://github.com/adventuregamestudio/ags/compare/9a683d6b52f950ce9b2b3bca36b61a21db95b0c7..daff42c9256d63d3a45e9ae27b0603c560d66be3)

ivan-mogilko:
I think I must clarify the potential consequences of adding the static method `AudioClip.GetCacheFileName`, for the reference.

Previously the method of generating a cache name was hidden from public, it just required a AudioClip structure, but did not say which of its fields are used in this name. The implementation details were private.

After this function is added, it now explicitly tells which fields are used. This means that if we ever would like to change how CacheName is generated, for example - generate from AudioClip's Name instead of index, then we would have to also change this static method's signature.

The two alternatives to this that I see are:
1. Create a temporary AudioClip object to learn the future cachename.
2. Introduce a static callback property in AudioClip, that should return a CacheName, and do the implementation inside the editor.

ericoporto:
The Cachename is based on Fixed ID and the source filename extension, the Fixed ID is... Fixed, so the operations like changing ID actually change the other id property. The source filename, well, I guess I am changing now so I can track that too.

So I could move the GetCacheName static method into the AudioComponent and set both cacheFileName and cacheFileNameWithoutPath of the audioClip at appropriate time. The only issue is this would require cacheFileName and cacheFileNameWithoutPath to be writable (or keep both read only but add a method for setCache.

I could also instead use the strategy of fake audioClip creation - but I will say I find it really ugly too.

ivan-mogilko:
Making CacheFileName a writeable property sounds like a better idea to me (i forgot to add this to the list of options).
CacheFileNameWithoutPath might remain readonly, as it is derived directly from CacheFileName.

I might try thinking about any bad consequences, but I cannot see anything critical at the first thought.

ivan-mogilko:
> I could also instead use the strategy of fake audioClip creation - but I will say I find it really ugly too.

By the way, I don't think it's ugly. AudioClip is a pure struct wrapping some data. Creating cache name is something like generating a Hash. So creating a temp struct for this purpose seems fine.

ericoporto:
Made an attempt to move the cache filename logic into the Editor in https://github.com/adventuregamestudio/ags/pull/2074/commits/0c59adef88ffc0a5f93557e263c80d669b1fae88, by going the _4. Make CacheFileName a settable property_ route. I also reversed the CacheFileName without path relationship so the responsability for telling about the cache dir is in the Editor.

ivan-mogilko:
Tested both commands (and property editing), looks working.

PR #2075, opened 2023-08-06
Upgrade SDL to 2.28.2

ericoporto:
I think we want to follow SDL 2 releases, from what I monitored, the regressions that were introduced since 2.24.1 are fixed.

I tested on macOS, Windows and Android, but not thoroughly. I haven't tested on Linux yet.

Emscripten is the only one not upgraded here - I haven't upgraded Emscripten upstream yet, I will do that and adjust later, but possibly in a new PR.

Ah, this SDL2 upgrade was a bit painful in my Android Studio and CLion IDEs, I had to delete all cache/intermediate build files to get it to build correctly. In the CI, because things are always built cleanly, this is already taken care.

Also, it's possibly a good idea if someone has an Android device to check if things are still working ok - I only tested in emulator, and it seems alright, but there are lots of android changes here in SDL2, although nothing changed API or behavior wise.

### DIFF
--- a/Android/library/runtime/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
+++ b/Android/library/runtime/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
@@ -186,7 +186,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
     // Because on Chromebooks we show up as a dual-mode device, it will attempt to connect TRANSPORT_AUTO, which will use TRANSPORT_BREDR instead
     // of TRANSPORT_LE.  Let's force ourselves to connect low energy.
     private BluetoothGatt connectGatt(boolean managed) {
-        if (Build.VERSION.SDK_INT >= 23) {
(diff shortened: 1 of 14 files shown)


ericoporto:
This looks ok, but it would be nice if more people tested to see if there's some regression introduced.

ericoporto:
Although this makes no changes to the Cirrus.yml file, I would like to rebase this after #2079 is merged.

ericoporto:
Yeah, the Android SDL project is from SDL2 repository and it needs to be updated accordingly - matching the SDL version number.

Additionally, there's a tiny patch to keep per letter writing working - in Android you can compose a word in the keyboard and later send the entire word in the field, and SDL2 goes the route of supporting that through the SDL text input. My tiny patch gives the previous per letter behavior back as the other requires a few changes in the engine - the same as the ones mentioned in the IME input issue.

ivan-mogilko:
Could not test this on my Android device atm, because of #2095.

ivan-mogilko:
Fixed #2095, @ericoporto could you rebase once more?

ericoporto:
I rebased this!

PR #2077, opened 2023-08-07
CI: CMake desktop flows to GitHub Actions

ericoporto:
Removes macOS and Linux CMake builds from CirrusCI.

Moves these to GitHub Actions. GH Actions won't trigger because they aren't yet in the repository.


ericoporto:
I am closing this and will try a different way before.

PR #2078, opened 2023-08-07
4.0 Editor: move translation src migration process to TranslationComponent

ivan-mogilko:
AGS.Types had been used as a Editor's "API" for a long while now, it's commonly linked to Editor plugins and third-party tools (such as FontEditor app). Therefore we cannot make any assumptions about filesystem around AGS.Types library, and should not implicitly delete or move project files.

Besides that, implicitly loading from legacy format and deleting files in random time is inconsistent with how Editor normally works.

With this change I move migration from old TRS format to the new PO format into the TranslationComponent class, where it runs it triggered by GamePostLoad event.

In addition, rise format index, and add a constant that tells the first version the PO became new translation source format.

Made a PR because I still have concerns that I could miss something.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -106,14 +106,16 @@ namespace AGS.Editor
          * 
          * 3.99.99.00     - BlendMode for various objects, Character.Transparency.
          * 3.99.99.01     - Open rooms
+         * 3.99.99.07     - PO translations
          *
         */
-        public const int    LATEST_XML_VERSION_INDEX = 3999901;
+        public const int    LATEST_XML_VERSION_INDEX = 3999907;
         /// <summary>
         /// XML version index on the release of AGS 4.0.0, this constant be used to determine
         /// if upgrade of Rooms/Sprites/etc. to new format have been performed.
         /// </summary>
-        public const int    AGS_4_0_0_XML_VERSION_INDEX = 3999901;
+        public const int    AGS_4_0_0_XML_VERSION_INDEX_OPEN_ROOMS = 3999901;
+        public const int    AGS_4_0_0_XML_VERSION_INDEX_PO_TRANSLATIONS = 3999907;
         /*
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
Agh, I got distracted, and forgot about my own PR.

So, I've been thinking about the current situation, and what irritates me is that the project migration is mostly done in the Editor application, where it has distinct tasks for that, which know what files to check, what to modify, and what to delete in the upgraded project. But Translations are doing their own upgrade right in AGS.Types. This is inconsistent, and imo also raises questions about the range of responsibility of AGS.Types library.

The way I see this, it's also a problem of how translations are added into the project: they cannot be *imported*, only created. Rooms, for instance, may be imported, and their migration is done as a part of this import, regardless of whether it's done when upgrading whole project, or when importing an individual room from another location. But there's no "import" action for translations.

This opens a potential for situations where the data migration may occur unexpectedly. For example, a user creates a Translation item in the editor, then goes to file explorer, deletes the PO file, and copies a TRS file over into the project dir. Next time a translation is loaded, it suddenly:
* loads from TRS, even though it was supposedly created as a new format in a new project;
* saves a new PO;
* deletes the TRS.

I realize these are narrow cases, but this means that the editor's behavior is kind of... *volatile* (not sure which term to use here), where it may recreate project's item from a matching file and delete the previous file.

I recall that in the past @morganwillcock was suggesting that in the future the game project could be defined simply by source files of supported format present in a directory (and IDE would hook them up automatically). Whether this is a good or a bad approach is an interesting matter for discussion, but *right now* the game project's contents are defined by the lists of items found in Game.agf. This is why I *suppose* that for *consistency* the upgrading of these items should also be an explicit operation. That is, an operation done in two cases:
* When older project is loaded in the newer editor for the first time (and we may test the project version index to know that it has to be upgraded);
* When a user commands to *import* a translation from an older file, e.g. using a menu command.

ericoporto:
> This is why I suppose that for consistency the upgrading of these items should also be an explicit operation.

This sounds reasonable.

PR #2082, opened 2023-08-10
AGS 4: When saving Room only save images that were actually modified

### LINKED ISSUE #2071: AGS 4: saving a room sometimes changes their background and mask file binary contents because of pHYs chunk
**Problem**

Testing recent ags4, I noticed that git sometimes detects changes to room image files, such background and mask pngs after I save the room, or rebuild whole game, although neither of these were changed in the editor.

~The changes seem to occur always in IHDR section of PNG.~
~The changes occur in one of the leading sections, prior to IDAT. There are sRGB, gAMA and pHYs sections present, but maybe something else.~
As [explained in the comment](https://github.com/adventuregamestudio/ags/issues/2071#issuecomment-1671813956), this is caused by pHYs chunk.

At first I was suspecting recent change to loading PNGs (#1772), but reverting to a previous version did not fix this.

**Steps to reproduce**
1. Open a project in latest ags4.
2. Make sure there's at least 1 room.
3. Make a backup copy of room N files.
4. Do Build -> rebuild all files.
5. Run a diff tool (like kdiff or similar) on old and new room image files, and notice that there are small changes.
Also if you use git, it will detect differences.

You may repeat Rebuild all. The results can be strange. Sometimes it changes the same file 1-2 more times in a row. Sometimes it stops. But restarting the Editor, and running Rebuild all again will create new differences.

**Expected behavior**

Ideally, if nothing was changed, then files should remain identical. Right now this will make working with source control annoying.

ivan-mogilko:
Fixes #2071 (for the most part).

This is an attempt to ensure that the RoomComponent does not unnecessarily resave room images (backgrounds, masks), unless they were actually edited by the user inside the room editor. (For room bgs this is only a case when a new bg was added.)

This is achieved by pairing each cached bitmap (bg and mask) with a Modified flag, and making sure they are set accordingly.

Did a small refactor in RoomComponent to reduce code duplication and make it easier to track actual modifications to masks.

Did few changes to the order of actions when recompiling and saving. Also removed saving script file when recompiling all rooms, as that's likely not necessary anymore since we have a room conversion on import. (According to the comment in code, this was previously done because importing old 2.72 rooms could lead to automatic changes in script.)

NOTE: Room class in AGS.Types was not changed at all, these mentioned Modified flags are added inside RoomComponent, and not exposed in any interface.

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -551,7 +551,13 @@ namespace AGS.Editor.Components
 					return;
 				}
 
-				PerformPreSaveChecks(room);
+                // First test the Room for valid content
+                if (!EnsureScriptNamesAreUnique(room, errors))
+                    return;
+
+                // Do any automatic fixups that do not raise errors
+                // TODO: perhaps a wrong place to do so, find another time and place for this?
+                room.Modified |= (PerformPreSaveChecks(room) > 0);
 
 				if (!_agsEditor.AttemptToGetWriteAccess(room.FileName))
 				{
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@persn could you also check this, if you have spare time?

ericoporto:
I tested this and couldn't get crashes and it looks like it works, but I couldn't reproduce the change in binary contents without this PR reported in #2071.

PR #2084, opened 2023-08-11
Support multiple mouse clicks per game frame

ivan-mogilko:
Implement mouse event buffer similar to key buffer, and process it all in check_controls().
Process Mouse.Click() simulation by sending appropriate events.

This theoretically allows to get multiple number of `on_mouse_click` callbacks per game frame, with different buttons too (in case 2 mouse buttons were pressed simultaneously).

NOTE: I was not able to register more than 1 click with 1 button per game frame, no matter the game speed. ~Possibly this is also restricted by SDL2 somehow (might need to investigate).~ Looks like i'm not clicking fast enough, or my mouse is unable to send clicks fast enough.

It's possible to receive all 3 `on_mouse_click` events with 3 different buttons in a row though.

Also possible to simulate unlimited number of clicks using Mouse.Click() script command.

TODO:
~* investigate why there's no more than 1 event per button in a row; this is either a SDL2 restriction, or some of the engine code is a culprit.~
* update all the special game states by processing all available mouse events in a loop, similar to keys: Dialog options, Inventory, built-in GUI, etc.

### DIFF
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -458,6 +458,7 @@ struct DialogOptions
     bool needRedraw;
     bool wantRefresh;
     bool usingCustomRendering;
+    bool newCustomRender; // using newer (post-3.5.0 render API)
     int orixp;
     int oriyp;
     int areawid;
@@ -484,6 +485,12 @@ struct DialogOptions
     // Process single key event
     // returns whether should continue to run options loop, or stop
     bool RunKey(const KeyInput &ki);
+    // Process all the buffered mouse events;
+    // returns whether should continue to run options loop, or stop
+    bool RunMouseControls();
+    // Process single mouse event
+    // returns whether should continue to run options loop, or stop
+    bool RunMouse(eAGSMouseButton mbut, int mwheelz);
     void Close();
 };
 
@@ -630,6 +637,7 @@ void DialogOptions::Show()
     if (!is_textwindow)
       areawid -= data_to_game_coord(play.dialog_options_x) * 2;
 
+    newCustomRender = usingCustomRendering && game.options[OPT_DIALOGOPTIONSAPI] >= 0;
     orixp = dlgxp;
     oriyp = dlgyp;
     needRedraw = false;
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
> investigate why there's no more than 1 event per button in a row; this is either a SDL2 restriction, or some of the engine code is a culprit.

No, apparently I am not clicking fast enough, or my cheap mouse cannot click so fast.
I was able to get more clicks per frame but only if I set game speed to below 10.

ivan-mogilko:
Was able to test this using [AutoHotKey](https://www.autohotkey.com/), with this primitive script:
```
#Requires AutoHotkey v2.0

^Z::
{
    MouseClick "Left", 500, 500
    MouseClick "Left", 500, 500
    MouseClick "Left", 500, 500
    MouseClick "Left", 500, 500
}

^X::
{
    ExitApp 
}
```
(this make Ctrl+Z send mouse clicks and Ctrl+X stop the ahk script)

ivan-mogilko:
Tested all game states after the code move above, and things seem to still work as expected.

ivan-mogilko:
No, this is wrong; there should be only 1 queue for input events, not separate, because the more input devices will be supported, the worse this will become. I need to redo this.

ericoporto:
I guess this modifies all the skipping in the Joystick PR... Please add a few more comments because right now I can't quite figure it out how it's expected to integrate in this merged queue...

ivan-mogilko:
> I guess this modifies all the skipping in the Joystick PR... Is it expected to have the gamepad and joystick in this merged queue?

Right, and the skipping check should be inside the added `for` loop
```for (InputType type = ags_inputevent_ready(); type != kInputNone; type = ags_inputevent_ready())```

I checked out your pr's commit, and believe it should be trivial to transit to the new code, although manual code merge will be required likely.

I picked out "check controls" code into separate functions everywhere in this PR, it's still kind of ugly, but this is probably the best possible at this point, until #1349 is done.

ericoporto:
Once this is merged, please merge in ags4 too so I can work on adjusting the joystick PR.

ivan-mogilko:
> Once this is merged, please merge in ags4 too so I can work on adjusting the joystick PR.

Yes definitely, I will try to finish this today, only have to test this once more after the recent update.

ericoporto:
Hey, I think I found a bug

https://github.com/adventuregamestudio/ags/pull/2084/commits/3a38e69212378a75064265eb14fb158067f9a069

Before this, AGS handled clicks on mouse UP, but after this it starts handling on mouse DOWN.

ivan-mogilko:
I see, I would have to fix and re-release the latest beta.

PR #2086, opened 2023-08-13
Editor: LipSync pane with scalable table layout

ericoporto:
This PR adjusts the LipSync pane to use a table layout instead of hardcoded control positions. This makes it work in both 100% and 200% scaling. 

In Windows 10, I need to use the following compatibility setting in the application properties so it's scaling can be tested.

I opened #2087 to list all the various pane that have issues.

### DIFF
--- a/Editor/AGS.Editor/Panes/LipSyncEditor.Designer.cs
+++ b/Editor/AGS.Editor/Panes/LipSyncEditor.Designer.cs
@@ -30,40 +30,128 @@ namespace AGS.Editor
         {
             this.label1 = new System.Windows.Forms.Label();
             this.label2 = new System.Windows.Forms.Label();
+            this.tableLayoutPanelLP = new System.Windows.Forms.TableLayoutPanel();
+            this.splitContainerMain = new System.Windows.Forms.SplitContainer();
+            this.flowLayoutPanelTop = new System.Windows.Forms.FlowLayoutPanel();
+            this.flowLayoutPanel1 = new System.Windows.Forms.FlowLayoutPanel();
+            ((System.ComponentModel.ISupportInitialize)(this.splitContainerMain)).BeginInit();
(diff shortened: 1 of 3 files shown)


ericoporto:
I guess this is one of the changes that doesn't show up in my Windows 10 theme, can you screenshot and point what you mean?

ivan-mogilko:
I did not test this on Win10 yet, I tried how this looks like on Win7 first.

Here's the original Editor's look:
editor-lipsync-original


Here's how it looks with the flow table:
editor-lipsync-flowtable

ivan-mogilko:
So, I never worked much with this control, but I found this tutorial:
https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-anchor-and-dock-child-controls-in-a-tablelayoutpanel-control?view=netframeworkdesktop-4.8

And tried to adjust your code where you create labels in this table like this:
```
label.TextAlign = ContentAlignment.MiddleRight;
```

This centered the label's text, which made it look okay-ish, imho. Perhaps using MiddleRight is even better, then it looks almost like before. But of course we also need to check that this remains good after the scaling that you're trying to fix.

Another, smaller thing, the upper panel may benefit from extra Padding, imho, because right now the text appears too close to the borders.

ericoporto:
> now the text appears too close to the borders.

That line is a selection mark, if something else is in focus it should disappear. Actually the splitter is fixed, it should not be selectable, don't know why that is happening. :/

I will adjust this later with the alignment you mentioned and see how it looks with scaling!

ivan-mogilko:
> That line is a selection mark, if something else is in focus it should disappear. 

Well, the text is still close to the panel borders, I feel like setting Padding 5-10 gives it a clearer look.

ericoporto:
image

image

It appears it's working!

Issue #2076, opened 2023-08-06
Reduce CirrusCI usage for new policy, possibly offloading some tasks to GitHub Actions


ericoporto:
as of 1st of Setember, Cirrus CI will be limiting FOSS repositories in 50 compute credits per month, according to [their blogpost](https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/). Unfortunately, we use a tad more per month, around 140 compute credits.

I think we can move things that aren't use in a release to GitHub Actions - so macOS builds and the Linux CMake builds. This should lower the compute credit amount of Cirrus CI a bit so we can have some breathing time. Right now, I think the harder to migrate are the Windows stuff as we use VS with Platform Toolset v140, I don't know if this is available in the VS in GitHub Actions.

My fear with GitHub Actions is it seems its a bit harder to keep using older things there, and also overall Docker usage in GitHub Actions is a bit more work - you need to push Docker images to GitHub Package Registry and deal with it more manually.

## Proposed plan

- [x] Reduce use of macOS and non-releaseable tasks (#2079 and  #2083 )
- [x] Leverage [`changesInclude` and `changesIncludeOnly` functions](https://cirrus-ci.org/guide/writing-tasks/#supported-functions) to avoid builds when changes don't ask for - say the commit only updated a README for example. (#2083 )
- [x] For macOS CMake builds we can also look into leveraging ccache (it requires a C++17 capable compiler for building, so we can't use it in other places right now...)
- [x] Look into offloading to GitHub Actions
  -  #2077 has a version of this, but we can explore alternatives here too
  - made a new branch with a possible suggestion here https://github.com/ericoporto/ags/tree/ci-add-gh-actions
  - added in #2091 

## Additional context

There are probably other projects with similar problem, we could take a look into them

- https://github.com/scikit-learn/scikit-learn/issues/26879
- https://github.com/bitcoin/bitcoin/issues/28098
- https://github.com/numpy/numpy/issues/24280
- https://github.com/scipy/scipy/issues/19006
- https://github.com/MDAnalysis/mdanalysis/issues/4216
- https://github.com/opencontainers/runc/issues/3964

## Our previous monthly usage

<img src="https://github.com/adventuregamestudio/ags/assets/2244442/12301650-3f2d-4c4c-9124-94dbced2e8e9)" width=600>

ivan-mogilko:
[...]

PR #2091, opened 2023-08-14
Ci add gh actions

ericoporto:
this mostly disables CI for macOS and linux CMake in Cirrus CI, and instead expect those to happen in GitHub Actions - contributing to the goal of #2076

I think I would like to have GitHub Actions checking on a per commit basis and use Cirrus CI like before a PR merge (maybe triggering manually?), for releases, and maybe through a cron job that checks difference and builds nightly for master and ags4. 

I won't do this in this PR though. I am leaving this as draft for a moment to think about this a bit more.


ivan-mogilko:
What are these skipped tasks for, are they used in preparing a Editor package (like a Linux build target), or no?
I'm mostly interested, whether I still will be able to make a release in the meantime.

PR #2098, opened 2023-08-16
PNG compression

eclecticanimal:
Hello, I tried to do the PNG compression implementation.
I was not able to test if it works, I just get there reading documentation, but not really sure if it would work.


ivan-mogilko:
I'm sorry, but this PR is not good as it is. There are multiple issues with this.

This is honestly the major one:
> I was not able to test if it works, I just get there reading documentation, but not really sure if it would work.

Compilation issues:

0. First of all, we must clarify that it **requires adding zlib library as a dependency**. Our engine nor editor do not link it, but without it your code won't work.
1. New files png.h and png.cpp are not actually added to the project's compilation. They have to be added to MSVS project Common to be compiled on Windows, and they also have to be added to Common/CMakeLists.txt to be compiled along the CMake builds on other platforms.
2. ZLib library will have to be added to dependencies: in MSVS project, CMake script (I will need to double check where), and Engine/Makefile-defs.linux (maybe Makefile-defs.osx too, I will need to double check).

Code issues:
1. png.cpp contains lots of useless code that you copied from lzw.cpp for some reason. I'm speaking about `insert` and `delete` functions, as well as global variables and macros. These are used in lzw algorithm, but not png algorithm, so they should not be there.

Logical issues:
1. The new code is not put into use, so it will never get called (this is why you was not able to test it). To be a complete feature it must be added as an option for saving sprites and/or room backgrounds.
2. PNG compression option has to be added to the [enum SpriteCompression](https://github.com/adventuregamestudio/ags/blob/6244e20d0d71746a91c08ade8a5dfb4741ad6dc0/Common/ac/spritefile.h#L71), and this option should be checked in [SpriteFileWriter::WriteBitmap](https://github.com/adventuregamestudio/ags/blob/6244e20d0d71746a91c08ade8a5dfb4741ad6dc0/Common/ac/spritefile.cpp#L696) and [SpriteFile::LoadSprite](https://github.com/adventuregamestudio/ags/blob/6244e20d0d71746a91c08ade8a5dfb4741ad6dc0/Common/ac/spritefile.cpp#L442) functions, with pngcompress and pngdecompress called respectively. 
4. PNG compression option has to be added in the Editor too, so that users could select it for their game. It's done by adding it to the [SpriteCompression enum](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Types/Enums/SpriteCompression.cs).

After this is done, then will then be possible to actually test it on real game projects, see if it adds benefit for compression sizes, and so on.

eclecticanimal:
Thank you. I'll follow the guidelines

ivan-mogilko:
Okay, I was not clear about the zlib library.

We do not push whole library sources into the repository, unless there's no other choice. This is redundant, and also this prevents platforms like Linux use the libs installed on their systems.

Instead we add corresponding library to the list of linked libs. I will probably have to explain how.

ivan-mogilko:
I tried to explain how to add a lib dependency in MSVS here:
https://github.com/adventuregamestudio/ags/wiki/Adding-third‐party-libraries-to-the-project

I think it's best to begin with Windows, since it's required to test this feature with the Editor.

If you have questions, please ask!

eclecticanimal:
Hello I did some more changes following your guidelines, but not clear about the dependencies.

ivan-mogilko:
I will try making a properly linked test in a separate PR, based on your code, and test if that works in principle.

ivan-mogilko:
Made #2111 based on your suggestion. I will close this one.

PR #2102, opened 2023-08-20
fix the no static arrays in autocomplete issue

### LINKED ISSUE #2101: Autocomplete doesn't work for static attribute arrays
**Describe the bug**
When you get the autocomplete for `Screen` object in the editor, the `Viewports` array doesn't show up.

**AGS Version**
v.3.6.1.6 or current master (c4947b21842b525692fac80033198cc4a75b2573).

**To Reproduce**
Try to get any struct that contains a member that is a static array, as an example, the Screen object, type `Screen` in the script editor and once you press the dot, the autocomplete shows up but the `Viewports` member is not there.

**Expected behavior**
There should be an additional item on the autocomplete for `Viewports` 

**Screenshots**
image

**Desktop:**
 - OS: Windows
 - Version: 10

ericoporto:
fix #2101

image

I couldn't find any unintended side effect on this change in the autocomplete!

### DIFF
--- a/Editor/AGS.Editor/Entities/AutoCompleteParserState.cs
+++ b/Editor/AGS.Editor/Entities/AutoCompleteParserState.cs
@@ -68,7 +68,7 @@ namespace AGS.Editor
             get { return PreviousWords[2]; }
         }
 
-        public string[] PreviousWords = new string[6];
+        public string[] PreviousWords = new string[7];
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Could you please explain how that fixes the bug, add some comment why is there 7 words?

PR #2103, opened 2023-08-22
Script API: implement Overlay.GetAtScreenXY, GetAtRoomXY, Visible

ivan-mogilko:
After numerous user requests...

This adds:
* `Overlay.Visible property;`
* `static Overlay.GetAtScreenXY(x, y, bool room_overlays);`
* `static Overlay.GetAtRoomXY(x, y);`

`Overlay.Visible` property is an explicit way to turn overlays on and off rather than setting transparency to max. This makes overlays close to all the rest of the drawable objects in game in set of properties they use, and their logic. But the main reason to add it here is that GetAtScreen/Room functions historically ignore the transparency and other visual effects, and only check for Enabled/Visible/Clickable properties (where applicable).

For GetAtScreen/Room functions I added a z-sorted vector of overlay indices, which is updated when:
* overlay is added,
* overlay is removed,
* overlay.ZOrder property changes
* save is restored

This vector is used to hit-test overlays in their z-order, as opposed to test them in the order of their storage in memory until the best match is found. This should speed GetAt* functions up to some degree in case of many overlapping overlays. However, I'm not certain if and how updating this vector will affect other actions, like adding/removing overlays.

Still these functions may be slow in case there are many thousands of overlays around. I believe there are potential ways to improve this, such as doing a space partitioning, but I left this for the later. Right now they simply iterate all the existing overlays in the their sorted z-order and test against their bounding rects.
Therefore I don't think they should be used for something complex, such as collision checks, and similar. IMO it's best to limit their use for reacting to user input (like mouse clicks).

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1343,6 +1343,14 @@ builtin managed struct Overlay {
   /// Gets/sets the overlay's z-order relative to other overlays and on-screen objects.
   import attribute int ZOrder;
 #endif
+#ifdef SCRIPT_API_v361
+  /// Gets the overlay that is at the specified position on the screen.
+  import static Overlay* GetAtScreenXY(int x, int y, bool roomOverlays = true);  // $AUTOCOMPLETESTATICONLY$
+  /// Gets the overlay that is at the specified position within this room.
+  import static Overlay* GetAtRoomXY(int x, int y);      // $AUTOCOMPLETESTATICONLY$
+  /// Gets/sets whether this Overlay is currently visible.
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Hm, I had another idea about how to manage the sorting vector. Since in games that use overlays to construct large scenes zorder property is changed often, but GetAtScreen probably will be called sparingly, then it may be faster to not update "order" vector each time a property changes, but instead mark it as "dirty" and update first time GetAtScreen is called (or any other function that reads it). I might try that and compare the performance diff.

ericoporto:
Hey, I won't have time to test this soon, can you wait a bit before merging this? I want to test the performance impact because I am using a few thousand overlays. :/

PR #2114, opened 2023-08-27
Add Deflate algorithm to Sprite Compression using MiniZ

ericoporto:
a different version of #2111 

just for reference of how it looks like with miniz sources directly in our repo. The source files are reasonably small, so it may be ok. Miniz has interface compatible with zlib, so it could be useful in the future for libpng or something similar to libpng - which would enable png glyphs in freetype, and possibly also png as a format for dynamic sprite file saving/loading.

I am also working in a version that instead fetches miniz in cmake or has miniz in dockerfiles (Windows, Linux) - and then I would add also zlib compatibility in the build scripts, so either of these two would be alright, since we are using the zlib interface already.

There is also an alternative to pick only the specific deflate code from miniz and skip the zlib interface, so that the source files are even smaller - the disadvantage is if we want to add PNG capabilities in the future we would not be able to leverage libpng and compatible libraries that uses zlib interface.


ivan-mogilko:
> @ivan-mogilko , I think you mentioned something like using a new stream like zipstream that would leverage deflate/inflate. I imagine it would have a specific direction - say it reads doing inflate (decompression) and writes doing deflate (compression). And probably some fixed chunk size (4kB ?) - maybe there's some way to adjust it.

My idea was quite raw, and i did not thought it through. The first thought was that it could be a kind of a proxy stream with a inflate/deflate "filter" inside. The works could be similar to the buffered stream.
Decompressing stream:
- read compressed data from input stream;
- pass through the deflate;
- store in a internal buffer;
- return bytes from the internal buffer on request.
- when internal buffer is read out, go back to reading compressed data.
Compressing stream does similar in opposite direction.

ivan-mogilko:
I was not planning the new stream class specifically for the sprite compression, but as a thought of something that might be convenient to make uniform reading code if more kinds of data will have to be decompressed, and larger data too which has to be read rather from file directly.

In regards to compression types, there are 3 compression formats supported already: RLE, LZW and now will be Deflate.

ivan-mogilko:
Tested this on the game which I previously used for the test, because I wanted to double check that the format does not depend on library.

The compressed size appear to be slightly bigger, about 1% larger (280 MB vs 278 MB on the project I used).

ericoporto:
That's interesting, for some reason I never thought about the size being different, I thought that the libraries just were different mostly in performance (as in CPU and memory usage).

If you want I can try to redo the zlib-ng version using pure zlib. I think though if we use an external library I would prefer to hold a fork in the ags organization so I can apply platform/compiler specific fixes more easily - zlib has like 200 open PRs so it's not very friendly to work with it. :/

ivan-mogilko:
The difference is within 1%, I don't think this matters.

ericoporto:
I don't have idea regarding performance impact, I guess my game projects aren't asset intensive in a way it matters, did you pickup any difference at all?

PR #2116, opened 2023-08-27
Add Deflate algorithm to Sprite Compression using ZLIB-NG

ericoporto:
Alternative to #2114, without sources in the repository (fewer line changes) and using zlib-ng instead.

I prefer sources in the repository for small things, zlib-ng is quite huge though, so that's why it's not added - miniz though is small enough it makes more sense to add it.

It's expected that zlib-ng is slightly faster than miniz but the tradeoff is a bigger footprint. I haven't validated this affirmation!!!


ivan-mogilko:
Sorry, but why have multiple PRs? Was not the previous one working?

ivan-mogilko:
I'm not sure what to do with this, am I supposed to review and test both #2114 and this? Is there a reason why not use a ready 2114 implementation?

ericoporto:
Sorry, I wasn't sure if you preferred sources in repository or external library, this was more to have alternatives.

I think I prefer the other one with sources in the repository because always there is something wrong in some platform that I need to fix - apparently zlib-ng already has a wrong ifdef for x86 MSVC when the build is static.

ivan-mogilko:
I do not know what I prefer, I was going to investigate, but now since you took over this, then let's use the one that you think is more convenient.

ericoporto:
I think I prefer the miniz route then but this is here as a reference if minds are changed.

Issue #2058, opened 2023-07-16
Distinct separation of data and runtime classes, clear Common (shared lib)


ivan-mogilko:
This is related to #1833 and #1834, but I decided to split this out into the separate ticket, as this task requires to work out a good single approach for several game classes.

**Problem**

The Common library has a number of classes that are mixing 3 purposes:
* game data serialization,
* design-time preview (like drawing in the editor),
* runtime behavior, including registering class instances in the script vm.
These classes are:
* CharacterInfo (partial character class),
* All GUI classes.

This is bad because:
* it's not always clear how to expand these classes;
* class is mixing purposes, sometimes unrelated ones;
* adding runtime behavior to a class requires either hacks or to provide different method implementations in each program that uses it; for example: a real function in the engine and stubs in the editor or tools;
* adds redundant dependencies when one wants to reuse the class in a different program.

In case of CharacterInfo the situation is also somewhat more complicated because these instances are stored inside another class, GameSetupStruct, which is also used for both data serialization (at design and runtime) and referenced at runtime. They could be moved out from that class, but I believe it's best to have them in some Game class anyway, because eventually we should not store game objects in multiple global variables and arrays.

**Requirements**

Before doing something it's detrimental to define what is required. I think that is:
1. We need structs that define serialized data, with functions that load and save these structs. This is useful for Editor (at least until it's rewritten to not rely on C++ code anymore) and any tool which reuses common engine code. I guess this may be generally useful for engine too. Let's call these "data structs" for convenience.
2. The instances of these structs have to be grouped in another "data struct" for convenience, for example "data structs" Character, Inventory etc may be stored inside a Game "data struct".
3. We need runtime classes that contain both data loaded from the game files, and any additional runtime data, as well as methods with runtime behavior. Their relation to "data" classes is a question of realization, they may inherit them, or contain inside, or just have a copy of their fields.
4. If shared lib still needs a "preview" behavior, like drawing a GUI, then it may include some kind of a partial runtime class, with only necessary behavior. While the full runtime class in the engine would inherit this base class.

NOTE: This separation on Common lib and Engine originally was meant to make it easier to separate functionality between Engine and the Editor. I guess that eventually we should not have this separation on Common lib and the Engine, but instead other programs would cherry pick and include files from the Engine. But to get there, Engine must have a very clear structure without mixed and redundant dependencies between its components. So that lies far outside the scope of this task.
[...]

PR #2117, opened 2023-08-29
Change the rules of how game objects are updated after sprite's update/deletion

### LINKED ISSUE #2104: Resolve a problem of updating game after dynamic sprite change/deletion
This is written after #1880, and doing small experiment as noted [in these couple of comments](https://github.com/adventuregamestudio/ags/issues/1880#issuecomment-1642899768).

The problem is following. The DynamicSprites are not assigned to game objects using a pointer, but instead using their IDs, just like the regular sprites.

One consequence of that is that when the dynamic sprite is deleted, the objects still keep their ID in their properties. Also as soon as ID becomes unused, it may later be assigned to another created dynamic sprite. Thus, if object's sprite property remains (e.g. if user forgot to reset/update it in script), then it may point at non-existing sprite, or a completely unrelated sprite.

Historically AGS did not reset Graphic property for all of the game objects, it only reset it for RoomObjects and Buttons (afaik). I thought to make this consistent and also added everything else that may reference a sprite to this list (I think this was done around 3.5.1, or maybe 3.6.0), but this of course slows things down considerably when alot of dynamic sprites are created and deleted consequently during a single game frame, because the current algorithm is very dumb and simply checks every possible sprite property in every single object in game's memory. In an ordinary adventure game the primary culprit will be Views, because this algorithm must check every single view frame for its graphic. If a game creates alot of Overlays, then they may become the main slowdown factor for this case.

So, indeed, this was a rather stupid thing to do.

The silly thing here is that above does not really cover all the situations where Graphic property has a wrong value, as currently nothing prevents user from assigning a non-existent sprite ID to a Graphic property (and I believe that will cause engine to crash badly).

There are already real games where this caused a slowdown, this was reported by a user to me. Mostly this was because of a suboptimal script that created too many dynamic sprites, but this means a engine regression nevertheless.
Hence I'd like to fix this situation in 3.6.1.

I have some ideas on how this could be improved, but first of all we must decide on what is required at all.
Trying to summarize, there are following related questions here:

1. Do we have a *logical* necessity to reset Graphic property when a DynamicSprite is deleted? By "logical" I mean - if we forget of technical issues, such as potentially trying to get non-existant sprite using saved ID and getting null pointer instead, - as these may have a solution of their own.
I know one actual problem with not resetting the property: because sprite's ID is freed by a deleted DynamicSprite, next DynamicSprite may get the same ID. If the Graphic property remains, then it will implicitly refer to a different sprite.
(Ironically this behavior was utilized in some of the old games, possibly unintended, where the coder recreated dynamic sprite without resetting Graphic, and it still pointed to a new valid sprite.)
Strictly speaking, it's a question of responsibility. Do we want users to be fully responsible for updating object properties after deleing a DynamicSprite, and get all kinds of weird consequences if they forget to, or we want the engine to do the automatic fixup for them?

2. In relation to the above, do we have a logical obligation to prevent setting a wrong value to Graphic property by hand in user script?
That is - test if this ID really refers to a sprite, and if not, then do something, like keep old value, reset to 0, issue a warning to the log, etc.

3. Is it possible to safeguard the program from crashing in case the Graphic property refering a non-existant sprite?
I believe that it is, without adding any more overhead. Currently SpriteCache already does checks for the sprite's existance, because it must know when to reload a missing one. I think SpriteCache could simply return a sprite 0, or even some programmatically generated placeholder, in case there's no real sprite, thus preventing engine from receiving a nullpointer, and also making it unnecessary to litter the code with nullpointer checks everywhere where engine wants to check a sprite.

4. What are the ways to speed up the algorithm of on-screen object updates when the DynamicSprite is updated (transformed, or drawn upon) or deleted? This still has to be done, in order to update the object texture and/or mark object for redraw.
I only thought briefly about this, and overall there are two approaches: either map an object to its sprite, or map a sprite to objects that use it. The mapping does not have to be direct. For example, it may be possible to mark the sprite as "assigned to some object" or "some viewframe" as a simple flag, which would narrow down the amount of search necessary to do after sprite update/deletion. This will also let completely avoid doing a search for temporary dynamic sprites that are only used to draw to other surfaces, but not assigned anywhere.

ivan-mogilko:
Fixes #2104 . Please read the ticket and comments for detailed explanation of the problem.

There are following changes done here.

1. SpriteCache now has a "placeholder" field initialized to a 1x1 transparent bitmap by default. operator `[]` is ensured to always return this placeholder if an invalid sprite was requested. This is a safeguard against program exceptions. Note that it's still possible to test for sprite's existence explicitly by calling DoesSpriteExist.
I think it may be feasible to add a method for setting custom placeholder. Possibly a copy of sprite 0 may be assigned as a placeholder. I may add this behavior if someone thinks that's a good idea. At first I thought to just use sprite 0 implicitly, but it's a real possibility that sprite 0 does not exist in game. So I went with a dedicated member instead. After all it's a exception safeguard, so should work regardless of the game data.

2. Deleting dynamic sprite no longer resets Graphic property of objects which had it set, at all. The property values stays as they are, which may result in them referring to non-existing sprite, or another dynamic sprite, would the same ID be assigned to them further.

3. In regards to updating objects on screen after sprite's change or deletion, I implemented a new method that does that. The details may be found in commit called "rewrote dynamic sprite change notification method". In short: for texture-based render updating shared texture seem to be enough. For software mode there's a vector<bool> that tells when the sprites were modified. The elements are marked when the sprite is updated or deleted. The game objects test these flags along with other conditions that tell whether object has been modified and should be redrawn. This is done for all objects except for GUI (see below).

**NOTE:** I found a potential problem with this new method, which is explained in [this comment](https://github.com/adventuregamestudio/ags/issues/2104#issuecomment-1698071807). In short, some object properties or actions may depend on sprite's size, and if they *cache* sprite's size, or any other value derived from the sprite's size, then the objects also have to be updated as soon as sprite is resized. This may currently occur upon DynamicSprite.Resize, Rotate and ChangeCanvasSize.
Examining current objects, it *seems* like this is not an *immediate* problem, as the sprite size is never cached in them (it's only cached in draw caches, but that is still updated properly in this pr). In case this becomes a problem in the future, there remains a solution: track sprite use upon a Graphic property assignment too.

**Potential TODO:**
1. Support setting placeholder to sprite 0, if it exists in game data?
2. GUI: this is the only major problem left. Because they are still drawn differently, sometimes combining a sprite with their own drawing (and some have >1 sprite, like sliders that have base graphic and a handle graphic), they would need a special approach, which I did not invent yet. For now I left them updated as before: whenever a dynamic sprite is changed or deleted it will run through all guis and controls that may have sprites, and marks those that have this sprite assigned. I've got a feeling that GUIs may also use this "sprites modified" vector somehow, but I left this for the future.

For now I'm interested to know if other contributors could see actual flaws in this approach.

Also, this has to be tested for any performance changes. I guess there should be 2 kinds of tests:
1. Regular test with many drawn objects, just to see that drawing itself did not slow down.
2. Test that creating and deleting a lot of dynamic sprites is not slowing things down more than in earlier versions (prior to 3.5.0 probably).

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -19,12 +19,14 @@
 
 using namespace AGS::Common;
 
+// Internal SpriteCache's flags, not serialized in game data
+//
 // Tells that the sprite is found in the game resources.
 #define SPRCACHEFLAG_ISASSET        0x01
 // Tells that the sprite is assigned externally and cannot be autodisposed.
 #define SPRCACHEFLAG_EXTERNAL       0x02
-// Tells that the sprite index was remapped to the placeholder (sprite 0).
-#define SPRCACHEFLAG_REMAP0         0x04
+// Tells that the asset sprite failed to load
+#define SPRCACHEFLAG_ERROR          0x04
 // Locked sprites are ones that should not be freed when out of cache space.
 #define SPRCACHEFLAG_LOCKED         0x08
 
@@ -49,11 +51,9 @@ SpriteCache::SpriteCache(std::vector<SpriteInfo> &sprInfos, const Callbacks &cal
     _callbacks.InitSprite = (callbacks.InitSprite) ? callbacks.InitSprite : DummyInitSprite;
     _callbacks.PostInitSprite = (callbacks.PostInitSprite) ? callbacks.PostInitSprite : DummyPostInitSprite;
     _callbacks.PrewriteSprite = (callbacks.PrewriteSprite) ? callbacks.PrewriteSprite : DummyPrewriteSprite;
-}
 
-SpriteCache::~SpriteCache()
-{
-    Reset();
+    // Generate a placeholder sprite: 1x1 transparent bitmap
+    _placeholder.reset(BitmapHelper::CreateTransparentBitmap(1, 1));
 }
 
 size_t SpriteCache::GetSpriteSlotCount() const
@@ -115,7 +115,7 @@ void SpriteCache::SetEmptySprite(sprkey_t index, bool as_asset)
     ResourceCache::Dispose(index); // make sure it's free
     if (as_asset)
         _spriteData[index].Flags = SPRCACHEFLAG_ISASSET;
-    RemapSpriteToSprite0(index);
+    RemapSpriteToPlaceholder(index);
 }
 
 Bitmap *SpriteCache::RemoveSprite(sprkey_t index)
@@ -175,17 +175,17 @@ sprkey_t SpriteCache::GetFreeIndex()
 
 bool SpriteCache::SpriteData::IsAssetSprite() const
 {
-    return (Flags & SPRCACHEFLAG_ISASSET) != 0; // found in game resources
+    return (Flags & SPRCACHEFLAG_ISASSET) != 0;
 }
 
-bool SpriteCache::SpriteData::IsRemapped() const
+bool SpriteCache::SpriteData::IsError() const
 {
-    return (Flags & SPRCACHEFLAG_REMAP0) != 0; // was remapped to placeholder (sprite 0)
+    return (Flags & SPRCACHEFLAG_ERROR) != 0;
 }
 
 bool SpriteCache::SpriteData::IsExternalSprite() const
 {
-    return (Flags & SPRCACHEFLAG_EXTERNAL) != 0; // assigned externally
+    return (Flags & SPRCACHEFLAG_EXTERNAL) != 0;
 }
 
 bool SpriteCache::SpriteData::IsLocked() const
@@ -195,7 +195,7 @@ bool SpriteCache::SpriteData::IsLocked() const
 
 bool SpriteCache::DoesSpriteExist(sprkey_t index) const
 {
-    return index >= 0 && (size_t)index < _spriteData.size() && // in the valid range
+    return (index >= 0 && (size_t)index < _spriteData.size()) && // in the valid range
         _spriteData[index].IsValid(); // has assigned sprite
 }
 
@@ -208,19 +208,21 @@ Bitmap *SpriteCache::operator [] (sprkey_t index)
 {
     // invalid sprite slot
     assert(index >= 0); // out of positive range indexes are valid to fail
-    if (index < 0 || (size_t)index >= _spriteData.size())
-        return nullptr;
+    if (!DoesSpriteExist(index) || _spriteData[index].IsError())
+        return _placeholder.get();
 
-    // Resolve potentially remapped sprites
-    index = GetDataIndex(index);
     // Try get image from cache
     auto &image = ResourceCache::Get(index);
     if (image)
         return image.get();
     // If no ready image, but has an asset, then try loading one
     if (_spriteData[index].IsAssetSprite())
-        return LoadSprite(index);
-    return nullptr;
+    {
+        auto *bitmap = LoadSprite(index);
+        if (bitmap)
+            return bitmap;
+    }
+    return _placeholder.get();
 }
 
 void SpriteCache::DisposeAllCached()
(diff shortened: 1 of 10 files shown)


ivan-mogilko:
@messengerbag, in regards to crashing the engine in case of user error, curiously, I missed this earlier, but found when testing these changes. There are already 2 checks for invalid sprites when drawing room objects and characters only. So, this pr brings back this old behavior probably, where deleting a sprite and forgetting to reassign a new one will stop the game with error message, like:
> There was an error drawing object N. Its current sprite, X, is invalid.

ericoporto:
Hey this PR instantly crashes my Future Flashback with the error below - but I checked and using the acwin.exe from master works fine.
image

ivan-mogilko:
This breaks some things, need to do more testing.

ivan-mogilko:
Hmm okay, I did some stupid oversights here, and gui is not updated along with dynamic sprites.

Also, I see now there's probably ways to simplify this further. So I have to redo half of this pr.

ivan-mogilko:
I did some thinking and code research again, and realized that for the texture-based rendering most of the object checking and marking for update is simply not needed anymore. After introducing shared textures (even before the texture cache, but more so after) the objects will already get updated as soon as you update the texture data itself, or mark it deleted. This refers to characters, room objects and overlays. GUI is different though (see below).

What was left is software drawing mode. My thinking was: so when you have a shared texture, you may mark texture updated, and all objects that use it will receive this update. So why not do something similar in software mode too?
What I tried is a shared *notification block*, which is a simple struct with just a `Modified` bool in it. There's a map which stores these "control blocks", and whenever an object draws with a dynamic sprite for the first time, it gets assigned this block from the map.
When a dynamic sprite is deleted or changed, the Modified flag is set in a control block, and so every object that still uses it will read this flag when checking whether a redraw is necessary.

GUI: this is the only problem left. Because they are still drawn differently, sometimes combining a sprite with their own drawing, they would need a special approach, which I did not invent yet. For now I left them updated as before: whenever a dynamic sprite is changed or deleted it will run through all guis and controls that may have sprites, and marks those that have this sprite assigned.

I've got a feeling that GUIs may also use this "notification block" somehow, I will think about it for bit more.

ivan-mogilko:
Note: this does not need #2123, but merging that first will simplify some code here.

ivan-mogilko:
I did performance tests, and judging by them, the primary culprit in slowing things down was checking properties and resetting them on sprite deletion. I presume that Views did the most impact. Removing these checks and property reassignments alone already seem to give good results.
And of course, not having to test through all Overlays, which would be very bad in case there are thousands of those.

In comparison to that, disabling searching through all GUIs and marking those containing deleted dynamic sprite for redraw makes an almost unnoticeable effect (maybe couple of % fps gain).

**For the reference:** I've been testing a real game scene where around **200+** temporary dynamic sprites were created and deleted per game frame, which contained **90** GUIs and **380** buttons, each of them was tested.

So while it would be good to improve this further, and only notify the guis that matter, I think it would be fine even like it is now.  If I won't be able to figure out a nice way to handle the gui notification.

ivan-mogilko:
So, in regards to software mode, I redid back to the variant which uses vector of bools to mark "recently modified sprites". This might be faster to check regularly at a cost of extra memory use. From what I heard before, the std::vector<bool> is a specialized container which stores elements in bits, meaning its 8 sprites per byte, which in worst case of game with many tens of thousands sprites this will result in kilobytes of extra data. Note that a previously tested "notification block" solution would require at least a size of a pointer (4 to 8 bytes) per a drawn object + something like 2 ints (8 bytes more) per a  dynamic sprite drawn at least once.
I could not decide which is best or worst.
Hypothetically, the sprite flags vector may be accessed from GUI drawing code, in order to check whether it needs update too. But I did not go there still. I think I prefer to leave this as is for the time being, and come back to this after a planned refactor is done, and there are clear and distinct "runtime" classes for GUIs (see #2058), then it will be perhaps easier to see which is a more convenient way to "connect" GUI elements with dynamic sprites to receive notifications about latter being modified.
For now, I'm afraid, I'm spending too much time thinking this over and over again...

ericoporto:
The recent version of this doesn't error in Future Flashback anymore. I tried some other games and they all seem to work well. I tried to compare performance but can't tell any regression - it looks the same? At least on the test games I had. The only one I didn't test was the sandwalker because it's using ags4, but I assume it's not significantly impacted because it doesn't do sprite deletion at any point - afaict.

ivan-mogilko:
The truth is, this becomes noticeable (at least on systems with a good CPU) only when the game creates and deletes a lot of dynamic sprites at once. Like in AGD's "Mage's Initiation" there's a particle system for magic spells which may repeatedly create and delete 200+ dynamic sprites each frame (it's not done very efficiently).
In that case I got a jump from like 30 fps to 110 fps with this PR. But I guess that's a very rare situation.

PR #2118, opened 2023-08-30
Android project sdk upgrade and ci adjustments

ericoporto:
- upgrade SDK build tools to 34.0.0
- upgrade Android Gradle Plugin to 8.1.1 client, and upgrade Gradle to 8.3
- better caching for android builds and configs, and in CI too (fine cache only pushes in master branch)
- additional gradle properties tuning


ericoporto:
This is still not building, so I think after Cirrus updates their docker (https://github.com/cirruslabs/docker-images-android/issues/58) I will update our android project to use it and see if it fixes.

For some reason its trying to install the ndk we already have in the docker image (??) and it looks like it locks waiting to accept the license - or something else at the same time.

ericoporto:
Gradle is still timing out in the CI and I don't know why :( things are building fine locally... I will try switching the base docker image to see if it changes anything.

ericoporto:
@ivan-mogilko I force pushed #2064 as an example, but Cirrus CI people have figured the issue and fixed on their side! So the android build is working again!

I will rework these commits as I thin the summary of this - upgrade docker image, upgrade version of things and all that - is still valid and useful, but I need to properly organize this. But this means releases should be working again.

PR #2123, opened 2023-09-03
Store custom overlay's image as dynamic sprite

ivan-mogilko:
This is something I experimented with few months ago, but work on #2117 reminded me of this code.

The main purpose of this is to make overlays work in a more uniform way, making things (and code) somewhat simpler. Historically overlays always stored some image in their struct. Since 3.6.0 they may optionally store a shared sprite number instead. This makes them work in double mode, where they may or not store a sprite internally. This sprite is either generated for textual overlays, or cloned from existing sprite if a "clone" argument is passed to overlay's factory method. But regardless this means that there's some kind of a special image which is non-accessible, not counted in sprite cache statistics, and written in its own way in game saves.

What the pr does: it makes the overlay store this generated image in the sprite cache as a "dynamic sprite". There's just one difference from common dynamic sprites: this one does not have a script DynamicSprite object referencing it. The ownership is fully kept by overlay itself.

As a result, the sprite becomes a part of standard sprite container, counts toward sprite statistics, saved among the rest of dynamic sprites in saves. In fact, it is even possible to reference it using Overlay.Graphic, so you may learn its parameters using Game.SpriteWidth/Height and draw it somewhere. You cannot draw *upon* this sprite, as DynamicSprite struct is not exposed as a property for this sprite, but I guess this may be added later if really wanted, and someone sees a good purpose.

Second commit is not directly related to this sprite, but it further simplifies Overlay struct by removing a texture pointer (DDB) from it, and makes overlays use ObjTexture struct, similar to Characters and Room Objects. The purpose is to make all these graphical objects closer in how their are processed.


ericoporto:
This is a pretty good change! Great work!

ivan-mogilko:
Unfortunately, there's a number of errors occuring after restoring an older save. I fixed most of them, but still missing something that is apparently causing memory corruption in a managed pool. So this cannot be merged right now.

ivan-mogilko:
Fixed one more bug. Tested by loading older saves with multiple textual overlays created, and creating them in the new version, saving and loading. Things seem to work now, as far as i can tell...

PR #2127, opened 2023-09-07
Editor: support multiframe and multiloop selection

### LINKED ISSUE #2124: Editor: in a View editor, support multiple frame selection
This is a repeating user request from old times until today: let select multiple frames in a View editor, which would allow to set property values for all of them at once. Usually people want to set same delay or same Flipped value for a loop, or multiple loops at once.

Properties grid already supports editing multiple objects at once, you may see how this is done for Sprites.

So the main problem here, it seems, to add multiple frame selection.

My proposal is to add classic controls, where:
* Ctrl+Click would select another individual frame,
* Shift+Click would select all frames between last selected one and clicked. I suppose that this should select even multiple loops: e.g. if user selected a frame 0 in loop 0, then Shift+Clicks on a frame 10 in loop 10, then all the frames in all the loops in the middle will also get selected.

**UPD**
About the code, the View preview pane (ViewEditor class) is constructed of multiple Loop preview panes (ViewLoopEditor class), each controlling its loop. Selection handling is shared between the Loop preview class and View pane class.

EDIT: also, there are methods in ViewLoopEditor working exclusively with "selected frame", instead of accepting a frame as a parameter. This all has to be refactored.

ivan-mogilko:
Resolves #2124.

This adds support for the multi-frame selection to the View editor. This includes *selecting frames in different loops*.
When multiple frames are selected, the Property Grid will display shared values for them, and let set values for all frames at once (this is an existing feature, previously only utilized for Sprites).

Emulates classic selection controls:
* Click - reset selection to the given frame;
* Ctrl + Click - add or remove selection on the given frame;
* Shift + Click - select a range of frames, starting from the last *individually* selected frame and to the clicked one.

Ranged selection is supposed to work across loops too.

In terms of code, I added enum called MultiSelectAction that describes selection type, because I don't know any better. If there's an existing equivalent to this in C#, the I may replace this.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -123,6 +123,7 @@
     <Compile Include="Components\TextParserComponent.cs" />
     <Compile Include="Components\ViewsComponent.cs" />
     <Compile Include="Entities\LogBufferEventArgs.cs" />
+    <Compile Include="Entities\MultiSelectAction.cs" />
     <Compile Include="GUI\AssignToView.cs">
       <SubType>Form</SubType>
     </Compile>
--- a/Editor/AGS.Editor/Entities/MultiSelectAction.cs
+++ b/Editor/AGS.Editor/Entities/MultiSelectAction.cs
@@ -1,0 +1,22 @@
+﻿using System;
+
+namespace AGS.Editor
+{
+    /// <summary>
+    /// Describes an action in a list control supporting multi-selection.
+    /// CHECKME: does .NET have a standard type for this already?
+    /// </summary>
+    public enum MultiSelectAction
+    {
+        // Set a new single selection
+        Set,
+        // Add a single item to the selection list
+        Add,
+        // Remove a single item from the selection list
+        Remove,
+        // Add a range to the selection list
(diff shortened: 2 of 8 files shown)


ivan-mogilko:
Something is wrong with CI not building windows part, so here's a manually built editor for testing:
https://www.dropbox.com/scl/fi/b3h3w3oxdwhfe4cuve5xr/361-multiviewframeselect.zip?rlkey=hckjbxjfozq7fsn7nd9xzm8qg&dl=0

PacketLauncher:
A great improvement to the View editor. Thanks! I can confirm that delaying , flipping, setting different images and setting a sound on frames in multiple loops work great! Below some feedback for other things I checked that don't currently work correctly:

1. Upon trying to delete (Del Key) a selection of frames in multiple loops, only the frames of the first loop are getting deleted while the selected frames on different loops remain intact.

2. The right-click context menu on any of the frames from the multiple selection removes the multiple selection and affects only the single frame clicked upon.

ivan-mogilko:
Updated, fixed selection reset on right click, made separate set of context commands depending on number of selected frames and whether they are selected in multiple loops.

PR #2128, opened 2023-09-09
Test docker rebuild

ericoporto:
@ivan-mogilko The error was a misconfigured DNS apparently. I pushed forced here just to confirm that the Windows builds run without issue.


ericoporto:
https://github.com/cirruslabs/cirrus-ci-docs/issues/1232

Here is the description and tracking of the issue that happened, apparently it’s all fixed now.

There is an occasional issue that appeared, I reported, but it seems it doesn’t happen often.

I will close this to avoid confusion, as there is no  need to recache the build.

PR #2130, opened 2023-09-11
Editor: allow downscale Character preview & make frame preview in View pane fill whole size

### LINKED ISSUE #2020: Possibility to _down_-zoom characters in Editor
**Describe the problem**
When you preview a character in the Editor, there is a zoom field to blow up visuals that are small. However, I've got the opposite problem: My characters are too large to fit the screen; I'd need the possibility to shrink them down to, e.g., 25 %. The zoom field doesn't offer any possibility to enter values smaller than 100 %. Thus there's no way to see all of the character with my monitor.

image

Similar problem in the View editor preview:

image

**Suggested change**
Make the zoom range start at 25 % or even at 10 % instead of 100 %.

AGS 3.6.0

ivan-mogilko:
Resolves #2020, and suggestion in comments by @ericoporto :
> In the view, originally I meant the left pane to have no relationship to the zoom slider at right, and instead just fill the panel as well as possible. I still think this works best for the animation preview, as it's more a quick preview of an animation that will be only properly viewed in-game, later, in a context. I think the slider should only affect the right pane there.

1. Adjust zoom sliders to allow downscale down to min value of 20%, changed step to 10% mostly to let more precise downscale than with 25% step. Made zoom control wider on existing panels, for convenience (idk why MSVS did so many changes to Designer code everywhere, apparently this happens when regenerating the control looks).
2. Added AutoFrameFill mode to the ViewPreview class. When this mode is enabled the view frame will be stretched to fill any available client size, while keeping aspect ratio (it's actually based on sprite height). The ZoomLevel property is ignored in this mode. This mode is used for preview in the ViewEditor.

### DIFF
--- a/Editor/AGS.Editor/Panes/CharacterEditor.Designer.cs
+++ b/Editor/AGS.Editor/Panes/CharacterEditor.Designer.cs
@@ -34,8 +34,8 @@ namespace AGS.Editor
             this.btnMakePlayer = new System.Windows.Forms.Button();
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
             this.sldZoomLevel = new AGS.Editor.ZoomTrackbar();
-            this.viewPreview1 = new AGS.Editor.ViewPreview();
             this.viewPreview2 = new AGS.Editor.ViewPreview();
+            this.viewPreview1 = new AGS.Editor.ViewPreview();
             this.groupBox1.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)(this.splitContainer1)).BeginInit();
             this.splitContainer1.Panel1.SuspendLayout();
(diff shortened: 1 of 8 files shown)


ericoporto:
It works as it says. I noticed that when the view size is lower than some amount of zoom, the preview panel can use white instead of pink for whatever reason, but only when the Editor is maximized. I couldn't figure the explanation, and I think it's fine, just the only thing I found curious.

https://github.com/adventuregamestudio/ags/assets/2244442/ce13b9aa-2837-409e-952b-def1ee17b35c

The character zoom can also downscale, can't figure any problem with that.

ivan-mogilko:
After using this, I am not certain if having preview always stretched was a good idea. This may be fine for hi resolution games, but in low res the character's image is just too pixelated. This may be inconvenient when testing animation, as I do not see it the way it will be in game.

PR #2133, opened 2023-09-16
Editor: add open compiled folder in file explorer

ericoporto:
Adds a command that opens the compiled directory in file explorer

image

image

Usually I just open the Compiled folder to grab the built executables, which is why I added that.

But I added also the File menu to open and show Game.agf in File Explorer.

I don't know why the art direction of all menu commands have a faded grayish color, so I made the icon like this and saved an over variant that has normal color just in case this is added in a toolbar later.

Forum: https://www.adventuregamestudio.co.uk/forums/index.php?topic=61091.0


ivan-mogilko:
Would it be a good idea to also add "Open project folder" to , say, File menu?

morganwillcock:
For the "Compiled" directory, it is a little confusing that the menu item does nothing when the directory isn't present. I think it will be present for projects created from scratch, but probably would not be present if the project was checked out from version control.

ericoporto:
@morganwillcock add a message in the last commit

ericoporto:
Not sure if desired (I can just drop the commit), but I added it in the top of the script tab.

image

I followed the code in #1693

ivan-mogilko:
I think "Open containing folder" for a script will be mostly useful in ags4, where room scripts may be in room subfolders.

ericoporto:
I copied the command name from Visual Studio, but beyond opening the folder it also highlights the file itself.

ivan-mogilko:
I noticed that when MSVS does this, it does not reopen same folder multiple times, and opens existing explorer window containing desired folder if one is there.

Is it possible to do this?

ericoporto:
If you want I can add that shell stuff under Hacks.cs and then call that through the Utilities stuff, in case it's a filepath.

ivan-mogilko:
> If you want I can add that shell stuff under Hacks.cs and then call that through the Utilities stuff, in case it's a filepath.

Why only in case of a filepath?

> There is apparently an additional `/e` flag that keeps the same window

I tried that, but it did not do any difference. Could be i used it wrong, idk.

ericoporto:
I switched to use the Shell API and it appears to work keeping the same window!

PR #2135, opened 2023-09-16
Linux: Makefile uses pkgconfig for SDL2_sound

### LINKED ISSUE #2049: Could not build with make, only with cmake
I don't know too much about compilation, so I followed the Linux README under ags/blob/master/debian/README.md.

It said to install the SDL_Sound library, and then to run `make --directory=Engine`. The SDL_Sound installation installed this header file:
```
-- Installing: /usr/local/include/SDL2/SDL_sound.h
```

But `make --directory=Engine` then didn't find that file, and gave me the following error:
```
In file included from ../Engine/media/audio/openalsource.h:26,
                 from ../Engine/media/video/video.h:27,
                 from ac/global_api.cpp:68:
../Engine/media/audio/sdldecoder.h:22:10: fatal error: SDL_sound.h: No such file or directory
   22 | #include <SDL_sound.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
```

I then used cmake, and with cmake everything compiled without problems. I don't actually know why make didn't work and cmake did. Is cmake maybe easier on newbie people like me who don't to C++ compilations that often?

Since it worked for me, I think it could perhaps help people to spell out the steps for a cmake build in the Linux README.

ericoporto:
Marking this as draft because I am not sure if this fixes the issue 

https://github.com/icculus/SDL_sound/blob/main/cmake/SDL2_sound.pc.in

You can see it's intended that the included file is in SDL2 directory, but from what I understand it could support two SDL2 directories.

fix #2049

### DIFF
--- a/Engine/Makefile-defs.linux
+++ b/Engine/Makefile-defs.linux
@@ -31,11 +31,14 @@ else
 endif
 LIBS += $(shell pkg-config --libs vorbisfile)
 
+SDL2_SOUND_CFLAGS = $(shell pkg-config --cflags SDL2_sound)
+SDL2_SOUND_LIBS = $(shell pkg-config --libs SDL2_sound)
+
 SDL2_CFLAGS = $(shell sdl2-config --cflags)
 SDL2_LIBS = $(shell sdl2-config --libs)
 
-CFLAGS += $(SDL2_CFLAGS)
-LIBS += $(SDL2_LIBS) -lSDL2_sound
+CFLAGS += $(SDL2_CFLAGS) $(SDL2_SOUND_CFLAGS)
+LIBS += $(SDL2_LIBS) $(SDL2_SOUND_LIBS)
 
 LIBS += -ldl -lpthread -lm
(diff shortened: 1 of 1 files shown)


ericoporto:
Actually, even if this doesn't particularly fix that issue, I have a feeling this is the right thing to do, because the other dependencies also use pkg-config.

@rathann , hi! I know you maintain the ags in Fedora, what do you think of using pkg-config for SDL2_sound ?

rathann:
> Actually, even if this doesn't particularly fix that issue, I have a feeling this is the right thing to do, because the other dependencies also use pkg-config.
> 
> @rathann , hi! I know you maintain the ags in Fedora, what do you think of using pkg-config for SDL2_sound ?

I think using pkg-config or cmake is the right thing to do. SDL2_sound comes with a pkg-config file and a cmake file, so you can use either to detect it.

ivan-mogilko:
I confirm this build for me.

PR #2136, opened 2023-09-16
Scale Character.z and view offsets along with the character

### LINKED ISSUE #1304: Character.z does not scale
**Describe the bug**
`Character.z` defines the point where the character "connects with the ground" (measured from the bottom of the sprite). In other words, how many pixels above/below the bottom of the sprite the character's Y-position is. Typical use-cases:

-When the feet of the character are not at the bottom of the sprite (e.g. character has a shadow)
-To make a character levitate or even sink into the ground (combined with a suitable walkbehind)

However, AGS does not scale `Character.z` when the character scales. This means that the z-offset is too large when the character is scaled down, and too small when it is scaled up. See [this thread](https://www.adventuregamestudio.co.uk/forums/index.php?topic=57776.0)

I hold that this is _never_ the appropriate behavior, and therefore an AGS bug.

**Expected behavior**
`Character.Z` should scale proportionally with the size of the character.

**Additional context**
Because of rounding, there might potentially be some jittering in the vertical positioning of the character that is moving vertically while scaling gradually.

ivan-mogilko:
Resolves #1304.

I make this as a draft for now, to have something that may be tested, as I have a uneasy feeling about this (i explain this below).

The purpose of this change is to make 3 Character sprite's offsets scale along with the regular Character's Scaling property, whether manually set, or got from the walkable area. This refers to following offsets:
* [Character.z](https://adventuregamestudio.github.io/ags-manual/Character.html#characterz)
* x and y offsets assigned by [Character.LockViewAligned](https://adventuregamestudio.github.io/ags-manual/Character.html#characterlockviewaligned) and [Character.LockViewOffset](https://adventuregamestudio.github.io/ags-manual/Character.html#characterlockviewoffset) methods.

Please note that properties themselves do not change with scaling, what changes is their effect, e.g.:
`real_z = character.z * character.Scaling / 100`

I added a game setting that enables this, called "Scale Character sprite offsets", in a "Backwards Compatibility" section. First, this is necessary for backwards compatibility, as imported projects would have to adjust their scripts, if they scaled these properties manually. Second, this is a backup option if someone will find themselves unable to achieve wanted effect with the new behavior.

Now, about this problem, which is also a root of my doubts. Thinking about potential uses of Character.z, in particular, I found that there may be 2 main uses for it:
1. Changing sprite's origin along Y axis. In my honest opinion it's a logical mistake to use this property for this, but to be fair, LockViewOffset is currently unusable for this purpose, as it does not apply to clicking and collision tests (as stated in the manual). Thus Character.z is the only option left. What I think is that AGS currently lacks proper "Origin X/Y" setting that would affect every interaction with a sprite. But that's something that has to be solved later (in ags4, perhaps).
2. Applying a "elevation" effect, where character would appear either levitating, or sinking in ground, water, and so forth.

Looking at these uses, use no1 should normally always be scaled along with the sprite itself, otherwise that would defeat the purpose, so no questions there.
Purpose 2 *depends on a visual perspective*. If a scene emulates a side-view, where character is scaled down as it travels "further", and scaled up as it comes "closer", then this "elevation" effect definitely should be scaled along.
However, if the game features a different kind of perspective, where character's scaling is not changed due to it walking "further", then this elevation effect should likely not be scaled either, as it may be considered a property of "terrain" rather than character. In such situation, if character has to be scaled for any special reason, this may cause an issue with Character.z.

The reason I decided to make this PR nevertheless, is that, it seems, the situations where this offset has to scale are much more often, at least in regular adventure games.

There have been an interesting suggestion by @ericoporto in the #1304 thread:
> it's better to introduce a new OffsetY property and make this the behavior of the new property and make a note that the other property is deprecated, because I think updating this behavior in scripts is going to be non-trivial - I imagine a lot of games use this property very intensively, in different use cases, so if we are changing behavior it's best to introduce a new property with the new behavior. Later in ags4 we can remove the old behavior. 

Personally, I think that both such properties may have a purpose, and kept even in ags4 (maybe in a different form and with different names, if desired). Basically, I come to think that there would be a use of 2 offsets: one applied before scaling (and therefore - always scaled), and another applied after scaling (and therefore - never scaled by character's scaling). This is due to the nature of AGS meant to emulate various kinds of environment.

With this PR, as it's done now, it would be possible to get old behavior back by using the aforementioned switch. This is a backup.

I believe that in ags4 the sprite transformation properties should be rethought, especially properties that define sprite origin (and not only for characters).

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -116,7 +116,6 @@ struct CharacterInfo {
     char  scrname[MAX_SCRIPT_NAME_LEN];
     char  on;
 
-    int get_effective_y() const;     // return Y - Z
     int get_baseline() const;        // return baseline, or Y if not set
     int get_blocking_top() const;    // return Y - BlockingHeight/2
     int get_blocking_bottom() const; // return Y + BlockingHeight/2
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -83,7 +83,8 @@
 #define OPT_GAMETEXTENCODING 49 // how the text in the game data should be interpreted
 #define OPT_KEYHANDLEAPI    50 // key handling mode (old/new)
 #define OPT_CUSTOMENGINETAG 51 // custom engine tag (for overriding behavior)
-#define OPT_HIGHESTOPTION   OPT_CUSTOMENGINETAG
+#define OPT_SCALECHAROFFSETS 52 // apply character scaling to the sprite offsets (z, locked offs)
+#define OPT_HIGHESTOPTION   OPT_SCALECHAROFFSETS
 #define OPT_NOMODMUSIC      98 // [DEPRECATED]
 #define OPT_LIPSYNCTEXT     99
 
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -101,8 +101,9 @@ namespace AGS.Editor
          * 3.6.0.20       - Settings.GameTextEncoding, Settings.UseOldKeyboardHandling;
          * 3.6.1.2        - GUIListBox.Translated property moved to GUIControl parent
          * 3.6.1.3        - RuntimeSetup.TextureCache, SoundCache
+         * 3.6.1.9        - Settings.ScaleCharacterSpriteOffsets
         */
-        public const int    LATEST_XML_VERSION_INDEX = 3060103;
+        public const int    LATEST_XML_VERSION_INDEX = 3060109;
         /*
          * LATEST_USER_DATA_VERSION is the last version of the user data file that used a
          * 4-point-4-number string to identify the version of AGS that saved the file.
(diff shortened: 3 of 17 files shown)


ivan-mogilko:
I tested this myself, and it works as stated; so unless someone has a reason not to, I will merge this.

ericoporto:
In the Editor, I wouldn't put this in Backwards Compatibility, it doesn't feel like going from old to new but more just an option for change of behavior.

>  it seems, the situations where this offset has to scale are much more common, at least in regular adventure games.

I get this, but this would at least be something that would break things I make, mostly because I try to do things a bit out of the regular. I think Character movement would be a better place to be in the settings.

ivan-mogilko:
> In the Editor, I wouldn't put this in Backwards Compatibility, it doesn't feel like going from old to new but more just an option for change of behavior.

That's a good point, I might move this to "Character movement", except I also now think to rename this category to "Character behavior", since it will be not only about movement.

ericoporto:
Yeah, that could work for me. Just second guessing myself here, does it goes to backwards compatibility because this flag would be something removed in ags4 - supposing this is redone there at some point?

ivan-mogilko:
> does it goes to backwards compatibility because this flag would be something removed in ags4 - supposing this is redone there at some point?

Yes, I think there should not be such switch, but character properties that work one way all the time.
One of the suggestions is to have separate offsets (x,y), one of which is applied before scaling and therefore is scaled, and another which is applied after scaling and therefore is never scaled.

There definitely have to be Origin property(s), because LockViewOffset is not usable for the general case (e.g. you cannot use it for normal walking animation). But I do not know if this Origin prop. will replace current "z" or not; this is something to think through.

PR #2138, opened 2023-09-17
Editor: write INI files using IniUtils from AGS.Native

### LINKED ISSUE #2137: Support acsetup.cfg in utf-8 encoding (and maybe in ansi too)
**Describe the problem**
acsetup.cfg is the last file that only supports ansi encoding, but unfortunately there are some free strings in the file that can have utf-8 names in them.

**Suggested change**
It would be nice if acsetup.cfg could support utf-8. It would be even nicer if utf-8 and ansi could be supported, somehow. Supposing it can only support one encoding, then we just always assume things are utf-8 in acsetup.cfg from now on.

The editor has to be modified so that the acsetup.cfg files it creates are utf-8 too.

**Additional context**

- See forums here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=61002.0

ivan-mogilko:
Resolves #2137

Because I also wish to patch this in release-3.6.0 branch, I chose the most straightforward method, and used IniUtils from C++ part of the Editor to merge new options with the config file.

Gather up dictionary of sections in C# code, and then pass to NativeProxy, where it converts to the native strings using TextHelper, and writes to ini using IniUtils.

I tested this and confirmed that the resulting file contains "titletext" setting in UTF-8, and that winsetup can successfully display this text in its caption.

Same is done for android.cfg.

I do not think that there are any other options that may contain utf-8 text (?), but will leave to others to comment on this.

### DIFF
--- a/Common/util/ini_util.h
+++ b/Common/util/ini_util.h
@@ -84,8 +84,6 @@ namespace IniUtil
     // If item already exists, only value is overwrited, if section exists,
     // new items are appended to the end of it; completely new sections are
     // appended to the end of text.
-    // Source and destination streams may refer either to different objects,
-    // or same stream opened for both reading and writing.
     // Returns FALSE if the file could not be opened for writing.
     bool Merge(const String &file, const ConfigTree &tree);
     // Similar to the above, but merges the key-value tree into the provided
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1415,7 +1415,7 @@ namespace AGS.Editor
             return result;
         }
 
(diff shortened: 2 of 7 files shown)


ericoporto:
Small question, does the engine already support acsetup.cfg in utf-8? (I see this only changes the Editor)

ivan-mogilko:
> Small question, does the engine already support acsetup.cfg in utf-8? (I see this only changes the Editor)

Engine does not need anything extra to support it, unless there are options that it must interpret as either ansi or utf-8.

ericoporto:
I don't think such option is needed, thanks for clarifying it!

ericoporto:
Pretty cool, I tried this, and it works! I can even pass an emoji to the Setup Window Title!!!
image

Issue #2125, opened 2023-09-07
Ordering Character to walk while it is already walking will make it halt for 1+ frames


ivan-mogilko:
**Describe the bug**
If a Character is walking, and is ordered to walk again (e.g. by calling Character.Walk), then the character will visibly halt on a current position for 1 or more frames (this is unclear) before starting to move towards the new destination.

This behavior may be quite irritating for players, who may have a habit to click around multiple times while character is walking. Similarly this is annoying to see when you change your mind in the middle of the walk, and order to walk elsewhere. Generally, this seems like a mistake in the engine.

I noticed that this behavior is best observed with "movement linked to animation" and slow movement speed, but it also actually happens if that setting is off too, only may be harder to notice.

A curious thing is that although movement stutters, the walking animation does not have same problem.

**AGS Version**
As far as I know, this happens for a very long time. At least in every version in 3.* generation, but maybe in earlier ones too.

**To Reproduce**
1. Configure player character to have MovementLinkedToAnimation enabled, and MovementSpeed 1. This is not required for the error to happen, but will make it easier to notice and thus test it.
2. Start the game, order character to walk. Then start clicking repeatedly on the same spot.
3. Notice how character "stutters" whenever there's another walking order. If you click fast enough, it will never change position at all (this is because movement speed is low). On another hand, notice that animation runs without extra delays.

**Expected behavior**
If a new movement order is issued while Character is walking, there has to be a smooth transition onto the new path. That is - a character does not halt and starts walking on a new path same game frame. In other words, this should look as if it keeps walking same path as before.

**Additional context**
There's a number of timers related to walking. I counted 3 that may be relevant: `wait`, `walkwait` and `walkwaitcounter`.
This may also be related to how pathwalking works in AGS, where the path is split into a number of "steps", and the rules of how these steps are performed. Or not, I could not figure out yet whether that has anything to do with this.

morganwillcock:
> If a new movement order is issued while Character is walking, there has to be a smooth transition onto the new path.

Some people may prefer to have the delay as an indication that the instruction to move was registered, otherwise something like a mouse click may not have any feedback to indicate that it was successfully received.

If the delay gets removed entirely it is probably worth checking that it would be  possible to recreate the original behaviour with some scripting.

ivan-mogilko:
I possibly found a culprit. It appears to be a part of the MoveList logic.

[...]

PR #2141, opened 2023-09-17
Allow to change walking speed during walking

ivan-mogilko:
Done after a conversation in #2125 , not directly related, but I wanted to experiment with recalculating movelists.

Try to support changing Character's movement speed while character is moving. This is done by recalculating movelist's stages and updating current stage's distance (`onpart` variable).
Recalculation is done as a rather trivial math, where we divide on the old speed factor and multiply on the new speed factor. This does not involve any pathfinding at all, as we only need to "replace" a speed factor in x/ypermove arrays, so to speak.

I did this only in case of a "new" pathfinder, because old one is used in old games only, where changing walkspeed during walk was prohibited. But technically this code is applicable to both.

Preliminary tests show a good progress, so I decided to put a PR for a test.

Example of a game that may help to test this, contains a slider to change walking speed:
https://github.com/adventuregamestudio/ags/files/12659434/test--361walkspeed.zip

Things to test:
- [x] Changing walk speed works without character jumping back or forth, regardless of the speed difference (old vs new).
- [x] Character keeps walking visibly along the same line after changing speed.
- [x] Destination check, after changing speed during a walk: character should arrive onto exact position where ordered (it should not depend on speed or whether speed has changed).
- [x] Writing and reading saves done in this version with walking character(s).
- [x] Reading saves done in older versions with walking character(s): they should continue walking with the same speed.

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -107,7 +107,7 @@ struct CharacterInfo {
     short pic_xoffs; // this is fixed in screen coordinates
     short walkwaitcounter;
     uint16_t loop, frame;
-    short walking;
+    short walking; // stores movelist index, optionally +TURNING_AROUND
     short animating; // stores CHANIM_* flags in lower byte and delay in upper byte
     short walkspeed, animspeed;
     short inv[MAX_INV];
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -1928,7 +1928,7 @@ int doNextCharMoveStep (CharacterInfo *chi, int &char_index, CharacterExtras *ch
 
         if ((chi->walking < 1) || (chi->walking >= TURNING_AROUND)) ;
         else if (mls[chi->walking].onpart > 0) {
-            mls[chi->walking].onpart --;
+            mls[chi->walking].onpart -= itofix(1);
             chi->x = xwas;
             chi->y = ywas;
         }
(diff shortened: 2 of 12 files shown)


ericoporto:
Pretty impressive! I tried and it works!

ivan-mogilko:
Turned MoveList::onpart value (progress) into a fractional value. This lets to update the progress very accurately when changing walk speed, and fixes position "jumps" in case of a big change in speed.

TODO:
1. Update the save format, because engine must know to convert this value from int to fixed point on loading older saves.
2. Potentially, engine could enforce that the first step after recalculation will bring `onstep` to a full integer, to make walking consistent (although I am not sure which issues may occur if I don't do that).

ivan-mogilko:
> A question, because of the change in save format, have you considered moving to the code in ags4 where the fixed math was replaced with floats? (just thinking about the merge with that branch later)

I do not want to backport that particular change; frankly it was not that important for the common case, maybe only have better potential in the long-term. I did backport the good idea of saving movelists in their own component, which is how they should have been saved from the start...

Other than that, the save format change done inside MoveList in this PR is very small, and should be fine to merge to with ags4.

ivan-mogilko:
I'm actually beginning to wonder why make `onpart` fixed point, as it is actually ever used in calculations being converted to float:
https://github.com/adventuregamestudio/ags/blob/855e8b24514fcd06b3071ba584a2543635bc5657/Engine/main/update.cpp#L152-L153

If so, then it may as well be turned into `float` for better precision right away. That should not change anything in the old games either.

Issue #1685, opened 2022-06-12
[Feature request] Right Click variable to find in tree or a search function.


tarnos12:
**Describe the problem**
In larger projects finding a character/item/GUI might be a problem as there is no way to search the tree.
This can be somehow avoided by using folders, but if I need to find something specific that I am currently referencing in a script then I have to go through a big chunk of menus to find it(usually a problem when looking for GUI or an Item, even more so when looking for a GUIControl, you have to either remember which GUI holds that specific control or search through them all)

**Suggested change**
Right click a variable in a script should give us an option similar to "Navigate(in tree)" option.

Second possible way is to add a search bar that would search by name of items/GUI/GUIControl/Character/rooms/scripts etc.(like a filter)
This could include auto complete(like in script which is already aware that we have `cPlayer or Label1`)
Ideally it would work like a filter and hide things that do not match the search(might also include special characters for extra filtering options such as || which works as "or" so I can search for `cPlayer || gGui1 || myScript || iCup`)

The filter would be extra useful for cases where I have multiple states of an item such as "iCupEmpty, iCupCoffee, iCupTea, iCupJuice` I could search for `iCup` and get all of those in a list.

tarnos12:
> In my opinion these are very different things. A find in tree that jumps and select imposes few changes to the tree ui, but the filter search has many implications.

I am referring to the already implemented feature which allows you to right click a tab of a script/character/item/gui to find it in a tree.
Ideally it would extend to the script itself, so I can right click a variable and search in tree.
So there are probably no changes needed to the tree ui.
I don't know how it's done(perhaps it is required to have the specific tab open/active for the navigation to work)

It's been implemented here: [Forum post](https://www.adventuregamestudio.co.uk/forums/?topic=46378.0)
https://i.imgur.com/4kJX7.png

ericoporto:
> It's been implemented here: [Forum post](https://www.adventuregamestudio.co.uk/forums/?topic=46378.0)
> https://i.imgur.com/4kJX7.png

Thanks for bringing up this thread, I wasn't aware of that! 

But there's been a confusion, I meant the navigate in tree feature is of a much lesser scope than the project search, which can be done in many different ways. Here's an example of project search, with autocomplete, but without the filter part... Once filtering is added, than the questions I posted need answer during design to be able to be implemented.

![search_box](https://user-images.githubusercontent.com/2244442/173562970-917ad672-6148-448e-af78-c276274bcc8d.gif)

[...]

PR #2142, opened 2023-09-18
Editor: support go-to for few game instances

ericoporto:
Adds support for AudioClip, Characters, Dialogs, GUIs and Inventory Items.

This is part of #1685

### DIFF
--- a/Editor/AGS.Editor/Panes/ScriptEditorBase.cs
+++ b/Editor/AGS.Editor/Panes/ScriptEditorBase.cs
@@ -541,11 +541,59 @@ namespace AGS.Editor
             {
                 if (foundInScript.FileName == AGSEditor.BUILT_IN_HEADER_FILE_NAME)
                 {
-                    Factory.GUIController.LaunchHelpForKeyword(_goToDefinition);
+                    if (_goToDefinition == "player")
+                    {
+                        CharactersComponent charactersComponent = Factory.ComponentController.FindComponent<CharactersComponent>();
+                        charactersComponent.ShowPlayerCharacter();
+                    }
+                    else
+                    {
+                        Factory.GUIController.LaunchHelpForKeyword(_goToDefinition);
+                    }
                 }
                 else if (foundInScript.FileName == Tasks.AUTO_GENERATED_HEADER_NAME)
                 {
-                    Factory.GUIController.ShowMessage("This variable is internally defined by AGS and probably corresponds to an in-game entity such as a Character or Inventory Item.", MessageBoxIcon.Information);
+                    if (found is ScriptVariable)
+                    {
+                        ScriptVariable sVar = found as ScriptVariable;
+                        string varType = sVar.Type;
+                        string varName = sVar.VariableName;
+
+                        // FIX-ME: at some version (ags4?) we can break IEditorComponent compatibility and just move the script type to the component
+                        // then we can add FindComponentThatManageScriptType to the ComponentController and get the managed script types from the components 
+                        BaseComponent component = null;
+                        switch (varType)
+                        {
+                            case "InventoryItem":
+                                component = Factory.ComponentController.FindComponent<InventoryComponent>();
+                                break;
+                            case "Character":
+                                component = Factory.ComponentController.FindComponent<CharactersComponent>();
+                                break;
+                            case "Dialog":
+                                component = Factory.ComponentController.FindComponent<DialogsComponent>();
+                                break;
+                            case "GUI":
+                                component = Factory.ComponentController.FindComponent<GuiComponent>();
+                                break;
+                            case "AudioClip":
+                                component = Factory.ComponentController.FindComponent<AudioComponent>();
+                                break;
+                        }
+
+                        if(component != null)
+                        {
+                            component.ShowItemPaneByName(varName);
+                        }
+                        else
+                        {
+                            Factory.GUIController.ShowMessage("This variable is internally defined by AGS and probably corresponds to an in-game entity such as a GUIControl or Inventory Item.", MessageBoxIcon.Information);
+                        }
+                    }
+                    else
+                    {
+                        Factory.GUIController.ShowMessage("This is internally defined by AGS.", MessageBoxIcon.Information);
+                    }
                 }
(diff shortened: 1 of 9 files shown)


Editor/AGS.Editor/Panes/ScriptEditorBase.cs
```diff
@@ -541,11 +541,57 @@ private void GoToDefinition(string structName, string memberName)
...
                 }
                 else if (foundInScript.FileName == Tasks.AUTO_GENERATED_HEADER_NAME)
                 {
-                    Factory.GUIController.ShowMessage("This variable is internally defined by AGS and probably corresponds to an in-game entity such as a Character or Inventory Item.", MessageBoxIcon.Information);
+                    if (found is ScriptVariable)
+                    {
+                        ScriptVariable sVar = found as ScriptVariable;
+                        string varType = sVar.Type;
+                        string varName = sVar.VariableName;
+                        if (varType == "InventoryItem")
```
ivan-mogilko, line 561: Hmm, Component has a base class. Would it be possible to:
- make virtual method SelectByName
- make method FindComponentByTypeName or similar?

in which case all this code would be replaced by something like:
```
Component(?) comp = Factory.ComponentController.FindComponentByName(varType);
if (comp)
     comp.SelectByname(varName);
```

which will also support components created by plugins.

ericoporto: I don't know how to do that, it would require some sort of type list so a component can say all things it can handle - Audio has the clips but maybe also the types?

ivan-mogilko: I also suggest to rename SelectByName to something that makes it clear what is selected and where, like SelectItemInProjectTree.
The reason for such suggestion is that the Component works a wide range of things (project explorer, panes, some with files), so it has to be certain. For example, maybe we will have a method for opening an panel to edit an item, etc.

ivan-mogilko: Also, tbh I doubt `SelectItem` is a good term here, as judging by the implementation, this method not just selects it in a project tree, but opens a pane for preview/edit. The method's name must clearly tell what is going to happen.

ivan-mogilko: > I don't know how to do that, it would require some sort of type list so a component can say all things it can handle - Audio has the clips but maybe also the types?

Well, if it's difficult now, perhaps at least make a function that does this selection and returns a component, just to have something that may be called elsewhere if a need appears.

ericoporto: About `SelectItem`, what about `ShowItemPaneByName` ?

ericoporto: There is a big note there to not modify the IEditorComponent or it will break plugins....

ivan-mogilko:
This is not directly related to this pr, but because there's a added comment about breaking compatibility, I'd like to elaborate a bit on changing the IEditorComponent interface, for the future.

When changing an interface leads inevitably to breaking plugins, there's a common method: make a new interface (e.g. literally `IEditorComponent2`), inheriting the first one, and support plugins registering either under old or new one. This requires some managing of these interfaces internally, because you have to distinguish objects registered as old and new ones. In C# this may be done by comparing types, among other things.

There's already a practical example of this method used for the font renderer interface in the engine, see:
https://github.com/adventuregamestudio/ags/blob/1be534c371d0631495f6035ea2beb019a6644c6e/Common/font/agsfontrenderer.h#L41
https://github.com/adventuregamestudio/ags/blob/1be534c371d0631495f6035ea2beb019a6644c6e/Engine/plugin/agsplugin.h#L545

On an additional note, if I remember correctly, in C and C++ it's possible to extend the interface without breaking stuff so long as you append new methods after old ones, as that makes them last in the virtual table (I honestly don't remember if this is by standard in C++, but that at least works in common), so it's still safe to call the old methods. But in order to make this safe you also need to distinguish the "interface version", so there have to be a way to check what kind of object you are dealing with.
I do not remember if this works in C#, as it may have its own rules, but that would be interesting to check out.

Editor/AGS.Editor/Panes/ScriptEditorBase.cs
ivan-mogilko: I still find it strange to have this list inside ScriptEditorBase class, because this makes a script editor know everything about which component does what; this is why I was proposing to have this selection in some global class, like ComponentController, or other, where would make more sense.

But i can look into adjusting this later myself, if you don't want to bother with this.

ericoporto: Uhm, maybe if the component controller iterates in the IEcomponent `_components` and then checks if they are BaseController, and then ask for the script type the component manages, maybe it could work.

ivan-mogilko: > Uhm, maybe if the component controller iterates in the IEcomponent _components and then checks if they are BaseController, and then ask for the script type the component manages, maybe it could work.

Hmm, that's actually one possible thing to do this,
but really, if that will be done with this big hardcoded switch - it would be okay for now too. My point is simply that in my opinion this search belongs to a controller class, rather than a script editor class.

ericoporto: @ivan-mogilko made a tentative commit with a different approach, see what do you think about it.

ivan-mogilko: Yes, I think it's fine; and it does not add anything to external interface, meaning this may be reverted if there's a better idea in the future.

ericoporto: Cool, I will mark this conversation as resolved then!

ivan-mogilko:
Tested, and this works well, both from the regular script and dialog.

Regarding Gui controls, I guess there may be an extra method for this implemented later in gui component, which searches all guis for them.

PR #2145, opened 2023-09-20
Support smooth transition when ordering Character to Walk while it is already walking

### LINKED ISSUE #2125: Ordering Character to walk while it is already walking will make it halt for 1+ frames

ivan-mogilko:
Fixes #2125.

This is an attempt to keep character moving seamlessly when a new walking order is made while the character is already walking.

WARNING: this PR is based on top of #2141, because it requires fractional `MoveList::onstep`, which is already implemented there.

Historically, engine had a problem with this, because of how a walking step is calculated. In short, a MoveList struct increments a number of "steps", but these "steps" may be less than pixel, in which case it takes >1 frames to make a visual change.
When a new walking order is made, this step counter is reset. This means that if player, for example, keeps clicking around, the character will be "stuttering".

Here I tried following **solution**: if the character is already walking, then save last value of a fraction of "unfinished" pixel step, and apply it to the new movelist as a starting step counter.
This idea sounded logical, and works pretty well for in case of manual orders.

I must note however, that still, due to the nature of this counter, there may be issues if a new Walk order is assigned each game frame. This may be done e.g. in script. In this case, depending on various factors (relation between x and y distance, walking speed, linking to animation, etc) the character may visually walk in a "ladder" instead of a straight line. This is because there's not enough "step value" each time to make a move along two axes smoothly.
This is probably will not be noticed when commanded by a player, as players commonly do not click that frequently.

At this point I'm uncertain whether fixing the above would be only a matter of changing the types of variables (fixed -> float, in attempt to reduce conversion mistakes), or a matter of changing the core movelist mechanics, and transfer more of a "progress" from previous movelist to the new one.

---

This is the test project I've been using:
[test--361walkspeed-source.zip](https://github.com/adventuregamestudio/ags/files/12677334/test--361walkspeed-source.zip)
Compiled:
[test--361walkspeed.zip](https://github.com/adventuregamestudio/ags/files/12677337/test--361walkspeed.zip)

This is similar project to the one I used in #2141, except I added a button that toggles continuous automatic reorder of Walk.


ivan-mogilko:
Per @morganwillcock 's concern:
> Some people may prefer to have the delay as an indication that the instruction to move was registered, otherwise something like a mouse click may not have any feedback to indicate that it was successfully received.

> If the delay gets removed entirely it is probably worth checking that it would be possible to recreate the original behaviour with some scripting.

I believe this may be done by calling `Character.StopMoving()` before the next Walk(). As bare StopMoving will reset current movelist without saving any progress, and the next Walk call will not have anything to get a progress data from.

ericoporto:
> WARNING: this PR is based on top of https://github.com/adventuregamestudio/ags/pull/2141, because it requires fractional MoveList::onstep, which is already implemented there.

Do I need to recompile my game with the version here to have this behavior or simply running a previous existing game will already have this behavior? I just wanted to test this change, this is why I ask.

ivan-mogilko:
> Do I need to recompile my game with the version here to have this behavior or simply running a previous existing game will already have this behavior? I just wanted to test this change, this is why I ask.

There's no changes to game options or script, so no need to recompile anything.

ericoporto:
I tested this in my older games and in Future Flashback and it looks like all keeps working nicely, so at least on backwards compatibility, this is all good!

ivan-mogilko:
Note: I'm currently looking into the #935, being given a test game which reproduces this 100%. Mainly I like to find out if this is caused by fixed point format limitations, that is - rounding mistakes. Because if it is, then it may make sense to backport float math into 3.6.1, and then change of `onpart` value to fixed will be redundant.

Issue #1175, opened 2021-01-14
[feature suggestion] Views as enums


ericoporto:
Today AGS creates a define macro for each view according to it's name (`NAME_OF_VIEW`). 

I was thinking about having it be a `eVIEW_NAME_OF_VIEW` or `eView_name_of_view` or similar. 

This way parts of the AGS Script API that accepts Views could use the enum type in it's parameters, instead of the current `int view` a `VIEW view` or what will be the type name used for all views. This way the views will appear as suggestions in the autocomplete on the editor when writing that parameter.

Since enums are just int I don't think this will break anything.

It's a new feature so I think it would be for AGS4.

ivan-mogilko:
Unless I miss anything, this might be very easy to do, just replace whatever is generated (`#define`s probably) with enum.

ivan-mogilko:
I recall @fernewelten was suggesting to do similar thing for room numbers https://github.com/adventuregamestudio/ags/issues/1135#issuecomment-733844485

morganwillcock:
Is it enforced that View names are unique?

ericoporto:
> Unless I miss anything, this might be very easy to do, just replace whatever is generated (#defines probably) with enum.
> (and adjust function prototypes)

Yeahp, exactly that :)

ivan-mogilko:
> Is it enforced that View names are unique?

I don't remember if it's checked at creation (probably yes), but every object name in AGS is forced to be unique simply because it is used to generate variables and macros for script compilation.

fernewelten:
> Is it enforced that View names are unique?

`enum` values don't need to be unique, either, so this wouldn't alter anything from that viewpoint. 

But seemingly it does: I've just tried to assign an identical name to two views in 3.5, and the Editor complains, “This script name is already used by another item.“

ericoporto:
Apparently the method is `IsScriptNameAlreadyUsed` in the `Editor/AGS.Types/Game.cs`.

rofl0r:
is this the same thing happening here ?

https://github.com/rofl0r/ascc/commit/948d5642f52ff6671b2a4c582f1bd0d2e1e095af

```
define ICONBAR FindGUIID("ICONBAR")
```

morganwillcock:
This seems like something that is very useful for how everything works right now, but if this is implemented as a `#define` doesn't that preclude any dynamic creation of resources?

[...]

PR #2148, opened 2023-09-23
Editor: support go-to for GUI controls, Cursors, Audio Types and Fonts

### LINKED ISSUE #1685: [Feature request] Right Click variable to find in tree or a search function.

ericoporto:
Btw, I looked into creating a ScriptElement in EditorFeatures/AutoComplete, which both ScriptVariable and ScriptEnumValue could extend from, but ScriptVariable name is VariableName and ScriptEnumValue name is just Name, so I couldn't easily make both extend from the same thing without breaking AGS Types api.

Ah, `#define` macros don't have the concept of what they belong to, so there isn't an easy way to add VIEW macro support (in ags4, we could have view as enum after #1175).

I also noticed the when launching the help, if it detects that the found thing is an enum, it could use it's type to instead search the help for that, because currently the enum values in the built-in header are not index, only their enum type. I am not doing this here because I am thinking if I can add them to the index somehow in the ags-manual.

fix #1685

### DIFF
--- a/Editor/AGS.Editor/Components/GuiComponent.cs
+++ b/Editor/AGS.Editor/Components/GuiComponent.cs
@@ -193,7 +193,7 @@ namespace AGS.Editor.Components
             _guiController.AddOrShowPane(document);
 		}
 
-        public override IList<string> GetManagedScriptTypes()
+        public override IList<string> GetManagedScriptElements()
         {
             return new string[] { "GUI" };
         }
--- a/Editor/AGS.Editor/Components/GuiComponent.cs
+++ b/Editor/AGS.Editor/Components/GuiComponent.cs
@@ -195,7 +195,7 @@ namespace AGS.Editor.Components
 
         public override IList<string> GetManagedScriptElements()
         {
-            return new string[] { "GUI" };
+            return new string[] { "GUI", "Label", "Button", "Slider", "ListBox", "TextBox", "InvWindow" };
         }
 
         public override void ShowItemPaneByName(string name)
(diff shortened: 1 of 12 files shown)


Editor/AGS.Editor/Components/GuiComponent.cs
```diff
@@ -209,6 +209,19 @@ public override void ShowItemPaneByName(string name)
...
                 }
+                
+                foreach(GUIControl gctrl in g.Controls)
+                {
+                    if(gctrl.Name == name)
+                    {
+                        _guiController.ProjectTree.SelectNode(this, GetNodeID(g));
+                        ShowOrAddPane(g);
+                        GUIEditor guiEditor = _documents[g]?.Control as GUIEditor;
+                        guiEditor?.SelectGUIControl(gctrl);
```
ivan-mogilko, line 220: I think that calling SelectGUIControl may be redundant, if i remember right the gui editor should react to property grid selecting a new object.

ericoporto: You are correct, adjusted it!

ivan-mogilko:
Hah, it works! this is going to be a useful feature :).

ivan-mogilko:
There's just one oversight: one message still mentions that Gui controls are not supported:
```
"This variable is internally defined by AGS and probably corresponds to an in-game entity such as a GUIControl or Room Object."
```

I propose to simplify this and simply say "This variable is internally defined by AGS and does not support Go to Definition at the moment".

ericoporto:
Adjusted!

PR #2150, opened 2023-09-24
Fix "walking on place" bug in 3.6.1

### LINKED ISSUE #935: Walk animation keeps going after character arrived at destination

ivan-mogilko:
Fixes #935.

AGS historically has a hidden problem: its pathfinder and walking algorithm uses math with relatively low precision. This may result in rounding mistakes. What this means in practice is that under some circumstances the "movement direction" (vector) is slightly off, e.g. more towards X than Y (or other way). In such case the character may reach the X destination while Y destination is still few pixels away.

How is this issue solved currently in AGS (and for a long time): when AGS finds out that one coordinate is complete, but another is not (for example X done and Y is still not reached), it starts waiting until character reaches Y, but also accepts a certain "mistake", in 2-3 pixels. If character is already within 2-3 pixels from the destination, then it just stops. This makes its real stopping point inaccurate, but nothing else happens. However, if character is further away, then it keeps "walking", but because the movement vector is still pointing in the same direction, usually it takes few moments for it to reach the proper Y coordinate. This results in a buggy effect known as "walking in place".

The most natural way to fix this would be to use floating-point math instead of a fixed-point math, as that would increase precision. But there's a concern about backwards compatibility. Knowing a history of AGS games, there is a non-zero chance that some old game had position checks in script, requiring character to pass or stop at the exact pixel.
I recall someone already found a issue in an old game after we introduced a new pathfinder, so we actually had to return an old one for old games.
This is why I'd rather leave this for ags4 branch (where it is already implemented).

This commit introduces a more "compatible" solution. What it does is: turn the walking vector to make it point right along the unfinished axis, while keeping its length (that means - walking speed), in order to reach the destination faster.
This should make character make some extra short move, properly animating, but this may look more natural than just moving legs in place.

NOTE: I still keep the "destination fix" for the old games, as they may be scripted specifically to test where character has stopped moving.

NOTE 2: This PR also contains MoveList refactor, backported from ags4 branch, but it does not change a save format.
But here I went further and also picked some generic code into a separate function that deals with "finishing move".

### DIFF
--- a/Common/util/geometry.h
+++ b/Common/util/geometry.h
@@ -118,6 +118,11 @@ struct Point
     {
         return Point(X - p.X, Y - p.Y);
     }
+
+    inline bool Equals(const int x, const int y) const
+    {
+        return X == x && Y == y;
+    }
 };
 
 struct Line
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -1896,7 +1896,7 @@ int has_hit_another_character(int sourceChar) {
 int doNextCharMoveStep (CharacterInfo *chi, int &char_index, CharacterExtras *chex) {
     int ntf=0, xwas = chi->x, ywas = chi->y;
 
-    if (do_movelist_move(&chi->walking,&chi->x,&chi->y) == 2) 
+    if (do_movelist_move(chi->walking, chi->x, chi->y) == 2) 
     {
         if ((chi->flags & CHF_MOVENOTWALK) == 0)
             fix_player_sprite(&mls[chi->walking], chi);
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -182,7 +182,7 @@ void Character_AddWaypoint(CharacterInfo *chaa, int x, int y) {
         return;
     }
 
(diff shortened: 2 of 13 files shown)


ericoporto:
Tested in some projects and appears to work and get the character to reach the destination correctly. Appears safe to merge, but not sure if you plan to put #2141 and #2145 in here before or rebase those after.

ivan-mogilko:
> but not sure if you plan to put https://github.com/adventuregamestudio/ags/pull/2141 and https://github.com/adventuregamestudio/ags/pull/2145 in here before or rebase those after.

I think I will merge this and rebase others, because this pr contains code cleanup.

ivan-mogilko:
Argh! i just realized that i did a "small" mistake here, which practically duplicates the speed of finishing the last step.
This may be noticed if the "smaller" axis is completed earlier than the "longer" one.

PR #2151, opened 2023-09-27
Editor: more fixes for compiled files cleanup

### LINKED ISSUE #1976: Editor: delete/rename old compiled binaries when the "Game Filename" is changed in project
Since the Editor allows to change the game's "filename" by editing a property in General Settings, following problem may occur: an old *.ags and possibly *.exe file would be left in Compiled/... folders. During further game compilations the extra *.ags files may get copied along from the Compiled/Data folder. Also this may cause problems when running a game using a standalone engine.

The proposal is to either **delete** or **rename** old compiled binaries found in Compiled folder and respective subfolders when "Game Filename" is modified.

---

Did a test of all builds to see the current situation:

* Data folder does not keep an old .ags file, so it's probably **fine**;
* Linux build does not keep old .ags, but it keeps old **startup script**;
* Web build keeps **old .ags**;
* Windows keeps an **old .ags** in case you have "Attach game data to exe" disabled in general settings, but not the old .exe;
* Android does not keep old .ags, so it's probably **fine**;

ivan-mogilko:
Fixes #1976, again.

* Picked out DeleteCommonGameFiles(), to be able to call same from any BuildTarget;
* Added cleanup of split resources.
* Call cleanup for the current filename too, just to be sure everything is deleted beforehand. This fixes split resources not being deleted if this option changes in General Settings.

Also:
* Save user data file (aka WorkspaceState) right after game building, to ensure that any information amended during build is always saved on disk regardless whether user saves or not the project itself.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -937,15 +937,6 @@ namespace AGS.Editor
             return errorToReturn;
         }
 
-        private void DeleteAnyExistingSplitResourceFiles()
-        {
-            string dir = Path.Combine(OUTPUT_DIRECTORY, DATA_OUTPUT_DIRECTORY);
-            foreach (string fileName in Utilities.GetDirectoryFileList(dir, this.BaseGameFileName + ".0*"))
-            {
-                Utilities.TryDeleteFile(fileName);
-            }
-        }
-
         private void CreateAudioVOXFile(bool forceRebuild)
         {
             List<string> fileListForVox = new List<string>();
@@ -989,6 +980,10 @@ namespace AGS.Editor
             var buildNames = Factory.AGSEditor.CurrentGame.WorkspaceState.GetLastBuildGameFiles();
             foreach (IBuildTarget target in BuildTargetsInfo.GetSelectedBuildTargets())
             {
+                // Primary cleanup
+                target.DeleteMainGameData(Factory.AGSEditor.BaseGameFileName);
+
+                // Old files cleanup (if necessary)
                 string oldName;
                 if (!buildNames.TryGetValue(target.Name, out oldName)) continue;
                 if (!string.IsNullOrWhiteSpace(oldName) && oldName != Factory.AGSEditor.BaseGameFileName)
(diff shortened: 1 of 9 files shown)


ivan-mogilko:
@ericoporto made editor to save user data right after building the game, which should resolve the issue you mentioned in comments to #1976.

morganwillcock:
Is it definitely safe to save WorkspaceState when the Editor is started with the `/compile` parameter?

(I imagine that is probably is safe, but potentially there is a risk that WorkspaceState gets mistakenly reset if the state is never loaded when the Editor is told to just compile and exit).

ericoporto:
That is interesting, do you know how to check for that command line parameter when the editor is running? Ages ago persn did an experimental version of the Editor that had ifdefs to not do certain things so it could run as a web server for compiling games. I remember it was good at not showing pop-ups.

ivan-mogilko:
> Is it definitely safe to save WorkspaceState when the Editor is started with the `/compile` parameter?
> 
> (I imagine that is probably is safe, but potentially there is a risk that WorkspaceState gets mistakenly reset if the state is never loaded when the Editor is told to just compile and exit).

That's a good question, and it raises much more questions, like, is it safe to save the game itself, and so forth, because this is what editor is doing whenever building.

But I noticed that /compile flag actually leads into a different place than the ones I edited:
https://github.com/adventuregamestudio/ags/blob/fc336554a25bb763f721fb1a694a09771ad09b07/Editor/AGS.Editor/GUI/GUIController.cs#L868-L881

This is already bad that there's this code duplication. But does this mean that I also have to add saving user data there too?

ivan-mogilko:
Hrm, I kind of thinking that this SaveGameFiles call I just found is a mistake. Here where it comes from: 6d5e060f031f616e270ff7b6d7b99b1f350b524e

```
 		if (forceRebuild) 
 			_agsEditor.SaveGameFiles(); 
```

I suspect this is done purely to save the changed Workstate, because NeedsRebuildForDebugMode does this:
https://github.com/adventuregamestudio/ags/blob/fc336554a25bb763f721fb1a694a09771ad09b07/Editor/AGS.Editor/AGSEditor.cs#L1169-L1176

In other words, the intent there is to save Workstate (user data), but in order to do so it just saves whole game, which is wrong.

So, the point here, I think, is that WorkspaceState has to be saved anyway, because it serializes last build configuration (regardless of how build has started). But the full game should not be saved in case editor is run with "/compiled", that's completely unnecessary.

I think that because game is loaded before compilation the WorkspaceState will be in correct state, so it should be safe to save again.

On another hand, all this situation also indicates potential question of how things will work in the future when we allow to use standalone tools. We will have to return to this later.

ivan-mogilko:
@morganwillcock I double checked, and WorkspaceState is loaded within LoadGameFromDisk call (if the file is present), which means that it will be in correct state during "/compile" call as well.

ericoporto:
@ivan-mogilko it looks like this is working correctly for the Datafile, Windows (with and without `.ags` file attached), Linux, Web and Android! Did a bunch of checks, cold boots of the Editor and it all appears to work! We should merge this!

PR #2160, opened 2023-10-05
Editor: normalize color transparency for "Leave as-is" import

### LINKED ISSUE #2156: Editor: Import GIF transparency lost when transparent palette index is not zero
**Describe the bug**
Importing a transparent gif  with "Leave as-is" will lose transparency if the transparent color is not index 0.

**AGS Version**
Tested on 3.6.1 and 4

**To Reproduce**
1. Import attached gif and check "Leave as-is"
2. Note that the preview is correct in import window
3. Click import
4. From Sprites pane it can be seen that the sprite has lost transparency

**Expected behavior**
Transparency should be applied no matter at which palette index is set at.

Example gif with transparency at palette index 5:
options_chkbox_sel

AlanDrake:
GIF and 8bit PNG can specify which palette entry is to be used for transparency. 
fix  #2156

This PR normalizes the transparent color while importing 8bit images, where transparency is set to an index different from 0.

It should be possible to backport to master.

### DIFF
--- a/Editor/AGS.Native/agsnative.cpp
+++ b/Editor/AGS.Native/agsnative.cpp
@@ -1732,6 +1732,33 @@ Common::Bitmap *CreateBlockFromBitmap(System::Drawing::Bitmap ^bmp, RGB *imgpal,
 	if (colDepth == 8)
 	{
 		cli::array<System::Drawing::Color> ^bmpPalette = bmp->Palette->Entries;
+    
+    // determine actual transparency index, in case of GIF or other formats that can mark a specific palette entry
+    int transparency_index = 0;
+    for (int i = 0; i < bmpPalette->Length; ++i) {
+      if (bmpPalette[i].A == 0) {
+        transparency_index = i;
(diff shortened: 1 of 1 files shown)


AlanDrake:
I also looked into the palette mangling of 8bit images.
The palette information is being reduced to VGA registers (0-63) because that's what allegro uses in the subsequent conversion to 32bit.

Is there a ready alternative to use the full 0-255 range of an image palette?

ivan-mogilko:
Cherry picked to master branch too : 94346059e946aae803aadc7d9d10f574e4603ce1

AlanDrake:
But maybe double check for 8bit on 8bit game if the palette needs the `bmpPalette[i].R / 4` etc for that specific condition.
It doesn't seem to matter in practical terms in AGS4, and the 0-255 palette PR is almost ready, but I never tested properly on AGS3...

The palette is allegedly used by allegro with `select_palette`, but perhaps it's only effective with higher bit depths so it doesn't really matter.

PR #2162, opened 2023-10-05
fix palettes being altered for 8bit sprites 

AlanDrake:
This change should prevent AGS from mangling palette colors when importing 256 color sprites in hi-color/true-color mode.
The problem was using allegro's palette, which restrict the color to 0-63 values, a legacy from the DOS VGA era.

All the palette logic appears to be crippled by Allegro's 0-63 restriction, so it might be worth rewriting at some point to just use 0-255.
Perhaps it could be a good time when we upgrade the inner color format from 16 to 32 bits.

I added checks to ensure this fix only affects hi-color/true-color, because the implications for 256 color games are unclear.

Caveats:
- Magenta will still be altered to not match the magic pink color.
- The original transparent color will appear black, if one of the corners is used instead (already existing issue, specific to palette index 0)

Note: this PR is based on #2160, because it contains related fixes to transparency for importing sprties

### DIFF
--- a/Editor/AGS.Native/agsnative.cpp
+++ b/Editor/AGS.Native/agsnative.cpp
@@ -1778,28 +1778,15 @@ Common::Bitmap *CreateBlockFromBitmap(System::Drawing::Bitmap ^bmp, RGB *imgpal,
       {
         // BMP files can have an arbitrary palette size, fill any
         // missing colours with black
-			  imgpal[i].r = 1;
-			  imgpal[i].g = 1;
-			  imgpal[i].b = 1;
+			  imgpal[i].r = 0;
+			  imgpal[i].g = 0;
+			  imgpal[i].b = 0;
       }
       else
       {
-			  imgpal[i].r = bmpPalette[i].R / 4;
-			  imgpal[i].g = bmpPalette[i].G / 4;
-			  imgpal[i].b = bmpPalette[i].B / 4;
-
-        if ((needToFixColourDepth) && (i > 0) && 
-            (imgpal[i].r == imgpal[0].r) &&
-            (imgpal[i].g == imgpal[0].g) && 
-            (imgpal[i].b == imgpal[0].b))
-        {
-          // convert any (0,0,0) colours to (1,1,1) since the image
-          // is about to be converted to hi-colour; this will preserve
-          // any transparency
-          imgpal[i].r = (imgpal[0].r < 32) ? (imgpal[0].r + 1) : (imgpal[0].r - 1);
-			    imgpal[i].g = (imgpal[0].g < 32) ? (imgpal[0].g + 1) : (imgpal[0].g - 1);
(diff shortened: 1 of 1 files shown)


ericoporto:
I think this PR can be rebased and doesn't need to be a draft.

AlanDrake:
I'm doubtful, re-reading the code made me realize that the part after `// convert any (0,0,0) colours to (1,1,1) since the image` is completely nonsensical.
What was even the original rationale in shifting all colors +/-1, instead of just changing the specific case for black color? 

I don't think I was thinking straight when I wrote this.

AlanDrake:
Alright I get it.
The point was altering the color so it's not the same as index 0, the transparent color.

I'll make it do a +/- 4 so it makes a difference with the 565 format restrictions.

ericoporto:
Erh, maybe improving the comments

AlanDrake:
Actually there were a couple pitfallf i need to doublecheks

AlanDrake:
Okay, final solution.
Removed that junk code and addressed magic pink cases directly in the other routine for both 16 and 32 bit.

I decided to add green instead, since it's hard to notice some extra green in magenta, and the green channel has 6bits in 16bit format, which makes the alteration less noticeable.

ivan-mogilko:
Is it safe to backport to ags3? I presume this only affects an import process in the editor.

AlanDrake:
I think it should be safe

AlanDrake:
But maybe double check for 8bit on 8bit game if the palette needs the `bmpPalette[i].R / 4` etc for that specific condition.
It doesn't seem to matter in practical terms in AGS4, and the 0-255 palette PR is almost ready, but I never tested properly on AGS3...

The palette is allegedly used by allegro with `select_palette`, but perhaps it's only effective with higher bit depths, which are hijacked by the manual blitting, so it doesn't really matter.

AlanDrake:
Ok, I had overlooked that the palette is indeed used for color remapping, and since I'm hitting some roadblocks with the PR for extending Allegro's palette to 0-255, I fixed it real quick.

I just readded the previous `/ 4` for when the game is 8bit.
```
      if (i >= bmpPalette->Length)
      {
        // BMP files can have an arbitrary palette size, fill any
        // missing colours with black
			  imgpal[i].r = 0;
			  imgpal[i].g = 0;
			  imgpal[i].b = 0;
      }
      else if (thisgame.color_depth == 1)
      {
        // allegro palette is 0-63
        imgpal[i].r = bmpPalette[i].R / 4;
        imgpal[i].g = bmpPalette[i].G / 4;
        imgpal[i].b = bmpPalette[i].B / 4;
      }
      else
      {
			  imgpal[i].r = bmpPalette[i].R;
			  imgpal[i].g = bmpPalette[i].G;
			  imgpal[i].b = bmpPalette[i].B;
      }
```

In theory we could also use a more accurate conversion.
From 8bit to 6bit the ideal algortihm should be `(x * 259 + 33) >> 6` but I guess this may be something for another PR.

PR #2165, opened 2023-10-06
Added #include directives that seem to be required

MatteoPiovanelli-Laser:
Getting the engine and editor to build in windows 10 with vs2022, the IDE wanted to update the SDK to 10 and the platform toolset to 143.
In order for the solutions to build (besides some configuration changes that are probably not retrocompatible), these file changes were required.

If I understand things correctly, they have to do with the fact that some declarations have been moved from `std` to more "specific" headers.
Please, test these out with older toolchains.


ivan-mogilko:
This makes sense, thank you.

PR #2168, opened 2023-10-09
Record inited SDL subsystems and use this info during uninit

ivan-mogilko:
This is done after comments in #2166. Unfortunately I realized the problem too late, after merging the pr.

Keep record of the SDL subsystems we init in a static variable inside sys_main, and use this info when shutting SDL subsystems down. The idea is that this is incapsulated in sys_main, so that external code would not interfere with this.
Somewhere in the future we could remake sys_main into a singleton class, for instance.

EDIT: i have a deja vu, because this looks like something I was trying to add a while ago... but maybe this was in a different program.

@ericoporto could you please also try that this still works as expected?

This should be applied to a 3.6.0 branch too.

### DIFF
--- a/Engine/media/audio/audio.cpp
+++ b/Engine/media/audio/audio.cpp
@@ -684,7 +684,7 @@ void shutdown_sound()
     stop_all_sound_and_music(); // game logic
     audio_core_shutdown(); // audio core system
     soundcache_clear(); // clear cached data
-    if(usetup.audio_enabled) sys_audio_shutdown(); // backend
+    sys_audio_shutdown(); // backend; NOTE: sys_main will know if it's required
     usetup.audio_enabled = false;
 }
 
--- a/Engine/platform/base/sys_main.cpp
+++ b/Engine/platform/base/sys_main.cpp
@@ -22,6 +22,13 @@
 using namespace AGS::Common;
 using namespace AGS::Engine;
 
+// TODO: prehaps refactor the whole sys_main into the singleton class at some point
+struct SysMainInfo
+{
+    // Indicates which subsystems did we initialize
+    uint32_t SDLSubsystems = 0u;
+} static gl_SysMainInfo;
+
 // ----------------------------------------------------------------------------
 // INIT / SHUTDOWN
 // ----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


ivan-mogilko:
Erm, also, why do we init "game controller" in 3.* if the engine has no support for controllers or joysticks in this version? I suppose plugins that implement joystick should init this by themselves?

ericoporto:
> Erm, also, why do we init "game controller" in 3.* if the engine has no support for controllers or joysticks in this version?

I didn't knew why, since the code was like this I just assumed there was some reason you knew. If there's is no need for it in 3.x just remove it! It's necessary on 4.x because of the gamepad API.

ivan-mogilko:
> I didn't knew why, since the code was like this I just assumed there was some reason you knew.

Oh. Maybe this was just in original sonneveld's draft, and I copied without giving it a thought.

> Btw, I think whatever we use at init the subsystems, quitting the subsystems should be the same call just reversed in order.

I probably understand what do you mean, but just in case, you mean the separate Init calls? Because order of flags in one call is inessential.

ericoporto:
> I probably understand what do you mean, but just in case, you mean the separate Init calls? Because order of flags in one call is inessential.

Yeah, separate. Just because if you do something like "init events" and then "init gamecontroller", you end up with events +2, joystick +1, gamecontroller +1, and then doing "quit (events + gamecontroller)", will now set you to events +1, joystick 0, gamecontroller 0. This is just so that the accounting of these are correct.

ericoporto:
Hi, I added a bit of detail above. I don't know the way to go here, if it's an array that gets consumed reversely on quit or some object that wraps the init/quit to account for disposal, or something else.

The flag approach I think may hit the issue I mentioned.

Also, this all may be non-issue because I think SDL_Quit will smash everything in the end, but I think the API is meant to not rely on SDL_Quit - I can't tell what would be the issue though.

ivan-mogilko:
Rewrote records keeping using stack, and enforce use of helper functions to init/quit.

ivan-mogilko:
Sorry, reverted back to previous branch version. I need to think this over again, because I'm unsure about this problem.

ivan-mogilko:
> See: https://github.com/libsdl-org/SDL/blob/6fc17b0e2b421c6f8b996b28f76447ea0956c08e/src/SDL.c#L190-L193
> 
> This means that when the audio opens, we have also a +1 in events, so closing audio and everything else in ONE call, will actually keep a dangling +1 in events. Like...

SDL apparently takes cares of this situation, because it does the reverse action itself on QuitSystem:

https://github.com/libsdl-org/SDL/blob/6fc17b0e2b421c6f8b996b28f76447ea0956c08e/src/SDL.c#L384

ivan-mogilko:
So, I wrote a stack unwinding version here:
https://github.com/ivan-mogilko/ags-refactoring/commit/70f2e5dff92e1f633ea00c61ca757c7bb9599f09

That works, but then I started to doubt if this is really necessary.
The point of this reference counting is mainly to not disable something that is used by a parallel component (like a plugin).
In the worst case, things will not get disabled until SDL_Quit, and some extra components will stay working for a while longer. I do not think that may be an issue.
Docs for SDL_Quit say that you *should* call SDL_Quit anyway:
https://wiki.libsdl.org/SDL2/SDL_Quit

The biggest reason why I dont want to write a more complex code for this is that this case is not about some dynamically allocated and unallocated objects, but something that is inited and uninited once per engine run.

ericoporto:
> wrote a stack unwinding version here

I don't understand that flag that you try to do to avoid already initialized systems, but that looks wrong it would be just the push to the stack as it will increment the ref count if it initializes successfully OR if it's already initialized. That looks almost correct, if you delete that.

> Docs for SDL_Quit say that you should call SDL_Quit anyway

We can just leave at that then. Maybe the quit of subsystems is only relevant for libraries or if you want to do some sort of check after the exit of a system but before ending the application.

ivan-mogilko:
I've been investigating and testing this issue with refcounts that you found, and realized that things may be worse, as there's also an opposite scenario which may lead to subsystems **removed too early**.
I wrote a ticket in SDL2's repo, for I now believe it's a serious error:
https://github.com/libsdl-org/SDL/issues/8381

I think this will not be a problem for us at the moment, because we only have 1 subsystem which may be turned on and off separately (AUDIO), but things may become inconvenient if we have more of these.

Issue #1974, opened 2023-04-08
Some styles not applied/missing in editor modern theme configuration


edmundito:
**Describe the bug**
The modern theme does not seem to be able to apply the following styles to the editor:

* script-editor/text-editor/indent-guide
* script-editor/text-editor/current-statement
* script-editor/text-editor/current-statement-2

Additionally, when a theme is applied, char values in the text editor appear to have a white background and black text, there seems to be no way to set the color values for a char.

**AGS Version**
3.6.0.47

**To Reproduce**
1. Install themes from https://github.com/adventuregamestudio/ags-themes
2. Select [C] VisualStudioDark and restart
3. Open the GlobalScript, note the indent guide
4. Select [M] Dark theme and restart
5. Open the GlobalScript, note the lack of an indent guide
6. Edit the [M] Dark theme and change the styles above to background: blue, foreground: red
7. Restart the editor

**Expected behavior**
The styles are visible

**Screenshots**
Char style missing, also showing lack of indent guide:
![image](https://user-images.githubusercontent.com/168664/230727056-795e1d1c-ef61-4c0a-a214-1869f7fe0c88.png)

edmundito:
Another thing I noticed: When first loading the editor, some theme colors were not applied properly to the panel heading buttons (the ones in red below) Later on they do fix themselves (when I focus on each panel)

![image](https://user-images.githubusercontent.com/168664/230727939-cb3ba591-56ba-413f-87fd-0b022dcf0440.png)

AlanDrake:
The "indent-guide" we're using are the vertical ones. They work but you only see them when indenting twice or more.
"current-statement" is the marker arrow. It works.
"current-statement-2" is a typo, it should be "current-statement2".

What doesn't work is overriding the foreground color of the marker line, but perhaps that should be done differently, because the MarkerSymbol Background is not doing anythign with the foreground color. Or it could be a scintilla bug, or we didn't implement that piece correctly. Needs investigation.

ericoporto:
I think the colors not applying in pin and cross buttons is from here  https://github.com/dockpanelsuite/dockpanelsuite/issues/509

ivan-mogilko:
Char style background was fixed by #2176, I will backport to 3.6.0 branch, as this is a trivial fix.

I did not look into the themes much, could anyone update me on this ticket, what would be required to fix these, and is this applicable to 3.6.0 release at all, or better left for later?

AlanDrake:
[...]

PR #2176, opened 2023-10-12
Editor: fix char style has wrong back color

ericoporto:
this was noticed in #1974


ivan-mogilko:
This fixes char literals resetting to default looks, but do we have a setting for char literals in the themes at all to let test their custom looks?

PR #2178, opened 2023-10-13
Basic support for "#else" preprocessor in AGS script

ivan-mogilko:
I could not find if there was there a ticket for this, maybe I had a false memory about it.

In any case, I got weary of not being able to have ```#ifdef N #else #endif``` kind of construct, and tried to implement basic support  for "#else".

This is a *most trivial and naive* implementation, because it assumes only ifdef/else/endif combination. This won't work with "else if", but these are not supported by AGS preprocessor at the moment.

Added to:
* Editor's preprocessor (C# one);
* Editor's Autocomplete. Although autocomplete seems to break often when preprocessor definitions are around, but that might be a separate issue. At least it seems to work when the `#if/else/endif` are on top of the script (outside of function).
* C++ preprocessor, to sync with the embedded C# one.

If this proves to be working, and merged, then we could also tidy builtin header where declarations depend on SCRIPT_API macros.

PS. I still think we should get a proper complete preprocessor into AGS from somewhere else at some point.

### DIFF
--- a/Compiler/preproc/preprocessor.cpp
+++ b/Compiler/preproc/preprocessor.cpp
@@ -87,6 +87,9 @@ namespace Preprocessor {
                 case ErrorCode::IfWithoutEndIf:
                     cc_error("Missing #endif");
                     break;
+                case ErrorCode::ElseIfWithoutIf:
+                    cc_error("#else has no matching #if");
+                    break;
                 case ErrorCode::MacroDoesNotExist:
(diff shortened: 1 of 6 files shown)


ericoporto:
It would be nice to add a demonstrative test case in https://github.com/adventuregamestudio/ags/blob/master/Compiler/test/preprocessor_test.cpp

ericoporto:
Looks good to merge! :)

PR #2180, opened 2023-10-13
Add autocomplete tests and backport editor tests

### LINKED ISSUE #1543: Task: Write Editor AutoComplete Tests
There are some small issues with the autocomplete. Once Tests are in place though, autocomplete logic should be easier to modify and verify correctness.

In ags4, there's a Test project in the AgsEditor solution. An autocomplete test file would probably be an `AutoCompleteTests.cs` file with a content like:

```C#
using AGS.Types;
using NUnit.Framework;

namespace AGS.Editor
{
    [TestFixture]
    public class AutoCompleteTests
    {
        private static Script CachedAutoCompletedScriptFromCode(string scriptCode)
        {
            Script script = new Script("_test", scriptCode, false);
            AutoComplete.ConstructCache(script);
            return script;
        }

        [SetUp]
        public void SetUp()
        {
        }

        [Test]
        public void CheckAutoCompleteVariables()
        {
            string scriptCode = $@"
int a;
";
            Script scriptToTest = CachedAutoCompletedScriptFromCode(scriptCode);

            Assert.That(scriptToTest.AutoCompleteData.Variables.Count, Is.EqualTo(1));            
        }
    }
}
```

The challenge mostly lies that [`ScriptAutoCompleteData`](https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Types/EditorFeatures/AutoComplete/ScriptAutoCompleteData.cs) contains lists of complex types, so whoever is writing the tests need to test if the generated data contains any element with certain specific attributes, instead of a more direct logic of `element in list`.

There's probably some magic way with LINQ I am not seeing yet that makes writing these tests easy though... 

Also probably ags4 tests in autocomplete which yield ags4 fixes in autocomplete, may have such fixes backported, using git.

ericoporto:
fix #1543

I am backporting from ags4 the Editor tests. I didn't pick everything from there, there are some theme related refactors that I skipped and also all the room controller stuff also was skipped. I am more interested in the preprocessor tests.

I really wanted to add a few autocomplete tests, so here they are!


ivan-mogilko:
@ericoporto I noticed you keep adding the tests, please tell when this is ready to be merged.

ericoporto:
This is ready. :)

I may make a new PR later that adds a few more tests and do a small refactor in the auto complete parser, but for now I won't have time to look into that, so what is here is what I have for now.

ivan-mogilko:
This pr makes unnecessary change to Sprite class reverting its DefaultProperty.
Also, it adds IToXml interface to Sprite class, but no method implementation, is it safe?

ericoporto:
Uhm, let me drop that part of that commit.

ericoporto:
> This pr makes unnecessary change to Sprite class reverting its DefaultProperty.
> Also, it adds IToXml interface to Sprite class, but no method implementation, is it safe?

I dropped changes to the Sprite class in https://github.com/adventuregamestudio/ags/pull/2180/commits/7f20cce7236e3ed4a19586abaed14d564ce7a0bb ! Yes, that was a mistake, thanks for catching it.

PR #2186, opened 2023-10-17
Common: avoid if in indexed bmp unpack

ericoporto:
this is done to squeeze that for to be a little faster when loading something from the spritefile. The change is almost negligible but in the VS Profiler what happens is the fraction of `UnpackIndexedBitmap` is reduced compared to `BitmapToVideoMem` (in gfxdriverbase).

### DIFF
--- a/Common/ac/spritefile.cpp
+++ b/Common/ac/spritefile.cpp
@@ -13,6 +13,7 @@
 //=============================================================================
 #include "ac/spritefile.h"
 #include <algorithm>
+#include <array>
 #include <time.h>
 #include "core/assetmanager.h"
 #include "gfx/bitmap.h"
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Actually, copying to a local pal[256] may not even be necessary, because the palette passed into this function is already a local array of 256:
https://github.com/adventuregamestudio/ags/blob/93c3f7fa510481c04924be82ca9adae4c984dbed/Common/ac/spritefile.cpp#L416

Perhaps you could remake UnpackIndexedBitmap to accept `const std::array<uint32_t, 256>` to enforce that, since this is a internal helper anyway.

ericoporto:
That is a good point, I hadn’t looked up to who was calling it, will adjust to this.

ericoporto:
Done! After this last change, it lowered relevance from `UnpackIndexedBitmap` of contributing 10% to 1.5% to same scene using the VS Profiler!

PR #2190, opened 2023-10-19
Engine: refactoring BitmapToVideoMem

ericoporto:
Hi, tried a bit of moving code around in BitmapToVideoMem. In my tests, this benchmark below that clears fullscreen bitmap and pushes to memory using an Overlay went from 22.6 FPS using what we currently have on master to 26.9 FPS after the modifications from this PR. It's nothing ground breaking, but still somewhat an improvement. Please verify this improvement.

Test Game: [ClearBenchMark.zip](https://github.com/adventuregamestudio/ags/files/13035487/ClearBenchMark.zip)
Project: [ClearBenchmarkProject.zip](https://github.com/adventuregamestudio/ags/files/13035484/ClearBenchmarkProject.zip)

I am not sure about the many commits and how to organize them or if I should merge them in a single one, and also if the code that I got to in the end is acceptable.

Ah, I don't know any 16-bit game, so I didn't test the 16-bit path, it would be nice to get more testing from this branch.

I also noticed that if original source bitmap had a 1 pixel transparent margin, the linear interpolation would not require all checks it has.

### DIFF
--- a/Engine/gfx/gfxdriverbase.cpp
+++ b/Engine/gfx/gfxdriverbase.cpp
@@ -393,10 +393,10 @@ template <typename T> void get_pixel_if_not_transparent(const T *pixel, T *red, 
     ( (((a) & 0xFF) << _vmem_a_shift_32) | (((r) & 0xFF) << _vmem_r_shift_32) | (((g) & 0xFF) << _vmem_g_shift_32) | (((b) & 0xFF) << _vmem_b_shift_32) )
 
 
-template <typename T, bool HasAlpha> void
+template <typename T, bool HasAlpha, bool UsingLinearFiltering> void
 VideoMemoryGraphicsDriver::BitmapToVideoMemImpl(
         const Bitmap *bitmap, const TextureTile *tile,
-        uint8_t *dst_ptr, const int dst_pitch, const bool usingLinearFiltering
+        uint8_t *dst_ptr, const int dst_pitch
 )
 {
     bool lastPixelWasTransparent = false;
@@ -413,7 +413,7 @@ VideoMemoryGraphicsDriver::BitmapToVideoMemImpl(
 
             if (is_color_mask<T>(*srcData))
             {
-                if (!usingLinearFiltering)
+                if (!UsingLinearFiltering)
                     memPtrLong[x] = 0;
                     // set to transparent, but use the colour from the neighbouring
                     // pixel to stop the linear filter doing black outlines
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
I think it's best to start with turning these function into template, without the commit that splits into 3 functions, because otherwise this pr first does one change, and then completely different change overriding first.

ericoporto:
Thanks, I will adjust and also do templates on the opaque variant too. Did you get some speed up with in the test game too? I was more looking into speed up things but it's nice we get shorter code.

The `lastPixelWasTransparent` variable weird approach there, if you happen to know a game or test case that makes evident why that is used, I may look into some alternative way to it.

ericoporto:
Adjusted the PR commits!

> there's also another function nearby called BitmapToVideoMemOpaque (i think), its code may be turned into template too probably.

I also included that too this time!

ericoporto:
Adjusted the PR commits! (sorry for the few push-forces, but I think it's done now!)

> there's also another function nearby called BitmapToVideoMemOpaque (i think), its code may be turned into template too probably.

I also included that too this time!

ivan-mogilko:
The "benchmark" test shows an increase of only about 1 fps for me, from 12.6 to 13.5 (it's rather stable), running with provided config (d3d9, linear filter).
Running with stdscale filter adds ~0.2-0.3 to that.
OpenGL has less fps for some reason, and less difference.

This seem practically negligible improvement, much less then the stats posted in this ticket (22 to 26 fps). This might depend on CPU and other factors maybe.

---

> Ah, I don't know any 16-bit game, so I didn't test the 16-bit path, it would be nice to get more testing from this branch.

Majority of games made before 2010 were 16-bit, they may be found in AGS game database.

---

> The lastPixelWasTransparent variable weird approach there, if you happen to know a game or test case that makes evident why that is used,

I have definitely seen the problems caused by this once, but I cannot remember where, and running random game without this variable does not show any immediate problems. Some investigation may be necessary. Maybe it's not just this variable, but whole `get_pixel_if_not_transparent` section works for this.

In regards to optimizing this, if your aim is to avoid `if` operators where possible, there's this trick of replacing a conditional bool with a factor to use in math, either as a multiplier or a bitwise `&` factor. For example:
```
uint32_t lastPixelTrsFactor = 0;

// lastPixelWasTransparent = true; is replaced with -
lastPixelTrsFactor = 0xFFFFFFFF;

// applying this factor is replaced with -
idst[x - 1] = (idst[x - 1] & ~lastPixelTrsFactor) | ((idst[x] & 0x00FFFFFF) & lastPixelTrsFactor);
```
This means that if factor is set to 0, then the [x-1] pixel keeps full value from itself, and 0 value from [x], and if factor is set to 0xFFFFFFFF, then [x-1] gets 0 from itself and full value from [x].

Unfortunately, it seems that this cannot be applied directly in this code, as you must to test that x > 0 (otherwise this may cause array underflow). Overall, checking for surface bounds result in most conditions there (for the linear filter case), so if that may be restructured somehow maybe it will result in higher speed. But idk if it's worth it.

ericoporto:
The black border issue can be evidenced by making a game in some low resolution (say 320x180) and then having it set to use linear filtering, and run it at a bigger resolution. Then one needs to import assets that are 32-bit, with parts that are fully transparent and parts that aren't fully transparent with some color different than black. Say a color like 0xF000F0FF (pink) now have the fully transparent pixels be 0x00000000. What happens is the linear interpolation from Directx/OpenGL will make the interpolation between all parts (R,G,B,A), meaning R and B in this case will approximate to 0, as the alpha also approaches, resulting in a black border. I will make a project that showcases this issue.

I found it weird you did not get as much improvement as I with what is there already. :/

ivan-mogilko:
I tested 16-bit and 8-bit games, and things seem to work as before. So I will merge this, as this may have a benefit of having a single universal algorithm.

PR #2193, opened 2023-10-20
AGS 4: Remove AlignedStream

ivan-mogilko:
AlignedStream was used initially in attempt to automate reading legacy formats, where structs were dumped from memory into filestream directly, resulting in alignment padding inserted into the file format (according to 32-bit system alignment, where first AGS games were created).

While using AlignedStream made working with old formats easier, this approach also has certain problems:
1. This makes extra padding in file formats implicit and non-obvious.
2. Object Read/Write functions will depend on the type of the passed stream.
3. Adding anything into Read/Write functions will also be affected by AlignedStream, and in theory may lead to additional unnecessary padding inserted simply by keeping same serialization logic (I think this never really happened to the game files, fortunately, but could have. OTOH it might have happened to legacy saves, this is something that needs separate investigation.)
4. AlignedStream does not support Seek, and in general you cannot trivially skip parts of struct by skipping or reading N bytes, you'll have to read it out step by step to keep alignment calculation working.

These concerns made me make the decision to remove AlignedStream, and restore hardcoded padding instead. It's explicit and honest.

This PR does this in ags4, but I suppose same may be done in 3.* branch, for the same reasons, but primarily for clarity of formats. In 3.* it will take additional effort to do, as it's used in more cases, including legacy saves.


ivan-mogilko:
Must double check that I did not miss a case where same function was called with regular and aligned stream on two different occasions. In such case either have 2 separate functions or a function with argument telling to skip alignment padding.

PR #2195, opened 2023-10-21
3.6.1: Remove AlignedStream

ivan-mogilko:
This is a remake of #2193 for the backwards compatible 3.* branch.

AlignedStream was used initially in attempt to automate reading legacy formats, where structs were dumped from memory into filestream directly, resulting in alignment padding inserted into the file format (according to 32-bit system alignment, where first AGS games were created).

While using AlignedStream made working with old formats easier, this approach also has certain problems:
1. This makes extra padding in file formats implicit and non-obvious.
2. Object Read/Write functions will depend on the type of the passed stream.
3. Adding anything into Read/Write functions will also be affected by AlignedStream, and in theory may lead to additional unnecessary padding inserted simply by keeping same serialization logic (I think this never really happened to the game files, fortunately, but could have. OTOH it might have happened to legacy saves, this is something that needs separate investigation.)
4. AlignedStream does not support Seek, and in general you cannot trivially skip parts of struct by skipping or reading N bytes, you'll have to read it out step by step to keep alignment calculation working.

These concerns made me make the decision to remove AlignedStream, and restore hardcoded padding instead. It's explicit and honest.

This PR is currently a draft, 3.* branch contains more cases of padded data in files, and in legacy save format too, so this will take an extra care to do properly, and test corresponding game versions.

- [x] - Games with legacy Interactions & Command lists;
- [x] - Games with ViewStruct272 data;
- [x] - Games with legacy saves (3.2.1 - 3.4.0);
- [x] - Games with saved property values in legacy saves (3.4.0);
- [x] - Games with animated buttons in legacy saves.
- [x] - Games with Overlays in legacy saves.


ericoporto:
I feel that for things used in new save file format and new things this is a good move! Looking at the code changes, I am curious if it would be a good idea for future save formats to have some strategy towards strings, like saving the length of the string, because that manual padding after strings (like in the case of audio clips file name and script name) looks super fragile.

ivan-mogilko:
AlignedStream was never used in new save format.

These paddings also are not related to writing or not writing a length of string, it's all related to how structs are positioned in memory.

ericoporto:
I think I confused with the gamesetupstruct, which I think is the thing that initializes the game. I was trying to understand the audio clip comment about padding to int16 with a single read8.

ivan-mogilko:
This padding depends on existing format which was written in file, that does not depend on which system this runs on.

The rules of padding was mentioned in comments to Alignment stream, but from my memory, the idea is to pad when smaller types are followed by larger types, and the number of written smaller types is **odd**.

example 1:
- int8
- int8
- int16

Here 2 x int8 result in 1 x int16, so no padding is necessary

example 2:
- int8
- int8
- int8
- int16

Here 3 x int8 result in 1.5 x int16, so we have to insert another int8 before int16 is read.

Same happens when int32 appears after int8 and int16. Same would happen with int64, but AGS never had these in formats.

ivan-mogilko:
Mostly done, only testing remained (according to the todo checklist in the ticket).

ericoporto:
I think a test that can be done is to load an old save and save in the current version, and then do the same in this PR version. The later generated save file should match exactly - an md5sum check can be run in both files to verify their contents are equal.

ivan-mogilko:
> I think a test that can be done is to load an old save and save in the current version, and then do the same in this PR version. The later generated save file should match exactly - an md5sum check can be run in both files to verify their contents are equal.

Which purpose would this test achieve? The contents of a save don't depend on what was restored before them, and contemporary saves should not be affect by this PR at all, as they never used AlignedStream.

ivan-mogilko:
Tested everything I could remember, found and fixed 1 typo, but this seems work overall.

PR #2196, opened 2023-10-22
Add mingw basic support

ericoporto:
This is to be able to build on Windows by using MinGW instead of MSVC. Of course, we will forever use MSVC because of the Editor (and C++/CLR), but still, wanted to investigate possibilities - just to safeguard.

Here is a 64-bit build through MinGW: [ags.zip](https://github.com/adventuregamestudio/ags/files/13062121/ags.zip)
_Known issue:_ SDL2 is statically linked and SDL2 statically linked has some issue with fullscreen on Windows I couldn't yet figure it out (this happens in MSVC too, so it's not mingw specific)

There is no rush here, so this can be reviewed with calm.

### DIFF
--- a/Engine/platform/windows/setup/winsetup.cpp
+++ b/Engine/platform/windows/setup/winsetup.cpp
@@ -882,9 +882,9 @@ void WinSetupDialog::OnGfxModeUpdate()
     DWORD_PTR sel = GetCurItemData(_hGfxModeList);
     switch (sel)
     {
-    case kGfxMode_Desktop:
+    case static_cast<DWORD_PTR>(kGfxMode_Desktop):
         _winCfg.FsSetup = WindowSetup(_desktopSize, kWnd_Fullscreen); break;
-    case kGfxMode_GameRes:
+    case static_cast<DWORD_PTR>(kGfxMode_GameRes):
         _winCfg.FsSetup = WindowSetup(_winCfg.GameResolution, kWnd_Fullscreen); break;
     default:
         {
--- a/Engine/util/library_windows.h
+++ b/Engine/util/library_windows.h
@@ -15,6 +15,7 @@
 #define __AGS_EE_UTIL__LIBRARY_WINDOWS_H
 
 #include <utility>
+#include "core/platform.h"
 #include "debug/out.h"
 #include "platform/windows/winapi_exclusive.h"
 #include "util/path.h"
@@ -103,7 +104,7 @@ public:
     {
         if (!_library)
             return nullptr;
-        return GetProcAddress(_library, fn_name.GetCStr());
+        return reinterpret_cast<void *>(GetProcAddress(_library, fn_name.GetCStr()));
     }
 
 private:
--- a/Plugins/agsblend/AGSBlend.cpp
+++ b/Plugins/agsblend/AGSBlend.cpp
@@ -16,14 +16,11 @@
 #define MIN_EDITOR_VERSION 1
 #define MIN_ENGINE_VERSION 3
 
-#if AGS_PLATFORM_OS_WINDOWS
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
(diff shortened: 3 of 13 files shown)


Plugins/agsblend/AGSBlend.cpp
```diff
@@ -42,18 +42,16 @@ typedef unsigned char uint8;
 #define DEFAULT_RGB_B_SHIFT_32  0
 #define DEFAULT_RGB_A_SHIFT_32  24
 
-#if !AGS_PLATFORM_OS_WINDOWS
-#define min(x,y) (((x) < (y)) ? (x) : (y))
-#define max(x,y) (((x) > (y)) ? (x) : (y))
-#endif
+#define math_min(x,y) (((x) < (y)) ? (x) : (y))
+#define math_max(x,y) (((x) > (y)) ? (x) : (y))
```
ivan-mogilko, line 46: It's better to go opposite way and disable macros coming from windows.h. Not only having to rename macros annoying, but also windows.h macros will prevent using std::min/max if one wishes to switch to these.

First I'd double check why windows.h is included here, but likely it's because of allegro.h
Then there are two ways the window header may be dealt with:
1. Undef-ing conflicting macros after the header.
2. Defining `#define NOMINMAX` before the header - in which case windows.h will skip these macro definition.

For example see 
https://github.com/adventuregamestudio/ags/blob/master/Common/platform/windows/windows.h

Engine/platform/windows/setup/winsetup.cpp
```diff
@@ -428,9 +428,9 @@ class WinSetupDialog
 public:
     enum GfxModeSpecial
     {
-        kGfxMode_None    = -1,
-        kGfxMode_Desktop = -2,
-        kGfxMode_GameRes = -3,
+        kGfxMode_None    = 1,
+        kGfxMode_Desktop = 2,
+        kGfxMode_GameRes = 3,
```
ivan-mogilko, line 433: This will break the logic, as iirc these enums are used in a mix with array indexes.
It may be better to cast the type on assignment.
Alternatively, make these constants very large values, and possibly adjust the logic of their usage further in code.

Engine/util/library_windows.h
```diff
@@ -103,7 +104,11 @@ class WindowsLibrary final : public BaseLibrary
     {
         if (!_library)
             return nullptr;
+#if AGS_PLATFORM_WINDOWS_MINGW
+        return reinterpret_cast<void *>(GetProcAddress(_library, fn_name.GetCStr()));
+#else
```
ivan-mogilko, line 109: I think this does not need a platform switch, as this may be a valid type cast regardless.

Engine/platform/windows/setup/winsetup.cpp
ericoporto: Right, the cast was what I was not able to do with the minus value, the compiler was giving me an error, I think the high bits strategy may be easier, so I will possibly may be going that way! Thanks fo catching it!

ericoporto: I think I made a confusion, moving the cast to the switch clause does fix!

Plugins/agsblend/AGSBlend.cpp
ericoporto: I noticed a few things were actually supposed to be built only in non-builtin plugins, so I adjusted that, but also used the NOMINMAX define there and used std::min and std::max instead.

Engine/util/library_windows.h
ericoporto: I removed the check and used the reinterpret_cast as the way to go there.

ericoporto: I wonder if it could just have auto in the method itself to not have any cast and work with all implementations somehow, but I will look into this later - I don't remember if this was already a thing in c++11.

PR #2197, opened 2023-10-23
Build Editor for x64

### LINKED ISSUE #2088: Make the Editor project 64-bit compatible
**Describe the problem**
AGS Editor works nicely as is, but sometimes when using high resolution images one can hit easily some ram limitations (#702), which may cause the editor to crash -  [this sometimes gets a mention in forums](https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/crash-bug-when-importing-sprites-from-an-animated-gif-of-500-frames/msg636656879/#msg636656879). While there may be some possible fixes, it sounds reasonable that the expectation of throwing more memory at a computer would solve a lack of memory problem, but this isn't currently true, as the editor is a 32-bit application.

**Suggested change**
The editor could be 64-bit compatible! Now the windows engine can still be 32-bit at the same time, there isn't anything currently that ties the editor "bitness" to the engine - maybe in the future there could be some way to have both possible build targets but this issue isn't about that. It's not about providing Editor releases for 64-bit too, at least not yet. For now, I think only making it possible to build the Editor for 64-bit in the VS Solution files for the Editor would suffice. At a later time once there isn't technical barriers we can think of how to solve this.

From top of my head, here are a few issues that will be faced making the editor 64-bit compatible

- the References directory doesn't allude any x86/x64 duality, it's just a directory for placing dlls for irrKlang and AGSNative, this makes it harder to alternate between 32-bit and 64-bit builds
- some project files are not tied to any architecture (either x86 or x64) but they may include x86 specifics
- We use a Image Magick NuGet package currently, but we get specific the x86 one and have x86 specific code to work with it
- I don't think that the Editor uses SDL2.dll anywhere, but if it does, we need to remind it's x86 for acwin and not x64.
- probably the solutions/project files should follow the changes made for the Engine dirs in #1870

ericoporto:
Hey, I tried to preserve things as is as possible for the x86 builds of the Editor but some things DID change, please review this with care and evaluate if we update this in 3.6.1 or leave for 3.6.2.

fix #2088

The Editor does successfully build on the x64 target and it does starts up! 

- [x] Verify the x86 target still builds
- [x] Adjust CI for x86 builds
- [x] Adjust installer and packaging scripts for x86 builds
- [x] Test x64 editor and check possible broken features
- ~~check x64 p/invoke and other direct access to windows dll still work or adjust~~ Apparently it works but I don't understand how
- ~~Handle x86 engine `.dll` plugins in x64 Editor (and possibly vice-versa)~~ abandoned here, leaving for the future

CI for x64 builds of the Editor will not be done in this PR.


AlanDrake:
I suggest not to bother trying to bridge x86<->x64 plugins. They should be compiled as separate dlls just like it's done in every other game engine.

Only thing necessary is to make sure x86 builds still work.

Fixing x64 editor issues can be done later, it's more important to have the build option available so we can test it in real scenarios.

ericoporto:
To explain the problem with plugins, it’s that Windows plugins for the Engine ALSO have Editor features, for one it reads the header but also it can do extra things, like showing additional options in the plugin context menu in the editor. I think it needs a Windows directory there where the different archs of the engine can be placed, along with their plugins.

But the core issue is the x64 editor will crash when trying to build with plugins for windows right now.

ericoporto:
this looks like it's working at least in the sense that the x86 Editor appears to work the same (import some gif here to test the Magick isn't broken), and that the project also correctly builds the x64 version of the editor and at least appears to run ok.

It would be nice if more people could play a bit with this version of the Editor project.

ericoporto:
I need to use the AnyCPU package to make it work with both x86 and x64 builds of the Editor, but I can't figure it out how to make it copy the Native dll in place correctly (x86 when building the x86 Editor and x64 when building the x64 Editor), right now no Native dll at all is being copied.

ericoporto:
Found out that this is wrong in 64-bit

https://github.com/adventuregamestudio/ags/blob/b0b63f1e4b6c60ec8b3aaf4c1d4bc0e5cad9658a/Editor/AGS.Editor/Utils/Utilities.cs#L517

Obviously the address can't be an int32, this will crash.

ericoporto:
> Only thing necessary is to make sure x86 builds still work.
> 
> Fixing x64 editor issues can be done later, it's more important to have the build option available so we can test it in real scenarios.

I chose this approach and upgraded the Editor project in a way it does builds for x64 but that it still builds for x86.

Still, I had to remove the AnyCPU and Mixed Platforms as a general target for the Solution (they kinda make sense in fully managed dlls, but not when there is Native dlls involved).

ivan-mogilko:
This is also related: #2274, although I think it may be an issue in case of very large compiled data.

ivan-mogilko:
For the record, this is also an issue related to 64-bit support, perhaps it makes sense to do next: #2274

ericoporto:
Yes, this has to be tackled next before providing builds of the Editor to ensure we don't get weird things when compiling script. I did a `git grep long` in the Compiler dir and doesn't look like a lot of things so I hope to be able to start and finish that tomorrow.

PR #2203, opened 2023-10-28
Theme support for Log Panel

AlanDrake:
This PR adds theme support to the log panel and fixes a couple things:
- added default margins to the RichTextBox
- resized log panel buttons to match icon size

Was there anything new that was missing theming?

### DIFF
--- a/Editor/AGS.Editor/GUI/LogPanel.Designer.cs
+++ b/Editor/AGS.Editor/GUI/LogPanel.Designer.cs
@@ -100,7 +100,7 @@ namespace AGS.Editor
             // 
             // toolStrip
             // 
-            this.toolStrip.ImageScalingSize = new System.Drawing.Size(20, 20);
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
> Was there anything new that was missing theming?

Looks like the LogPanel's Settings pane is not themed. You make it visible by pressing "Show Settings" toolbar button.

AlanDrake:
It's the propertyGrid (already themed), but I noticed I should also ltheme the splitter bar.

ivan-mogilko:
> It's the propertyGrid (already themed)

For some reason "[C] VisualStudioDark" theme does not apply colors to the LogPanel's property grid. Some other themes do apply though. Do you know what may be the difference?

AlanDrake:
> Do you know what may be the difference?

Not all compat themes have global definitions, but I just finished updating themes, so it shouldn't be a problem.

ivan-mogilko:
Recent themes seem to fix this.

Issue #2159, opened 2023-10-05
Link to the license is invalid


ivan-mogilko:
We have a link which is supposed to open Artistic License 2 text in the License.txt, and every code file header. The link is:
http://www.opensource.org/licenses/artistic-license-2.0.php

Recently I tried to open this link and was told that this page cannot be found. I do not know whether it was available earlier, and when did it become invalid.

Searching the opensource.org for Artistic License 2 points to this page instead:
https://opensource.org/license/artistic-2-0/

I guess we should replace the link everywhere through the repository, in the maintained branches (starting with release-3.6.0), in the manual, and on the website, if it also is present there.
(Also see #2131)

ivan-mogilko:
Since there's a need to adjust all the source file headers, in them we have the copyright years written as:
```// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others```

I always wondered if this is permitted to do something like this. Others (major projects at least) seem to just update each year, for example:
https://github.com/libsdl-org/SDL/commit/fde78d12f247a776b52b007479e5274d4bd4e3fe

ivan-mogilko:
I think the best way would be to write another python script in Scripts/ dir and add one that runs through the whole repo and applies this to all files that need and contain this form. Values may be read from version.json as well.

ivan-mogilko:
Fixed by #2212 in multiple branches.

PR #2212, opened 2023-11-07
Update copyright years and license link in the code headers

ivan-mogilko:
For #2159 (but don't close yet, also would need to apply on release-3.6.0 branch)

Added a python script called "push_codeheader.py". It scans all the *.h, *.c and *.cpp files and replaces AGS copyright years and license link, using "version.json" for the reference.

Contains result of running this script for master branch.

Please tell if I forgot any paths, these may be amended in "push_codeheader.py".

### DIFF
--- a/Common/ac/audiocliptype.cpp
+++ b/Common/ac/audiocliptype.cpp
@@ -2,13 +2,13 @@
 //
 // Adventure Game Studio (AGS)
 //
-// Copyright (C) 1999-2011 Chris Jones and 2011-20xx others
+// Copyright (C) 1999-2011 Chris Jones and 2011-2023 various contributors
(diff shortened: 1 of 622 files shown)


ericoporto:
I checked the diffs for the actual files and they look correct. Common and Engine have their own libsrc dirs and apparently this correctly ignore those, so it all seems to work.

ivan-mogilko:
By the way, this is a good moment to decide if following copyright sentence is good:

> Copyright (C) 1999-2011 Chris Jones and 2011-2023 **others**

i mean, who are "others"? Perhaps this could be changed to "AGS team" or "contributors", or something similar.
Our Copyright.txt uses "various contributors".

PR #2214, opened 2023-11-08
Common: use Android asset compatible methods

ericoporto:
fixes Android error reported here https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/ogg-audio-makes-the-game-freeze-android/msg636658856/#msg636658856

> In other news, I seem to have trouble building for Android on this new version. I get the following error every time I try to launch the .apk on my phone :
> "AGSNative
> Loading game failed with error:
> Main game file not found or could not be opened.
> Filename: ac2game.dta..
> 
> The game files may be incomplete, corrupt or from unsupported version of AGS."
> 
> Until I manage to fix that, I won't be able to verify that the issue is fixed in the 3.6.1....

The issue is the previously used methods can't test for file existence inside the Android asset file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ bin/
 gen/
 out/
 
+# ags game in Android single game project used for testing
+Android/mygame/app/src/main/assets
+
 ## end Android related
 
 ## Xcode Build generated
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ out/
 # ags game in Android single game project used for testing
 Android/mygame/app/src/main/assets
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
The fix is logical, and further in code FindFileCI is using DirectoryIterator, which also searches using android's aassetmanager.

What bothers me is that although Android fails this preliminary test, it could continue further and find necessary files using DirectoryIterator, but apparently does not. This means either `AGS_CASE_SENSITIVE_FILESYSTEM` is not defined, or something else is not working well.

PR #2220, opened 2023-11-09
Fix warnings

MatteoPiovanelli-Laser:
As discussed on discord, I cleared a few of the compilation warnings from the Editor.Full solution


Editor/AGS.Editor/Panes/ScintillaWrapper.cs
```diff
@@ -2017,35 +2017,24 @@ private bool ShowAutoCompleteForEnum(string typeName)
         private bool ShouldShowThis(ScriptToken token, List<ScriptDefine> defines)
         {
             Settings gameSettings = Factory.AGSEditor.CurrentGame.Settings;
-            if ((token.IfNDefOnly == "STRICT") && (gameSettings.EnforceObjectBasedScript))
+            // Some settings have become obsolete and now have default values
```
MatteoPiovanelli-Laser, line 2020: Github shows this edit weird. Every 2 `if` conditions on the original file there is now only one, because one ended up being alwasy false.

PR #2221, opened 2023-11-10
Editor: recreate GeneralSettings and DefaultSetup documents when a new game is opened

### LINKED ISSUE #1005: General Settings do not work correctly at first after importing a pre-3.* game project
**Describe the bug**
When you import a pre-3.0 AGS game project into modern editor, the General Settings page does not reflect actual game properties. Instead, it shows properties from the temporary Game object, which is created when the editor launches.
Note that it's not only about displaying values, but it links to wrong data object. Thus, changing any properties won't actually save with the game.
It only works correctly after you close and restart the editor and load already upgraded project.

**AGS Version**
3.5.0.21
but this also happens in AGS 3.4.3.

**Game**
Simple game project made in 2.72: 
[ags-2.72-game.zip](https://github.com/adventuregamestudio/ags/files/3935643/ags-2.72-game.zip)

**To Reproduce**
1. Create a game in a pre-3.0 AGS editor (or load a test project above).
2. Launch AGS 3.5.0 and load that project up.
3. Pass through the import wizard.
4. When it's done, open the General Settings. If you checked the game properties in a AGS 2.72, you'll imediately notice that the settings do not match.
5. Change several properties in General Settings, save and close the project.
6. Reopen project and observe that General Settings now display properties from original project and your changes done as p.5 were lost.
7. Default Setup page has the same behavior.

**Expected behavior**
General Settings and Default Setup should reflect actual game state and changes done to their contents should be saved and compiled along with the rest of the game at all times.

**Additional context**
From what I could find out, General Settings and Default Setup are created early and assigned the dummy Game object before any actual project is loaded. When the 3.+ project is being loaded, it loads up directly into same Game object. However, when you import an old project from 2.* format, a new Game object is allocated, but data references are not replaced in these pages.

To think of it, there could be other effects that I did not yet notice.

ivan-mogilko:
Fix #1005

This makes GeneralSettings and DefaultSetup behave consistently with the other unique persistent panels (such as GlobalVariables, LipSync, etc), and get recreated when the new game is opened. This fixes them in case CurrentGame reference has changed in the editor.

In practice this fixes editor not acting properly after importing a pre-3.* game.

### DIFF
--- a/Editor/AGS.Editor/ApplicationController.cs
+++ b/Editor/AGS.Editor/ApplicationController.cs
@@ -88,10 +88,8 @@ namespace AGS.Editor
             DataFileWriter.TextEncoding = _agsEditor.CurrentGame.TextEncoding;
         }
 
-        private void _events_GamePostLoad()
+        private void _events_GamePostLoad(Game game)
         {
-            Game game = Factory.AGSEditor.CurrentGame;
-
             // TODO: this may be noticably slow especially for sprites. Display some kind of
             // progress window to notify user.
             // Convert absolute paths to relative paths. This is an automatic fixup from when the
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
@Crystal-Shard, if you like to test this, the editor build may be got from here:
https://cirrus-ci.com/task/6176499896680448

Crystal-Shard:
Yes, that works for me.

PR #2223, opened 2023-11-12
Reimplement fade effects outside of GraphicsDriver classes

ivan-mogilko:
This is for #1349 (one of the preliminary steps), but also a proper thing to do.

Targets ags3, but I may redo in ags4 too. In theory it would be nice to have #1349 done in ags3, because that impacts Emscripten performance (according to @ericoporto ), but completing it will take a while. This is just a first step.

If done for ags3, I'd postpone merging this until after 3.6.1 is released anyway, for safety.

In this PR:
* Reimplemented each effect in screen.cpp, with a set of screen_effect_*() functions as transition "interface". Bring them to certain level of consistency in how they render and update frame.
* Removed FadeIn(), FadeOut() and BoxOutEffect() methods from IGraphicsDriver, and respective implementations.
* Removed redundant callbacks from gfx drivers.
* Mark GameState.no_hicolor_fadein flag as deprecated, as it's purely an optimization, and should perhaps be a part of engine config rather than a part of script API.
* Implemented Render(to_texture) and GetCopyOfScreenIntoDDB methods for graphics drivers.

PS. Reimplementing transition effects is done as one big commit, but these functions are purely graphic animation which has no effect on game logic (sans setting `screen_is_faded_out` flag), so I think it's okay to do in a batch.


ericoporto:
This specifically has and additional benefit of reducing the amount of things in the graphics driver - say a new graphic driver is written for Metal, Vulkan, or for an accelerated SDL renderer, this is one less thing to implement.

ivan-mogilko:
While testing this found and fixed several cases in the "render target" feature which was introduced for GUI in 3.6.0.
Turned out it never had fully correct logic, but worked in the existing circumstances.

These fixes has to be applied to 3.6.0 and 3.6.1, regardless of this pr.

ivan-mogilko:
A note to self, following need to be changed:
- for fadeout create screenshot by asking graphics renderer to redraw last frame (using a copy of data they have);
- add new Render method which accepts render target texture as an argument, and draws without calling Present.

ivan-mogilko:
In general, this is done and may be tested (I will convert from draft when 3.6.1 is fixed in a separate release branch).

Had to do some changes to OpenGL and Direct3D classes, necessary for implementing rendering whole game on a any arbitrary texture.

Issue #2226, opened 2023-11-13
Plugin API: proper stream interface and a method for opening data stream


ivan-mogilko:
**Problem**

Engine's Plugin API lacks any means to open game data stream. There are IAGSEngine->FRead and FWrite functions, they require file "handle", an arbitrary int32 value provided by the engine, but in practice these are only usable when saving or reading a game save.

Because of this, if plugin wants to load any custom data, it will have to either use standard FILE* stream, or implement its own reader. This is bothersome, and also not always suitable (some ports may have their own restrictions to how data is packaged or which system API may be used).

**Solution**

The most minimal solution would be to add some kind of IAGSEngine->FOpen function that would resolve a script path with location tokens, open a data stream, and return stream "handle". But after bit of thinking, this may not be enough, because operations in plugin may require few other things such as knowing stream's length and seeking.

Therefore I propose to add a new interface object, with minimal necessary functions for stream use:

```
class IAGSStream {
    virtual int Read(buffer, len);
    virtual int Write(buffer, len);
    virtual int Length();
    virtual int Position();
    virtual void Seek(int offset, int base);
};
```

And two functions in IAGSEngine that open and close it:
```
    IAGSStream* OpenStream(path, mode);
    CloseStream(IAGSStream*);
```

Open function would need to support everything that engine supports internally, that is: opening files on disk, assets in game package (for reading), and stream from any platform-specific packages (e.g. AAssetStream on Android).

**Implementation**

Internally IAGSStream would just wrap our Stream object pointer, hence supporting any stream source.

**Additional notes**

Would need to decide what to do with FRead, FWrite and game saving process in the context of this new interface.
Of course old functions need to keep working for backwards compatibility.

ivan-mogilko:
I am planning to make a draft of this and test with the "spritevideo" plugin I'm currently (struggling) to port to Android.
https://github.com/ivan-mogilko/ags-spritevideo/tree/android

ivan-mogilko:
Resolved by #2227.

PR #2227, opened 2023-11-13
Plugin API: a proper File Stream API

ivan-mogilko:
For #2226.

~This is a draft made primarily for testing these changes on Android.~

~Done in a more primitive way first, without a IAGSStream struct, but added F* functions (FSeek, FLength etc) into IAGSEngine instead. Plugin works with stream using int32 handles. This is a working variant, although an "ugly" one.~

UPDATE: added a IAGSStream interface to plugin API.

Overall changes to plugin API look like this:
```
#define AGSSTREAM_FILE_OPEN         0
#define AGSSTREAM_FILE_CREATE       1
#define AGSSTREAM_FILE_CREATEALWAYS 2

#define AGSSTREAM_MODE_READ  0x1
#define AGSSTREAM_MODE_WRITE 0x2
#define AGSSTREAM_MODE_READWRITE (AGSSTREAM_MODE_READ | AGSSTREAM_MODE_WRITE)

#define AGSSTREAM_SEEK_SET 0
#define AGSSTREAM_SEEK_CUR 1
#define AGSSTREAM_SEEK_END 2

class IAGSStream {
public:
  // Flushes and closes the stream, deallocates the stream object.
  // After calling this the IAGSStream pointer becomes INVALID.
  virtual void   Close() = 0;
  // Returns an optional stream's source description.
  // This may be a file path, or a resource name, or anything of that kind.
  virtual const char *GetPath() = 0;
  // Reads number of bytes into the provided buffer
  virtual size_t Read(void *buffer, size_t len) = 0;
  // Writes number of bytes from the provided buffer
  virtual size_t Write(void *buffer, size_t len) = 0;
  // Returns the total stream's length in bytes
  virtual int64_t GetLength() = 0;
  // Returns stream's position
  virtual int64_t GetPosition() = 0;
  // Tells whether the stream's position is at its end
  virtual bool   EOS() = 0;
  // Seeks to offset from the origin, see AGSSTREAM_SEEK_* constants
  virtual void   Seek(int64_t offset, int origin) = 0;
  // Flushes stream, forcing it to write any buffered data to the
  // underlying device. Note that the effect may depend on implementation.
  virtual void   Flush() = 0;
};
```

```
  // *** BELOW ARE INTERFACE VERSION 28 AND ABOVE ONLY
  // Opens a data stream, resolving a script path.
  // File mode should contain one of the AGSSTREAM_FILE_* values,
  // work mode should contain flag set of the AGSSTREAM_MODE_* values.
  // Returns IAGSStream object, or null on failure.
  // IAGSStream must be disposed by calling its Close() function.
  AGSIFUNC(IAGSStream*) OpenFileStream(const char *script_path, int file_mode, int work_mode);
  // Returns IAGSStream object identified by the given stream handle.
  // This lets to retrieve IAGSStream object from a handle received in a event callback.
  // Returns null if handle is invalid.
  AGSIFUNC(IAGSStream*) GetFileStreamByHandle(int32 fhandle);
```


ericoporto:
> without a IAGSStream

Do we need all iagstream or can we make a smaller part of it, so we have more flexibility? I am just trying to figure what is really needed, because once this is external, it has to be kept in that API - I remember you were thinking about some change to streams.

ivan-mogilko:
Okay, I made a successful test with plugin, and made theora video run on Android through this interface, being streamed by the engine:
https://github.com/ivan-mogilko/ags-spritevideo/tree/android--newfileapi


> Do we need all iagsstream (or at all) or can we make a smaller part of it, so we have more flexibility? I am just trying to figure what is really needed, because once this is external, it has to be kept in that API - I remember you were thinking about some change to streams.
> 
> Edit: now I see the GitHub issue refers to a different IAGSStream and not [IAGSStream.h](https://github.com/adventuregamestudio/ags/blob/master/Common/util/iagsstream.h), so I may be making a confusion.

It's all under question. Years ago I've been thinking to make Stream base class to implement plugin interface directly. That's an option. But existing IAGSStream is way too cumbersome, there's no need for that.
So I'd probably get rid of that, and make a wrapper around Stream ptr for now.

ivan-mogilko:
Added IAGSStream variant of stream API, updated PR description.

ivan-mogilko:
So, to clarify on implementation, I am making our base Stream class to implement an interface (contract) identical to the one exposed in plugin API, because it already matches the requirements. But this does not impose any restrictions on any future changes to these classes. If this becomes inconvenient at any point to pass Stream object to plugins directly, we may as well switch to another class, or write a wrapper over Stream. It is all good so long as the returned object implements exact interface.

ivan-mogilko:
Alright, updated once more. Please see PR's description for API details. In summary, now the IAGSStream closes itself with Close method, which is supposed to deallocate the stream object (so pointer becomes invalid after this). Also added Flush and EOS methods, for sake of completeness.

I had to make a wrapper over our Stream object for this to work.

I'm testing this using my plugin here in this temp branch, where it utilizes this interface to read video:
https://github.com/ivan-mogilko/ags-spritevideo/tree/android--newfileapi3

ivan-mogilko:
...forgot to update plugin version.

ericoporto:
I am a bit lost on one specific API

```
  // Returns IAGSStream object identified by the given stream handle.
  // This lets to retrieve IAGSStream object from a handle received in a event callback.
  // Returns null if handle is invalid.
  AGSIFUNC(IAGSStream*) GetFileStreamByHandle(int32 fhandle);
  ```

What is the use-case here? I am trying to figure what event callback this is meant to work with.

**Comment by ivan-mogilko** [comment]
> What is the use-case here? I am trying to figure what event callback this is meant to work with.

Right now there are 2 event hooks that pass int32 "file handler" into plugins:
AGSE_RESTOREGAME and AGSE_SAVEGAME. These "handlers" are meant to be used with IAGSEngine::FRead and FWrite functions.
I'd like to provide an option to get the IAGSStream for these callbacks, but there's no way to pass a pointer unless I change the callback prototype. So for the time being I added this, as sort of a bridge between old-style file handle to new-style iagsstream object.

In the future we might redesign how the events work, maybe even provide more specific prototypes for each event. But for the time being, this is a working solution: pass a int32 filehandle and let plugin get IAGSStream from it if it needs to.

**Comment by ericoporto** [comment]
That is fine. The new api looks ok API wise when comparing with the rest of the current API.

My minor nitpick would be the following

```
Seek(int64_t offset, int origin) = 0;
```

Origin here is not a position, it's one of the macros that work like enums here - I guess for some C reason. But the macro doesn't mention they are related to origin.

```
#define AGSSTREAM_SEEK_SET 0
#define AGSSTREAM_SEEK_CUR 1
#define AGSSTREAM_SEEK_END 2
```

I would suggest renaming origin to origin_type and mention it in the macro.

```
Seek(int64_t offset, int origin_type) = 0;

#define AGSSTREAM_SEEK_ORIGIN_SET 0
#define AGSSTREAM_SEEK_ORIGIN_CUR 1
#define AGSSTREAM_SEEK_ORIGIN_END 2
```

It's verbose but it makes it easier to understand (I think).

**Comment by ivan-mogilko** [comment]
No, sorry, I do not want to do this. I've been already struggling to keep these macro names short. It's one thing when the macro is used as a global setting that is set once per program (like SDL hints), but this is a common operation.
"Origin" for a "seek" is a well defined term in stream interfaces too, it's used in many libraries (another term is "whence"). Again, this is not something unique to AGS api, so once a person learns how stream seek works, they can do it everywhere.
Here's an example from SDL, for instance:
https://github.com/libsdl-org/SDL/blob/4722269fb62315640cbdbd8916cfc937ee3603d4/include/SDL3/SDL_rwops.h#L70-L76

We may plan to rework plugin API for ags4, put everything in a namespace and turn macros to enums/enum classes (for compilers that support them). That may shorten typing things too.

**Comment by ericoporto** [comment]
The issue is the comment is "Seeks to offset from the origin" and origin being an int, it's hard to understand that origin is a small set of possibilities - and not some other offset.  I would suggest at least rewording the comment to "Seeks to offset from the beginning, current position or end of file".

**Comment by ivan-mogilko** [comment]
I tried this:

```
// Stream seek origins
// Seek from the beginning of a stream (towards positive offset)
#define AGSSTREAM_SEEK_SET 0
// Seek from the current position (towards positive or negative offset)
#define AGSSTREAM_SEEK_CUR 1
// Seek from the end of a stream (towards negative offset)
#define AGSSTREAM_SEEK_END 2

  // Seeks to offset from the origin defined by AGSSTREAM_SEEK_* constants:
  //  * AGSSTREAM_SEEK_SET - seek from the beginning;
  //  * AGSSTREAM_SEEK_CUR - seek from the current position;
  //  * AGSSTREAM_SEEK_END - seek from the end (pass negative offset)
  virtual void   Seek(int64_t offset, int origin) = 0;
````

**Comment by ericoporto** [comment]
Looks good! Thanks, this is easier to understand! :)

**Comment by ivan-mogilko** [comment]
Also made Seek return new position instead of bool, following examples from SDL and .NET framework.

**Comment by ivan-mogilko** [comment]
After thinking this over for a while, I can see only two potential solutions for the problem of "dispose" method, that would allow to avoid using wrappers. This refers to any objects shared between engine and plugin, that may be freely deleted: stream, audio and video processors (decoders, players, filters), and anything of that kind.

These solutions are:
1. Actually have a Dispose() method in the internal object interface in the engine,  but add a comment to never use it directly because it frees the object's memory (especially if they are allocated on stack).
For objects received from plugins, these may be stored in a std::unique_ptr (or shared_ptr) with custom deleter that calls this Dispose() method instead of `delete` operator. This would let to not handle this difference manually.

2. An alternative curious idea is to pass these objects in plugin API in extra small structs, pairing object's pointer and a pointer to deleter function. As an example:
```
template <typename T>
struct DisposableInterface
{
    TInterface *Object;
    void (*Dispose)(T *obj);
};
```
So when the stream, or any other disposable object is passed from plugin to engine, it's done using this struct instead, for example:
```
typedef DisposableInterface<IAGSStream> DisposableStream;

IAGSAudioDecoder *OpenAudioDecoder(DisposableStream *stream);
```

PR #2228, opened 2023-11-14
Android in the Editor support plugins

ericoporto:
Make sure the Android port can package the plugins when building from the editor.

@ivan-mogilko can you test this works with your plugin?

### DIFF
--- a/Android/README.md
+++ b/Android/README.md
@@ -18,6 +18,7 @@ Additionally, there are other directories too:
 
 - `gradle` : to hold shared gradle scripts when needed
 - `misc` : has an example `android.cfg` file 
+- `plugins` : this is an empty directory for placing prebuilt plugins that will be used by the project in `mygame/` dir.
 
 ### Building Requirements
 
--- a/Android/and-build.sh
+++ b/Android/and-build.sh
@@ -22,11 +22,13 @@ function prepare_package {
   cp -r misc package_debug/misc
   cp -r mygame package_debug/mygame
   cp -r gradle package_debug/gradle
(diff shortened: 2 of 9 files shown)


ivan-mogilko:
Please add explanation on how this works, where to put plugins, how they are copied etc?

ericoporto:
In the editor directory, there is the `Android/plugins` folder, you put the plugins that you have built for Android there, under each specific architecture directory - there is a txt file there with a text with a quick reminder of the available architectures currently and how this should work.

Then, assuming you have already placed the dlls for Windows, you open the Editor, right click in the plugins node in the project explorer and select the plugins you want to activate in your game. Now, when you build for Android using the Editor it will additionally look for and grab any plugins and place in your game Compiled/Android/plugins folder and it will warn of any missing plugins (either partially or entirely).

Note that since there are possibly plugins it's ok to miss entirely (like agssteam) It won't error for a missing plugin, but instead only warn about it.

ericoporto:
Added additional notes in the README about it.

ivan-mogilko:
Yes, this works.

> Note that since there are possibly plugins it's ok to miss entirely (like agssteam) It won't error for a missing plugin, but instead only warn about it.

May the same thing be done for Linux, for consistency?

ericoporto:
> May the same thing be done for Linux, for consistency?

Done. I think at some point it would be nice to redo the Linux build script for N archs, but I didn't had a good idea on how to redo, so decided to do the least amount of changes to the code.

ivan-mogilko:
Alright, thank you.

Issue #614, opened 2019-02-07
Editor: support moving objects beyond room background


ghost:
Imagine user likes to position object partially or fully beyond the room's border. At the moment this is only possible by modifying object X and Y property fields. It's curious that characters may be moved freely at the same time.

Consider marking these objects in some way to indicate they won't be visible at room start (or even shade only invisible part if the object is partially in the room).

~Note that #613 is complementary to this, because without free pan it's only possible to drag objects in positive coordinates.~

LATER EDIT: unfortunately #613 was closed without actually implementing unrestricted panning, but only panning within room background. We'd need to have *real* free panning, where a user could pan outside of the room background too. Which would also mean changing scrollbars to some kind of custom control. Perhaps open a new ticket for this, using #613's description and discussion as a reference.
Alternatively: close this ticket and write a "free panning" feature ticket, as technically dragging objects and characters beyond the background will be a direct consequence of such feature.

PR #2229, opened 2023-11-14
Editor: enhance coordinates hint when moving Objects and Characters around the room

### LINKED ISSUE #984: show coordinates when moving object with arrowkeys
when dragging an object around the current coordinates are shown on the viewport. Can we please also have some display of the current coordinates when moving an object (or character I suppose) when moving the object with arrow keys.

special cases are when zoomed in and when the object is not even in the zoomed in viewport.

ivan-mogilko:
Fix #984, addresses a minor problem mentioned in #614.

* Display coordinates hint when Characters or Objects are moved with keys too.
* Fixed hint's position was only constrained by top-left border; made sure it is always visible even if a moved object is out of visible bounds (e.g. if user zoomed into a different part of room).
* Allow to move Objects outside of room bg, similar to Characters.
* Brought Characters and Objects moving code to more consistency (although imo they should get a base class, but that's for separate task).

On a side note, the coordinates may be difficult to see sometimes, depending on background.
For a moment I've been thinking about letting user choose a color in Editor prefs, but even a custom color will still be fixed and have same problem. So this is an open issue.
EDIT: added a black shade under the text.

There is ALOT of code duplication in these two classes, so I opened #2230 as a separate task, don't have spare time for that right now.

### DIFF
--- a/Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs
@@ -33,6 +33,7 @@ namespace AGS.Editor
         private int _mouseOffsetX, _mouseOffsetY;
         private bool _movingCharacterWithKeyboard = false;
         private int _movingKeysDown = 0;
+        private Timer _movingHintTimer = new Timer();
 
         public Character SelectedCharacter { get { return _selectedCharacter; } }
 
@@ -53,6 +54,9 @@ namespace AGS.Editor
                 cmp.OnCharacterIDChanged += OnCharacterIDChanged;
                 cmp.OnCharacterRoomChanged += OnCharacterRoomChanged;
             }
+
+            _movingHintTimer.Interval = 2000;
+            _movingHintTimer.Tick += MovingHintTimer_Tick;
         }
 
         public bool MouseDown(MouseEventArgs e, RoomEditorState state)
@@ -141,7 +145,7 @@ namespace AGS.Editor
         {
             if (_selectedCharacter == null)
             {
-                _movingCharacterWithMouse = false;
+                ClearMovingState();
             }
             else
             {
@@ -152,8 +156,23 @@ namespace AGS.Editor
                     _selectedCharacter.StartY = newY;
                     // NOTE: do not mark room as modified, as characters are not part of room data
                 }
+                _movingHintTimer.Stop();
             }
             return true;
+        }
+
+        private void ClearMovingState()
+        {
+            _movingCharacterWithMouse = false;
+            _movingCharacterWithKeyboard = false;
+            _movingKeysDown = 0;
+            _movingHintTimer.Stop();
+        }
+
+        private void MovingHintTimer_Tick(object sender, EventArgs e)
+        {
+            ClearMovingState();
+            Invalidate();
         }
 
         private void CharCoordMenuEventHandler(object sender, EventArgs e)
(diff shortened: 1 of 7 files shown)


ericoporto:
> the coordinates may be difficult to see sometimes

I think I would collect the code that draws it around like here:

https://github.com/adventuregamestudio/ags/blob/31f6d270379b914314ae4b48cccbc2d7747ae465/Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs#L221

And then make an alternative to System.Drawing graphics DrawString that puts an outline around it - like say a black outline around the yellow text. I looked around and there isn't an easy way to do this, so it would require some amount of fiddling. But if this is acceptable approach l can figure somewhere to add a utility method for this and then PR this change later.

ivan-mogilko:
I just added a black shade under whole text for now.

Editor/AGS.Editor/Panes/Room/RoomEditFilters/CharactersEditorFilter.cs
```diff
@@ -222,6 +222,7 @@ public void Paint(Graphics graphics, RoomEditorState state)
 
                 if (_movingCharacterWithMouse || _movingCharacterWithKeyboard)
                 {
+                    Brush brush = new SolidBrush(Color.FromArgb(200, Color.Black));
```
ericoporto, line 225: Can this be renamed `shadeBrush`? Just because it’s created a bit far from its usage.

ericoporto:
https://github.com/adventuregamestudio/ags/assets/2244442/8b93614c-b4f2-4f9b-8e54-d113dfa55f8a

The PR does what it says. I noticed when tapping keys that the coordinate is only briefly shown, not sure if this is an issue or not.

ivan-mogilko:
Yes, that's a problem, because with mouse user may keep button down and have the hint for as long as necessary, but with keys you cannot do that. So this solution does not work.

Options:
* make coordinates display all the time so long as object is selected;
* make coordinates display after it's been moved by keys until it's moved by mouse or deselected.
* hide hint by a timer (but then users may argue that timeout is not enough)

ericoporto:
From the options above 

> make coordinates display all the time so long as object is selected

It appears this is the simplest rule.

ivan-mogilko:
> Maybe just making a function for the drawing of the coordinates and then when drawing the selection rectangle it would just call such function.

All this is already in the same function.

ericoporto:
I think it's best to merge this PR as is.

I played a bit with the mentioned alternatives and the "hide hint by a timer" is actually the one that felt best but there isn't a way to make the code for this super short, so this is best left to test it out later, probably after #2230 .

ivan-mogilko:
I think i will ask on forums for more opinions.

From the code's perspective the easiest is to draw coordinates always. But I don't know if people will be okay with that.

The way it is right now won't let to find coordinates comfortably enough with key moving, which will defeat the purpose of pr.

ericoporto:
if the zoom is one that makes the room really small, the coordinates occupy all of the room, and it's not possible to see what is selected, but by using the timer approach it's still useable. By showing when the object is selected, it becomes weird in this specific edge case, but the timer approach works alright.

ivan-mogilko:
Okay, I added the timer to remove the hint. It ended up with pretty small amount of code.

ericoporto:
I tested and the timer works great!!! This is a nice addition! Nudging with the arrow keys works nicely now, no more flickering when tapping the keys. :)

Issue #2231, opened 2023-11-14
Editor: Pasting into the Dialog's option text will paste into the dialog script instead


ivan-mogilko:
**Problem**

If you have a Dialog open, and paste a text from clipboard into the option's text, it will instead paste into the dialog script, as if it had a focus or a command precedence.

**AGS Version**
3.6.1, does not reproduce in earlier version.

I have suspicion that this might be related to picking out ScriptEditor's base class.

**To reproduce**
1. Open a dialog for editing, create few options.
2. First click onto the script to set a cursor there.
3. Copy any text to clipboard.
4. Set cursor into one of the option fields and paste (ctrl + v).
5. Notice how the text pastes into the script instead, at the position where the latest cursor position was.

**Expected behavior**
Nothing is pasted into script unless it has a focus.

ericoporto:
Now that you mentioned, indeed in the previous dialog editor there wasn't a `PASTE_COMMAND`, by looking into the diff in the PR #1743

ivan-mogilko:
> Now that you mentioned, indeed in the previous dialog editor there wasn't a PASTE_COMMAND, by looking into the diff in the PR https://github.com/adventuregamestudio/ags/pull/1743

I'm not sure what is the deal with the old code, but I tested a number of older versions (starting with 3.2.1) and they have Copy/Cut/Paste in context menu, and Ctrl+C/Ctrl+V combinations work.

ericoporto:
OK, the issue is the edit menu that we add has commands with shortcut keys that match the default ones (e.g. Ctrl+C for copy) and so on. The menus are changed through a weird way by the `MainMenuManager.cs` under `ShowPane` method. They are define in the ScriptControlBase and then they are passed to the matching component.

So, when you do the shortcut it will call `OnClickCommand` passing the specific string, not sure how that works, and unfortunately when it does so, it doesn't pass what object the command was done, as there isn't one, as this is just calling the method on top of the edit.

Given this, I had the following not bright idea to add on top of `OnClickComman` call...
```C#
            if(!_scintilla.Focused)
            {
                Control c = Utilities.GetControlThatHasFocus();
                if (command == CUT_COMMAND) (c as TextBox)?.Cut();
                else if (command == COPY_COMMAND) (c as TextBox)?.Copy();
                else if (command == PASTE_COMMAND) (c as TextBox)?.Paste();
                else if (command == UNDO_COMMAND) (c as TextBox)?.Undo();

                return;
            }
```

[...]

PR #2232, opened 2023-11-15
Editor: if textbox is focused, edit menu affects it

ericoporto:
additional for #2231 

There are still small details there, but in this way, using an action in the edit menu won't affect the scintilla editor if a textbox has focus.

Checking for focus in scintilla instead won't work, so this is why it's done in such way.

### DIFF
--- a/Editor/AGS.Editor/Panes/DialogEditor.cs
+++ b/Editor/AGS.Editor/Panes/DialogEditor.cs
@@ -62,6 +62,7 @@ namespace AGS.Editor
                 if (textBox != null)
                 {
                     textBox.GotFocus += dialogOptionsEditorTextBox_GotFocus;
+                    textBox.MouseUp += dialogOptionsEditorTextBox_MouseUp;
                 }
             }
         }
@@ -74,6 +75,7 @@ namespace AGS.Editor
                 if (textBox != null)
                 {
                     textBox.GotFocus -= dialogOptionsEditorTextBox_GotFocus;
+                    textBox.MouseUp -= dialogOptionsEditorTextBox_MouseUp;
                 }
             }
         }
@@ -238,13 +240,23 @@ namespace AGS.Editor
             UpdateUICommands(force:true);
         }
 
-        private void dialogOptionsEditorTextBox_GotFocus(object sender, EventArgs e)
+        private void dialogOptionsEditorTextBox_Event(object sender, EventArgs e)
         {
             TextBox tbox = sender as TextBox;
             if (tbox == null) return;
 
             bool can_copy_cut = tbox.SelectionLength > 0;
-            EnableStandardEditCommands(copy: can_copy_cut, cut: can_copy_cut, paste: true, undo: tbox.CanUndo, redo:false);
+            EnableStandardEditCommands(copy: can_copy_cut, cut: can_copy_cut, paste: true, undo: tbox.CanUndo, redo: false);
+        }
+
+        private void dialogOptionsEditorTextBox_MouseUp(object sender, EventArgs e)
+        {
+            dialogOptionsEditorTextBox_Event(sender, e);
+        }
+
+        private void dialogOptionsEditorTextBox_GotFocus(object sender, EventArgs e)
+        {
+            dialogOptionsEditorTextBox_Event(sender, e);
         }
 
         protected override void OnCommandClick(string command)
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
But this is not enough for properly applying the action over the textbox. Currently for scintilla it also checks for availability. For example: whether it is possible to cut or copy, whether it is possible to undo or redo, and enables and disables the commands accordingly. With this change the commands will act inconsistently.

ericoporto:
> applied on controls that may not belong to the panel

So this can't happen because if a different pane is focused the AGS Editor will readjust the menus to that pane, so while that focus query can get focus from any control in the editor, it's only called when something in the pane is already in focus.

> Make the commands enabled only when Scintilla is active.

I made an attempt at something like this, if scintilla is gets focus, the edit menu gets updated.

> If you like them to apply to the Dialog options too, then this should be handled inside DialogEditor (as it's the class that knows that it has other input fields), and make sure that it affects only its children

Took a stab at that.

> For the real proper behavior it would also need to check which control is under focus right now, and change the command states according to the active control and its state.

OK, I thought that was a bit too much, and instead just enable everything if in a textbox.

ivan-mogilko:
> > applied on controls that may not belong to the panel

> So this can't happen because if a different pane is focused the AGS Editor will readjust the menus to that pane, so while that focus query can get focus from any control in the editor, it's only called when something in the pane is already in focus.

This can happen (and happens!) if a text box is not on a ContentDocument control. For instance, if it's a Properties Grid.

ivan-mogilko:
I still do not understand why the textbox handling has to be inside ScriptEditorBase class: that is a parent class for a script editor, which is meant to handle only Scintilla control. It does not know that there may be any textboxes.
If anything, then DialogEditor should be handling its own textboxes.

But, the real problem is that this PRs goal is not well defined imo, and it's not working consistently.
First of all, why there's a need to make these Edit menu options work on the option text boxes at all?
Secondly, what about the other options, below the Copy/Paste etc? They are still active in this menu when text box is focused, but they do not work when you click on them. I'm speaking about "Find", "Goto Line", "Show autocomplete", and others.

I suppose it would be simpler to disable this menu completely when the focus leaves scintilla.

ivan-mogilko:
I still think that the textbox handling should not be done inside ScriptEditorBase class: that is a parent class for a script editor, which is meant to handle only Scintilla control. It does not know that there may be any textboxes or anything else.
If anything, then DialogEditor should be handling its own textboxes.

It is possible to override OnCommandClick() in DialogEditor, and intercept some commands. Possibly these command constants in ScriptEditorBase have to be turned from `private` into `protected` to let the derived classes know about those.
Then you could handle these differently in DialogEditor.

UpdateUICommands() may also be overridden as necessary, instead of passing a slightly strange "force" parameter into the base class.

Besides the above, there's an issue with the rest of the menu: commands such as "Find", "Goto Line", "Show autocomplete", and others. They are kept enabled while the focus is in the textbox, but do not do anything.
IMO some of those, like "Goto Line" and "Find", may keep working even if focus is in the textbox, since it's obvious they are about the script.
If OnCommandClick() UpdateUICommands() were handled in DialogEditor class, then it could accurately determine which commands to disable and which to keep, which should have special meaning for textboxes, and which to delegate to the base class.

ericoporto:
I moved the textbox things out of the ScriptEditorBase and into the DialogEditor and added additional code in the dialog editor to make sure the Edit menu updates correctly when the cursor is in either the scintilla or in a textbox. I am not dealing with the other things it could be because there isn't really a good way to do so. Using something that is not cut/copy/paste/undo/redo will force focus into Scintilla, even if it's not in scintilla before.

ivan-mogilko:
I think this might be fine, but could you add a UpdateUICommands after DialogEditor does OnCommandClick, similar to how it's done in the base class? This lets to enable Undo command, otherwise it's stuck in disabled state unless you switch to another text box and return back.

ericoporto:
OK, added also an extra KeyUp event there because I noticed it picked almost everything except selecting text with the keyboard arrows + shift! Now it's really checking all possible situations!

ivan-mogilko:
Unfortunately I missed this, but `force` argument in UpdateUICommands has to be redone somehow. This is a virtual method, which means that it may be overidden in subclasses, but "force" only has a meaning for limited number of commands in the base class.

Issue #1926, opened 2023-02-17
Editor: substantial performance issue in a script editor


ivan-mogilko:
**Problem**
There seem to be a serious performance issue in the script editor, which becomes easily noticeable when editing large scripts, and when typing a *long sequence of characters*. In other words, if you open any big script, such as Tween module, or any game script with 1000+ lines that you have, and start typing random garbage non-stop.

In my case the CPU load easily goes up to 30%.

Under more close inspection, proper script syntax also causes this, except that may be harder to notice, because of the shorter lengths of the words.

This happens in 3.6.0 and ags4, but also in previous versions (tried 3.5.0, 3.5.1), which means that this is an old issue, which was not recorded in our issue tracker.

**AGS Version**
3.5.0, 3.5.1, 3.6.0, ags4; maybe earlier too.

**To Reproduce**
1. Create a game or open existing one in the Editor.
2. Make sure that you have a big script with real contents. For example, download a [Tween module](https://github.com/edmundito/ags-tween/releases) and add it to the game project.
3. Open that script for editing.
4. Put a cursor in any random place and press and hold some letter key.
5. In a few moments you will notice how slow the typing became.
6. If you run any system performance monitor (e.g. a Windows Task Manager) you shall also notice a significant CPU load while you type.

**Expected behavior**
Typing random words should not cause such a performance hit.

Need to investigate this under profiler and find out a cause of the slowdown.

morganwillcock:
If you revert commit 9979c304a14b43a855c601f30ca3cbdedf9aab1e does it help?

morganwillcock:
It looks like it is the construction of the autocomplete cache. The majority of the CPU time is used for external string operations. I couldn't find an easy way to export a profile report as readable text, but here is a screenshot:
![autocomplete_cpu](https://user-images.githubusercontent.com/14958747/219983706-e74efdb5-7389-4cc4-a7d3-0f8b5d99e045.png)

ivan-mogilko:
There's something strange I noticed about ConstructCache in the past, while updating Scintilla to the new version. Judging by the existing code, it was planned to run ConstructCache periodically in the background thread. But, if I recall right, it's never run in background, but always in sync with main thread, when user types, etc.

ericoporto:
[...]

PR #2233, opened 2023-11-19
Editor: slightly perf improve in FastString StartsWith

ericoporto:
I noticed the StartsWith comparison is using the default `CurrentCulture`, in the `FastString` implementation, which from the [docs here](https://learn.microsoft.com/en-us/dotnet/api/system.string.compare?view=netframework-4.6.2) I think is more meant for regular text, but for variables and other elements in AGS Script, I think we expect binary matches. `FastString` is used only in a context of processing AGS Script language, so I think it's safe to change it's behavior to such.

The motivation is I noticed this makes StartsWith to run a little faster. To profile I used the description of #1926, and just loaded my ImGi script module and typed a lot of things in it's script module to make the ConstructCache run a lot.

I am not super sure of how much this improves because for some reason I had to build the Editor for Debug to actually find the FastString.StartsWith calls in the ones listed.

### DIFF
--- a/Editor/AGS.CScript.Compiler/Entities/FastString.cs
+++ b/Editor/AGS.CScript.Compiler/Entities/FastString.cs
@@ -50,7 +50,7 @@ namespace AGS.CScript.Compiler
 			{
 				return false;
 			}
-            return (_data[_offset] == text[0]) && (_data.Substring(_offset).StartsWith(text));
+            return (_data[_offset] == text[0]) && (_data.Substring(_offset).StartsWith(text, StringComparison.Ordinal));
         }
 
         public int IndexOf(string text, int offset)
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Question: is it desired that FastString worked with `StringComparison.Ordinal` in general, or only in this particular method?

I'm asking because there's also a call IndexOf below your changes, and that also accepts "StringComparison" as an argument (at least when the input is a string). I suppose that this class should do all comparisons alike.

ericoporto:
First, the perf only shows StartsWith on top ([as seen in this comment](https://github.com/adventuregamestudio/ags/issues/1926#issuecomment-1436141014)), which is why I looked into it.

After, StartsWith doesn't even show up on top - it's like this

image

> Question: is it desired that FastString worked with StringComparison.Ordinal in general, or only in this particular method?

Good question, I think it's alright to make it like this in general - for consistency.

PR #2236, opened 2023-11-20
CI: run auto-test in GitHub Actions pipeline

ericoporto:
This is does an additional run of the Engine in the GitHub Actions CI running auto-test

The project is here: 

https://github.com/adventuregamestudio/ags-test-games

Example run:

https://github.com/adventuregamestudio/ags/actions/runs/7340091125/job/19985502228?pr=2236#step:7:63

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,11 +17,11 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows,         os: windows-latest, artifact: 'ags-windows', shell: cmd }
-        - { name: Ubuntu,          os: ubuntu-latest,  artifact: 'ags-linux',   shell: sh  }
-        - { name: macOS,           os: macos-latest,   artifact: 'ags-macos',   shell: sh  }
+        - { name: Windows,    os: windows-latest, artifact: 'ags-windows', shell: cmd, runbin: ".\\ags.exe", bindir: "Release" }
+        - { name: Ubuntu,     os: ubuntu-latest,  artifact: 'ags-linux',   shell: sh,  runbin: "./ags",      bindir: ""        }
+        - { name: macOS,      os: macos-latest,   artifact: 'ags-macos',   shell: sh,  runbin: "./ags",      bindir: ""        }
 
-    steps:    
+    steps:
     - uses: actions/checkout@v4
 
     - name: Configure CMake
@@ -44,3 +44,11 @@ jobs:
           ${{github.workspace}}/build/${{env.BUILD_TYPE}}/ags.exe
           ${{github.workspace}}/build/ags
           ${{github.workspace}}/build/ags.exe
+            
+    - name: Run game auto-test  # we are using tap.py so that the pipeline fails if a test fails
+      working-directory: ${{github.workspace}}/build/${{matrix.platform.bindir}}
+      run: |
+        pip install tap.py
+        curl -sLo auto-test.ags https://github.com/adventuregamestudio/ags-test-games/releases/latest/download/auto-test.ags
+        ${{matrix.platform.runbin}} --no-message-box --log-stdout=script:info,main:info --user-data-dir . auto-test.ags        
+        tappy agstest.tap
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,22 +9,28 @@ jobs:
   build:
     name: ${{ matrix.platform.name }}
     runs-on: ${{ matrix.platform.os }}
+    defaults:
+      run:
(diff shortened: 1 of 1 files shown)


ericoporto:
I have been trying to build the editor in the GitHub Actions CI, [but I am failing at being able to figure it out](https://github.com/ericoporto/ags/actions/runs/6997546824/job/19034637583#step:10:1)...

So far I think the way to go is a separate repo so the test game can be built with a released ags version - probably the upcoming 3.6.1.13 or later, with the `/compile` issues fixed. Then this pre-built can be released and we can fetch the latest release to here.

ericoporto:
Minor reminder that this is a possible "run ags game as test in CI" PR, assuming there is an external game to be run, but it assumes such game is maintained in a separate repository that would work best being in ags org in github.

ivan-mogilko:
So you need me to create a "test game" repository in ags org?
I could create an empty repository for this purpose, and give you full contributor rights there.

ericoporto:
Do you think this is a good approach? My idea would be to create some game that does things fast that could as well as be running headless because the idea would be to use it's output as a test that things in the engine are kept working.

My idea to do it in a separate repository is because I think it would be easier to play with it and evolve it's repository and then it could later be fed back in the ags repo to be used here too, to check the engine is still running things ok.

ivan-mogilko:
I am not sure what is a question here? I do think that we may have a repository in ags org, that could contain multiple test games. Having that repository in ags org would mean that it's close to the engine, and that it may have multiple contributors. That does not prevent making PRs from personal repositories either.

I am not sure what does "feed back in the ags repo" mean in this context, does this mean put a game into ags organization, or main ags repository? If latter, then I don't think it's a good idea, because that repository is already hosting alot of stuff, and usually it's a proper thing to have separate repositories for external utilities.

On a side note, I also do not know what "running headless" means...

ericoporto:
oh, right, headless as in the game is meant to be run without needing a monitor. This doesn't mean the test game doesn't put anything on screen, but if it does, it's nice to screenshot (in ags script) and save the image as means to test if they are correct (we could have our own "golden" images, something pre-generated stored in the repository meant to be compared to see if images are equal.

And by feedback in ags repo, I mean, the built `game.ags` could be downloaded by the current ags ci so it runs it also in the current pipeline, so we can check things are ok (like in this PR).

ivan-mogilko:
Oh, I see, I completely missed the point. Now I read the commits more carefully.

So, the "approach" is to have a compiled game *.ags somewhere, like attached to another repo's "release"; and then engine's repo has a action that temporarily downloads this *.ags to an environment with compiled engine binary, and runs, and checks the output. Is this correct?

I think this is curious, and would be enough for runtime testing.
For a full compilation + runtime testing we'd have to finish the standalone compilation tools. Or recompile only scripts and have them as separate assets on "disk". Or compile using "AGSEditor.exe /compile" (if that's doable)

ivan-mogilko:
Created a repository, and given you write access to it:
https://github.com/adventuregamestudio/ags-test-games

ericoporto:
Thanks! I started work on it, trying to think now on how releasing there will work - thinking about if just making releases with numbers that somewhat match main repo, like 3.6.1.14.1, and also how to deal with ags4, like, should it be a different branch or just keep ags4 and ags3 folders.

ericoporto:
Erh, my simple tests already found something! 

https://adventuregamestudio.github.io/ags-manual/Set.html#setgetitemsasarray
https://adventuregamestudio.github.io/ags-manual/Dictionary.html#dictionarygetkeysasarray

In both Set and Dictionaries, it's written in the manual that `GetItemsAsArray` should "Creates a dynamic array filled with keys in the same order as they are stored in the Dictionary.", and gives an example.

I thought this "order as they are stored" meant the order in which `.Set("key","value")` was used, but just to confirm, it's actually just whatever the internal memory maintains, correct? I noticed the order is different between Windows and Linux/macOS. I mention because I noticed the tests failed, so I think I need to make tests that don't care for the order.

ivan-mogilko:
> I thought this "order as they are stored" meant the order in which .Set("key","value") was used, but just to confirm, it's actually just whatever the internal memory maintains, correct?

No, the order is:
- for sorted container it's the sorted order (alphabetical)
- for non-sorted container it's undefined (the hash order).

ivan-mogilko:
I'd like to double check, is this PR for testing CI only, or it is intended to be merged?

ericoporto:
The idea here is to have it merged, and then the auto-test game built in ags-test-games would always be ran in the GitHub Actions here. Notice that because the building of the Engine, Editor and Releases today are done in Cirrus, the changes here doesn 't affect any of this.

PR #2237, opened 2023-11-20
Script API: add Game.PrecacheSprite() and PrecacheView()

ivan-mogilko:
This is the last thing i wanted to try adding in 3.6.1.

It's been always a problem in AGS that game developers have little control over the resource management. That is a big topic, and has to be thought through well before writing a good API, but I wanted to add couple of simple temporary methods, acting as a  replacement for existing hacks that let precache a number of sprites and fix performance in certain scenes.

* Game.PrecacheSprite(int sprnum) precaches a single sprite;
* Game.PrecacheView(int view, int first_loop, int last_loop) precaches sprites and *linked sounds* for all frames in the given range of loops for the view.

Precaching a sprite in both of these methods does not load only a raw sprite, but also prepares a texture for it, since we have a texture cache now too, and converting sprites to textures may also have an impact on performance.

An important note is that these functions work according to the existing cache rules. That is - cached assets are normally counted towards the cache limit, and if the limit is exceeded, then older objects will be freed. In addition, the sounds are only cached if their individual size does not exceed "stream_threshold" value from ags config.

This means that in the end the result of using these functions also depends on active game config, and this has to be honestly stated.

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -230,7 +230,7 @@ void SpriteCache::DisposeAllCached()
     ResourceCache::DisposeFreeItems();
 }
 
-void SpriteCache::Precache(sprkey_t index)
+void SpriteCache::PrecacheSprite(sprkey_t index)
 {
     assert(index >= 0); // out of positive range indexes are valid to fail
     if (index < 0 || (size_t)index >= _spriteData.size())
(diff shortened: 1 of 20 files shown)


ericoporto:
I can't see `LockSprite` get called, I thought DynamicSprites had to be locked. Additionally, I also thought that the mouse cursor **had** to be locked - as in it would crash if not, there is/was a comment in the code about this.

It seems, from the comments, this changes the rules so that `LockedSprites` don't count to the cache limit. What I am concerned is in the Windows build where ram is limited to 2GB, so I guess the cache is more limited, can this impact in some way?

ivan-mogilko:
> I can't see LockSprite get called, I thought DynamicSprites had to be locked. 

It's locked by passing a lock flag to SetSprite.

> Additionally, I also thought that the mouse cursor had to be locked - as in it would crash if not, there is/was a comment in the code about this.

I removed the code that depended on that in commit named "cleaned up mouse cursor caching", it has bit more info in extended description.

> It seems, from the comments, this changes the rules so that LockedSprites don't count to the cache limit. What I am concerned is in the Windows build where ram is limited to 2GB, so I guess the cache is more limited, can this impact in some way?

It's been always like this historically, until I changed it when implementing texture cache. Now I am returning this back.
Prior to this only player's view and mouse cursors were locked (but never unlocked). After this PR even them are not locked. Actually nothing is locked now, unless someone reintroduces locking of particular views.

ivan-mogilko:
But I will think more about this. Maybe it's more correct to count locked regular sprites along with the cached ones, if we consider "max sprite cache" setting to define the total allowed size of regular sprites in memory.

ericoporto:
Well if I understood, currently regular sprites don't get locked simply because the calls that did so got removed.

Ah, it is necessary to test the Editor too just in case this changes anything there - I think not, but would just check.

ivan-mogilko:
I removed the change to the locked items calculation, but added a comment about this to ResourceCache class.

ivan-mogilko:
Fixed a bug in PrecacheView (forgot to convert from 1-based view index to 0-based).

Added diagnostic logging.

ivan-mogilko:
Alright, I guess I will merge this, because this formally works and may even be useful under some circumstances.

PR #2238, opened 2023-11-21
Editor: compile flag turns alert in log messages

### LINKED ISSUE #666: AGSEditor /compile should turn alert into log messages

ericoporto:
fix #666

The goal here is so that using `/compile` flag the Editor won't show a pop-up or require human input in some way it could hang a continuous integration pipeline - because no one would be there to click the button.

- `/compile` will make the Editor alert messages get logged to the terminal
- Additionally, messages that goes to the output panel also gets logged to the terminal
- The software exit code will be non-zero if the compilation failed when using `/compile`
  - This can be used to fail a pipeline in case we have a build error in a game

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -376,6 +376,7 @@
     <Compile Include="Utils\ConcurrentCircularBuffer.cs" />
     <Compile Include="Utils\MathExtra.cs" />
     <Compile Include="Utils\ScriptGeneration.cs" />
+    <Compile Include="Utils\StdConsoleWriter.cs" />
     <Compile Include="Utils\Validation.cs" />
     <EmbeddedResource Include="GUI\frmMain.resx">
       <SubType>Designer</SubType>
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -373,6 +373,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Utils\AndroidUtilities.cs" />
     <Compile Include="Utils\ArgumentBuilder.cs" />
+    <Compile Include="Utils\CommandLineOptions.cs" />
     <Compile Include="Utils\ConcurrentCircularBuffer.cs" />
     <Compile Include="Utils\MathExtra.cs" />
     <Compile Include="Utils\ScriptGeneration.cs" />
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Could you please add an explanation to this PR, saying what is an expected behavior after this change?

ericoporto:
Right, I had simply removed the beta warning when running from the command line, but other message boxes are indeed logged instead. I didn't pick the output that normally goes to the output panel and logged - it would be nice but it's more complicated.

ericoporto:
I went ahead and changed so the beta message is still printed, messages in the output panel get printed and build error return non-zero exit code.

ivan-mogilko:
I'm sorry, but I still do not know what is expected from this change, and how to test this.

This is an example of what I see when I run with the /compiled flag from the windows console:
editor-console

Notice two things:
1. For some reason the messages are not printed normally, but appended to the current directory line, practically into the console input (I do not know how to express this).
2. The "This is BETA" warning is printed not as the first, but as the last message for some reason.

Is this an intended result?

ericoporto:
> trying to reroute output to a file does not work (file remains empty)

this is because by attaching to console it won't be redirected to the file. Luckily, there is [`Console.IsOutputRedirected`](https://learn.microsoft.com/en-us/dotnet/api/system.console.isoutputredirected?view=net-8.0&redirectedfrom=MSDN#System_Console_IsOutputRedirected), which I can use to detect and not attach in this case. This is easy to fix, thanks!

About 2., simply it happens later in the code, what I can do is figure some way to pass the compileAndExit flag forward so I can run the compile and exit outside of the function that process the command line parameters.

About 1., I am not sure because I don't observe this on the new Windows Terminal which I am using, I will see if I can find how to run the old terminal.

ericoporto:
@ivan-mogilko  I think I figured all the mentioned issues, can you test it again?

I think it's good enough we could use this to build a test game (as mentioned in #2236) and then we could provide releases of this test game we could use to test ags itself - to reduce catching up errors after release.

ivan-mogilko:
Hmm, the latest variant works better. Running directly from console still has this problem where messages are visually inserted into command line for some reason. But running same command from a batch file, or redirecting output into a text file produces correct result. So this is usable.

PR #2249, opened 2023-12-05
Remove deprecated inheritance from std::x_function

### LINKED ISSUE #2247: remove use of `std::binary_function` from Common
Common uses `std::binary_function` which was deprecated in C++11 and removed in C++17. If you want to build the engine using C++17 or newer for experimental reasons, checking compiler behavior or something else, the only error you will meet is where `std::binary_function` is used.

The following patch can keep the `std::binary_function` behavior if desired.

```diff
diff --git a/Common/core/assetmanager.h b/Common/core/assetmanager.h
index 73f85f17f..e52a21601 100644
--- a/Common/core/assetmanager.h
+++ b/Common/core/assetmanager.h
@@ -129,8 +129,11 @@ private:
     std::vector<std::unique_ptr<AssetLibEx>> _libs;
     std::vector<AssetLibEx*> _activeLibs;

-    struct LibsByPriority : public std::binary_function<const AssetLibInfo*, const AssetLibInfo*, bool>
+    struct LibsByPriority
     {
+        using first_argument_type  = const AssetLibInfo*;
+        using second_argument_type = const AssetLibInfo*;
+        using result_type          = bool;
         AssetSearchPriority Priority = kAssetPriorityDir;
         bool operator()(const AssetLibInfo*, const AssetLibInfo*) const;
     } _libsByPriority;
diff --git a/Common/util/string_types.h b/Common/util/string_types.h
index 8c0f45e09..0e34237ed 100644
--- a/Common/util/string_types.h
+++ b/Common/util/string_types.h
@@ -78,8 +78,11 @@ namespace Common
 //

 // Test case-insensitive String equality
-struct StrEqNoCase : public std::binary_function<String, String, bool>
+struct StrEqNoCase
 {
+    using first_argument_type  = String;
+    using second_argument_type = String;
+    using result_type          = bool;
     bool operator()(const String &s1, const String &s2) const
     {
         return s1.CompareNoCase(s2) == 0;
@@ -87,8 +90,11 @@ struct StrEqNoCase : public std::binary_function<String, String, bool>
 };

 // Case-insensitive String less
-struct StrLessNoCase : public std::binary_function<String, String, bool>
+struct StrLessNoCase
 {
+    using first_argument_type  = String;
+    using second_argument_type = String;
+    using result_type          = bool;
     bool operator()(const String &s1, const String &s2) const
     {
         return s1.CompareNoCase(s2) < 0;
```

I can't quite figure it out the why of the std::binary_function inheritance in these three places though, and I think simply deleting it should be just fine. Still, wasn't completely sure to just submit a PR. :/

ivan-mogilko:
Fix #2247

std::unary_function and std::binary_function have been deprecated in C++11, and removed completely in C++17.
The std algorithms are supposed to accept any functor that defines correct operator.

Additionally, in AssetManager replaced struct LibsByPriority with a function pointer and 2 sorting functions. This is to avoid making redundant checks for a priority type during sort operations.

### DIFF
--- a/Common/core/assetmanager.cpp
+++ b/Common/core/assetmanager.cpp
@@ -39,20 +39,19 @@ bool AssetManager::AssetLibEx::TestFilter(const String &filter) const
         (std::find(Filters.begin(), Filters.end(), filter) != Filters.end());
 }
 
-
-bool AssetManager::LibsByPriority::operator()(const AssetLibInfo *lib1, const AssetLibInfo *lib2) const
+// Asset library sorting function, directories have priority
+bool SortLibsPriorityDir(const AssetLibInfo *lib1, const AssetLibInfo *lib2)
 {
-    const bool lib1dir = IsAssetLibDir(lib1);
-    const bool lib2dir = IsAssetLibDir(lib2);
-    if (lib1dir == lib2dir)
-        return false; // both are equal, none is less
-    if (Priority == kAssetPriorityLib)
-        return !lib1dir; // first element is less if it's library file
-    if (Priority == kAssetPriorityDir)
-        return lib1dir; // first element is less if it's directory
-    return false; // unknown priority, just fail
+    // first element is less if it's a directory while second is a lib
+    return IsAssetLibDir(lib1) && !IsAssetLibDir(lib2);
 }
 
+// Asset library sorting function, packages have priority
+bool SortLibsPriorityLib(const AssetLibInfo *lib1, const AssetLibInfo *lib2)
+{
+    // first element is less if it's a lib while second is a directory
+    return !IsAssetLibDir(lib1) && IsAssetLibDir(lib2);
+}
 
 /* static */ bool AssetManager::IsDataFile(const String &data_file)
 {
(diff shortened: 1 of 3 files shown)


ericoporto:
> Additionally, in AssetManager replaced struct LibsByPriority with a function pointer and 2 sorting functions. This is to avoid making redundant checks for a priority type during sort operations.

The code also looks a lot more readable this way! Pretty cool.

I don't understand much why it's a struct with overloaded `()` operator instead of a function in the other places, but I guess this is to be able to persist things in there somehow or due to something `std` internals expect in somewhere.

ivan-mogilko:
> why it's a struct with overloaded () operator instead of a function in the other places, but I guess this is to be able to persist things in there somehow (functors + data?) or due to something std internals expect in somewhere. 

This is because there is a number of predefined types like -
```
typedef std::unordered_map<String, String, HashStrNoCase, StrEqNoCase> StringIMap;
```
and template arguments are expected to be Types. You cannot pass an actual function there. You could pass a function type though, but then you'd also have to specify exact function every time you make this map variable.

PR #2252, opened 2023-12-07
Engine: ali3dogl iOS fix for SDL renderbuffer

ericoporto:
- SDL iOS OpenGL view ([SDL_uikitopenglview.m](https://github.com/libsdl-org/SDL/blob/SDL2/src/video/uikit/SDL_uikitopenglview.m)) generates its own framebuffer, use it instead of 0
- Ensure that iOS is initialized with a 0 depth size to avoid generated depth framebuffers
- Add missing GL_FRAMEBUFFER_COMPLETE_EXT definition

@edmundito is working on the iOS port I left unfinished. Most of the changes are in the `iOS` directory. I want to pick the ones that aren't in small bits to make it easier to review, as there isn't a lot of stuff.

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -53,6 +53,7 @@ const char* fbo_extension_string = "GL_OES_framebuffer_object";
 const void (*glSwapIntervalEXT)(int) = NULL;
 
 #define GL_FRAMEBUFFER_EXT GL_FRAMEBUFFER
+#define GL_FRAMEBUFFER_COMPLETE_EXT GL_FRAMEBUFFER_COMPLETE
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
> I want to pick the ones that aren't in small bits to make it easier to review, as there isn't a lot of stuff.

Would it not be more convenient to gather all of these fixes in a single PR and merge when it's ready?

ericoporto:
ok, the diff of the PR will look huge because it starts by deleting everything in iOS directory and there's a lot of old things there - Xcode project that is completely useless, patches that aren't needed, headers for entire libraries and other things that don't make sense. Passed this, the other changes are quite contained.

ivan-mogilko:
For information, I had to add a fix: 7a0680ddb5c296ae95578a005108ba53c0ba52f9

I guess this could cause render mistakes with "render in screen resolution" mode.

PR #2255, opened 2023-12-08
Remove length restrictions from remaining ScriptName and Name object properties

### LINKED ISSUE #1734: AGS 4: remove limits of game objects' script name lengths
There are still length limits to the script names of certain game objects. These should be removed and script names either made unrestricted or limited by a larger length (it's desired for that restriction to be uniform among all the objects).

This is primarily a compiled data format / engine issue; after this is fixed in the engine, the editor should be adjusted to comply to the same format.

Known cases:
* AudioClips
* Characters

ivan-mogilko:
Fix #1734 

I got reminded about this recently, unfortunately there still a number of game objects, such as Character and Inventory item, which names are restricted in length by a hard number of bytes. Inventory item's name is the shortest, being 24 bytes.
This may not be a serious limitation for ASCII games, where 1 char = 1 byte (although some games may feature very long item descriptions), but things get worse in Unicode, where text in utf-8 may have 1 char = 1-4 bytes, thus same field may contain up to 4 times less characters.
In the sense this change is complementary to the Unicode support in 3.6.0, but also done for consistency with other non-restricted script name and name fields.

Although user asked only about Inventory names, I went through the game data and gathered all fields with a hard-limit, wanting to cover them all in one bulk change, if i'm addressing name limits anyway.

There's a number of nuances in this change. I added comments to each commit, and also in code, but here's a summary.

1. Affected properties are:
    - Game name (title);
    - Game's savegame folder name;
    - Character's script name and name (description);
    - Inventory item's name (description);
    - Mouse cursor's script name; this one is not much important, since it's not used anywhere, besides compilation time, but I did it nevertheless for consistency (and in case it will be used at some point).
    - Audioclip's script name and filename.

3. CharacterInfo struct must retain old fixed char array fields for compatibility with plugin API. I also write values there too, as much as they may take. Would the plugin API be adjusted and have direct access to a character data struct removed, then these fields may be removed as well. (This is already true in ags4 branch)

4. All the unrestricted name values are added as an "extension" to the game data, without changing the existing data chunks. This is done for both backward and forward compatibility, as explained in this project wiki article: https://github.com/adventuregamestudio/ags/wiki/Extending-game-and-room-data-format

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -853,6 +853,32 @@ HError GameDataExtReader::ReadBlock(int /*block_id*/, const String &ext_id,
     return HError::None();
 }
 
+// Search and read only data belonging to the general game info
+class GameDataExtPreloader : public GameDataExtReader
+{
+public:
+    GameDataExtPreloader(LoadedGameEntities &ents, GameDataVersion data_ver, Stream *in)
+        : GameDataExtReader(ents, data_ver, in) {}
+
+protected:
+    HError ReadBlock(int block_id, const String &ext_id,
+        soff_t block_len, bool &read_next) override;
+};
+
+HError GameDataExtPreloader::ReadBlock(int /*block_id*/, const String &ext_id,
+    soff_t /*block_len*/, bool &read_next)
+{
+    // Try reading only data which belongs to the general game info
+    read_next = true;
+    if (ext_id.CompareNoCase("v361_objnames") == 0)
+    {
+        _ents.Game.gamename = StrUtil::ReadString(_in);
+        _ents.Game.saveGameFolderName = StrUtil::ReadString(_in);
+        read_next = false; // we're done
+    }
+    return HError::None();
+}
+
 
 HGameFileError ReadGameData(LoadedGameEntities &ents, Stream *in, GameDataVersion data_ver)
 {
@@ -983,6 +1009,16 @@ void PreReadGameData(GameSetupStruct &game, Stream *in, GameDataVersion data_ver
     GameSetupStruct::SerializeInfo sinfo;
     game.ReadFromFile(in, data_ver, sinfo);
     game.read_savegame_info(in, data_ver);
+
+    // Check for particular expansions that might have data necessary
+    // for "preload" purposes
+    if (sinfo.ExtensionOffset == 0u)
+        return; // either no extensions, or data version is too early
+
+    in->Seek(sinfo.ExtensionOffset, kSeekBegin);
+    LoadedGameEntities ents(game);
+    GameDataExtPreloader reader(ents, data_ver, in);
+    reader.Read();
 }
 
 } // namespace Common
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -603,13 +603,15 @@ void UpgradeCharacters(GameSetupStruct &game, GameDataVersion data_ver)
     // Fixup charakter script names for 2.x (EGO -> cEgo)
     if (data_ver <= kGameVersion_272)
     {
-        String tempbuffer;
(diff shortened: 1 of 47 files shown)


Common/game/main_game_file.cpp
```diff
@@ -783,33 +783,73 @@ HError GameDataExtReader::ReadBlock(int /*block_id*/, const String &ext_id,
...
+        size_t num_invitems = _in->ReadInt32();
+        if (num_invitems != _ents.Game.numinvitems)
+            return new Error(String::FromFormat("Mismatching number of inventory items: read %zu expected %zu", num_invitems, (size_t)_ents.Game.numinvitems));
+        for (int i = 0; i < _ents.Game.numinvitems; ++i)
+        {
+            _ents.Game.invinfo[i].name = StrUtil::ReadString(_in);
+        }
+        size_t num_cursors = _in->ReadInt32();
+        if (num_cursors != _ents.Game.mcurs.size())
+            return new Error(String::FromFormat("Mismatching number of inventory items: read %zu expected %zu", num_cursors, _ents.Game.mcurs.size()));
```
ericoporto, line 834: The mismatch is in the number of cursors

ericoporto (changes requested):

ericoporto:
I think the ticket for this #1734

ivan-mogilko:
> I think the ticket for this #1734

Thank you, added to the ticket description.

ivan-mogilko:
Also updated game name and savegame folder, while I'm at this.

ivan-mogilko:
I found a silly problem with the game title + savegame folder name. There's a "game preload" action that loads only a small portion of data necessary to know basic info about this game. This is used for knowing where to store a config file, and also for "--tell" commands. But if there's "extended" unrestricted fields saved in the extension, then this has to either know how to find that extension (skipping the main data prior to it). I'll check out if this may be done without too much hacks...

ivan-mogilko:
Okay, I wrote extensions offset into one of the reserved ints inside the main data block (I think this has been originally suggested by rofl0r, back when we discussed data extensions).
This lets engine jump to the extensions list easily when preloading data, instead of coding skipping through the whole thing item by item.

ericoporto:
Probably stupid question, this extension is for the `game28.dta` file, and it being a 32-bit int is not a problem for a 64-bit engine, is this correct?

ivan-mogilko:
> this extension is for the game28.dta file, and it being a 32-bit int is not a problem for a 64-bit engine, is this correct?

Yes, they are all appended to game28.dta.
32-bit int is a file offset and therefore is not related to the program address bitness, but to how long a file may be. I suppose game28.dta should not exceed 4 GB... The biggest data in it are compiled scripts, and those may be moved out as separate assets.

Editor/AGS.Editor/DataFileWriter.cs
```diff
@@ -1392,7 +1397,8 @@ public static bool SaveThisGameToFile(string fileName, Game game, CompileMessage
             //   foreach (cap in caps)
             //       FilePutString(cap.Name);
             //
-            WriteGameSetupStructBase(writer, game);
+            long ext_off_pos; // position to write extensions offset to
```
ericoporto, line 1400: long behaves different between Windows 64-bit and other platforms when they are 64-bit, can it be other type?

ericoporto (changes requested):

Editor/AGS.Editor/DataFileWriter.cs
ivan-mogilko: Yes, it's C#, and I've been using exact type as returned from Stream.Position:
https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-8.0

Issue #1981, opened 2023-04-16
A plugin interface for adding video codecs


ericoporto:
**Describe the problem**
OGV works, and is ok for the video needs in a low resolution game. But it's compression is not very effective, which becomes noticeable in bigger resolutions. At the same time, video decoding libraries that are effective can be ten times the size of AGS itself, which make adding new codecs in the engine a hard problem.

**Suggested change**
Have some way in the plugin API to add new video codecs to AGS itself, in a way that the new format becomes playable by AGS, and the regular script functions for video are used. The idea is, if in the future some new video functionality is added in AGS, it should be able to leverage this API so it can access more codecs. It would probably be nice that, if the codec is not available it just degrades the experience by not playing the videos.

I don't have the API design tied down, but it should be so that the media/video internals of AGS could leverage such interface.

As an example, [here are builds of libvpx](https://github.com/ShiftMediaProject/libvpx), which would enable decoding of vp9 and vp8 `.webm` videos, you can see the dll files are around 20MB.

ivan-mogilko:
Decoding video is a combination of decoding audio and decoding images, therefore I propose to start with these.
FFmpeg has a very flexible interface, although I do not suggest adding fully same, but it may be a good reference for the design.

The basic idea is this: a codec is something that converts from input data to output data. So in the most primitive case you have a function Convert with input and output buffers as arguments. But there's a problem: a codec may require certain number of data to even begin decoding, so it may ask you to keep passing more until it's ready to return something. On another hand, our engine not necessarily will be able to accept whole returned data (it may be inconvenient).
This brings us to extended different API, where we have separate PutInputData / GetOutputData methods. The user would put data while it can and until codec will tell that no more is necessary / possible to be put; and retrieve as much as wanted and as much as codec can give back.
This kind of interface would work virtually for anything (image, audio, video, any thing else really).

The rest is to decide the structs for input and output data.
Again, in the most primitive case we need just byte arrays. So long as we know input and output formats - we will be able to work with these.
But for the purpose of video (at least, although I may imagine a case when audio will need that too) these packages will also require "timestamps". These will help to sync video and audio pieces (in ffmpeg these all are called Frames).

As for the codec object itself, it needs to be able to register in the engine, possibly providing a list of extensions it would like to decode. Then engine will be able to try all registered plugin codecs, and then fall back it its own.

ericoporto:
[...]

PR #2256, opened 2023-12-09
DRAFT: IAGSAudioPlayer plugin api

ivan-mogilko:
Adds a IAGSAudioPlayer api for plugins, which wraps our OpenALSource object, and lets render any arbitrary sound data using engine's audio output.

This is immediately for the sake of the ["Sprite Video" plugin](https://github.com/ivan-mogilko/ags-spritevideo/), which cannot play sounds at the moment, having no audio output of its own. But of course this may have any other potential future uses, such as playing custom sounds from plugins.

I referred to the conversation in #1981 when organizing api.

The practical use of this may be found in this experimental branch of the "Sprite Video" plugin:
https://github.com/ivan-mogilko/ags-spritevideo/tree/experiment--agsaudioplayer

New API:
```
#define AGS_AUDIOPLAY_FLD_VOLUME        0x0001

#define AGS_AUDIOFRAME_FLD_DATA         0x0001
#define AGS_AUDIOFRAME_FLD_TIMESTAMP    0x0002

// use sdl2 format for test, but maybe switch to our own custom values just in case?
#define AGS_AUDIOFORMAT_F32             0x8120

struct AGSAudioFormat
{
  // Set which fields are valid, see AGS_AUDIOFORMAT_FLD_* flags
  uint32_t Fields;
  int Format; // s16, s32, f16, f32, etc
  int Channels;
  int Freq;
};

struct AGSAudioPlayConfig
{
  // Set which fields are valid, see AGS_AUDIOPLAY_FLD_* flags
  uint32_t Fields;
  float Volume;
};

struct AGSAudioFrame
{
  // Set which fields are valid, see AGS_AUDIOFRAME_FLD_* flags
  uint32_t Fields;
  void *Data;
  size_t DataSize;
  int64_t Timestamp;
};

// Something that plays the sound data
class IAGSAudioPlayer {
public:
  // Tells which version of the plugin API this object corresponds to;
  // this lets users know which of the following methods are valid to use.
  virtual int    GetVersion() = 0;
  virtual void   SetConfig(AGSAudioPlayConfig *config) = 0;
  // Flushes and closes the AudioPlayer.
  // After calling this the IAGSAudioPlayer pointer becomes INVALID.
  virtual void   Close() = 0;
  virtual void   Pause() = 0;
  virtual void   Resume() = 0;
  // *copies* data on call, does not hold the frame (??? need different rule?)
  // returns amount of data copied, or 0 if data cannot be accepted at the moment.
  virtual size_t PutData(AGSAudioFrame *frame) = 0;

protected:
  IAGSAudioPlayer() = default;
  ~IAGSAudioPlayer() = default;
};
```

and...

```
class IAGSEngine {
  <...>

  // open audio player, telling the format of the sound data that will be SENT into this player by user.
  // optionally pass AGSAudioPlayConfig setting playback parameters (these may be reset later).
  AGSIFUNC(IAGSAudioPlayer*) OpenAudioPlayer(AGSAudioFormat *in_format, AGSAudioPlayConfig *config);
}
```

### DIFF
--- a/Engine/media/audio/audio_core.cpp
+++ b/Engine/media/audio/audio_core.cpp
@@ -77,7 +77,9 @@ AudioCoreSlot::AudioCoreSlot(int handle, std::unique_ptr<SDLDecoder> decoder)
     : handle_(handle), _decoder(std::move(decoder))
 {
     _source = std::make_unique<OpenAlSource>(
-        _decoder->GetFormat(), _decoder->GetChannels(), _decoder->GetFreq());
+        _decoder->GetFormat(), _decoder->GetChannels(), _decoder->GetFreq(),
+        2 /* TODO: make this a constant,
+          and add comment explaining why we keep this queue limit low for now */);
 }
 
 void AudioCoreSlot::Init()
--- a/Engine/media/audio/openalsource.cpp
+++ b/Engine/media/audio/openalsource.cpp
@@ -95,12 +95,13 @@ static struct
 // OpenAlSource
 //-----------------------------------------------------------------------------
 
-OpenAlSource::OpenAlSource(SDL_AudioFormat format, int channels, int freq)
+OpenAlSource::OpenAlSource(SDL_AudioFormat format, int channels, int freq, uint32_t max_queue)
 {
     _inputFmt.format = format;
     _inputFmt.channels = static_cast<Uint8>(channels);
     _inputFmt.rate = freq;
-    _alFormat = OpenAlFormatFromSDLFormat(_inputFmt, _recvFmt);;
+    _alFormat = OpenAlFormatFromSDLFormat(_inputFmt, _recvFmt);
+    _maxQueue = max_queue;
     alGenSources(1, &_source);
     dump_al_errors();
     _resampler.Setup(_inputFmt, _recvFmt);
@@ -156,7 +157,7 @@ size_t OpenAlSource::PutData(const SoundBuffer &data)
 {
     Unqueue();
     // If queue is full, bail out
-    if (_queued >= MaxQueue) { return 0u; }
+    if (_queued >= _maxQueue) { return 0u; }
     // Input buffer is empty?
     if (!data.Data || (data.Size == 0)) { return 0u; }
     // Check for free buffers, generate more if necessary
--- a/Engine/media/audio/openalsource.h
+++ b/Engine/media/audio/openalsource.h
@@ -33,12 +33,9 @@ namespace Engine
 class OpenAlSource
 {
 public:
-    // Max sound buffers to queue before/during processing
-    static const ALuint MaxQueue = 2;
-
     // Initializes Al source for the given format; if there's no direct format equivalent
     // found, setups a resampler.
-    OpenAlSource(SDL_AudioFormat format, int channels, int freq);
+    OpenAlSource(SDL_AudioFormat format, int channels, int freq, uint32_t max_queue = 1024u);
     OpenAlSource(OpenAlSource&& src);
     ~OpenAlSource();
 
@@ -87,7 +84,13 @@ private:
     PlaybackState _playState = PlayStateInitial;
     float _speed = 1.f; // change in playback rate
     float _predictTs = 0.f; // next timestamp prediction
-    unsigned _queued = 0u;
+    // Max sound buffers to queue during processing.
+    // TODO: replace limit of buffer count with limit of memory and/or duration.
+    // TODO: add a comment explaining why do we have to set very small queue limit
+    // for regular audio slot play (to decrease custom effects lag),
+    // alternatively have this logic externally to OpenALSource, somewhere in AudioCoreSlot.
+    uint32_t _maxQueue = 1024u;
+    uint32_t _queued = 0u;
 
     // SDL resampler state, in case dynamic resampling in necessary
     SDLResampler _resampler;
--- a/Engine/media/video/video.cpp
+++ b/Engine/media/video/video.cpp
@@ -36,9 +36,10 @@
 #include "gfx/ddb.h"
 #include "gfx/graphicsdriver.h"
 #include "main/game_run.h"
-#include "util/stream.h"
 #include "media/audio/audio_system.h"
 #include "media/audio/openal.h"
+#include "util/memory_compat.h"
+#include "util/stream.h"
 
 using namespace AGS::Common;
 using namespace AGS::Engine;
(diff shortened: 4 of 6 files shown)


ivan-mogilko:
Although formally working, there are few things that I do not like in this approach. Primarily, my mistake is that engine implements this interface as a specialized wrapper around existing "player" (OpenALSource).
The proper approach would be instead for the engine to use the "audioplayer" interface itself right in the audioslot, and implement this interface in OpenALSource class. Then similarly make exposable interface for the audio decoder and implement it in SDLDecoder.
This way we would both be able to export these to plugins, and be potentially ready to use custom implementations from plugins, if that will be wanted in a future.

ericoporto:
Uhm, the FontRenderer is an external thing that renders text and the Engine gets to use but this AudioPlayer is an external thing that (possibly decodes audio and) puts the audio data for the Engine to play. So the FontRenderer is for you to implement a font renderer and the audio player is for you to implement an audio ??? - a decoder?

ivan-mogilko:
> the FontRenderer is an external thing that renders text and the Engine gets to use

Not necessarily external, engine has two internal font renderers: WFNFontRenderer and TTFFontRenderer. Both implement IAGSFontRenderer interface, identical to the plugin's. This allows the engine to use a font renderer similarly regardless of what it is (with some small exceptions).

> but this AudioPlayer is an external thing that (possibly decodes audio and) puts the audio data for the Engine to play
> and the audio player is for you to implement an audio ??? - a decoder?

AudioPlayer does not decode audio, it only plays it. It has to be decoded separately.

What I mean is that AudioPlayer may potentially be switched by something given by plugin.

Same if I add IAGSAudioDecoder (and other parts mentioned in #1981), in ideal way engine should be able to replace internal decoders and players with the ones implemented by plugin. This may be achieved only if engine uses these exact interfaces inside its video/audio playback logic.

ericoporto:
> I mean that internal audio player (currently - OpenALSource) could potentially be also switched by something given by plugin

This is more sort of what I meant, the OpenALSource is the audio player but the interface named AudioPlayer is not for it, but for someone implementing their own AudioClip+Decoder of sorts - you read the audio file from the video in the plugin, decode and push the results here.

So when you compare with FontRenderer it's confusing because it's meant for you to implement your own FontRenderer.

ivan-mogilko:
> This is more sort of what I meant, the OpenALSource is the audio player but the interface named AudioPlayer is not for it, but for someone implementing their own AudioClip+Decoder of sorts - you read the audio file from the video in the plugin, decode and push the results here.

But if somebody implements their own IAudioPlayer, then in order to use it inside the engine, then engine would also need to use this interface IAudioPlayer, and not OpenALSource directly.

ivan-mogilko:
I'd like to repost my notes on a problem of "disposing" the objects involved in plugin api, which initially were posted in the comments to already closed stream api PR: https://github.com/adventuregamestudio/ags/pull/2227#issuecomment-1850532997
for the record, as they are also be related to this pr, and practically to any future similar api, where some object may be provided from engine to plugins and other way, and that object's ownership is passed along. In other words, when the receiver must delete the object after use.

Specifics of dynamically loaded plugin is that object allocated in plugin must be deallocated in plugin as well, same goes for the object allocated in the engine (it must be deallocated in the engine). The reason for this is that in theory engine and plugin may be compiled using different compilers, and even have different standard lib linked to them. This is why these objects cannot be deleted with "delete" outside of the module that created them.
This leads to a requirement that there has to be factory methods for creating and destroying each object type. If an object is created in the engine, then engine api must have "create"/"destroy" functions. If an object is created by plugin, then plugin should also provide "create"/"destroy" functions for this object type.
(This refers only to dynamically created objects which ownership is passed along with them. Anything that engine or plugin passes but does not give ownership for do not need this.)

While "create" function is always a part of the factory interface, there are two approaches to "destroy" function:
1. Have it in a factory interface as well, with the object passed as an argument,
2. Have it in an object interface itself, so called from that object.

*

**Having "destroy" function in an object's interface** has a advantage of easy access, because all you need is to have the object pointer itself. But if we want to pass internal engine's objects out to plugins, then we also would have to implement this "destroy" method in them. In other words, our own engine class should have a method that deletes object itself. This may be contradicting the class design, and not look pretty. This may also be not safe if the object of such class can be created as a local variable on stack, because calling that function then would corrupt the stack.

Possible solutions here are:
- Enforce that this object cannot be created on stack. (This may be done by hiding its constructor and providing factory method returning the pointer.) This will change how the class is used in the program. Still the "destroy" method has to be commented with a warning that you should use either `delete` or this method, but not both. Of course if the object in some particular place in the engine may theoretically come *from plugin*, then only "destroy" method must be used.
- Hide this "destroy" method itself by making it private in the class. It will be public in the interface, but private in the implementation. This means that if you have an object as a interface pointer, then you may use the "destroy" method to dispose it. And at the same time you may create a subclass object on stack, and the "destroy" method won't be accessible.
- Write a wrapper over the engine's object, which replicates the base class methods, but adds just this extra "destroy" function. The problem with wrappers is that they may make code cumbersome. The situation becomes worse if plugin may pass interface pointer back to the engine for some purpose. Since engine does not know if this interface has implementation by engine or plugin, it cannot safely cast it to its internal class. This would mean that engine would have to use the "wrapper" interface with "destroy" function in places where it would instead use original class object.

In any case, it's important to note that whenever we have a place where the object may come both from engine and plugin, we must use the interface pointer and not any specific implementation. In such case also we must dispose such object using "destroy" method, and not `delete`, as we don't know where it came from.
This may be automated using smart pointers, as they support custom deleters. So we store the interface in std::unique_ptr or std::shared_ptr, with custom deleter that calls "destroy" method for the object instead of deleting it regular way.

*

**Having "destroy" function in factory interface**, besides the "create" function. This is easy at first, but if the object interface may be implemented by plugins too, there appear a new issue: you cannot tell which "destroy" function to use, as the object's interface itself has no means to tell this.

It means that we have to somehow pair object's interface with a pointer to its "destroy" function anyway. This may be done by passing them together in any api function that accepts this interface. Either as 2 arguments, or grouped into a struct.

To give an example of such struct:
```
template <typename TInterface>
struct DisposableInterface
{
    TInterface *Object;
    void (*Dispose)(TInterface *obj);
};
```

So when any disposable object is passed from plugin to engine, it's done using this struct, for example:
```
typedef DisposableInterface<IAGSStream> DisposableStream;

IAGSAudioDecoder *OpenAudioDecoder(DisposableStream *stream);
```

The struct itself is not owned, and should not be deleted by the engine, of course. We simply take values out from it.
Similar to the first solution, the received object ptr may be stored in a unique_ptr or shared_ptr with custom deleter, except that custom deleter would be using this provided Dispose function pointer.

PR #2259, opened 2023-12-10
iOS port fixes and Xcode project improvements

### LINKED ISSUE #1505: Port AGS 3.6.0 to iOS (SDL2 backend)

edmundito:
Resolves #1505 

This includes some changes in the iOS port to make it fully compatible with SDL and bring it up to par to where it was before, and added some minor improvements to the Xcode project.

### iOS Port improvements
* Added configure calls to the GL driver to set the GL RGB bits to 8 8 8, so that the SDL ios OpenGL view set the color format to `kEAGLColorFormatRGBA8` 
https://github.com/libsdl-org/SDL/blob/release-2.28.5/src/video/uikit/SDL_uikitopenglview.m#L98-L102
* Fix type warnings in the iOS platform driver cpp
* Update iOS platform driver to show iOS alert through SDL
* Update delay call in the iOS platform driver to use `SDL_delay`
* Update agstouch plugin to show, hide, and check status of virtual keyboard by using SDL calls

### Xcode improvements
* Split the targets into three parts: One for the third party libs, one for AGS as an engine, and one for the game application
* Remove files in Xcode project that were removed in master
* Update iOS XCode file settings for ali3dogl and acplios to indent with 2 spaces (because that's how they are right now)

#### Target splitting

The Xcode project now builds into three parts, which simplifies a lot of things:
Captura de pantalla 2023-12-10 a la(s) 1 55 17 p m

* `AGS_ThirdParty_iOS` builds all the external libraries from libsrc (excluding SDL) into a static library (.a) file
* `AGS_iOS` is an [binary framework ](https://developer.apple.com/documentation/xcode/configuring-your-project-to-use-mergeable-libraries) containing the Common, Engine, and Plugins
* `AGS Game` is the actual aplication itself. It only builds the files added to the game and references the two libraries above and the SDL framework, and serves as an example setup for an Xcode project

This change will developers porting to iOS to be able to export the AGS and SDL into their own Xcode project, instead of having to fork the whole source code per game.


Engine/platform/ios/acplios.cpp
```diff
@@ -359,15 +356,13 @@ void AGSIOS::DisplayAlert(const char *text, ...) {
...
-//  ios_show_message_box(displbuf);
   
-  //while (ios_wait_for_ui)
-  //  usleep(200);
+  Debug::Printf(kDbgMsg_Warn, displbuf);
+  SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_WARNING, nullptr, displbuf, nullptr);
 }
 
 void AGSIOS::Delay(int millis) {
-  usleep(millis);
```
edmundito, line 370: I'm pretty sure this was incorrect because `usleep` accepts seconds and not milliseconds, so a conversion was needed. The SDL Implementation uses the best option for the platform and takes milliseconds as the input.

edmundito:
This should be ready for review after speaking with @ericoporto and realizing there were a few changes that needed to be pulled in. Tested locally with a real AGS game to ensure that it builds and runs properly. Only tested the simulator build.

Engine/platform/ios/acplios.cpp
```diff
@@ -357,11 +357,7 @@ void AGSIOS::DisplayAlert(const char *text, ...) {
   vsnprintf(displbuf, 2000, text, ap);
   va_end(ap);
   
-  // TODO: Print this properly?
-  printf("%s", displbuf);
-  
-//  Debug::Printf(kDbgMsg_Info, text);
-  
+  Debug::Printf(kDbgMsg_Warn, displbuf);
```
ivan-mogilko, line 360: The safer form would be 
`Debug::Printf(kDbgMsg_Warn, "%s", displbuf);`
just like with printf, in case input text contains formatting tokens.

ivan-mogilko, line 360: The safer form would be similar to one previously used with printf:
`Debug::Printf(kDbgMsg_Warn, "%s", displbuf);`
This is in case input contained formatting tokens.

ivan-mogilko:
Question: why does iOS port need the search for games? Is not it supposed to receive game location passed into cmd args similar to Android?

ericoporto:
> Question: why does iOS port need the search for games? Is not it supposed to receive game location passed into cmd args similar to Android?

Guess that is on me, it's true, this is a possibility, I think in Android I had a better understanding of Java and the system to wire-up the things that make that work - note you never set the name of the game manually in the Android single game project and it still manages to make it work!

In iOS we have the world of ObjectiveC, the Xcode project and things like that, I didn't had a good enough understanding of it to figure a way to make it pass the internal path of the game without one have it hardcoded in acplios itself (like `game.ags` or similar), but leaving AGS to discover the ags game file in the bundle does work - and one doesn't have to rename their game to `game.ags` or adjust the code.

edmundito:
Pushed a couple of changes:
* Fix for @ivan-mogilko 's feedback on `Debug::Printf`
* Moved building the libsrc c files in Engine and Core from the `AGS_iOS` framework to `AGS_ThirdParty_iOS` lib

edmundito:
Pushed a couple of more changes after testing on device:

1. Remove the default bundle identifier because setting it to something specific may register with Apple causing conflicts. Developers should define an identifier for their domain before testing.
2. Added the allow highdpi flag (suggestion by @ericoporto) after testing the high-dpi game [Stranger in Utopia](https://lorenzob.itch.io/stranger-in-utopia)

ivan-mogilko:
> In iOS we have the world of ObjectiveC, the Xcode project and things like that, I didn't had a good enough understanding of it to figure a way to make it pass the internal path of the game without one have it hardcoded in acplios itself (like game.ags or similar), but leaving AGS to discover the ags game file in the bundle does work - and one doesn't have to rename their game to game.ags or adjust the code.

Could you please add a TODO comment somewhere around that code, explaining this case?

ericoporto:
I will write this in better details and add as a TO-DO in either the Readme or in acplios.

Issue #601, opened 2019-02-01
Duplicate/Clone GUI


selmiak:
I noticed I wanted to clone a gui and change some minimal parameters, why is this not possible? You can export the gui to a *.guf file and reimport that one, but direct cloning seems a more common thing to do than exporting. Exporting is important too, but just cloning would be a nice feature to have. Just keep interating the gGUIname through gGUIname2, gGUIname3, etc.

I noticed this on GUIs, but I think a cloning function for characters, inventory items, and more with the exact same settings will be useful too.

ivan-mogilko:
The task should be following:

1. Implement "Duplicate" or "Clone" context menu command for each eligible project item (see below).
2. The command should take the current item and create a shallow copy, that is copy only the item itself, but not the resources it uses or other items it references.
3. The copy is assigned the first free ID, and a unique name, preferably using existing method `Factory.AGSEditor.GetFirstAvailableScriptName`.

Eligible items:
* Character
* Cursor
* Font
* GUI
* InventoryItem
* View

Others:
* Dialogs: no idea if may be useful. Maybe cloning dialogs may be useful if there's a need to have repeating options, or other reasons. But then there's a question whether they should be copied with a script or not.
* Scripts: probably not? easy to do by hand.
* Rooms: may be done, but even if so, I'd suggest doing this as a separate task, as it involves copying lots of material.

PR #2260, opened 2023-12-12
Fix editor config invalid

### LINKED ISSUE #2206: Linux cfg file lists Windows driver
**Describe the bug**
Compiling a game for Linux, by default the Linux/data/acsetup.cfg file will contain the following line which is not valid for Linux,
`driver=D3D9`.
This produces a warning in the logs shown to stdout when the game starts up. I'm not sure _which_ driver is loaded instead, but the game does work (so this is not particularly urgent).

**AGS Version**
3.6.0.52

**To Reproduce**
1. Compile a game for Linux, with graphics driver set to D3D9
2. Open acsetup.cfg in a text editor
3. See error

**Expected behavior**
When compiling a game for any system, its acsetup should contain values valid for that system, or be blank (I'm assuming that writing "driver=" will use the system's default graphics driver, just like with audio).

Ubuntu 22; I haven't checked if this also applies to Android or Mac builds.

ericoporto:
Fix #2206


ivan-mogilko:
I think cloning may be done using reflection, about same way as automatic serialization is done in SerializeUtils: run through all properties that have setters, optionally skip properties with certain attributes (or not?), and copy property values. That might result in a similar effect, but skipping writing and parsing XML doc in memory.

If you add Clone method, probably it makes sense to add ICloneable to the list of interfaces this class implements.

ericoporto:
> it makes sense to add ICloneable to the list of interfaces this class implements.

There's a weird note in .net docs to not put ICloneable in public APIs ([at the last sentence here](https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-8.0#notes-to-implementers)).

ivan-mogilko:
> There's a weird note in .net docs to not put ICloneable in public APIs ([at the last sentence here](https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-8.0#notes-to-implementers)).

This sais:
> Because callers of [Clone()](https://learn.microsoft.com/en-us/dotnet/api/system.icloneable.clone?view=net-8.0#system-icloneable-clone) cannot depend on the method performing a predictable cloning operation, we recommend that [ICloneable](https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-8.0) not be implemented in public APIs.

Is not the same true for this Clone method? It does not seem to guarantee that all fields are cloned, does not let choose whether to do shallow or deep copy. Actually, since the current implementation secretly uses xml serialization, the result would probably be a deep copy (latter is probably not important for RuntimeSetup class in particular).

ericoporto:
@ivan-mogilko , here is what I did

- I added a test for RuntimeSetup, the real focus is I wanted to test Cloning it, so a lot of other things there (like serialization) is not tested.
- for rewriting `Clone()`
  - I understood that I had to create a generic way for cloning, but I could not quite figure how (RuntimeSetup doesn't have an empty constructor for instance), so I instead created a method for copying properties from one object to the other in the Utilities unit. This method lacks some safety checks - objects of different types will result in things going wrong...
  - I then rewrote RuntimeSetup Clone method to use it.

I think better understanding of what cloning mean will have to be exercised at a later time when solving #601.

ivan-mogilko:
The problem that I was referring to in my previous comment is this: if you are adding a method to the public interface, then you have to clearly specify what is expected from this method, and this expectation has to be kept in the future (and likely repeated for each similar class, since this is a universal kind of function).
If you're certain about this now then fine; but if not, then it's better to not add anything in the class itself, and use helper method(s) for this for the time being.

> I understood that I had to create a generic way for cloning, but I could not quite figure how (RuntimeSetup doesn't have an empty constructor for instance)

Hmm, yes, that's an issue. It depends on a use case then... but property copying function that you wrote seems fine already.

ivan-mogilko:
> I will try to hide this inside BuildTargetBase, but keep it public (in Editor project) just for testing

I recall that there's a "internal" access modifier which makes things be seen only within an assembly, idk if this will help here (are tests located in a separate one?)

ericoporto:
Hey, I can squash the last 5 commits if needed. But did as I mentioned and hidden the cloning in BuildTargetBase

PR #2261, opened 2023-12-14
Small fixes for Android and iOS

ivan-mogilko:
- Use FindFile class when searching for translations;
- Uniform DisplayAlert code.
- Use our Path functions when getting parent path.
- Add `const` to `char*` where makes sense.

Cannot test this on iOS myself.

On a side note, these likely could benefit from a MobilePlatform base class.

### DIFF
--- a/Engine/platform/android/acpland.cpp
+++ b/Engine/platform/android/acpland.cpp
@@ -15,9 +15,7 @@
 
 #if AGS_PLATFORM_OS_ANDROID
 #include <ctype.h>
-#include <dirent.h>
 #include <stdio.h>
-#include <sys/stat.h> 
 #include <unistd.h>
 #include <SDL.h>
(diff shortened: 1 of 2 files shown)


ericoporto:
> these likely could benefit from a MobilePlatform base class

I guess if one is written, it could go here: https://github.com/adventuregamestudio/ags/blob/master/Engine/platform/base/mobile_base.h

I wonder if the Display Alert could be in base itself and then we just delete it from the other places - maybe not the Windows one. It may need the parent window passed there for the desktop platforms instead of the nullptr.

ericoporto:
I verified this still works fine with the game! Here is an example of placing an empty game and having an error message appear, to demo the Display Alert

image

In a second run after removing the breakpoint

Screenshot 2023-12-15 at 22 00 01

PR #2262, opened 2023-12-15
Refactor Stream hierarchy into a Stream - StreamBase pair

ivan-mogilko:
This is for a post-3.6.1 time. Optionally I may add this for ags4 right away.

This is a change that we've been discussing with Nick Sonneveld several years ago.
Refactors existing Stream classes hierarchy, separating a basic "device" class and Stream helper wrapper, in a similar vein to how C++ standard library (stream / streambuf) and many frameworks has this.

**Old state:**
- IStream is a pure virtual interface for the basic stream class.
- Stream is a base class for all streams, every stream type is derived from it, or from DataStream specifically.
- Stream interface mixes both low-level operations and helper methods, e.g. ones that read and write independent of system endianess.
- IStream is matched by the engine's plugin API (IAGSStream).

**New state:**
- IStreamBase is a pure virtual interface for the basic stream class.
- StreamBase is a parent class for derived stream "device" implementations (File, Memory, etc).
- Stream is a thin wrapper over a IStreamBase pointer with no virtual members.
- Stream owns IStreamBase object, unless ordered to give it away it by calling ReleaseStreamBase().
- DataStream class is removed (because Stream class now has the necessary endianess conversions).
- IStreamBase is matched by the engine's plugin API (IAGSStream).

IStreamBase provides basic operations over a stream.
Stream works with IStreamBase and provides a multitude of helper methods, similar to what DataStream had been before.
Stream owns IStreamBase and disposes it on destruction by default, but it's possible to release StreamBase's ownership with a call to ReleaseStreamBase(). This ability is left for cases when you need to wrap StreamBase temporarily, but keep it after use.

BufferedStream is no longer a child of FileStream, but a wrap around actual stream base, such as FileStream or AAssetStream.
As a consequence, AAssetStream no longer needs to support section, it's done either using SectionStream or BufferedStream, or any similar proxy object.

Besides changes to the stream classes, changes to the rest of the engine code are:
- Constructing a Stream object now requires to pass a IStreamBase ptr, similar to:

    `Stream s(std::make_unique<FileStream>(...args...));`

**Reasoning**

1. In some cases we may need a basic stream object that has only basic Read/Write methods, while having anything extra will add unnecessary dependencies.
2. We may expect basic stream object to be provided by an external code (like plugin), and thus cannot be of our Stream class. This is why Stream, as a host for helper read/write methods, has to be a wrapper over a base stream.
3. Multiple i/o devices may need buffering, so BufferedStream must not be a child of FileStream. One option is to have it a parent to FileStream, and any other basic stream class that might need buffering. But then there may be other cases where we may need extra processing over a read data before letting engine to read from it. Good example is zip compression: in such case there has to be a ZipStream wrapping basic stream class. Since ZipStream would have buffering too, it would not be optimal to have basic stream class also have a buffer; instead it may be better even to have ZipStream derived from BufferedStream.


Common/util/stream.cpp
```diff
@@ -107,12 +111,42 @@ size_t Stream::WriteByteCount(uint8_t b, size_t count)
...
-    _start = std::max((soff_t)0, std::min(start, end));
-    _end = std::min(std::max((soff_t)0, end), base->GetLength());
+    OpenSection(base, start, end);
+    _base = base;
+}
+
+void StreamSection::Open(IStreamBase *base)
+{
+    if (!base)
+        throw std::runtime_error("Base stream invalid.");
```
ericoporto, line 134: Hey, can you try adding `#include <stdexcept>` on top?

I gave a try [here](https://github.com/ericoporto/ags/commit/97e16415a2993c886f6ae7c9478b1b60ac8450af) and it fixed the build errors [in my GHA CI](https://github.com/ericoporto/ags/actions/runs/8445607277)

ivan-mogilko:
Okay, I mark this PR as ready. It took bit long not because of technical difficulties, but rather because i had doubts for a while. I spent extra time experimenting with various use cases, and also checking some other programs, such as ScummVM.
The doubts were not so much about the new internal structure of stream classes, but about convenience of use. But I hope that any issues there will be resolved eventually; it's always possible to add some helper functions which create certain stream setup more simply.

I also dropped other refactor steps from this PR, but I will restore them as a separate PR.

ericoporto:
Small question: is the endianness work in streams finished? I swear I saw a mention of this somewhere... I see the reading functions have an ifdef to deal with endianness. And the writing have methods like below

```
size_t WriteInt32(int32_t val)
    {
        val = BBOp::Int32FromLE(val);
        return Write(&val, sizeof(int32_t));
    }
```

So the writing is always little endian.

I ask because I think I read at some point that endianness was going to be dealt with after the PR but it looks like it's already done. :)

ivan-mogilko:
The endianness conversion is still performed, but there are 2 differences in code:
1. Previously it was performed by a DataStream class (which was derived from Stream). Now it is performed by the base Stream class itself, as DataStream is removed.
2. Previously there has been an explicit variable that told the target endianness, now there's no such variable and target endianness is calculated from the compilation flags.

ericoporto:
Hey, I don't know how to test this besides throwing a few games to see if they still work alright, and so far that's what I did, and they do. I think it's better to merge this since you already have additional work on top of this in a different pr.

PR #2263, opened 2023-12-16
Engine: report extra mem info on OoM failure

ericoporto:
The error message for Out of Memory would look like this:

image

Here is a demo game that causes an OoM error: [OoMTest.zip](https://github.com/adventuregamestudio/ags/files/13691155/OoMTest.zip)

- https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex
- https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo

### DIFF
--- a/Engine/platform/windows/win_ex_handling.cpp
+++ b/Engine/platform/windows/win_ex_handling.cpp
@@ -18,14 +18,16 @@
 #include <new.h>
 #include <cinttypes>
 #include <stdio.h>
+#include "platform/windows/windows.h"
+#include <Psapi.h>
 #include "ac/common.h" // quit
 #include "ac/common_defines.h"
 #include "debug/out.h"
 #include "util/ini_util.h"
 #include "main/main.h"
 #include "platform/base/sys_main.h"
-#include "platform/windows/windows.h"
 #include "script/cc_common.h"
+#include "ac/spritecache.h"
 
 #if !AGS_PLATFORM_DEBUG
 #define USE_CUSTOM_EXCEPTION_HANDLER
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
@ericoporto almost 1.5 GB of extended sprite memory, this means either memory is filled by dynamic sprites, or spritecache is not working right.

ivan-mogilko:
To double check, I think the quit message should also be printed to the log, correct?

ericoporto:
Ohh, right, I forgot about the log. Yeah, that would be a good idea.

ivan-mogilko:
I had a false memory that it's done in quit function, but I cannot find it there.

ivan-mogilko:
Actually, first thing to do should be to bring DisplayAlert function to consistency. Right now different platforms do 3 things in various combinations there:
- Some display message box using system api or SDL,
- Some print into stdout or stderr depending on a `_logToStdErr` setting,
- Some print to log. (Also, I realized that printing to log uses wrong message type, it should use Alert, but uses Warn)

This all has to be brought to a uniform rule IMO.

ericoporto:
I think Windows default of displaying a message box unless `--no-message-box` is passed is a good default.

I just am not super sure of Windows in particular using the SDL version of the function because the version it does has been working fine and I worry about switching to that this late in beta.

ericoporto:
Otoh, maybe using SDL in Windows can fix #1034, although I forgot how to reproduce that to check if it's still a problem.

ivan-mogilko:
If using SDL's message box on Windows is a concern, then platform driver could have a virtual MessageBox method, overriding it.

Issue #1919, opened 2023-02-14
[AGS4] Unhandled Error in the editor when left clicking on a line in a script.


tarnos12:
**Describe the bug**
When left clicking at a line of code the error pops up.
Perhaps it happens only for newly created functions before/right after saving the file.(just a hunch tho)
It happens only for specific lines which seems to be related to a specific function hence the above statement.
It could also be related to my layout:
![d5dfc238b64881be36cc1c5272dd1a53](https://user-images.githubusercontent.com/10388576/218611869-44b979a8-9c9d-4f03-a11b-70a34cb9c1b8.png)

```
Error: InvalidArgument=The value '4' is not a valid value for 'index'.
Parameter name: index
Version: AGS 3.99.101.0

System.ArgumentOutOfRangeException: InvalidArgument=The value '4' is not a valid value for 'index'.
Parameter name: index
    at System.Windows.Forms.ComboBox.ObjectCollection.get_Item(Int32 index)
    at AGS.Editor.ComboBoxCustom.<OnCreateControl>b__4_2(Object s, DrawItemEventArgs a)
    at System.Windows.Forms.DrawItemEventHandler.Invoke(Object sender, DrawItemEventArgs e)
    at System.Windows.Forms.ComboBox.OnDrawItem(DrawItemEventArgs e)
    at System.Windows.Forms.ComboBox.WmReflectDrawItem(Message&m)
    at System.Windows.Forms.ComboBox.WndProc(Message&m)
    at AGS.Editor.ComboBoxCustom.WndProc(Message&m)
    at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message&m)
    at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message&m)
    at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

**AGS Version**
AGS 4.0
https://cirrus-ci.com/build/5180877416693760

**To Reproduce**
I can't reproduce it.

ivan-mogilko:
Also reproduced in 3.6.0:
#2268

AlanDrake:
I'm still not sure what's going on there, but I expect the problem is happening inside the `DrawItem += (s, a) =>` event inside ComboBoxCustom.

That's where the ineex is being accessed for drawing the item.  I wonder if the event is being added multiple times, or if it's a concurrency issue. Maybe the latter?

[...]

PR #2272, opened 2023-12-21
Editor: ensure themed combobox item index is valid

ericoporto:
for #1919

Not the proper fix but it should at least prevent things from crashing.


ivan-mogilko:
Could you please add a TODO or FIXME comment somewhere around this code, mentioning to check how incorrect index is getting there?

ericoporto:
I think it may be an issue where something takes a certain time to run (why Tarnos get in his long db code and Edmundito gets in his Tween module I guess), and also where something is ran in a draw that should be in an UI thread or similar ([example here](https://jonskeet.uk/csharp/threads/winforms.html)), but I still haven't figured it out exactly.

PR #2275, opened 2023-12-24
More fixes for AAssetStream

ivan-mogilko:
I've been doing adjustments in Stream classes lately, and while reading AAssetStream code noticed that OwnHandle and WrapHandle methods do not actually work, because the associated constructor does not assign a member pointer (and does not init few other members). I guess this was never used nor tested. So I refactored the class a little: picked out a method for opening AAsset pointer by name, and have Open/OpenSection methods take this pointer and "own" parameter. Then they may be called from that non-working constructor too.

Other than that, fixed couple of minor mistakes (minor in the sense that they would rarely be an issue in practice, but still):
* in Seek method "current offset" has to be remembered only after successful asset seek, and old pos kept if there was an error. At least according to fseek documentation, it is supposed to keep old stream position on error. AAsset docs do not explicitly mention this, but I make a guess that it has similar logic.
* in OpenSection method fixed offsets clamping to clamp against real start of the stream (this is in case someone passes a negative offset), and changed the exception message to something more suitable for this case.

### DIFF
--- a/Common/util/aasset_stream.cpp
+++ b/Common/util/aasset_stream.cpp
@@ -28,25 +28,24 @@ namespace AGS
     namespace Common
     {
         AAssetStream::AAssetStream(const String &asset_name, int asset_mode, DataEndianess stream_endianess)
-                : DataStream(stream_endianess)
-                , _aAsset(nullptr)
+            : DataStream(stream_endianess)
         {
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Tested this, and appears working...

PR #2284, opened 2024-01-07
Common: fix typo leads out of range memcpy

ericoporto:
Hey, I think this is a typo, `LEGACY_MAX_CHAR_NAME_LEN` should really be `LEGACY_MAX_SCRIPT_NAME_LEN` here (they have different max sizes)

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -603,12 +603,12 @@ void UpgradeCharacters(GameSetupStruct &game, GameDataVersion data_ver)
     // Fixup charakter script names for 2.x (EGO -> cEgo)
     if (data_ver <= kGameVersion_272)
     {
-        char namelwr[LEGACY_MAX_CHAR_NAME_LEN];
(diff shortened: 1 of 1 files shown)


Common/game/main_game_file.cpp
```diff
@@ -608,7 +608,7 @@ void UpgradeCharacters(GameSetupStruct &game, GameDataVersion data_ver)
         {
             if (chars[i].legacy_scrname[0] == 0)
                 continue;
-            memcpy(namelwr, chars[i].legacy_scrname, LEGACY_MAX_CHAR_NAME_LEN);
+            memcpy(namelwr, chars[i].legacy_scrname, LEGACY_MAX_SCRIPT_NAME_LEN);
```
ivan-mogilko, line 611: This is correct, but then namelwr should also be made an array of this size, for consistency. (It is declared few lines above.)

ericoporto: Ah, thanks, didn't notice it!

Issue #816, opened 2019-06-05
MAXSAVEGAMES  50


rofl0r:
with quest for infamy, i'm constantly hitting this limit. this is pretty annoying, i constantly have to delete old savegames.

the gui part even seems to be unable to display more than the latest 20 savegames (macro gets redefined). can't really figure out a good reason why it shouldn't be possible to display all savegames in the listbox.

ivan-mogilko:
This is one of those things that continiously get postponed because they are "less important". (This limit only persists if developer used built-in functions to fill the save list)

Should be trivial to remove though.

ivan-mogilko:
> It should be trivial to remove this limit.

Hmm... ok, not so trivial, because of this: https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/game.cpp#L2504

sonneveld:
Is it possible for the game to determine if there's enough space for a new save game? Or does it have to rely on an error message occurring on save?

ivan-mogilko:
> Is it possible for the game to determine if there's enough space for a new save game? Or does it have to rely on an error message occurring on save?

Do you mean a disk space, an available save slot, or available slot in the internal array?

sonneveld:
> Do you mean a disk space, an available save slot, or available element in the internal array limited by MAXSAVEGAMES?

I guess all of the above. I know some games already check if the savegamelist has 50 items so they can show their own message.

sonneveld:
`The SaveGameDialog function uses slots numbered from 1 to 20` <-- from the manual, is this still true?

sonneveld:
Also, some games use custom save slot numbers (like 101, or maybe even 21) for auto save. They can pick these numbers because the engine dialog reserves slots 1-20.

ivan-mogilko:
MAXSAVEGAMES is an arbitrary limit only used in default API functions which display list of saves. Probably intended to emulate old Sierra games.
AGS itself technically supports save slots from 0 to 999, with 999 usually reserved for automatic checkpoint.
I don't think it tests for available disk space before trying to save.


> Also, some games use custom save slot numbers (like 101, or maybe even 21) for auto save. They can pick these numbers because the engine dialog only reserves slots 1-20.

Yes, I did not consider that at first. This could be a real problem.
Removing the list limit itself is not that hard, but this has to be somehow consistent with the script API.
Game developers must be able to limit what is shown in these automatic dialogs.

[...]

PR #2289, opened 2024-01-09
Script API: in FillSaveGameList() accept a range of slots

ivan-mogilko:
I don't like adding "last minute" things  to an upcoming release, but this request seemed easy enough to try.

Historically FillSaveGameList has a hardcoded range of 0 to 99 slots, and for some reason also additionally limits total shown slots to 50 items. This change lets user to tell which range of slots to fill, and removes item count limit, because it makes no sense anymore.

```
/// Fills the list box with the current user's saved games in the given range of slots.
import int  FillSaveGameList(int min_slot = 0, int max_slot = 99);
```

**Backwards compatibility**
Old compiled games will be using same old range (0-99) and 50 items limit.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1869,8 +1869,8 @@ builtin managed struct ListBox extends GUIControl {
 	import void Clear();
 	/// Fills the list box with all the filenames that match the specified file mask.
 	import void FillDirList(const string fileMask);
-	/// Fills the list box with all the current user's saved games.
-	import int  FillSaveGameList();
+	/// Fills the list box with the current user's saved games in the given range of slots.
+	import int  FillSaveGameList(int min_slot = 0, int max_slot = 99);
 	/// Gets the item index at the specified screen co-ordinates, if they lie within the list box.
 	import int  GetItemAtLocation(int x, int y);
 #ifndef STRICT_STRINGS
--- a/Engine/ac/global_game.cpp
+++ b/Engine/ac/global_game.cpp
@@ -228,7 +228,7 @@ void FillSaveList(std::vector<SaveListItem> &saves, unsigned bot_index, unsigned
 int GetLastSaveSlot()
 {
     std::vector<SaveListItem> saves;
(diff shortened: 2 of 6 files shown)


ericoporto:
This is related to #816 (I think)

ivan-mogilko:
It is related, but the #816 was about solving this when running already existing games without rescripting them.

This pr only lets game authors to show more in their games. There may have to be a separate change to let override this limit in existing games.

ericoporto:
> I might actually add the config option which hacks this number for testing / running old games, which will likely be enough to close

Not sure if we need another hack for changing old things, if it's working as is, it's fine, better to look into the future - less options are less surface for bugs.

ivan-mogilko:
This is for running old games with the backwards compatible engine.

ivan-mogilko:
I am actually going to suspend this, so converted to draft for the time being.

While the change to FillSaveGameList itself is really trivial, the consequence is that people may use any arbitrary range of saves.
And then this old DeleteSaveSlot behavior gets in a way:
https://adventuregamestudio.github.io/ags-manual/Globalfunctions_General.html#deletesaveslot
> for historical reasons, when deleting slots in the range of 1 to 50, this function also moves highest slot found in that range to fill the freed slot. For example, if there were slots 1, 2, 3, 4, 5 and DeleteSaveSlot removes slot 3, then save in slot 5 will be renamed to become slot 3.
See commit: https://github.com/adventuregamestudio/ags/pull/2289/commits/8e07fcd191ea55f9a9cee51cae4de07462346209

If I allow to list any arbitrary range of saves in the listbox, then this behavior must also be either adjusted or removed.
As a consequence, there's a chance this will break some game logic, because currently games might account for this, or even use this behavior on purpose. 
Which in turn may lead to users requesting additional setup or commands (like RenameSaveSlot, for a hypothetical example).
In other words, this may require some extra api design, which should not be done in a rush.

ivan-mogilko:
I might remake this PR for ags4 instead, and later see if that is safe to backport or anything for some tiny 3.6 update.

ivan-mogilko:
Implemented in ags4 by #2293. This may be later backported to ags3 branch if really wanted.

PR #2292, opened 2024-01-10
Editor: fix '\[' sequences lost in textual properties

### LINKED ISSUE #2288: Escape of `[` character no longer possible in Label
**Describe the bug**
Using `\[` should escape a `[` character when used in a label. Instead, we get a new line. This can be seen directly in the Editor but it will also show as the same when running the game.

**AGS Version**
v3.6.1.16

**To Reproduce**
1. Create a game from a template
2. Either use an existing GUI or create a new GUI and add a Label
3. In the Label, set it's text property to something like `Test \[ a ]`
4. See it will instead show as
```
Test
 a ]
```

**Expected behavior**
It should read as `Test [ a ]` in the label.

**Screenshots**
image

**Desktop:**
 - OS: Windows
 - Version: 10

**Additional context**
I noticed this in my Don't Give Up the Cat game, because I use `[ ]` and `[x]` as checkbox in the menu options for selecting either Windowed or Fullscreen.

ivan-mogilko:
Fix #2288

The '[' is still supported as a linebreak in 3.* branch, for backwards compatibility. If user wants to have literal '[' they have to escape it like '\\['. But recent change 432ae84 have introduced Unescape call for properties (in order to support '\n' etc). Unescape call removes backslashes in unrecognized sequences. So now we have to add this hack, escaping a backslash with another backslash, this will keep it intact.

### DIFF
--- a/Editor/AGS.Editor/DataFileWriter.cs
+++ b/Editor/AGS.Editor/DataFileWriter.cs
@@ -292,7 +292,10 @@ namespace AGS.Editor
         /// <param name="text"></param>
         public static string TextProperty(string text)
         {
-            return Regex.Unescape(text);
+            // Escape backslashes before brackets: for '\[' support;
+            // this is needed because Unescape will delete '\' in unrecognized sequence.
+            string escapedText = text.Replace("\\[", "\\\\[");
+            return Regex.Unescape(escapedText);
         }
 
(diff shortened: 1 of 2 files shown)


ericoporto:
This successfully fixes #2288 in my tests!

PR #2293, opened 2024-01-10
AGS 4: Script API: in FillSaveGameList() accept a range of slots

ivan-mogilko:
Historically FillSaveGameList has a hardcoded range of 0 to 99 slots, and for some reason also additionally limits total shown slots to 50 items, which normally means that it's limited to slots 0-50. This change lets user to tell which range of slots to fill, and removes item count limit, because it makes no sense anymore.

Same change is done to function RestoreGameDialog and SaveGameDialog. These are old, but may still be used sometimes, so I decided to still do this.

The new functions are declared as
```
/// Fills the list box with the current user's saved games in the given range of slots.
import int  FillSaveGameList(int min_slot = 0, int max_slot = 50);

import void RestoreGameDialog(int min_slot = 0, int max_slot = 50);
import void SaveGameDialog(int min_slot = 0, int max_slot = 50);
```

In addition, removed the historical effect of DeleteSaveSlot which would move topmost save down to fill the gap in the list. This is unexpected to users, and generally is a ugly solution.

In order to let users do the similar behavior if they like (or for any other purpose), added function MoveSaveSlot that moves a save game from one slot to another; effectively this works as a "rename file".

```
/// Moves the save game from one slot to another, overwriting a save if one was already present at a new slot.
import void MoveSaveSlot(int old_slot, int new_slot);
```

**Backwards compatibility**
Old function will be passing same old range (0-50).
The 50 items limit is removed completely though, as there's little purpose.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1409,8 +1409,8 @@ builtin managed struct ListBox extends GUIControl {
 	import void Clear();
 	/// Fills the list box with all the filenames that match the specified file mask.
 	import void FillDirList(const string fileMask);
-	/// Fills the list box with all the current user's saved games.
-	import int  FillSaveGameList();
+	/// Fills the list box with the current user's saved games in the given range of slots.
+	import int  FillSaveGameList(int min_slot = 0, int max_slot = 50);
 	/// Gets the item index at the specified screen co-ordinates, if they lie within the list box.
 	import int  GetItemAtLocation(int x, int y);
 	/// Inserts a new item before the specified index.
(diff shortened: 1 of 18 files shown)


ivan-mogilko:
May we need something like "MoveSaveSlot" or "RenameSaveSlot", in case user wants to emulate the shifting behavior after deleting an item?

The original behavior was ugly, but also unreliable, because it did not account for more possible gaps after player deleting saves by hand.

AlanDrake:
Maybe the shifting behavior could be implemented with a module?
Renaming all saves doesn't feel like something anyone should ever do though...

ivan-mogilko:
Did similar arguments extension to RestoreGameDialog and SaveGameDialog. (I was not sure if we want to keep these, but still these are in still in api, I figured this makes sense to do).

Added MoveSaveSlot function, in case someone wants this ability for their save system.

PR #2295, opened 2024-01-13
Fix Common/Engine/Tools analysis warnings

ericoporto:
This is a bunch of different warnings and analysis picked from different tools, they looked safe enough to apply.

### DIFF
--- a/Engine/ac/parser.cpp
+++ b/Engine/ac/parser.cpp
@@ -269,7 +269,7 @@ int parse_sentence (const char *src_text, int *numwords, short*wordarray, short*
                 // if it's an unknown word, store it for use in messages like
                 // "you can't use the word 'xxx' in this game"
                 if ((word < 0) && (play.bad_parsed_word[0] == 0))
-                    strcpy(play.bad_parsed_word, thisword);
+                    snprintf(play.bad_parsed_word, 100, "%s", thisword);
             }
 
             if (do_word_now) {
--- a/Common/ac/spritefile.cpp
+++ b/Common/ac/spritefile.cpp
@@ -559,7 +559,6 @@ int SaveSpriteFile(const String &save_to_file,
 
     std::unique_ptr<Bitmap> temp_bmp; // for disposing temp sprites
     std::vector<uint8_t> membuf; // for loading raw sprite data
-    std::vector<uint32_t> palette;
 
     const bool diff_compress =
         read_from_file &&
--- a/Common/game/interactions.cpp
+++ b/Common/game/interactions.cpp
@@ -115,6 +115,10 @@ void InteractionCommand::Write(Stream *out) const
 
 InteractionCommand &InteractionCommand::operator = (const InteractionCommand &ic)
 {
+    if (this == &ic) {
+        return *this;  // prevent self-assignment
+    }
+
     Type = ic.Type;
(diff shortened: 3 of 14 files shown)


ivan-mogilko:
I have to honestly say, I don't like use of emplace_back, for two reasons:
* It makes reading code more difficult, because it's not clear which object is constructed.
* MSVS does not hint args, so when you write it you have to lookup the constructor args yourself.

Engine/ac/parser.cpp
```diff
@@ -269,7 +269,7 @@ int parse_sentence (const char *src_text, int *numwords, short*wordarray, short*
                 // if it's an unknown word, store it for use in messages like
                 // "you can't use the word 'xxx' in this game"
                 if ((word < 0) && (play.bad_parsed_word[0] == 0))
-                    strcpy(play.bad_parsed_word, thisword);
+                    strncpy(play.bad_parsed_word, thisword, 100);
```
ivan-mogilko, line 272: Please use snprintf instead of strncpy, strncpy *does not guarantee null-terminator*:
https://en.cppreference.com/w/c/string/byte/strncpy

Another variant is to use newer strncpy_s, it seems, but I've been replacing str(n)cpy to snprintf for quite some time, so that might be more consistent.

ericoporto: I replaced it with `snprintf(play.bad_parsed_word, 100, "%s", thisword);`

ericoporto:
I dropped the emplace_back commits. I get the warnings for prefering emplace and emplace_back from the VS Sonar Lint plugin (I also did the substitutions using it instead of manually writing the code).

ivan-mogilko:
Another thing that I disagree with, is reordering fields in struct, where it sais "reorder to reduce size". I suspect this is related to padding. But the order of fields is commonly done without such intent, unless there's a critical reason to do so, usually it is a matter of organization. The way I do this, for example, is that I try to have most important fields first (like "object type", "id" or "mode") and derivatives or optional settings later. Sometimes I group member variables by "topic". If we will instead put fields trying to prevent few bytes of padding in memory, this will make things annoying.

ericoporto:
> I suspect this is related to padding

So, from what I understand a smaller struct means better cache performance, the smaller struct is achieved by reducing padding through reordering in this case.

> But the order of fields is commonly done without such intent

The two commits are literally the only places where this isn't done in the entire codebase.

ivan-mogilko:
> So, from what I understand a smaller struct means better cache performance, the smaller struct is achieved by reducing padding through reordering in this case.

The thing is, some of this padding occurs simply because some variables are *shorts* and not ints. If they would be replaced by ints that would give same size and no warning of extra padding.

If the size of a struct is critical, then this should not be fixed by simply reordering, but by purportedly reducing the variable sizes, or even packing several values in a single int, and similar.

One of the structs changed in this commit is a single global struct with few ints and bools in it...

ericoporto:
I agree with you so I will drop the commits.

About the running engine, I know that things are not so simple because I have tried to reorder`cc_instance` in the past and after trying many different combinations the one that we currently have was indeed the best performance.

Anyway, I just have to do this later because I need a computer to do it.

ericoporto:
Sorry for the delay, dropped the struct reordering commits now.

PR #2297, opened 2024-01-13
Fix Editor analysis warnings

ericoporto:
These are the things that were brought up as very weird when running the Editor codebase through different code analysis.


Editor/AGS.Editor/Panes/Room/RoomEditNode.cs
```diff
@@ -66,7 +66,7 @@ public RoomEditNode(string uniqueID, IAddressNode[] children, bool visible, bool
         public bool IsVisible 
         { 
             get { return _control == null ? true : _control.IsVisible; }
-            set { if (_control != null) _control.IsVisible = true; }
+            set { if (_control != null) _control.IsVisible = value; }
```
ivan-mogilko, line 69: I do not remember how RoomEditNode works, but apparently this property is never changed after initial initialization; that's why this did not break anything until now.

PR #2298, opened 2024-01-13
Engine: fixed CharacterInfo breaking script compatibility again

### LINKED ISSUE #2296: Backwards Compatibility: comparison between Character pointer and `character` array can "incorrectly" fail
**Describe the bug**
A comparison like `character[0] == cEgo` will always work, but any succeeding character will fail, like `character[1] == cOne`

**AGS Version**
This started to happen in 3.6.1.14, and it did not happened in 3.6.1.13. The issue appears to be this commit https://github.com/adventuregamestudio/ags/commit/4be15565755bc54f5d8f48be0e6aebc6eee117c1

**Game**
[TestCharacterArray.zip](https://github.com/adventuregamestudio/ags/files/13928630/TestCharacterArray.zip)

**To Reproduce**
You can run the test game with 3.6.1.13 and see no error with
```
./acwin_3.6.1.13  --no-message-box --console-attach --log-stdout=all:all
```
```
Script : C: cEgo C*: cEgo Equal: 1 - 1
Script : C: cCharOne C*: cCharOne Equal: 1 - 1
Script : C: cTwo C*: cTwo Equal: 1 - 1
Script : C: cThree C*: cThree Equal: 1 - 1
Script : C: cFour C*: cFour Equal: 1 - 1
Script : C: cFive C*: cFive Equal: 1 - 1
```

And then run in 3.6.1.14 and see the error in the logs
```
./acwin_3.6.1.14  --no-message-box --console-attach --log-stdout=all:all
```
```
Script : C: cEgo C*: cEgo Equal: 1 - 1
Script : C: cCharOne C*: cCharOne Equal: 0 - 0
Script : C: cTwo C*: cTwo Equal: 0 - 0
Script : C: cThree C*: cThree Equal: 0 - 0
Script : C: cFour C*: cFour Equal: 0 - 0
Script : C: cFive C*: cFive Equal: 0 - 0
```
Note the equals should all be true (1), but are false (0).

**Note:** this only errors for characters other than the 0 index.

1. Create a new game from Sierra template and add a new character `cOne`.
2. Go to the room and create a new function in After Fade-in
3. Write it as 
```
function room_AfterFadeIn()
{
  Display("equality: %d \n\n 1 is error,\n 0 is not.",  cOne != character[1]);
}
```
5. Build and run and see the error 

**Expected behavior**
Equality should work (look into additional context below!!!)

**Desktop:**
 - OS: Windows
 - Version: 10

**Additional context**
Note that this code should probably give an error when compiling, because we are freely comparing or casting as pointer a struct that isn't a pointer (the element from character, as that array is an array of characters and not readonly pointers of characters), something like: `Character * cWhatever; import character[]; cWhatever == character[1]`.

A good fix for this is probably https://github.com/adventuregamestudio/ags/issues/2273 , but we can't do that in a backwards compatibility manner (in ags3)

ivan-mogilko:
Fix #2296

This fixes a mistake introduced in commit 4be1556, which in turn followed b02814b8bb3bb7a2cd00110466e2bd3047e772d1

The problem is, that CCStaticArray does not work as intended, and seems won't (at least not without some extra workarounds).

When the script compiler is told to get an Nth element of a global struct array, such as character[n], it calculates the memory address as

    Character* ptr = (array address + sizeof(Character) * n).

If this address is used for the read/write operations, these ops can be intercepted by interpreter and remapped into the real fields (see IScriptObject::ReadN, WriteN interface). But if this address is used IN POINTER COMPARISON BY VALUE, then we cannot do anything. And if our real struct in the engine is stored on a different relative memory offset than one expected by compiler, then this pointer comparison will fail. This breaks script expressions like e.g.:

      if (player == character[n])

---

This PR moves two new "unrestricted" name and scriptname fields out of CharacterInfo and into a new small CharacterInfo2 struct. These structs are stored alongside with CharacterInfo array in GameSetupStruct.

One thing, I left original fixed-length CharacterInfo::scriptname used in engine warnings (logging and error messages), because replacing all those with access to new struct would be too annoying of a change. I'd rather leave this be for now, and look for a better way later.

I do believe that there *might* be a more elegant solution to this issue, but I'd like to keep this as a hotfix for now, and do not want to do much experiment patching upcoming 3.6.1 release.

NOTE: only data organization in memory has changed, file data formats should not be affected (neither game files, nor saves).

### DIFF
--- a/Common/ac/characterinfo.cpp
+++ b/Common/ac/characterinfo.cpp
@@ -114,11 +114,11 @@ void CharacterInfo::WriteBaseFields(Stream *out) const
     out->WriteInt16(acty);
 }
 
-void CharacterInfo::ReadFromFile(Stream *in, GameDataVersion data_ver)
+void CharacterInfo::ReadFromFile(Stream *in, CharacterInfo2 &chinfo2, GameDataVersion data_ver)
 {
     ReadBaseFields(in);
-    StrUtil::ReadCStrCount(legacy_name, in, LEGACY_MAX_CHAR_NAME_LEN);
-    StrUtil::ReadCStrCount(legacy_scrname, in, LEGACY_MAX_SCRIPT_NAME_LEN);
+    StrUtil::ReadCStrCount(name, in, LEGACY_MAX_CHAR_NAME_LEN);
+    StrUtil::ReadCStrCount(scrname, in, LEGACY_MAX_SCRIPT_NAME_LEN);
     on = in->ReadInt8();
     in->ReadInt8(); // alignment padding to int32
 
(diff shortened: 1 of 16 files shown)


ericoporto:
I made a few tests and this worked correct in both Don't Give Up the Cat, in the test game attached to the issue and in a new test I created for the automatic that should run on the pipeline.

PR #2299, opened 2024-01-14
AGS 4: Script API: replace all global arrays of structs with arrays of pointers

### LINKED ISSUE #2273: AGS 4: In Script API replace all arrays of objects with arrays of pointers
This refers to arrays like `Character character[]` and all the similar ones.

Because of how these arrays are declared, compiler can make assumptions about the way these objects are allocated, and it will use their **sizes** (as declared in a API header) when calculating element access. This in turn causes compatibility issues.

There are multiple things that may be planned regarding the way objects are accessed in AGS, such as having instead an indexed property to access them (like `Game.Characters[]`), or having dynamically created objects, which would possibly require to design a different way to access them altogether.

But as a first iteration, these arrays of objects should be replaced with *arrays of pointers*. This step will be fully syntax-compatible, meaning no change is necessary to user scripts.
The changes will be required inside the engine, where we would likely have to export an array filled with the managed handles from respective registered script objects. 

See also: #1729 suggests that arrays of managed structs should be forbidden by the new compiler.

ivan-mogilko:
Resolves #2273

This affects following arrays, some of which are declared in the prepared header, and others are generated from the game:
```
import readonly Character *character[N];
import readonly Dialog *dialog[N];
import readonly GUI *gui[N];
import readonly InventoryItem *inventory[N];
import readonly Object *object[N];
import readonly Hotspot *hotspot[N];
import readonly Region *region[N];
```

`readonly` modifier restricts changing array elements. This does not prevent changing objects themselves, only their pointer values.
Because the syntax change is minimal, all scripts remain *syntactically* compatible and do not have to be adjusted, only recompiled.

**WARNING:** this completely breaks compatibility with *previously compiled scripts*, as in they will no longer work correctly and might cause runtime crashes. In case of emergency that would be possible to support a old-style array variant, but I would not do that without a critical need. After this PR game projects must be fully recompiled.


ericoporto:
I noticed the following comment

> // NOTE: there's always a dummy inv item at index 0

Does this means the `inventory[]` array will always exist? If such, this fixes this issue in the Tween module:

https://github.com/edmundito/ags-tween/issues/20

ivan-mogilko:
> Does this means the inventory[] array will always exist

No, it's not created if there are no items at all.

Besides, there are other arrays that has exact same issue: gui and dialogs.

ivan-mogilko:
But thinking about this, since these are pointer arrays now, I might generated a 1-sized array with a nullpointer in it for the case of no objects.

PR #2302, opened 2024-01-19
Add a script for updating SDL_Sound revision in readmes and cmake files

ivan-mogilko:
I noticed that SDL_Sound's revision is changed by hand all the time, and some files already have different revisions mentioned (even multiple different ones in a single readme).

Added another python script for updating the revision number. This requires a new values group in version.json:
```
    "sdlsound": {
        "revision": "c5639414c1bb24fb4eef5861c13adb42a4aab950",
        "urlHash": "d1037daa0e444efa8a09defaa1a6321f"
    }
```

Affected files are hardcoded in the script, and may be amended as necessary. Currently they are:
* CMake/FetchSDL_Sound.cmake
* debian/README.md
* OSX/README.md
* Windows/README.md

---

The full script's work may be tested by inserting random values into json, and running "push_sdlsound.py".

### DIFF
--- a/CMake/FetchSDL_Sound.cmake
+++ b/CMake/FetchSDL_Sound.cmake
@@ -2,7 +2,7 @@ FetchContent_Declare(
     sdlsound_content
     DOWNLOAD_EXTRACT_TIMESTAMP TRUE
     URL https://github.com/icculus/SDL_sound/archive/c5639414c1bb24fb4eef5861c13adb42a4aab950.tar.gz
-    URL_HASH MD5=d1037daa0e444efa8a09defaa1a6321f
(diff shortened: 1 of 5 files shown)


ericoporto:
Two things that I update manually is there is a sha1sum check for the libraries in that are fetch (tar.gz) using `libsrc/download.sh` and there is md5sum for the same file used in the [`CMake/FetchSDL_Sound.cmake`](https://github.com/adventuregamestudio/ags/blob/master/CMake/FetchSDL_Sound.cmake). These were arbitrarily chosen. CMake supports `URL_HASH SHA1` so we could instead switch to that and use the same sha1sum in both cases.

PR #2303, opened 2024-01-19
Unify DisplayAlert in all platform classes

### LINKED ISSUE #2279: macOS engine: Show error message as as a dialog window
**Describe the problem**
Currently, error messages displayed by the macOS (OSX) engine are only shown in the log, which means that when the game crashes for a Mac user, there is no visible error message. The only way a mac user can see the error is by running the app directly from the Terminal, so that the engine logs display on the terminal console.

Here's a reference to the OSX `DisplayAlert` implementation, as you can see there it only prints the console
https://github.com/adventuregamestudio/ags/blob/9cc72e49301120fc2e1a8bd672cc1328f7e5a600/Engine/platform/osx/acplmac.cpp#L71-L81

**Suggested change**
Leverage [`SDL_ShowSimpleMessageBox`](https://wiki.libsdl.org/SDL2/SDL_ShowSimpleMessageBox) to show the error message using the native dialog window. This was done recently for iOS and works great:

https://github.com/adventuregamestudio/ags/blob/9cc72e49301120fc2e1a8bd672cc1328f7e5a600/Engine/platform/ios/acplios.cpp#L325-L334

ivan-mogilko:
Fix #2279

1. Implement general DisplayAlert logic in the base AGSPlatformDriver class. This method does 3 things: print to log, write to stdout and display message box, depending on current settings.
2. Add new method called DisplayMessageBox which purpose is strictly to display a message box if platform can do that. Platforms that cannot or should not may override this method, or override _guiMode variable.
3. Added a "dirty" fix for duplicate alert in stdout, which may appear because of passing the message to the log system. The platform class will store a pointer to a function that has to be called when printing a message. This pointer is assigned according to "_logToStdErr" setting. When printing, use this pointer instead of checking for the setting. This also lets setting this pointer to null, or a stub that does nothing, hence temporarily disabling an output.

### DIFF
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -114,12 +114,13 @@ void AGSPlatformDriver::DisplayAlert(const char *text, ...)
         DisplayMessageBox(displbuf);
 }
 
-void AGSPlatformDriver::WriteStdOut(const char *fmt, ...) {
+void AGSPlatformDriver::WriteStdOut(const char *fmt, ...)
+{
     va_list args;
     va_start(args, fmt);
-    vprintf(fmt, args);
+    vfprintf(stdout, fmt, args);
     va_end(args);
-    printf("\n");
+    fprintf(stdout, "\n");
     fflush(stdout);
 }
 
@@ -130,7 +131,7 @@ void AGSPlatformDriver::WriteStdErr(const char *fmt, ...)
     vfprintf(stderr, fmt, args);
     va_end(args);
     fprintf(stderr, "\n");
-    fflush(stdout);
+    fflush(stderr);
 }
 
 void AGSPlatformDriver::DisplayMessageBox(const char *text)
--- a/Engine/platform/base/agsplatformdriver.cpp
+++ b/Engine/platform/base/agsplatformdriver.cpp
@@ -45,6 +45,11 @@ AGSPlatformDriver *platform = nullptr;
 
 // ******** DEFAULT IMPLEMENTATIONS *******
 
+AGSPlatformDriver::AGSPlatformDriver()
+{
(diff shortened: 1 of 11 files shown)


ericoporto:
The check for `_guiMode` can be in only one place (I guess in `DisplayAlert` and not in `DisplayMessageBox`, unless `DisplayMessageBox` is called "naked" elsewhere).

I think the `#ifdef` below here could also be removed in this PR

https://github.com/adventuregamestudio/ags/blob/fba9544af4e652f831a43ff6557d35dd7538950a/Engine/main/main.cpp#L160-L162

ivan-mogilko:
> The check for _guiMode can be in only one place (I guess in DisplayAlert and not in DisplayMessageBox, unless DisplayMessageBox is called "naked" elsewhere).

I am checking it there because DisplayMessageBox is in public interface.

ivan-mogilko:
By the way, I kept the Windows implementation for DisplayMessageBox, because the SDL is using something different, and it looks bit strange.

ericoporto:
I agree with keeping our own windows implementation there, I also thought that their windows one looked weird.

Engine/platform/base/agsplatformdriver.cpp
```diff
@@ -86,6 +87,27 @@ void AGSPlatformDriver::GetSystemTime(ScriptDateTime *sdt) {
     sdt->year = newtime->tm_year + 1900;
 }
 
+void AGSPlatformDriver::DisplayAlert(const char *text, ...)
+{
+    char displbuf[1024];
```
ericoporto, line 97: I am probably paranoid, but it seems we use 2000 and now this is 1024. I would keep it at 2000 just to be safe.

ericoporto:
this looks good!

PR #2304, opened 2024-01-20
AGS 4: Script API: implement FaceDirectionRatio

### LINKED ISSUE #2243: AGS 4: implement Character's direction ratio property
### Problem

AGS has a hardcoded factor for switching directional character view loops depending on the direction they are facing. There were numerous requests to let user configure these in the past, when they like to emulate different perspectives, such as isometric movement and whatnot. Yet it's not possible to do in AGS without scripting switching loops yourself.

---

### Potential solution

The basic suggestion is explained on ags forums here by @messengerbag, with illustrations:
https://www.adventuregamestudio.co.uk/forums/editor-development/feature-request-ability-to-create-16-character-facing-directions/msg636657477/#msg636657477

Quote:
> As you can see, as the circle and cross get more and more flattened by **foreshortening**, the front and back view angles get wider, while the side view angles get narrower.
> We can represent this perspective distortion as **a single number**: the height/width ratio of the circle or ellipse.

Coincidentally, there have been already a feature made by @AlanDrake in his experimental "draconian" version, I think it's been around for many years, but never made it to main ags version. It's called "DirectionRatio", and is represented by a script function `Game.SetDirectionRatio(float ratio)`.

Example commit: https://github.com/AlanDrake/ags-draconian/commit/bdf8cffe4922b98df06933985f0b43dd4401e07e

From the comment's description:
> Allows to change the way the direction view is calculated. Can be useful for better matching the room perspective.
What it does is change the importance of the X axis in the original calculations.
Default is 1.0, from 2.0 to 3.0 could be good values for a room in perspective.

---

### Suggestion

1. Decide on the API terminology; would that be a "foreshortening" or "direction ratio", or else. It should be clear enough for users to understand its purpose: it is for choosing the facing/moving angles at which to switch from one directional view loop to another.
2. Implement a number of properties that let user configure this factor for various scenarios. For the most basic effect a global `Game.DirectionRatio` would be enough (I am using example name here). But I'd suggest to consider having several overriding properties: `Game.DirectionRatio` -> `Room.DirectionRatio` -> `WalkableArea.DirectionRatio` -> `Character.DirectionRatio`, where each property has a default value meaning "use parent", and character's individual property (when set) overrides all the above. EDIT: I suppose that value 0.0 may be "use parent" / "undefined", unless there may be scenarios where it's valid? If so, then we may add a constant with a -1.0 value to be used here.
3. It's more essential to have this in script API, to let change dynamically on need, but of course would be nice to have in Editor properties too.

ivan-mogilko:
Resolves #2243

This implements new property called FaceDirectionRatio for Game, Room, WalkableArea and Character. This property affects selection of directional loop while Character moves and turns.

```
/// Gets/sets the default y/x ratio of character's facing directions, determining directional loop selection for all Characters in game.
static attribute float Game.FaceDirectionRatio;

/// Gets/sets the optional y/x ratio of character's facing directions, determining directional loop selection for each Character in the current room.
static attribute float Room.FaceDirectionRatio;

/// Gets/sets the optional y/x ratio of character's facing directions, determining directional loop selection for each Character while on this area.
static attribute float FaceDirectionRatio;

/// Gets/sets the optional y/x ratio of character's facing directions, determining directional loop selection while Character moves and turns.
attribute float Character.FaceDirectionRatio;
```

Ratio's implementation is loosely based on the experimental commit in @AlanDrake 's custom "Draconian" engine. The meaning of ratio is (Y / X) relation. Default value is 1.0, where loop selection strictly corresponds to movement and facing dir. Values >1.0 mean that Y axis has more "weight", and vertical loops are chosen for the wider range of angles. Values <1.0 mean that X axis has more "weight", and horizontal loops are chosen for the wider range of angles.

This may be also illustrated with this pic:

The properties are checked in following order:

1. Individual Character.FaceDirectionRatio, if it's > 0;
2. WalkableArea.FaceDirectionRatio, if it's > 0;
3. Room.FaceDirectionRatio, if it's > 0;
4. If none of above, then Game.FaceDirectionRatio

---

Potential TODOs:
- [x] Add respective properties in General Settings and Room properties in the Editor. Idk if Characters need an individual design-time property for this, I suppose character's property serves rather as an override in some narrow use cases.
- [x] Add property for Walkable Areas in the Editor, in which case it will be checked after Character's but before Room's.
- [x] Float properties are parsed depending on Culture, which should not be happening. To clarify: on some systems it will require "2.5" (with a dot) and on others "2,5" (with a comma). Need to figure out how to enforce InvariantCulture in the PropertyGrid.
- [x] Add script property for Walkable Areas, but I don't want to do this right away here, as WAs don't have their respective struct. I'd rather wait until their api is replaced with OO-style api first (opened pr for this #2307).
- [x] Save properties that may be changed at runtime.

~[ ] Optimize direction ration calculation by caching few things in character class.~

---

Test game, made with assets by user who requested this feature recently:
[IsoBikeDemo.zip](https://github.com/adventuregamestudio/ags/files/14316758/IsoBikeDemo.zip)

### DIFF
--- a/Common/ac/gamesetupstruct.h
+++ b/Common/ac/gamesetupstruct.h
@@ -52,9 +52,6 @@ struct GameSetupStruct : public GameSetupStructBase
     std::vector<FontInfo> fonts;
     InventoryItemInfo invinfo[MAX_INV]{};
     std::vector<MouseCursor> mcurs;
-    // CLNUP old interactions
-    //Interaction     **intrChar;
-    //Interaction      *intrInv[MAX_INV];
     std::vector<PInteractionScripts> charScripts;
     std::vector<PInteractionScripts> invScripts;
     // TODO: why we do not use this in the engine instead of
@@ -84,6 +81,10 @@ struct GameSetupStruct : public GameSetupStructBase
     int               numGameChannels = 0;
     // backward-compatible channel limit that may be exported to script and reserved by audiotypes
     int               numCompatGameChannels = 0;
+
+    // Extended global game properties
+    // Character face direction ratio (y/x)
+    float             faceDirectionRatio = 1.f;
     
     // TODO: I converted original array of sprite infos to vector here, because
     // statistically in most games sprites go in long continious sequences with minimal
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -526,6 +526,12 @@ HError GameDataExtReader::ReadBlock(int /*block_id*/, const String &ext_id,
             _in->Seek(sizeof(int32_t) * 11);
         }
     }
+    else if (ext_id.CompareNoCase("v400_gameopts") == 0)
+    {
+        _ents.Game.faceDirectionRatio = _in->ReadFloat32();
+        // reserve few more 32-bit values (for a total of 10)
+        _in->Seek(sizeof(int32_t) * 9);
+    }
     else
     {
         return new MainGameFileError(kMGFErr_ExtUnknown, String::FromFormat("Type: %s", ext_id.GetCStr()));
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -330,6 +330,31 @@ HError ReadExt399(RoomStruct *room, Stream *in, RoomFileVersion data_ver)
     return HError::None();
 }
 
+// Extended walkable areas and related room properties
+HError ReadExt_400_WalkOpts(RoomStruct *room, Stream *in, RoomFileVersion data_ver)
+{
+    // New room properties
+    room->Options.FaceDirectionRatio = in->ReadFloat32();
+    // reserve few more 32-bit values (for a total of 4)
+    in->ReadInt32();
+    in->ReadInt32();
+    in->ReadInt32();
+
+    size_t wa_count = in->ReadInt32();
+    if (wa_count != room->WalkAreaCount)
+        return new Error(String::FromFormat("Mismatching number of walkable areas: read %zu expected %zu", wa_count, room->WalkAreaCount));
+    for (size_t i = 0; i < wa_count; ++i)
+    {
+        auto &wa = room->WalkAreas[i];
+        wa.FaceDirectionRatio = in->ReadFloat32();
+        // reserve few more 32-bit values (for a total of 4)
+        in->ReadInt32();
+        in->ReadInt32();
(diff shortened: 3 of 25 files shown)


ivan-mogilko:
Added design-time properties for Game (in General Settings), Room and Walkable area.

Two problems:
1. I don't know how to write description brief and clear enough for regular users to understand :).
2. Float properties are parsed depending on Culture, which should not be happening. To clarify: on some systems it will require "2.5" (with a dot) and on others "2,5" (with a comma). Need to figure out how to enforce InvariantCulture in the PropertyGrid.

ivan-mogilko:
After #2307 added a script property to WalkableArea struct too.

ivan-mogilko:
I just realized, that the Editor already has a workaround for floats implemented during serialization, a very long time ago:
https://github.com/adventuregamestudio/ags/blob/b148832100f1ebc3fc826085577c6a4e74b6cbc1/Editor/AGS.Types/SerializeUtils.cs#L143-L152
https://github.com/adventuregamestudio/ags/blob/b148832100f1ebc3fc826085577c6a4e74b6cbc1/Editor/AGS.Types/SerializeUtils.cs#L241-L250

So, these values will save and load using InvariantCulture at least. The only issue remaining is how they are displayed in the property grid, but this may be a lesser priority.

(I was not able to find how to apply the "culture" to a property grid control or particular fields, except for writing a custom typeconverter, which I don't like...

ivan-mogilko:
Following a proposal by @messengerbag, allowed negative ratios, which basically swap up/down loops. This maybe will work for "going downhill" kind of effects in game.

ivan-mogilko:
Added a test game, with assets made by user who requested this feature recently:
https://github.com/adventuregamestudio/ags/files/14254038/test--ags4-directionratio.zip

ivan-mogilko:
In regards to this subtask:
> Optimize direction ration calculation by caching few things in character class.

I decided to leave this out, and only have couple TODOs in code. I found that with the current order of character update checking "area" under the feet will be called about same amount of times anyway; and caching things is often bug prone in AGS, as character state (like position) may be changed in a random script callback. So i'd rather leave this kind of improvement for later.

ivan-mogilko:
Updated test game, now made by blaholtzen, with 2 characters featuring different face ratios:
https://github.com/adventuregamestudio/ags/files/14316758/IsoBikeDemo.zip

PR #2306, opened 2024-01-20
3.6.0: update to the latest SDL_Sound + add py script for easier updating

ivan-mogilko:
This is like #2302 but for 3.6.0.

Added a py script for pushing SDL_Sound revision into each script and readme where necessary.
I thought we might update SDL_Sound for 3.6.0 release as well, since these are mostly bug fixes to decoders, so uped it to latest official release v2.0.2 (previously it was an arbitrary commit from June 2022).


ericoporto:
Hey, you need to remove this line here to fix the pipelines

https://github.com/adventuregamestudio/ags/blob/78464fe34b79bc26ad9f88082da6c7ec59e76d48/CMake/FetchSDL_Sound.cmake#L25

I got rid of it in this commit: https://github.com/adventuregamestudio/ags/commit/1d6acae994a1bff8fec8b7401f78926ed64d4502

The Windows build in GitHub Actions will still fail because it will try to build with C++17 and fail because of the binary/unary functions that you removed in master. I can fix it later by adding something to force it to use C++11.

ivan-mogilko:
Ah, from that commit I can see that there are more instances of revision number which I did not include in the script.

PR #2308, opened 2024-01-21
Update to SDL2 2.28.5 and Add a script for updating SDL revision in Dockerfiles, scripts and cmake files

ericoporto:
This seems to work, just wanted to test if it was possible after #2302

It's possible there is some SDL2 version I forgot somewhere.

One thing this doesn't do, is updating the Android project, which has to be done manually so it's keyboard can be patched.

The added release type is because I made this for myself to test the new SDL2 release (which is in prerelease)

https://github.com/libsdl-org/SDL/releases/tag/prerelease-2.29.1


ivan-mogilko:
I'd rather have this in a single script, which updates all libs. It may use separate function with different parsing for a library

ericoporto:
I tried to make a new script that unifies both. I didn't put a lot of effort to shorten the code, but I think this could be achieved later. I think SDL_sound and SDL2 are the main dependencies to take on now.

PR #2315, opened 2024-01-26
OpenGL: use GL_CLAMP_TO_EDGE with Linear filter

ivan-mogilko:
This supposedly fixes occasional edge lines around sprites, occurring in "render in screen resolution" mode with Linear filter, and certain scaling factors.

A quote from Khronos specs, about a difference between GL_CLAMP and GL_CLAMP_TO_EDGE :

> GL normally clamps such that the texture coordinates are limited to exactly the range [0;1]. When a texture coordinate is clamped using this algorithm, the texture sampling filter straddles the edge of the texture image, taking half its sample values from within the texture image, and the other half from the texture border. It is sometimes desirable to clamp a texture without requiring a border, and without using the constant border color.

> <...> CLAMP_TO_EDGE, clamps texture coordinates at all mipmap levels such that the texture filter never samples a border texel. The color returned when clamping is derived only from texels at the edge of the texture image.

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -36,8 +36,6 @@
 
 #if AGS_OPENGL_ES2
 
-#define GL_CLAMP GL_CLAMP_TO_EDGE
-
 const char* fbo_extension_string = "GL_OES_framebuffer_object";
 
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
User confirmed that this fixes the problem.

PR #2317, opened 2024-01-26
Refactor `our_eip=X` to use `set_our_eip(X)` function

ericoporto:
Hey, just to make it easier to search our codebase when someone reports a `PP=X` in a crash / exception. There was a mix of SET_EIP, `our_eip=X` and `our_eip = X` (varying blank spaces). If all eip changes are guarded in a macro, it's much easier to find using the code search feature of an IDE or `git grep` in the command line.

I am letting it be a draft because the `SET_EIP` macro is in `"gui/guimain.h"` and perhaps it should be in somewhere else, but not sure where.

### DIFF
--- a/Common/ac/common.cpp
+++ b/Common/ac/common.cpp
@@ -15,6 +15,7 @@
 #include "util/string.h"
 
 using namespace AGS::Common;
+int our_eip = 0;
 
 void quit(const String &str)
 {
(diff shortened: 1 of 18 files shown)


ivan-mogilko:
There's a set_our_eip and get_our_eip functions in common.h.

ericoporto:
ok, removed the macro and instead used `set_our_eip()` directly, also picked out `our_eip` usage that was "naked" and guarded too where appropriate using `get_our_eip()`. I also moved the implementation to common instead and out of the font, and also removed the need for a separate implementation in Native.

PR #2324, opened 2024-01-31
Print graphics adapter info + OpenGL video memory detection

ivan-mogilko:
Print available adapter info for Direct3D and OpenGL (this basically means GPU / system driver description).

Print approximate available video memory to the log.

For OpenGL: use ATI / Nvidia extensions if available to get their gpu memory info.

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -684,9 +684,9 @@ void init_draw_method()
         size_t tx_cache_size = usetup.TextureCacheSize * 1024;
         // If graphics driver can report available texture memory,
         // then limit the setting by, let's say, 66% of it (we use it for other things)
-        size_t avail_tx_mem = gfxDriver->GetAvailableTextureMemory();
+        uint64_t avail_tx_mem = gfxDriver->GetAvailableTextureMemory();
         if (avail_tx_mem > 0)
-            tx_cache_size = std::min<size_t>(tx_cache_size, avail_tx_mem * 0.66);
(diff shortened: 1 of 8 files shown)


ericoporto:
In `Engine/gfx/graphicsdriver.h`, in the comment above `GetAvailableTextureMemory`, it would be nice to mention in which unit it's expected to return the amount of memory (I can't tell from reading the code and comments if it's supposed to return in bytes or kilobytes).

ivan-mogilko:
Ok, I fixed these two things.

ericoporto:
I tested this with my Nvidia card and got the information below

```
OpenGL adapter info:
        OpenGL: 4.6.0 NVIDIA 536.23
        Vendor: NVIDIA Corporation
        Renderer: NVIDIA GeForce RTX 2080 SUPER/PCIe/SSE2
OGL: Transparency shader program created successfully
OGL: Tinting shader program created successfully
OGL: Lighting shader program created successfully
Graphics driver set: OpenGL
Graphics mode set: 2560 x 1440 (32-bit) fullscreen desktop
Graphics mode set: refresh rate (optional): 60, vsync: 1
Graphics driver texture memory (approx): 8388608 KB
Render frame set, render dest (0, 0, 2559, 1439 : 2560 x 1440)
Requested gfx filter: linear
Graphics filter set: 'Linear', filter dest (0, 0, 2559, 1439 : 2560 x 1440)
Using gfx filter: Linear
Texture cache set: 131072 KB
```

It's indeed an 8GB video memory card as reported so it looks correct.

PR #2325, opened 2024-02-01
Windows: enable /LARGEADDRESSAWARE for 32-bit builds

ivan-mogilko:
I've been looking into a Dave Gilbert's problem with "out of mem" errors, and it turned out that the texture cache functionality is not behaving as I expected it to. The problem is that engine allocates system memory proportional to video memory allocation when caching textures. It feels as if there were a duplicate of each texture's surface in RAM. It might be that this is normal and I simply don't have enough theoretical knowledge about how textures are stored to judge; but it might also mean that we have something done wrong.
In any case, as an effect the total usage of system memory becomes approximately equal to the sum of sprite cache and texture cache.

Strictly speaking, if we have a large texture cache allocated, then we don't need spritecache at all. But that does not change the fact that creating texture images allocates system memory of comparable amount.

I also noticed strange issue that OpenGL allocates around x1.5-x2 more system memory in comparison with Direct3D.

It does not look like a memory leak though, as the memory usage resets immediately if sprite/texture caches are reset at runtime.

This is quite worrying and has to be investigated. But in the meantime, I'd like to try enabling /LARGEADDRESSAWARE program setting for 32-bit windows builds. This presumably will increase program's address range to 3 or 4 GB, while the program will likely stay compatible with existing 32-bit windows plugins.

Quick test shows that it's enough to run "Old Skies" with (unnecessarily) large caches, with memory usage getting close to 3 GB even. I sent the engine to Dave for a test, and he confirmed that this stopped OOM crashes with OpenGL and 1 GB of texture cache.

So maybe we could use this as a safety workaround for now.

### DIFF
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -622,6 +622,15 @@ if (WIN32)
     set_target_properties(ags PROPERTIES
         WIN32_EXECUTABLE TRUE
     )
+    if (CMAKE_SIZEOF_VOID_P EQUAL 4)
+        # enable more than 2 GB in a 32-bit application
+        if (MINGW)
+            target_link_options(ags PUBLIC "-Wl,--large-address-aware")
+        endif ()
+        if (MSVC)
+            target_link_options(ags PUBLIC "/LARGEADDRESSAWARE")
+        endif ()
+    endif ()
 endif()
 
 # Test
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -1446,7 +1446,7 @@ size_t OGLGraphicsDriver::RenderSpriteBatch(const OGLSpriteBatch &batch, size_t 
         if (e.skip)
             continue;
 
-        switch (reinterpret_cast<intptr_t>(e.ddb))
+        switch (reinterpret_cast<uintptr_t>(e.ddb))
         {
         case DRAWENTRY_STAGECALLBACK:
             // raw-draw plugin support
--- a/Engine/gfx/gfxdriverbase.h
+++ b/Engine/gfx/gfxdriverbase.h
@@ -131,9 +131,9 @@ public:
 
 protected:
     // Special internal values, applied to DrawListEntry
-    static const intptr_t DRAWENTRY_STAGECALLBACK = 0x0;
-    static const intptr_t DRAWENTRY_FADE = 0x1;
-    static const intptr_t DRAWENTRY_TINT = 0x2;
+    static const uintptr_t DRAWENTRY_STAGECALLBACK = 0x0;
+    static const uintptr_t DRAWENTRY_FADE = 0x1;
+    static const uintptr_t DRAWENTRY_TINT = 0x2;
 
     // Called after graphics driver was initialized for use for the first time
     virtual void OnInit();
--- a/Engine/platform/windows/gfx/ali3dd3d.cpp
+++ b/Engine/platform/windows/gfx/ali3dd3d.cpp
@@ -1439,7 +1439,7 @@ size_t D3DGraphicsDriver::RenderSpriteBatch(const D3DSpriteBatch &batch, size_t 
         if (e.skip)
             continue;
 
-        switch (reinterpret_cast<intptr_t>(e.ddb))
+        switch (reinterpret_cast<uintptr_t>(e.ddb))
         {
         case DRAWENTRY_STAGECALLBACK:
             // raw-draw plugin support
--- a/Engine/script/cc_instance.cpp
+++ b/Engine/script/cc_instance.cpp
@@ -1279,13 +1279,13 @@ int ccInstance::Run(int32_t curpc)
             int32_t instId = codeOp.Instruction.InstanceId;
             // determine the offset into the code of the instance we want
             runningInst = loadedInstances[instId];
-            intptr_t callAddr = reg1.PtrU8 - reinterpret_cast<uint8_t*>(&runningInst->code[0]);
-            if (callAddr % sizeof(intptr_t) != 0)
+            uintptr_t callAddr = reg1.PtrU8 - reinterpret_cast<uint8_t*>(&runningInst->code[0]);
+            if (callAddr % sizeof(uintptr_t) != 0)
             {
                 cc_error("call address not aligned");
                 return -1;
             }
-            callAddr /= sizeof(intptr_t); // size of ccScript::code elements
+            callAddr /= sizeof(uintptr_t); // size of ccScript::code elements
 
             if (Run(static_cast<int32_t>(callAddr)))
                 return -1;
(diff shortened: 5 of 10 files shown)


ericoporto:
First, pretty cool this works! I couldn't figure the OoM issues and was trying to work into getting 64-bits working because of this.

One thing I noticed is that in draw for some reason some intermediary bitmaps are created and destroyed every frame, and because they are a considerable range of sizes I think they contribute a lot to memory fragmentation - I couldn't figure any test that proves this. But memory is fragmented enough you get OoM crashes (without large address aware) at around 1.15 GB.

ivan-mogilko:
@ericoporto could you tell, what do I need to change in CMake scripts to make this linker setting into windows project that they create?

> One thing I noticed is that in draw for some reason some intermediary bitmaps are created and destroyed every frame, and because they are a considerable range of sizes I think they contribute a lot to memory fragmentation

It is worth documenting these cases, and open a ticket about them. I suppose cases like that may be resolved by allocating and keeping a bitmap of a larger size and use portions of it through subbitmaps, or plain math.

---

On the topic of cache sizes, I have a thought that current settings are not done well. If it's normal that textures require to have a copy in system memory, maybe the idea is that this serves as a backup, and video memory works as a cache of its own, where only active (recently displayed) textures are stored.

Then, with a texture cache, sprite cache is necessary only for 2 things: GUIs, because they are still drawn partially in software mode, and using sprites for raw drawing. Which means that for the most sprites having textures is enough, and there's no need to have these in sprite cache.

Which leads me to a thought that maybe the settings could be redone. For example, instead of having 2 seemingly separate cache sizes, ags could have a setting of "total sprite cache" and a subsetting dedicated for textures, where the raw sprite cache would only take what's left after texture cache.

ericoporto:
I will look into the largeaddressaware later, but I think there isn't a setting for it native on CMake, so I would need to add it manually with checks for 32-bit and a specific call for MSVC and other for MinGW.

It would be nice to still get 64-bit builds pretty soon, once some way to work with plugins is figured out. Does using largeaddressaware requires the DLLs to be rebuilt with this setting too? I don't understand much how this work in Windows - I think in other OSes the 32-bit build just access the initial 4GB normally.

ivan-mogilko:
>  Does using largeaddressaware requires the DLLs to be rebuilt with this setting too?

According to what I've found on the internet on this topic, /LARGEADDRESSAWARE setting applies to everything loaded into the current process, which means DLLs too. If that's so, then *rebuilding* plugins is not necessary, as all modules will be treated similarly.

OTOH the problem may arise if any of the plugins has incompatible pointer math, such as casting pointers or pointer diffs to int32.

ivan-mogilko:
Hmm, in regards to the potential issue that I mentioned above, thinking more on this, since we have working 64-bit builds, then engine don't have these, otherwise it won't work in 64-bit correctly.
Something that might require checking and fixing are casts to `intptr_t`, they may have to be replaced with `uintptr_t` (but this depends on a context).

ivan-mogilko:
There's another way to significantly reduce the memory use that may be tried (it's a shame that I did not realize earlier).
If there's a ready texture in cache, there's no need to keep original sprite in a raw sprite cache most of the times.
So, the texture cache may be adjusted to exclude the bitmap from the sprite cache when it's loaded for the texture.
The spritecache would still keep sprites requested for gui and raw drawing.

ivan-mogilko:
I have a thought about another memory optimization that may be tried: when a sprite is requested directly as a texture, and it was not loaded before, then skip storing it in raw form a spritecache, and only store it as a texture. That might save alot on regular sprites that are not being used on GUI or in raw drawing (most character and object Views basically).

It's just that there has to be some good rule about when to do or not do that, based on texturecache's size. E.g. allow this if texturecache limit is comparable to spritecache. Because if texturecache is too small, then such behavior may lead to more repeated reloads.

ericoporto:
The power of 2 code, some time ago I found in the internet some known "hack" to speed up finding the next power of 2 and it looked like this

https://gist.github.com/ericoporto/c925c4d1e61bfaa96bc7addbe93dc833

ivan-mogilko:
> The power of 2 code, some time ago I found in the internet some known "hack" to speed up finding the next power of 2 and it looked like this
> 
> https://gist.github.com/ericoporto/c925c4d1e61bfaa96bc7addbe93dc833

Yes, I've seen it too.

ivan-mogilko:
> It turned out that OpenGL has been forcing all textures to the power of 2 size, which is a very old standard restriction.

Ah, but this means that the texture space calculation is wrong, as it was never accounting for this. Which can make texture cache to occupy more space than requested by settings. This has to be fixed too.

ericoporto:
> @ericoporto could you tell, what do I need to change in CMake scripts to make this linker setting into windows project that they create?

Hey, I think this patch below should work! I only tested with MSVC, I don't currently have a MinGW toolchain for 32-bit builds.

```diff
diff --git a/Engine/CMakeLists.txt b/Engine/CMakeLists.txt
index 2ca19a418..2f2ee3c60 100644
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -622,6 +622,15 @@ if (WIN32)
     set_target_properties(ags PROPERTIES
         WIN32_EXECUTABLE TRUE
     )
+    if (CMAKE_SIZEOF_VOID_P EQUAL 4)
+        # enable more than 2 GB in 32 bit application.
+        if (MINGW)
+            target_link_options(ags PUBLIC "-Wl,--large-address-aware")
+        endif()
+        if (MSVC)
+            target_link_options(ags PUBLIC "/LARGEADDRESSAWARE")
+        endif ()
+    endif ()
 endif()

 # Test
```

ivan-mogilko:
Meanwhile, I did another experiment: 5cdc87b4c84624c713544d48f3499e4202b03f8b
Added a rule for the TextureCache, to not store loaded raw sprites in a spritecache, if the capacity of texturecache is close to larger. Because logically in that case there's little or no point at all in keeping a raw version, as when texturecache runs out of space the spritecache also runs out of space, so older sprites will get disposed about same often as older textures.

Hopefully that makes sense. This option may be easily disabled in code (it's one variable).

Normally, I think, if there's a big texturecache, then the spritecache may be left small. But this rule will work as a check in case user config makes both spritecache and texturecache equally big.

Running with this setting (or simply low spritecache) saves ALOT of memory in "Old Skies". OpenGL seems to work much better now, possibly it uses texture compression, but I noticed to get only around 250-300 MB RAM used in the main menu, even if the whole intro animation was cached in the process.

ericoporto:
Hey, I think we will only know if something is wrong with enabling this if it makes in a release and people use it, so far all I can say is it appears to work in the small things I was able to test. From what I understand the largeaddressaware fail in some specific cases where the pointer is treated like an int but it will actually be an unsigned int or something like this. I don't know any specific code analysis to run to catch this.

ivan-mogilko:
I looked around for intptr_t casts, and ones left are two big groups:

1. Memory offsets in scripts. These are *relative* offsets, counting either from bytecode array address, or a dynamic object address. They may be converted to uintptr_t, but that won't have any practical effect, because these values are formed as int32 in the script VM. This is a similar problem in 32-bit and 64-bit engine, and it means that any *compiled script* or any *object in script* cannot be larger than 2 GB.
2. Casts to intptr_t when passing into a function that may accept either an int or a pointer. That is not a problem, because the receiver (called function) is supposed to cast this argument back to pointer before using it. The two cases where this happens are:
    * calling plugin function from script vm
    * calling built-in GUI dialog callbacks: they have a ugly recreation of a WinAPI events system.

ivan-mogilko:
Alright, this seems to be working, I was able to reach 2.5 GB ram usage with the build from CI using higher cache settings (not that it's necessary...)

PR #2326, opened 2024-02-03
Remove long usage in Compiler 

### LINKED ISSUE #2274: Old script compiler: clean up "long" variables
The old compiler works with "long" type alot, its code probably was originally written having 16-bit systems in mind.

"Long" type is non-portable, as it has different sizes on 64-bit Windows and Unix platforms. It's hard to tell whether this is actually causing any problems, but it's better to be safe imo.

This could be related to #2088, but only if Editor's native parts could be built on other platforms/compilers; so perhaps not so much.
On another hand, compiler may be built as a standalone program now.

ericoporto:
fix #2274

I tried to keep changes at minimum.

### DIFF
--- a/Compiler/fmem.h
+++ b/Compiler/fmem.h
@@ -19,16 +19,16 @@
 struct FMEM
 {
   char *data;
-  long len;                     // length of data in array
-  long size;                    // size of data array allocated
-  long magic;
-  long pos;                     // current seeked position
+  size_t len;                  // length of data in array
+  size_t size;                 // size of data array allocated
+  int magic;
+  size_t pos;                  // current seeked position
 };
 
 extern FMEM *fmem_create();
 extern FMEM *fmem_open(const char *);
 extern void fmem_close(FMEM *);
-extern void fmem_write(char *, long, FMEM *);
+extern void fmem_write(char *, size_t, FMEM *);
 extern void fmem_putc(char, FMEM *);
 extern void fmem_puts(char *, FMEM *);
 extern int fmem_getc(FMEM *);
--- a/Compiler/fmem.cpp
+++ b/Compiler/fmem.cpp
@@ -55,8 +55,8 @@ void fmem_close(FMEM*cloo) {
 // of malloc and freeing
 #define BUFFER_INCREMENT_SIZE 2000
 
-void fmem_write(char* towr,long mei,FMEM*fill) {
-  long curdatlen=fill->len;
+void fmem_write(char* towr ,size_t mei, FMEM*fill) {
+  size_t curdatlen=fill->len;
   if (curdatlen + mei >= fill->size) {
     char*datwas=fill->data;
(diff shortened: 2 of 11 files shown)


ericoporto:
Did an additional commit just to kick the CI and see how it goes, but I think that the packfile struct was from the time the system was DOS and 16-bit so I instead switched things there from long to int.

Compiler/fmem.h
```diff
@@ -19,16 +19,16 @@
...
 {
   char *data;
-  long len;                     // length of data in array
-  long size;                    // size of data array allocated
-  long magic;
-  long pos;                     // current seeked position
+  size_t len;                  // length of data in array
+  size_t size;                 // size of data array allocated
+  int magic;
+  int pos;                     // current seeked position
```
ivan-mogilko, line 25: pos also should be size_t, as it should be comparable to size and len fields.

libsrc/allegro/include/allegro/file.h
```diff
@@ -70,9 +70,9 @@ struct PACKFILE_VTABLE
    AL_METHOD(int, pf_fclose, (void *userdata));
    AL_METHOD(int, pf_getc, (void *userdata));
    AL_METHOD(int, pf_ungetc, (int c, void *userdata));
-   AL_METHOD(long, pf_fread, (void *p, long n, void *userdata));
+   AL_METHOD(int, pf_fread, (void *p, int n, void *userdata));
    AL_METHOD(int, pf_putc, (int c, void *userdata));
-   AL_METHOD(long, pf_fwrite, (AL_CONST void *p, long n, void *userdata));
+   AL_METHOD(int, pf_fwrite, (AL_CONST void *p, int n, void *userdata));
```
ivan-mogilko, line 75: It's better to make these size_t, for compliance with classic stream reading and writing.

ericoporto: Switched to size_t!

Compiler/fmem.h
ericoporto: Done

ivan-mogilko:
Sorry, I made a mistake about allegro packfile!

I noticed now that it returns -1 or EOF (which is also -1) on error. This means that the return value is supposed to be signed.
Could you please revert this last commit and leave them as int32?
Otherwise we'd need to also fix return values, and scan all uses of these functions in libraries to see how the return values are checked.

ericoporto:
That is fine, I will revert.

ivan-mogilko:
>  I think the idea would be to return the maximum possible value of size_t?

Standard read and write return strictly number of processed bytes, so in case of EOF or error they should return 0 instead.

ericoporto:
I decide to simply drop that commit (https://github.com/adventuregamestudio/ags/commit/aa77a424cc1886544ec9f9ccc8e755a13dcdafa3) for the time being, as I think it's not relevant now and the long in these files are a good reminder to update that at some point.

PR #2333, opened 2024-02-07
3.6.0: change Editor platform config to Win32

ivan-mogilko:
This is based on commits from #2197 , except no x64 configuration is added, only cleaned up existing Win32 config.

The reason I do this in release-3.6.0 branch is that after #2197, switching between branches in the same workspace may cause wrong platform selected in MSVS, which results in "weird" build errors. Since I still do patches for 3.6.0, I thought to make this branch compatible with recent config changes.


ericoporto:
Hey, is there a button here for you to rerun the building of the Android docker image here? I think the image is gone from cache.

https://cirrus-ci.com/task/5300300806881280

ivan-mogilko:
> Also I think you mentioned that there's something wrong in the Android build from the Editor, if you can create an issue just so I don't forget to look into it - doesn't need to detail it.

No, I don't recall saying that. I was refering to the Android not building in CI for release-3.6.0 branch.

ivan-mogilko:
Ok, finally, everything is building.

ericoporto:
Nice it worked! I think I misunderstood then.

The docker caches failing is a mix of a bug in CirrusCI and Google cloud I think. I think anytime it mysteriously doesn't work we should rebuild the cache - I don't have access for doing this in ags repository, so if I notice it I will tell you.

PR #2334, opened 2024-02-11
Windows: don't link amstrmid and quartz libs, no longer used

ivan-mogilko:
These were likely related to Windows-only functionality cut in AGS 3.6.0.

I only just noticed these while checking the video-related flags in Cmake script.


ericoporto:
On master branch, these are gone from the CMake build but they are still in the Visual Studio vcxproj project.

PR #2336, opened 2024-02-18
ci: use ags4 specific auto-test game

ericoporto:
ags4 specific variant of  #2335

It would be nice to get a new release of ags4 because I think the specifics of the global arrays (inventory, character, ...) changed.


ericoporto:
To be clear, the ags4 tests for global arrays will fail until a new ags4 release is made, as I believe this is fixed in ags4 branch.

ivan-mogilko:
What exactly is the problem with global arrays? I expected that the scripts that use them should still compile.

Issue #2155, opened 2023-09-29
Video playback API overhaul


ivan-mogilko:
### Problem

Video playback API in AGS is rudimentary, it only allows to play a single video, blocking whole game, and nothing else may be updated nor drawn on screen while this video is playing. While this is not the most popular kind of request, there are still users who would like to be able to make proper FMV-style games with AGS, or at least have more control over cinematic cutscenes.
At the moment this may only be achieved using a plugin, but using plugins is always a last resort, as they don't always have full sync with what engine does (think of audio), and one has to ensure that the plugin is built for all wanted platforms.

Long story short, there's a certain demand to have a better video playback API. The basic requirements are:
* be able to run video while the rest of the game is running too;
* be able to render video on an arbitrary game object (gui, overlay, room object);
* be able to run multiple videos at once;
* and all the above should support playing video's audio track too.

We had a discussion about this on forums a while back:
https://www.adventuregamestudio.co.uk/forums/engine-development/feature-request-videos-non-blocking-playback-with-guis/

Following is based on my comment there, but this is a draft idea, so may be discussed and rethought further.

### Proposal

In regards to the Script API:

1. There should be a kind of `VideoPlayback` struct in script. `VideoPlayback` object is returned from `PlayVideo` (or similar) function, and provides means for controlling video state, including but not necessarily limited to: Pause, Resume, Seek functions, Length and Position properties.
2. `VideoPlayback` does not draw anything on screen itself. Instead it has a `Graphic` property, that lets access a dynamic sprite, created and "owned" by the Playback object, where the video frames are decoded to. This property may then be used to draw or assign this sprite anywhere, in a same way as any other sprite is used in the engine.
3. For Audio support, `VideoPlayback` should have a `AudioChannel*` property, which lets to access a channel where video's audio track is being played. NOTE: if we modify audio API prior to this, then this should be changed accordingly.
**EDIT:** later I had strong doubts about using `AudioChannel` for this purpose, as that gives audio playback controls, which will overcomplicate things. Please refer to my comment below: https://github.com/adventuregamestudio/ags/issues/2155#issuecomment-1869161654

In regards to the underlying implementation:

1. The decoding must happen on a separate thread to not unnecessarily load up or block the game's thread. It is important to remember that the video playback may work with its own **fps setting**, not necessarily matching the game's fps. So a frame may theoretically be decoded once per a number of game updates, or multiple times per a single game update.
[...]

PR #2338, opened 2024-02-20
AGS 4: non-blocking video playback

ivan-mogilko:
For #2155.

This is not a complete functionality, but a draft, meant to test out the idea. Random mistakes are to be expected.

1. Rewrote VideoPlayer class, so that it buffers video frames and prepares a "ready frame" when the time is right, but does not draw anything itself. The plan for the VideoPlayer is that it can work in 2 modes: autoplay and frame-by-frame mode. In the first mode it calculates timing when the next frame should be presented, and switches the "ready frame". In the frame-by-frame mode it will be in paused state but advance to next frame by a call to NextFrame(). Audio is not played in the frame-by-frame mode.
2. Picked out BlockingVideoPlayer class which is basically a "game state" singleton that runs VideoPlayer, blocking the rest of the game. This is used for the old PlayVideo command.
3. Implemented "video core", similar to existing "audio core", with a running thread, which polls video players.
4. Wrote a video object control, that starts and stops videos, and acquires their "ready frames", placing them into the spriteset, as dynamic sprites (so that they don't get deleted). Besides being video frames, these act entirely like any sprite, so may be assigned anywhere, e.g. to Overlay. When frame is changed, game objects are notified of a sprite's change as usual, and redraw themselves as necessary.
5. Implemented new VideoPlayer API as follows:
```
enum PlaybackState {
  ePlaybackOn = 2,
  ePlaybackPaused = 3,
  ePlaybackStopped = 4
};

builtin managed struct VideoPlayer {
  import static VideoPlayer* Open(const string filename, bool autoPlay=true, RepeatStyle=eOnce);
  /// Starts or resumes the playback.
  import void Play();
  /// Pauses the playback.
  import void Pause();
  /// Advances video by 1 frame, will pause video when called.
  import void NextFrame();
  /// Changes playback to continue from the specified frame.
  import void SeekFrame(int frame);
  /// Changes playback to continue from the specified position in milliseconds.
  import void SeekMs(int position);
  /// Stops the video completely.
  import void Stop();

  /// Gets current frame index.
  import readonly attribute int Frame;
  /// Gets total number of frames in this video.
  import readonly attribute int FrameCount;
  /// Gets this video's framerate (number of frames per second).
  import readonly attribute float FrameRate;
  /// Gets the number of sprite this video renders to.
  import readonly attribute int Graphic;
  /// The length of the currently playing video, in milliseconds.
  import readonly attribute int LengthMs;
  /// Gets/sets whether the video should loop.
  import attribute bool Looping;
  /// The current playback position, in milliseconds.
  import readonly attribute int PositionMs;
  /// The speed of playing (1.0 is default).
  import attribute float Speed;
  /// Gets the current playback state (playing, paused, etc).
  import readonly attribute PlaybackState State;
  /// The volume of this video's sound, from 0 to 100.
  import attribute int Volume;
};
```

A key point is a Graphic property that returns an index of a sprite where the video renders to. This sprite may then be assigned to anything that may normally have a sprite, such as Overlay, GUI button, etc.

Basic script example may be:
```
VideoPlayer *video = VideoPlayer.Open("video.ogv", true, eRepeat);
if (video)
{
    Overlay *over = Overlay.CreateGraphical(0, 0, video.Graphic);
    Wait(100);
    over = null;
    video = null;
}
```

---

TODO:
- [x] Proper script API
- [x] Fix timing after resuming
- [ ] Autofix audio/video desync
- [x] Support speed setting
- [ ] Support seek, APEG does not provide this!
- [x] Stubs for AGS_NO_VIDEO_PLAYER build (?)

POTENTIAL OPTIMIZATIONS
- [ ] Buffer frame textures along with the bitmaps
- [ ] Store "opaque" flag per sprite, so that updating sprite's texture would be a bit faster. This may be still done separately, as a general improvement, if "buffer texture" path is taken.
- [ ] Might have separate locks for buffering and processing/control, as buffering only touches the queue and frame pool, and might as well keep running while the ready frame is being retrieved (or rather, we should let retrieve the ready frame while the decoding and buffering is run). This may be done in "audio core" too, but with video this is more important, as video frames take more time to process.

OTHER THINGS:
- [ ] Consider finding/writing a new library code for working with ogg theora, because APEG is very old, and does not have all required functionality prepared.

### DIFF
--- a/Engine/media/video/video.cpp
+++ b/Engine/media/video/video.cpp
@@ -16,15 +16,21 @@
 #ifndef AGS_NO_VIDEO_PLAYER
 #include "core/assetmanager.h"
 #include "ac/draw.h"
+#include "ac/dynamicsprite.h"
+#include "ac/game.h"
 #include "ac/game.h"
 #include "ac/gamesetupstruct.h"
 #include "ac/gamestate.h"
 #include "ac/global_audio.h"
 #include "ac/joystick.h"
+#include "ac/spritecache.h"
 #include "ac/sys_events.h"
+#include "ac/dynobj/dynobj_manager.h"
+#include "ac/dynobj/scriptvideoplayer.h"
 #include "debug/debug_log.h"
 #include "gfx/graphicsdriver.h"
 #include "main/game_run.h"
+#include "media/video/video_core.h"
 #include "media/audio/audio.h"
 #include "media/video/video_core.h"
 #include "util/memory_compat.h"
@@ -279,7 +285,8 @@ std::unique_ptr<BlockingVideoPlayer> gl_Video;
 
 
 //-----------------------------------------------------------------------------
-// Blocking video API
+// Legacy Blocking video API
+//-----------------------------------------------------------------------------
 // Running the single video playback
 //-----------------------------------------------------------------------------
 // Checks input events, tells if the video should be skipped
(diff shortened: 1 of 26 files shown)


ericoporto:
I know this is ugly but it was the best place I found at the time to throw the polling for audio

https://github.com/adventuregamestudio/ags/blob/84ab2b592f5611b22380ed52a3b42080c9c68487/Engine/ac/timer.cpp#L72-L74

The video thread will require polling there too as we can't use threads in the Emscripten port currently.

ericoporto:
The CI failure in Android just looks like it failed to download the SDL sources it uses when building it, maybe just rerunning that task should make it complete now.

ivan-mogilko:
Pushed a proper Video API (still a draft though), which now looks like this:

```
enum PlaybackState {
  ePlaybackOn = 2,
  ePlaybackPaused = 3,
  ePlaybackStopped = 4
};

builtin managed struct VideoPlayer {
  import static VideoPlayer* Open(const string filename, bool autoPlay=true, RepeatStyle=eOnce);
  /// Starts or resumes the playback.
  import void Play();
  /// Pauses the playback.
  import void Pause();
  /// Advances video by 1 frame, will pause video when called.
  import void NextFrame();
  /// Changes playback to continue from the specified frame.
  import void SeekFrame(int frame);
  /// Changes playback to continue from the specified position in milliseconds.
  import void SeekMs(int position);
  /// Stops the video completely.
  import void Stop();

  /// Gets current frame index.
  import readonly attribute int Frame;
  /// Gets total number of frames in this video.
  import readonly attribute int FrameCount;
  /// Gets this video's framerate (number of frames per second).
  import readonly attribute float FrameRate;
  /// Gets the number of sprite this video renders to.
  import readonly attribute int Graphic;
  /// The length of the currently playing video, in milliseconds.
  import readonly attribute int LengthMs;
  /// Gets/sets whether the video should loop.
  import attribute bool Looping;
  /// The current playback position, in milliseconds.
  import readonly attribute int PositionMs;
  /// The speed of playing (1.0 is default).
  import attribute float Speed;
  /// Gets the current playback state (playing, paused, etc).
  import readonly attribute PlaybackState State;
  /// The volume of this video's sound, from 0 to 100.
  import attribute int Volume;
};
#endif
```

Most of the functionality works, except Seek and few less important getters.
Unfortunately APEG library does not support Seek, so either this has to be written, or a completely new library found/written for working with ogg theora decoder.

ericoporto:
Uhm, maybe add a CanSeek in the video API for now and return false? Is there a chance we sometime use a new stream class type that can't seek?

ivan-mogilko:
Rebased, squashed commits.

Made Seek return new position or error value. Seek(0) now works as a rewind so long as stream may be reused.

Engine/media/video/video.cpp
```diff
@@ -228,17 +224,16 @@ bool BlockingVideoPlayer::Run()
...
-
-    // update/render next frame
     std::unique_ptr<Bitmap> frame;
     {
-        std::lock_guard<std::mutex> lk(_videoMutex);
-        _playbackState = _player->GetPlayState();
-        frame = _player->GetReadyFrame();
+        auto player = video_core_get_player(_playerID);
+#if defined(AGS_DISABLE_THREADS)
+        if (!_player->Poll())
```
ericoporto, line 231: This is a typo, it should be `player` here.

ivan-mogilko:
Hmm, actually I removed the video polling inside BlockingVideoPlayer, because it's polled in WaitForNextFrame in case of "no threads" build.

PR #2346, opened 2024-03-05
Emscripten: upgrade to 3.1.55

ericoporto:
- This upgrades the Emscripten SDL2 port to 2.28.4
- We also need to upgrade node version because emscripten updated it
- Adjusted docker file for bug in `--lto` handing in embuilder in 3.1.55

### DIFF
--- a/Emscripten/install_dependencies.sh
+++ b/Emscripten/install_dependencies.sh
@@ -7,8 +7,8 @@ mkdir emscripten
 pushd emscripten
 git clone https://github.com/emscripten-core/emsdk.git
 pushd emsdk
-./emsdk install 3.1.28
-./emsdk activate 3.1.28
+./emsdk install 3.1.55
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
> This upgrades the Emscripten SDL2 port to 2.28.4

3.6.1 is using 2.28.5, does it need this update too?

ericoporto:
I will get there eventually, I just like using the Emscripten port instead of building from source because this forces more tests to run and more people to review it before upgrading. But other than I think two or three gamepads there's not much difference between 2.28.4 and 2.28.5. I unfortunately had mostly tested thoroughly 2.28.4 in Emscripten.

ericoporto:
Argh, sorry brain farted a bit and thought the 3 whatever was the Emscripten and not ags version... Yeah, this is safe to apply to 3.6.1.

PR #2347, opened 2024-03-05
Engine: allow disabling exclusive fullscreen checkbox

ericoporto:
if you have `fullscreen=full_window`, using 

```
[disabled]
fullscreen=1
```

Will lock it. This is from my own suggestion here for the Direct3D alt+tab issue

https://github.com/adventuregamestudio/ags/issues/1287#issuecomment-1894818412

Now, when I built ags here and tried to trigger the alt+tab hanging issue I could not cause it anymore - not sure if something magically fixed it in either my computer or ags, will have to experiment more later.


ivan-mogilko:
> Now, when I built ags here and tried to trigger the alt+tab hanging issue I could not cause it anymore - not sure if something magically fixed it in either my computer or ags, will have to experiment more later.

Is this behavior consistent with release-3.6.1 branch too, or only changed in master branch?

PR #2348, opened 2024-03-06
Separate VideoPlayer and BlockingVideoPlayer game state classes

ivan-mogilko:
This is a limited backport from #2338, containing only improved VideoPlayer class, but no changes to script API, and no global working thread for non-blocking videos. Its main purpose is to keep code in branches at least somewhat consistent, but also make it easier to do improvements to video playback.

* Separated VideoPlayer and BlockingVideoPlayer classes, where VideoPlayer is a generic player, while BlockingVideoPlayer defines a game state which runs VideoPlayer in particular way.
* Turned BlockingVideoPlayer as another GameState implementation (complementary to the recent #2332).
* VideoPlayer now has buffer queue for video frames, supports both autoplay and manual frame-by-frame work modes.
* No actual working thread is implemented in this PR, but it may be done later, to improve playing videos with higher framerates.

There's a potential issue with FLIC videos that I found. It looks like existing games that use FLIC has videos coded with 100 fps (!), but it's played at 60 fps as a effect from setting vsync in script, and that's how it matches the audio (which is played separately). This makes me concerned about maintaining compatibility with these games in case if a video will be run on a working thread on its true fps rate. I suppose this could be worked around by running the FLICs in manual mode using NextFrame() call from the main thread, rather than starting it with Play(). This is possible because FLICs don't have any audio in them.
**EDIT:** fixed this, but used an easier method.

### DIFF
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -400,8 +400,14 @@ target_sources(engine
     media/audio/sound.h
     media/audio/soundclip.cpp
     media/audio/soundclip.h
+    media/video/flic_player.cpp
+    media/video/flic_player.h
+    media/video/theora_player.cpp
+    media/video/theora_player.h
     media/video/video.cpp
     media/video/video.h
+    media/video/videoplayer.cpp
+    media/video/videoplayer.h
     platform/base/agsplatformdriver.cpp
     platform/base/agsplatformdriver.h
     platform/base/agsplatform_xdg_unix.cpp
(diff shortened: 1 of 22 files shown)


ivan-mogilko:
Ouf, I had to change few things in VideoPlayer compared to #2338, after finding some logical mistakes. But it seems working now.
I will need to update #2338 accordingly, maybe after merging this pr to master and ags4 branch, and rebasing 2338.

PR #2353, opened 2024-03-09
Implemented a RAII wrapper over IUnknown, for Direct3D

### LINKED ISSUE #2343: Smart pointer for Direct3D interfaces
Direct3D interface requires objects to have a Release call. Copying an object pointer normally should have a matching pair of AddRef/Release calls. A missing Release may cause objects stacking up in memory, but also errors when resetting device (e.g. when changing display mode).

Each of those interfaces is derived from IUnknown, a base interface of Windows COM system, and AddRef/Release methods are part of IUnknown.

1. Implement a custom smart pointer template for IUnknown* and children, which calls AddRef on **copy assignment** and Release on deletion (but not `delete`!). The move assignment should not do AddRef, obviously. Possibly we'll need to have two variants, one mimicking std::unique_ptr (with no copy operator) and another std::shared_ptr. I'm not entirely certain if std smart pointers may be utilized for this, but *maybe* it's possible to do through the use of helper functions that AddRef on copy. It's just preferred to have this guaranteed somehow, for safety.
2. Use these smart pointers throughout the D3DGraphicsDriver class.

ivan-mogilko:
Resolves #2343.

An automatically releasing pointer for IDirect3D objects.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -152,6 +152,7 @@ target_sources(common
     util/path.h
     util/resourcecache.h
     util/scaling.h
+    util/smart_ptr.h
     util/stdio_compat.c
     util/stdio_compat.h
     util/stream.cpp
--- a/Common/util/smart_ptr.h
+++ b/Common/util/smart_ptr.h
@@ -1,0 +1,178 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-2024 various contributors
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// https://opensource.org/license/artistic-2-0/
+//
+//=============================================================================
+//
+// Custom smart pointer types and helpers.
+//
+//=============================================================================
+#ifndef __AGS_CN_UTIL__SMARTPTR_H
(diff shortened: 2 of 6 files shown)


ericoporto:
I think the code for IUnknownPtr needs to be wrapped in a `#if AGS_PLATFORM_OS_WINDOWS` clause

ivan-mogilko:
> I think the code for IUnknownPtr needs to be wrapped in a #if AGS_PLATFORM_OS_WINDOWS clause

It has zero dependency on windows headers, and may work with anything that has AddRef/Release methods.

But I have second thoughts about this approach, will revise this later.

ivan-mogilko:
Hmm, maybe this is still a not so bad approach.
I found that it's not much different to a variant found in this MS dev article, and they even have a similar method of retrieving internal pointer field address for assignment (their function is called "GetAddressOf", while mine is "Acquire"):
https://learn.microsoft.com/en-us/archive/msdn-magazine/2015/february/windows-with-c-com-smart-pointers-revisited

I still need to reorganize this a bit, and I found there's some bug, occuring when resizing a window (but strangely, not when changing from windowed to fullscreen). I will have to carefully recheck all usages of the new ptr type.

ericoporto:
I  threw a bunch of games on this and they all seem to work.

Small things in `smart_ptr.h`.

- You can reduce `ComPtr<TInterface>` to only `ComPtr`, it should be the same.
-  in `uint64_t ref_cnt = _obj->Release()`, `ref_cnt` can be made `const`

ericoporto:
Oh, I meant just replacing. The `<TInterface>` next to `ComPtr` could omitted.

It can be left there too if it makes things clearer.

PR #2355, opened 2024-03-11
3.6.1: remove "supersampling" setting from OpenGL renderer

ivan-mogilko:
This "supersampling" setting comes from the original mobile port of AGS, but it's not clear whether it was useful. My understanding is that it increases resolution in which game sprites are drawn to a fixed multiplier of 2, but there's already is a "Render in screen resolution" setting that does the similar thing.

More importantly, there are issues with this:
1. It was never implemented in another hardware accelerated driver (Direct3D), being exclusive to OpenGL.
2. It was never added to standard setup, so not very visible.
3. It's not clear whether it was working properly all this time. There have been multiple changes to graphic renderers, and some of them may assume that rendering at "native resolution" is done at exactly, well, native resolution.

UPDATE:
Also removed unused "video" "framedrop" setting.

### DIFF
--- a/Android/README.md
+++ b/Android/README.md
@@ -144,14 +144,12 @@ The preferences modifies the `android.cfg` file, either the local or global one.
 
 #### Graphics `[graphics]`
 
-- **Renderer** : Software (SDL2) or Hardware (OpenGL ES2)  
-  `renderer = 0, 1`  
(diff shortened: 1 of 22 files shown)


ericoporto:
I noticed the CI didn't pickup neither the version commit after the merge or the new tag. On GitHub Status there's an existing incident going on with pull requests.

I emailed CirrusCI support in the meantime.

Issue #2352, opened 2024-03-08
Editor: color properties are clamping RGB values inconsistenly with the engine


ivan-mogilko:
**Problem**

Because in AGS color index <-> RGB mapping is done based on 16-bit RGB values, both Editor and Engine fixup any assigned RGB value by clamping its components to R5G6B5 format (i think). (On a side note, there's a task to remove this clamping in ags4, see #1980)

However, it seems that Editor's mapping is different from the Engine's.
Sometimes the RGB value typed into a property field is clamped to one result, but using `Game.GetColorFromRGB` in script produces a different result.

For example:
Input value: 167, 181, 147
Editor clamps to: 160, 180, 144
Engine clamps to: 165, 182, 148

UPDATE: user reports there's another inconsistency, when using Colours Pane for making color numbers and then trying to insert them into the color number properties. Following is user's explanation:

> I plugged in 73, 106, 69 into the colour selector in the colour panel
it spat out a decimal number
I plugged in that number into the speechview colour
it reclamped it to 72, 104, 64
but if I go to the Colour Panel and tap the FIND COLOUR next to the generated number of 19272
it shows 73, 106, 69

EDIT: in regards to this addition, maybe it's a mistake in user's report, the result of converting 19272 back in the Color Picker will be 72, 104, 64
In other words,
input 73, 106, 69 ==> number is 19272
input 19272  ==> RGB is 72, 104, 64
You should reset and input the number again to see this.

**AGS Version**

Any 3.* version, I think.

**Expected behavior**

I believe that the Editor should comply to the engine's formula when clamping colors.

Issue #1980, opened 2023-04-16
Support true 32-bit colors in game data and script API


ivan-mogilko:
### Problem

Historically AGS stores a script color value in 16-bit format (iirc it is R5-G6-B5) in both 16-bit and 32-bit games. This causes obvious problems in 32-bit games:
* you do not have access to a full range of colors when selecting a color in editor or script,
* you do not have access to the alpha channel.

Besides that, colors with values 0-31 (this corresponds to lower blue hues in ARGB format) have special meaning: they are forced to refer to the palette indices 0-31, even in 32-bit games. The reason why this is done is, AFAIK, because certain hardcoded graphics in the engine use these palette slots to get drawn.

EDIT: upon quick check of the code, it's not the real game palette where these colors are taken from, but the special 32-slot hardcoded palette. So any changes to the 256-colors game palette do not affect these.
This also means that these 0-31 color indices are not "dynamic", they are always resolving to the same RGB values

### Proposed changes

In brief, the goal is this:
1. The color values in game properties and script variables are treated either as full 32-bit color values (ARGB8888) or 8-bit indexes depending on the context. If there's any conversion necessary to the game's native color depth (like 16-bit), this conversion is performed in the code as close to drawing as possible, and result stored in internal or temporary variables.
2. If 16-bit is still supported, the Editor may provide a help in finding closest supported RGB color when editing color properties.
3. 0-31 color values are no longer locked to the palette indices in anything beyond 8-bit games. If we need hardcoded color for any purpose, then in 16/32 bit games the RGB should be hardcoded instead, and assigned in accordance to the current color depth.
4. In 32-bit games the alpha channel has a meaning. It may be set, and affects the drawing operations. NOTE: this might require an additional work to implement consistently, so may be picked out into separate tickets.

Further notes:
1. `Game.GetColorFromRGB` is superceded or accompanied with `Game.GetColorFromRGBA(r, g, b, a)`
2. To clarify: any `int color` value in script API in 32-bit games should reliably be A8R8G8B8, meaning that users may also just make color themselves using bitwise operations. This must be documented in the manual (same for 16-bit games, if it's not already).
3. Must look out for any hardcoded color values set in the engine. Not sure what to suggest here to ease the search. These has to be replaced with something like `GetHardcodedRGB(index)` (or better name).
EDIT: good start is to search for GetCompatibleColor() calls which have hardcoded numbers as arguments.
[...]

PR #2356, opened 2024-03-11
AGS4: Extend Allegro palette to 0-255

AlanDrake:
This is a WIP attempt to increase Allegro's internal palette from 0-63 to 0-255.

EDIT2: I managed to fix the bestfit algorithm, but I'm still converting to 6bit for the rbg_map

I upgraded some pieces from the ancient demo quest, which can be used as a test project.
[DemoQuest_ags4.zip](https://github.com/adventuregamestudio/ags/files/14556155/DemoQuest_ags4.zip)

You can also import other 256 games, but they will need a full rebuild to update the palettes in game data, and manual palette manipulation that were previously restricted to 0-63 will obviously work differently.


ericoporto:
I am curious about this being rebased after the recent changes.

AlanDrake:
Alright, I rebased and revised this PR.
I made `create_color_table` and `bestfit_color` create a temporary 6bit palette for compatiblity with their uber haxx algorithms, which I'm not good enough to upgrade.

Should work fine both for antialiased sprites on and off.
I didn't test BMP/PCX/FLI load/save/display.

ivan-mogilko:
> I upgraded some pieces from the ancient demo quest, which can be used as a test project.

Note, I added "test-4.0.0" branch to the old demo repository:
https://github.com/adventuregamestudio/ags-demo-quest-old/tree/test--4.0.0

Editor/AGS.Editor/ImportExport.cs
```diff
@@ -694,9 +694,9 @@ public static void ExportCharacter272(Character character, string fileName, Game
             writer.Write((int)6);
             for (int i = 0; i < 256; i++)
             {
-                writer.Write((byte)(game.Palette[i].Colour.R / 4));
-                writer.Write((byte)(game.Palette[i].Colour.G / 4));
-                writer.Write((byte)(game.Palette[i].Colour.B / 4));
+                writer.Write((byte)(game.Palette[i].Colour.R));
+                writer.Write((byte)(game.Palette[i].Colour.G));
+                writer.Write((byte)(game.Palette[i].Colour.B));
```
ivan-mogilko, line 699: You have adjusted palette fixups in ExportCharacter272, but not ImportCharacter272. Is that intentional?

Personally, I believe that these two must not be adjusted, because they relate to a old data format (and by changing palette range you are changing format).
Either keep these as-is (maybe with some commentary), or remove both functions completely as obsolete.

Engine/ac/room.cpp
```diff
@@ -603,12 +603,12 @@ void load_new_room(int newnum, CharacterInfo*forchar) {
...
-            if (palette[ff].g > 63)
-                palette[ff].g = 63;
-            if (palette[ff].b > 63)
-                palette[ff].b = 63;
+            if (palette[ff].r > 255)
+                palette[ff].r = 255;
+            if (palette[ff].g > 255)
+                palette[ff].g = 255;
+            if (palette[ff].b > 255)
+                palette[ff].b = 255;
```
ivan-mogilko, line 611: This code chunk may be removed completely, because palette components are declared as `unsigned char` and can never be > 255.

libsrc/allegro/src/allegro.c
```diff
@@ -75,84 +75,14 @@ PALETTE _current_palette;
...
 {
-   { 63, 63, 63, 0 },   { 63, 0,  0,  0 },   { 0,  63, 0,  0 },   { 63, 63, 0,  0 },
-   { 0,  0,  63, 0 },   { 63, 0,  63, 0 },   { 0,  63, 63, 0 },   { 16, 16, 16, 0 },
-   { 31, 31, 31, 0 },   { 63, 31, 31, 0 },   { 31, 63, 31, 0 },   { 63, 63, 31, 0 },
-   { 31, 31, 63, 0 },   { 63, 31, 63, 0 },   { 31, 63, 63, 0 },   { 0,  0,  0,  0 }
+  { 255, 255, 255, 0 },{ 255, 0,  0,  0 },{ 0,  255, 0,  0 },{ 255, 255, 0,  0 },
+  { 0,  0,  255, 0 },{ 255, 0,  255, 0 },{ 0,  255, 255, 0 },{ 65, 65, 65, 0 },
+  { 125, 125, 125, 0 },{ 255, 125, 125, 0 },{ 125, 255, 125, 0 },{ 255, 255, 125, 0 },
+  { 125, 125, 255, 0 },{ 255, 125, 255, 0 },{ 125, 255, 255, 0 },{ 0,  0,  0,  0 }
 };
```
ivan-mogilko, line 84: Apparently, `desktop_palette` may also be removed, as it's not used anywhere in our stripped allegro sources.

ivan-mogilko:
Besides the in-code comments above,

1. Game data version (GameDataVersion enum) should be raised, because palette unit range has changed. Project version may be left, I think, as palette was saved in 256-unit range there.
2. I am bothered by temporary palette in `bestfit_color`, because that function is run per pixel, meaning this temporary palette will also be recreated over and over again for each processed pixel. We need to find way to optimize this, such as create a fixed palette beforehand and pass into this function instead of original palette.
3. I am also bothered by the potential use of `rgb_map` in `fixup_loaded_bitmap`, which I do not fully understand. But it looks like the engine only goes this path if a **hi/true-color** image is being converted **to 8-bit** image, which is unlikely to happen in practice.
4. Need to test FLIC playback, to see if its looks changed (but I guess the worst that may happen is that there will be more color variation?)

AlanDrake:
Good news.

Thanks to divine inspiration I managed to fix the bestfit algorithms for real.
Not sure if formally correct, but I finally found the weights and values to make it looks like it's working without glaring colro mismatches.

Now sprites can finally match the 252-gray that used to become full white.

The rgb_table is still being wrapped to a 6bit palette. I don't have time for figuring that one. I duped the bestfit_init, so it can still generate the weights it needs.

> 1. Game data version (GameDataVersion enum) should be raised, because palette unit range has changed. Project version may be left, I think, as palette was saved in 256-unit range there.
>
> 2. For a proper behavior, there should be a palette upgrade from 64-range to 256-range on game load in the engine (UpdateGameData) if loaded a game of lower version.
>
> 3. "Game State" save component version should be raised, for the same reasons, as palette is written there. And palette upgraded on loading an older save.

Seems a waste of time, since the format is essentially unchanged and we only need a full rebuild to get everything on track.
Savestates should be expected to be unsafe on a alpha build.

> 4. I am bothered by temporary palette in `bestfit_color`, because that function is run per pixel, meaning this temporary palette will also be recreated over and over again for each processed pixel. We need to find way to optimize this, such as create a fixed palette beforehand and pass into this function instead of original palette.
 
Sounds good, but in my opinion supporting 8bit mode in the way it's implemented is a waste of time, because shaders could do all these transformations in more convenient ways.

> 5. I am also bothered by the potential use of `rgb_map` in `fixup_loaded_bitmap`, which I do not fully understand. But it looks like the engine only goes this path if a **hi/true-color** image is being converted **to 8-bit** image, which is unlikely to happen in practice.
 
I suppose that's what the rgb_table is supposed to optimize. In fact there are several sections where the table overtakes the individual bestfit, if initialized.

> 6. There's also BestFitColor in PaletteUtilities.cs, which i ported to C# for remapping backgrounds. Likely it should have same fix?

I managed to fix both to full 0-255 color matching.

> 8. There's still this statement in the PR's description:

Fixed

ivan-mogilko:
> Seems a waste of time, since the format is essentially unchanged and we only need a full rebuild to get everything on track.

Format is changed, because you change the palette units from 64-based to 256-based.
This will make loading older data incorrect.
You may ignore loading older files, but at least the version should be incremented following the change in format.

Editor/AGS.Editor/Utils/PaletteUtilities.cs
```diff
@@ -233,20 +232,20 @@ private static void RemapPixels(byte[] pixels, Color[] oldPal, Color[] newPal, b
         }
 
         // 1.5k lookup table for color matching
-        private static uint[] ColDiffTable = new uint[3 * 128];
+        private static uint[] ColDiffTable = new uint[3 * 512];
```
ivan-mogilko, line 235: The above comment does not match the table now, please fix it, or remove the comment.

libsrc/allegro/src/color.c
```diff
@@ -261,7 +261,7 @@ int geta(int c)
 
 
 /* 1.5k lookup table for color matching */
-static unsigned int col_diff[3*128]; 
+static unsigned int col_diff[3*512];
```
ivan-mogilko, line 264: Same here, the comment no longer matches the table.

Editor/AGS.Editor/Utils/PaletteUtilities.cs
ivan-mogilko, line 235: There's a comment about "1.5k table", now it does not reflect the table size. Should be either fixed, or removed.

libsrc/allegro/src/color.c
ivan-mogilko, line 264: Also comment about 1.5k table, same as above.

ivan-mogilko:
Tested the palette remap, it seems to work fine (results are equal or close to the existing version).

ivan-mogilko:
Somehow I completely missed this, but this PR does 2 changes that are not related to the palette in any way, and instead change regular color properties from 16-bit R6G5B6 to 32-bit R8G8B8, which refers relates to task #1980. But since these changes are not complemented by others, on their own they break some functionality in both Editor and Engine.

Noteably these are changes to ColorMapper and `__my_setcolor` function in AGS.Native.

For example, same "Color" property numbers now produce a different color, so loading existing project suddenly make colored things look different. And because `__my_setcolor` was only modifed in AGS.Native, but not Engine, they will also look different at runtime compared to the Editor.

Note: I've already reverted the `__my_setcolor` back during small recent refactor, now this function is shared, but ColorMapper still has this change. I might just keep this, since I started working on #1980 now. But I thought that should note this here just in case.

ericoporto:
Uhm, is this also related to https://github.com/adventuregamestudio/ags/issues/2352 somehow?

ivan-mogilko:
> Uhm, is this also related to #2352 somehow?

This may be the related place in code, but the change itself is not related, because it was done in ags4, while the mistake is found in 3.* branch.

PR #2358, opened 2024-03-16
Script API: implement System.GetEngineInteger() and GetEngineString()

ivan-mogilko:
This implements two script functions: System.GetEngineInteger() and System.GetEngineString(), which return some runtime engine parameter by its ID (and an optional index).

The idea of this is to be able to get a number of predefined types of parameters from the engine in script, meant mainly for diagnostic purposes (logging, debugging, performance testing, etc). Because this purpose is not related to a game's logic, I believe that it does not make sense to litter script API with properties, but instead have just a couple of functions that return a value by id.

This is sort of inspired by similar cases in some popular APIs, such as OpenGL, OpenAL, and some others.
The reasons why I did not want to use strings as IDs is that the list of value IDs for those functions will likely be persistent, and also that strings won't autocomplete in script editor. If there will be a need to support some non-standard values, we could add a new function that would accept a String ID instead.

The functions are declared as members of `System` struct as:
```
/// Gets a runtime engine value represented as integer by the given identifier; is meant for diagnostic purposes only
import static int GetEngineInteger(EngineValueID value, int index = 0);
/// Gets a runtime engine string by the given identifier; is meant for diagnostic purposes only
import static String GetEngineString(EngineValueID value, int index = 0);
```

Where first function returns only parameters that are integers by nature, and the second returns those that are strings and also can return integer parameters formatted in a string. The optional "index" argument is meant for values that are arrays.

Currently supported following value IDs:
```
enum EngineValueID {
  ENGINE_VALUE_UNDEFINED = 0,            // formality...
  ENGINE_VALUE_SI_VALUENAME,             // get engine value's own name, by its index
  ENGINE_VALUE_S_ENGINE_NAME,
  ENGINE_VALUE_S_ENGINE_VERSION,         // N.N.N.N (with an optional custom tag)
  ENGINE_VALUE_S_ENGINE_VERSION_FULL,    // full, with bitness, endianess and any tag list
  ENGINE_VALUE_S_DISPLAY_MODE_STR,
  ENGINE_VALUE_S_GFXRENDERER,
  ENGINE_VALUE_S_GFXFILTER,
  ENGINE_VALUE_I_SPRCACHE_MAXNORMAL,
  ENGINE_VALUE_I_SPRCACHE_NORMAL,
  ENGINE_VALUE_I_SPRCACHE_LOCKED,
  ENGINE_VALUE_I_SPRCACHE_EXTERNAL,
  ENGINE_VALUE_I_TEXCACHE_MAXNORMAL,
  ENGINE_VALUE_I_TEXCACHE_NORMAL,
  ENGINE_VALUE_I_FPS_MAX,
  ENGINE_VALUE_I_FPS,
  ENGINE_VALUE_LAST                      // in case user wants to iterate them
};
```

This enumeration may be expanded easily in the future.
Note that the constant name contains a hint of how this value should be retrieved: I - integer, II - indexed integer, S - string, SI - indexed string.

Quick example of use:
```
// Print all supported engine values as strings
for (int i = ENGINE_VALUE_UNDEFINED + 1; i < ENGINE_VALUE_LAST; i++)
{
    Display("(%d) %s =\n%s", i, System.GetEngineString(ENGINE_VALUE_SI_VALUENAME, i), System.GetEngineString(i));
}
```

A more practical example:
```
lblSpriteCache.Text = String.Format("%d / %d (KB)",
    System.GetEngineInteger(ENGINE_VALUE_I_SPRCACHE_NORMAL),
    System.GetEngineInteger(ENGINE_VALUE_I_SPRCACHE_MAXNORMAL));

lblFPS.Text = String.Format("FPS: %d / %d",
    System.GetEngineInteger(ENGINE_VALUE_I_FPS),
    System.GetEngineInteger(ENGINE_VALUE_I_FPS_MAX));
```

### DIFF
--- a/Engine/ac/system.cpp
+++ b/Engine/ac/system.cpp
@@ -15,19 +15,23 @@
 #include "ac/common.h"
 #include "ac/draw.h"
 #include "ac/dynobj/cc_audiochannel.h"
+#include "ac/game.h"
 #include "ac/gamesetup.h"
 #include "ac/gamesetupstruct.h"
 #include "ac/gamestate.h"
 #include "ac/global_debug.h"
 #include "ac/global_translation.h"
 #include "ac/mouse.h"
+#include "ac/spritecache.h"
 #include "ac/string.h"
 #include "ac/system.h"
 #include "ac/dynobj/scriptsystem.h"
 #include "debug/debug_log.h"
 #include "debug/out.h"
 #include "gfx/graphicsdriver.h"
+#include "gfx/gfxfilter.h"
 #include "main/config.h"
+#include "main/game_run.h"
 #include "main/graphics_mode.h"
 #include "main/engine.h"
 #include "main/main.h"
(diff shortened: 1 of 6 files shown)


ericoporto:
If one method only returns integers and the other only returns strings, and the enums have the type coded there in a way, isn't easier to have an enum for strings and other enums for integers?

> meant mainly for diagnostic purposes 

Alternatively, did you consider making all the returned values being strings?

ivan-mogilko:
GetEngineString appears to be a universal way to get everything, but GetEngineValue is there for convenience. It may be also called GetEngineInteger; and there also may appear a need for GetEngineFloat at some point.

ivan-mogilko:
Made some changes.
Renamed ENGINE_VALUE_I_DISPLAY_TYPE to ENGINE_VALUE_I_DISPLAY_MODE, but I am still not sure how to make it easier to understand, other than introduce a new enum for display modes.

Having anything in these "values" does not preclude from adding actual properties later. For example, we may still have a DisplayMode struct for reading or changing current display mode.

Engine/ac/system.cpp
```diff
@@ -216,6 +220,140 @@ void System_SetRenderAtScreenResolution(int enable)
     usetup.RenderAtScreenRes = enable != 0;
 }
 
+int System_GetEngineInteger(int value_id, int index)
+{
+    int value = 0;
+    if (!GetEngineInteger(value, static_cast<EngineValueID>(value_id), index))
+        debug_script_warn("System.GetEngineValue: undefined engine value %d (#%d), or not an integer value", value_id, index);
```
ericoporto, line 227: this debug message needs to also be renamed to `System.GetEngineInteger`

ericoporto:
> Renamed ENGINE_VALUE_I_DISPLAY_TYPE to ENGINE_VALUE_I_DISPLAY_MODE, but I am still not sure how to make it easier to understand, other than introduce a new enum for display modes, or splitting this in two values (Fullscreen yes/no, Borderless window yes/no).

I think the way that is, is ok as I understand this is meant for debugging issues. My feeling is in general if something would be better to return from an enum, it's better to instead return a string in this API, so it can be easily changed or upgraded later.

In this sense, all integer returns that are explained in comments I would instead switch for strings and just return the explanation. For example, in the endianess, I would return either `"LE"` or `"BE"`, or to keep it as integer I would rename it as BIGEND and return 0 if it's Little Endian (something that makes the 0 or 1 being obvious).

ivan-mogilko:
I changed the list of values to a minimalistic variant, where there's no individual version or display mode parts, but only strings. This way they are meant for checking out by a human rather than using in some program algorithm. Presumably latter is not much useful currently, since there's no way to change display mode, and checking engine version in script is not meaningful either as capabilities are tested differently.

ericoporto:
Hey, does `frames_per_second` really have the maximum frame set? If I call this in AfterFadeIn event I get a `-1` for `System.GetEngineString(ENGINE_VALUE_I_FPS_MAX)` but I can see the maximum FPS on Screen (using `Debug(4,1);`) and I also can get it using `GetGameSpeed()`.

ivan-mogilko:
There's just a missing "break" in case of ENGINE_VALUE_I_FPS_MAX.

ericoporto:
I didn't catch the missing break. Well, it looks alright now. :)

ivan-mogilko:
Forgot to fix the debug_script_warn typo mentioned above...

PR #2359, opened 2024-03-20
CI: Add MinGW build in GitHub Actions

ericoporto:
adding MinGW to the GitHub Actions CI. This is mostly to verify it's still building.

We aren't really using this target for anything, but MinGW could have some uses for testing - someone in Linux could build for Windows using and then run it on Wine, and the someone could be a CI environment.

---

From the warnings, the one that caught my attention was this one

```

2024-03-20T01:22:51.7305404Z D:/a/ags/ags/Engine/debug/debug.cpp: In function 'int check_for_messages_from_debugger()':
2024-03-20T01:22:51.7307229Z D:/a/ags/ags/Engine/debug/debug.cpp:507:36: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
2024-03-20T01:22:51.7308632Z   507 |             editor_window_handle = (HWND)atoi(windowHandle);
2024-03-20T01:22:51.7309372Z       |                                    ^~~~~~~~~~~~~~~~~~~~~~~~
```

https://github.com/adventuregamestudio/ags/blob/317e31a5c868fb6648172a583446cd7dd3badb85/Engine/debug/debug.cpp#L507

I guess in a 64-bit environment the HWND is 64-bit, according to the docs the parts that matters are still 32-bit, and the cast is correct as they are safe to sign-extend the handle : https://learn.microsoft.com/en-gb/windows/win32/winprog64/interprocess-communication?redirectedfrom=MSDN . So nothing is required there (I think).


ivan-mogilko:
I also looked into HWND cast, and found the same info about window handle being strictly 32-bit.
But the warning may be suppressed by replacing with C++ static_cast, I think.

ericoporto:
Yeah, I think also maybe somewhere else (in the Editor?) where this handle is written to a string, it needs to be cast to an int before it's converted to string. Haven't looked at that yet.

PR #2360, opened 2024-03-20
Editor: add room description in tab name

### LINKED ISSUE #1066: show more relevant info in the ROOM tab headers in the editor
when you have a lot of rooms and scripts open seeing only room1.asc, room2.asc, etc makes it hard to read what you are actually doing. 
having the descriptive name I give the rooms in the tab headers would make this easier to read.
having this customizable would be best, like having the option to chose from these display methods in the preferences:
(1)room1name
room1name(1)
room1.asc

where "room1name" is the decriptive name I give the room script in the project explorer of course.

For all other tabheaders this works already, I wouldn't mind having only "I:" in front of an inventory name, but for invitems the tabs work. 
For the rooms it is complicated as there are "room scripts" and the "room room" with images and hotspots both with the same name, but a "S:" for script in front of the room script and the complete room name for the room room should work.

some small icons at the left start of the tab or different colors for different kind of tabs would also make the open tabs easier to distinguish but that is the bigger change... a simple "image"icon and a "script" icon could work...

ericoporto:
fix #1066 

~~I didn't add the icon as requested, it would be good to disambiguate between room and script, but I didn't want to deal with ALL possible tabs.~~ Welp, turns out someone already did that in the past, added a way to turn it on.

It also looked interesting to have some method to deal with the tab name, that could perhaps be overriden in some way, but I didn't figure a common place to them.

image

Please evaluate.

### DIFF
--- a/Editor/AGS.Editor/Panes/ScriptEditor.cs
+++ b/Editor/AGS.Editor/Panes/ScriptEditor.cs
@@ -49,11 +49,24 @@ namespace AGS.Editor
             _agsEditor = agsEditor;
             _showMatchingScript = showMatchingScript;
             _room = null;
-            _roomNumber = 0;
+            _roomNumber = -1;
             Init(scriptToEdit);
 
             this.Load += new EventHandler(ScriptEditor_Load);
             this.Resize += new EventHandler(ScriptEditor_Resize);
+        }
+
+        public string GetScriptTabName()
+        {
+            if (_roomNumber >= 0)
+            {
+                UnloadedRoom room = Factory.AGSEditor.CurrentGame.FindRoomByID(_roomNumber);
+                if(room != null && room.Number == _roomNumber && !string.IsNullOrEmpty(room.Description))
+                {
+                    return Script.FileName + (IsModified ? " *" : "") + ": " + room.Description;
+                }
+            }
+            return Script.FileName + (IsModified ? " *" : "");
         }
 
         private void Init(Script scriptToEdit)
--- a/Editor/AGS.Editor/Components/BaseComponentWithScripts.cs
+++ b/Editor/AGS.Editor/Components/BaseComponentWithScripts.cs
@@ -55,7 +55,7 @@ namespace AGS.Editor.Components
 
         protected void UpdateScriptWindowTitle(ScriptEditor editor)
         {
-            string newTitle = editor.Script.FileName + (editor.IsModified ? " *" : "");
+            string newTitle = editor.GetScriptTabName();
             ContentDocument document = GetDocument(editor);            
             if (document != null && document.Name != newTitle)
             {
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -695,7 +695,7 @@ namespace AGS.Editor.Components
                 scriptEditor.Room = _loadedRoom;
             }
             _roomScriptEditors[selectedRoom.Number] = new ContentDocument(scriptEditor,
-                selectedRoom.Script.FileName, this, SCRIPT_ICON);
+                scriptEditor.GetScriptTabName(), this, SCRIPT_ICON);
             _roomScriptEditors[selectedRoom.Number].ToolbarCommands = scriptEditor.ToolbarIcons;
             _roomScriptEditors[selectedRoom.Number].MainMenu = scriptEditor.ExtraMenu; 
         }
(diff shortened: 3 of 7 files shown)


Editor/AGS.Editor/Panes/ScriptEditor.cs
```diff
@@ -56,6 +56,19 @@ public ScriptEditor(Script scriptToEdit, AGSEditor agsEditor, Action<Script> sho
             this.Resize += new EventHandler(ScriptEditor_Resize);
         }
 
+        public string GetScriptTabName()
+        {
+            if (_roomNumber > 0)
+            {
+                UnloadedRoom room = Factory.AGSEditor.CurrentGame.FindRoomByID(_roomNumber);
+                if(room != null && room.Number > 0 && !string.IsNullOrEmpty(room.Description))
```
ivan-mogilko, line 64: `room.Number > 0` condition is wrong here, because room number 0 is allowed.

Editor/AGS.Editor/Panes/ScriptEditor.cs
```diff
@@ -56,6 +56,19 @@ public ScriptEditor(Script scriptToEdit, AGSEditor agsEditor, Action<Script> sho
...
         }
 
+        public string GetScriptTabName()
+        {
+            if (_roomNumber > 0)
+            {
+                UnloadedRoom room = Factory.AGSEditor.CurrentGame.FindRoomByID(_roomNumber);
+                if(room != null && room.Number > 0 && !string.IsNullOrEmpty(room.Description))
+                {
+                    return Script.FileName + ": " + room.Description + (IsModified ? " *" : "");
```
ivan-mogilko, line 66: I think it's better to have `*` mark right after filename, because when description is too long, the trailing part is hidden, and `*` will not be visible.

E.g. the tab name may look like: "room3.asc: Bla bla bla ..."

Editor/AGS.Editor/Panes/ScriptEditor.cs
ericoporto: Uhm, probably `room.Number == _roomNumber` is the correct condition!

ivan-mogilko:
Sorry for delay, I finally started testing this, and it looks like a simple but useful additions.

Besides the small issues mentioned above, one noteable thing is that when you edit Room's description, it is only applied over Room's tab, and not script's tab (if it was opened at the same time).
I did only a quick search, and RoomComponent appears to have a way to search opened room scripts using a room number:
https://github.com/adventuregamestudio/ags/blob/bfc99b00879abefd426880dcfa1def0ad96c964b/Editor/AGS.Editor/Components/RoomsComponent.cs#L318-L326

Perhaps same method could be used to force-update the script's tab name.

Editor/AGS.Editor/Panes/ScriptEditor.cs
ericoporto: fixed it in this push https://github.com/adventuregamestudio/ags/compare/49ca91d1d367bb35d6b0a7db9feff1b948423ebe..56ad1d6b2d3e0e5edc52c4d4f94f36c03e441544

Editor/AGS.Editor/Panes/ScriptEditor.cs
ericoporto: fixed it in this push https://github.com/adventuregamestudio/ags/compare/49ca91d1d367bb35d6b0a7db9feff1b948423ebe..56ad1d6b2d3e0e5edc52c4d4f94f36c03e441544

ericoporto:
I did a commit https://github.com/adventuregamestudio/ags/pull/2360/commits/0e22d334d2aec0d440da99362ecc100e8987293d that updates the room script name with the suggested method.

The only issue is this. The `BaseComponentWithScripts` has the `UpdateScriptWindowTitle` method that does what we want. Problem is I could figure how to acces the script component from the room component, for that specific script. So I slightly copy pasted six lines of code.

https://github.com/adventuregamestudio/ags/pull/2360/commits/0e22d334d2aec0d440da99362ecc100e8987293d#diff-8b972772a20e32bc3210074e87211dc903fcfc04b1ccf53a8c42acdbf21c1170R1120-R1125

ivan-mogilko:
> The only issue is this. The BaseComponentWithScripts has the UpdateScriptWindowTitle method that does what we want. Problem is I could figure how to acces the script component from the room component

Getting components is done with `Factory.ComponentController.FindComponent<Type>()`.
Although, I wonder if that's at all necessary, as ScriptEditor could probably do it for itself, if it had a function like "UpdateTitle", doing same actions.

ivan-mogilko:
Erm, nevermind the above. It appears that you don't need to add any new code.

RoomsComponent inherits BaseComponentWithScripts class, which already has UpdateScriptWindowTitle method.

ericoporto:
> Following may work:

It does!!! I squashed that in the first commit. :)

PR #2361, opened 2024-03-21
Reorganize DebugManager and make it thread-safe

### LINKED ISSUE #2339: Logging system is not thread-safe
Somehow I missed this, but we don't have much logging done on separate threads yet, and audio polling coincidentally is performed not at all times, but is triggered by certain "sync" actions on main thread, and then halts again; so likely this was not causing trouble in practice.

Existing logging is not designed with thread-safety in mind. Initialization is done before more threads are created, yet there's no reason to not make it safe as well. But besides that the two general concerns are:
1. Output implementations; if these use any non-local data.
2. "Supress" mechanism, see this: 
https://github.com/adventuregamestudio/ags/blob/84ab2b592f5611b22380ed52a3b42080c9c68487/Common/debug/debugmanager.cpp#L218-L220
I suspect that the code in p2 may be reviewed, and possibly a rule enforced that a output implementation must not print to log itself.

A solution is likely to introduce a mutex in the DebugManager class, and guard either PrintMessage call or external call that iterates all outputs. As well as any modification to the output list and settings.

ivan-mogilko:
Resolves #2339

This does a number of changes to DebugManager, with a purpose to make its interface simpler, and thus easier to safeguard its innards when logging from multiple threads. Adds a mutex to protect public methods.

1. Moved DebugOutput inside DebugManager, and removed its access from public interface.
2. Added a mutex, which guards every public method in DebugManager.
3. Moved MessageBuffer object also inside DebugManager, let it handle message buffering instead of hacking this externally.
4. Made DebugManager owning output objects and dispose them when shutting logging down. I had to create a separate class for logging to stdout, called StdOutLogger, which is created by AGSPlatformDriver, as we cannot let DebugManager own whole AGSPlatformDriver.


ericoporto:
I will have to test this tomorrow, but only mentioning here it's worth to check it out if the Emscripten port is still running, because it has no threads.

ericoporto:
ok tested the web port and it seems it's all good. (I remembered https://github.com/libsdl-org/SDL/issues/6344, which is why I was worried, but turns out the mutex used here doesn't crash things and everything works fine).

I am not sure how to test this besides trying a bunch of games and seeing that it still works.

PR #2362, opened 2024-03-22
Editor: move folders between folders and move item downwards in Project Explorer

### LINKED ISSUE #1526: Editor: cannot order an item or folder among other folders in Project Tree
**Describe the bug**
If you have subfolders in any list inside the Project Tree, you cannot drag the item in between or above these folders.
As a result:
1) all items outside of subfolders will always remain at the bottom of the list;
2) you cannot change folder's order.

This is not a very critical problem, but is a quite annoying inconvenience.

The root of the problem, it seems, is that item dragging operation expects the dragged item to be dropped over another _item_ in order to place former _above_ the latter. But if you drag something on top of a folder's name - it works as placing an item _inside_ that folder.

The solution therefore is likely to correctly handle dropping _in between_ the items. Usually in applications this is done by displaying a drop position as a line between two items.

**AGS Version**
Starting from 3.2.2 and up, 3.5.1, 3.6.0.

**To Reproduce**
1. Start the Editor and open any game.
2. In Scripts, for example, create two folders (F1,F2) and two scripts (S1,S2).
3. Try to position F2 above F1. Observe the failure.
4. Try to position script S1 in between two folders. Observe the failure.
5. Try to position script S1 above all folders. Observe the failure.

**Expected behavior**
It should be possible to drag and position items and folders between any items or folders and above them.

ericoporto:
Edit: I copied the things I haven't checked yet to the bottom of this.

fix #1526 

Please test this, also any tips on the things that are problematic let me know.

**NOTE:** This won't fix moving _items_ BETWEEN _folders_. This would require much bigger changes. TBH, I don't even think this is a good idea.

I think the commits would be better if they all got squashed in a single commit, but leaving as is currently. In the PR the files changed tab can show the changes squashed.

https://github.com/adventuregamestudio/ags/assets/2244442/9ac7f6fe-ab3c-4ae8-a768-3dba963114c4

- My own feeling is that this makes dragging and dropping a bit finnicky, because the nodes are very small, the tree nodes could be made slightly bigger in height, I think it would work better for this interaction.
- There is some autoscroll that happens that I don't know what causes.
- [x] The length of the "insert line" depends on the length of a item name, if item's name is short then this line also becomes very short, which is not very convenient. I think it would be nice to have it either fixed length (maybe something like half of explorer's width) or at least have a minimal length.
- [x] As you stated, dragging an item between folders is not supported, but the "insert line" still shows up. This causes wrong expectations: user may think that they are going to place item between folders, but in reality it will get inside the folder.
- [x] It looks like the "insert line" can also place itself above or below the parent node if you hover a cursor over it while dragging.
- [x] Dragging any item at the parent node will make it move to the bottom of the list.
- [ ] If you select an item A by clicking, then hold and drag another item B, the selection returns to item A (but you still drag B successfully).

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -191,6 +191,9 @@
     <Compile Include="GUI\ImportTTFDialog.Designer.cs">
       <DependentUpon>ImportTTFDialog.cs</DependentUpon>
     </Compile>
+    <Compile Include="GUI\LineInBetween.cs">
+      <SubType>Component</SubType>
+    </Compile>
     <Compile Include="GUI\LogPanel.cs">
       <SubType>Form</SubType>
     </Compile>
--- a/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
+++ b/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs
@@ -256,7 +256,7 @@ namespace AGS.Editor.Components
                     (commandId == TOP_LEVEL_COMMAND_ID));
         }
 
-        private bool ProjectTreeItem_CanDropHere(ProjectTreeItem source, ProjectTreeItem target)
+        private bool ProjectTreeItem_CanDropHere(ProjectTreeItem source, ProjectTreeItem target, TargetDropZone dropZone)
         {
             FolderType targetFolder;
             ItemType targetItem;
@@ -305,7 +305,7 @@ namespace AGS.Editor.Components
             }
         }
 
-        private void ProjectTreeItem_DropHere(ProjectTreeItem source, ProjectTreeItem target)
+        private void ProjectTreeItem_DropHere(ProjectTreeItem source, ProjectTreeItem target, TargetDropZone dropZone)
         {            
             FolderType targetFolder;
             ItemType targetItem;
@@ -328,7 +328,16 @@ namespace AGS.Editor.Components
             {
                 if (targetFolder == null)
                 {
-                    DragItemToBeBeforeItem(sourceItem, targetItem);
+                    if (dropZone == TargetDropZone.Bottom || dropZone == TargetDropZone.MiddleBottom)
+                    {
+                        DragItemToBeAfterItem(sourceItem, targetItem);
+                        //Console.WriteLine("drop after");
+                    }
+                    else
+                    {
+                      //  Console.WriteLine("drop before");
+                        DragItemToBeBeforeItem(sourceItem, targetItem);
+                    }
                 }
                 else
                 {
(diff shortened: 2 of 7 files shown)


ivan-mogilko:
> NOTE: This won't fix moving items BETWEEN folders. This would require much bigger changes. TBH, I don't even think this is a good idea.

Could you elaborate why don't you think that's a good idea?

The original ticket was written for Scripts, because in Scripts the order of items matters, and I had some trouble not being able to move some single item above a folder with scripts. But if moving folders works, then this may be worked around by creating a folder even for a single item and moving it upwards.

---

About this PR, I started testing this, and there are following problems I noticed so far:
1. While dragging, the "insert line" control seems to be repeatedly jumping to (0,0) and reposition itself back, causing very noticeable blinking. Also I think that Project Explorer repaints itself often: moving mouse up down while dragging causes it to blink heavily. But the latter is an older problem, not a new one.
2. The length of the "insert line" depends on the length of a item name, if item's name is short then this line also becomes very short, which is not very convenient. I think it would be nice to have it either fixed length (maybe something like half of explorer's width) or at least have a minimal length.
3. As you stated, dragging an item between folders is not supported, but the "insert line" still shows up. This causes wrong expectations: user may think that they are going to place item between folders, but in reality it will get inside the folder.
5. It looks like the "insert line" can also place itself above or below the parent node if you hover a cursor over it while dragging.
6. Dragging any item at the parent node will make it move to the bottom of the list.
7. If you select an item A by clicking, then hold and drag another item B, the selection returns to item A (but you still drag B successfully).
8. If you have a folder A inside a folder B, you cannot move A outside of B by dragging it to a place above or below B. This also makes it impossible to move subfolder out of parent folder if there are no more folders present.

ericoporto:
Thanks for the review! I will analyze and see how to handle things and answer soon, but just wanted to mention one thing

> While dragging, the "insert line" control seems to be repeatedly jumping to (0,0) and reposition itself back, causing very noticeable blinking. Also I think that Project Explorer repaints itself often: moving mouse up down while dragging causes it to blink heavily. But the latter is an older problem, not a new one.

This will sound ridiculous but there is no event in WinForms for a drag being canceled! Afaict, I would need to somehow query the dragging of the TreeNode whereas all the drag over and drop logic is handled in the ProjectTree. I looked around a lot and could not figure how to do this, so the way I went to deal with this is with a timer and the timer cause the blinking.

The problem here is to avoid repaint on the ProjectExplorer, instead of overriding it's paint method or something, the way I approached this was using a control that is a line and simply placing it on top, but if the drag is cancelled I need to hide it, otherwise it stays in place.

I spent probably more time than anything else in this PR trying to figure how to know if the drag is cancelled. I will set it as draft for now because I don't expect to figure any answer for this soon, and would appreciate if anyone has any knowledge in WinForms to handle this.

ericoporto:
> If you select an item A by clicking, then hold and drag another item B, the selection returns to item A (but you still drag B successfully).

I can't think of anyway to prevent this, but I think I took care of everything else.

ivan-mogilko:
Looks working nicely now.

PR #2364, opened 2024-03-24
Editor: add feature add existing ash/asc files

### LINKED ISSUE #589: Editor: allow to import scripts as ash/asc pair
Right now editor only lets you import scripts in the exported module format (scm), but that's inconvenient because you cannot even import your own scripts from another game without packaging them first or doing some workaround.

I suggest to let user choose ash or asc file as well.
* if scripts are in another folder, they get copied to the current project folder first;
* if the second of the pair does not exist it will be created as empty file.
* if the script with same name already exists in the project, then share behavior with scm import (right now it creates a duplicate appending a number to to the module's name; idk if that's a good thing in general but better keep things consistent imo).

ericoporto:
fix #589

This took way more lines of code than I thought it would. I broke things down in multiple steps and when I went to implement there was a lot that I thought would be written that just wasn't.

<details><summary>Hidden completed checklist because it's not as relevant now!</summary>

- I think I can't figure the encoding from the file itself, so I am using the one set in the project.
- ~Are special characters in file names allowed? (Specifically basename)~ I think the filesystem already took care of this
- [x] Are special character in the script name allowed? ~~(Is there a function for this?)~~ There is a function but turns out they are already checked by the filesystem itself, so no need to call it again
- [x] ~~I still need to~~ test this a bit more for spaces in filenames. Apparently they don't matter
- [x] I think if a person is adding an existing script file that is in the game project directory  and it has a header there that pairs but they didn't select, it should magically figure it out and pick it up, otherwise their content may be deleted in the process (need to check)
- [x] Same as before but switch script for header and header for script.
- [x] forgot to refresh the project tree after adding scripts, doing that correctly now!

~~Leaving as a draft until I can test a bit more.~~ 

</details>

Tested and it appears to work alright!

### DIFF
--- a/Editor/AGS.Editor/ImportExport.cs
+++ b/Editor/AGS.Editor/ImportExport.cs
@@ -311,6 +311,24 @@ namespace AGS.Editor
             return importErrors;
         }
 
+        public static List<Script> AddImportedScriptModule(ScriptModuleDef module)
+        {
+            string author = module.Author;
+            string description = module.Description;
+            string name = module.Name;
+            string version = module.Version;
+            string scriptText = module.Script;
+            string headerText = module.Header;
+            int uniqueKey = module.UniqueKey;
+
+            List<Script> scriptsImported = new List<Script>();
+            Script header = new Script(null, headerText, name, description, author, version, uniqueKey, true);
+            Script mainScript = new Script(null, scriptText, name, description, author, version, uniqueKey, false);
+            scriptsImported.Add(header);
+            scriptsImported.Add(mainScript);
+            return scriptsImported;
+        }
+
         public static List<Script> ImportScriptModule(string fileName, Encoding defEncoding)
         {
             BinaryReader reader = new BinaryReader(new FileStream(fileName, FileMode.Open, FileAccess.Read));
@@ -328,9 +346,9 @@ namespace AGS.Editor
 
             Encoding enc = defEncoding;
 
-            var author = ReadNullTerminatedStringAsBytes(reader);
-            var description = ReadNullTerminatedStringAsBytes(reader);
-            var name = ReadNullTerminatedStringAsBytes(reader);
+            var bAuthor = ReadNullTerminatedStringAsBytes(reader);
+            var bDescription = ReadNullTerminatedStringAsBytes(reader);
+            var bName = ReadNullTerminatedStringAsBytes(reader);
             var version = ReadNullTerminatedString(reader);
 
             int scriptLength = reader.ReadInt32();
@@ -361,15 +379,24 @@ namespace AGS.Editor
             }
             catch (ArgumentException) { }
 
-            List<Script> scriptsImported = new List<Script>();
-            Script header = new Script(null, enc.GetString(headerBytes), enc.GetString(name),
-                enc.GetString(description), enc.GetString(author), version, uniqueKey, true);
-            Script mainScript = new Script(null, enc.GetString(scriptBytes), enc.GetString(name),
-                enc.GetString(description), enc.GetString(author), version, uniqueKey, false);
-            scriptsImported.Add(header);
-            scriptsImported.Add(mainScript);
-            return scriptsImported;
+            string author = enc.GetString(bAuthor);
+            string description = enc.GetString(bDescription);
+            string name = enc.GetString(bName);
+            string script = enc.GetString(scriptBytes);
+            string header = enc.GetString(headerBytes);
+
+            ScriptModuleDef module;
+            module.Author = author;
+            module.Description = description;
+            module.Name = name;
+            module.Version = version;
+            module.Script = script;
+            module.Header = header;
+            module.UniqueKey = uniqueKey;
+
+            return AddImportedScriptModule(module);
         }
+
 
         public static void ExportScriptModule(Script header, Script script, string fileName, Encoding defEncoding)
         {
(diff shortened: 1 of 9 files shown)


Editor/AGS.Editor/ImportExport.cs
```diff
@@ -311,6 +311,38 @@ public static CompileMessages ImportOldEditorDatFile(string fileName, Game game,
...
+        public static List<Script> AddImportedScriptFromAscAsh(Encoding defEncoding, string scriptName, int uniqueKey, byte[] scriptBytes, byte[] headerBytes)
+        {
+            byte[] b_empty = defEncoding.GetBytes(string.Empty);
+
+            byte[] author = b_empty;
+            byte[] description = b_empty;
+            string version = string.Empty;
+            byte[] name = defEncoding.GetBytes(scriptName);
+
+            return AddImportedScriptModule(defEncoding, author, description, name, version, scriptBytes, headerBytes, uniqueKey);
```
ivan-mogilko, line 343: I know this is a draft, but I'd like to comment on these 2 functions right away.
I think it's better to have the "final" function with convenient argument types, which is "string" in this case, and if input data is in different type (like `byte[]`), then convert that to the final type. In this code this is done "backwards", which makes string converted to byte[] and then byte[] converted to string for no good reason.

ivan-mogilko: Another thing that I might propose is to make a struct for holding this module's "meta info", create that struct from input, and pass into the "final" function that imports the module.

ericoporto: Did a push that tries to address this comment: https://github.com/adventuregamestudio/ags/compare/d9a0902208c100097560f56739fe723bcef2b63b..4cf2490ae863ccb539d541976460d563ce46a62a

the code became a bit longer but I like to break things in separate ways so I can debug it more easily. I can try to reduce the number of lines if needed.

ivan-mogilko:
There are 2 general things that I need to note:
1. This pr lacks comments for new functions and structs, sometimes it's not clear what exactly the function does or what the struct is meant for. Just a brief explanation above it would be nice to have.
2. New "Add existing script" command does not report anything if operation fails or forbidden. For example, trying to add a room script (roomN.asc) does nothing, but nothing is reported to user.

Something I am not sure about, is it intentional that if I don't select full pair (ash + asc) for adding, then only selected one is copied and another generated as a new blank file?

ericoporto:
I will answer only the last thing because the rest I will need a computer, but the rule is 

- if you are adding a file that comes from a different folder, you have to explicitly add all files you want. 
- If you are adding a file that you previously had already copied in the project dir, it will find the missing pair instead of overwriting it with an empty one. This is done to avoid you losing work you may have done in the file.

I also don't give any errors because in this way I could just select all files from my project dir and it would just add what is necessary - I was imagining the use case of how I would use such feature which would be just picking and choosing my own stuff from previous project and pasting in a new game project folder so I would later open the project in AGS Editor and just add all files.

ivan-mogilko:
> I also don't give any errors because in this way I could just select all files from my project dir and it would just add what is necessary

There are different situations; if a file was not added because it's already in a project, then it's not a error. But if file cannot be added because it's forbidden to add, or there was any exception during this process, then it is an error, and user has a right to know what have happened.

If there's a concern of not blocking a batch operation, then this may be solved by gathering all error/warnings into a list and posting them into the output panel.

ericoporto:
> If there's a concern of not blocking a batch operation, then this may be solved by gathering all error/warnings into a list and posting them into the output panel, similar to how some other processes do

Uhm, if I add something there it clears the output panel before adding, but if I don't add anything there any previous error stays there. Should I clear the output panel if no error occurs?

ivan-mogilko:
> Uhm, if I add something there it clears the output panel before adding, but if I don't add anything there any previous error stays there. Should I clear the output panel if no error occurs?

This is unfortunately inconsistent throughout the Editor, but I suppose that in case of a import action it should not clear anything.

OTOH, I think that if user requested only for 1 file, and it failed, then it should display this error in a message box instead.

ericoporto:
I had an idea for something like an event log, where it would write things with the option of showing a "balloon" of sorts that points to it highlighting a summary of issues. Something that would be similar to the output panel, but multiline and primarily text - maybe it could use red to highlight errors. Not something to do it here, but just a general ideal. This event log would autoscroll and not clear itself.

ericoporto:
Threw some message box there but I absolutely hate it.

ivan-mogilko:
Well, I'm certain there could be ways to improve error reporting to make it both visible and less annoying...

There's just one other thing I'd like to ask, is regarding the code style. I noticed that you sometimes write a single statement directly after `if`; there are multiple examples of this in function `GetNormalizedScriptHeaderPairs` where there are returns, and also in the latest commit, e.g.
```
if (isMsgBoxShownOnError) _guiController.ShowMessage("Some of the selected files could not be added, see output panel for more details", MessageBoxIcon.Warning);
```

This goes against our code style, and may make the code harder to read especially when you try to quickly grasp the code branching in a function. Could you get these statements on a second line please?

ericoporto:
I adjusted the if clauses. I added a thing there to conditionally show the message box and tried to tie it to the config on compile error message box, I could create a new specific property only for it but I think for the time being it's good enough.

PR #2368, opened 2024-03-30
Editor: fix script editor fold in comment in scintilla lexxer

### LINKED ISSUE #2323: Script editor collapsable block matching problems lead to the blocks collapsing incorrectly or not at all.
**Describe the bug**

There are a couple of areas in the script editor that don't collapse correctly. Described below.

**AGS Version**
Editor 3.6.0.56

**To Reproduce**
1. Open a script editor
2. Paste the code below

```agsscript
void bug(int value) {
  if (value == 0) {
  } else { // Not collapsable
    Display("Not zero");
  }

  if (value < 0) {
  }
  else {
    // if (value > 10) {
    //  Display("Greater than 10");
    //  return;
    //} // Collapses here. Notice that // } is fine, but when without space //} it breaks
    Display("Greater than 0");
  }
}
```

3. Try collapsing the statements.

**Expected behavior**
* `} else {` should be collapsable
* Should ignore commented out code

**Screenshots**
image6f)

 - OS: Windows
 - Version 10

ericoporto:
fix #2323

The issue here is if foldExplicitStart is empty it is defined as `"//{"`  and foldExplicitEnd it is defined as `"//}"`.

This causes the confusing behavior that if foldCommentExplicit  is true, the scintilla cpp lexxer can fold in comments following the previously defined start and terminator string. We are setting it to false to disable this unexpected behavior.

### DIFF
--- a/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
+++ b/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
@@ -324,6 +324,8 @@ namespace AGS.Editor
             // to make this work properly we need to supply keywords for preprocessor,
             // otherwise lexer will not know about external defines.
             scintillaControl1.SetProperty("lexer.cpp.track.preprocessor", "0");
+            // avoid confusing comment fold feature, see https://github.com/adventuregamestudio/ags/issues/2323
+            scintillaControl1.SetProperty("fold.cpp.comment.explicit", "0");
 
             Factory.GUIController.ColorThemes.Apply(LoadColorTheme);
             UpdateAllStyles();


ivan-mogilko:
I'm sorry for a long delay, the issue sounded confusing and I needed to find a spare moment to grasp what is it about.
I still don't quite understand internal logic of this setting, so I'll leave this to you.

But it's not good to hack the setting inside the library. There seem to be a way to set it from outside, like we do to another property here:
https://github.com/adventuregamestudio/ags/blob/72edc8af3a2992a954c366c54b7238b065c21489/Editor/AGS.Editor/Panes/ScintillaWrapper.cs#L323-L326

This command should be equivalent to your change:
```
scintillaControl1.SetProperty("fold.cpp.comment.explicit", "0");
```

ericoporto:
Ah, nice, I didn't saw this interface in the C# side! I will adjust this PR to use it instead!

Issue #2087, opened 2023-08-13
Support Editor Scaling when DPI Aware is enabled


ericoporto:
In Windows 10, enabling DPI Aware gives crisp fonts in the Script Editor. Unfortunately, if your windows scaling is beyond 125%, due to some high-dpi monitor, a few issues may arise - these are better visualized using a 200% scaling. I don't believe we need to support further scaling than 200% so I don't recommend testing with beyond it. Here are all the panes in current (https://github.com/adventuregamestudio/ags/tree/2030bdab74933e13300765f2b66edbe811b2ab81) AGS 3.6.1.5 that require improvements.

- [ ]  The Welcome Pane
  ![scal_welcome_page](https://github.com/adventuregamestudio/ags/assets/2244442/c735768b-de82-4077-8b00-112c3ed202f1)
  - Mostly a few texts being cut-off and the lack of auto scrolling

- [ ] The Text Parser Pane
  ![scal_text_parser](https://github.com/adventuregamestudio/ags/assets/2244442/94afe797-e53c-456d-b29b-a4bcb6c4cffe)
  - text being cut and label being cut too, labels are terrible to scale in their own...

- [ ] The Font Pane
  ![scal_font_pane](https://github.com/adventuregamestudio/ags/assets/2244442/b9a6d21c-e11d-4e91-b5df-641ae21e9d8e)
  - there is a lack of auto scrolling, so we can't scroll to the right here

- [ ] The Global Variables Pane
  ![scal_global_variables](https://github.com/adventuregamestudio/ags/assets/2244442/b031380c-cda8-43c7-a458-8f3853bf7e1a)
  - Text being cut, we also need a more dynamically sized table as the columns are narrow here if you consider the font size...

- [ ] The Dialog Editor pane
  ![scal_dialog_pane](https://github.com/adventuregamestudio/ags/assets/2244442/a5c528c0-7603-47a0-9d95-b439738fc57a)
  - The left-side elements, the column cuts the two checkbox that should be visible there

- [ ] The View Pane
  ![scal_view_pane](https://github.com/adventuregamestudio/ags/assets/2244442/16a1b96b-bb24-42b2-9f9a-f2d6e9b938cc)
  - various issues, probably needs a lot of restructuring

- [ ] The Colours Pane
  ![scal_colours_palette](https://github.com/adventuregamestudio/ags/assets/2244442/2d0fdfe6-15d1-4d45-9066-afff377c8d1b)
[...]

PR #2369, opened 2024-03-30
Editor: fix welcome pane scaling under dpiaware

ericoporto:
Allow scaling of the Welcome Pane, this is step 1/7, partial of #2087

I plan to work on each pane slowly when I find time, in separate PRs, also because these panes have lots of specific details, I think it's better to work them one by one.

<details>
<summary>Editor at 125% scaling</summary>

image

</details>

<details>
<summary>Editor at 200% scaling</summary>

image

</details>

To test this PR I recommend editing the `Editor/AGS.Editor/app.manifest` and uncommenting the dpi aware setting (lines 52 to 58).

### DIFF
--- a/Editor/AGS.Editor/Panes/WelcomePane.Designer.cs
+++ b/Editor/AGS.Editor/Panes/WelcomePane.Designer.cs
@@ -29,136 +29,203 @@ namespace AGS.Editor
         private void InitializeComponent()
         {
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(WelcomePane));
-            this.panel1 = new System.Windows.Forms.Panel();
+            this.panelTitle = new System.Windows.Forms.Panel();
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Something I noticed: if I repeatedly click "Show another tip", sometimes changing a tip resizes the tip panel horizontally, and the big "what's new" panel shifts to right and left. This was not happening previously.

ericoporto:
Can you take a screenshot of this? Does it become larger?

ivan-mogilko:
I don't think a screenshot will tell much, the difference is tiny, but the shifting is noticeably if you repeatedly click on it.

It looks like the left panel becomes a bit wider, and the main panel becomes a bit smaller.

ericoporto:
> The question here is mostly why are they allowed to resize like this?

I made so they could grow horizontally a little too if necessary. But now I removed this and fixed the width, so it can only grow vertically. It appears to work alright at least from 100% to at most 300%. I think this if fine.

PR #2370, opened 2024-03-30
Editor: create a parent class for Object and Character room filters

### LINKED ISSUE #2230: Editor: base class for ObjectsEditorFilter and CharactersEditorFilter
This is a strictly refactor task, no change in functionality implied.

CharactersEditorFilter was added much later in development, seemingly by copy-pasting ObjectsEditorFilter's code.
These two classes share much things in common. It would be great to have a parent class picked out, having as much shared code as possible, where any differences would be handled in specialized classes either by configuring behavior settings, or handling extra things in virtual methods.

This would of course be also beneficial in case any new type of object is added in the future which has similar editing methods.

ivan-mogilko:
Resolves #2230

Introduces BaseThingEditorFilter<TThing>, a generic parent class for room filters which handle "things": objects which have precise coordinates and are drawn individually.
Picks out much of the shared code out from ObjectsEditorFilter and CharactersEditorFilter classes.

Although BaseThingEditorFilter is generic, it cannot access actual object properties, because, unfortunately, Character and RoomObject do not have a base class which would share common properties... This is why when there's a need to get something out of them, such as position or a script name, it has to call a virtual method to let child class handle this.

This does not have to be seen as a "final" implementation; as interfaces are not modified, and all classes are internal to the editor, these classes may be changed and optimized further.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -60,7 +60,7 @@
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
     <LangVersion>6</LangVersion>
     <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>    
+    <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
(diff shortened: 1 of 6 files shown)


ericoporto:
I think there is an issue with characters, I am testing the build from the CI of the Editor and I can't see the character 0 (cEgo) and I added a second character to the same room and I can't move it.

image

ivan-mogilko:
The second commit have broken the object list initialization.

ericoporto:
Yes, it now works correctly now with characters and objects!

PR #2371, opened 2024-04-01
Editor: Change room icon for state saving rooms

AlanDrake:
Implements different rooms icon, depending if the room saves state or not.
Should be a nice-to-have if we stop relying on room number to decide if a room saves state.

I reused pre-existing icons, but I could make an alternate version where the diskette remains colored even when the room is not loaded to make it more visible:
immagine

Opened:
immagine


ericoporto:
They are so cute! Now I noticed that the other room icon has the area masks there, super interesting.

AlanDrake:
I did a quick mockup, diskettes only look readable when black or blue

immagine

So, do I make a new icon with colored diskette? (blue or black?)

ericoporto:
I think the blue one is the best one.

Sorry for the trouble 😬

AlanDrake:
Alright, redone with the blue one.

ericoporto:
Weird, the files change tab only contains the change to csproj file and the change in script but no new icon added. :/

AlanDrake:
My bad, I had resetted to make less useless changes in the csproj and frogot to readd the new icon.

ericoporto:
I think it's still missing one icon (either the grayed out one or the non-grayed out one)

AlanDrake:
I only created the "grayed out with blue disk" icon, everything else are pre-existing resources, so this is correct.

PR #2378, opened 2024-04-04
3.6.1: rewrote dynamic sprite to texture notification mechanism

ivan-mogilko:
The previous attempt, using a boolean `spritemodified` array, proved to be having logical loopholes. In simple words, because `spritemodified` flag was resetting each draw pass, it won't affect objects that are temporarily not visible, and they failed to reupdate when becoming visible again. This affected Software renderer.

Texture-based renderers have a different problem, where they fail to recognize when an old sprite was deleted and a completely new dynamic sprite created accidentally with the same ID. The cause was simply that the check for this is done inside sync_object_texture(), but sync_object_texture itself is not called unless draw() function thought that object had changed. So there has to be a check outside too...

Here I decided to reimplement notification mechanism, and use it for both types of renderers, for the sake of more straightforward code logic. (Although, strictly speaking, texture-based renderers could check their texture's "ref id" instead.)

Anyway, the idea of a new implementation is this:
1. There's a map of "control blocks" per each active sprite (that is - sprite that was drawn as a part of object). The "control block" is simply a integer with a sprite ID in it, allocated and wrapped by shared_ptr.
2. When the ObjectTexture syncs its texture with a new sprite, the control block is retrieved (or new created), and from now on it may be shared among multiple objects which display same sprite.
3. When the sprite gets either updated or deleted two things happen: the previous control block is assigned an "invalid value", serving as a marker, and then it is removed from the map, left only if referenced by any ObjectTexture.
4. When checking whether texture needs to be re-synced we test the control block too, to see if it was invalidated. In such case this tells us to call sync_object_texture() again.

NOTE that dynamic sprites that do not appear on screen, and are only used for drawing on another surface, do not allocate control blocks, nor notify anything.

---

This mechanism was, ironically, similar to the first idea I had when doing #2117, and is mentioned in following commit:
https://github.com/adventuregamestudio/ags/pull/2117#issuecomment-1705778549
unfortunately, I had a change of mind and did this "spritemodified" array solution instead:
https://github.com/adventuregamestudio/ags/pull/2117#issuecomment-1717920385
But, looking at my oldest implementation saved in my repository here: https://github.com/ivan-mogilko/ags-refactoring/commit/7c4fcdc3e787fce738d7ca07c8ee11ccf6a99231 it appears that I did not do it completely right, so it would have same problem as there's now.

Something that I might try to change in this PR, is only create notification block for **dynamic** sprites. That will change few conditions, but might save memory.

But first I need to be sure that this works overall.

And of course I would need to do performance tests again, just in case.

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -246,7 +246,8 @@ struct SpriteInfo
         : Width(w), Height(h), Flags(flags) {}
 
     inline Size GetResolution() const { return Size(Width, Height); }
-
+    // Gets if sprite is created at runtime (by engine, or a script command)
+    inline bool IsDynamicSprite() const { return (Flags & SPF_DYNAMICALLOC) != 0; }
     //
     // Legacy game support
     //
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -156,11 +156,10 @@ struct ObjTexture
         }
     }
 
-    // Tests the sprite notification block to ensure that the texture
-    // is synchronized with the latest sprite version
-    inline bool IsSynced() const
+    // Tests if the sprite change was notified
+    inline bool IsChangeNotified() const
     {
-        return SpriteNotify && (*SpriteNotify == SpriteID);
+        return SpriteNotify && (*SpriteNotify != SpriteID);
     }
 
     ObjTexture &operator =(ObjTexture &&o)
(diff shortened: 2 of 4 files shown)


ivan-mogilko:
I tested this PR with the "Mage Initiation" game, where it has a scene with an image made of dozens of dynamic sprites, and performance seems to be more or less identical to release-3.6.1 branch.

I think, for a good performance test we'd need a game where alot of objects switch sprites (both static and dynamic) in a quick succession, but idk if we have something like that.
OTOH a game with just very large number of objects on screen may be tested to, just to see that new "has changed" check did not  have a noticeable impact on drawing pass.

ivan-mogilko:
@ericoporto Alright, I tested for performance using the 3 games you sent to me, and it looks like the general level of max fps did not change. There may be a little change, but it's hard to tell judging only by the fps counter on screen. I guess in the long run we better have lowest/highest fps recorded and logged periodically or by demand, and a proper test game which has a stable "workload" (in the regular games it's only ever stable when you load a scene and don't do anything).

ericoporto:
I've thought about in the past of making a benchmark game that would run in the CI to get measurements, but in the end I decided against because the hardware we get in the CI run is unknown, as is any possible additional load it may be running without our knowledge. If you think such game would have some use even if we don't run it automatically I can look into making something up that is meant to be run manually later and put in the test game repository - outside of the scope for this PR.

As I said to you, I didn't notice the bug mentioned in any game I tried. The changes here apparently also made no performance or behavior change I could detect in any game of mine.

So, I would say this PR looks fine.

PR #2381, opened 2024-04-12
call ReadPlugin already with version 2.55

rofl0r:
Zak McKraken 2 is game version 2.55 and uses the ags flashlight plugin. using this patch it successfully parses the plugin data.

https://www.maniac-mansion-mania.com/index.php/en/resources-mainmenu-76/downloads/episoden/fanadventures/95-the-new-adventures-of-zak-mckracken/file.html

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -948,7 +948,7 @@ HGameFileError ReadGameData(LoadedGameEntities &ents, std::unique_ptr<Stream> &&
         return new MainGameFileError(kMGFErr_GameEntityFailed, err2);
     game.numgui = ents.Guis.size();
 
-    if (data_ver >= kGameVersion_260)
+    if (data_ver >= kGameVersion_255)
     {
         err = ReadPlugins(ents.PluginInfos, in);
         if (!err)


ivan-mogilko:
Confirmed that random 2.55 game has a valid data for plugin infos (format version and number of plugins = 0).

rofl0r:
@ivan-mogilko does above game work for you ? it fails to load room128 later on in my case, but my ags version is a few months old already.

ivan-mogilko:
> @ivan-mogilko does above game work for you ? it fails to load room128 later on in my case, but my ags version is a few months old already.

I've only tested that it starts up and launches the intro. I will test again and see what happens...

ivan-mogilko:
Hmm, after watching the intro I get this error:

> Error: Unable to load the room file 'room107.crm'. Error: Failed loading a room from file 'room107.crm'.
There was error reading room data.
This engine cannot handle requested room content.
Room animations are no longer supported.

This error refers to "room animations" data which meaning is currently unknown to us.
But the loading code is simply commented out
```
in->ReadArray(&fullanims[0], sizeof(FullAnimation), fullanim_count);
```
maybe it could be restored and read, with a warning to the log, because we do not exactly know how it should work.

rofl0r:
i just submitted a PR that fixes the game in my few months old AGS version. in my case i didn't even get the first room or intro loaded before the change.

rofl0r:
ah, now it crashes at room 110 after intro. maybe both fixes are needed.

ivan-mogilko:
Well, I pushed a fix 890a5db403e07144b37b1602f300417b328377c2, and it starts the game well now.
I have only tested couple of first rooms though.

PR #2382, opened 2024-04-12
room file: fix loading of zak mckraken 2

rofl0r:
roomver is 2.53, and the skip_rle code skips way too many bytes way over the script data, this change fixes the game.

### DIFF
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -316,7 +316,7 @@ HError ReadMainBlock(RoomStruct *room, Stream *in, RoomFileVersion data_ver)
         room->BgFrames[0].Graphic = load_rle_bitmap8(in);
 
     // Area masks
-    if (data_ver >= kRoomVersion_255b)
+    if (data_ver >= kRoomVersion_253)
         room->RegionMask = load_rle_bitmap8(in);
     else if (data_ver >= kRoomVersion_114)
         skip_rle_bitmap8(in); // an old version - clear the 'shadow' area into a blank regions bmp (???)


ivan-mogilko:
I have no errors with this game currently, could you tell which commit or branch you are building from?

Also, you say that "skip_rle" skips too many bytes, in which case that should be fixed instead.

ivan-mogilko:
There was this fix to skip_rle function a while ago, could it be the same problem in your case?
3df6b7185685bcd6a2d6cbdb7b5c3b62ea5dc9b0

rofl0r:
i'm on 38922c2eb62b8082ae7ba07f07a8fe2192f1bac5 - i didn't update yet because rebuilding everything from scratch is so slow, and i've seen some more C++ features are in use, which may no longer be compatible with my C++-14 compiler (gcc 6.5.0).
skip_rle skips

```
(gdb) p (w * h) + (3 * 256)
$101 = 64768
```
~64 kbytes, but the next data block starts way earlier.
when skip is called, position is `0x4c83` but next non-rle data is ~`0x68cd`, so only 7242 bytes. since rle is compressed, i think you cant just multiply the sizes but have to actually decode it.

ivan-mogilko:
> i'm on https://github.com/adventuregamestudio/ags/commit/38922c2eb62b8082ae7ba07f07a8fe2192f1bac5

So, likely the fix I mentioned above (https://github.com/adventuregamestudio/ags/commit/3df6b7185685bcd6a2d6cbdb7b5c3b62ea5dc9b0) should fix this problem.

> and i've seen some more C++ features are in use, which may no longer be compatible with my C++-14 compiler (gcc 6.5.0).

We're not using C++14 yet, we use only couple select features from C++14 which we had to reimplement in case a compiler does not support them. I don't think we have anything beyond 14.

ivan-mogilko:
I do not try or test old games much nowadays, so when regressions happen they might go unnoticed for some time...

PR #2395, opened 2024-04-27
4.0: Overlays always store their explicit position, reimplemented autopositioning

### LINKED ISSUE #1180: Overlay.X and Y properties set/get inconsistency
Comparing the setter and getter for Overlay.X and Y one may notice that setters apply x and y fields directly:
https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/overlay.cpp#L83

However the code of Overlay.X and Y getters retrieve calculated value from get_overlay_position which may return an offseted coordinate:
https://github.com/adventuregamestudio/ags/blob/master/Engine/ac/overlay.cpp#L73

Note this is the original behavior from <= AGS 3.2.1.

The value returned by getters may have following meaning:

* If overlay.x is OVR_AUTOPLACE then engine _treats overlay.y as a character ID_, and calculates overlay rect relative to character's position in the room, then translates to screen coordinates through the viewport.
* Otherwise, following may happen:
    * overlay may be flagged as being "relative to room" (it has RelativeToScreen flag which is usually treated inversely), in which case its coordinates will be again translated to screen coordinates through the viewport.
    * additionally if it's an automatic overlay with text window drawn on it (created for speech etc), then coordinates are shifted by offset (which is a text window border/padding). AFAIK the offsets may only be negative (thus moving coords up-left). (For more historical context see #1098, #1104)

In short, the position returned by Overlay.X and Y may be not at all what user may have set, and even in different coordinate system.

My personal view on this is that if there's an API property X then it should return exactly same thing as was set, with only exception when engine could not apply a required value because it was out of range and thus corrected it. And definitely there must be a rule about coordinate system (if setter expects room coordinates then getter returns same, and similar with screen coords). If there's a separate calculated value then it should be returned by a different (perhaps readonly) property or not at all.

It's also interesting to note that there's no check for special OVR_AUTOPLACE in setters. I am very curious what will happen if one sets their script overlay to x = OVR_AUTOPLACE and y = some character ID. That might as well work, in which case it's another unexpected and undocumented feature.

ivan-mogilko:
This fixes #1180 for AGS 4. Initially I had a deeper reimplementation in mind for ags4, where not overlays keep links to characters, but characters have a link to their speech overlays, but then decided to step back for a simpler one for a time being.
I also think that this is safe to be backported to ags3, as this is really a bug fix, does not modify a save format; and the only thing different will be returned property values, which may be wrapped in a "script api" check. But I'll look into that separately.

---

The purpose of this change is to have x,y parameters of ScreenOverlay to contain explicit coordinates at all times, and use coordinates strictly depending on the Overlay's layer (screen or room).

Historically, non-blocking character speech overlays stored a magic OVR_AUTOPLACE value in their x parameter, and character's index in y parameter. This told the engine to calculate this overlay's drawing position based on linked character's position. Furthermore, for blocking speech the flag kOver_PositionAtRoomXY told to assume Overlay stores its position in room coordinates, even though it's an overlay on a screen layer.

This made the meaning of x,y parameters complicated, and contributed to discrepancy between values set to Overlay.X and Y properties, and values returned by reading them.

In this change:
* Magical OVR_AUTOPLACE value is removed.
* Overlay's flag kOver_PositionAtRoomXY is replaced with kOver_AutoPosition; - that is safe to do, because PositionAtRoomXY was ever only set along with OVR_AUTOPLACE.
* bgSpeechForChar parameter is renamed to speechForChar, and is assigned regardless of whether speech is blocking or non-blocking.
* Overlays now always store explicit position in x,y.
* Auto-positioning is performed during game update, and new x,y parameters are assigned.
* Overlay.X,Y properties now get and set same values, corresponding to overlay's layer (screen or room).

**POTENTIAL TODO:**
- I suppose the next logical step could be to remove dependency on "offsets". Right now the assigned position corresponds to the position of "text" in textual overlays, and any borders (from TextWindow) are positioned "outside". Historical reference: #1098, #1104.
So the question is whether this should be changed to have position mark exactly top-left corner of the generated image, including borders, or not.
**Regardless**, X,Y properties must assign and return same position (either accounting or not accounting these borders).

EDIT: sudden thought: maybe these offsets could be redesigned as "origin" parameter, which was planned (potentially) for all objects in the future ags4.

**TESTING**
- [ ] Display boxes should not have any changes in behavior.
- [ ] Non-blocking speech should position correctly above character even if it walks around; also - in scrolling rooms.
- [ ] Blocking speech should position correctly above character; also - in scrolling rooms.
- [ ] Setting X or Y property to a autopositioned overlay should disable its automatic behavior. **I forgot to do this**, needs to be amended.
- [ ] Overlay.X and Y properties must always return the value that was set to it by user, and always in coordinates consistent with the overlay's layer: screen overlays use screen coords, room overlays use room coords.

### DIFF
--- a/Engine/ac/screenoverlay.cpp
+++ b/Engine/ac/screenoverlay.cpp
@@ -103,7 +103,7 @@ void ScreenOverlay::ReadFromSavegame(Stream *in, bool &has_bitmap, int32_t cmp_v
     x = in->ReadInt32();
     y = in->ReadInt32();
     timeout = in->ReadInt32();
-    bgSpeechForChar = in->ReadInt32();
+    speechForChar = in->ReadInt32();
     associatedOverlayHandle = in->ReadInt32();
     if (cmp_ver >= kOverSvgVersion_36025)
     {
@@ -112,8 +112,9 @@ void ScreenOverlay::ReadFromSavegame(Stream *in, bool &has_bitmap, int32_t cmp_v
     else
     {
         in->ReadBool(); // [DEPRECATED] has alpha
-        if (!(in->ReadBool())) // screen relative position
-            _flags |= kOver_PositionAtRoomXY;
+        // "is screen relative pos" flag; historically room relative == autopos
+        if (!(in->ReadBool()))
+            _flags |= kOver_AutoPosition;
     }
 
     if (cmp_ver >= kOverSvgVersion_35028)
@@ -172,7 +173,7 @@ void ScreenOverlay::WriteToSavegame(Stream *out) const
     out->WriteInt32(x);
     out->WriteInt32(y);
     out->WriteInt32(timeout);
-    out->WriteInt32(bgSpeechForChar);
+    out->WriteInt32(speechForChar);
     out->WriteInt32(associatedOverlayHandle);
     out->WriteInt16(_flags);
     // since cmp_ver = 1
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -2575,10 +2575,10 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
     int tdxp = xx,tdyp = yy;
     int oldview=-1, oldloop = -1;
     int ovr_type = 0;
+    int autoplace_at_char = -1; // depends on speech style below
     text_lips_offset = 0;
     text_lips_text = texx;
     Bitmap *closeupface = nullptr;
-    bool overlayPositionFixed = false;
     int charFrameWas = 0;
     int viewWasLocked = 0;
     if (speakingChar->flags & CHF_FIXVIEW)
@@ -2851,7 +2851,6 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
             if (facetalkchar->blinktimer < 0)
                 facetalkchar->blinktimer = facetalkchar->blinkinterval;
             textcol=-textcol;
-            overlayPositionFixed = true;
             // Process the first portrait view frame
             const int frame_vol = charextra[facetalkchar->index_id].GetFrameSoundVolume(facetalkchar);
             CheckViewFrame(facetalkview, facetalkloop, facetalkframe, frame_vol);
@@ -2859,6 +2858,9 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
         else if (useview >= 0) {
             // Lucasarts-style speech
             set_our_eip(154);
+
+            // autoplace at character
+            autoplace_at_char = aschar;
 
             oldview = speakingChar->view;
             oldloop = speakingChar->loop;
@@ -2903,8 +2905,9 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
 
         }
     }
-    else
+    else {
         allowShrink = 1;
+    }
 
     // it wants the centred position, so make it so
     if ((xx >= 0) && (tdxp < 0))
@@ -2918,7 +2921,7 @@ void _displayspeech(const char*texx, int aschar, int xx, int yy, int widd, int i
         char_thinking = aschar;
 
     set_our_eip(155);
-    display_main(tdxp, tdyp, bwidth, texx, nullptr, DISPLAYTEXT_SPEECH, FONT_SPEECH, textcol, isThought, allowShrink, overlayPositionFixed);
+    display_main(tdxp, tdyp, bwidth, texx, nullptr, DISPLAYTEXT_SPEECH, FONT_SPEECH, textcol, isThought, allowShrink, autoplace_at_char);
     set_our_eip(156);
     if ((play.in_conversation > 0) && (game.options[OPT_SPEECHTYPE] == 3))
         closeupface = nullptr;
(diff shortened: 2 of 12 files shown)


Engine/ac/screenoverlay.cpp
```diff
@@ -112,8 +112,8 @@ void ScreenOverlay::ReadFromSavegame(Stream *in, bool &has_bitmap, int32_t cmp_v
     else
     {
         in->ReadBool(); // [DEPRECATED] has alpha
-        if (!(in->ReadBool())) // screen relative position
-            _flags |= kOver_PositionAtRoomXY;
+        if (!(in->ReadBool())) // logically room relative == autoplace
```
ericoporto, line 115: I think the comment here as it was made more sense because of the negation there? Also you are removing the `OVR_AUTOPLACE` and using autoposition as the new name and commenting autoplace. If I understood this is an old parameter that was removed in a new savegame update but the comment is being less helpful than the code itself.

ericoporto:
oh, this is pretty cool! Amazing work!

Fun looking into `autoposition_overlay` logic because I sort of did it using AGS Script [here in fancy](https://github.com/ericoporto/fancy/blob/83681075d18cbf11932e8436249417a6bff85e70/fancy_demo/fancy.asc#L48-L82). I was dumb and did by trial and error - forgot I could look into the code.

Engine/ac/screenoverlay.cpp
ivan-mogilko: Ok, I changed to 
```
// "is screen relative pos" flag; historically room relative == autopos
```

ericoporto: thanks! Reads easier now!

ivan-mogilko:
I had to add a fix: overlay position has to be updated right before the drawing stage, but after the camera update, because autopositioned overlays depend on both character pos and scale AND camera.

Moved out camera update out of drawing function.

ericoporto:
I tested lucasarts, sierraTransparent, sierraWithBackground and Wholescreen types of speech and all four appear to work correctly: the returned x and y is the same x and y I set.

> Setting X or Y property to a autopositioned overlay should disable its automatic behavior. I forgot to do this, needs to be amended.

You made this comment above, but either you already fixed this or I can't figure what it means.

ivan-mogilko:
> You made this comment above, but either you already fixed this or I can't figure what it means.

Yes, this should be already fixed.

ivan-mogilko:
I merged this too soon; there is a problem with blocking speech which I've been planning to investigate, but forgot about it.

(Also, there is at least one case which I outright missed: such as SayAt, which must not set autoposition. But that should be easy to fix.)

The reason why blocking speech was storing position in room coordinates was to keep it in one place if camera scrolls while it's displayed (this may be achieved in "repeatedly_execute_always").

At first I thought that this will be solved by setting blocking speech to "Autoposition", but this turned out to be not trivial, because blocking speech has different positioning rules, compared to non-blocking ones, in terms of its offset from the top of character's "head", and possibly some other things, such as width limit, and centering. There's also "Thought" mode, which has extra difference in positioning (it's placed higher above the head than the speech).

There's also another quirk: since blocking speech is defined in room coordinates and never updated, then, if camera is moving around, speech's position is retained, causing it to move outside of view, as if it were a room overlay. I suppose this is not intentional, and may be considered a bug, so preserving this may not be a good idea.

This leaves the issue with difference in position. I have never investigated how it is calculated, and why it's different between blocking, non-blocking speech, and "thinking", so I have no idea what is a good thing and not, and whether it makes sense to unify these rules. But I don't think any such change should be a part of this overlay fix.

I am trying to figure out the easiest fix, that would let keep at least positioning rule. That is, not recalculate whole thing with overlay width etc, but only recalculate x,y position in relation to the character and camera, while keeping overlay's size the same.

The important nuance here is that I *do not want to add more parameters to Overlay itself*, rather than the link to character that it already has. Overlays should stay as simple as possible.
So, perhaps, the better idea would be to add the "current speech" parameters to either a speaking character, or a global game state instead. Since these belong strictly to the blocking speech (for now), they do not have to be in game saves too.

ivan-mogilko:
Seeing how complicated the coordinate math is in these functions, I'm also considering an alternative:
1. Restore kOver_PositionAtRoomXY flag (kOver_AutoPosition will be kept of course, only have its value changed).
2. Blocking speech has kOver_PositionAtRoomXY and not kOver_AutoPosition, and stores its position in room coords internally.
3. X,Y properties of blocking speech return screen coords when read. When user writes these, the kOver_PositionAtRoomXY flag **is removed** (just like kOver_AutoPosition is removed), and it becomes a proper screen overlay.

ivan-mogilko:
Ah, but maybe I am overthinking this again.

If blocking speech is fixed, either in room coords (old behavior) or screen coords (supposed new behavior), there will always be this problem that if Character itself is moved (by directly changing its position in "rep-exec-always"), the speech location is not updated. It's been like that before, it will be like that now. The difference is only that with old behavior moving camera will move the speech (but it may also move offscreen). With the new behavior the speech will always stay on same place on screen. Maybe it's a better way, if choosing between these two?

ericoporto:
I don't recall observing this behavior of the speech moving when it's onscreen in any old game, not sure if all conversations people have always made them static enough to not have that behavior observed. If the speech moves it feels it's hard to read - unless that is the intent. At least from the sierra with view I always expect it to be static on screen, but I didn't play enough lucasarts style games to know the intended behavior in that case.

ivan-mogilko:
Ok, I pushed 002ab503355aa7b2928446cc64c2ccdd80647cad.

PR #2396, opened 2024-04-28
3.6.1: fix unsafe buffer use in String.Replace and removed result limit of 2048 bytes

ivan-mogilko:
Somehow this problem was ignored after all these years (even though the String API was updated with Unicode support).

The implementation of script function String.Replace was limiting resulting string to 2048 bytes, and furthermore used that buffer without any size checks. This may potentially lead to a buffer overflow and memory corruption.

NOTE: buffer limit has been 3000 originally, but reduced to 2048 in the latest master branch.

Example test case:
```
function TestLongReplace()
{
  String source = "REPLACE THIS BIT"; // test replace in the beginning
  while (source.Length < 4096)
  {
    source = source.Append("xxxxxxxxxxxxxxxx REPLACE THIS BIT xxxxxxxxxxxxxxxx");
  }
  source = source.Append("REPLACE THIS BIT"); // test replace in the end
  
  String dest1 = source.Replace("REPLACE THIS BIT", "THIS IS A NEW BIT", eCaseSensitive);
  String dest2 = source.Replace("replace this bit", "this is a new bit", eCaseInsensitive);
  
  File *f = File.Open("$SAVEGAMEDIR$/replace.txt", eFileWrite);
  f.WriteRawLine(source);
  f.WriteRawLine(dest1);
  f.WriteRawLine(dest2);
  f.Close();
}
```

Above will crash the engine in any past version of AGS (most likely).

This PR should fix the above for both ASCII and Unicode text format setting.

### DIFF
--- a/Engine/ac/string.cpp
+++ b/Engine/ac/string.cpp
@@ -160,51 +160,42 @@ int String_EndsWith(const char *thisString, const char *checkForString, bool cas
 
 const char* String_Replace(const char *thisString, const char *lookForText, const char *replaceWithText, bool caseSensitive)
 {
-    char resultBuffer[STD_BUFFER_SIZE] = "";
-    size_t outputSize = 0; // length in bytes
-    if (caseSensitive)
-    {
-        size_t lookForLen = strlen(lookForText);
-        size_t replaceLen = strlen(replaceWithText);
-        for (const char *ptr = thisString; *ptr; ++ptr)
-        {
-            if (strncmp(ptr, lookForText, lookForLen) == 0)
-            {
-                memcpy(&resultBuffer[outputSize], replaceWithText, replaceLen);
-                outputSize += replaceLen;
-                ptr += lookForLen - 1;
-            }
-            else
-            {
-                resultBuffer[outputSize] = *ptr;
-                outputSize++;
-            }
-        }
-    }
-    else
-    {
-        size_t lookForLen = ustrlen(lookForText);
-        size_t lookForSz = strlen(lookForText); // length in bytes
-        size_t replaceSz = strlen(replaceWithText); // length in bytes
-        const char *p_cur = thisString;
-        for (int c = ugetxc(&thisString); *p_cur; p_cur = thisString, c = ugetxc(&thisString))
-        {
-            if (ustrnicmp(p_cur, lookForText, lookForLen) == 0)
-            {
-                memcpy(&resultBuffer[outputSize], replaceWithText, replaceSz);
-                outputSize += replaceSz;
-                thisString = p_cur + lookForSz;
-            }
-            else
-            {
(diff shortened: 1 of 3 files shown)


ericoporto:
I tested a few regular cases and additionally:

- Works correct if replace string contains string
```
String mystr = "abab";
String adjusted_mystr = mystr.Replace ("ab", "abab");
bool ok = adjusted_mystr == "abababab";
```
- Tested if either loofForText or replaceText are given `String empty;`, the engine will error on null string, but correctly.
- I tested replacing emoji with regular characters (mostly because I didn't have a font with emoji at hand to display), and it appears to work too.

ivan-mogilko:
Sorry, something was bugging me about this function, so I rewrote it into a somewhat faster variant, which uses a function pointer to select a function that searches for a string (between `strstr` and unicode-aware no-case search).

ericoporto:
Not for here, but in the future I think we should move the unicode stuff in allegro out of allegro, in some util in common or something. It's just weird to continuously add new code there.

ivan-mogilko:
Yes, that is true.

ericoporto:
sorry for the offtopic again, but I checked now and it appears there is a lot that is interwoven with the allegro file stuff, so it isn't as trivial. There are some functions there that looks like ok to remove (`uatof`, `ustrtol`, `ustrtod` which I guess are useless, and , `register_uformat`, `set_ucodepage`, I think we will never use, and `uinsert`, which for some reason we don't use).

Anyway, reran my tests in the new implementation and it worked alright.

ivan-mogilko:
> Ah, I noticed that the code around uses a lot of CreateNewScriptString(std::move(buf)), that function looks like the only one with similar buffer not using this, not sure is needed, just commenting about it.

This is meant for a StringBuffer allocated when the final size is known. I suppose it's possible to run Replace in two steps, first calculating the final size, then writing; but I cannot tell which will be faster.

ivan-mogilko:
Ok, I did some speed tests. There's no clear "winner", it depends on a combination on factors.

With small strings the difference seem negligible.
With long strings (> 1024 long), in general, preliminary counting works pretty much faster with case-sensitive replace, but around x1.5 times slower with case-insensitive replace, because it has to do case-insensitive compare which has its quirks in with unicode strings.

Perhaps, I will keep the pre-counting variant, as case-sensitive replace sounds as a more common use-case with long strings....?

ivan-mogilko:
I'm going little crazy with all the variants, but now it is the new variant, which saves match locations before allocating a buffer and then uses the recorded match positions to copy data over.

Tried few different tests, and this proved to be the fastest in all of them. Sometimes it's noticeably faster, sometimes about the same. Even though it still allocates some memory dynamically, but since it copies a bunch of ints, that does not affect performance as much.

PR #2399, opened 2024-04-30
Hotfix function pointer assignment in String_Replace independent of libc

ivan-mogilko:
Various implementations of libc may have non-standard overloads of `strstr`, making it impossible to use static_cast or reinterpret_cast with them directly (either would work with one libc, but may fail with another).

Experiment to see which approach may work here.

### DIFF
--- a/Common/util/string_compat.h
+++ b/Common/util/string_compat.h
@@ -30,4 +30,10 @@ char *ags_strdup(const char *s);
 }
 #endif
 
+#if defined (__MINGW32__)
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Hmm, not sure how truly "universal" this fix is, but at least it builds for all the compilers in our CI.

ivan-mogilko:
I made a mistake again, only realizing after merging this. This was not a correct fix, it's not proper to base this selection on declared compilation flags, because it's not reliable nor universal.

Reverted this fix again.

PR #2401, opened 2024-05-02
[ags4] add string Split and Join

ericoporto:
adds String Split - it's working afaict. I am leaving as a draft because I would like to add String Join but I don't know yet how to read a String array.

### DIFF
--- a/Engine/ac/dynobj/cc_dynamicarray.cpp
+++ b/Engine/ac/dynobj/cc_dynamicarray.cpp
@@ -189,6 +189,21 @@ DynObjectRef DynamicArrayHelpers::CreateStringArrayFromBuffers(std::vector<Scrip
     return arr;
 }
 
+DynObjectRef DynamicArrayHelpers::CreateScriptArray(std::vector<DynObjectRef> &&items)
+{
+    DynObjectRef arr = globalDynamicArray.CreateOld(items.size(), sizeof(int32_t), true);
+    if (!arr.Obj)
+        return arr;
+
+    int32_t *slots = static_cast<int32_t*>(arr.Obj);
+    for (auto const& obj : items)
+    {
+        ccAddObjectReference(obj.Handle);
+        *(slots++) = obj.Handle;
+    }
+    return arr;
+}
+
 bool DynamicArrayHelpers::ResolvePointerArray(const void* arrobj, std::vector<void*> &objects)
 {
     objects.clear();
--- a/Engine/ac/dynobj/cc_dynamicarray.cpp
+++ b/Engine/ac/dynobj/cc_dynamicarray.cpp
@@ -171,6 +171,23 @@ DynObjectRef DynamicArrayHelpers::CreateStringArray(const std::vector<const char
     return arr;
 }
 
+DynObjectRef DynamicArrayHelpers::CreateStringArrayFromBuffers(std::vector<ScriptString::Buffer> &&items)
+{
+    DynObjectRef arr = globalDynamicArray.CreateOld(items.size(), sizeof(int32_t), true);
+    if (!arr.Obj)
+        return arr;
+    // Create script strings and put handles into array
+    int32_t *slots = static_cast<int32_t*>(arr.Obj);
+    for (auto &buf : items)
+    {
+        DynObjectRef str = ScriptString::Create(std::move(buf));
+        // We must add reference count, because the string is going to be saved
+        // within another object (array), not returned to script directly
+        ccAddObjectReference(str.Handle);
+        *(slots++) = str.Handle;
+    }
+    return arr;
+}
 
 bool DynamicArrayHelpers::ResolvePointerArray(const void* arrobj, std::vector<void*> &objects)
 {
--- a/Engine/ac/dynobj/cc_dynamicarray.h
+++ b/Engine/ac/dynobj/cc_dynamicarray.h
@@ -17,6 +17,7 @@
 #include <vector>
 #include <unordered_map>
 #include "ac/dynobj/cc_agsdynamicobject.h"
+#include "scriptstring.h"
 
 
 #define ARRAY_MANAGED_TYPE_FLAG    0x80000000
@@ -89,8 +90,10 @@ extern CCDynamicArray globalDynamicArray;
 // Helper functions for setting up dynamic arrays.
 namespace DynamicArrayHelpers
 {
-    // Create array of managed strings
+    // Create array of managed strings from strings that exists somewhere
     DynObjectRef CreateStringArray(const std::vector<const char*>);
+    // Create array of managed strings from script strings
+    DynObjectRef CreateStringArrayFromBuffers(std::vector<ScriptString::Buffer> &&);
     // Resolves a dynamic array of managed pointers (handles) to a list of object addresses
     bool ResolvePointerArray(const void* arrobj, std::vector<void*> &objects);
     // Resolves a dynamic array of managed pointers (handles) to a list of DynObjectRef structs,
--- a/Engine/ac/dynobj/cc_dynamicarray.h
+++ b/Engine/ac/dynobj/cc_dynamicarray.h
@@ -94,6 +94,8 @@ namespace DynamicArrayHelpers
     DynObjectRef CreateStringArray(const std::vector<const char*>);
     // Create array of managed strings from script strings
     DynObjectRef CreateStringArrayFromBuffers(std::vector<ScriptString::Buffer> &&);
+    // Create array from managed list of dynamic objects
+    DynObjectRef CreateScriptArray(std::vector<DynObjectRef> &&items);
     // Resolves a dynamic array of managed pointers (handles) to a list of object addresses
     bool ResolvePointerArray(const void* arrobj, std::vector<void*> &objects);
     // Resolves a dynamic array of managed pointers (handles) to a list of DynObjectRef structs,
--- a/Engine/ac/string.cpp
+++ b/Engine/ac/string.cpp
@@ -49,6 +49,64 @@ int String_IsNullOrEmpty(const char *thisString)
     return 0;
 }
 
+const char * String_Join(void* arrobj, const char *separator)
+{
+    if(arrobj == nullptr)
+        return CreateNewScriptString("");
+
+    std::vector<void*> items{};
+    if (!DynamicArrayHelpers::ResolvePointerArray(arrobj, items))
+    {
+        // somehow this is not a managed array, this would have to be a compiler error
+        return nullptr;
+    }
+
+    const size_t count = items.size();
+    if(count == 0)
+        return CreateNewScriptString("");
+
+    if(separator == nullptr) {
+        // I am not sure this can actually happen or the compiler will prevent this
+        separator = "";
+    }
+
+    int sep_len, sep_ulen;
+    ustrlen2(separator, &sep_len, &sep_ulen);
+
+    int total_len, total_ulen;
+    total_len = 0;
+    total_ulen = 0;
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        auto address = static_cast<const char*>(items[i]);
+        if (address != nullptr) {
+            auto const& sh = ScriptString::GetHeader(address);
+            total_len += sh.Length;
+            total_ulen += sh.ULength;
+        }
+    }
+    total_len += sep_len * (count - 1);
+    total_ulen += sep_ulen * (count - 1);
+    
+    auto buf = ScriptString::CreateBuffer(total_len, total_ulen);
+
+    // concatenate the array elements with a separator between elements: el1+sep+el2+sep...elN
+    char* p = buf.Get();
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        auto address = static_cast<const char*>(items[i]);
+        if (address != nullptr) {
+            auto const& sh = ScriptString::GetHeader(address);
+            p = std::copy(address, address +sh.Length, p);
+        }
+        if (i != count - 1) {
+            p = std::copy(separator, separator + sep_len, p);
+        }
+    }
+    buf.Get()[total_len] = 0; // Null-terminate joint string
+    return CreateNewScriptString(std::move(buf));
+}
+
 const char* String_Copy(const char *srcString) {
     return CreateNewScriptString(srcString);
 }
@@ -425,6 +483,12 @@ RuntimeScriptValue Sc_String_IsNullOrEmpty(const RuntimeScriptValue *params, int
     API_SCALL_INT_POBJ(String_IsNullOrEmpty, const char);
 }
 
+// const char * (void* arrobj, const char *separator)
+RuntimeScriptValue Sc_String_Join(const RuntimeScriptValue *params, int32_t param_count)
+{
+    API_SCALL_OBJ_POBJ2(const char, myScriptStringImpl, String_Join, void, const char);
+}
+
 // const char* (const char *thisString, const char *extrabit)
 RuntimeScriptValue Sc_String_Append(void *self, const RuntimeScriptValue *params, int32_t param_count)
 {
@@ -564,6 +628,7 @@ void RegisterStringAPI()
     ScFnRegister string_api[] = {
         { "String::IsNullOrEmpty^1",  API_FN_PAIR(String_IsNullOrEmpty) },
         { "String::Format^101",       Sc_String_Format, ScPl_String_Format },
+        { "String::Join^2",           API_FN_PAIR(String_Join) },
 
         { "String::Append^1",         API_FN_PAIR(String_Append) },
         { "String::AppendChar^1",     API_FN_PAIR(String_AppendChar) },
--- a/Engine/ac/string.cpp
+++ b/Engine/ac/string.cpp
@@ -28,6 +28,7 @@
 #include "script/runtimescriptvalue.h"
 #include "util/string_compat.h"
 #include "util/utf8.h"
+#include "ac/dynobj/cc_dynamicarray.h"
 
 using namespace AGS::Common;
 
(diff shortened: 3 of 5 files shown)


ivan-mogilko:
> but I don't know yet how to read a String array.

String dynamic array is an array of managed handles; these are 32-bit values, which may be resolved to a String object using `ccGetObjectAddressFromHandle`.

Engine/ac/string.cpp
```diff
@@ -257,6 +258,55 @@ const char* String_UpperCase(const char *thisString) {
...
+        buf.Get()[len] = 0;
+        buflist.push_back(std::move(buf));
+        items.push_back(buflist.back().Get());
+
+        ptr = found_cstr + seplen;
+    }
+
+    // this takes a vector of const char*, which means null terminated array of strings
+    // null terminated means we have to hold copies instead of references in-string
+    DynObjectRef arr = DynamicArrayHelpers::CreateStringArray(items);
```
ivan-mogilko, line 306: This DynamicArrayHelpers::CreateStringArray was originally implemented for strings retrieved from somewhere else, but since you already create ScriptString::Buffers here, you can add another variant, which accepts `std::vector<ScriptString> &&` and moves this data to ScriptString::Create().

Engine/ac/string.cpp
```diff
@@ -257,6 +258,55 @@ const char* String_UpperCase(const char *thisString) {
...
+        if (!found_cstr) {
+            found_cstr = end;
+        }
+
+        len = found_cstr - ptr;
+        auto buf = ScriptString::CreateBuffer(len, 0);
+        std::copy(ptr, found_cstr, buf.Get());
+        buf.Get()[len] = 0;
+        buflist.push_back(std::move(buf));
+        items.push_back(buflist.back().Get());
```
ivan-mogilko, line 299: This code not make much sense, `ScriptString::CreateBuffer` is meant for creating a specialized memory buffer holding scriptstring's meta-data. The purpose is to have a buffer that may be directly moved to ScriptString::Create. If you want only copies of strings, then create simple Strings.

Although here you'd rather make a std::vector<ScriptString::Buffer> and move it further, see another comment below.

ivan-mogilko:
Looking at the draft code, I think you misunderstand the purpose of ScriptString::Buffer. This Buffer object is meant for the case when you need to fill it in multiple steps, copying parts of other strings over, as opposed to a single string assignment.

Second thing is, this Buffer is for passing directly into ScriptString::Create, avoiding extra copies and length recalculation. Having it as a temporary string storage but not moved further defeats its purpose. For that you could using simple Strings (and they are easier to assign to, no need to use std::copy).

---

But here, for your purpose, it would make more sense to either
1. Write another variant of `DynamicArrayHelpers::CreateStringArray` that accepts `std::vector<ScriptString::Buffer> &` and std::moves its members to `ScriptString::Create` calls.
2. Write another variant of `DynamicArrayHelpers::CreateArray` that accepts `std::vector<DynObjectRef> &`; create ScriptStrings here directly using ScriptString::Create, and then pass DynObjectRefs into CreateArray, where it makes array, and stores handles in there.

Engine/ac/string.cpp
ivan-mogilko: Another alternative would be a variant of DynamicArrayHelpers::CreateArray that accepts a generic std::vector<DynObjectRef> &`. In this case you could create ScriptStrings here directly using ScriptString::Create, and then pass DynObjectRefs into CreateArray, where it makes array, and stores handles in there.

I cannot say which is more convenient at the moment.

Engine/ac/string.cpp
```diff
@@ -257,6 +258,55 @@ const char* String_UpperCase(const char *thisString) {
...
+    {
+        const char *found_cstr = strstr(ptr, separator);
+        if (!found_cstr) {
+            found_cstr = end;
+        }
+
+        len = found_cstr - ptr;
+        auto buf = ScriptString::CreateBuffer(len, 0);
+        std::copy(ptr, found_cstr, buf.Get());
+        buf.Get()[len] = 0;
```
ivan-mogilko, line 297: I think this could be simplified by using a single `snprintf(buf.Get(), "%s", ptr, len + 1);`

This also makes me wonder why I haven't added an assignment operator to ScriptString::Buffer, which would just copy an input string over...

Engine/ac/string.cpp
ericoporto: > since you already create ScriptString::Buffers here, you can add another variant,

I can't figure how to iterate the buffer array from this because I keep getting the following error:

> Call to implicitly-deleted copy constructor of 'ScriptString::Buffer' copy constructor is implicitly deleted because 'Buffer' has a user-declared move constructor

> you could create ScriptStrings here directly using ScriptString::Create, and then pass DynObjectRefs into CreateArray

This worked, so I went this way.

ericoporto:
> String dynamic array is an array of managed handles; these are 32-bit values, which may be resolved to a String object using `ccGetObjectAddressFromHandle`.

Thanks, I think I got it, I will try to update in the weekend with Join.

Engine/ac/string.cpp
ericoporto: Oh so it was double ampersand in the call like `&&items`, that's why it didn't work. :O

ericoporto: Sorry, decided to fix that "FIX-ME" and I tried the following:

```
    String type_str = !items.empty() ? items[0].Mgr->GetType() : "";
    auto type = std::find_if(ccInstance::GetRTTI()->GetTypes().begin(), ccInstance::GetRTTI()->GetTypes().end(),
                             [type_str](const RTTI::Type& i) -> bool {return type_str.Equals(i.name); });
```

This works and `type` correctly finds the type "String" in the "_BuiltInScriptHeader.ash", but the type size is 0 - instead of `sizeof(int32_t)`, so I can't use `CreateNew` because I don't understand how to correctly get the size dynamically.

ivan-mogilko: The thing is, RTTI may be missing. It has to be decided, whether this API actually requires type info or not.

ivan-mogilko: For faster searching for a type there's a lookup table - ccInstance::_rttiLookup, but it may be not exposed right now. May be worth adding an inlined function that returns a constant reference to this table.

ericoporto: It's alright, it is working with the CreateOld method, I can update that in a new PR. But thanks for the information!

The find wasn't as slow. About "fully qualified", this means just "_BuiltInScriptHeader.ash::String" or something that can be determined like this right? If so, then it should be fine anyway, at least for internal usage.

ericoporto:
I got String.Join to work!!!

> String dynamic array is an array of managed handles; these are 32-bit values, which may be resolved to a String object using `ccGetObjectAddressFromHandle`.

Thanks for this because without it, I wouldn't get anywhere! I will leave as a draft still because I want to see if I can improve it's code still.

ericoporto:
I got code that works and I think the code in general is looking ok now - but any possible changes please tell me!!

Other things to review!

- Do we want this? I feel these are useful extensions, but it's good to check how everyone feels.
- is the API ok? Please check the names of things in the API, parameter order, ...
- @AlanDrake suggested adding a flag for Trim in String.Split too, but maybe this would complicate the code significantly - it would need to be trimmed before allocating the buffer but using a specific function that did not copy the buffer but instead selected things... I liked the idea at first but now I would prefer this to be done in AGS Script as a String extension
  - I feel some additional string stuff could be in a script library of sorts...

Engine/ac/string.cpp
ivan-mogilko: About this, type's size is 0 because there are no plain fields in it. RTTI only knows the fields declared in script, and there are none in String.

But when you make dynamic array of pointers, you don't need type's size at all, as you don't store type itself in the array, you store pointers to that type. So you ever need only size of pointer. If you were making a dynamic array of primitive types or plain (non-managed) structs, then you would have to get their size.

ericoporto: Right, I think this means I need to check the ARRAY_MANAGED_TYPE_FLAG in this case. This flag is found in the type_id here in ags4 and was found in the elementCount in ags3 I think.

I think I understand now how to do it but I think I will leave this for later.

ivan-mogilko:
> @AlanDrake suggested adding a flag for Trim in String.Split too, but maybe this would complicate the code significantly - it would need to be trimmed before allocating the buffer but using a specific function that did not copy the buffer but instead selected things...

I don't think it would complicate much. You are already using 2 pointers: for beginning of a string and end of string. Trimming the source would mean moving these pointers right and left correspondingly.

Engine/ac/string.cpp
```diff
@@ -49,6 +49,67 @@ int String_IsNullOrEmpty(const char *thisString)
     return 0;
 }
 
+void calculate_joint_length(const size_t count, const uint32_t *&handles, int sep_len, int sep_ulen, int &total_len, int &total_ulen)
```
ivan-mogilko, line 52: I notice that you iterate over handles and do `ccGetObjectAddressFromHandle` on them twice in succession, first to calculate final length and second to actually do string copying. This seem suboptimal and complicates the code. You could instead make a function that gathers the resolved ScriptString's pointers into a std::vector, and then do whatever you want with them: get their headers, check lengths, then copy their contents.

Engine/ac/string.cpp
```diff
@@ -49,6 +49,67 @@ int String_IsNullOrEmpty(const char *thisString)
...
+
+    // concatenate the array elements with a separator between elements: el1+sep+el2+sep...elN
+    char* p = buf.Get();
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        auto address = (const char *) ccGetObjectAddressFromHandle(handles[i]);
+        if (address != nullptr) {
+            auto const& sh = ScriptString::GetHeader(address);
+            memcpy(p, address, sh.Length);
+            p += sh.Length;
```
ivan-mogilko, line 102: if you use std::copy, it returns iterator after copied fragment, letting you write slightly less lines:
```p = std::copy(address, address +sh.Length, p);```

File: Engine/ac/string.cpp [hunk 6]
```diff
@@ -49,6 +49,67 @@ int String_IsNullOrEmpty(const char *thisString)
...
+    total_ulen = 0;
+    for (uint32_t i = 0; i < count; ++i)
+    {
+        auto address = ccGetObjectAddressFromHandle(handles[i]);
+        if (address != nullptr) {
+            auto const& sh = ScriptString::GetHeader(address);
+            total_len += sh.Length;
+            total_ulen += sh.ULength;
+        }
+        if (i != count - 1) {
```
ivan-mogilko, line 64: I think you may simplify this by not adding separator lengths in a loop, but add them once afterwards, since this only depends on the `count`:
```
total_len += sep_len * (count - 1);
total_ulen += sep_ulen * (count - 1);
```

Engine/ac/string.cpp
ericoporto: good, I didn't knew this, changed but didn't push force - will squash commits things later

Engine/ac/string.cpp
ericoporto: that's true, I didn't notice that when writing the code.

Engine/ac/string.cpp
ericoporto: I got rid of the function but wasn't sure on making it a function because then I would iterate three times on the strings, because I still need to figure the length before I can create a buffer...

ivan-mogilko:
There's a curious question about null arguments in Join. I noticed that passing null array would result in returning null pointer, but having null Strings in that array have same effect as empty strings. Which means that if I pass an array of only nulls, that would return an actual string, which may be empty or containing a sequence of separators.

Just a observation...

Engine/ac/string.cpp
```diff
@@ -257,6 +258,52 @@ const char* String_UpperCase(const char *thisString) {
...
+{
+    if ((thisString == nullptr) || (thisString[0] == 0))
+        return nullptr;
+
+    auto const &header = ScriptString::GetHeader((void*)thisString);
+
+    if ((separator == nullptr) || (separator[0] == 0)) {
+        std::vector<const char*> items{};
+        items.push_back(thisString);
+        DynObjectRef arr = DynamicArrayHelpers::CreateStringArray(std::vector<const char*>(items));
```
ivan-mogilko, line 327: This is probably a typo, should be passing `items`, because right now you are constructing a new temporary vector.

ivan-mogilko:
Well, I find mostly minor things that may be optimized. I will be testing this properly tomorrow, but I guess it should be working.

Engine/ac/string.cpp
ivan-mogilko: Hmm... well, the function itself would iterate over handles and call ccGetObjectAddressFromHandle in order to gather strings, that may come handy later. Other 2 loops would be over a ready vector of string pointers, I don't think that's too bad.

Engine/ac/string.cpp
ericoporto: thanks! This was leftover from a code that did all in one line - but looked ugly...

Engine/ac/string.cpp
```diff
@@ -49,6 +49,62 @@ int String_IsNullOrEmpty(const char *thisString)
...
+    if(separator == nullptr)
+        separator = "";
+
+    const uint32_t *handles;
+    handles= static_cast<const uint32_t*>(arrobj);
+
+    int sep_len, sep_ulen;
+    ustrlen2(separator, &sep_len, &sep_ulen);
+
+    std::vector<ScriptString*> script_strings{};
```
ivan-mogilko, line 71: This is not correct, the ScriptString is only a manager, but the registered objects are buffers of type `char*`.

ivan-mogilko:
I need to honestly say that I will want to pick out an algorithm of resolving handles to object pointers as a separate function, that may be used in any API that has a dynamic array argument. I believe that's better to have this algorithm standalone rather than to mix it with data processing.

Engine/ac/string.cpp
ericoporto: Ah, this is why the autocomplete was making no sense with the data.

ericoporto:
If you write this function and put it somewhere in ags4 I can rebase and adjust the code to use it.

ivan-mogilko:
> If you write this function and put it somewhere in ags4 I can rebase and adjust the code to use it.

Okay, I wrote couple of variants, for cases when you do and do not need to know actual "managers": 387aa5e8ed8844e1e379f9dd4fb63749fb97284c

ericoporto:
@ivan-mogilko  I tried to adjust in the commit https://github.com/adventuregamestudio/ags/pull/2401/commits/99d408038b6912c2219e11f17c5a86e6ff988834

> EDIT: in regards to this PR, I tested that String.Join correctly recreates a string after String.Split, and it works as expected.

This exact case is failing for me in my commit, the `assert(header.IsPointerArray());` is failing, so I think I did something wrong building the string. :/

ivan-mogilko:
I just tried your latest commit, and it still works for me. It looks 99% like my own test (the only difference is that some typecast is not needed).

Can you post the script you're using for a test?

ericoporto:
ah, crap, that is exactly it, I recompiled the game but did not place in the right folder...

Engine/ac/dynobj/cc_dynamicarray.cpp
```diff
@@ -189,6 +189,28 @@ DynObjectRef DynamicArrayHelpers::CreateStringArray(std::vector<ScriptString::Bu
     return arr;
 }
 
+DynObjectRef DynamicArrayHelpers::CreateScriptArray(const std::vector<DynObjectRef> &items, const char* type_name)
+{
+    String type_str = type_name;
+    auto type = std::find_if(ccInstance::GetRTTI()->GetTypes().begin(), ccInstance::GetRTTI()->GetTypes().end(),
+                             [type_str](const RTTI::Type& i) -> bool {return type_str.Equals(i.name); });
+
+    uint32_t size = type->size == 0 ? sizeof(int32_t) : type->size;
```
ivan-mogilko, line 198: This is a wrong size calculation. If you are creating array of pointers, that's always int32_t, regardless of whether type has size >0 or not. On another hand, size == 0 does not indicate a pointer storage.

Engine/ac/dynobj/cc_dynamicarray.h
```diff
@@ -93,7 +93,9 @@ namespace DynamicArrayHelpers
     // Create array of managed strings from strings that exists somewhere
     DynObjectRef CreateStringArray(const std::vector<const char*>);
     // Create array of managed strings from script strings
-    DynObjectRef CreateStringArray(std::vector<ScriptString::Buffer> &&);
+    DynObjectRef CreateStringArrayFromBuffers(std::vector<ScriptString::Buffer> &&);
```
ivan-mogilko, line 96: There's really no need to have different name, it was fine as a function overload.

ivan-mogilko:
Please note, that you may also use CCDynamicArray::CreateOld if you don't want to bother with type names.
For array of pointers you simply can pass "is_managed" as true.

Engine/ac/dynobj/cc_dynamicarray.cpp
ericoporto: I adjusted it here https://github.com/adventuregamestudio/ags/pull/2401/commits/82488f296cb4e62b12933d7a82615fde1e0220d4 , I think dynamically in the api we will always be creating a dynamic array of pointers?

Engine/ac/dynobj/cc_dynamicarray.h
ericoporto: I had to remove the overload because the linter I use was not picking up the overload kept telling me this function was unused, I wasn't comfortable with it.

ivan-mogilko:
> fix crash if the string goes out of scope but not the array

I think I understand what the problem was.
You were not doing AddRef on the source string when you place it into array.

ivan-mogilko:
Could you elaborate the meaning of "fix scope crashes" commit?

Comment sais:
> fix crash if the string goes out of scope but not the array

Do you refer to C++ code, or AGS script?

ericoporto:
Sorry, I am readjusting things, but the problem is if you write in AGS Script something like this:
```AGS Script
  String splitted[];
  {
    String str = "text,to,split";
    splitted = str.Split(",");
  }
  for(int i=0; i<splitted.Length; i++)
  {
    System.Log(eLogInfo, "'%s'",  splitted[i]);
  }
```

ivan-mogilko:
The PR looks fine overall now. There's just one thing: since you added CreateArray helper, you could just use that instead of CreateStringArrayFromBuffers, if you'd create ScriptStrings from these buffers right away. Anyway, there's a room of improvement in the existing helper functions, and string factory methods.

ericoporto:
There's no rush to this and it doesn't block anything, so it's ok to take time. There was also no previous issue discussing it.

ivan-mogilko:
Well, I tried few script tests, and things appear working, including returning arrays from functions, splitting by multiple char sequences, etc. I may merge this now, and leave any possible code simplification for the later.

PR #2402, opened 2024-05-02
Implement `--no-plugins` command arg and config setting

ivan-mogilko:
Add `--no-plugins` command line argument, and related config setting, which tells to **not** load plugins from dynamic libraries.
Built-in plugins and stubs are still going to be used in this case (if they are available).

This may be used to either force engine to use built-in plugins, or as a safety measure to run engine without loading any extra binary.

### DIFF
--- a/Engine/plugin/agsplugin.cpp
+++ b/Engine/plugin/agsplugin.cpp
@@ -108,8 +108,8 @@ struct EnginePlugin
     EnginePlugin(const EnginePlugin&) = delete;
     EnginePlugin(EnginePlugin&&) = default;
 
-    AGS::Common::String  filename;
-    AGS::Engine::Library library;
+    String      filename;
+    Library     library;
     bool        available = false;
     std::vector<uint8_t> savedata;
     int         wantHook = 0;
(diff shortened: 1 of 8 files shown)


Engine/plugin/plugin_stubs.cpp
```diff
@@ -62,11 +62,12 @@ bool RegisterPluginStubs(const char* name)
     ccAddExternalStaticFunction("ShellExecute",                 Sc_PluginStub_Void);
     return true;
   }
-  else if (ags_stricmp(name, "agsappopenurl") == 0)
+  else if (ags_stricmp(name, "agsappopenurl") == 0 ||
+           ags_stricmp(name, "ags_SnowRain20") == 0)
```
ericoporto, line 66: I think this is a typo

ivan-mogilko: I put this new line into the wrong condition.

Engine/plugin/agsplugin.cpp
```diff
@@ -1061,47 +1082,31 @@ Engine::GameInitError pl_register_plugins(const std::vector<PluginInfo> &infos)
...
-            AGS::Common::Debug::Printf(kDbgMsg_Info, "Build-in plugin '%s' found and being used.", apl->filename.GetCStr());
-          }
-          else
-          {
-            // Plugin loading has failed at this point, try using built-in plugin function stubs
-            if (RegisterPluginStubs(apl->filename.GetCStr()))
-              AGS::Common::Debug::Printf(kDbgMsg_Info, "Placeholder functions for the plugin '%s' found.", apl->filename.GetCStr());
+            String expect_filename = apl->library.GetFilenameForLib(apl->filename);
+            Debug::Printf(kDbgMsg_Error, "Plugin '%s' could not be loaded (expected '%s'):\n\t%s",
+                apl->filename.GetCStr(), expect_filename.GetCStr());
```
ericoporto: Hey, this debug will crash things instantly because it's missing the last parameter.

ericoporto:
I fixed locally the `Debug::Printf` issue and then run and apparently it works, using `--no-plugins` correctly disables the plugins and removing loads them without a problem.

Issue #2313, opened 2024-01-25
Plugin API: add a CreateScriptArray method


ivan-mogilko:
When a plugin registers a script function that returns a `String`, it must use `IAGSEngine::CreateScriptString` in order to create an internal engine's ScriptString object, that engine knows how to deal with.

But there's no such function for creating a dynamic array object. This means that plugins cannot register script functions that return dynamic arrays.

Optionally same could be done for "user struct" object. The use case for this is when plugin wants to return a managed struct, but not manage that struct type itself: don't provide a type's manager, and have struct's data stored in engine's mem.

ericoporto:
The closest thing for this we already have is the `DynamicArrayHelpers` function to create string arrays - we use for retrieving keys from the dictionary and the set.

https://github.com/adventuregamestudio/ags/blob/ced74c55e6b24a6c64e6e8a76f75f925d7915cdc/Engine/ac/dynobj/cc_dynamicarray.cpp#L100-L117

ivan-mogilko:
> Don't know how to make this generic for any object as plugin interface. Take both void* and handle of each element?

If a plugin returns dynamic array, then it knows the type and size of elements. The engine only need to know the size of element, and its count, in order to create a buffer of necessary length. Plugin will be supposed to write the memory itself.

Since we support RTTI in ags4, then perhaps "type" could be another argument, but it's not clear to me how plugin may know which value to pass.

> We never in script take an array as input! So the reverse of receiving an array as input in a plugin would take more effort.

Plugin can receive arrays just like it can receive strings: a pointer to a buffer of memory.

PR #2403, opened 2024-05-02
Engine: replace a "plugin return value" hack with something appropriate

ivan-mogilko:
Necessary for #2313 (indirectly).

There is this very old hack which I invented while porting script VM to 64-bit back in 2012. When I've been rewriting script VM to work with RuntimeScriptValue struct, instead of a plain int32, plugins were becoming an issue, because there was no way to change the return values of their functions without modifying plugin API (and each existing plugin). And if we keep plugin functions cast to something returning an int, then we cannot know if that's a pointer, and which kind of a pointer that is (which exactly dynamic object type).

Back then I introduced a "temporary" hack: have a GlobalReturnValue object, which is assigned whenever plugin does *anything* related to the managed objects: create, register, resolve a handle, add reference, and so forth. Then *hope* that it is the value that the plugin is intending to return from its function. Silly as it is, this naive solution actually worked, for the actively used plugins at least. That's why it was forgotten about and never replaced.

But of course this is not a correct solution, simply because a plugin does not have to return a dynamic object that it touched last, it may return something completely different, previously saved instead.

The new solution I'm trying here is based on a simple but more logical assumption:
1. A plugin's return value is either a primitive type (int or float) or a memory address.
2. If it's a memory address, then this must be a managed object's address, as AGS script does not support plain pointers returned from a function (at least as of now).
3. If it's a managed object, then we may quickly search for it in the dynamic memory manager (it has a hashtable that resolves addresses to handles).

Therefore, we try the plugin function's return value as a address, and ask for its respective IScriptObject (dynamic object manager). If we succeeded, then we may assign a RuntimeScriptValue as a "Script Object". If we fail, we assign one as a "Plugin Argument", which may be anything else.

---

TEST LIST:

- [x] Win x32 engine
- [ ] Win x64 engine
- [ ] Linux x64 engine
- [ ] OSX x64 engine
- [ ] Android
- [ ] Performance test

### DIFF
--- a/Engine/ac/dynobj/dynobj_manager.cpp
+++ b/Engine/ac/dynobj/dynobj_manager.cpp
@@ -108,6 +108,15 @@ ScriptValueType ccGetObjectAddressAndManagerFromHandle(int32_t handle, void *&ob
     return obj_type;
 }
 
+ScriptValueType ccTryGetObjectManagerFromAddress(void *address, IScriptObject *&manager)
+{
+    if (address == nullptr) {
+        manager = nullptr;
+        return kScValUndefined;
+    }
+    return pool.FindManagerForAddress(address, manager);
+}
+
 int ccAddObjectReference(int32_t handle) {
     if (handle == 0)
         return 0;
--- a/Engine/ac/dynobj/dynobj_manager.h
+++ b/Engine/ac/dynobj/dynobj_manager.h
@@ -30,25 +30,28 @@ using namespace AGS; // FIXME later
 
 // register a memory handle for the object and allow script
 // pointers to point to it
-extern int32_t ccRegisterManagedObject(void *object, IScriptObject *, ScriptValueType obj_type = kScValScriptObject);
+int32_t ccRegisterManagedObject(void *object, IScriptObject *, ScriptValueType obj_type = kScValScriptObject);
 // register a de-serialized object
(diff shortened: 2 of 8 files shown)


ericoporto:
I don't understand most of this but if the issue is testing in other platforms I can test, but I need to be told how to test this - I can build and run things on them, but I don't know if there is any specific here beyond "plugin that already works keeps working".

ivan-mogilko:
A small update.

I have successfully tested changing return value of "call function" to intptr_t on Windows and Linux, so likely that part might work.
See commit that does only that: 8795c02d7b3a7bbe56eaf3b624327f1b53b844f1

On another hand, I realized a problem which I've missed previously: it's the case when the plugin's function is "void". Because we do not know the return type, we cast each function to a type that returns "int" (or "intptr_t"). If the pointed function is actually "void", then return value will be undefined, but most likely contain random garbage, like some nearby value from the stack.

This "works" so far, because engine does not do anything meaningful with this value, only stores it in a RuntimeScriptValue struct, which is later simply dropped. But if I try to test this value for being a registered script object, as I mentioned in my plan, then this may lead to all kind of unexpected results.

Which in turn means that this proposed solution is going to be even less reliable than the existing one.

ivan-mogilko:
Resolved this differently, see commits:
7ce589e1426afd204d6f7852090a5ab7542856dd
975ffc2b268cedbef7ab11864f185333bf574bdd

PR #2406, opened 2024-05-04
fix[Common/util/stdio_compat.c]: This commit fixes the use of *NIX pr…

bhbuehler:
…imitive I/O functions read and write.

The types of read_num and wrote_num in ags_file_copy were changed from the unsigned type size_t to the signed type ssize_t because read and write return -1 on an error and therefore the variables which hold the return values must be signed or errors will not be caught.

### DIFF
--- a/Common/util/stdio_compat.c
+++ b/Common/util/stdio_compat.c
@@ -192,7 +192,7 @@ int ags_file_copy(const char *src, const char *dst, int overwrite)
     int fd_src, fd_dst;
     int dst_flags;
     char buf[4096]; // CHECKME: larger buffer? malloc a bigger one on heap?
-    size_t read_num;
+    ssize_t read_num; // read returns -1 if there was an error, so we need a signed type.
 
     fd_src = open(src, O_RDONLY);
     if (fd_src < 0)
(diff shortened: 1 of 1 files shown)


bhbuehler:
I do not expect the fixed bug to introduce any problems, though I imagine that is difficult to predict. The code actually checks to see whether `wrote_num >= 0`, to the variables were clearly written to be signed. 

I was able to compile both Debug and Release using config flags `-DAGS_TESTS=On`, `-DAGS_BUILD_ENGINE=On`, `-DAGS_BUILD_TOOLS=On`, `-DAGS_BUILD_Compiler=On`, and `-DAGS_BUILTIN_PLUGINS=On`. I ran `common_test`, `compiler_test`, and `engine_test` for each build. I wouldn't have expected the tests to fail because there aren't actually any for the *NIX I/O functions. I would have written some, but don't know how to get them into a harness easily.

ivan-mogilko:
Looks like my oversight, read and write return ssize_t indeed.
(https://man7.org/linux/man-pages/man2/read.2.html, https://man7.org/linux/man-pages/man2/write.2.html)

PR #2407, opened 2024-05-05
AGS 4: remake ExtendedCompiler setting into ScriptCompiler selection

ivan-mogilko:
As suggested few time before, this remakes the "Use extended compiler" property in General Settings to "Script compiler" with a list of compilers to choose from.

At first I wanted to get compiler names (IDs) out of compiler libraries, but then met 2 issues:
1) I was not sure how to distinguish the "new" and "old" one, to choose one when upgrading a project.
2) I was not sure if their "displayed name" should be also defined by them, or left for the Editor.

At the moment I'm hardcoding their names in the AGSEditor class, but that is frankly not good.
I also would like to make this future compatible with potentially supporting standalone compiler(s).


ivan-mogilko:
Closed in favor of #2552.

PR #2408, opened 2024-05-05
AGS 4: replace "Remap 8-bit backgrounds" preference with Room property

ivan-mogilko:
After making #2393 I realized that the option of remapping 8-bit backgrounds no longer works well if it's located in the Editor Preferences. That is because room backgrounds are stored unremapped in folders, and remapped each time they are loaded into the editor, or a room is compiled for packaging. This means that rooms need persistent "import settings" for their backgrounds, similar to sprite import settings.

In this PR I removed "remap 8-bit backgrounds" from Editor Preferences, and added two settings:
1. DefaultRemap8bitRoomBackgrounds to the General Settings.
2. Remap8bitBackgrounds to the Room properties.

DefaultRemap8bitRoomBackgrounds is assigned to the room whenever a new room is created.
Importing a room from CRM forces this option to false, because the existing backgrounds presumably are already remapped.
When loading a background from the room's folder, Editor uses room's property instead of one from the preferences (which is now removed).

REMAINING ISSUES:
1. I noticed that Room data xml never had its version updated, and does not have a version check at all. It has to have something similar to game.agf version check, but it was never implemented apparently. This may be done separately from this PR though.
2. There's an unfortunate problem: ConvertRoomFromCrmToOpenFormat is also used when creating a room from template, even after generating an empty room programmatically... Possibly I'll have to set Remap8bitBackgrounds property per-case.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -112,9 +112,10 @@ namespace AGS.Editor
          * 4.00.00.00     - Raised for org purposes without project changes
          * 4.00.00.03     - Distinct Character and Object Enabled and Visible properties;
          *                - FaceDirectionRatio
-         *
+         * 4.00.00.05     - Settings.DefaultRemap8bitRoomBackgrounds
+         * 
         */
-        public const int    LATEST_XML_VERSION_INDEX = 4000003;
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
Hmm, no, I have second thoughts about this.
Maybe we should discuss the workflow with developer(s) who are making 8-bit games first, before changing anything.

Issue #2328, opened 2024-02-04
provide standalone builds of the new compiler in ags4


ericoporto:
The new compiler in ags4 is missing a command line front end. It's necessary to decide one of two approaches.

- Use the existing frontend for the old compiler but use macro/template so it's used in building the new one
- Write it's own frontend for the new compiler, use the same code as baseline but make it it's own thing

It needs it's own CMake script to build it too! 

After this, it also needs to be added to the CI - so it can build and be tested in other platforms.

ivan-mogilko:
The two compiler libraries are using same interface, and similar set of options, so I think it's better to use existing frontend code to build both using either a compile-time or runtime switch.

ericoporto:
I am looking into this

- the preprocessor is using bits of the old compiler, so I need to refactor to it's own thing
- the new compiler has a very different interface from the old one
  - all flags come into a int64, old compiler had an interface to set each
  - scripts must be concatenated before, old compiler had an interface to feed things
  - the new compiler uses a message handler, the old one you just got the error

I am thinking there's quite a lot of stuff different that I need to pickout to some file and then use a different file depending on compiler... 

But it would be super easy to make a new frontend from duplicating the code and later figure what the hell they have in common - which could then go into a compiler_frontend_lib of sorts.

PR #2410, opened 2024-05-07
Compiler: preprocessor can build without compiler

ericoporto:
This PR is targeting master branch but it is actually useful in ags4 and part of #2328 as it enables building the preprocessor with both the legacy compiler but also the new ags4 compiler.

- is_digit and is_alphanum are replaced by standard functions
- Fixes preprocessor not properly picking up escaped characters
- the preprocessor also needs the `NEW_SCRIPT_TOKEN_PREFIX` from that header.

With these changes I can build the preprocessor independently of the compiler.

### DIFF
--- a/Compiler/preproc/preprocessor.cpp
+++ b/Compiler/preproc/preprocessor.cpp
@@ -12,6 +12,7 @@
 //
 //=============================================================================
 #include <algorithm>
+#include "script/cs_parser_common.h"
 #include "preproc/preprocessor.h"
 #include "script/cc_common.h"
 #include "util/textstreamwriter.h"
@@ -34,6 +35,11 @@ namespace Preprocessor {
 #else
     static const char * li_end = "\n";
 #endif
+
+    bool IsScriptWordChar(int c)
+    {
+        return std::isalnum(c) || c == '_';
+    }
 
     class StringBuilder : TextWriter {
     private:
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
It's better to use standard library functions instead, unless there's a clear purpose for these?

https://en.cppreference.com/w/cpp/header/cctype

ericoporto:
> It's better to use standard library functions instead

Thanks, I will adjust to go that road instead. I will try to update this later.

ericoporto:
I found a bug in the cpp preprocessor that made it pass the tests accidentally, lol, working on a fix

ivan-mogilko:
I think you might be missing a standard header somewhere.

ericoporto:
Apparently MSVC in VS 2015 and 2017 requires cctype header here, but for some reason MSVC in VS 2019 and 2022 doesn't. Fixed.

ivan-mogilko:
> Apparently MSVC in VS 2015 and 2017 requires cctype header here, but for some reason MSVC in VS 2019 and 2022 doesn't.

Some compilers may have extra headers included into other headers, which results in them being added implicitly. It's safer to include required standard header always.

ivan-mogilko:
Maybe I don't understand something, you don't remove unnecessary functions like `is_digit` and `is_alnum` here?

PR #2417, opened 2024-05-08
AGS 4: catching missed bytecode overflows in the new Compiler

ivan-mogilko:
Small buffer overflows are never detected, because bytecode buffer is always allocated with extra capacity.
But #2411 made this problem come to surface.

This is not yet for merging, the purpose is to catch known overflow errors, and be able to investigate and fix them.

EDIT: need to find out which exception type would be best there, new Compiler introduces couple of exception types.


ivan-mogilko:
Made obsolete by #2510.

PR #2419, opened 2024-05-08
Support importing 4-bit bitmaps as 8-bit (for masks and sprites)

### LINKED ISSUE #2416: Cannot import 16-color BMP file
**Describe the bug**
When importing a room Walkbehind mask, selecting a 16-color BMP image doesn't work. The editor expects a 256-color BMP image using only the first 16 colors, which is a somewhat odd choice of file formats considering common bitmap editors (e.g. MS Paint) by default support 16-color bitmaps.
Opening a 16-color bitmap in MS Paint, saving it as 256-color BMP, and importing that doesn't help (it imports succesfully but doesn't get the full mask, presumably because MS Paint does not save the 16 colors as the _first_ 16 in a 256-color BMP).

**AGS Version**
3.6.0

**Game**
n/a

**To Reproduce**
1. Import this bitmap as a walkbehind mask: [parkmine-wb.zip](https://github.com/adventuregamestudio/ags/files/15241137/parkmine-wb.zip)

**Expected behavior**
Importing the 16-color bitmap works.

 - OS: Windows
 - Version 11

ivan-mogilko:
Fixes #2416

Implemented a conversion from imported 4-bit bitmaps into 8-bit ones supported by the Editor & Engine.

There were 2 potential approaches here, one is to convert on C# side, another in the native C++ side.
I chose to do this in native, and have a helper function in Common lib, to make this potentially usable in the engine. For example, this may be used when loading an image from file.

### DIFF
--- a/Common/gfx/bitmap.cpp
+++ b/Common/gfx/bitmap.cpp
@@ -99,6 +99,22 @@ Bitmap *CreateBitmapFromPixels(int width, int height, int dst_color_depth,
         return bitmap.release();
     }
 
+    // Copy 4-bit indexed image into 8-bit image
+    if (dst_color_depth == 8 && src_col_depth == 4)
+    {
+        uint8_t *data_ptr = bitmap->GetDataForWriting();
+        for (int y = 0; y < height; ++y)
+        {
+            for (int x = 0; x < width / 2; ++x)
+            {
+                uint8_t sp = pixels[y * src_pitch + x];
+                data_ptr[y * width + x * 2]     = ((sp >> 4) & 0xF);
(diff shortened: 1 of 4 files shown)


ericoporto:
I don't know if useful or not, but I think the engine does loads 4-bit bitmap already. I remember to have written it here

https://github.com/adventuregamestudio/ags/blob/01b73a09c0f806260d9ef5a6ef6cf076afa6c907/Common/gfx/bitmap.cpp#L344

ivan-mogilko:
Ah, this reading function unpacks 4-bit to 8-bit on read.
Well then, this means that the new function may not be necessary for loading bitmaps at runtime. It's only needed if an unpacked data is read.

PR #2422, opened 2024-05-08
Improve Room Editor's navigation bar's scrolling

### LINKED ISSUE #2415: Room editor's navigation bar: pulldown menu is inconvenient to scroll
**Describe the bug**
This pulldown menu has inconveniently tiny "up" and "down" buttons, and does not support the mouse wheel.
The same applies to the pulldown for hotspots and for walkable areas and for regions.
pulldown

**AGS Version**
3.6.0

**Game**
n/a

**To Reproduce**
1. Edit a room
3. Select walk behinds
4. Click on '....'

**Expected behavior**
Show all 16 entries of this pulldown menu, if the window size permits.
If not, then allow the mouse wheel to scroll through this screen.
When closing and reopening the pulldown, start from the top (so that selecting the eraser is convenient).

 - OS: Windows
 - Version 11

ivan-mogilko:
Fix #2415

Wrote a ToolStripDropDownMenuEx class which extends ToolStripDropDownMenu and amends or overrides its behavior.
Unfortunately had to use various hacks, including getting internal base methods using reflection.

WHAT IS DONE:
* Scroll dropdown menu with mouse wheel.
* Allow to change arrow buttons height to a bigger value.
* Set dropdown height is at least as big to accommodate 16 items at once, that is current max number of walkable areas, etc.

~* Don't close the open dropdown menu when hovering over other items in the Address Bar.~
EDIT: this fix is temporarily removed, because it turned to be not reliable and causing new problems.

### DIFF
--- a/Editor/AGS.Controls/AGS.Controls.csproj
+++ b/Editor/AGS.Controls/AGS.Controls.csproj
@@ -144,6 +144,9 @@
     <Compile Include="AssemblyInfo.cs">
       <SubType>Code</SubType>
     </Compile>
+    <Compile Include="Controls\ToolStripDropDownMenuEx.cs">
+      <SubType>Component</SubType>
+    </Compile>
     <Compile Include="ScintillaHelper.cs" />
     <Compile Include="ScintillaNET\Annotation.cs" />
     <Compile Include="ScintillaNET\AutoCSelectionEventArgs.cs" />
--- a/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
+++ b/Editor/AGS.Controls/AdressBarEx/AddressBarExt.cs
@@ -1,15 +1,8 @@
-#region Using Statements
-
-#region .NET Namespace
-
 using System;
 using System.ComponentModel;
 using System.Drawing;
 using System.Windows.Forms;
-
-#endregion
-
-#endregion
+using AGS.Controls;
 
 namespace AddressBarExt.Controls
 {
@@ -102,7 +95,7 @@ namespace AddressBarExt.Controls
         /// <summary>
         /// Drop down menu that contains the overflow menu
         /// </summary>
-        ToolStripDropDownMenu tsddOverflow = null;
+        ToolStripDropDownMenuEx tsddOverflow = null;
 
         #endregion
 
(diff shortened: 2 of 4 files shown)


ericoporto:
Could the arrow button height just be slightly bigger, like 24? 20 is in the almost sweet spot to be able to move the mouse without much effort to the place but I think this little pixels would make it just right.

ivan-mogilko:
Rewrote, adding a new ToolStripDropDownMenuEx class to AGS.Controls, that handles most things.

ericoporto:
I have no idea why it works but I tried this diff below and the artifacts when scrolling went away

```diff
diff --git a/Editor/AGS.Controls/Controls/ToolStripDropDownMenuEx.cs b/Editor/AGS.Controls/Controls/ToolStripDropDownMenuEx.cs
index 4afe1ef25..695d975f8 100644
--- a/Editor/AGS.Controls/Controls/ToolStripDropDownMenuEx.cs
+++ b/Editor/AGS.Controls/Controls/ToolStripDropDownMenuEx.cs
@@ -101,6 +101,16 @@ namespace AGS.Controls
             MinDisplayedItems = 0;
         }

+        protected override CreateParams CreateParams
+        {
+            get
+            {
+                CreateParams cp = base.CreateParams;
+                cp.ExStyle |= 0x02000000;  // Turn on WS_EX_COMPOSITED
+                return cp;
+            }
+        }
+
         /// <summary>
         /// Raises the System.Windows.Forms.ToolStripDropDown.Closing event.
         /// Here we intercept certain scenarios where we do not want our dropdown to close.
```

From https://stackoverflow.com/a/3718648, I had tried the code in the question which is what we use for our buffered panel in the room editor, but for some reason it doesn't work in this control specifically.

ivan-mogilko:
So, unfortunately, my fix to "Don't close the open dropdown menu when hovering over other items" is not reliable, and has 2 unexpected consequences:
1. You have to click twice on another "..." before it displays its dropdown menu
2. Sometimes the previous dropdown is not closed even when a new one is opened.

ericoporto:
There is a comment [here in the docs](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.systeminformation.mousewheelscrolllines).

> Be sure to handle the special case that occurs if the "One screen at a time" mouse option is selected. In this case, the [MouseWheelScrollLines](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.systeminformation.mousewheelscrolllines) property has a value of -1.

ivan-mogilko:
@Crystal-Shard, I tried to make scrolling speed relative to the system settings, using a formula found elsewhere, but I am not certain if did this right, because I don't know what is considered to be a "line" in there. Is it a line of pixels, or a line of text? And if it's a line of text, then should this equal to a single item?

That's how I calculate it now:
```
int linesPerWheelDelta = SystemInformation.MouseWheelScrollLines;
int heightOfLine = ItemHeight;
int scrollAmount = (linesPerWheelDelta * heightOfLine * e.Delta / SystemInformation.MouseWheelScrollDelta);
```

On my PC this results in approximately 3 items long scroll at a single wheel tick.

ivan-mogilko:
I had to calculate number of displayed items, in order to properly support "One screen at a time" .

ericoporto:
This works fine here too. I don't think many people set this (one sco) specific setting on windows other than for accessibility reasons. I put a reference to two examples in Winforms source code and it pretty much matches your code - it's written in a much more convoluted way though, but I think it looks fine as is now and it's made in a proper way that minor tweaks are easier to do on top of your code if needed in the future before a new control arises.

PR #2423, opened 2024-05-10
Compiler: removed duplicates of standard cctype functions

ivan-mogilko:
Complementary to #2410.

Remove unnecessary duplicates of standard cctype functions found in compiler code.
Share new IsScriptWordChar() function between parser and preprocessor.


ericoporto:
This kinda undos the goal I had to build the preprocessor without the legacy compiler...

ivan-mogilko:
> This kinda undos the goal I had to build the preprocessor without the legacy compiler...

Could you elaborate?

ivan-mogilko:
> This kinda undos the goal I had to build the preprocessor without the legacy compiler... The real issue I have is the preprocessor cannot depend on the legacy compiler code or I can't use it with the new compiler.

But the preprocessor still includes cs_parser_common.h, in order to get some constant.
So your goal is to NOT include this header?

ericoporto:
Ah, I see, you delete the cpp file so it's just header only, I see. Right, this works. It's alright then. After you merge this please push it to ags4 too, I think my next pr I will do in ags4 for the standalone of the new compiler.

ivan-mogilko:
I noticed that in ags4 there are headers that are shared among both compilers (located in Common/script), which are also used by the Engine.

However there are unfortunately number of duplicate declarations between two compilers, this refers to constants used only internally during compilation (this "__NEWSCRIPTSTART_" string is one of them), and perhaps some utility parsing functions, similar to checking if char can be a part of symbol.

It could be that more things would have to be picked out to shared headers in ags4 (otherwise they have to be duplicated).

ericoporto:
Perhaps there could be a `Compiler/common` or `Compiler/base` (not sure the right name) and then I think for now it would hold a header for things in common between both compilers - and maybe the preprocessor too.

PR #2431, opened 2024-05-20
Plugin API: implemented IAGSEngine.Log()

### LINKED ISSUE #2429: Add a Log function to engine plugin API
**Describe the problem**
Often it's useful to log from a Plugin, and usually this is done by simply throwing printfs or using iostream, but it would be nice to be able to use AGS own log functionality.

Plus there was a typo (https://github.com/adventuregamestudio/ags/commit/9acbba5f17e10921da19f8724bae75d080afd4b0) that made a bit hard to use the Engine script api for `System.Log`.

**Suggested change**
Add some function for logging from the plugin API and increase the plugin version once adding it so it can be detected by a plugin that such function is available.

ivan-mogilko:
Resolves #2429

Added `IAGSEngine.Log(int level, const char *fmt, ...);`.
Log levels are defined as:
```
#define AGSLOG_LEVEL_NONE   0
#define AGSLOG_LEVEL_ALERT  1
#define AGSLOG_LEVEL_FATAL  2
#define AGSLOG_LEVEL_ERROR  3
#define AGSLOG_LEVEL_WARN   4
#define AGSLOG_LEVEL_INFO   5
#define AGSLOG_LEVEL_DEBUG  6
```

The original plan was for the engine to dynamically allocate an individual logging group for each plugin, which would allow to distinguish plugin messages by their group names.
But this is currently complicated by the Editor which does not support arbitrary group ids. So I'm leaving this in plans for the future.

Instead all plugins use a single "Plugin" message group, but engine prepends plugin's name to the message.


ivan-mogilko:
No, this won't work, because Editor does not support receiving arbitrary log groups.

ivan-mogilko:
@ericoporto could you please tell if this works for you?

ericoporto:
I will test this tonight and reply

ericoporto:
I tested and verified this works, it uses the following output when logging `[Plugin][LOG_LEVEL]: plugindllname : log message`.

PR #2434, opened 2024-05-23
Backport importing indexed PNGs as room backgrounds and masks (AGS 4 -> AGS 3)

ivan-mogilko:
This ports minimal necessary code from AGS 4 to AGS 3, letting users to import indexed PNGs (of any palette size) as 8-bit room backgrounds and masks.

I am not certain why this was never added to ags3 branch, perhaps simply forgotten. I think this was addressed for ags4 in issues #1771 and #1772.

I copied only minimal necessary code from BitmapExtensions class, and it seems that, combined with the recent support for small palette image import, this allows to import any indexed PNGs as room bgs and masks.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -409,6 +409,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Utils\AndroidUtilities.cs" />
     <Compile Include="Utils\ArgumentBuilder.cs" />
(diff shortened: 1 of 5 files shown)


ericoporto:
Well, adding significant "weird" code at bitmap loading I felt in ags4 was easier to experiment. Now that this code has been used for a while I think it feels safe to add to ags3.

ivan-mogilko:
#1769 contains fixes for handling of rooms in managed code. But one small fix is applicable to ags3, which is saving masks in requested image format instead of always bmp.

PR #2435, opened 2024-05-25
Introduce BitmapData class: a plain buffer of pixels, may be used without full Bitmap dependency

ivan-mogilko:
This introduces BitmapData class, which is a simple non-owning wrapper over a plain buffer of pixels, combined with its description (width, height, pixel format etc). The purpose of this class is to be used instead of a Bitmap class where we do not need full set of drawing operations, but only read, write and storage of pixel data.

The PixelBuffer is a derived class, which contains owned buffer.

* BitmapData may be used to wrap a pixel buffer from Bitmap, but also from any other source, thus it's possible to use data coming from a third-party lib in some of our algorithms, without constructing full Bitmap object.
* BitmapData and PixelBuffer, and any structs and algorithms using these, may be used in standalone tools without bringing in dependency on full Allegro lib, as these tools don't need drawing, only pixel reading and writing.
* Image loading and saving won't need a full Bitmap class to be constructed, so unusual pixel formats may also get loaded into memory before getting converted to a pixel format supported by Bitmap class.

Changed Bitmap class to contain a pixel buffer compatible with PixelBuffer directly, instead of letting Allegro's BITMAP to allocate one. Added couple of functions to Allegro sources, which let create BITMAP struct without allocating actual data, and insert our custom buffer pointer instead. This means that Allegro's BITMAP won't own the data, but only reference it. This lets to construct Bitmap by *moving* (swapping) PixelBuffer contents, instead of copying.

Moved bitmap loading and saving into a separate files called "image_file.h/cpp". The loading functions now create and return PixelBuffer, saving function work with BitmapData. Bitmap object is only created after loading was successful, from a valid PixelBuffer object.
The functions themselves mostly did not change, only have Bitmap getters replaced with respective getters of BitmapData object.


ericoporto:
This appears to be working correctly upon testing saving a "bmp" bitmap from script and loading it.

But I am getting an assert error when trying to load a PCX file in `bitmapdata.h` in line 124:

```
    inline uint8_t *GetLine(size_t i)
    {
        assert(_stride * i < _dataSize);
        return &_buf[_stride * i];
    }
```

The problem is `_dataSize` is zero. The issue is the bpp of this pcx is 24-bits (it just has no alpha channel!), so it fails `ColorDepthToPixelFormat` in `bitmapdata.h`.

ericoporto:
I can't test the 15-bit as I don't know how to generate image for it but this seems like it's working alright with 24-bit and the other things.

PR #2438, opened 2024-05-30
Engine: add macOS free space check

ericoporto:
@edmundito can you check if this is correct for macOS ?

In my tests this worked alright. I made a separate repo to test the command here: https://github.com/ericoporto/size 

But please check this specifically in ags.

### DIFF
--- a/Engine/platform/osx/acplmac.cpp
+++ b/Engine/platform/osx/acplmac.cpp
@@ -28,6 +28,7 @@ using namespace AGS::Common;
 
 void AGSMacInitPaths(char appdata[PATH_MAX]);
 void AGSMacGetBundleDir(char gamepath[PATH_MAX]);
+int AGSMacGetFreeSpaceInMB(const char path[PATH_MAX]);
 //bool PlayMovie(char const *name, int skipType);
 
 static char libraryApplicationSupport[PATH_MAX];
@@ -70,8 +71,7 @@ int AGSMac::CDPlayerCommand(int cmdd, int datt) {
 }
 
 uint64_t AGSMac::GetDiskFreeSpaceMB(const String &path) {
-  // placeholder
-  return 100;
+  return AGSMacGetFreeSpaceInMB(path.GetCStr());
 }
 
 eScriptSystemOSID AGSMac::GetSystemOSID() {
--- a/Engine/platform/osx/alplmac.mm
+++ b/Engine/platform/osx/alplmac.mm
@@ -26,3 +26,15 @@ void AGSMacGetBundleDir(char gamepath[PATH_MAX])
   strcpy(gamepath, [parentDir UTF8String]);
   }
 }
+
+int AGSMacGetFreeSpaceInMB(const char path[PATH_MAX])
+{
+  @autoreleasepool {
+  NSString *dirpath = @(path);
+  NSError *error;
+  NSDictionary* fileAttributes = [[NSFileManager defaultManager] attributesOfFileSystemForPath:dirpath
+                                                                                       error:&error];
+  unsigned long long freeSpace = [[fileAttributes objectForKey:NSFileSystemFreeSize] unsignedLongLongValue];
+  return (int)(freeSpace / (1024*1024));
+  }
+}
--- a/Engine/platform/osx/alplmac.mm
+++ b/Engine/platform/osx/alplmac.mm
@@ -32,9 +32,22 @@ int AGSMacGetFreeSpaceInMB(const char path[PATH_MAX])
   @autoreleasepool {
   NSString *dirpath = @(path);
   NSError *error;
-  NSDictionary* fileAttributes = [[NSFileManager defaultManager] attributesOfFileSystemForPath:dirpath
-                                                                                       error:&error];
-  unsigned long long freeSpace = [[fileAttributes objectForKey:NSFileSystemFreeSize] unsignedLongLongValue];
+  unsigned long long freeSpace = 0;
+
+  if (@available(macOS 10.13, iOS 11.0, *)) {
+      NSURL *fileURL = [[NSURL alloc] initFileURLWithPath:dirpath];
+      NSDictionary *results = [fileURL resourceValuesForKeys:@[NSURLVolumeAvailableCapacityForImportantUsageKey] error:&error];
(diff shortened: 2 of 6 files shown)


Engine/platform/osx/acplmac.cpp
```diff
@@ -71,7 +72,7 @@ int AGSMac::CDPlayerCommand(int cmdd, int datt) {
 
 uint64_t AGSMac::GetDiskFreeSpaceMB(const String &path) {
   // placeholder
```
edmundito, line 74:
```suggestion
```

Engine/platform/osx/alplmac.mm
```diff
@@ -26,3 +26,15 @@ void AGSMacGetBundleDir(char gamepath[PATH_MAX])
   strcpy(gamepath, [parentDir UTF8String]);
   }
 }
+
+int AGSMacGetFreeSpaceInMB(const char path[PATH_MAX])
+{
+  @autoreleasepool {
+  NSString* dirpath = @(path);
```
edmundito, line 33: Coding convention nit:

```suggestion
  NSString *dirpath = @(path);
```

Engine/platform/osx/alplmac.mm
```diff
@@ -26,3 +26,15 @@ void AGSMacGetBundleDir(char gamepath[PATH_MAX])
...
+  NSError *error;
+  NSDictionary* fileAttributes = [[NSFileManager defaultManager] attributesOfFileSystemForPath:dirpath
+                                                                                       error:&error];
+  unsigned long long freeSpace = [[fileAttributes objectForKey:NSFileSystemFreeSize] longLongValue];
```
edmundito, line 37: Warning in the docs... I wonder if the project would need a PrivacyInfo file:
https://developer.apple.com/documentation/foundation/nsfilesystemfreesize

edmundito:
```suggestion
  unsigned long long freeSpace = [[fileAttributes objectForKey:NSFileSystemFreeSize] unsignedLongLongValue];
```

edmundito:
@ericoporto, looking at the code, it looks fine. I put some small comments here and there. Would you also like for me to test it?

It seems that this should also work for `iOS`, too.

ericoporto:
Thank you so much for the comments, I will do the changes. Out of curiosity to learn something new to me, how does one do these code suggestions in a GitHub PR? Is there a guide or docs for this? I found those super cool!

ericoporto:
Thanks! Lovely to know! I squashed your suggestions in my commit here.

> Would you also like for me to test it?

If you can test, unfortunately AGS doesn't print this information, so the only way I found to test was to edit the code and add an intermediate variable and throw a breakpoint there to read this variable.

> It seems that this should also work for iOS, too.

I don't want to do a change for iOS here because I am not sure how to do it. I guess it would mean adding the alplmac.mm file to the iOS project? Or would it make more sense to have also an `alplios.mm` or have an `alplapplecommon.mm`?

Engine/platform/osx/alplmac.mm
ericoporto: I created a PrivacyInfo file and added in the Xcode project, I only added the NSPrivacyTracking with it set to false. From what I am getting we would only need to list everything we use if we set that to true.

edmundito:
> I don't want to do a change for iOS here because I am not sure how to do it. I guess it would mean adding the alplmac.mm file to the iOS project? Or would it make more sense to have also an `alplios.mm` or have an `alplapplecommon.mm`?

@ericoporto I'm not too sure how much overlap there is between the two, but I suspect there may be more than I think. For now it could be duplicated into alpios.mm (maybe later in a separate PR?). Overall, I think the code can be cleaned up significantly for these platforms and it's due for a refactor where they can share common code. (maybe another future PR)

I'll give it a test this weekend on my Macs.

edmundito:
Tested on maOS via breakpoint and looks like it works! But looking online it seems that it's not completely accurate and reports a lot less than it's available. There's a new API that works with OSX 10.13 (or iOS 11), which is more accurate, but I guess it'd have to [check the OS version](https://developer.apple.com/documentation/xcode/running-code-on-a-specific-version) first:

https://stackoverflow.com/a/46317698/20494

Here is an example:
https://developer.apple.com/documentation/foundation/nsurlresourcekey/checking_volume_storage_capacity?language=objc

ivan-mogilko:
@ericoporto do you have plans on amending this with @edmundito 's suggestion?
If you're currently busy, I could merge and do this addition in a separate PR (although i cannot test myself, since I have no Mac)

ericoporto:
added a check using new api

edmundito (approved):
Tested locally for both types of space check and returns the expected data. I did not test the game with low storage.

Engine/platform/osx/acplmac.cpp
```diff
@@ -28,6 +28,7 @@ using namespace AGS::Common;
 
 void AGSMacInitPaths(char appdata[PATH_MAX]);
 void AGSMacGetBundleDir(char gamepath[PATH_MAX]);
+int AGSMacGetFreeSpaceInMB(const char path[PATH_MAX]);
```
ivan-mogilko, line 31: Argument should not be `const char path[PATH_MAX]`, but `const char*`, because there's no guarantee it's a array of PATH_MAX size.

edenwaith:
Especially with the newer APFS file system (introduced in iOS 10.3 and later made official in macOS 10.13 High Sierra), trying to get an "official" amount of the free disk size is hazy.  Even if one deletes some files, the Mac may not immediately report that there is more space (supposedly) available.  From what I saw in the `apple.mm` file, it looks like good checks for both newer and older techniques.

For more of the nitty-gritty details about reporting free space on a Mac: https://eclecticlight.co/2023/04/27/where-does-macos-get-its-volume-free-space-figures-from/

> Tested on maOS via breakpoint and looks like it works! But looking online it seems that it's not completely accurate and reports a lot less than it's available. There's a new API that works with OSX 10.13 (or iOS 11), which is more accurate, but I guess it'd have to [check the OS version](https://developer.apple.com/documentation/xcode/running-code-on-a-specific-version) first:

PR #2439, opened 2024-06-01
Engine: hotfix certain platforms to not save "warnings.log" in current working dir

### LINKED ISSUE #2437: Engine: Log file location hould consider platform app output directory as base path
**Describe the problem**
When debug mode is enabled and the engine outputs log files (such as `warnings.log`), it writes the file in the same directory as the application. This works fine for Windows games but on other platforms like macOS, iOS (and perhaps others - e.g., consoles) may not be allowed, or it may corrupt the validity of the application. For example, in Mac builds, it will create the file inside the app bundle (`/Applications/<AGS Game Name>.app/Contents/MacOS/`) instead of  where the app-generated files are located (`~/Library/Application Support/<AGS Game Name>/`)

One problem is that the implementation of outputting to warnings.log (or maybe the `create_log_output` function does not consider a path from the platform driver:

https://github.com/adventuregamestudio/ags/blob/be87b0c035763e0857a21b35c5fe48fe90da0319/Engine/debug/debug.cpp#L339

https://github.com/adventuregamestudio/ags/blob/be87b0c035763e0857a21b35c5fe48fe90da0319/Engine/debug/debug.cpp#L166

**Suggested change**
A simple solution would be to update the `create_log_output` function to ensure that the base path is always the app output directory from the driver so that each platform can provide the best path for the log files. However, this needs to be discussed first to ensure it does not impact other areas.

ivan-mogilko:
Fix #2437

There are platforms where this behavior is not suitable even for a Debug game build, because writing to current working dir will modify the game bundle, causing reaction from the operating system or app store. Introduce a IsLocalDirRestricted() method to the platform driver class, use it when "warnings.log" is created to know whether we are allowed to write it into cwd, or should be using a safer user dir.

For the time being, do not impose this restriction on Windows and Linux.

NOTE: this is a temporary hotfix. For the proper solution we'd need to revise "warnings.log" feature and possibly use config settings instead.

### DIFF
--- a/Engine/debug/debug.cpp
+++ b/Engine/debug/debug.cpp
@@ -161,9 +161,10 @@ void SDL_Log_Output(void* /*userdata*/, int category, SDL_LogPriority priority, 
 // Log configuration
 // ----------------------------------------------------------------------------
 
-PDebugOutput create_log_output(const String &name, const String &path = "", LogFile::OpenMode open_mode = LogFile::kLogFile_Overwrite)
+// Create a new log output by ID
+PDebugOutput create_log_output(const String &name, const String &dir = "", const String &filename = "",
+    LogFile::OpenMode open_mode = LogFile::kLogFile_Overwrite)
 {
-    // Else create new one, if we know this ID
     if (name.CompareNoCase(OutputSystemID) == 0)
     {
         return DbgMgr.RegisterOutput(OutputSystemID, AGSPlatformDriver::GetDriver(), kDbgMsg_None);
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
@edmundito would you be able to test if this works well and solves the reported problem (#2437)?

edmundito:
I tested and confirmed that `warnings.log` is saved to the applications support directory (next to the saves) or macOS and iOS 👍🏼

PR #2440, opened 2024-06-02
Engine,Editor: add Touch Script API

### LINKED ISSUE #1538: Enhancement: Proper touch input support

ericoporto:
fix #1538

A minimal touch api that can be used to detect touching a screen.

```
void repeatedly_execute_always() 
{
  TouchPoint* points[];
  points = Touch.GetTouchPoints();
  
  String s = "";
  for(int i=0; i<points.Length; i++)
  {
    TouchPoint* p = points[i];
    if(p.IsDown) {
      s = s.Append(String.Format("P(%d) _DOWN_ (%d, %d)\n", p.ID, p.X, p.Y));      
    } else {
      s = s.Append(String.Format("P(%d) __UP__ (%d, %d)\n", p.ID, p.X, p.Y));
    }
  }
  
  lbl_dbg.Text = s;
}
```

### DIFF
--- a/Engine/ac/touch.cpp
+++ b/Engine/ac/touch.cpp
@@ -1,0 +1,106 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-2024 various contributors
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// https://opensource.org/license/artistic-2-0/
+//
+//=============================================================================
+
+#include <array>
+#include "ac/dynobj/scriptuserobject.h"
+#include "ac/dynobj/cc_dynamicarray.h"
+#include "device/mousew32.h"
+#include "touch.h"
+#include "ac/common.h"
+
+enum class touch_state {
+    up,
+    motion,
+    down
+};
+
+struct touch_point {
+    int x;
+    int y;
+    bool down;
+};
+
+struct tp {
+    static const int MAX_POINTERS = 10;
+    int point_count = 0;
+    std::array<touch_point, MAX_POINTERS> touch_points = {};
+} _tp;
+
+void on_touch_pointer(int pointer_id, Point position, touch_state state)
+{
+    _tp.point_count = std::max(pointer_id+1, _tp.point_count);
+    _tp.touch_points[pointer_id].x = position.X;
+    _tp.touch_points[pointer_id].y = position.Y;
+    _tp.touch_points[pointer_id].down = state != touch_state::up;
+}
+
+void on_touch_pointer_down(int pointer_id, Point position)
+{
+    on_touch_pointer(pointer_id, position, touch_state::down);
+}
+
+void on_touch_pointer_motion(int pointer_id, Point position)
+{
+    on_touch_pointer(pointer_id, position, touch_state::motion);
+}
+
+void on_touch_pointer_up(int pointer_id, Point position)
+{
+    on_touch_pointer(pointer_id, position, touch_state::up);
+}
+
+DynObjectRef create_touchpoint(int n)
+{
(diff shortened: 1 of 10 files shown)


Engine/ac/touch.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void on_touch_pointer_up(int pointer_id, Point position)
+{
+    on_touch_pointer(pointer_id, position, touch_state::up);
+}
+
+DynObjectRef create_touchpoint(int n)
+{
+    if (n < 0 || n >= tp::MAX_POINTERS)
+        quit("!Touch: invalid index for touchpoint");
```
ivan-mogilko, line 66: This is not a good idea to use `quit` here, for something that is neither a fatal error nor a user's fault.
An assertion and returning `null` should be enough.

But seeing how this function is used, I am not sure why is it needed at all.
The loop in Touch_GetTouchPoints() below could do this instead, and limit iterations to MAX_POINTERS too.

ivan-mogilko:
Trying to understand how this api works, am I right that this returned array is always as long as the maximal number of touches done during the gameplay (it only grows and never shrinks)?

My big question in regards to API is: why return array, and not have an indexed attribute instead?
I mean something like
```
builtin struct Touch {
  import static TouchPoint* TouchPoints[];
  import static int TouchPointsCount;
};
```

ericoporto:
> My big question in regards to API is: why return array, and not have an indexed attribute?

That wasy first sketch of the API, I am not sure which is better.

Currently, each touch point is a position (x,y), if the finger is down or not, and a finger id that is unique.

Now apparently Safari on iPhone currently can only have at maximum 2 touch positions, while on Android Chrome it seems possible to go up to 10. On an Android app it seems, at least on my phone, it goes up to 10 (ignoring the hard limit set in ags in this PR), and in my iPhone I need to test again - I forgot. But I thought about using the max to limit the count to how many touches are actually being used in the application.

I was debating about which way is better, to handle the touch points, for instance, I think I like more the way you mentioned, but I was also thinking about dropping the "is down" property from the touch point and return an array and any items in the array means the finger is down, as I think the position in the touch release is not relevant.

ivan-mogilko:
Alternate PR merged: #2692

PR #2441, opened 2024-06-02
AGS 4: Implement Pathfinder scripting api

### LINKED ISSUE #1979: Script API: expose pathfinding to the script
### Problem

Currently pathfinding in AGS is used strictly for the built-in behavior. If a user would like to script their own movement they would have to use a workaround, like a invisible character moving around, and attaching real object to it.

### IMPORTANT

Below are original API proposals, but discussion that followed brought up several issues with these. Please refer to the comments below, or the [summarizing comment](https://github.com/adventuregamestudio/ags/issues/1979#issuecomment-1510406790), to see what we came up to. This ticket has to be updated after we have a better API draft.

### Proposal

Implement a script function that calculates path between starting and ending position. For example:
```
int[] Room.FindPath(int sx, int sy, int ex, int ey);
```
```
Point[] Room.FindPath(int sx, int sy, int ex, int ey);
```
There are two alternatives here, one returning an array of ints, another an array of managed Points.
Each may have pros and cons. Points are easier syntactically, but will require to allocate N managed objects, and will take bit more memory because of how managed objects are stored (unfortunately at this moment ags script does not support returning regular structs by values, neither dynamic arrays of regular structs for some reasons).
In case of ints, the x & y coordinates would be stored sequentially (x0, y0, x1, y1, and so forth).

In AGS4 the new compiler already supports .Length property in arrays, but if we want to also support compiler without this feature, then we might add one final element that signifies end of data: e.g. null pointer in case of `Point[]` array, and some obvious invalid value in case of `int[]` (like INT32_MIN).

Implementation would call current pathfinder (see `find_route`). Something to consider: I don't remember exactly, but I think that pathfinder uses "movement speed" parameter when calculating moves. In any case, we do not need step-by-step points for this, only node points.

See also, as mildly relevant: #1965

### Alternate proposal

After writing this ticket, I realized there may be an alternative to `int[]`/`Point[]` array. That is - a `MovePath` struct, which might look like:
```
builtin managed struct MovePath {
    import readonly attribute Point Pos[];
    // OR
    import readonly attribute int X[];
    import readonly attribute int Y[];
    //
    import readonly attribute int Length;
};
```
This is somewhat more complicated, but there are couple of advantages I can see:
* this struct may be easily expanded if necessary;
* we may use same struct for exposing Character's and Object's move paths (like - `Character.MovePath` and `Object.MovePath` properties).
* there's already an internal MoveList object, so it may have certain sense to make a script's "frontend" for the MoveList.

ivan-mogilko:
Resolves #1979

This implements following scripting API:

```
builtin managed struct Pathfinder {
  /// Tries to find a move path from source to destination, returns an array of navigation points.
  import Point*[] FindPath(int srcx, int srcy, int dstx, int dsty);
  /// Traces a straight path from source to destination, returns the furthermost point reached.
  import Point* Trace(int srcx, int srcy, int dstx, int dsty);
};

builtin managed struct MaskPathfinder {
  /// Creates a new MaskPathfinder, initialized with the given 8-bit mask sprite
  import static MaskPathfinder* Create(int mask_sprite); 
  /// Assigns a new mask to this MaskPathfinder
  import void SetMask(int mask_sprite);
};
```

Here Pathfinder class is a parent class that provides most basic methods, unrelated to how the pathfinding works. User cannot create objects of this class directly.
MaskPathfinder is a child class that implements a pathfinder based on 8-bit masks.

And the Room class now has a new static method:
```
struct Room {
  <...>
  /// Gets this Room's Pathfinder object that lets find route around walkable areas.
  import static readonly attribute Pathfinder *PathFinder;
};
```

Note that the Room returns a pointer of parent type Pathfinder, but it creates MaskPathfinder internally.
This is done to restrict user from changing the room pathfinder's mask, because this pathfinder object is meant to represent actual room's walkable areas.

<details>
<summary>
Example script: CLICK HERE
</summary>

**WARNING: ** player Character must be set Solid = false for this script to work! otherwise paths won't be found, because they start inside the Character's blocking area.

```
DynamicSprite *pathspr;
Overlay *pathover;

function on_mouse_click(MouseButton button)
{
	if (button == eMouseLeft)
	{
		handle_room_click(button);
		return;
	}
	
	if (button != eMouseRight)
	{
		return;
	}
	
	if (pathover == null)
	{
		pathspr = DynamicSprite.Create(Room.Width, Room.Height);
		pathover = Overlay.CreateRoomGraphical(0, 0, pathspr.Graphic);
		pathover.Transparency = 50;
	}
  
	Point *goal = Screen.ScreenToRoomPoint(mouse.x, mouse.y);
	Pathfinder *finder = Room.PathFinder;
	Point *path[] = finder.FindPath(player.x, player.y, goal.x, goal.y);
	Point *trace = finder.Trace(player.x, player.y, goal.x, goal.y);
	
	DrawingSurface *ds = pathspr.GetDrawingSurface();
	ds.Clear(30);
	if (path != null)
	{
		for (int i = 0; i < path.Length; i++)
		{
			if (i > 0)
			{
				ds.DrawingColor = 1;
				ds.DrawLine(path[i - 1].x, path[i - 1].y, path[i].x, path[i].y);
			}
			ds.DrawingColor = 4;
			ds.DrawRectangle(path[i].x - 1, path[i].y - 1, path[i].x + 1, path[i].y + 1);
		}
	}
	
	if (trace != null)
	{
		ds.DrawingColor = 12;
		ds.DrawLine(player.x, player.y, trace.x, trace.y);
	}
	
	ds.Release();
}
```

</details>

---

**KNOWN ISSUES:**
* MaskPathfinder is currently unsafe, internally the actual RouteFinder object stores Bitmap directly, and it does not track sprite's existence, so if you assign a dynamic sprite as a mask, and then delete the sprite, MaskPathfinder can crash the engine. I need to rethink how it is organized.
* Possibly need extra property or argument for the Pathfinder returned by the room, which tells to ignore solid objects and characters, because right now they may unexpectedly "get in a way" (both literally and figuratively).


ivan-mogilko:
Indeed, this is a strictly 2D pathfinder. If there's a problem with that, I can rename it to Pathfinder2D, for example.

>Dynamic Sprites use the bitdepth of the game and you can't specify it to be 8-bit.

There's a reserved "format" parameter in DynamicSprite.Create for this.

ericoporto:
The only "color" that matters is 0 and everything else is considered walkable, so in theory the format could be made to not matter? Just thinking.

ivan-mogilko:
> The only "color" that matters is 0 and everything else is considered walkable, so in theory the format could be made to not matter? Just thinking. I haven't looked at the code, don't remember if the code is properly optimized so it only works with 8-bit specifically.

The pathfinder expects a 8-bit image, because it does not use "getpixel" function, it uses raw arrays of bytes for performance:
https://github.com/adventuregamestudio/ags/blob/983c4a974495162f2891264e3aa8ac10c510d708/Engine/ac/route_finder_impl.cpp#L67-L68
https://github.com/adventuregamestudio/ags/blob/983c4a974495162f2891264e3aa8ac10c510d708/Engine/ac/route_finder_jps.inl#L60

ericoporto:
Ah, right, I forgot that alpha is a thing. So better leave it working only with 8-bit for now.

ivan-mogilko:
UPDATE.
* Disabled (commented out) MaskPathfinder in script, as it is unusable without support for importing/creating 8-bit sprites in hi-color games.
* Added Character.WalkPath, Character.MovePath and Object.MovePath functions that accept array of Points instead of calculating a new path. It is possible to pass a result of Pathfinder.FindPath, but also possible to just make your own path the way you please.

ericoporto:
I tested this, and I notice one thing, if I have a path and do `oBlueCup.MovePath(path, 2);`, it moves the blue cup through the set path, which is great. But if I happen to call AGAIN passing the SAME PATH the object won't restart the movement, I think because the object is already in the final position, it will do nothing. I need to instead do `oBlueCup.SetPosition(path[0].x, path[0].y);` before calling MovePath.

The other thing is if I understood a path is always at minimum TWO points, a beginning and an end, not sure if the function that receives a path should duplicate a point if it receives a single point to avoid something failing later in ags - perhaps not, but just checking.

messengerbag:
It occurs to me that it would be useful to have a method to retrieve the path a Character or Object is currently following. (The alternative is to make sure you don't use Walk or Move, but FindPath before Walk/MovePath.)

The MovePath behavior @ericoporto describes sounds like correct behavior to me (I guess the alternative is that if the starting point doesn't match the Character/Object's position, it first jumps to the starting position?), but perhaps it would be worthwhile to add a `Direction` argument to the movement functions, so they could move backwards along the same path without having to reverse the array.

ivan-mogilko:
1. It appears as a good idea to have Direction and RepeatStyle parameters in MovePath then, and Direction enum might have something like "Alternate" added (or "AlternateForwardFirst" and "AlternateBackwardFirst"). If Direction enum is expanded, then other things that use it might also implement this (Animate).
I've been considering expanding these direction options for Animation some time ago.
OTOH, this may be done differently: have RepeatStyle as Once, Repeat (as in "repeat same dir"), and RepeatAlternate, whereas Direction tells the starting direction.

2. In regards to exposing current character and object's moving path, I've been thinking about exposing a "MoveList" struct to script (it may be given a better name in script), as something that has at least a "CurrentStage" index in addition to the path. This was mentioned in the task ticket (#1979). The other MoveList parameters are velocity vectors (x and y floats telling the movement velocity in each stage), but idk if this is needed to be in script.

ivan-mogilko:
But then, the above are matters of convenience. Yet there's one major problem which I haven't thought through beforehand.

Besides the room walkable mask there are also solid objects in the room. Default pathfinding combines these two (it paints both original mask and solid objects onto the temp mask, but that's implementation details).

The Pathfinder interface in this PR does not let to control that in any way, it simply does not have any concept of additional blocking objects. When engine makes a path for a walking character, it knows to exclude the character's "blocking region", but if user will call Pathfinder.FindPath, there's no indication of which object are we searching path for. It may be for a real object in room, but also for anything "imaginary" as well.
That is important, because if we let use this interface for finding paths in the room itself, then we must let select at least two things:
1. whether to include solid objects into pathfinding;
2. if yes, then which object to exclude, since we're building path from its position.

I am not particularily certain whether "Pathfinder" should be the one knowing about these objects, as this is an interface for something that searches a path, using some implementation. It could as well be that there has to be a separate class that describes an "area" with walkable masks as a foundation, and a list of "colliders".

ericoporto:
In my head, a PathFinder2D can take a 2D grid of nodes and calculate a path on them, and you can convert a Sprite (8-bit?) in a 2D grid of nodes, and you can convert a Room to a Sprite, either picking only it's walkable areas or a set of walkable areas + solid things2d (object + characters in room).

So either something like...

```
DynamicSprite* walkable = Room.GetWalkMaskForCharacter(cEgo);
PathFinder2D p2d = PathFinder.Create(walkable.Graphic);
Point* path[] = p2d.FindPath(4, 4, 250, 150);
```

Or just have MaskPathFinder be instead something that is made focused on Room Walkable Area mask instead of sprite or instead have a proper RoomPathFinder that extends PathFinder2D.

messengerbag:
@ericoporto's example doesn't fully make sense to me (I like the original proposal of having a preexisting `Room.Pathfinder` better), but I do think it makes sense to put the API to create the apppropriate walkable area mask in `Room` and pass it to the Pathfinder. That requires giving `Room.Pathfinder` the MaskPathfinder API, but that's better than exposing a special RoomPathfinder type.

ivan-mogilko:
I've been also thinking about extra type RoomPathfinder recently, as something that deals with characters and objects: "knows" about their presence and may find a path from particular perspective. These are vague ideas atm.

>  I do think it makes sense to put the API to create the apppropriate walkable area mask in Room and pass it to the Pathfinder. That requires giving Room.Pathfinder the MaskPathfinder API, but that's better than exposing a special RoomPathfinder type.

The problem I see in this approach is that it seems redundant to combine persistent Room.Pathfinder object and feeding it with a mask you get elsewhere. With this nothing prevents from giving it a mask which is not related to the room at all, and so what's the point of having it a Room.Pathfinder then?

ivan-mogilko:
My apologies, I said something wrong in the previous comment, so must amend:

The reason why you will have to recreate a "mask for character" every time you call a FindPath is that the contents of this mask might change whenever **any other character or object moves**, and whenever room's mask is modified in any way (toggled walkable areas, manually redrawn walkable areas).

I believe that the public interface must not block any potential future optimizations, and therefore should not follow current suboptimal implementation.

ivan-mogilko:
The way I see this, there's a persistent "object" which may be called a "navigation grid", and this object has following features:
1. It may be generated from a different kinds of data (such as a mask bitmap);
2. It may have "toggleable" parts, that is - bits that block path and may be on or off depending on some setting.
3. It may have "dynamic" obstacles, that is - bits that potentially change their presence, position and state over time.
4. Provides a pathfinding method that searches for a path, *optionally* ignoring selected walkable areas or obstacles.

The last option is useful if we need pathfinding that skips dynamic obstacles, for example, but we don't want to keep reconfiguring navigation grid just for that.

This is, I think, how navigation works in game engines on a basic level (more or less).

If we apply this to AGS:
1. Navigation grid is commonly generated from a mask bitmap.
2. Grid consists of parts constructed from different mask colors (aka areas). These may be turned on and off.
3. Characters and objects provide optional non-static obstacles that may change their position, become solid or non-solid.

The exact way this "grid" is implemented internally should not be exposed to the scripting api from "pathfinder", in my opinion, because that might prevent having other, alternate or better optimized implementations. Instead, I believe, we should try to make an abstract interface that deals with basic concepts; and only provide extra methods that construct or return this "grid" in a certain form (such as mask bitmap) as a non-essential feature.

Supposedly, this means that a navigation grid, or "pathfinder" object, if it contains grid in itself as well, should have:
1. A factory method, which generates it from something, such as a mask.
2. Method(s) which toggle(s) certain parts of the basic map on and off.
3. Methods which add, remove, update or toggle dynamic obstacles.
4. Pathfinding method, that allows to optionally skip certain obstacles.

The full set of these above methods is needed for the generic pathfinder, which is created and configured by user script.
The "Room's own pathfinder" should not have most of these setup methods exposed, because it receives these settings from the room state. That is - when user calls `RemoveWalkableArea`, `RestoreWalkableArea` or sets `Character.Solid`, etc.
This makes "Room's pathfinder" a "higher-level" pathfinder, which functionality is programmed within the engine.

The question remaining is how to tell to ignore certain walk paths or certain obstacles when making a path.

It seems that the common way of doing this in the game engines is a flag set (a bit mask), which may be assigned to certain parts of a navigation grid (such as area) and given to a moving character. The character would only use the areas that have matching bits, or acknowledge obstacles that have matching bits set (this may be done other way around too).
When character is commanded to walk on its own, the engine can tell which flags to use, and also to prevent character from being an obstacle with itself.
But when a user wants to search a path for a certain character, they need a way to pass this character's reference in. Now, we have 2 kinds of game objects that may move along a path, and be an obstacle: Characters and Room Objects. Unfortunately, as of now they don't have a parent type that could be used as a shared reference type between them (so that we did not have to create separate methods for these two).

The solution here could be a new struct returned from both a Character's and Object's property, which:
* contains any configuration related to navigation, such as an aforementioned "flag set" telling which areas to include in pathfinding;
* may be passed as a "ignore this obstacle" argument into "find path" function.

We may call this struct a "Collider", for instance, or, if that term would rather be reserved in case of "physics" implementation in the engine, then perhaps something like "Navigation", "NavAgent", "NavBox", "NavObject" etc.

The `FindPath` function would then look something like:
```
Point*[] FindPath(int srcx, int srcy, int dstx, int dsty, int area_flags, NavObject *exclude);
```

ivan-mogilko:
Hmm, of course introducing a new struct such as "NavObject" comes with extra responsibilities, as we need to make sure that it may be expanded and may have more uses in the future. Such struct would have to be connected to its owner (character or object), and likely custom pathfinders need to be able to handle these too.

What may be the alternatives to a new struct, in a short term?

I see two options:
1. Instead of a struct, introduce something like a "navigation index" property for Characters and Objects. This will be an arbitrary integer value assigned by the engine, that may be passed into Pathfinder's FindPath.
2. Have two explicit methods FindPathForChar / FindPathForObject. But then, these methods should rather be in specialized Room's Pathfinder class, because they make no sense for the user-created pathfinder.

ivan-mogilko:
So, here's my experimental proposal:

1. Added `NavEntity` struct. Currently this looks like a dummy struct that exposes nothing to script, but keeps some data internally.
2. Added `readonly NavEntity* Character.Navent` property that returns NavEntity for a character (I did not really know how to call the property...).
3. `Pathfinder.FindPath` and `Pathfinder.Trace` accept an optional `NavEntity* exclude` pointer.

This allows to tell that particular character should be not considered an obstacle when searching for a path.

ivan-mogilko:
Fixed few things, besides others the character or object will now jump to the beginning of a path when using WalkPath/MovePath.
I suppose that's reasonable, because this api does not force the path's start to match character's current position. So it's either that, or the character must first walk towards the first point? idk how expected that might be.

In regards to other things. I played with the idea of excluding a chosen character from obstacles (as described in few of my previous comments), but in the end decided to drop this completely for the time being. Either way this would require to introduce some kind of a "obstacle identifier", either as a new struct, or an integer index which user has to calculate in certain way (may or not be equal to character.ID). And I don't feel like adding more new concepts right away.

Current API still allows to disable solid objects from pathfinding, by temporarily toggling Object.Solid or Character.Solid properties. That's enough for now.

Similarly, I don't want to add new parameters that change moving logic in this PR: such as direction and repeat option. I am totally for this, but I think I better do this separately. I remember that wanted to bring these parameters to consistency in a way how they are stored in memory and game saves, because right now every object have them in their own way, which is very ugly.

So, what's left here, is an idea to expose current paths for character and object. I will try to do this next.

ivan-mogilko:
Added simple methods for getting current Character's and Object's paths:

```
/// Returns the moving path of this character, or null if it's not moving
Point*[] Character.GetPath();
/// Returns the moving path of this object, or null if it's not moving
Point*[] Object.GetPath();
```

As mentioned earlier, I have a thought about adding something that returns a more sophisticated object, which describes not only the path, but also speeds and current position on this path, but perhaps I should do this as a separate task. Another task would be adding "repeat" and "direction" parameters to Move/WalkPath methods.

ericoporto:
I tested and verified that GetPath works, it returns the path correctly both with characters and objects here. FindPath is also working.

I think I don't understand what `Trace` is meant to do, is it to check if the final point can be the destination without using a pathfinder - thus in a way that uses less CPU in some way?

ivan-mogilko:
> I think I don't understand what Trace is meant to do,

Trace draws a ray from src to dst, and finds the point at which this ray intersects any wall. If this point is before dst, then the last position before wall is returned, otherwise dst is returned.

PR #2442, opened 2024-06-12
Engine: fix string overflow in parser

mausimus:
Under certain conditions sentence parsing function can read beyond the end of input string resulting in unpredictable behavior and/or crashes, this small fix prevents this.

The problematic scenario happens when a multi-word is the last alternative in the input and an earlier alternative has already been matched. For example: "one,two,three point five" where "three point five" is a multi-word and "two" has already been matched. The logic that is supposed to skip other alternatives will jump beyond the end of the input string.

The issue is FindMatchingMultiWordWord() sets text to point to the first character after matched multi-word which in this scenario is \0. The search continues and in the next iteration text is incremented to the first character after \0 which reads out-of-bounds memory. Most of the time this is benign (random invalid characters might go unnoticed) but if it accidentally reads ']' for example it will result in game crash due to malformed input (original issue we've observed).

### DIFF
--- a/Engine/ac/parser.cpp
+++ b/Engine/ac/parser.cpp
@@ -243,7 +243,11 @@ int parse_sentence (const char *src_text, int *numwords, short*wordarray, short*
                             thisword[text - textStart] = 0;
                             // forward past any multi-word alternatives
                             if (FindMatchingMultiWordWord(thisword, &text) >= 0)
+                            {
+                                if (text[0] == 0)
+                                    break;
                                 continueSearching = 1;
+                            }
                         }
                     }
 


ivan-mogilko:
@mausimus Hello, could you explain how to setup a test game which lets to reproduce this problem? I'd like to clearly understand what is fixed here, but do not have full knowledge about how text parser works.

I tried creating a dummy game with words: "look", and "chair", "sofa", "seat" as synonyms.
Then used following script:

```
Parser.ParseText("look chair");
bool result = Parser.Said("look chair,sofa,window seat");
Display("said result = %d", result);
```

For some reason this script always results in Said returning false if there's a multi-word in the input string.
But, running engine under debugger, I do not observe it ever getting beyond the `text` array; in fact `FindMatchingMultiWordWord` fails.

mausimus:
@ivan-mogilko This is the minimal repro for my particular problem, on original code I hit this breakpoint pointing at \0 and loop continues in the next iteration beyond the end of the string.

Screenshot 2024-06-13 084926
Screenshot 2024-06-13 084907
Screenshot 2024-06-13 085118

I agree with you there seem to be other issues in this function, "look chair" should work but doesn't because comma gets included in multi-word lookup and never gets backed out of. I've added a fix for this so that we shift the beginning of the potential multi-word each time we encounter a comma. Now this gives me all 1s:

Screenshot 2024-06-13 092126

ivan-mogilko:
Thank you, I see how this problem appears now.

I will merge this, and also copy the fixes to release-3.6.1 branch (it's the latest stable version which we apply patches to).

But there's an issue with the text parser: it's not used a lot, and not many people are clearly aware of how it is supposed to work.
We have a repository with a test game, which is used practically as an automatic unit test, here:
https://github.com/adventuregamestudio/ags-test-games

If you may be interested and have spare time, I would propose for you to add tests for Text Parser feature there.

ivan-mogilko:
Ah, unfortunately, I found that your last fix could be wrong.

Reading the text parser's documentation here:
https://adventuregamestudio.github.io/ags-manual/TextParser.html

It appears that in AGS parser comma-separated sequences may only include single words.

Specifically this example:
> if (Parser.Said("eat,take tablet"))

Here the two variants are "eat tablet" and "take tablet", but not "eat" and "take tablet".

This means that the example that you posted: "look chair,sofa,window seat", was treated by AGS like
- "look chair seat", or
- "look sofa seat", or
- "look window seat".

mausimus:
Thanks so much, I will take a look at the test game and possibly adding some tests. Agreed the documentation isn't entirely clear what is or isn't allowed. I would like to keep the text parser functionality alive, if there are any issues around it flagged please feel free to mention me and I'd be happy to investigate.

ericoporto:
@mausimus If you have time, the parser tests are here: https://github.com/adventuregamestudio/ags-test-games/blob/main/ags3/auto-test/test-parser.asc , you may need to additionally add things in the game project for the test as needed.

PR #2443, opened 2024-06-15
Engine: fix check for word boundaries in the parser

mausimus:
Parser should exclusively use is_valid_word_char to check for word boundaries (affects dash and apostrophe containing words).


ivan-mogilko:
One request, if you make more PRs: please add some reference to the text parser in commit's description. For example:
"Text parser: consistently check for word boundaries"
or similar.

This makes it easier for me to write changelogs and cherry pick fixes to other branches (when that's necessary).

PR #2444, opened 2024-06-15
AGS 4: RepeatStyle and Direction parameters for MovePath and WalkPath

ivan-mogilko:
This is to be applied post-#2441. (this branch contains all the 2441 contents atm)

Added `RepeatStyle` and `Direction` parameters to Character.MovePath, WalkPath and Object.MovePath.


ivan-mogilko:
Hmm, this produced an unexpected effect, where walking a path with eBackwards makes character to walk virtually backwards (chosen directional loops are inverted). I think this is not what is expected, @messengerbag ?

PR #2452, opened 2024-06-20
Editor: Fix properties for properties indicator in BusyDialog

edmundito:
Fixing something that bothered me for the longest time. When the game builds and the BusyDialog appears, the user can hover over the progress indicator, and the background will slightly change color. The reason behind this is that the indicator is a hacked button.

What I did to solve it was tor set the background for click/hover to transparent. Additionally, I removed a tab index and disabled validation since it's not needed.

Before - note that the screenshot does not include the mouse cursor, but it's placed over the progress indicator:
image


ivan-mogilko:
@edmundito Thank you, but please, if it's a fix to a non-serious problem, it's best to make PR to master branch instead.

PR #2453, opened 2024-06-21
Editor,Engine: declare game object imports as pointers

ericoporto:
This change affects GUIs, Inventory Items, Dialogs and Characters

These elements are imported using their direct raw memory references instead of handles but they are used in script as pointers, this change modifies the header generation to actually use pointers and also registers to them their handles.

This change can be visible when using the Watch Variable feature as the handle will show up there. I am leaving this as draft because I am not sure yet of the consequences of  this change.

### DIFF
--- a/Engine/ac/gui.cpp
+++ b/Engine/ac/gui.cpp
@@ -60,6 +60,7 @@ using namespace AGS::Engine;
 extern RoomStruct thisroom;
 extern int cur_mode,cur_cursor;
 extern std::vector<ScriptGUI> scrGui;
+extern std::vector<int> StaticGUIControlsArray;
 extern GameSetupStruct game;
 extern CCGUIObject ccDynamicGUIObject;
 extern IGraphicsDriver *gfxDriver;
@@ -611,14 +612,33 @@ void prepare_gui_runtime(bool startup)
     GUIE::MarkAllGUIForUpdate(true, true);
 }
 
-void export_gui_controls(int ee)
+void export_all_gui_controls()
 {
-    for (int ff = 0; ff < guis[ee].GetControlCount(); ff++)
-    {
-        GUIObject *guio = guis[ee].GetControl(ff);
-        if (!guio->Name.IsEmpty())
-            ccAddExternalScriptObject(guio->Name, guio, &ccDynamicGUIObject);
-        ccRegisterPersistentObject(guio, &ccDynamicGUIObject); // add ref for engine
+    size_t guictrlcount = 0;
+    for (int i = 0; i < game.numgui; ++i) {
+        auto const &gui = guis[i];
+        for (int j = 0; j < gui.GetControlCount(); j++) {
+            GUIObject const *guio = gui.GetControl(j);
+            if (!guio->Name.IsEmpty()) {
+                guictrlcount++;
+            }
+        }
+    }
+
+    StaticGUIControlsArray.resize(guictrlcount);
+
+    int gi = 0;
+    for (int i = 0; i < game.numgui; ++i) {
+        auto const &gui = guis[i];
+        for (int j = 0; j < gui.GetControlCount(); j++) {
+            GUIObject *guio = gui.GetControl(j);
+            int handle = ccRegisterPersistentObject(guio, &ccDynamicGUIObject); // add ref for engine
+            if (!guio->Name.IsEmpty()) {
+                StaticGUIControlsArray[gi] = handle;
+                ccAddExternalScriptObject(guio->Name, &StaticGUIControlsArray[gi], &GlobalStaticManager);
+                gi++;
+            }
+        }
     }
 }
 
@@ -632,6 +652,15 @@ void unexport_gui_controls(int ee)
         if (!ccUnRegisterManagedObject(guio))
             quit("unable to unregister guicontrol object");
     }
+}
+
+void unexport_all_gui_controls()
+{
+    for (int i = 0; i < game.numgui; ++i)
+    {
+        unexport_gui_controls(i);
+    }
+    StaticGUIControlsArray.clear();
 }
 
 void update_gui_disabled_status()
--- a/Engine/ac/gui.cpp
+++ b/Engine/ac/gui.cpp
@@ -658,10 +658,6 @@ void unexport_all_gui_controls()
         unexport_gui_controls(i);
     }
     StaticGUIControlsHandles.clear();
-
-    // FIX-ME: for some reason this method is called in DoBeforeRestore
-    // so we must reset arrays sizes below. Remove if removed from DoBeforeRestore
-    set_array_all_gui_controls_size();
 }
 
 void update_gui_disabled_status()
--- a/Engine/ac/gui.cpp
+++ b/Engine/ac/gui.cpp
@@ -60,7 +60,7 @@ using namespace AGS::Engine;
 extern RoomStruct thisroom;
 extern int cur_mode,cur_cursor;
 extern std::vector<ScriptGUI> scrGui;
-extern std::vector<int> StaticGUIControlsArray;
+std::vector<std::vector<int>> StaticGUIControlsHandles;
 extern GameSetupStruct game;
 extern CCGUIObject ccDynamicGUIObject;
 extern IGraphicsDriver *gfxDriver;
@@ -228,7 +228,7 @@ void GUI_SetSize(ScriptGUI *sgui, int w, int h) {
 
   if ((tehgui->Width == w) && (tehgui->Height == h))
     return;
-  
+
   tehgui->SetSize(w, h);
   tehgui->MarkChanged();
 }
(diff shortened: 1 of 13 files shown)


ericoporto:
I am apparently getting segfaults when running the test here : https://github.com/adventuregamestudio/ags-test-games/blob/main/ags4/auto-test/test-misc.asc

ivan-mogilko:
I think this should be done for room things too (objects, hotspots, walkable areas etc).

ericoporto:
Conceptually is this change good or bad? I can't tell, I think on one hand direct memory access is faster than handle, but otoh every use of pointer in pure script structs are using handles.

ivan-mogilko:
> Conceptually is this change good or bad? I can't tell, I think on one hand direct memory access is faster than handle, but otoh every use of pointer in pure script structs are using handles.

I think it's proper, because these are managed structs, and compiler currently has to provide a hack for "import" case which allows to declare a non-pointer variable of managed type.

Furthermore, if these types will be dynamically created at some point, then we'll have to have a consistent declaration of them.

I don't think a difference in access should make a notable difference in performance, and these are individual objects. It is changing their arrays to arrays of pointers that made most difference in this regard (because that affects iterating all objects in a loop), but that is already done.

Something I forgot to mention, these pointers must be declared as `readonly`, otherwise user will be able to reset them in script.

ericoporto:
Hey, you mentioned Walkable Areas... are they exported too here? They, and walkbehinds and regions, I am not sure I see where they are...

ivan-mogilko:
> Hey, you mentioned Walkable Areas... are they exported too here? They, and walkbehinds and regions, I am not sure I see where they are...

No, they are not, because they don't have script names.

ericoporto:
There's one thing I don't quite get that is missing here which are the GUI Controls

ivan-mogilko:
> There's one thing I don't quite get that is missing here which are the GUI Controls, there isn't an existing array or place where their handles are stored, I guess they would need to be stored somewhere.

The GUI Controls never had an array of handles created in the engine, because they are not exposed as a single array of pointers in script. But such array (vector) may be created. I suppose that it's enough to make a single array for all controls in game, and fill it inside `export_gui_controls` function.
https://github.com/adventuregamestudio/ags/blob/70513e44604b014be3bc4814e2c206d2e68673f2/Engine/ac/gui.cpp#L589
and clear in `unexport_gui_controls`:
https://github.com/adventuregamestudio/ags/blob/70513e44604b014be3bc4814e2c206d2e68673f2/Engine/ac/gui.cpp#L600
I guess it's not convenient to precalculate each control's index, so better just push into array. It has to be reset prior to calling `export_gui_controls` though.

ericoporto:
Uhm... I think resizing a vector changes the memory allocation, and so it changes the memory location of an element in a vector... So I would need to iterate twice in there to be able to create a static vector to be able to pass the memory address in `ccAddExternalScriptObject`. The other alternative would be having something like `ccAddExternalScriptObject` that can actually take handles.

ivan-mogilko:
Ah, right... well then, there's need for another function like export_gui_controls without gui index, which first calculates total size, allocates a vector, then runs actual export. But exporting should take this single array's index reference which it uses and updates.

ivan-mogilko:
Your new change is wrong, this function is called separately for each gui.
Resize must be done outside of this function to work properly.

ericoporto:
I am doing something wrong with GUI Controls somewhere, or at least, this broke them... Would this require to use something other than `ccDynamicGUIObject` to register the objects because the pointers are to Button, Slider and other different types?

ivan-mogilko:
As I mentioned, you are still registering them wrong, because this function is called multiple times for each GUI.
So you are recreating same array multiple times, instead of appending it.

You need to calculate the total size of all controls on all guis once, prior to calling this function `export_gui_controls`, allocate a vector with this total size, and then pass "last used index" reference into `export_gui_controls` as a argument. Maybe make a function `export_all_gui_controls` which wraps this whole thing up.

ericoporto:
> Perhaps a good alternative: first pass creates objects and pushes their handles into global array (without preliminary resizing), and then the second pass adds accumulated handles to exports table.

Ah, that sounds like a good idea! I will try that. It will take me a bit to get back because I need to briefly go out but I will try it soon.

ivan-mogilko:
Finally I got time to test this, and unfortunately this does not work, at all. Whenever an object pointer is passed to the engine api function, it seems that is not a real object, but a random memory (fields have garbage in them).

I must investigate what is happening, possibly something I missed something.

ivan-mogilko:
Erm, sorry, deleted my last comment. It turned out I mixed 2 different versions of engine and editor when testing.

So, this seem to work overall. I am confused by the tests result, going to investigate that now.

ivan-mogilko:
Ah, I think i understand, the automatic tests are not rebuilt, and the older compiled scripts won't work with these changes. So the tests are failing for the same reason why they failed for me when I mixed new engine with the older editor.

ivan-mogilko:
I realized that this PR is missing one important piece: re-registering objects after restoring a save.
You may find these by searching for `ccRegisterUnserializedPersistentObject` calls.

I suppose that for most objects it will be a trivial fix, but gui controls may require a special treatment again, because they are identified by a pair of gui index / control index, and there will have to be a method for translating that into the index of a single big array that you created.

ivan-mogilko:
There's a problem that I did not realize earlier. This was potentially an issue since I changed object arrays to arrays of pointers (#2273), but it never caused any errors in practice until now.

The problem is that when restoring a save, the object may potentially be registered using a different handle. This was never true (unless a number of objects change), but it is now, because you changed the order of registration for GUI controls.
Restoring a save made by an older engine now breaks the GUI* array.

The only direct solution that I see is to update these arrays of handles when restoring a persistent object.
These places may be found in code by searching for `ccRegisterUnserializedPersistentObject` calls.
So, after each such call, you also need to overwrite a restored handle value into the array, at respective index.

I suppose that for most objects it will be a trivial fix, but gui controls may require a special treatment again, because they are identified by a pair of gui index / control index, and there will have to be a method for translating that into the index of a single big array that you created. One way is to make a helper array that contains a "global" sequential index of a first control of each gui. There's an example of such array in drawing function:

https://github.com/adventuregamestudio/ags/blob/514865fe864cb4ff48cd3f5f669bbdf6ff8ddc2d/Engine/ac/draw.cpp#L369

https://github.com/adventuregamestudio/ags/blob/514865fe864cb4ff48cd3f5f669bbdf6ff8ddc2d/Engine/ac/draw.cpp#L683-L689

ericoporto:
Did a fix for the arrays in https://github.com/adventuregamestudio/ags/pull/2453/commits/b3265d3dfdd7976815902d7f31be7262afc7f087, while things appear to work in my tests I was a bit unsure, for the GUI Controls I instead went back on using a static array of handles for all GUI Controls and did an approach of using a "2D" array.

Engine/ac/gui.cpp
```diff
@@ -660,7 +657,11 @@ void unexport_all_gui_controls()
     {
         unexport_gui_controls(i);
     }
-    StaticGUIControlsArray.clear();
+    StaticGUIControlsHandles.clear();
+
+    // FIX-ME: for some reason this method is called in DoBeforeRestore
+    // so we must reset arrays sizes below. Remove if removed from DoBeforeRestore
+    set_array_all_gui_controls_size();
```
ivan-mogilko, line 664: Why is it necessary to recreate and resize this array after unregistering control exports?

ericoporto: There's a fix-me already there in `DoBeforeRestore`. I can't figure why it's called there, I wonder if this is something related to GUI Controls being different between saves - like, because the game was edited and then it's loading an old save on a newer game, but I couldn't figure how that would fix things.

ivan-mogilko: I know, but I was wondering why do you clear the array and then immediately recreate it. That code did not make any sense to me.

ericoporto: The issue is because it's an array of vector arrays the memory access when restoring when the "sub-array" is not created it will instantaneously crash. But even if it didn't, because the arrays is cleared it's size would not be correct - if the size dynamically adjusted it would be then cause the issue of the memory pointed inside the array being incorrect.

ivan-mogilko: i know that of course; my question was, why clear it in the first place, if you are going to recreate it in the very next line?

ivan-mogilko:
If you remove unexport_gui_controls from DoBeforeRestore, then I think export_gui_controls should be removed from DoAfterRestore? Otherwise calling that may lead to duplicate export (i did not test it).

ericoporto:
OK, I think I got it, the GUI used to be like this

https://github.com/adventuregamestudio/ags/blob/401f27ce52b00b8b94c73b47d31a884d4aa308f8/Engine/ac/gui.cpp

And we had this 

https://github.com/adventuregamestudio/ags/blob/401f27ce52b00b8b94c73b47d31a884d4aa308f8/Engine/acmain/ac_game.cpp#L1812-L1837

I think this is because the GUI Object was inside the GUI itself so the memory that was exported would change.

ivan-mogilko:
> I think this is because the GUI Object was inside the GUI itself so the memory that was exported would change.

Ah.... The engine was using same shared function to
* load gui from game data
* load gui save state

and all the gui arrays were freed and reallocated in memory.

ericoporto:
I now realize that things that weren't in arrays exposed to the game directly like audioclips I forgot. Should those have the same done to them?

ivan-mogilko:
> I now realize that things that weren't in arrays exposed to the game directly like audioclips I forgot. Should those have the same done to them?

Yes, definitely.

Issue #1879, opened 2023-01-04
Player can move into non walkable area and gets stuck


jumpjack:
In some cases it can happen that player moves into a nonwalkable area; once it's there, it is of course impossible to move it away.

A workaround is to add this code in repeatedly_execute() in global script:

```
 if (GetWalkableAreaAtRoom(  player.x,  player.y) == 0) {
   cEgo.PlaceOnWalkableArea()
}

```

But it shouldn't be difficult to implement it directly into AGS code.

I logged these two occurrences of the issue (player.x, player.y, GetWalkableAreaAtScreen, GetWalkableAreaAtRoom):

```
2023/01/04 12:09:21 - x,y,w,w = 164,83,1,1
2023/01/04 12:09:21 - x,y,w,w = 168,85,1,1
2023/01/04 12:09:21 - x,y,w,w = 171,86,1,1
2023/01/04 12:09:21 - x,y,w,w = 174,85,1,1
2023/01/04 12:09:21 - x,y,w,w = 178,83,1,1
2023/01/04 12:09:21 - x,y,w,w = 175,80,0,0
```

Here I even detected a move inside a non-walkable area :

```
2023/01/04 12:09:21 - x,y,w,w = 170,86,1,1
2023/01/04 12:09:21 - x,y,w,w = 173,82,1,1
2023/01/04 12:09:21 - x,y,w,w = 177,80,1,1
2023/01/04 12:09:21 - x,y,w,w = 178,79,1,1
2023/01/04 12:09:21 - x,y,w,w = 175,80,0,0
2023/01/04 12:09:21 - x,y,w,w = 171,82,0,0  <<<--- Note: move performed by engine while player is inside a nonwalkable area
```

jumpjack:
Character gets stuck in locations highlighted by yellow arrows, but NOT in the green arrow:
![image](https://user-images.githubusercontent.com/1620953/210587669-fb607c2d-7e5e-4991-9e70-8d8dfb69355b.png)

Moving by keyboard. Game based on demo Sierra game included in V 3.6.0 RC4.

This is the walkable mask:
![image](https://user-images.githubusercontent.com/1620953/210589428-af25dcbf-5766-4aaf-b73d-8ecdeead05b7.png)


It should be a 320x200 image, but game automatically starts "zoomed" and I don't know how to force it to real 320x200 size.

ivan-mogilko:
[...]

PR #2455, opened 2024-06-22
AGS 4: support creating and loading sprites in a non-default color depth

### LINKED ISSUE #2027: Script API: support pixel format selection for DynamicSprites
**Problem**

Currently AGS converts any sprite or loaded image at runtime to the native game color depth. This may be fine in the majority of cases, but there are still situations where you may want that to not happen.

First of all, user may like to import a 8-bit image to use as a mask in their game (in fact, they may even want 4, 2 or 1-bit mask, something that AGS never supported directly).

Another, perhaps less important, case, is if somebody will like to have a strictly opaque surface and thus have a 24-bit RGB sprite, to ensure that it's always stays opaque regardless of what you draw on it.
In the recent past I've removed the "has alpha channel" flag (#1808, #1813), because it was creating unnecessary difficulties when handling sprites, but in this case the pixel format of the bitmap itself will affect the draw/blend result, so no extra logic in the engine will be necessary (everything is handled by the bitmap vtable implementations).
Allegro bitmap impl already supports 24-bit, and that may be even replaced by a custom 32-bit RGB implementation that keeps alpha channel opaque (0xFF) at all times.
This part is not a priority here though, and may be omitted.

**Proposal**

We need to make sure that the engine does not do implicit conversions unless user lets it. The solution I'm proposing is to add an 
*optional* argument to the DynamicSprite.Create* and LoadFromFile functions, that tells the wanted pixel format. In Create* functions this argument will take place of a recently removed "has alpha" argument, and we may actually pick constant values that match the "has alpha" values for easier transition.

I'm thinking about something like

```
enum PixelFormat {
    // use default game's format for new sprites, or keep source image's format if possible
    ePixelFmt_Default = -1,
    // RGB in game's format (if supported, otherwise - default game's format, e.g. indexed 8-bit)
    ePixelFmt_RGB = 0,
    // ARGB in game's format (if supported, otherwise - see RGB)
    ePixelFmt_ARGB = 1,
    // 8-bit indexed format
    ePixelFmt_Indexed8bit = 2,
    // this may be extended further
};
```

The proposed look of the DynamicSprite functions then will be:
```
DynamicSprite* Create(int width, int height, PixelFormat fmt = ePixelFmt_ARGB);
DynamicSprite* CreateFromExistingSprite(int slot, PixelFormat fmt = ePixelFmt_Default);
DynamicSprite* CreateFromFile(const string filename, PixelFormat fmt = ePixelFmt_Default);
```

Add a new readonly property `readonly PixelFormat DynamicSprite.Format`, for informational purposes.

ivan-mogilko:
Resolves #2027

This implements an ability to create and load sprites in a different color depth than the game's default color depth.
This is not meant for a common use, but special cases, such as having 8-bit masks as sprites in 32-bit games.
Therefore I tried to not overdo with options here and keep new parameters to a reasonable minimum.

Introduced ColorFormat enum:
```
enum ColorFormat
{
  /// Tells to use default game color depth
  eColorFmt_Default = 0,
  eColorFmt_8bit    = 8
};
```

Added `ColorFormat color_fmt` parameter with `eColorFmt_Default` default value to the following functions:
* DynamicSprite.Create,
* DynamicSprite,CreateFromDrawingSurface,
* DynamicSprite.CreateFromExistingSprite
* DynamicSprite.CreateFromFile

This "color_fmt" parameter tells engine of which depth to create this sprite. Value 0 means "use game default". Another supported value that engine currently supports is eColorFmt_8bit (meaning 8-bit indexed bitmap). In theory we could also support 1-bit and 4-bit masks later.

In #2027 we discussed potentially having a "pixel format" enum instead, which lets more fine tuning here. As I mentioned in the conversation there, I had change of opinion regarding that. Such thing may be useful internally inside the engine, but IMO has a very little meaning in script, and will only confuse majority of the users. For 99% use cases it will be enough to have a either a default color depth, or a specific color depth which is cast to *default pixel format corresponding to that depth* by the engine.

If we'll have a reason for defining a precise pixel format, I think that existing "ColorFormat" may be expanded with:
* standard formats equal to corresponding color depth values;
* any extra formats have values above common color depths.

Finally, couple more properties are added which let to read color depth:
```
int DrawingSurface.ColorDepth;
int Game.SpriteColorDepth[];
```


ivan-mogilko:
Hmm, perhaps I found another argument for having enum for color format: it is to have a clear list of which values are supported.

I'll think about it...

ericoporto:
Perhaps it only has to be two formats: game color format and mask color format, and then if it's mask color format it would be possible to use the pathfinder API in this sprite.

ivan-mogilko:
> Perhaps it only has to be two formats: game color format and mask color format

Possibly... except there may be a use for multiple mask formats (starting with 1-bit).

For the starters the enum could be 
* GameDefault = 0 (use default game's color depth)
* 8bit

> Also is it missing enabling Mask Pathfinder here

Right, I tested MaskPathfinder, fixed couple of bugs (in the main repo), and rebased with MaskPathfinder re-enabled.
In my simple test, at least, it works with a loaded 8-bit sprite.

The only strangeness I found is that GetPixel returns -1 (COLOR_TRANSPARENT) for pixels with 0 value, which is inconvenient when you scan the mask yourself.

ivan-mogilko:
I will have to make a break in my work for several days, but I will try to re-think these options in the meantime.

ivan-mogilko:
New simpler variant.

Introduced ColorFormat enum:
```
enum ColorFormat
{
  /// Tells to use default game color depth
  eColorFmt_Default = 0,
  eColorFmt_8bit    = 8
};
```

Added `ColorFormat color_fmt` parameter with `eColorFmt_Default` default value to the following functions:
* DynamicSprite.Create,
* DynamicSprite,CreateFromDrawingSurface,
* DynamicSprite.CreateFromExistingSprite
* DynamicSprite.CreateFromFile

ericoporto:
[MaskPathFinder.zip](https://github.com/user-attachments/files/16047102/MaskPathFinder.zip)

This appears to work. I decided to play with MaskPathFinder, it seems it works exactly as I thought it would using 0 for the transparent 8 bit color and any other color to make it a walkable area, and it also blends with the mask colors when just randomly merging an 8-bit dynamic sprite in a non-8-bit one, which also makes things easier! Two things appeared, but they are in pathfinder and not in the dynamic color.

- Passing a null where a path is accepted moves things to position (0, 0), I believe it should just not do anything and print a warning on log.
- After clicking and walking a lot with the character at  the borders it is possible to click to somewhere where there is a path and get no path after a while. I think it's just it can't use the "fix" in the comment here https://github.com/adventuregamestudio/ags/issues/1879#issuecomment-1903709465.

ivan-mogilko:
> Passing a null where a path is accepted moves things to position (0, 0), I believe it should just not do anything and print a warning on log.

Fixed this one in ags4 branch, it was a mistake in MovePath/WalkPath functions.

> After clicking and walking a lot with the character at the borders it is possible to click to somewhere where there is a path and get no path after a while.

This must be a mistake in the JPS pathfinder. When testing new functions I also noticed that it "cuts corners" sometimes. Easy to see if you make a mask with strict rectangular walls.

PR #2457, opened 2024-06-26
Editor: Add shortcut keys for context menu items and Go to Definition, Find All Usages, Go to Sprite, and dd items to Edit menu

edmundito:
The change contains several improvements to the context menu and the Edit menu for the script editor:

1. Context menu items display the shortcut keys next to them (e.g., Copy shows Ctrl+C)
2. Added new shortcut keys for Go To Definition, Find All Usages, and Go To Sprite (based on Visual Studio conventions)
3. Added those options to the Edit menu
4. Code cleanup

## Screenshots

(My Windows is in Spanish. Mayús. = Shift)
image

image

### DIFF
--- a/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
+++ b/Editor/AGS.Editor/Panes/ScintillaWrapper.cs
@@ -32,7 +32,6 @@ namespace AGS.Editor
         private const string CONTEXT_MENU_CUT = "CtxCut";
         private const string CONTEXT_MENU_COPY = "CtxCopy";
         private const string CONTEXT_MENU_PASTE = "CtxPaste";
-        private const string CONTEXT_MENU_GO_TO_DEFINITION = "CtxDefinition";
 
         private const string THIS_STRUCT = "this";
         private const int INVALID_POSITION = -1;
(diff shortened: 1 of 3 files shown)


edmundito:
I noticed that the Editor's conventions for prepositions in menus may be different from what I was looking at in visual studio. It looks like prepositions should be lowercase, but maybe adjectives should have a capitalized first letter (except for Find next for whatever reason). I may need to update some of the wording to match the rest of the menu.

ericoporto:
Can you use `git rebase -i HEAD~4` and reword all 4 commit messages to start with `Editor: `? This will make it easier to pick them later.

Editor/AGS.Editor/Panes/ScintillaWrapper.cs
```diff
@@ -2450,12 +2449,23 @@ private void ScintillaWrapper_MouseUp(object sender, MouseEventArgs e)
...
-                menu.Items.Add(new ToolStripMenuItem("Cut", Factory.GUIController.ImageList.Images["CutIcon"], onClick, CONTEXT_MENU_CUT));
-                menu.Items[menu.Items.Count - 1].Enabled = this.CanCutAndCopy();
-                menu.Items.Add(new ToolStripMenuItem("Copy", Factory.GUIController.ImageList.Images["CopyIcon"], onClick, CONTEXT_MENU_COPY));
-                menu.Items[menu.Items.Count - 1].Enabled = this.CanCutAndCopy();
-                menu.Items.Add(new ToolStripMenuItem("Paste", Factory.GUIController.ImageList.Images["PasteIcon"], onClick, CONTEXT_MENU_PASTE));
-                menu.Items[menu.Items.Count - 1].Enabled = this.CanPaste();
+
+                ToolStripMenuItem menuItem;
+
+                menuItem = new ToolStripMenuItem("Cut", Factory.GUIController.ImageList.Images["CutIcon"], onClick, CONTEXT_MENU_CUT);
```
ericoporto, line 2455: Question, what about use the `using` keyword here? Like `using( var item = new ....){...item.Enabled = ...}`, I feel it would nicely group things but I don't know if it's the correct word.

edmundito: I wasn't super familiar with the using stement in C# so [I looked up the doc](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using), and it seems that the use is when the object needs to be disposed after being used. But in this case, the ToolStripMenuItem doesn't need to be disposed until the context menu is  disposed, since it's being added to a list and rendered later.

edmundito:
> Can you use `git rebase -i HEAD~4` and reword all 4 commit messages to start with `Editor: `? This will make it easier to pick them later.

Done and repushed. Are these commit conventions documented somewhere?

Editor/AGS.Editor/Panes/ScintillaWrapper.cs
persn: If the goal is to group things you can use brackets without attaching them to any keywords

```cs
var something = ....

{
    var foo = Factory.Create<IFoo>();
    foo.Bar = ?;
}
```

People usually don't do that though, keeping variable declaration and property assignments together is often considered enough

ivan-mogilko:
>  Are these commit conventions documented somewhere?

They were suggested many years ago on forums, but I don't remember if we have them in the project wiki.

One notable convention (and this is a general convention in git, afaik) to have commit title <= 72 symbols, because terminals and web interfaces tend to cut it after 72th character. So a rule of thumb is to keep title within 72 char length, and then put a detailed description of any necessary length after couple of linebreaks.

ivan-mogilko:
I do not see any problems here, but one mildly annoying thing is menu separators. I only ever noticed that because there's a commit removing one.

What is the convention for separators in menu, and by which principle are the items grouped?

For instance, "Show autocomplete" and "Match brace" refer to the current cursor position, same is "Go to definition", but they are in different groups.
At the same time "Go to Line" is not related to the cursor position, but it's in the same group as "Match brace".
"Switch to matching script or header" is a completely unrelated command, but it's grouped with the few other "context" commands.

Now, also, there's a separator after "Go to Sprite" in context menu, but not in the Edit menu.

edmundito:
@ivan-mogilko, do you have any suggestions on what you'd like to change about the separators?

My suspicion is because it's based on Visual Studio, there are many separators. I placed the context options above the breakpoint to match the order already in the context menu. Here's an example from VS:

image

ivan-mogilko:
No, I don't have any exact suggestion about grouping items at the moment. Perhaps someone will be willing to fix this later.

My only nitpick is that there was a separator between "base" items created by ScriptEditorBase and items created by derived classes, such as ScriptEditor. Now this separator is gone in the Edit menu, but it is *still present* in the context menu.

I take a guess that you removed it, because it caused extra trailing separator in the Dialog editor, as Dialog editor does not add any more commands.

So, perhaps it would be more proper to prepend it before extra commands when ScriptEditor adds them here:
https://github.com/adventuregamestudio/ags/blob/70513e44604b014be3bc4814e2c206d2e68673f2/Editor/AGS.Editor/Panes/ScriptEditor.cs#L137-L141

This will be consistent with a separator prepended for extra context menu commands, as seen here:
https://github.com/adventuregamestudio/ags/blob/70513e44604b014be3bc4814e2c206d2e68673f2/Editor/AGS.Editor/Panes/ScriptEditor.cs#L143-L148

---

For better command grouping, ScriptEditorBase might have to define "groups of menu commands", to which the child classes would add theirs. That would require a bigger change to how the menu is configured in these classes.

edmundito:
@ivan-mogilko At the very least I added a change that pretends a separator for the custom menu commands in the script editor, so that it's consistent with the context menu. I definitely think the menus need some rework, but that is a bigger task outside the PR.

PR #2460, opened 2024-06-30
AGS 4: allow to import sprites as 8-bit image regardless of game color depth

ivan-mogilko:
This complements #2455.

Adds a combobox to the "Import Sprite" window that lets select whether to import the sprite, converting to game's default color depth (default choice), or import as a 8-bit image. This list may potentially be expanded, if necessary.
spriteimport--importdepth2

The choices are currently named:
* "In Game Color Depth"
* "As 8-bit image"

At first I wanted the second option to be displayed only if the imported image is of indexed pixel format, but this may be complicated to do, as multiple different files may be selected at once.

Removed "Import alpha channel" option, as redundant.

Other things...
Sprite index file now stores additional array of bytes, defining sprite's color depth. This is to let know the sprite's color depth without loading it up.

---

**NOTE:** for testing 8-bit sprites as walkable masks following settings must be used:
* "As 8-bit image"
* Remap to game palette = FALSE
* Transparency = Leave As Is (otherwise color 0 is fixed to color 16 !!)

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -400,7 +400,7 @@ HError SpriteCache::InitFile(std::unique_ptr<Stream> &&sprite_file,
 {
     Reset();
 
-    std::vector<Size> metrics;
+    std::vector<GraphicResolution> metrics;
     HError err = _file.OpenFile(std::move(sprite_file), std::move(index_file), metrics);
     if (!err)
         return err;
@@ -417,6 +417,7 @@ HError SpriteCache::InitFile(std::unique_ptr<Stream> &&sprite_file,
             _spriteData[i].Flags = SPRCACHEFLAG_ISASSET;
             _sprInfos[i].Width = metrics[i].Width;
             _sprInfos[i].Height = metrics[i].Height;
+            _sprInfos[i].ColorDepth = metrics[i].ColorDepth;
             Size newsz = _callbacks.AdjustSize(Size(metrics[i].Width, metrics[i].Height), _sprInfos[i].Flags);
             _sprInfos[i].Width = newsz.Width;
             _sprInfos[i].Height = newsz.Height;
--- a/Common/ac/spritefile.cpp
+++ b/Common/ac/spritefile.cpp
@@ -166,7 +166,7 @@ SpriteFile::SpriteFile()
 }
 
 HError SpriteFile::OpenFile(std::unique_ptr<Stream> &&sprite_file,
-    std::unique_ptr<Stream> &&index_file, std::vector<Size> &metrics)
+    std::unique_ptr<Stream> &&index_file, std::vector<GraphicResolution> &metrics)
 {
     Close();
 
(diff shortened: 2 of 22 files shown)


ivan-mogilko:
Also, maybe it's the time to remove "Import alpha channel" option? How useful it is anyway?
The only difference it currently makes is that if sprite has <255 alpha anywhere, this option will turn the image to full opaque.
But I suppose that if user wants an opaque image, then they should make one in the first place.

ericoporto:
ahn, how is this supposed to work? I tried to add an 8-bit color sprite (
[test8bit.zip](https://github.com/user-attachments/files/16090916/test8bit.zip) this png) and the colors 
and got the following 

```
Color 0,0=16
Color 146,64=18
Color 288,182=10
```

by using 

```AGS Script
  dynspr = DynamicSprite.CreateFromExistingSprite(1, eColorFmt_8bit);
  DrawingSurface* surf = dynspr.GetDrawingSurface();
  int c1 = surf.GetPixel(0, 0);
  int c2 = surf.GetPixel(146, 64);
  int c3 = surf.GetPixel(288, 182);
  System.Log(eLogInfo, "Color 0,0=%d\nColor 146,64=%d\nColor 288,182=%d", c1, c2, c3);
```

I expected to get 0, 1, and 9 which are the indexes of the colors. :/

ivan-mogilko:
@ericoporto ok, I pushed a fix to the sprite import, and now your png is loaded correctly as a 8-bit.
The test returns this:

```
[Script][Info]: Color 0,0=-1
Color 146,64=1
Color 288,182=9
```

-1 is returned by GetPixel in case of a transparent color (0).
Curiously, your image did not have color 0 replaced by 16...

ericoporto:
Uhm, that is a tricky one, I think the color transparent is alright for bliting/blending and should give the correct result visually, but I think the index zero is expected to be the "wall"/"non-walkable" areas when using MaskPathFinder. My idea would be to secretly replace color transparent by zero for the mask pathfinder somehow.

ivan-mogilko:
It's not related exclusively to "maskpathfinder", reading pixels from room masks gives the same result, and has same convenience problem.

ivan-mogilko:
Two last commits are unrelated to this new setting, but fix couple of mistakes in importing indexed images. (also cherry picked them to the master branch).

From what I see, the only remaining annoyance here is the aforementioned GetPixel behavior, which returns COLOR_TRANSPARENT constant instead of actual color value for the mask color. I believe that's a matter of DrawingSurface api.

ericoporto:
I retried this and I think now it's correct

- The 0 index is properly zero for the MarkPathFinder, it will be correctly considered a wall
- It does get understood as COLOR_TRANSPARENT instead of zero when using GetPixel, but I think this can just be documented and left as is.
- The 8-bit image when draw uses the game palette, even in 32-bit mode, I think this is alright, for the purpose that I had in mind as masks, it looks consistent with the other masks.

PR #2462, opened 2024-07-03
[ags4] Adjust Audio Clips imports to use handles

ericoporto:
Continuing from https://github.com/adventuregamestudio/ags/pull/2453#issuecomment-2203865445

I am leaving as a draft for now because I am trying to figure if something has to change in `Game.AudioClips[i]` script interface

https://github.com/adventuregamestudio/ags/blob/4d5d29e87e4f00aa800dd18a66e2219cf92f84b8/Engine/ac/game.cpp#L778-L783

and

https://github.com/adventuregamestudio/ags/blob/4d5d29e87e4f00aa800dd18a66e2219cf92f84b8/Editor/AGS.Editor/Resources/agsdefns.sh#L2473

The reason I ask is because I noticed it returns a pointer to `ScriptAudioClip` and wasn't sure if it would be the same.

### DIFF
--- a/Editor/AGS.Editor/Tasks.cs
+++ b/Editor/AGS.Editor/Tasks.cs
@@ -780,7 +780,7 @@ namespace AGS.Editor
         {
             foreach (AudioClip clip in clips.AllItemsFlat)
             {
-                sb.AppendLine("import AudioClip " + clip.ScriptName + ";");
+                sb.AppendLine("import AudioClip *" + clip.ScriptName + ";");
             }
         }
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
You are exporting handles, under which `game.audioClips[i]` are registered. This means that when resolving such handle, engine will get `game.audioClips[i]`.

Game_GetAudioClip returns `game.audioClips[index]`. Which is literally the same thing.

ericoporto:
I did a few tests, and I noticed that it worked fine, including comparison like `aStream01 != Game.AudioClips[0]`, so it seems this is correct indeed.

PR #2464, opened 2024-07-05
Editor: add support for sub items in file menu and move global messages upgrade actions to submenu

ericoporto:
This is the minimal amount of code I could think that give support to submenus. The other approach would be to make the property read-only and then make so you can only pass such list in Constructor - which would make it more similar to the other properties.

I noticed the following todo:

```
// TODO: I do not see any way to schedule sub-menus in this system!?
// but if it's supported, maybe put these 2 global messages commands int a submenu
```

Decided to try to fix it. Not sure on the name of the group as "Global Messages upgrade".

image

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -261,6 +261,16 @@ namespace AGS.Editor
                     {
                         RegisterMenuCommand(command.ID, plugin);
                     }
+
+                    if(command.SubCommands != null && command.SubCommands.Count > 0)
+                    {
+                        foreach (MenuCommand scommand in command.SubCommands)
+                        {
+                            RegisterMenuCommand(scommand.ID, plugin);
+                            scommand.IDPrefix = plugin.ComponentID + CONTROL_ID_SPLIT;
+                        }
+                    }
+
                     command.IDPrefix = plugin.ComponentID + CONTROL_ID_SPLIT;
                 }
 				_menuManager.AddMenuCommandGroup(commands);
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -251,29 +251,30 @@ namespace AGS.Editor
 			_menuManager.AddMenu(id, title, insertAfterMenu);
 		}
 
-		public void AddMenuItems(IEditorComponent plugin, MenuCommands commands)
+        private void RegisterMenuItems(IEditorComponent plugin, IList<MenuCommand> commands)
+        {
+            foreach (MenuCommand command in commands)
+            {
+                if (command.ID != null)
+                {
+                    RegisterMenuCommand(command.ID, plugin);
+                }
+
+                command.IDPrefix = plugin.ComponentID + CONTROL_ID_SPLIT;
+
+                if (command.SubCommands != null && command.SubCommands.Count > 0)
+                {
+                    RegisterMenuItems(plugin, command.SubCommands);
+                }
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Perhaps name it simply "Global Messages"?

ericoporto:
Redid as following

image

Editor/AGS.Types/EditorFeatures/MenuCommand.cs
```diff
@@ -49,6 +49,7 @@ public MenuCommand(string itemID, string itemName, Keys shortcutKey, string icon
 		private bool _checked;
 		private Keys _shortcutKey = Keys.None;
         private string _shortcutKeyDisplayString = string.Empty;
+		public List<MenuCommand> SubCommands { get; set; }
```
ivan-mogilko, line 52: Please, move this SubCommands out away from the list of "private" fields, to make it stand out as a public property. This is just to help notice it.

Editor/AGS.Editor/GUI/GUIController.cs
```diff
@@ -261,6 +261,16 @@ public void AddMenuItems(IEditorComponent plugin, MenuCommands commands)
                     {
                         RegisterMenuCommand(command.ID, plugin);
                     }
+
+                    if(command.SubCommands != null && command.SubCommands.Count > 0)
+                    {
+                        foreach (MenuCommand scommand in command.SubCommands)
+                        {
+                            RegisterMenuCommand(scommand.ID, plugin);
```
ivan-mogilko, line 269: You need to test if scommand.ID is not null, similar to how it's tested for command.ID above.
Otherwise, having separators in submenu would crash the editor.

ivan-mogilko: But, on another hand, I just realized, that your submenu handling does not allow nested submenu levels.
Probably this would require recursive call to a helper function, having IList<MenuCommand> as a parameter

ivan-mogilko:
I confirm that submenu works as expected, but I think this system may be expanded to allow nested submenus (submenus in submenus), if you pick out menu item registration / addition into helper function(s) that is called recursively over SubCommands.

ericoporto:
I am not 100% sure, but I think I figured how to support it recursively to register the submenus. I am not completely sure because there are a few dictionaries in things like `AddMenuCommandGroup` and `RemoveMenuCommandGroup` I am not completely sure how it's supposed to work.

Editor/AGS.Types/EditorFeatures/MenuCommand.cs
ericoporto: I added a private variable with it and moved a public getter. Remade so it can only be set on creation.

Editor/AGS.Editor/GUI/GUIController.cs
ericoporto: I used your code and modified AddMenuItem so it could be recursive.

ivan-mogilko:
I think it's working right.

>  I am not completely sure because there are a few dictionaries in things like AddMenuCommandGroup and RemoveMenuCommandGroup

I quickly checked that out, and these dictionaries store MenuCommands objects in a list mapped to a key. All the items and submenus are stored internally inside MenuCommands, so should not affect these dicts.

Editor/AGS.Types/EditorFeatures/MenuCommand.cs
ivan-mogilko: > Remade so it can only be set on creation.

I don't think that matters tbh, all other properties have public setters, and if you give out List<MenuCommand> then this collection may be modified anyway.

PR #2465, opened 2024-07-07
Editor: add open recent games to file menu

ericoporto:
Adds a lot of machinery to allow replacing menu commands enough so that recent games can be updated

This adds the following menu

image

Please review this carefully as I had to do a lot of bolting between different things for this to work. This is also new surface of bugs, if we decided this is unwanted, it's alright! An alternative could be to have an entry to bring up the initial screen of AGS.

The good: it actually works
The bad: code is a mess (I couldn't figure what should be in GUIController and what should be in main menu manager!)

Still, I thought it was a good things to have in the editor to easily alternate between games in the editor.

Edit: actually I think I can always hide the most recent game as it will always be the one being edited

### DIFF
--- a/Editor/AGS.Types/EditorFeatures/MenuCommand.cs
+++ b/Editor/AGS.Types/EditorFeatures/MenuCommand.cs
@@ -112,6 +112,7 @@ namespace AGS.Types
 		public IList<MenuCommand> SubCommands
 		{
 			get { return _subCommands; }
+			set { _subCommands = value; }
 		}
 	}
 }
--- a/Editor/AGS.Types/EditorFeatures/MenuCommand.cs
+++ b/Editor/AGS.Types/EditorFeatures/MenuCommand.cs
@@ -14,15 +14,6 @@ namespace AGS.Types
 			get { return new MenuCommand(null, MenuCommand.MENU_TEXT_SEPARATOR); }
 		}
 
-		public MenuCommand(string itemName, IList<MenuCommand> subCommands)
-		{
-			_id = null;
-			_name = itemName;
-			_iconKey = null;
-			_enabled = true;
-			_subCommands = subCommands;
-		}
-
 		public MenuCommand(string itemID, string itemName)
 		{
 			_id = itemID;
@@ -44,10 +35,15 @@ namespace AGS.Types
 		}
 
 		public MenuCommand(string itemID, string itemName, Keys shortcutKey, string iconKey)
-			: this(itemID, itemName)
+			: this(itemID, itemName, shortcutKey)
 		{
-			_shortcutKey = shortcutKey;
 			_iconKey = iconKey;
+		}
+
+		public MenuCommand(string itemID, string itemName, Keys shortcutKey, string iconKey, IList<MenuCommand> subCommands)
+			: this(itemID, itemName, shortcutKey, iconKey)
+		{
+			_subCommands = subCommands;
 		}
 
 		private string _id;
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -412,6 +412,7 @@
     <Compile Include="Utils\BitmapExtensions.cs" />
     <Compile Include="Utils\CommandLineOptions.cs" />
     <Compile Include="Utils\ConcurrentCircularBuffer.cs" />
+    <Compile Include="Utils\Debounce.cs" />
     <Compile Include="Utils\MathExtra.cs" />
     <Compile Include="Utils\ScriptFileUtilities.cs" />
     <Compile Include="Utils\ScriptGeneration.cs" />
--- a/Editor/AGS.Editor/Components/FileCommandsComponent.cs
+++ b/Editor/AGS.Editor/Components/FileCommandsComponent.cs
@@ -12,6 +12,7 @@ namespace AGS.Editor.Components
     class FileCommandsComponent : BaseComponent
     {
         private const string OPEN_GAME_COMMAND = "OpenGame";
+        private const string OPEN_RECENT_GAME_COMMAND = "OpenRecentGame";
         private const string SAVE_GAME_COMMAND = "SaveGame";
         private const string GAME_STATS_COMMAND = "GameStatistics";
         private const string JUMP_TO_EVENTS_TAB_COMMAND = "JumpToEventsTab";
@@ -50,6 +51,7 @@ namespace AGS.Editor.Components
 
             MenuCommands commands = new MenuCommands(GUIController.FILE_MENU_ID, 0);
             commands.Commands.Add(new MenuCommand(OPEN_GAME_COMMAND, "&Open...", Keys.Control | Keys.L, "MenuIconOpen"));
+            commands.Commands.Add(GetOpenRecentMenuCommand());
             commands.Commands.Add(new MenuCommand(SAVE_GAME_COMMAND, "&Save", Keys.Control | Keys.S, "MenuIconSave"));
             commands.Commands.Add(new MenuCommand(GAME_STATS_COMMAND, "&Game statistics", Keys.F2, "MenuIconStatistics"));
             commands.Commands.Add(new MenuCommand(JUMP_TO_EVENTS_TAB_COMMAND, "&Go to Events grid", Keys.F4, "MenuIconGoToEventsGrid"));
@@ -68,7 +70,7 @@ namespace AGS.Editor.Components
                 new MenuCommand(EXPORT_GLOBAL_MESSAGES_TO_SCRIPT_COMMAND, "Export Global Messages to script"),
                 new MenuCommand(REMOVE_GLOBAL_MESSAGES_COMMAND, "Remove Global Messages")
             };
-            commands.Commands.Add(new MenuCommand("Global Messages", subCommands));
+            commands.Commands.Add(new MenuCommand(null, "Global Messages", Keys.None, null, subCommands));
             _guiController.AddMenuItems(this, commands);
 
             commands = new MenuCommands(GUIController.FILE_MENU_ID, 800);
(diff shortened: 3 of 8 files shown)


Editor/AGS.Editor/AGSEditor.cs
```diff
@@ -1400,6 +1400,7 @@ public bool SaveGameFiles()
             }
             Settings.RecentGames.Insert(0, recentGame);
             Settings.Save();
+            Factory.ComponentController.FindComponent<AGS.Editor.Components.FileCommandsComponent>()?.RefreshRecentGamesMenu();
```
ivan-mogilko, line 1403: Right from the start, this is better done using event, where AGSEditor declares event, and FileCommandsComponent subscribes to event. This way AGSEditor does not need to know that it must call some component.

Editor/AGS.Editor/Tasks.cs
```diff
@@ -169,6 +169,7 @@ public bool LoadGameFromDisk(string gameToLoad, bool interactive)
                 Factory.AGSEditor.Settings.RecentGames.Remove(recentGame);
             }
             Factory.AGSEditor.Settings.RecentGames.Insert(0, recentGame);
+            Factory.ComponentController.FindComponent<AGS.Editor.Components.FileCommandsComponent>()?.RefreshRecentGamesMenu();
```
ivan-mogilko, line 172: Same here, it's better to use event, which FileCommandsComponent subscribes to.

Editor/AGS.Editor/AGSEditor.cs
ericoporto: That event gets called more than once when it's called, and it's named as being a limit. I can do that, but this function will be called quite a few times where this approach calls it only once on update.

ivan-mogilko: > That event gets called more than once when it's called, and it's named as being a limit. I can do that, but this function will be called quite a few times where this approach calls it only once on update.

I'm sorry, I don't understand. Which event is "called more than once", and what does "named as being a limit" means?
How it will be different between calling a function here and calling a event?

ericoporto: I created a debouncer utility and used it to implement it on top of the `ListChangeUpdate` event and then subscribed to the event in `FileCommandsComponent`. There is an existing `ListChanged` for `RecentGames` but it's used for limiting it's size and every insertion or removal calls it (so it even calls itself once when it does a removal).

ivan-mogilko: Ah, I see now, you were referring to a "recentgames list changed" event inside the AppSettings.

Editor/AGS.Types/EditorFeatures/MenuCommand.cs
```diff
@@ -14,9 +14,9 @@ public static MenuCommand Separator
 			get { return new MenuCommand(null, MenuCommand.MENU_TEXT_SEPARATOR); }
 		}
 
-		public MenuCommand(string itemName, IList<MenuCommand> subCommands)
+		public MenuCommand(string itemName, IList<MenuCommand> subCommands, string itemId = null)
```
ivan-mogilko, line 17: I have a proposal here, to instead make 2 separate constructors, one of which has itemId as a first parameter, and another does not have it.

My feeling is that this will be more consistent with existing constructors which have itemId first param, as it's the priority property for the MenuCommand, and more obvious for coders.
(Note that since this is in AGS.Type, it's practically a Editor's api)

ivan-mogilko:
I looked through the code, and I think that understand what is being done here more or less.

There's indeed an issue with menu managing code split between GuiController and MenuManager, but that's an older problem, and that's all internal editor's code, so may be left as it is and considered for refactor later.

There's one thing that I found strange: why do you replace the parent menu command, and not just its subitems? I thought it's more logical to replace the submenu contents.

ericoporto:
> There's one thing that I found strange: why do you replace the parent menu command, and not just its subitems?

It would take a lot more IDs to do so, and the amount of those would be variable, replacing the parent is easier, I create a new one with the correct children and just remove the previous one to put the new one in place.

ivan-mogilko:
> It would take a lot more IDs to do so, and the amount of those would be variable

I mean, passing a parent's ID and telling to replace all of its subitems with a new list of subitems.

Editor/AGS.Types/EditorFeatures/MenuCommand.cs
ericoporto: I can't actually do this because a different constructor gets used if I move the `IList<MenuCommand>` to the right, the null for iconKey or shortcutKey gets passed to subCommands. I need to think how to do this.

ericoporto: I did it like this https://github.com/adventuregamestudio/ags/pull/2465/commits/20f7833f719beb8e97c181b50823ca4e3089c153

ericoporto:
> I mean, passing a parent's ID and telling to replace all of its subitems with a new list of subitems.

I think that method would have a more complex name and it would be harder to explain it, this way it looks more reusable.

Please tell me if there is something else, I would like to squash the commits to have it all in a single one.

ivan-mogilko:
> > I mean, passing a parent's ID and telling to replace all of its subitems with a new list of subitems.
> 
> I think that method would have a more complex name and it would be harder to explain it, this way it looks more reusable.

My opinion on this is opposite, imo the concept of updating submenu contents is something more common than a replacement of a menu item with another.

ericoporto:
I don't think I agree, but here is a commit that behaves like this. I am still replacing the toolstrip because it would be more complicated not to, but the commands and the other things I instead readjust the subcommands only.

It's still not complete (I think) for subcommands that have subcommands, but I think I will leave this for the future if the need arises.

ivan-mogilko:
Thank you.
I tested this and it appears working.

There are couple of things I found regarding the behavior.
1. When the "recent games" list is empty the menu item becomes a non-expanding item, clicking on which simply closes the menu. I recall that in such case programs usually add a single disabled sub-item with "(empty)" label. Would it be possible to add one here?
2. While the project is being loaded, there's no visual indication of this process. If it takes some time, the Editor appears hanging. I suppose there could be at least this usual BusyDialog box with "rotating circle" icon to let user know what is happening.
But this may be a planned to do separately.

ericoporto:
>  I recall that in such case programs usually add a single disabled sub-item with "(empty)" label. Would it be possible to add one here?

That looks easy to do, I can add that!

> While the project is being loaded, there's no visual indication of this process.

Yeah, I am using the same thing that open uses already in the editor, I need to look into how the rotating circle in ags editor works to see if I can add that.

ericoporto:
@ivan-mogilko  adding a busy dialog is more difficulty than I imagined because it breaks the Editor refresh due to being called from a different thread

```
An error occurred whilst trying to load your game. The error was:
Cross-thread operation not valid: Control 'frmMain' accessed from a thread other than the thread it was created on.

Error details: System.InvalidOperationException: Cross-thread operation not valid: Control 'frmMain' accessed from a thread other than the thread it was created on.
   at System.Windows.Forms.Control.get_Handle()
   at System.Windows.Forms.Control.set_WindowText(String value)
   at System.Windows.Forms.Form.set_WindowText(String value)
   at System.Windows.Forms.Control.set_Text(String value)
   at System.Windows.Forms.Form.set_Text(String value)
   at AGS.Editor.GUIController.GameNameUpdated()
   at AGS.Editor.AGSEditor.RefreshEditorAfterGameLoad(Game newGame, CompileMessages errors)
   at AGS.Editor.Tasks.LoadGameFromDisk(String gameToLoad, Boolean interactive)
   at AGS.Editor.InteractiveTasks.LoadGameFromDisk(String gameToLoad)
```

ericoporto:
I found a way that almost works in https://github.com/adventuregamestudio/ags/pull/2465/commits/d2644387b090b17859b8581ca606e18b1a544e73 , but the open recent menu entry becomes blank after this for some reason, after you click any of the Open Recent entries... There is some issue between loading in a different thread and the ui not loading correctly.

ivan-mogilko:
Maybe leave this without BusyDialog for now, and try that separately?

ericoporto:
I think I got it working, can you please try after this commit and see if it is working for you?

ivan-mogilko:
I think since it affects other things, it's better to do this change separately.

For example, there may be a design issue with RefreshEditorAfterGameLoad, a public interface calling Invoke in itself. This raises a question: how do you tell if a public method should call Invoke on its actions? maybe other methods should too?
Here it calls because it "guesses" that it may be called from another thread, but this looks like a inverted logic. It should likely be the caller to know whether to do Invoke if they call this from another thread.

ericoporto:
I dropped because that will be a lot more work that has to be done separately. Besides, this issue already happens - just hit `Open...` after loading a game and try to load a game that takes a little to load.

ericoporto:
Just to check, progress and busy dialog are different things here right? Like progress is for when there's some notion of the task going forward and busy is just "waiting for something we have no idea how it's advancing". Just trying to understand if what you mentioned has some relationship with #2062 , I am still not sure how to deal with "InvokeRequired" check.

ivan-mogilko:
> Just to check, progress and busy dialog are different things here right? Like progress is for when there's some notion of the task going forward and busy is just "waiting for something we have no idea how it's advancing".

They are different dialogs by form, but not by what they indicate. As I mentioned elsewhere (probably in #2062), in my opinion it's preferable to make both of them support same amount of information about progress (defined by IWorkProgress interface).
BusyDialog already implements IWorkProgress, and can display progress in a form of "X/Y completed" label. But there's no problem in adding a progress bar to the BusyDialog too, i just did not bother at the time.

The difference between them two is simply that BusyDialog is a modal form that blocks the rest of the editor, while Progress form is more suitable to be created as a child of another form to run in parallel with the editor, performing some background update.

> I am still not sure how to deal with "InvokeRequired" check.

"Invoke" is needed when a method, or operation, must be run on GUI thread. In your previous commit that i commented you had a large group of seemingly unrelated operations inside Invoke. This is what I thought as incorrect. IMO it's better to pinpoint particular operations that require to be run on a GUI thread, and then either
* implement InvokeRequired check and Invoke call in them
* do this check and call Invoke in the calling code, over these particular methods.

ivan-mogilko:
To give an example, there's a GUIController.GameNameUpdated(). This method updates GUI, so must be done on GUI thread.

GUIController already has some methods doing InvokeRequired check before changing GUI. E.g. see SetStatusBarText, ShowCallStack or ShowFindSymbolResults.

Following their example, GameNameUpdated may be changed as:
```
public void GameNameUpdated()
{
    if (InvokeRequired)
    {
        Invoke(new ParameterlessDelegate(GameNameUpdated));
        return;
    }

    _mainForm.Text = _titleBarPrefix + _agsEditor.CurrentGame.Settings.GameName + " - AGS Editor";
}
```

PR #2466, opened 2024-07-07
Editor: fix global variables and text parser pane scaling

ericoporto:
prevents text from being cut vertically, both panes are very similar (a group with a label and a control), so I just used docking at top for label and fill for the main control.

this is a minimal fix part of #2087


ivan-mogilko:
Is it possible to increase the top label's height and maybe add some padding to these controls? Right now they look bit too close to each other and frame borders:
globalvariables-narrow

ericoporto:
Adjusted it slightly, it looks like this to me

image

PR #2467, opened 2024-07-08
Editor: adds toggle line comments in script editor

ericoporto:
Two modes of comment/uncomment are supported through Editor Preferences (default `addsSpace` is false):

- addsSpace == false: comment is inserted as the first character of the line, uncommenting will not modify space characters

- addsSpace == true: comment is inserted as the first indentation character with and additional space after, uncommenting will remove one space character after if it exists.

I am trying to remake #1575 in a way that is acceptable, as this is a feature I constantly miss. The method I prefer to use for myself when coding is the one that adds a space.

I used `Shift` + `Ctrl` + `Q` for the shortcut, if a different one is preferred please tell me.

### DIFF
--- a/Editor/AGS.Editor/Entities/EditorPreferences.cs
+++ b/Editor/AGS.Editor/Entities/EditorPreferences.cs
@@ -718,6 +718,24 @@ namespace AGS.Editor.Preferences
         }
 
         [Browsable(true)]
+        [DisplayName("Add space when using toggle line comments")]
+        [Description("If it should add space between \"//\" line comments and what is commented when toggling.")]
+        [Category("Script Editor")]
+        [UserScopedSettingAttribute()]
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
I'm sorry for reviewing this late, I've got very distracted on the last two weeks.

I don't usually use such feature in text editors, so cannot judge on the idea itself. But it seems to work as stated, and the new code is contained in a single function, so should not conflict with anything and will be easy to fix if something wrong is found. So I think it's fine.

The only suggestion i have is to group "toggle comments" menu item with "toggle breakpoint" (place them close together), because both relate to changes in current script, while "Switch to Matching Script or Header" is a unique kind of action.

ivan-mogilko:
Did a small addition: ad31da24686e6f78360a8a6bc827b5b914d4f089 for consistency

PR #2469, opened 2024-07-11
Editor: resize room masks when changing the background to a different size instead of just resetting them.

KeyserSoSay0:
modified:   Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs

	Make it so that room masks are resized instead of reset when the resolution of the background changes

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1760,12 +1760,6 @@ namespace AGS.Editor.Components
                 _backgroundCache[background] = new RoomImage(newBmp, true);
             }
 
-            // If size or resolution has changed, reset masks
-            if (hasResolutionChanged)
-            {
-                ResetMaskCache();
-            }
-
             _loadedRoom.Modified = true;
         }
 
@@ -1842,6 +1836,41 @@ namespace AGS.Editor.Components
             if (x < 0 || y < 0 || x > mask.Width || y > mask.Height) return 0;
 
             return mask.GetRawData()[(y * mask.Width) + x];
+        }
+
+        void IRoomController.ResetMasks()
+        {
+            foreach (RoomAreaMaskType mask in Enum.GetValues(typeof(RoomAreaMaskType)))
+            {
+                if (mask == RoomAreaMaskType.None)
+                    continue;
+
+                if (_maskCache.ContainsKey(mask))
+                {
+                    _maskCache[mask]?.Dispose();
+                    _maskCache[mask] = new RoomImage(CreateMaskBitmap(mask, _loadedRoom.Width, _loadedRoom.Height), true);
+                }
+            }
+        }
+
+        void IRoomController.ResizeMasks(bool doScale, int newWidth, int newHeight, int xOffset, int yOffset)
+        {   
+            if (newWidth <= 0 || newHeight <= 0)
+            {
+                throw new ArgumentException("Negative width or height set for mask resize.");
+            }
+
+            foreach (RoomAreaMaskType mask in Enum.GetValues(typeof(RoomAreaMaskType)))
+            {
+                if (mask == RoomAreaMaskType.None)
+                    continue;
+
+                if (_maskCache.ContainsKey(mask))
+                {
+                    _maskCache[mask].Image = ResizeMaskBitmap(_maskCache[mask].Image, mask, doScale, newWidth, newHeight, xOffset, yOffset);
+                    _maskCache[mask].Modified = true;
+                }
+            }
         }
 
         void IRoomController.DrawRoomBackground(Graphics g, int x, int y, int backgroundNumber, int scaleFactor)
@@ -2102,6 +2131,20 @@ namespace AGS.Editor.Components
             return bitmap;
         }
 
+        /// <summary>
+        /// Resizes a mask bitmap for the given mask type of the requested size.
+        /// </summary>
+        private Bitmap ResizeMaskBitmap(Bitmap originalBitmap, RoomAreaMaskType mask, bool doScale, int newWidth, int newHeight, int xOffset, int yOffset)
+        {
+            // Adjust by the mask scale
+            double scale = _loadedRoom.GetMaskScale(mask);
+
+            int drawWidth = doScale ? newWidth : originalBitmap.Width;
+            int drawHeight = doScale ? newHeight : originalBitmap.Height;
+
+            return BitmapExtensions.ResizeScaleAndOffset(originalBitmap, drawWidth, drawHeight, (int)(newWidth * scale), (int)(newHeight * scale), xOffset, yOffset);
+        }
+
         private Bitmap LoadMask(RoomAreaMaskType mask) => BitmapExtensions.LoadNonLockedBitmap(_loadedRoom.GetMaskFileName(mask));
 
         private void RefreshMask(RoomAreaMaskType mask)
@@ -2183,21 +2226,6 @@ namespace AGS.Editor.Components
                 {
                     _maskCache[mask]?.Dispose();
                     _maskCache[mask] = null;
-                }
-            }
-        }
-
-        private void ResetMaskCache()
-        {
-            foreach (RoomAreaMaskType mask in Enum.GetValues(typeof(RoomAreaMaskType)))
-            {
-                if (mask == RoomAreaMaskType.None)
-                    continue;
-
-                if (_maskCache.ContainsKey(mask))
-                {
-                    _maskCache[mask]?.Dispose();
-                    _maskCache[mask] = new RoomImage(CreateMaskBitmap(mask, _loadedRoom.Width, _loadedRoom.Height), true);
                 }
             }
         }
--- a/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
+++ b/Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
@@ -545,6 +545,9 @@ namespace AGS.Editor
                     bool doImport = true;
                     bool deleteExtraFrames = false;
                     bool roomSizeChanged = false;
+                    AdjustMaskOptions MaskOption = AdjustMaskOptions.ResetMask;
+                    int MaskXOffset = 0;
+                    int MaskYOffset = 0;
 
                     if ((Factory.AGSEditor.CurrentGame.Settings.ColorDepth == GameColorDepth.Palette) && (bmp.GetColorDepth() > 8))
                     {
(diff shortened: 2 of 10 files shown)


KeyserSoSay0:
oh man, I have no idea what I'm doing with git or github.  I was just trying to put this into my own branch of ags but this seems like I put in a pull request for the real ags?  I guess I'll just go with it. 

I'm not sure this is ready yet but this is the gist of what I want to do.  Basically, for me it's a fairly common workflow to want to change a background slightly or just resize it down without completely losing all of the masks.  So I made it so that the masks scale instead of reset when the background size changes.

I don't know if people actually use the reset functionality though, so it might be better to do a yes/no/cancel box and make it so that yes is resize, no is reset, and cancel just cancels it.

I've only tested this by changing the background on one room with a walk area mask, so it probably needs more testing.  It seems like it should work fine though.

KeyserSoSay0:
My particular use case was that I kept realizing that one of my scenes was scaled slightly differently than my other scenes, so I just wanted to scale the image down or up, but then I would lose all of my mask data.

You're definitely right that resizing is also a common problem.  I was going to look into that as well before I accidentally created a pull request because I have no idea what I'm doing with git.

Resizing is the more complicated problem though, since you also have to offset the mask in most cases.  Like, if you add 300 pixels on to the left of an image, then you have to move the mask to the right by 300 pixels.  There's no easy way to know which way you have to offset the mask other than having the user put in the offsets during import or having some sort of way to offset masks post-import.  In theory you might also want to both scale and offset the masks.

I'm not sure how easy it is to add checkboxes and input boxes to a dialog, but if they're not a huge pain then I could add a yes/no/cancel dialog box that has a checkbox on it for enabling/disable scaling, an input box for the x offset, and an input box for the y offset (both of which would default to 0).  That way, yes could mean "copy the masks then scale and/or offset them", no could mean "discard the masks", and cancel would just cancel the whole operation.

ivan-mogilko:
I spoke to people on AGS Discord, and the consensus seems that resizing masks without scaling (that is - expanding or cropping the mask canvas, while keeping existing pixels) is the common use case.

In regards to the dialog, I quickly made a dirty mockup of how it could look like:

adjustmaskstoroombg

Making all the above may not be necessary at first, even if there is a dedicated dialog with 3 options that would be enough, and we may add the rest ourselves later.

ericoporto:
If a new dialog is added would be nice if it was using some layout facility otherwise this new dialog will break when scaling to a new size.

KeyserSoSay0:
Hopefully I did the git commands correctly to update this...

I added a dialog for the mask adjustment.  I ended up not adding the placement widget, so everything assumes an offset of the top left.  The widget just seemed like it was more work than it's worth, but if people really want it then it would be easy to add later.  I made it so the dialog doesn't resize so that we don't have to worry about layout changes.

image

I moved the bitmap code to bitmap extensions and made it slightly more generic.

I tested on one room and tried resizing, scaling, and resetting.  I actually haven't used AGS much yet, so it might be good to have some other people test it out.

Things to keep in mind while looking more in depth to the code (just so you can be on the lookout for types of errors): 

1. I hardly ever use C# and I've never actually even made a dialog before.  I'm mostly a C++ programmer.  So make sure I'm not copying a bunch of data that I don't have to or not cleaning up objects or something like that.
2. This is my first change in AGS and I haven't even actually used the tool that much, so definitely let me know if anything is in the wrong place or is likely to cause problems.
3. I really do not know git.  I'm pretty sure I included all the files I actually added/edited but I could have missed something.

ivan-mogilko:
The dialog looks do not comply the rest of AGS UI style, but on itself it is not a big concern, since it may be adjusted or even remade later by one of the developers.

In the above changes I'm mostly concerned about modification to AGS.Types and IRoomController, since that's practically the Editor's public interface (may be used by Editor plugins, etc), so additions should be done with a forethought, and changes to existing functions likely should not be done at all (creating an overloaded variant of a method may be an option).
If it's possible to avoid changing AGS.Types at all here, - such approach may be better, - but I'd need to investigate the situation more closely before I can come to an opinion.
I would also like to ask @persn to check this out if he has got spare time, since he was the one designing IRoomController.

Unfortunately, I've got an unexpected thing and must disappear for several days, but I'll try to review this more closely in the end of this week.

KeyserSoSay0:
Ah I see.  I basically only need the AGS.Types code so that I could pass down the resize info through IRoomController.SetBackground.  I could just pass a bunch of parameters directly though, it just felt wrong to have that many parameters not directly related to the background.

I wonder if it would make sense to break out the mask resizing?  So basically just remove the maskcache reset/resize code from IRoomController.SetBackground, but add a new IRoomController.ResizeMasks.  Then we could call SetBackground followed by ResizeMasks and pass in any parameters directly.  That might still break plugins though if they expect SetBackground to still reset the mask?

I'll wait for persn to chime in since they'll probably have a good idea.

ivan-mogilko:
IRoomController is a relatively new interface, it is unlikely that there are plugins that depend on it already; but my point here rather is that it's better keep the interface as simple as possible.

The idea of separating SetBackground and resizing masks, that you mentioned, sounds reasonable and a step to a right direction.
But I would even go further and suggest to not have an idea of "resizing/rescaling" of anything within the room controller interface at all. I tend to think that we may see IRoomController as a provider of most basic actions, such as "Set background" and "Set mask", while the external user of this interface may decide what to do with the existing masks when the new background is set.

In other words, my proposal would be to try following:
* IRoomController.SetBackground either resets the masks of unmatching size, or does nothing at all (keeps things as-is)
* RoomSettingsEditor does the size check, suggests options (using that dialog that you made), and does the mask rescale, saving results in a temporary masks.
* Then RoomSettingsEditor calls IRoomController.SetBackground and IRoomController.SetMask with all the new modified masks, as necessary.

Editor/AGS.Types/Interfaces/IRoomController.cs
```diff
@@ -37,7 +37,8 @@ public interface IRoomController
         /// </summary>
         /// <param name="background">The background index to set.</param>
         /// <param name="bmp">The image to use for the frame.</param>
-        void SetBackground(int background, Bitmap bmp);
+        /// <param name="maskAdjustParams">The options to use for the masks when the background is changing sizes.</param>
+        void SetBackground(int background, Bitmap bmp, MaskAdjustParameters maskAdjustParams);
```
persn, line 41: If we ARE going to put this in `IRoomController` then it should not be a mandatory param, most people probably aren't that interested in this

```cs
public interface IRoomController
{
     void SetBackground(int background, Bitmap bmp);
     void SetBackground(int background, Bitmap bmp, MaskAdjustParameters maskAdjustParams);
}

public class RoomController : IRoomController
{
     // Map to the "real" method
     void SetBackground(int background, Bitmap bmp) => SetBackground(background, bmp, new MaskAdjustParameters { MaskOption = 0 };

     void SetBackground(int background, Bitmap bmp, MaskAdjustParameters maskAdjustParams)
     {
            // Put real implementation here
     }
}
```

Editor/AGS.Types/MaskAdjustParameters.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace AGS.Types
+{
+	public struct MaskAdjustParameters
+	{
+		public int MaskOption { get; set; }
```
persn, line 8: Magic undocumented numbers? `MaskOption` should be enum with descriptive names

persn (commented):
I can't comment on whether this is a meaningful change or not, however in principle I don't mind it as long as it's not mandatory

persn:
> IRoomController is a relatively new interface

It's not.. just to be clear, I did not invent this interface. It has existed in the source code since before I learned to code. However I did brush the dust off it and expand upon it with functionality that previously only existed in AGS.Native

> AGS.Types and IRoomController, since that's practically the Editor's public interface

If we cannot modify the interfaces we are in a lot of trouble since that is exactly what I did when I added most of the methods that are on there. The idea of incrementing major version from 3 to 4 was that we could break things was it not? To add on this, the SetBackground method that OP is suggested changing/adding is a new method that doesn't even exist in AGS3. For compatibility reasons with plugins this is an OK change to me.

persn:
My interpretation is that the RoomController is just a big hotpot that does all work on Room instances. Controller is an awfully generic word in the first place. It's very highly coupled with the GUI which it ideally shouldn't be, but that was how it already was before I found it. Ideally it should be broken up into smaller sub systems, but it was too much of a headache so I didn't want to deal with it

Editor/AGS.Types/Interfaces/IRoomController.cs
KeyserSoSay0: I ended up breaking out the mask resizing/resetting code from SetBackground.  So SetBackground will be unchanged and RoomSettingsEditor will call ResetMasks or ResizeMasks as needed.

Editor/AGS.Types/MaskAdjustParameters.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace AGS.Types
+{
+	public struct MaskAdjustParameters
```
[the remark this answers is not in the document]
KeyserSoSay0: This got removed since I didn't need to pass it down into SetBitmap anymore

KeyserSoSay0:
Ok, I think I addressed all of the suggestions.  Let me know if I missed anything or if anything else comes up.

ericoporto:
A file is missing here - the mask enum one. Also, I recommend taking a look at the commit messages in the main repository because you are writing things that aren't necessary like the specific file names of each change and the first commit message is way too long. You can rewrite and squash commit messages using git rebase if necessary and I recommend using GitHub Desktop as an easy interface of authoring your commits.

ivan-mogilko:
AdjustMaskOptions.cs file is missing in the latest changes. Also, this file (AdjustMaskOptions.cs) is still mentioned in the AGS.Types project for some reason, although it's now apparently used only in AGS.Editor.

IRoomController.SetBackground still mentions maskAdjustParams parameter in the description, although parameter was removed.

In regards to the commit description: the common convention is to have a short (<= 72 characters) summary as a first line, and then a detailed commentary of any length after couple of linebreaks. This way it is displayed well in majority of git frontends.

KeyserSoSay0:
Whoops, nope that wasn't it...

KeyserSoSay0:
Oh wait, maybe that was correct?  Comparing the ags4 branch to this change looks right.  I guess maybe I was just comparing the original branch with this change or something because I saw 1000+ changes.

Thanks for the help @ivan-mogilko, GitHub Desktop made it WAY clearer what was going on.  It turns out that when I ran the old CMake project gen stuff a while ago it added a TON of files that weren't in .gitignore.  I was thinking it was totally normal for git to just have hundreds of untracked files, but the GUI made it more obvious that that wasn't the case.

So here's to hoping that I successfully rebased on top of the ags4 branch and squashed all the commits into one CL description that isn't insane anymore.  I'll give it a 50% chance at least.

Editor/AGS.Editor/Components/RoomsComponent.cs
```diff
@@ -2187,21 +2230,6 @@ private void ClearMaskCache()
             }
         }
 
-        private void ResetMaskCache()
```
KeyserSoSay0, line 2190: This became IRoomController.ResizeMasks

Editor/AGS.Editor/Panes/Room/RoomSettingsEditor.cs
```diff
@@ -566,19 +569,24 @@ private void ImportBackground(int bgIndex)
                             }
                             else
                             {
-                                roomSizeChanged = true;
```
KeyserSoSay0, line 569: I think this shouldn't have been here because farther down the doImport could be canceled by the dialog box and the room size wouldn't change.  It wouldn't have mattered in practice though since roomSizeChanged only gets used if doImport is true.

ivan-mogilko:
I tested the functionality, and it works fine.
The interesting thing is that you can also crop masks with negative offset, moving central part of the larger mask to the top-left corner of a smaller mask (not sure if there's a practical use case, but wanted to note that this is possible).

In regards to the code, I'd still move AdjustMaskOptions from AGS.Types to AGS.Editor project, as it's only used there.
Also, the dialog's looks need to be readjusted to match the rest of Editor dialogs, but that is something that one of the developers can do after this PR is merged.

@persn , would you like to comment anything else?

persn:
I concur that the `AdjustMaskOptions` should be moved out of AGS.Types since it's not used there

Everything else is fine I think

ivan-mogilko:
Okay, I'll merge this and do a small fix myself, that will be quicker than keeping updating PR.

PR #2470, opened 2024-07-12
Pass mouse position into "on_mouse_click" script callback

### LINKED ISSUE #2468: Expand on_mouse_click callback parameters with click position
I've kept forgetting to write this down, but here it is. In v3.6.0 we expanded `on_key_press` with `mod` parameter, and made it called for each key press. In v3.6.1 (iirc) `on_mouse_click` is called actually as many times as a mouse was clicked between game frames.

However, there's still something missing: mouse position is not recorded for the time of click and `mouse.x/y` properties will only show mouse position at the time of the callback is run, but not at the time of actual events. Which may lead to random mistakes when player moves mouse quickly across the screen while clicking, or when the mouse "twitches" while clicking.

Therefore the task is to define `on_mouse_click` as following:

```
function on_mouse_click(MouseButton button, int mx, int my)
```

Because of the changes done to the engine when expanding on_key_press, this will be fully backwards compatible, as if callback's parameters are missing (not declared) in older script, engine will simply skip these.

It should be easy to get the click's coordinate from SDL mouse event, and save along with the button id. There may be a question whether have these saved in real or native game coords, but in the end they have to be converted to game coords before passing to script.

**UPDATE**

Also will be necessary to expand `dialog_options_mouse_click` for custom dialog options:

```
function dialog_options_mouse_click(DialogOptionsRenderingInfo *info, MouseButton button, int mx, int my)
```

**UPDATE 2**

Also must investigate `on_event` with `eEventGUIMouseDown`/`eEventGUIMouseUp`. Unfortunately, GUI processing is currently written in not a good way, and may ignore multiple button down/ups during game frame.

ivan-mogilko:
Resolves #2468

1. Refactored "game events" code inside the engine. This is the first big commit, details may be found in commit's description. Functionality must remain the same, but the event parameters are stored more explicitly now.
2. Added coordinate parameters to "on_mouse_click" callback, now it should be declared as:
```
function on_mouse_click(MouseButton button, int mx, int my)
```
3. Added similar parameters to "dialog_options_mouse_click", now it should be declared as:
```
function dialog_options_mouse_click(DialogOptionsRenderingInfo *info, MouseButton button, int mx, int my)
```
These changes are fully backwards compatible, as the engine was previously taught to skip trailing parameters if the script has declared these functions with less number of arguments.

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -140,7 +140,7 @@ void Character_AddInventory(CharacterInfo *chaa, ScriptInvItem *invi, int addInd
             if (charextra[charid].invorder[ee] == inum) {
                 // They already have the item, so don't add it to the list
                 if (chaa == playerchar)
-                    run_on_event (GE_ADD_INV, RuntimeScriptValue().SetInt32(inum));
+                    run_on_event (kScriptEvent_InventoryAdd, RuntimeScriptValue().SetInt32(inum));
                 return;
             }
         }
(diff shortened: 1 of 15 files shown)


ericoporto:
The refactor changes are good and the other changes appear alright too. Later, is it needed to include position for the OnClick GUI/Button events, created by the editor, too?

ivan-mogilko:
> Later, is it needed to include position for the OnClick GUI/Button events, created by the editor, too?

I don't think so, the OnClick event are corresponding to the sole fact of control being interacted, not where (this corresponds to interaction events of characters and others).

We might need a separate event for clicking on a gui layer, or merge this in "on_mouse_click".

PR #2476, opened 2024-07-20
Editor: add a template save command line option

ericoporto:
I am trying to figure a command line option to build a template in the editor from the command line, so it can be ran in a CI environment.

The idea would it be using it for automation purposes, like https://github.com/adventuregamestudio/ags-template-source/issues/27 .

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -933,6 +933,26 @@ namespace AGS.Editor
             this.ExitApplication();
         }
 
+        public void SaveAsTemplateAndExit(string projectPath)
+        {
+            bool error = true;
+            if (_interactiveTasks.LoadGameFromDisk(projectPath))
+            {
+                error = false;
+                bool forceRebuild = _agsEditor.NeedsRebuildForDebugMode();
+                _agsEditor.CompileGame(forceRebuild, false);
+                if (forceRebuild)
+                    _agsEditor.SaveUserDataFile(); // in case pending config is applied
+
+                _batchProcessShutdown = true;
+
+                error = !SaveGameAsTemplate(_agsEditor.CurrentGame.Settings.GameFileName + ".agt");
+            }
+            if (error) Program.SetExitCode(1);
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Is there a reason why not to share same chunk of code found in `GUIController.SaveGameAsTemplate`?
https://github.com/adventuregamestudio/ags/blob/ad31da24686e6f78360a8a6bc827b5b914d4f089/Editor/AGS.Editor/GUI/GUIController.cs#L1177-L1189
i.e. pick that out to a private method and call from both, passing a filename as an argument.

ivan-mogilko:
Maybe call command `/maketemplate`, for clarity, and in case there will be another operation involving template at some point?

ericoporto:
Yes, these changes are easy to do, I can try they.

What's your thoughts on rebuilding the game and saving here? I added those to make sure room.crm is updated, but Empty Game template for example has no room and will error when building (the error is ignored here). The other approach is too strictly only run the SaveGameAsTemplate and not recompile, and then the user would usually do both in a sequence in a CI.

ivan-mogilko:
Are there situations when the project is modified on load without user's interference?
- when an older project gets upgraded;
- when some errors are fixed automatically (such as disconnected sprites in a sprite file).

Upgrading a project may theoretically involve changing file structure: the easiest example is ags3 -> ags4 upgrade, where room contents are unpacked into the subfolders. If a project file is not saved afterwards, then there may be inconsistencies between project version and file structure. This makes me think that saving the project is mandatory here.

As for compilation, I am not certain. Maybe it's useful as a test that the game is valid...

ericoporto:
I left it exactly as it was - including it recompiling and saving. But I renamed the flag to `maketemplate` and did the small refactor on GUI Controller to move the template saving code there.

The name of the template is set automatically as `_agsEditor.CurrentGame.Settings.GameFileName + ".agt"` , which will cause it to be generated in the current working dir, which I think is always the game directory.

Another approach is to take two command line arguments so that it can receive a second parameter that is a fullpath of the target name of the template.

PR #2478, opened 2024-07-25
AGS 4: Script API: add Overlay.Tint and SetLevel, complying to character and object settings

ivan-mogilko:
Support for tinting overlay, for consistency with characters and objects.

Added following to the Overlay struct in script:
```
  /// Tints the overlay to the specified colour.
  import void     Tint(int red, int green, int blue, int saturation, int luminance);
  /// Sets the light level for this overlay.
  import void     SetLightLevel(int light_level);
  /// Removes an existing colour tint or light level from this overlay.
  import void     RemoveTint();
  /// Gets whether the overlay has a tint set.
  readonly import attribute bool HasTint;
  /// Gets whether the overlay has a light level set.
  readonly import attribute bool HasLightLevel;
  /// Gets the individual light level for this character.
  readonly import attribute int  LightLevel;
  /// Gets the Blue component of this overlay's colour tint.
  readonly import attribute int  TintBlue;
  /// Gets the Green component of this overlay's colour tint.
  readonly import attribute int  TintGreen;
  /// Gets the Red component of this overlay's colour tint.
  readonly import attribute int  TintRed;
  /// Gets the Saturation of this overlay's colour tint.
  readonly import attribute int  TintSaturation;
  /// Gets the Luminance of this overlay's colour tint.
  readonly import attribute int  TintLuminance;
```

### DIFF
--- a/Common/game/roomstruct.cpp
+++ b/Common/game/roomstruct.cpp
@@ -220,7 +220,7 @@ int RoomStruct::GetRegionLightLevel(int id) const
 int RoomStruct::GetRegionTintLuminance(int id) const
 {
     if (id >= 0 && id < MAX_ROOM_REGIONS)
-        return HasRegionTint(id) ? (Regions[id].Light * 10) / 25 : 0;
+        return HasRegionTint(id) ? GfxDef::Value250ToValue100(Regions[id].Light) : 0;
     return 0;
 }
 
--- a/Common/gfx/gfx_def.h
+++ b/Common/gfx/gfx_def.h
@@ -124,6 +124,20 @@ private:
 
 namespace GfxDef
 {
+    // Converts value from range of 100 to range of 250 (sic!);
+    // uses formula that reduces precision loss and supports flawless forth &
+    // reverse conversion for multiplies of 10%
+    inline int Value100ToValue250(int value100)
+    {
+        return (value100 * 25) / 10;
+    }
+
+    // Converts value from range of 250 to range of 100
+    inline int Value250ToValue100(int value100)
+    {
+        return (value100 * 10) / 25;
+    }
+
     // Converts percentage of transparency into alpha
     inline int Trans100ToAlpha255(int transparency)
     {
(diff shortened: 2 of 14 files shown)


ericoporto:
What is LightLevel and why it isn't a property that can be set too?

ivan-mogilko:
> What is LightLevel and why it isn't a property that can be set too?

Light level brightens or darkens sprite. It may be set using function SetLightLevel(), and read using LightLevel property. This is done for consistency with characters and objects, and also to make it clear that it is mutually exclusive with the Tint. Either Tint or LightLevel may be set at the same time, so calling function resets the other setting. With settable properties that may be less obvious.

ivan-mogilko:
Oops, I just noticed that SetLightLevel is declared as "function" instead of "void". Probably a result of copy-paste.

ericoporto:
Is it possible to add the ranges right in the comments too? Like does it goes to 0-100 or 0-255 or 0-250 ...?

ivan-mogilko:
> Is it possible to add the ranges right in the comments too? Like does it goes to 0-100 or 0-255 or 0-250 ...?

Yes, I shall add these.

ericoporto:
@ivan-mogilko you are hitting a weird error below according to the CI
```
libsrc/allegro/src/file.c:508:18: error: passing argument 1 of 'utf8_getx' from incompatible pointer type [-Wincompatible-pointer]
```

ivan-mogilko:
That is not a problem of this PR, this happens in ags4 branch.

ivan-mogilko:
> I think if I could in a single API call set an overlay position and size too, it would be great! (like `Overlay.SetDimensions(int x, int y, int width, int height)`)

That will be fine to add; although i am not certain about the name. "Dimensions" term is never used in ags api; the only instance of a function that sets all 4 values is "Viewport.SetPosition" (probably because I found it easier to remember), and the equivalent that i found in C# is "Control.SetBounds".

ericoporto:
Other possible naming would be something like Geometry. I know it's not related to this PR it's just that looking at SetTint it would be awesome for performance that does more of setting the overlay in a single call for uses of overlays in things like a particle or in this alternative renderer of mine.

ivan-mogilko:
Added SetPosition and SetSize in a separate PR:  #2483

ivan-mogilko:
Rebased to be after mingw fixes.

PR #2479, opened 2024-07-27
AGS 4: add custom properties to AudioClips, Dialogs, GUI, Regions and Walkable Areas

ivan-mogilko:
This adds custom properties functionality to AudioClips, Dialogs, GUI, Regions and Walkable Areas.

For the most part this is a simple change, although there's lots of code duplication, as unfortunately the script types in AGS still don't have a proper base class.

**NOTE:** AudioClips are not writing their custom properties to saves. They are considered a "game resource", and not part of the game's state, and are kind of assets that may be freely added and removed in the game project, similar to sprites.
This may be reviewed later, when we find good use cases for their custom properties, and it will become apparent that having them in saves is more convenient than reinitializing them in script after restoring a save.
Although hopefully this won't be necessary. Personally, I'd rather wish we reduce number of things written to game saves. E.g. don't save other resources, such as dynamic sprites.
If we must, eventually, then audio clip's runtime changes should not saved by "index", but use a script name for identification and matching, so that any change to a list of audio clips in game won't break older saves.

### DIFF
--- a/Common/ac/dialogtopic.cpp
+++ b/Common/ac/dialogtopic.cpp
@@ -29,7 +29,7 @@ void DialogTopic::ReadFromFile(Stream *in)
     topicFlags = in->ReadInt32();
 }
 
-void DialogTopic::ReadFromSavegame(Common::Stream *in)
+void DialogTopic::ReadFromSavegame(Common::Stream *in, DialogSvgVersion /*svg_ver*/)
 {
     in->ReadArrayOfInt32(optionflags, MAXTOPICOPTIONS);
 }
--- a/Common/ac/dialogtopic.h
+++ b/Common/ac/dialogtopic.h
@@ -29,6 +29,16 @@ using namespace AGS; // FIXME later
 #define DTFLG_SHOWPARSER    1  // show parser in this topic
 #define DCHAR_NARRATOR  999
 #define DCHAR_PLAYER    998
+
+// RoomStatus runtime save format
+// TODO: move to the engine code
+enum DialogSvgVersion
+{
+    kDialogSvgVersion_Initial  = 0,
+    kDialogSvgVersion_40008    = 4000008, // custom properties
+};
+
+
 struct DialogTopic {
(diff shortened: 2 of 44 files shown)


ivan-mogilko:
I wanted to spend some time, thinking on how audio clips are writing their custom properties in saves.
But then I came to a decision to not write them into saves at all.
I think we might want to write less things in saves in the future.
Of course if this is absolutely necessary, then they may be added to the save components.
Left a note in the PR's description about this (and a small comment in code).

ericoporto:
I am all in for writing less things in saves but I wouldn't mind having something for ini files or even json or xml files right in script api so handling writing your own text data would be easier in script.

ivan-mogilko:
> Dynamic sprites having some way to not write them would be useful too. Maybe they would need some way to be tagged so their bitmap data could be recreated - not sure on this.

This is not related to the PR, but since I mentioned this myself, I suppose that we may exclude bitmaps from save states completely (DynamicSprite struct must still be saved, as it's a part of a managed memory, and pointers must remain valid, but it may be "invalidated").
The sprites may be recreated after restoring a save, in script. If it's more convenient to have them saved and loaded, they may be written to files in save folder, and read back from them. I suppose that this is going to be approximate in speed and disk space requirement to having them inside a save state.

ivan-mogilko:
Hmm, but what of this PR itself? I think it's pretty trivial, so unless there's an obvious issue, i'd just merge it.

ericoporto:
I am trying to understand how the new v400 custom property extension works. In the original custom properties there was a default value and only properties different than default value were stored on saves/gamedata or were they all always stored?

ivan-mogilko:
New properties are supposed to work exactly as existing properties work. There's a dictionaries with design-time values found in GameSetupStruct, and dictionaries with runtime set values found in GamePlayState struct. Only latter are saved.

ivan-mogilko:
Had to fix Room Regions and Walkable Areas, it appeared they were not saved into compiled rooms.

PR #2482, opened 2024-07-28
Editor: error preprocessor in invalid script characters

### LINKED ISSUE #2480: Preprocessor gets infinite loop meeting a non-latin unicode char in script
**Problem**
If the game project is configured into Unicode mode, and there's a non-latin character met in plain script (except comments or character string literals in single or double quotes respectively), then the script preprocessor would stuck in an infinite loop while trying to process the scripts.

**Steps to reproduce**
1. Create a project in Unicode mode.
2. Open a script and place a random non-latin char on a new line.
3. Run game build.
4. Observe that the process is stuck at the script compilation forever.

**Expected behavior**
I suppose it's expected that preprocessor should produce a warning or error about invalid character found in script.

**Additional**
That would be a good idea to also test compilers with the same case, as both old and new one were written back when scripts were strictly ASCII.

On a side note, we might plan adding a "Abort/Cancel" button to a BusyDialog that runs a working process. It's weird how this was not featured from the start.

ericoporto:
possible fix #2480 

It would certainly be nice to have more tests to add here.

### DIFF
--- a/Editor/AGS.CScript.Compiler/Preprocessor.cs
+++ b/Editor/AGS.CScript.Compiler/Preprocessor.cs
@@ -165,6 +165,15 @@ namespace AGS.CScript.Compiler
 			{
 				i++;
 			}
+
+			if (i < text.Length && text[i] > 127)
+			{
+				RecordError(ErrorCode.InvalidCharacter, "Invalid character detected in script at position " + i.ToString() + " in '" + text + "'");
+				string res = text;
+				text = string.Empty; // need to end line
+				return res;
+			}
+
 			string word = text.Substring(0, i);
 			text = text.Substring(i);
 			if (trimText)
--- a/Editor/AGS.CScript.Compiler/Entities/ErrorCode.cs
+++ b/Editor/AGS.CScript.Compiler/Entities/ErrorCode.cs
@@ -46,6 +46,7 @@ namespace AGS.CScript.Compiler
         MemberFunctionNotDefined,
         ParentIsNotAStruct,
         InvalidUseOfStruct,
-        UnterminatedString
+        UnterminatedString,
+        InvalidCharacter
 	}
 }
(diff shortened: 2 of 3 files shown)


Editor/AGS.CScript.Compiler/Preprocessor.cs
```diff
@@ -158,19 +170,32 @@ private string GetNextWord(ref string text)
 		private string GetNextWord(ref string text, bool trimText, bool includeDots)
 		{
 			int i = 0;
-			while ((i < text.Length) && 
-				   (text[i].IsScriptWordChar() ||
-				    (includeDots && (text[i] == '.')))
-				   )
+			while (i < text.Length)
 			{
+				if (char.IsWhiteSpace(text[i]) || text[i] == ';' || text[i] == ',' || text[i] == '(' || text[i] == ')' || (!includeDots && text[i] == '.'))
```
ivan-mogilko, line 175: I don't think that's a fully correct inversion of the previous condition: this does not include other symbols that may be met in script, such as `+`, `-`, `*`, `/`, square brackets, figure brackets, and anything else I forgot about.

Would not it be possible to keep the previous condition, but also detect invalid chars right away?

ivan-mogilko: It looks like the following change would be enough, if added right after the original loop:

```
            if (i < text.Length && text[i] > 127)
            {
                RecordError(ErrorCode.InvalidCharacter, "Invalid character detected in script at position " + i.ToString() + " in '" + text + "'");
                string res = text;
                text = string.Empty; // need to end line
                return res;
            }
```

ericoporto: Thank you! I think that would work, yes. I won't be back to my computer until a few days (using my phone), if you want to do this change directly feel free.

If you want to modify this branch too with this change I think you can do it either as a moderator of ags or alternatively I added you as a collaborator in my fork.

At a later time I plan to contribute some additional (more complete) tests.

ericoporto:
Found a little time and managed to fix this!

Editor/AGS.CScript.Compiler/Preprocessor.cs
ericoporto: I found time and implemented your change and it works fine!

PR #2483, opened 2024-07-28
Script API: added helper functions Overlay.SetPosition and SetSize

ivan-mogilko:
This is my proposal:

```
/// Sets this overlay's position, and optionally its size.
import void SetPosition(int x, int y, int width = 0, int height = 0);
/// Changes the size of the overlay.
import void SetSize(int width, int height);
```

In SetPosition() width and height are optional parameters, if they are not set (or rather <= 0), then the old size is kept.
Having them combined in SetPosition makes 1 function less, and also we might expand existing SetPosition for other objects with optional width and height too.
But if someone thinks that this is not a good idea, i may reduce SetPosition to strictly x,y and make a third function that has 4 parameters.


ericoporto:
This looks good! I won't be able to test it for a few days I think.

PR #2484, opened 2024-07-30
AGS 4: Script API: add global events for Dialogs, and properties to read current dialog state

### LINKED ISSUE #1328: Script: add generic events for dialog start/end
A suggestion is to add two new events for the `on_event` callback:
1) One run just before a dialog starts;
2) Second run just after a dialog ends;

Both events' parameter defines dialog's numeric ID.

If possible to implement, this will provide a simple and convenient opportunity to handle any changes when entering/exiting a dialog state in game.

ivan-mogilko:
Resolves #1328

This adds following EventTypes for use in "on_event" callback:

```
eEventDialogStart.
eEventDialogStop,
eEventDialogRun,
eEventDialogOptionsOpen,
eEventDialogOptionsClose,
```

* eEventDialogStart - run when the dialog state begins; arg1 = dialog number;
* eEventDialogStop - run when the dialog state ends; arg1 = last dialog number;
* eEventDialogRun - run when any dialog's entry is executed, whether starting entry or one of the options; arg1 = dialog number, arg2 = option number (0 = starting entry);
* eEventDialogOptionsOpen - run right before the dialog options are shown on screen; arg1 = dialog number
* eEventDialogOptionsClose - run right after the dialog options are removed from the screen; arg1 = dialog number, arg2 = chosen option number

Additionally, added following properties to a Dialog struct:
```
/// Gets the currently running dialog, returns null if no dialog is run
static readonly attribute Dialog* CurrentDialog;
/// Gets the currently executed dialog option, or -1 if none is
static readonly attribute int ExecutedOption;
/// Gets if the dialog options are currently displayed on screen
static readonly attribute bool AreOptionsDisplayed;
```

These may be accessible in dialog script, any functions called from dialog script, custom dialog option callbacks, and "repeatedly_execute_always" either when there's a waiting inside a dialog script or during options and "run game loops while dialog options" is enabled.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1812,6 +1812,14 @@ builtin managed struct Dialog {
   /// Gets the script name of this dialog.
   import readonly attribute String ScriptName;
 #endif
+#ifdef SCRIPT_API_v362
+  /// Gets the currently running dialog, returns null if no dialog is run
+  import static readonly attribute Dialog* CurrentDialog; // $AUTOCOMPLETESTATICONLY$
+  /// Gets the currently executed dialog option, or -1 if none is
+  import static readonly attribute int ExecutedOption; // $AUTOCOMPLETESTATICONLY$
+  /// Gets if the dialog options are currently displayed on screen
+  import static readonly attribute bool AreOptionsDisplayed; // $AUTOCOMPLETESTATICONLY$
+#endif
 
   readonly int reserved[2];   // $AUTOCOMPLETEIGNORE$
 };
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1327,6 +1327,13 @@ enum EventType {
   eEventLeaveRoomAfterFadeout = 11,
   eEventGameSaved = 12,
 #endif
+#ifdef SCRIPT_API_v400
+  eEventDialogStart = 13,
+  eEventDialogStop = 14,
+  eEventDialogRun = 15,
+  eEventDialogOptionsOpen = 16,
+  eEventDialogOptionsClose = 17,
+#endif
 };
 
 #ifdef SCRIPT_API_v350
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -54,6 +54,7 @@
 
 using namespace AGS::Common;
 using namespace AGS::Engine;
+struct DialogExec;
 
 extern GameSetupStruct game;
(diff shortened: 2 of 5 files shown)


ivan-mogilko:
Made a simple demo game, only to illustrate the potential use case:
[ags4--demo-dialogevents.zip](https://github.com/user-attachments/files/16429277/ags4--demo-dialogevents.zip)
Compiled game:
https://www.dropbox.com/scl/fi/jzotaao8pdzcjtehehvke/ags4-demo-dialogevents-win.zip?rlkey=kdc9qzpliaz7jqtf0dy5u35on&st=b6wcw1eq&dl=0

ericoporto:
I think this is meant to address #1328 but it’s not clear how the issue described in comments (on blocking) is solved.

ivan-mogilko:
I forgot about #1328. My first attempt was a failure, but I cannot remember what I did wrong, because it works correctly now.

ericoporto:
Regarding api naming, I think instead of `Dialog.OptionsShown` naming it as `Dialog.OptionsVisible` or `Dialog.OptionsEnabled` would sound more like a boolean because it would sound more like the other APIs in ags that are boolean. For me `OptionsShown` sound like a number, list or range (like"what options are shown?" or something).

ivan-mogilko:
My understanding is that the difference between words "Shown" and "Visible" is that "Shown" implies a past change of state (a result of action), and "Visible" is simply a description of the current state. If OptionsShown may be understood as "number of / which options shown", then so OptionsVisible may  be understood similarly as "number of / which options visible".

In properties that describe to the current object's state we usually leave the "Is" out (this.Visible vs IsVisible, this.Enabled vs IsEnabled, and so forth). But I cannot tell whether same would work linguistically for properties referring to a another object connected to the current one.

I can add "Are" for clarity.

Another thing, there's Dialog.DisplayOptions, which uses term "to display" for options. "AreOptionsDisplayed" may be therefore a more consistent variant.

ericoporto:
Hey, this change I think sounds better! I think the attribute registration in the engine script API still is using the old string that needs to also be updated.

ericoporto:
All looking good!

I see when running at first there's a call that is as follows

```
run_on_event(kScriptEvent_DialogRun, RuntimeScriptValue().SetInt32(dlgnum));
```

In this case no value for "arg2 = option number" is explicitly set, just to confirm, this would cause AGS to pass 0 which is equal to the starting option, correct?

I can't test this now but the code, logic and names all looking great.

ivan-mogilko:
> In this case no value for "arg2 = option number" is explicitly set, just to confirm, this would cause AGS to pass 0 which is equal to the starting option, correct?

It is; but in this case I should have passed an explicit parameter.

PR #2490, opened 2024-08-07
Editor: Fix room editor flood fill

KeyserSoSay0:
-It looks like images can have a different _paddedWidth that needs to be taken into account when calculating where in the byte array we are.  The flood fill algorithm was using the normal width instead of the padded width, so it would basically always flood the entire screen since the pixels would be off.

-Also, added a minor optimization.  In our inner loop were were checking image[i] == initial && image[i] != replacement.  The only time you could ever pass the initial check but fail the replacement check is if initial == replacement.  In that case you're just doing if (a == b && a != b), so we can just skip the entire function for cases where the initial and replacement values are the same.

### DIFF
--- a/Editor/AGS.Editor/Utils/BitmapExtensions.cs
+++ b/Editor/AGS.Editor/Utils/BitmapExtensions.cs
@@ -623,7 +623,7 @@ namespace AGS.Editor
         public void FillArea(int color, Point position, double scale)
         {
             Point positionScaled = new Point((int)(position.X * scale), (int)(position.Y * scale));
-            FloodFillImage(_pixels, positionScaled, _bmp.Size, _pixels[(positionScaled.Y * _bmp.Width) + positionScaled.X], (byte)color);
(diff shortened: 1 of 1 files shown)


KeyserSoSay0:
I think I didn't completely mess up the pull request this time?  Fingers crossed for slightly figuring out git.

ericoporto:
Can you give some steps to reproduce the bug you are reporting?

KeyserSoSay0:
It looks like my image is 358x205.  Here's the actual background I'm using: 
starshipbridgeSm

Here's the walkbehinds I was editing when I noticed it (but it's probably easier to repro by just making your own lines)
walkbehinds

ivan-mogilko:
I confirm the error, and that the fix works. The code changes make perfect sense to me too.

While testing the issue I was expecting something like a attempt to access a byte outside array, but it never occured. What happened instead was that the fill operation was cut at random diagonal lines. Here's a screenshot for the reference:
ags4-floodfill-error

PR #2496, opened 2024-08-13
AGS 4: disable 16-bit games, and convert them to 32-bit on import

### LINKED ISSUE #2493: AGS 4: remove 16-bit game color depth
I've been looking at #1980 again, and I think we might remove 16-bit game support first, thus reducing number of concerns in #1980.

For the new projects it is a matter of removing "16-bit" choice in the general settings. The value itself may have to stay in GameColorDepth enum to avoid errors when importing an older project, but tagged with `[Browsable(false)]` and `[Obsolete]` attributes, which will make it not present in the list.

I mostly wonder about older project import here, in case they are in 16-bit. AFAIK AGS does not have any actual conversion mechanism when switching the color depth in general settings. In theory, following has to be done:
- All sprites reimported, converting them to new final color depth. Normally this is done by running "recreate spritefile" task, but this will only work if their sources are intact. If they are not, then this will have to be done other way. Perhaps it's a matter of writing an alternate process that reads sprites from old spritefile, converts color depth, and saves into a new one.
- Rooms: since there's a room conversion (from compiled to open format) in place already, it may be amended with saving background PNGs in the specified format. (BTW, I never tried 16-bit in ags4, so I don't know if new rooms even work in it)
- Updating "color" properties: they are already in 16-bit as we know, so upgrading these will be a part of #1980.

For the engine... I don't want to remove 16-bit bitmap support from the graphics library just yet, after all we might want to support loading an arbitrary image for sprites with conversion.
But any algorithm in the engine itself, which has a branch for 16-bit color, may be cut out.

---

On testing, there is a number of old 16-bit games in this list of opensource games:
https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/a-sticky-list-with-open-source-ags-games/

Quick example: "Aeronuts" (http://shatten.sonores.de/wp-content/uploads/2014/09/AeroNuts_OSS.7z)

ivan-mogilko:
Resolves #2493

* Marks "16-bit" Game ColorDepth's option as disabled;
* When loading a project test whether Game's ColorDepth is 16-bit, and if it is, then warn the user and update to 32-bit.
* Recreate spritefile, automatically converting every sprite of previous game's color depth to a new one (16->32 in this case).

Rooms are converted from CRM to open format after the game setting is updated, and save backgrounds in 32-bit, so that seems to be covered.

Remaining tasks:
- [x] - Investigate the editor and cut any code that depends on "Hi-color" setting, if there's anything.
- [x] - Investigate the engine and cut any code branch that depends on having 16-bit game setting or 16-bit sprites. Possible exceptions are graphic library and any utility functions (like converting a bitmap to a texture).


ericoporto:
The `InitSpritesAfterGameLoad` that was added and refactored in the Editor LoadFromDisk function looks good, I think that part specifically should be backported to master. (It's in the commit https://github.com/adventuregamestudio/ags/pull/2496/commits/efb8b27e22d902aa6df8e855410c785f15bce0fe, which is doing other things too, so not the other stuff).

ivan-mogilko:
Reorganized commits a little, but the changes are the same.

I used "AeroNuts" game to test 16-bit game conversion (sprites + rooms):
http://shatten.sonores.de/wp-content/uploads/2014/09/AeroNuts_OSS.7z
although one must import this game to 3.6.1 first and do couple of script fixes before importing further into AGS 4.

PR #2497, opened 2024-08-14
Editor: fix case recent games menu error on null path

ericoporto:
Not sure, but possible fix for https://www.adventuregamestudio.co.uk/forums/ags-engine-editor-releases/ags-4-0-early-alpha-for-public-test/msg636665024/#msg636665024

Edit:

OK, to cause the error it's not that the path tag is empty but that it doesn't exist at all. I had only tested it for it being empty, which instead makes it be an empty string instead of null.


ericoporto:
I put it on master but this should also go to ags4 branch.

PR #2498, opened 2024-08-14
Engine: IndexOf add string index, count and case sensitivity params

### LINKED ISSUE #2487: Script API: add "startIndex" parameter to String.IndexOf
Reasoning: String.IndexOf will only find the first instance of a "needle". If you like to keep finding further instances, you must create substrings of original string, which slows things down.

Proposal therefore is to add a "startIndex" to IndexOf, which lets to keep searching for the "needle" after the first instance.

EDIT: amended with "endIndex" and "caseSensitive" from proposals below.

```
int String.IndexOf(const string needle, bool caseSensitive = true, int startIndex = 0, int endIndex = -1);
```

Example of a use case:
```
int found_at = -1;
int found_count;
do
{
    found_at = input_str.IndexOf(find_what, true, found_at + 1);
    if (found_at >= 0)
        found_count++;
}
while (found_at >= 0);
```

ericoporto:
fix #2487

- I don't remember how to deal with the previous script API being only 1 parameter in ags4 branch! (Should I also link to  `String::IndexOf^1` and then have both be able to set through ifdefs in the header? Or is this not needed in ags4 ? the defaults will keep them working but rebuild is still needed.
- I am not convinced that the final code is actually getting proper offset in unicode characters, even as is in the current branch, did we test this at some point?
- I stole bits of the substring function to make the substring that the IndexOf will apply to. I wasn't sure if I should error here too or just throw a warn and return -1.

### DIFF
--- a/Engine/ac/string.cpp
+++ b/Engine/ac/string.cpp
@@ -27,6 +27,7 @@
 #include "debug/debug_log.h"
 #include "script/runtimescriptvalue.h"
 #include "util/string_compat.h"
+#include "util/string_utils.h"
 #include "util/utf8.h"
 #include "ac/dynobj/cc_dynamicarray.h"
 
@@ -427,28 +428,49 @@ int StringToInt(const char*stino) {
     return atoi(stino);
 }
 
-int StrContains (const char *s1, const char *s2) {
-    VALIDATE_STRING (s1);
-    VALIDATE_STRING (s2);
-    char *tempbuf1 = ags_strdup(s1);
-    char *tempbuf2 = ags_strdup(s2);
-    ustrlwr(tempbuf1);
-    ustrlwr(tempbuf2);
+int String_IndexOf (const char *thisString, const char *lookForText, int startIndex, int count, bool caseSensitive) {
+    VALIDATE_STRING(thisString);
+    VALIDATE_STRING(lookForText);
 
-    char *offs = ustrstr(tempbuf1, tempbuf2);
-
-    if (offs == nullptr)
-    {
-        free(tempbuf1);
-        free(tempbuf2);
+    const auto &header = ScriptString::GetHeader(thisString);
+    if ((startIndex < 0) || ((size_t) startIndex > header.ULength)) {
+        debug_script_warn("String.IndexOf: invalid start index %d.", startIndex);
         return -1;
     }
 
+    if (count == 0 || count >= header.ULength - startIndex)
+        count = header.ULength - startIndex;
+
+    if (count < 0) {
+        debug_script_warn("String.IndexOf: invalid count %d. Did you meant to use 0 for until string end?", count);
+        return -1;
+    }
+
+    size_t start = uoffset(thisString, startIndex);
+    size_t end = uoffset(thisString, startIndex + count);
+
+    auto thistmpbuf = StrUtil::Substring(thisString, start, end - start);
+    auto looktmpbuf = StrUtil::Duplicate(lookForText);
+
+    if (!caseSensitive)
+    {
+        ustrlwr(thistmpbuf.get());
+        ustrlwr(looktmpbuf.get());
+    }
+
+    char *offs = ustrstr(thistmpbuf.get(), looktmpbuf.get());
+
+    // return -1 if it can't find it
+    if (offs == nullptr)
+        return -1;
+
     *offs = 0;
-    int at = ustrlen(tempbuf1);
-    free(tempbuf1);
-    free(tempbuf2);
+    int at = startIndex + ustrlen(thistmpbuf.get());
     return at;
+}
+
+int StrContains (const char *s1, const char *s2) {
+    return String_IndexOf(s1, s2, 0, 0, false);
 }
 
 int String_GetLength(const char *thisString) {
@@ -552,6 +574,12 @@ RuntimeScriptValue Sc_String_CompareTo(void *self, const RuntimeScriptValue *par
 RuntimeScriptValue Sc_StrContains(void *self, const RuntimeScriptValue *params, int32_t param_count)
 {
     API_OBJCALL_INT_POBJ(const char, StrContains, const char);
+}
+
+// int  (const char *s1, const char *s2)
+RuntimeScriptValue Sc_String_IndexOf(void *self, const RuntimeScriptValue *params, int32_t param_count)
+{
+    API_OBJCALL_INT_POBJ_PINT2_PBOOL(const char, String_IndexOf, const char);
 }
 
 // const char* (const char *srcString)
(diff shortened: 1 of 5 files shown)


ivan-mogilko:
>  (Should I also link to String::IndexOf^1 and then have both be able to set through ifdefs in the header? Or is this not needed in ags4

I am not sure myself, but if need arises then this will be trivial to add another table entry.

Engine/ac/string.cpp
```diff
@@ -427,30 +427,55 @@ int StringToInt(const char*stino) {
...
+    if (endIndex == -1 || endIndex >= header.ULength)
+        endIndex = header.ULength;
 
-    if (offs == nullptr)
+    size_t start = uoffset(thisString, startIndex);
+    size_t end = uoffset(thisString, endIndex);
+    size_t byte_len = end - start;
+    size_t u_len = endIndex - startIndex;
+
+    auto buf = ScriptString::CreateBuffer(byte_len, u_len);
```
ivan-mogilko, line 451: There's virtually no reason to use ScriptString::CreateBuffer here, this function is only ever needed when you need to create and return a new ScriptString object. This function creates a buffer with a meta header prepended to it.
Here you just run a standard string algorithm over a pair of buffers, so ags_strdup is correct.

ivan-mogilko: Alright, I see, you need a substring.
In such case you could simply use a common String, which has a constructor accepting char pointer and length.
You may use String for both buffers, and it will automatically delete itself too.

ericoporto: > then it may be convenient to have helper functions mimicing "strdup" but return a C++ container

I think this sounds best but wasn't sure where to put this helper. I thought about mimicking but taking (optional?) start and end parameters so it can either duplicate or substring.

ivan-mogilko:
> I am not convinced that the final code is actually getting proper offset in unicode characters, even as is in the current branch,

It looks like returning a correct result in the current branch, because it does this after finding a "needle":
```
*offs = 0;
int at = ustrlen(tempbuf1);
```
This puts a null terminator right where the "needle" was found, and calculates the ulength of the string before that.

You can do the same. Or you may try to optimize this, and make a separate version of ustrstr that also has a pointer to "index" as an argument, and fills in an index at which a substring was found.

Engine/ac/string.cpp
ivan-mogilko: > I think this sounds best but wasn't sure where to put this helper

StrUtil namespace is a good place for this, in string_utils.h/cpp

ericoporto:
Right, this reminds me, stupid question, but the returned index is relative to the 0 character of the original string right? Not the substring.

ivan-mogilko:
> Right, this reminds me, stupid question, but the returned index is relative to the 0 character of the original string right? Not the substring.

Yes, it should be an index in the full string; the input parameters are also indexes in full string, so return value is consistent with them.

Engine/ac/string.cpp
ericoporto: > If using String, it will need to be added unicode-aware MakeLower/Upper methods (similar to existing String::ReverseUTF8).

Erh, just to be sure I am going the right way, the reason I thought to not go this way is because string.cpp is currently independent from allegro, so the only way to keep it this way would be to duplicate the utolower/utoupper functions that apply on a single char.

ivan-mogilko: > the reason I thought to not go this way is because string.cpp is currently independent from allegro, so the only way to keep it this way would be to duplicate the utolower/utoupper functions that apply on a single char.

Hmm, right, getting allegro dependency there will be not good.
So either the raw buffer way where a substring is returned as `std::vector<char>`, or if in String then it has to be fully recreated there, using at most only Utf8 utilities header.

ericoporto: I ended up adding to the StrUtil namespace.

ericoporto:
ok, I think I fixed all things. I also ended adding a table entry for `IndexOf^1` pointing to `StrContains` as it was.

ericoporto:
ok, I think I fixed all things. I also ended adding a table entry for `IndexOf^1` pointing to `StrContains` as it was.

Also I had a [copy paste error](https://github.com/adventuregamestudio/ags/compare/9e2c6f4bf75f828d3764216187279af3bd99c61b..c6568fb44d64a37dee9201e663a9e578b1bd91a1) that is why it wasn't working at all, fixed now.

ivan-mogilko:
We must clarify what does "end index" mean in IndexOf: is it a *last* character of a searched range, or a first index *after* range. Right now the function works as if it were the index *after*. In other words startIndex is inclusive and endIndex is exclusive, the search is done in range `[startIndex, endIndex)`. Is that a expected behavior?

In regards to the code, I'd argue that for the new Substring function it may be better to have start and length parameters, that's common for "substring" functions.

ericoporto:
I think Length is less ambiguous than end index, wouldn't it be better to use it instead as input parameter?

ivan-mogilko:
Personally I don't mind "end index" so long as it's clearly documented; and in C and C++ it's traditionally used as an exclusive range ending (i.e. how iterator parameters work).

@messengerbag, you have suggested this parameter, what do you think about this?

ericoporto:
I think C# uses a "count" which would be length in the [C# String.IndexOf](https://learn.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-8.0). There are a lot of overloads, so not so sure if there's one that would also use index instead.

ivan-mogilko:
Alright, we may use "count" too. It's easy to convert to count if you know ending index.

ericoporto:
> BTW in C#'s example the "comparison type" is the last argument after startIndex and count.

Uhm, we also have it as last parameter in the other String methods we have, I think switching it here too for consistency may be a good idea. Do you agree?

ivan-mogilko:
Maybe call this parameter "count"? Count is number of characters after index, length is a length of an object, like a string.

ivan-mogilko:
Maybe call this parameter "count"? Count is number of characters after index, length is a length of an object, like a string.

> we also have it as last parameter in the other String methods we have, I think switching it here too for consistency may be a good idea. Do you agree?

Yes, I think that will be alright.

ericoporto:
ok, just a check, if we are inverting and switching for count, I think 0 would be better than -1 to be evaluated as instead going through the full length. I think that would be easier to remember as default for it.

ivan-mogilko:
> if we are inverting and switching for count, I think 0 would be better than -1 to be evaluated as instead going through the full length.

That makes sense, because 0 is an invalid value for this operation.

ericoporto:
ok, I think now all is correct!

messengerbag:
> @messengerbag, you have suggested this parameter, what do you think about this?

I suppose this has already been settled, but since I was asked, I personally find that endIndex is usually more convenient, and I think making it non-inclusive would be preferable. That is the way which, in my experience, most often allows you to simply use the variable value you already have, without any kind of arithmetic.

But obviously, the conversion between the alternatives is pretty trivial, so it's not a huge deal.

ericoporto:
Yeah, I had made endIndex non inclusive and also because in my head it was more convenient but in the end I think using count is easier to understand, less ambiguous, plus I think it will be easier to figure it out that 0 causes it to run until string end.

I am trying to make the API easier to get by reading parameter names and the tooltip comment and I think this achieved design with count does make sense in a somewhat more intuitive way.

PR #2501, opened 2024-08-17
AGS 4: Support proper 32-bit color values

### LINKED ISSUE #1980: Support true 32-bit colors in game data and script API

ivan-mogilko:
Resolves #1980

What is done:

* Color numbers encode 32-bit R8G8B8,
* Color properties correctly map R-G-B input to a 32-bit color number,
* Color Finder updated to display correct 32-bit values and matching colors, ignoring special 0-31 entries,
* 8-bit games should still interpret color numbers as palette indexes,
* Game.GetColorFromRGB script function updated to return a 32-bit R8G8B8,
* Old projects are **updated on import**, all color properties converted from 16-bit R5G6B5 to 32-bit R8G8B8 (except when it's a 8-bit game).
* Reimplemented use of hardcoded colors in some parts of the game, like built-in dialogs, etc.
* Cursor's hotspot crosshair colors are saved as int32 in game data (was saved as int16).
* Default color values in the Editor (Character speech color, GUI colors etc), - these are currently set as 0-31 palette indexes. We cannot have "default values" for more than 1 mode simultaneously; so I left these for now, and made them converted to the current mode when adding an item in the editor.
* Editor now converts all color properties when Color Depth setting is switched (between palette index and RGB number).
* Engine updates color values when loading older games. I also increased save versions, but did not do any upgrade there (it may be added later if really needed).

Additionally:
* Merged pairs of Color/ColorNumber properties into a single one (keeping data format unchanged), which displays a color value differently depending on the Game's ColorDepth setting. The color picker only works in 32-bit mode for now, but some kind of "palette picker" may be implemented in the future if desired.

What is left:
* Alpha value is not settable yet, and is always assumed as 255 when color is used to draw anything in the engine. But alpha should rather be addressed as a separate task. Also color properties should maybe changed from `int` to `uint` for this; alternatively they should use a custom editor attribute that displays unsigned numbers (or maybe in hex).

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -96,6 +96,7 @@ enum CharacterSvgVersion
     kCharSvgVersion_36115   = 4, // no limit on character name's length
     kCharSvgVersion_400     = 4000000, // extended graphic effects (blend, rotate,...)
     kCharSvgVersion_400_03  = 4000003, // compat with kCharSvgVersion_36115
+    kCharSvgVersion_400_09  = 4000009, // 32-bit color properties
 };
 
 struct CharacterExtras;
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -127,6 +127,8 @@ BlendModes
 Raised for org purposes without format changes
 4.0.0.8:
 Palette component range changed from 64 to 256
+4.0.0.9:
+32-bit color properties
 */
 
 enum GameDataVersion
@@ -140,8 +142,9 @@ enum GameDataVersion
     kGameVersion_399            = 3999999,
     kGameVersion_400            = 4000000,
     kGameVersion_400_08         = 4000008,
+    kGameVersion_400_09         = 4000009,
     kGameVersion_LowSupported   = kGameVersion_360_21,
-    kGameVersion_Current        = kGameVersion_400_08
+    kGameVersion_Current        = kGameVersion_400_09
 };
 
 // Data format version of the loaded game
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -275,8 +275,8 @@ void GameSetupStruct::ReadFromSavegame(Stream *in)
     // of GameSetupStructBase
     playercharacter = in->ReadInt32();
     dialog_bullet = in->ReadInt32();
-    hotdot = static_cast<uint16_t>(in->ReadInt16());
-    hotdotouter = static_cast<uint16_t>(in->ReadInt16());
+    in->ReadInt16(); // [DEPRECATED] uint16 value of a inv cursor hotdot color
+    in->ReadInt16(); // [DEPRECATED] uint16 value of a inv cursor hot cross color
     invhotdotsprite = in->ReadInt32();
     default_lipsync_frame = in->ReadInt32();
 }
@@ -289,8 +289,8 @@ void GameSetupStruct::WriteForSavegame(Stream *out)
     // of GameSetupStructBase
     out->WriteInt32(playercharacter);
     out->WriteInt32(dialog_bullet);
-    out->WriteInt16(static_cast<uint16_t>(hotdot));
-    out->WriteInt16(static_cast<uint16_t>(hotdotouter));
+    out->WriteInt16(0); // [DEPRECATED] uint16 value of a inv cursor hotdot color
+    out->WriteInt16(0); // [DEPRECATED] uint16 value of a inv cursor hot cross color
     out->WriteInt32(invhotdotsprite);
     out->WriteInt32(default_lipsync_frame);
 }
--- a/Common/ac/gamesetupstruct.h
+++ b/Common/ac/gamesetupstruct.h
@@ -39,7 +39,6 @@ namespace AGS
     }
 }
 
-//using AGS::Common::Interaction;// CLNUP stuff for old interactions
 using AGS::Common::PInteractionScripts;
 using AGS::Common::HGameFileError;
 
@@ -102,9 +101,6 @@ struct GameSetupStruct : public GameSetupStructBase
     // we could use a kind of hash map containing fixed-sized arrays, where size of
     // array is calculated based on key spread factor.
     std::vector<SpriteInfo> SpriteInfos;
-
-    // Get game's native color depth (bits per pixel)
-    inline int GetColorDepth() const { return color_depth * 8; }
 
 
     GameSetupStruct();
--- a/Common/ac/gamesetupstructbase.cpp
+++ b/Common/ac/gamesetupstructbase.cpp
@@ -93,8 +93,8 @@ void GameSetupStructBase::ReadFromFile(Stream *in, GameDataVersion game_ver, Ser
     color_depth = in->ReadInt32();
     target_win = in->ReadInt32();
     dialog_bullet = in->ReadInt32();
-    hotdot = static_cast<uint16_t>(in->ReadInt16());
-    hotdotouter = static_cast<uint16_t>(in->ReadInt16());
+    in->ReadInt16(); // [DEPRECATED] uint16 value of a inv cursor hotdot color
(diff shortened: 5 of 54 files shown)


ivan-mogilko:
I've been wondering, would that make sense to handle existing duality of Color / ColorNumber properties by hiding one of this pair depending on the current game's color depth? That is - display ColorNumber if game is 8-bit and Color (rgb) if game is 32-bit.

But then, it would also be nice to have color displayed on ColorNumber field. Maybe even add a [...] ColorNumber property which opens a palette for selecting a color.

Alternatively, how feasible that would be to merge Color and ColorNumber properties together?
Only ColorNumber is serialized, and Color (rgb) field is there only for editing at design time. This makes it possible to remove one, rename ColorNumber to just Color. But the biggest question is whether it would be possible to change the *looks* of the field depending on game settings. Supposedly the field may have a custom editor attribute, that draws the field and provides [...] button depending on some condition.

ivan-mogilko:
The "Github Actions" are failing because of the test games. I suspect the DrawingSurface tests will have to be adjusted after this is merged.

ivan-mogilko:
I managed to figure out how to make a property that may have two meanings and textual representation in a PropertyGrid, depending on game settings (see two commits above).

This lets merge a pair of properties ColorNumber and Color (rgb) into a single one. I used "ColorNumber", since it's the one that was serialized, so to keep data format unmodified.

The idea is that this property will now look like a "color number" if the game is 8-bit, and like a RGB string when the game is 32-bit. Internally it is always integer (either palette index, or encoded RGB). In 32-bit mode it opens color picker on pressing a button. In 8-bit mode button is removed, but in theory it might be possible to implement a "palette selector" in the future.

Merged all existing color property pairs into a single property (in character, gui, etc), and even made "cursor hotspot colors" properties to work the same way, so you can have a colored rectangle and color picker there too, without adding more properties.

ivan-mogilko:
If I understood right, @AlanDrake have tested this PR (he mentioned this on Discord).

AlanDrake:
Yes. I've tested upgrading a 32bit project, a 8bit one, and some simple DynamicSprite drawings.

ivan-mogilko:
Alright, fingers crossed...

PR #2504, opened 2024-08-19
Editor: update ConstructBasicFileListForTemplate() for AGS 4

### LINKED ISSUE #2242: AGS 4: adjust a list of files packaged in template
Since in ags4 we have rooms in separate subfolders now, the old template rules that package *.crm files are no longer applicable.

Need to fix this:
- remove *.crm files from the list of packaged files;
- add all the room sources from Rooms/N/ subfolders.

The related code may be found here:
https://github.com/adventuregamestudio/ags/blob/9a65d6e2b4cea69f628e7fb0c604a38a4ca9becd/Editor/AGS.Editor/Tasks.cs#L44

ivan-mogilko:
Fix #2242

Seems like a very trivial fix, but please check out in case I missed anything.


ericoporto:
The code looks alright but I don't have a way to test this right now. If you want to be super specific the room data is named `data.xml` and the room script is `room*.asc`.

Issue #2063, opened 2023-07-28
Editor: for making templates, support a include/ignore file in a project folder


ivan-mogilko:
As per @ericoporto 's suggestion [here](https://github.com/adventuregamestudio/ags-template-source/issues/40#issuecomment-1580759552)

**Problem**

Currently the template packaging from AGS Editor has a hardcoded list of files, and does not let to package custom files, let alone subfolders. At the same time it might be desired to get other assets into the template (sprite sources, for instance), which are located in subdirs for convenience.

**Solution**

We could implement a GUI dialog in the editor that selects files for template, but then it would also be wanted to save this information for the future use, as a user may want to recreate template from the same game project later, multiple times, like we do with our default template projects. This means that the GUI is not a solution, we need to begin with a file that would be saved in the project folder, and won't be a part of the game project itself (no data saved in Game.agf, Game.agf.user).

Note that this approach is not something new for template building, this process already supports template.txt and template.ico, which define optional template's description and an icon respectively.

As @ericoporto suggested, this may be done using a file formatted similar to .gitignore, as a list of patterns. It may be interpreted either as a "include list" or "exclude list", whatever seems convenient. It does not have to be exactly .gitignore kind of syntax, may be of standard regex syntax, or even classic wildcards.
In terms of implementation, .NET framework already has a regex support, and C++ that we use in our tools has std::regex, which may be more or less enough to support using such file when gathering a package.

**Tasks**

- [ ] Decide on a file name and format.
- [ ] Implement using this file by the Editor when creating a template from the project, along with hardcoded standard game files.
- [ ] Need to decide how Editor uses this file: either as a complete override of hardcoded list, or as an addition to hardcoded list. This choice is important, as it will define how this feature is going to be used. Given a list may be passed to other tools, including possibly third-party tools, perhaps "override" is better (although that will require bit more effort from user to configure it).
- [x] Optionally, support such file passed as an extra arg to our agspak tool (This may be moved out to a separate ticket, as with command line tools there's more flexibility in which files they are applied to already).

ericoporto:
It appears it's just redoing the function below here

https://github.com/adventuregamestudio/ags/blob/b403471f7e4737cf88fb6f46b89556b049c33e1a/Editor/AGS.Editor/Tasks.cs#L44

There is some additional details that when writing the CLIB package it appears there are some hardcoded conditions.

[...]

PR #2505, opened 2024-08-19
Tools: agspak with ignore file support

ericoporto:
partial for #2063 , a later change would be to actually use agspak for building template.

I played a bit with it and it seems it's working.

I noticed I can begin the file with the first line being * and then have the later lines all be prepended by `!` to turn into an include pattern instead. A possbile enhancement for people who prefer the include patter could be to also check for both a `template.include` and `template.ignore` file and adjust the behavior by this.

### DIFF
--- a/Tools/data/include_utils.cpp
+++ b/Tools/data/include_utils.cpp
@@ -1,0 +1,227 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-2024 various contributors
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// https://opensource.org/license/artistic-2-0/
+//
+//=============================================================================
+#include "data/include_utils.h"
+#include "util/file.h"
+#include "util/path.h"
+#include "util/textstreamreader.h"
+#include <regex>
+
+using namespace AGS::Common;
+
+namespace AGS
+{
+namespace DataUtil
+{
+
+using AGS::Common::Stream;
+
+enum PatternType {
+    eInclude = 0,
+    eExclude
+};
+
+struct Pattern {
+    PatternType Type;
+    std::regex Regex;
+    String TextualTranslatedPattern; // for debug purposes
+    String TextualOriginalPattern; // for debug purposes
+};
+
+static String translate_to_regex_string(const String &pattern)
+{
+    int i = 0;
+    int n = (int)pattern.GetLength();
+    String result;
+
+    while (i < n) {
+        char c = pattern[i];
+        ++i;
+
+        if (c == '*') {
+            result.Append(".*");
+        } else if (c == '?') {
+            result.AppendChar('.');
+        } else if (c == '[') {
+            int j = i;
+            /*
+             * The following two statements check if the sequence we stumbled
+             * upon is '[]' or '[!]' because those are not valid character
+             * classes.
+             */
+            if (j < n && pattern[j] == '!')
+                ++j;
+            if (j < n && pattern[j] == ']')
+                ++j;
+            /*
+             * Look for the closing ']' right off the bat. If one is not found,
+             * escape the opening '[' and continue.  If it is found, process
+             * the contents of '[...]'.
+             */
+            while (j < n && pattern[j] != ']')
+                ++j;
+            if (j >= n) {
+                result.Append("\\[");
+            } else {
+                String stuff = pattern;
+                stuff.ClipMid(i, j-i);
+                stuff.Replace("\\"," \\\\");
+                char first_char = pattern.GetAt(i);
+                i = j + 1;
+                result.Append("[");
+                if (first_char == '!') {
+                    result.Append("^");
+                    stuff.ClipLeft(1);
+                    result.Append(stuff);
+                } else if (first_char == '^') {
+                    result.Append("\\");
+                    result.Append(stuff);
+                } else {
+                    result.Append(stuff);
+                }
+                result.Append("]");
+            }
+        } else {
+            // I thought this may cause issue with utf-8 but the python approach is to escape all characters
+            if (isalnum(c)) {
+                result.AppendChar(c);
+            } else {
+                result.Append("\\");
+                result.AppendChar(c);
+            }
+        }
+    }
+    return result;
+}
+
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
I cannot agree with a decision to hardcode the ignore file's name, and have doubts about hardcoded ignored list of files too.
This might work for the Editor, but looks very much off for agspak.

agspak was supposed to be a simple universal tool that packs any chosen set of files into the ags package format. It may be used to pack anything, and so it may be necessary to be able to give it any include or exclude list of files as a command line argument, regardless of the purpose (template or not template). E.g. one might want to package custom list of files for the game release too.

There is more than 1 type of template in AGS, besides the game template there's room template that has different set of files.

As for the hardcoded "automatic ignore list", this makes this tool depend on a particular version of the Editor.
I can also see at least 1 entry that is not a common project file, but something depending on a operating system (DS_Store), which seems strange.

ericoporto:
I think I changed it as requested

ivan-mogilko:
> Now for the packaging, I find it weird to be a different tool for packing and other to unpacking - besides being also terrible ergonomics that I can't quickly rerun the tool in the other direction to check if it packaged correctly. I may be more used to something like tar which has many commands with many configs.

These 2 tools may of course be merged into one; I cannot remember why I chose to make 2 separate tools, and I won't insist on keeping them separate.

In regards to the agspak use with the include/ignore lists: the idea was that agspak is simple and just does what user tells it to, without any hidden rules. If there are special rules that have to be applied often - that's definitely something to think about, but, for example, that may be done with the use of a batch script. For instance, the Editor may come with a batch script that runs agspak for making a game template, passing standard include / ignore files.

ericoporto:
So the big question of the include rule, any idea how to name this? I just named the command name as `--file` because anything I write sounds weird... Like `--include-file` kinda sounds like it expects a file that is included xD

This is more like a package "project" file of sorts. But I can't find good names.

ivan-mogilko:
Alright, I finally made a proper test, and the pattern file seems to work, both for including and excluding files.

There are few things that I would suggest to change.

Firstly, regarding the option name, one idea i had is simple `--pattern-file` but then idk about one-letter option, since "p" is already taken by "partition". Then there's `--file-list`, and either `-l` or `-f` for single char.
But I was wondering if it also makes sense to have a future option of passing a list of patterns directly in a command line. If it's short, then this may be easier than making a file. So makes sense to at least reserve some name for that too.

So what about this:
* `--pattern-file` and `-f` - for a file with include/exclude list
* `--pattern-list` and `-l` - for passing this pattern as cmd arg.

---

Other than that, in regards to "verbose" setting and printing included files.
I propose to print files that have been successfully packed, similar to how `agsunpak` prints files that have been successfully unpacked. This may be seen here:
https://github.com/adventuregamestudio/ags/blob/5e8ffad9e153a77282ff03f1b058752782e1e473/Tools/data/mfl_utils.cpp#L63-L67
IMO such approach has an advantage that:
- it clearly states which files were processed successfully and which not;
- in certain case it allows to watch the progress, e.g. if program crashed you will know which was the last processed file.

Tools/data/include_utils.h
```diff
@@ -26,10 +26,10 @@ namespace DataUtil
 using AGS::Common::HError;
 using AGS::Common::String;
 
-// Pass a list of files that is filtered in place by ignore-like patterns in a file
-HError IgnoreFiles(std::vector<String> &files, String parent, String ignore_file, bool verbose);
+// Pass a list of files that is filtered in place by include-like patterns in a file
+HError IncludeFiles(std::vector<String> &files, const String& parent, const String& include_pattern_file, bool verbose);
```
ivan-mogilko, line 30: Please, make this function have separate arguments for input and output files vector. IMHO that's more convenient because there may be different situations, and also easier to understand that this function requires *input files* (I did not understand this first, and though that this function also tests the directory somehow). The caller may replace source vector with resulting one if they need to.

Tools/data/include_utils.cpp
```diff
@@ -152,24 +152,24 @@ std::vector<Pattern> description_to_patterns(const std::vector<String>& descript
 std::vector<String> match_files(std::vector<String> files, std::vector<Pattern> patterns)
```
ivan-mogilko, line 152: Both input vectors should be passed as a const reference here.

Tools/data/ignore_utils.cpp
```diff
@@ -0,0 +1,216 @@
...
+            } else {
+                result.Append("\\");
+                result.AppendChar(c);
+            }
+        }
+    }
+    return result;
+}
+
+String normalize_separators_in_text(String path)
```
ivan-mogilko: Input path should be passed as a const reference.

Tools/data/mfl_utils.cpp
```diff
@@ -70,6 +70,34 @@ HError UnpackLibrary(const AssetLibInfo &lib, const String &lib_dir, const Strin
...
+    for (FindFile ff = FindFile::Open(parent, "*", true, false, do_subdirs ? -1 : 0);
+         !ff.AtEnd(); ff.Next())
+    {
+        String filename = ff.Current();
+        files.push_back(filename);
+    }
+    return HError::None();
+}
+
+HError MakeAssetListFromFileList(std::vector<String> files, std::vector<AssetInfo> &assets, const String &asset_dir)
```
ivan-mogilko, line 86: vector of filenames should be passed as a const reference here too.

Tools/data/mfl_utils.h
```diff
@@ -35,6 +35,10 @@ namespace DataUtil
     // if the asset name contains directories, they will be created as sub-
     // directories inside dst_dir.
     HError UnpackLibrary(const AssetLibInfo &lib, const String &lib_dir, const String &dst_dir);
+    // Gather a list of files from a given directory as a vector of strings
+    HError MakeListOfFiles(std::vector<String> &files, const String &asset_dir, bool do_subdirs);
+    // Prepare list of assets from a list of filenames
+    HError MakeAssetListFromFileList(std::vector<String> files, std::vector<AssetInfo> &assets, const String &asset_dir);
```
ivan-mogilko, line 41: vector of filenames should be passed as a const reference here too.

ericoporto:
I think I fixed the mentioned issues. I squashed the commits too, so it's ready for review again.

ivan-mogilko:
I think this is fine now.

ivan-mogilko:
Hmm, weirdly enough, the recursive file search was broken, it was only including *first subdirectory* on each level.
So, full recursive packaging was never working correctly before.

Issue #2418, opened 2024-05-08
New compiler: potential code buffer overflow when compiling certain snippets


ivan-mogilko:
CC @fernewelten 

As demonstrated by #2417.

Certain code snippets cause write or read beyond the valid data in `code` buffer.
This was never noticed before, because compiler allocates this buffer with an extra capacity, and offset mistakes are quite small (few bytes).

Errors occur in functions `AGS::ForwardJump::Patch` and `AGS::BackwardJumpDest::Set`.

Specifically, following tests cause the error:

Bytecode1.Ternary5
```
        float main()
        { 
            int I1a = 0 ? 10 : 20;
            int I1b = 2 ? 30 : 40;
            int I2a = 0 ?: 50;
            int I2b = 3 ?: 60;
            int I3a = 0 ? I1a : (7 + I1b);
            int I3b = 4 ? I2a : (7 + I2b);
            int I4a = 0 ? 70 : I3a;
            int I4b = 4 ? 80 : I3b;
            int I5a = 0 ? I4a : 90;
            int I5b = 5 ? I4b : 100;
            int I6 = 0 ? : I5a;
            return 0.;
        }
```

Compile0.Ternary02
```
        int main()
        {
            return 2 < 1 ? 1 : 2.0;
                    break;
        } 
```

Compile1.CompileTimeConstant2
```
        int main() {
            while (1)
            {
                const int CI2 = 4712;
            }
            float CI2;
        }
```

fernewelten:
Oops … I'm just seeing this. I'm onto it.
I was just going to report that the compiler is failing its googletests; this issue is the reason for that.

ivan-mogilko:
Fixed by #2510

PR #2508, opened 2024-08-20
building ags4 compiler in gcc

ericoporto:
So I am trying to make it possible to run the ags4 compiler as standalone. 

The biggest issue I am having right now is actually being able to build it in GCC. 

Edit: managed to fix a lot of issues and it actually builds now! I think I still may need to do some adjustments for clang and apple-clang in macOS.


ericoporto:
I have no idea what to do with the Makefile, like, ideally I would prefer having two makefiles but I have no idea how to do that...

ericoporto:
ok, getting to build the compiler but now getting test failures... Except the `Compile1` `CTEvalIntModulo` test, this one is exiting with SIGSEGV. :/

Compiler/test2/cc_parser_test_0.cpp
```diff
@@ -56,7 +56,7 @@ int cc_compile(std::string const &inpl, AGS::ccCompiledScript &scrip)
     AGS::MessageHandler::Entry const &err = mh.GetError();
     static char buffer[256];
     ccCurScriptName = buffer;
-    strncpy_s(
+    strncpy(
```
ivan-mogilko, line 59: If strncpy_s cannot be used, it's better to use snprintf instead, because strncpy has unexpected quirks. For instance, it does not guarantee null-termination if string is longer than the buffer.
https://en.cppreference.com/w/c/string/byte/strncpy

ivan-mogilko: On another hand, the way ccCurScriptName is a horrible hack (and dangerous too), it must be investigated and possibly replaced with simply std::string. Maybe separately from this PR.

ericoporto: strncpy_s is a Microsoft extension. We could make an equivalent somewhere and use that where it's used otherwise it won't work in non-MSVC compilers.

ericoporto:
Hey there's an error I am getting that isn't happening in the CI somehow - I am using Ubuntu 16.04 to use a compiler closer to the one we are using in debian. Anyway, I have a different compiler test causing SIGSEGV. If you have some time at some point later, please try building and running the tests on after building with GCC on Linux (or MinGW on Windows).

ivan-mogilko:
Which tests exactly cause sigsevs?
There's this know problem, for instance: #2417

ericoporto:
OK, I found what was causing. There was an error message for the modulo calculation that had a `%` sign that was supposed to be just a percent sign and not part of a format string, but no matter what it would eventually hit somewhere in the code that thought it should be a format string. I managed to fix by removing the percent sign altogether since escaping didn't help. Somehow this was not an issue in Windows.

ericoporto:
One thing that I won't do here, but it should be done at some point is to adjust the namespace use in the new compiler it is very inconsistent.

ivan-mogilko:
> There are places in the code that uses Long that are problematic. I don't know how to fix those, they are causing test errors. We should really never use long ever anywhere because of the way it behaves differently in different archs...

I can look tomorrow, but I'm pretty much certain that any `long` in compiler may be changed to `int32`, because ags bytecode is based on 32-bit signed ints.

Compiler/script2/cc_symboltable.cpp
```diff
@@ -564,7 +571,7 @@ AGS::Vartype AGS::SymbolTable::VartypeWithArray(std::vector<size_t> const &dims,
         aoa_dims.reserve(old_dims.size() + dims.size());
         aoa_dims = dims;
         aoa_dims.insert(aoa_dims.end(), old_dims.begin(), old_dims.end());
-        // Cut off the first '[�]', it will be replaced by the index of the joint array
+        // Cut off the first '[�]', it will be replaced by the index of the joint array
```
ivan-mogilko, line 574: Is it intended to insert these symbols here? Not sure if it's UTF-8 or another encoding, but perhaps it's better to replace with 3 dots.

Compiler/script2/cs_compiler.cpp
```diff
@@ -201,6 +201,12 @@ static std::unique_ptr<ScriptTOC> ccCompileDataTOC(const SymbolTable &symt, cons
         new ScriptTOC(std::move(tocb.Finalize(rtti))));
 }
 
+#ifdef __GNUC__
+inline int strncpy_s(char* dest, size_t destsz, const char* src, size_t count)
+{
+    strncpy(dest, src, std::min(destsz, count)); return 0;
+}
+#endif
```
ivan-mogilko, line 209: The replacements for standard string functions are better put into string_compat.h/cpp.
strncpy_s seems to be a standard now, starting with C11, according to https://en.cppreference.com/w/c/string/byte/strncpy

Compiler/script2/cs_message_handler.h
```diff
@@ -23,7 +23,7 @@ namespace AGS
 class MessageHandler
 {
 public:
-    enum Severity
+    enum Sev
```
ivan-mogilko, line 26: Why replace the name? Severity is not too long, and "Sev" is more difficult to guess.

Compiler/test2/cc_bytecode_test_1.cpp
```diff
@@ -1548,7 +1549,7 @@ TEST_F(Bytecode1, Attributes08) {
 TEST_F(Bytecode1, Attributes09) {
 
     // Function call to 'set_Visible()' mustn't go awry
-    // After loading 'true' to AX�, must protect AX from being clobbered (push it)
+    // After loading 'true' to AX�, must protect AX from being clobbered (push it)
```
ivan-mogilko, line 1552: Another case of strange symbols, maybe automatic fix by a visual studio?

Compiler/script2/cs_parser.cpp
```diff
@@ -4199,7 +4199,7 @@ void AGS::Parser::ParseVardecl_Import(Symbol var_name)
     }
 
     _sym[var_name].VariableD->TypeQualifiers[TQ::kImport] = true;
-    int const import_offset = _scrip.FindOrAddImport(_sym.GetName(var_name));
+    int const import_offset = _scrip.FindOrAddImport(_sym.GetName(var_name).c_str());
```
ivan-mogilko, line 4202: This change looks unnecessary, FindOrAddImport accepts `const std::string&", why this is done?

Compiler/test2/cc_bytecode_test_1.cpp
ericoporto: this file is not encoded in UTF-8 and is instead using some other encoding I can't figure it out...

Compiler/script2/cs_compiler.cpp
ericoporto: C11 is not C++11, you can't access it from a C++ code unit.

Compiler/script2/cs_message_handler.h
ericoporto: For some reason no matter what I do the Severity will cause issue in the compiler like it's picking the wrong thing there - the build error gives a very weird complain. Specifying the field won't work.

Compiler/script2/cc_symboltable.cpp
ericoporto: I don't know what the alien head means there in the other file. I am using CLion on Linux and it keeps complaining, ONLY in the new compiler files, that the files isn't UTF-8 and that it doesn't know the right encoding.

Compiler/script2/cs_parser.cpp
ericoporto: this is probably a mistake, will fix later, for now I just want to get it building so I can switch back to the Windows PC and polish the commits there...

Compiler/script2/cs_compiler.cpp
ivan-mogilko: Yes I know that C11 and C++11 are different things. What I meant is that's a part of a standard C library, and C library may be used by C++ compilers.

I am not aware of such details whether C11 standard is widely supported by C++ compilers now, but, for example, C99 is already supported by most C++11 compatible compilers.

Compiler/script2/cc_symboltable.cpp
ericoporto: oh, now I noticed you saw three dots, I see them on macOS but on Linux it was showing an Alien Head emoji

Compiler/script2/cs_parser.cpp
ericoporto: fixed

Compiler/test2/cc_bytecode_test_1.cpp
ericoporto: fixed by removing this symbol

Compiler/script2/cc_symboltable.cpp
ericoporto: fixed by using the three dots suggested - now I understand the suggestion.

ericoporto:
I attempted to fix the long usage but it is still wrong and the tests Bytecode1 and Compile0 are still failing

ivan-mogilko:
Unrelated to above, but when trying to build a x64 build in MSVS, compiler note this line:

```
sprintf(appendage, "^%d", _sym.FuncParamsCount(name_of_func) + 100 * _sym[name_of_func].FunctionD->IsVariadic);
```

the argument type here is size_t, but formatting uses `%d`, should be `%zu`, or this will break on 64-bit build.

ivan-mogilko:
I might suggest to disable following tests temporarily:
- Bytecode1.Ternary5
- Compile0.Ternary02
- Compile1.CompileTimeConstant2

These tests are known to cause incorrect memory access (#2418), which hypothetically may lead to other glitches.

Compiler/script2/cs_message_handler.h
ericoporto: I think I found a fix but am using the Apple Clang now, will recheck later with gcc 5.4 in Ubuntu.

ericoporto:
So I commented the said tests but I still am getting an error. I also adjusted the mentioned appendage section and the tests kept failing. I will look for more possible printf like errors.

Compiler/script2/cs_message_handler.h
ivan-mogilko: Is the error only caused by member variable declaration `Severity Sev`, or something else besides that?

If only by variable, then my meaning was that it's better to rename a variable than a type.

ericoporto: I removed the renaming. It looks like it worked.

Compiler/test2/cc_parser_test_0.cpp
ericoporto: ok, I wrote a better one using snprintf and abusing the format string, I think it's really forcing the buffer to be null-terminated and has no easy area for errors I think.

Compiler/script2/cs_compiler.cpp
ericoporto: I wrote one and placed in the string_compat

ericoporto:
The failing test `Bytecode1.LongMin1` looks like this

https://github.com/adventuregamestudio/ags/blob/2bb36fe514f0a80f125bfc2db3da1d828e825787/Compiler/test2/cc_bytecode_test_1.cpp#L3296-L3310

image

ericoporto:
Btw, the tests that aren't failing aren't failing because of this PR, they just weren't ran at all before in the CI because they just weren't built in CMake at all. This is the first time these are being built using GCC and Clang. You can see the CMake test that is built using MSVC still makes all tests pass.

ivan-mogilko:
There has to be a difference in how C++ compilers treat type conversions somewhere, same code works in 64-bit msvc, but results in integer overflow in others.

ericoporto:
I found it! There was a LONG_MIN I forgot to replace! There is still ONE test that fails , the `Scan.ConsecutiveStringLiterals1`, but only fails in gcc and doesn't in Apple Clang builds.

ivan-mogilko:
> There is still ONE test that fails , the Scan.ConsecutiveStringLiterals1, but only fails in gcc and doesn't in Apple Clang builds.

Is there any way to know which exact assertion fails? There are 4 of them in this test, but logs only have a message that the test failed.

ivan-mogilko:
Unrelated, by searching the logs for warnings in compiler tests, I found this:

> /ags/Compiler/script2/cc_symboltable.h:481:126: warning: no return statement in function returning non-void [-Wreturn-type]
inline Symbol *FindStructComponent(Symbol strct, Symbol component) const { FindStructComponent(strct, component, strct); }

Too good this function is not used anywhere...

ericoporto:
So I understand exactly why the test is failing, I actually don't understand why it should work... the issue is this.

In the [`ReadInStringLit`](https://github.com/adventuregamestudio/ags/blob/2bb36fe514f0a80f125bfc2db3da1d828e825787/Compiler/script2/cs_scanner.cpp#L549), this is the function that globs something `"ABC" "DEF"` into a single `"ABCDEF"` string literal.

When it hits the second `"` it will try to see if there is only whitespace followed by another string. But ags has a special string literal that is a line with a `" __NEWSCRIPTSTART_scriptname";`, and so it has a minimal code that does the disambiguation. Here is the code

https://github.com/adventuregamestudio/ags/blob/2bb36fe514f0a80f125bfc2db3da1d828e825787/Compiler/script2/cs_scanner.cpp#L602-L608

Now this code is using  standard c++ stream code here, and notice the `_eofReached` here. In this test the `"__NEWSCRIPTSTART_"` prefix is longer than the remaining characters that are there (`"expialidocious"`). So what happens is `_eofReached` is set but once things are undone, it's not unset. Because it remains set, once the while loop loops, it breaks and throws the error message: `Input ended within a string literal`. 

I don't understand why this is working in Windows or macOS.

ivan-mogilko:
If I remember correctly, in standard lib `eof()` returns `true` only when a stream *tries to read again* when being past the last char. It is not yet set when it's read everything out first time.
https://en.cppreference.com/w/cpp/io/basic_ios/eof

Debugging under MSVS shows that the stream always have at least 17 chars ready.

```
            pos_before_skip = _inputStream.tellg();
            _inputStream.seekg(0, std::iostream::end);
            auto pos_end = _inputStream.tellg();
            size_t len = pos_end - pos_before_skip;
```

ericoporto:
I don't think so because it reproduces in my old gcc 5.4, and in the new GCC on the Ubuntu 24.04 on CI and in the MinGW (also GCC) on the Windows CI - check the actions log here.

I guess these issues are the things we bypass when we use our own streams that are in ags. I am not sure how to fix this yet.

ivan-mogilko:
There's another similar explanation on stackoverflow:
https://stackoverflow.com/a/42135904

> To know that you are at the end of the file you should attempt to read further.

Although it's curious that the OP claimed that it returned "false" with gcc, but "true" with MSVC.

In any case, eof is supposed to only return true when there was an attempt to read again after the cursor already reached the end.
Perhaps debugging or verbose logging could help to figure out what happens in these tests.
We'd need to know whether the position was >= ending position of the stream prior to the "get" call. If it was before the end, then eof should not be set even if we tried to read more than stream had.

ivan-mogilko:
Hmm, no, I seem to be wrong here, maybe I misunderstood certain things.

If I do this:
`_inputStream.seekg(-5, std::iostream::end);`
right before the 
`_inputStream.get(tbuffer, kNewSectionLitPrefixSize + 1u, 0);`

then it will always return eof == true.

ivan-mogilko:
So, here's my proposal for a test:

```
pos_before_skip = _inputStream.tellg();
//<------------------------------------------------
_inputStream.seekg(0, std::iostream::end);
auto pos_end = _inputStream.tellg();
size_t has_count = pos_end - pos_before_skip;
_inputStream.seekg(pos_before_skip);
assert(has_count >= kNewSectionLitPrefixSize + 1u);
//<------------------------------------------------
```

This assets that the stream contains enough chars to read "kNewSectionLitPrefixSize + 1u".
This assertion never failed for me on MSVC, it always contained at least "kNewSectionLitPrefixSize + 1u" chars.
What will happen on GCC?

ericoporto:
image

A print from that code copy pasted and ran in the CLion IDE (you can see the values). I need to sleep now and I will look at this tomorrow. Also thanks for the input on the other PR, those are easy to adjust I should be able to adjust that one tomorrow.

ivan-mogilko:
Yes I had same values.
I forgot to tell to test _eofReached after this though, it will be interesting if it sets eof in this case, or there's another reason it appears set.

On a side note, this line in the test must be safeguarded by checking if `strings` is not empty:
```
EXPECT_STREQ("Supercalifragilisticexpialidocious", &string_collector.strings[0] + value);
```
because if scanner fails for any reason, this leads to a index out of range exception.

ericoporto:
Not super sure on the last commit but it makes all tests pass. I will wait for the CI to build everything and mark this as ready to review.

This PR is NOT meant to fix the standalone compiler issue. Now that I can get it to build and play with it, I will rethink on how to make the existent front end to properly glue with it, but the existing placeholder code should at least be there to make it easier to run it. I want to work on a next PR that actually has both the proper code for the standalone compiler plus builds both it and the other one in the CI and saves their binaries.

ivan-mogilko:
I think we should ask @fernewelten to check out this problem with eof.

ericoporto:
@fernewelten can you check the mentioned eof() issue? I explain it here [here](https://github.com/adventuregamestudio/ags/pull/2508#issuecomment-2307937530) and put a small fix in https://github.com/adventuregamestudio/ags/pull/2508/commits/3da2faee00986eb9c5f0dc9a1d3da7714830eb99 and https://github.com/adventuregamestudio/ags/pull/2508/commits/ff2eb6d8f27eee4b45cfa833802b6c5a774822d6 .

ivan-mogilko:
I think I'll merge this now, as all the relevant tests pass, and I was able to recompile and run a large existing project after these changes too. The fix to the "eof" check is contained in a single place of code and is easy to see; it may be revisited later. I will add a small comment there explaining what was found.

PR #2510, opened 2024-08-23
New compiler: Fix bugs uncovered by stricter 'compiledscript' code

fernewelten:
Fixes #2417

Parser code for ternaries could sometimes patch code cells that had once existed but then were ripped out of the codebase.

In 'while(true)' loops with bodies that didn't contain code, the parser tried to look at the first byte beyond the end of the codebase.


ivan-mogilko:
Thank you, appears fixed now.

PR #2511, opened 2024-08-25
Compiler: port preprocessor updates from C# to Cpp

ericoporto:
This is essentially https://github.com/adventuregamestudio/ags/commit/cc465235b5a20b2a5a377aa4140ee3cbbd6d66d8 (https://github.com/adventuregamestudio/ags/pull/2482) and https://github.com/adventuregamestudio/ags/commit/69704a20c7bb70e34609d6ec78d3549b2a2e8a9f (small part of https://github.com/adventuregamestudio/ags/pull/2180)

- remove undefined behavior from IsScriptWordChar function
- port detection of non-latin characters in the preprocessor
- port escaping of the script name by the preprocessor
- minor readjustment on the error logging to ensure easier error testing this was needed to ease comparison when the error string contained non-latin characters.

**NOTE:** I am modifying the IsScriptWordChar function which is used by the script compiler. It had an undefined behavior - the comment text is from cppreference website. It looks like it's still working but previous isalnum was returning true for non-latin characters, I don't think this breaks anything because in the script compiler this is used only once and we pass a `char` type variable instead of an int. I actually think this function should take a `char` as parameter since both existing uses are only passing a char.

### DIFF
--- a/Compiler/preproc/preprocessor.cpp
+++ b/Compiler/preproc/preprocessor.cpp
@@ -37,6 +37,8 @@ namespace Preprocessor {
     static const char * li_end = "\n";
 #endif
 
+    const Error Preprocessor::NoError = Error();
+
     size_t FindIndexOfMatchingCharacter(String text, size_t indexOfFirstSpeechMark, int charToMatch)
     {
         size_t endOfString = NOT_FOUND;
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
About IsScriptWordChar, somehow I did not realize this earlier, but this function is not correct for couple of reasons.
First, as you say, it should accept `char` parameter, and second, using std::isalnum seems incorrect, because it may return positive for a number of extended characters, and also depends on current locale. But these are not suitable for the "script wo

The C# version of preprocessor looks more correct, where it does not use "is alphanumeric", but explicitly compares to the only valid range of chars that may be used in script code:
```
        public static bool IsScriptWordChar(this Char theChar)
        {
            return ((theChar >= 'A') && (theChar <= 'Z')) ||
                ((theChar >= 'a') && (theChar <= 'z')) ||
                ((theChar >= '0') && (theChar <= '9')) ||
                (theChar == '_');
        }
```

PR #2512, opened 2024-08-25
Editor: implement Window Config, save SpriteChooser window state between sessions

### LINKED ISSUE #1484: Make AGS Editor Sprite View remember folder tree area size
> I have an issue that barely qualifies as a nitpick, but rather a small annoyance that I wonder if anything could be done about. I've been organizing all my sprites into folders and subfolders. This means I have to increase the window size to accommodate the expanded folder trees in the "select a sprite to use" window. Normally that's no big deal, but the window size resets to the original (too small) size every time I want to select a sprite. Is there any way the size of this window can be "saved" so I don't need to do that every time? See the image below to see what I mean.

image

Originally from here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=59607.0

ivan-mogilko:
Fix #1484

* Adds IObjectConfig interface and ObjectConfigJson implementation, taking idea from existing ColorThemeJson class.
* GUIController provides WindowConfig property that returns IObjectConfig for anyone to use.
* WindowConfig is written to a file at "%LOCALAPPDATA%\AGS\WindowConfig.json". (That's the same place where Layout.xml is saved)
* SpriteChooser stores its window position and splitter position in WindowConfig.
* SpriteManager stores its splitter position.

Few things that may be improved here right away.
1. **[DONE]** Safety checks for the window position. This is a must, I suppose that window could clamp itself to the parent or desktop size, in case config is broken.
2. **[DONE]** Enhance window pos serialization. Right now it saves 4 properties explicitly, but normally there should be a method that does that, for uniformity (so that every window that wants that would use same way to save things). Either serializing Point and Size types (for Form.Location and Form.Size respectively), or serializing control's position. In the latter case we might introduce a IControlConfig interface, extending IObjectConfig and providing extra methods suitable for Controls and Forms.
3. **[DONE]** Also save splitter pos in SpriteManager, since it uses same "SpriteSelector" control. I tried to do this, but met couple of issues, one of which I could not yet solve properly. The problem is that, unlike regular dialog window, these panes have minimal size before they are resized by the docking manager, so loading config in OnLoad method fails to restore splitter position reliably. There has to be another method called after a EditorContentPanel gets shown resized on its tab. If it does not exist, then it may be added.
4. **[DONE]** Oh right, and - find if it's possible to make Newtonsoft.Json to save json file formatted for human reading, because right now it just dumps it as a single string, apparently.....


ericoporto:
> find if it's possible to make Newtonsoft.Json to save json file formatted for human reading

I think the JsonTextWriter should have a Formatting property that can be set to `Formatting.Indented` for this.

ivan-mogilko:
Okay, this looks more interesting now.

Added saving of a splitter in SpriteManager, for consistency, and also to experiment saving & restoring config for EditorContentPane classes, which have some quirks in how they are opened and closed (details are in commits).

ivan-mogilko:
Well, I suppose i'll merge this since no one comments.
The new file is entirely optional, so this may be redone or removed without breaking editor.

PR #2514, opened 2024-08-26
Engine: implement "accessibility" config for speech and text skip

ivan-mogilko:
Per a *very old* users request.

This allows players to override game's speech skip style to their preference if the game does not provide such option.

Adds two config options under "[access]" group:
```
[access]
speechskip = <none, input, any, time>
textskip = <none, input, any, time>
```

Both may take following values:
* "none", or empty string (default) - this actually means that the option is not used
* "input", meaning any player input
* "any", meaning any player input, or time
* "time", meaning time only

These options override in-game settings for skipping speech and displayed messages respectively.

The override is done:
- when the game is loaded
- when the save is restored (because these options may be saved)
- when these options are changed by a script, these overriding settings may prevent assignment of a new value.

Note that reading a Speech.SkipStyle property in script will actually return a true overridden value. This is done in case game script uses this property as a reference when implementing custom speech, for example.

### DIFF
--- a/Common/util/string_utils.h
+++ b/Common/util/string_utils.h
@@ -95,25 +95,47 @@ namespace StrUtil
 
 
     // Parses enum value by name, using provided C-string array,
-    // where strings are compared as case-insensitive; returns def_val if failed;
+    // where strings are compared as case-insensitive; returns def_val if failed
     template<typename T, std::size_t SIZE>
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
Alright, I'll merge this, although I never got a response from a user who asked for this recently.
But this works, and might work as a workaround.

Ideally these settings should be in winsetup, I think. But winsetup is badly designed (ui-wise), and should be either transformed into multi-paged window with tabs (short term), or completely rewritten into something cross-platform (long term).

PR #2515, opened 2024-08-27
Tool: agfexport - export miscellaneous data from the game source

ericoporto:
@ivan-mogilko , ages ago you made the pr #1375, with agfexport but then decided to not go forward.

I am making here a tentative idea, I picked your code from that and from the glvar and autoash tools and put together a new tool idea, that makes the three tools into one, using subcommands.

agfexport - parses the game project file (*.agf), and exports various stuff depending on command.

Usage: `agfexport <COMMAND> [<OPTIONS>] <input-game.agf> <out-file>`
Commands:
  `autoash <file>`         Generate auto script header
  `glvar <header> <body>`  Generate global variables scripts
  `script-list`            Exports ordered list of script modules
  `room-list`              Exports list of active rooms
  `-h, --help`             Show help message

### DIFF
--- a/Solutions/Tools.App/agfexport.vcxproj
+++ b/Solutions/Tools.App/agfexport.vcxproj
@@ -1,0 +1,159 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <ItemGroup>
+    <ClCompile Include="..\..\Common\debug\debugmanager.cpp" />
+    <ClCompile Include="..\..\Common\util\bufferedstream.cpp" />
+    <ClCompile Include="..\..\Common\util\cmdlineopts.cpp" />
+    <ClCompile Include="..\..\Common\util\file.cpp" />
+    <ClCompile Include="..\..\Common\util\filestream.cpp" />
+    <ClCompile Include="..\..\Common\util\path.cpp" />
+    <ClCompile Include="..\..\Common\util\stdio_compat.c" />
+    <ClCompile Include="..\..\Common\util\stream.cpp" />
+    <ClCompile Include="..\..\Common\util\string.cpp" />
+    <ClCompile Include="..\..\Common\util\string_compat.c" />
+    <ClCompile Include="..\..\Common\util\string_utils.cpp" />
+    <ClCompile Include="..\..\Common\util\textstreamreader.cpp" />
+    <ClCompile Include="..\..\libsrc\tinyxml2\tinyxml2.cpp" />
+    <ClCompile Include="..\..\Tools\agfexport\main.cpp" />
+    <ClCompile Include="..\..\Tools\data\agfreader.cpp" />
+    <ClCompile Include="..\..\Tools\data\scriptgen.cpp" />
+  </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="..\..\libsrc\tinyxml2\tinyxml2.h" />
+    <ClInclude Include="..\..\Tools\data\agfreader.h" />
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{776E38BA-D6A4-431E-8053-299310D85301}</ProjectGuid>
+    <RootNamespace>MakeRoomHeader</RootNamespace>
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
I agree that this is a good idea. Will test this today.

Another thing that may be planned for the future is exporting dialog scripts, but that would require decoupling dialog converter from agf2dlgasc into something that inputs a dialog script from a file, and possibly establish an extension for dialog scripts (that is #1008).

ericoporto:
I haven't yet deleted autoash and glvar as standalone tools but they may be removed later.

I can later look into decoupling agf2dlasc into some dl2asd command and a asd2asc converter but I would prefer to work on this in a separate PR.

I have some work in a tool that can use the existing tools to generate a building script that can be run to build a game too.

I also don't know if the agf2dta tool should be standalone or go here too - I think I can make a simple "library"/utils file and then just use the function in a simple command here in agfexport. I have some sketch of it.

ivan-mogilko:
> I also don't know if the agf2dta tool should be standalone or go here too - I think I can make a simple "library"/utils file and then just use the function in a simple command here in agfexport. I have some sketch of it.

`agf2dta` (i assume this created a main game data file) should be a separate tool, because it does not export, but will have to do a lot of job, converting all game classes and compiling a data file out of the source.

The game data writing functions may be placed close to the already existing data reading functions in respective files (such as main_game_file.cpp)

But besides the tool itself, there will be a question of dependency, because right now using existing classes to store intermediate data after conversion will bring Allegro 4 in, as some may have fields of types like Bitmap or RGB, or methods that reference Bitmaps, and others.

Which means that either extra work has to be done on decoupling this dependency, or duplicate classes and/or functions written.

ericoporto:
I think I can remove dependency on allegro on imagefile without much work, I will try this later and see where I get to, I noticed that bitmapdata is already decoupled.

ericoporto:
I don't know if it's the same problem but on AGS Script I have a struct that is data only which I use then for serialization and another struct that extends it (in a different script header) that has the methods that are used. Would it be possible to use a similar approach here?

ivan-mogilko:
> I don't know if it's the same problem but on AGS Script I have a struct that is data only which I use then for serialization and another struct that extends it (in a different script header) that has the methods that are used. Would it be possible to use a similar approach here?

This was sort of the plan, yes. Something like proposed in "Proposal" section here #2058.

ivan-mogilko:
Anyway, in regards to this PR, I tested it, and overall this works, and is a simple program.
The issues I found are mostly cosmetic and organizational.

1. Code style is mismatching our defined style in some places, i.e.
```
if ( ) {
}
```
vs
```
if ( )
{
}
```
(but also mixing both together sometimes)

2. The help calls this tool:
> AGS game's miscellaneous export tool

I find this may be misleading, or rather not specific enough. "Game" may refer to compiled game too. I think there should be a disambiguation. Perhaps "game project export", or similar.

3. There seem to be a common phrasing mistake (maybe this classifies as a grammatical mistake) in the command descriptions, where "to" precedes "what/from", like:
> This will write to <file> the auto-generated script header from <input-game.agf>

or
> This will print to the command line the list of scripts

In English it's correct to put this other way around, e.g.:
> This will write the auto-generated script header to a file
> This will print the list of scripts to the command line

5. 2 commands out of 4 are printing to stdout by default and require extra option to write into a file. This is inconsistent with the rest of commands, and with the "export" nature of this tool. Is this intentional, and what is the purpose of such decision?

6. The "room-list" command prints a pair "roomN.crm:Description" for some reason, instead of just list of files. Is that intentional? What is the exact purpose of this command, how is it supposed to be used?
Should not it also print room script filenames? (These are not included into "script-list")
On another note, I think the room list should be sorted, for convenience (the AGF::ReadRoomList probably does not do that itself).

ericoporto:
I think I managed to fix all mentioned issues.

ivan-mogilko:
I don't quite understand the use case for exporting separate script and script header files. What is the purpose of these actions?

If one builds the game, they would need following:
- list of scripts
- list of headers to include in each of these scripts.

ericoporto:
Yeah, exactly as you mentioned, you need the headers list as you essentially concatenate them all and then you just build the scripts, not necessarily in order.

ivan-mogilko:
I think my question is rather, why have these as separate lists, and not a single list with only script names?

When building we either must assume that each script has a respective header of the same name, and then this list of headers can be generated from a list of script names preceding or equal to current script. Or have explicit "include list" per a script which may contain anything.

So, having 2 separate full lists right now seems redundant.

ericoporto:
Nah, don't want to do that, filenames are not case sensitive and module names are, there's enough to go wrong there. They are separate properties in the XML, they have to be separate commands.

ivan-mogilko:
I still don't understand, but alright.

ericoporto:
Headers list has to somehow make into here: https://github.com/adventuregamestudio/ags/blob/6ce71b446cb7a561dc4fbcec57f71a29547c5b1e/Compiler/main.cpp#L30

The script list goes to your build system

ivan-mogilko:
I think this tool needs to be exposed further and also export:
- fonts used by the game
- plugins used by the game
- translations? but probably not, since these may be considered standalone (even though they are mentioned in the project tree)

ivan-mogilko:
So, I was confused earlier, but I realize my doubts now.

There are separate tasks, which require different kind of lists.

1. Compiling scripts - requires a list of scripts, *including room scripts*. This will allow to have a single list of scripts to pass into the script compiler.
2. Packaging a game "library" file - requires a list of assets, but does not need to know their distinct types. It only need filenames. So my original idea was that such tool should just give a list of files and say "these should be packaged".
3. Gathering a full distributive. This would require a list of plugins, etc.

ericoporto:
Perhaps files for packaging can be exported as a list of files for packaging. Note that the agspak can already get such list since it can interpret it as an include file pattern.

Note some files are not specified in agf but are to be packed - they are somewhere in the build task in the Editor, things like ogv files at the root of the directory, speech files, preload.pcx and maybe others.

ericoporto:
But perhaps agspak can have an additional command where it gets a list of files and tries to package the exact list.

PR #2516, opened 2024-08-30
New compiler: Error message: Undeclared identifier

### LINKED ISSUE #1839: [AGS4] Wrong(not precise) error in an "if" statement with an undefined variable.
**Describe the bug**
```c
if(5 != myVariable)
```
There might be a bug in the compiler(?) or something which gives seemingly wrong error.
`Expected a term following the '!=', didn't find it`

Basically it can't find `myVariable` variable inside `if` statement, but the error doesn't specify that.

**AGS Version**
4.0(not latest)

**To Reproduce**
1. Create if statement with an operator `!=` where right hand side uses a variable that is not defined.
2. Run the project to see the error.
3. It seems to be the same for most other operators such as `==`, `<`, `>` etc.

**Expected behavior**
The error should be more consistent with similar message where left hand of the condition is not defined:
```c
if(myVariable != 1)
```
`Expected an expression, found 'myVariable' instead`

fernewelten:
Fixes #1839 

When an expression is expected and an undeclared symbol follows or comes instead, then this unknown symbol is most probably an identifier whose declaration has been forgotten. Complain with a hand-crafted error message.

Example:
```
function game_start()
{
    int i;
    while (i < Holzschuh)
    {}
}
```
→ Identifier 'Holzschuh' is undeclared

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -3890,6 +3890,14 @@ void AGS::Parser::ParseExpression(SrcList &src, EvaluationResult &eres)
     size_t const expr_start = src.GetCursor();
     SkipToEndOfExpression(src);
     SrcList expression = SrcList(src, expr_start, src.GetCursor() - expr_start);
+    Symbol potential_ident = src.PeekNext();
+    if (_sym.IsIdentifier(potential_ident) &&
+        !_sym.IsPredefined(potential_ident) &&
+        _sym.kNoSrcLocation == _sym.GetDeclared(potential_ident))
+    {
+        UserError("Identifier '%s' is undeclared", _sym.GetName(potential_ident).c_str());
+    }
+
     if (0u == expression.Length())
         UserError("Expected an expression, found '%s' instead", _sym.GetName(src.GetNext()).c_str());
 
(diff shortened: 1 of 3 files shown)


fernewelten:
Taking out the googletest 'Ternary6'; will reinstate that exact test when I release the ternary patch

ivan-mogilko:
> Taking out the googletest 'Ternary6'; will reinstate that exact test when I release the ternary patch

Could you please comment it rather than completely erasing the code?

ivan-mogilko:
>  I had hoped that this would make the PR pass the acceptance checks, but no luck with that: There are still some googletests that fail with some compiler runs. But as far as I can see, these googletests have nothing to do with the changes in this PR.

These tests are all run using gcc, as far as I know.

ericoporto:
Not all, the Windows 32 one runs on MSVC, you can see it's logs.

ivan-mogilko:
> Not all, the Windows 32 one runs on MSVC, you can see it's logs.

Hmm, I got a wrong impression seeing the mention of MSYS2, but now I see that it's disabled in the "Windows" run.

I have no such errors in MSVS, so wonder what could be the difference?

ericoporto:
@fernewelten , the issue was figured in https://github.com/adventuregamestudio/ags/issues/2518 and @ivan-mogilko put a fix in #2519 , after that is merged I think you can just rebase to pickup the fix.

ivan-mogilko:
I cherry picked and pushed the commit from this PR into the ags4 branch:
b598d5d7842d3af75afa5e27254521b39dfad24f

Will close this PR without merging now.

PR #2517, opened 2024-08-30
CI: gives more detailed error message on failure

ericoporto:


ivan-mogilko:
I assume it is safe to copy same change to the ags4 branch?

PR #2521, opened 2024-08-30
Script API: support eNullFont, a pseudo font ID which draws nothing

ivan-mogilko:
This is a bit of a exotic feature, it's meant to be used in a situation when you want to have an object with a valid text property, but does not really have that text drawn. One of the use cases is a button which text is used to form a sentence on a action label (MI-style), but you like to have a button image only and not the text.
Maybe there are others, like invisible speech, for instance, or else.

In script this adds `eNullFont` constant, as a part of `eFontType` enum.
Adds "Null Font" in a font selection for properties.

Codewise this is a tiny change, and does not seem to bring much with itself. Our font system already reacts safely on a "invalid" font number, reports 0 width, height, no outline, and not drawing anything with it when requested.
Works for GUI controls, and DrawingSurface.DrawString function.

NOTE: It's important to mention that this is different from "transparent" text, because the "no font" has no measurements, or rather all of them are 0. So anything that relies on the text's *render size* will not get any meaningful calculations.

### DIFF
--- a/Common/ac/common_defines.h
+++ b/Common/ac/common_defines.h
@@ -121,4 +121,7 @@
 // Animates once and stops, resetting to the very first frame
 #define ANIM_ONCERESET         2
 
+// An identifier of a "null font", a pseudo font used when you don't want a text to be drawn
+#define FONT_NULL (-1)
+
 #endif // __AC_DEFINES_H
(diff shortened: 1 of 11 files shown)


ericoporto:
> Codewise this is a tiny change, and does not seem to bring much with itself. Our font system already reacts safely on a "invalid" font number

We do this safely even before passing to a plugin like the sprite font one ?

ivan-mogilko:
> We do this safely even before passing to a plugin like the sprite font one ? (Haven't checked)

Yes, this check is done before using a registered Font in any way, including calling a font renderer. The plugins install their fonts in the valid slot numbers, and "null font" is outside this range, so plugins won't ever get called for this "font".

ericoporto:
I tested this and using it does make the text disappear without giving any errors.

PR #2528, opened 2024-09-09
New compiler: Revamp google tests

fernewelten:
Rewrite the googletests to call the compiler proper directly, i.e., to call `cc_compile(source_code, options, scrip, messages)` – instead of calling `cc_compile(source_code, scrip)` and passing the options and getting the error messages in a roundabout way. 

In that way, any global static variables are sidestepped from the get-go and so can't foul the googletests even when they are clobbered due to the googletests being run in parallel. (IIRC, that's what the Editor does, too, when it runs instances of the new compiler for all the ags files of a game in parallel instead of one after the other.)

Apart from the calling protocol and some slight variable renaming in some few tests to bring them in line to the other tests, the tests themselves remain unchanged. 

Add another googletest file and distribute the tests to use that additional file.

Typical old calling protocol:
```
ccSetOption(SCOPT_RTTIOPS, true);

int compileResult = cc_compile(inpl, scrip);
ASSERT_STREQ("Ok", (compileResult >= 0) ? "Ok" : last_seen_cc_error());
```

Typical new calling procotol:
```
FlagSet const options = SCOPT_RTTIOPS;

int compile_result = cc_compile(inpl, options, scrip, mh);
std::string err_msg = mh.GetError().Message;
ASSERT_STREQ("Ok", mh.HasError() ? err_msg.c_str() : "Ok");
```

(In both versions the `ASSERT` is written in such a way that any error message that the compiler emits gets sent to the googletest protocol so that it is immediately visible)


fernewelten:
The failed checks don't seem to be related to the googletests.
The new file `cc_parser_test_2.cpp` doesn't seem to be checked automatically yet.

ericoporto:
If is a test file you can add here in the CMakeLists.txt for the compiler and it will be picked by the compiler CMake project

https://github.com/adventuregamestudio/ags/blob/a34c7483a590a1c24af668351374c45c56079d44/Compiler/CMakeLists.txt#L88

PR #2529, opened 2024-09-09
New compiler: Initializer functions for `new`

fernewelten:
Please test this!
Please note that this PR relies on the PR #2528.

Managed structs can have an initializer: i.e, a struct function that must be named `initialize`.
- The initializer __must__ return `void` (or `int` so that it is possible to define it with the `function` keyword).
- It can have any parameters or none, but note that we don't have overloaded functions in AGS yet, so when you define a parameter list for an initializer then you must stick with it forever. 

When a struct has an initializer that the compiler knows about, the compiler will insert a call to this function directly after an object has been created with `new`. You __must__ give the parameters for the function call directly after `new XX`, the compiler will no longer allow you to create objects of that type without any parameter ist.

On the other hand, when a struct does __not__ have an initializer function, you can use `new` on it either with an empty parameter list or without any parameter list.

```
managed struct Stx
{
    int Payload;
    import function initialize(float f);
};

managed struct Sty
{
    short Payload;
};

function game_start()
{
    Stx *st1 = new Stx(7.0); // ← okay
    Stx *st2 = new Stx(7); // ← syntax error, parameters don't match
    Stx *st3 = new Stx(); // ← syntax error, parameter expected but not found
    Stx *st4 = new Stx; // ← syntax error, parameter list missing

    Sty *st5 = new Sty(); // ← empty list is okay, nothing will be called
    Sty *st6 = new Sty; // ← okay
    Sty *st7 = new Sty(4); // ←syntax error
}
```


ivan-mogilko:
This would be a great addition!, but I am bothered by the chosen hardcoded name `initialize`.

Would it work if a classic constructor-like syntax was used instead, where the method has the same name as a struct?

fernewelten:
> Would it work if a classic constructor-like syntax was used instead, where the method has the same name as a struct?

We could have it that way, too. 

Something along the lines of the following?
```
managed struct Stx
{
    int Payload;
    import function Stx(float f);
};
```

ivan-mogilko:
I also propose that constructor should be enforced to be `void`, because there's no way to receive its return value.

I'd even say that it should be typeless, but idk if ags script can handle that now.

fernewelten:
IIRC correctly, this started when someone (I think @ericoporto ) noted that the old compiler allowed the syntax `new Strct()` without doing anything with the pair of parens or the info in between.

So the first remedy was to make the new compiler accept this syntax, too, so that it could grok ‘old’ code. But I stuck a warning into it that this wasn't really implemented yet. 

At that time, I brought up the idea to have an initialize function that would be called with the parameters between the parens whenever a struct would have one.  We hadn't decided on anything definite yet, and as I said, the old compiler seemingly doesn't do anything whatsoever with this syntax.

fernewelten:
Edit: Here's the thread.
#2181

ivan-mogilko:
I tested this, and it works in principle.

But, as I mentioned previously, in my opinion, the "initializer" should rather comply to classic constructor syntax, where:
* it must have a name identical to the struct's name;
* it should have either strictly void type, or no declarable type at all;
* it should not be permitted to call this function directly from script (it is possible to call it currently).

fernewelten:
Okay, thanks for testing and thanks for the feedback. 

I'll take that as a “go” and implement the modifications you suggest, within this PR.

ivan-mogilko:
Note: I applied all the google tests commits from this PR directly to ags4 branch.

ivan-mogilko:
Merged the feature remake #2582

PR #2531, opened 2024-09-11
Switch to use full character's script name in voice-over clips

### LINKED ISSUE #1456: Consider expanding voice clip naming rules
**Problem**

Currently voice-over clip filenames are restricted to NNNNX.ext format, where NNNN is four first letters of character's script name *except the leading small 'c'* (this is because characters were often called `cNAME`) and X is a number in the range of positive 32-bit integer. Older engines supported only 4-digit numbers 1-9999, but that restriction was since removed. I believe this name format may be a legacy of DOS support (8.3 filenames).

The name rule may cause conflicts between characters with similar script names and usually forces gamedev to edit script names to make first 4 letters unique.

Limit of 9999 clips per character is reached less often, but sometimes being ran into. In such case users usually use dummy characters to increase number of slots.

Neither of the issues above is game breaking, but may cause much annoyance and inconvenience. So this task is not ultra-critical, yet something to consider.

**Suggestion**

Instead voice clip format should be "NAMEX.ext", where NAME is full character's script name.

NOTE: @messengerbag has suggested an alternative to this: introduce new Character's property to define voice-name, which is equal to its script name stripped of small `c` prefix by default. This will help if a user does want to keep the script naming convention which suggests start names with `c` for characters, in which case all the voice clip names would have to begin with redundant "C".
Keeping a rule to strip `c` prefix may be causing confusion if characters have all lowercase script names, as their names and this prefix are not enforced neither by engine nor tools.

ivan-mogilko:
Resolve #1456.

Change the rule of how voice-over clip names are generated from character script name:
* still cut the 'c' prefix, nothing is changed in this regard (since characters traditionally are still called `cEgo` in AGS)
* use full name instead of first 4 letters
* insert a dot '.' between the name and a clip number, in order to distinguish number from script names having digits.

In other words: if you have a character `cMyCharacter11` and have a 11th voice line, the filename will be:
`MyCharacter11.11.ext` (case insensitive).

I had to add a switch in "Backwards compatibility" section that toggles between old and new rule. This is for importing old projects, where users might not want to rename all the voice clips.


ericoporto:
Curious, should the Narrator also change to "Narrator" in the new style?

ivan-mogilko:
Ok, i did it, it seems consistent with the "narrator" used in dialog scripts.

> I am not completely sure though because from the manual I can't readily tell if this is still supported or if it should be the character assigned in game.narrator_speech.

Hardcoded "narrator" name is used when no character is set to game.narrator_speech.

ivan-mogilko:
> Does this new format also apply to sync files used by the built-in lip sync, so that voice clips and sync files use the same format (with different file extensions)?

It does apply, but it turns out that the lipsync has a hardcoded filename limit of 14 characters. I did not know about this, and nobody tested this apparently.

ivan-mogilko:
Actually, I am not sure which filename is that. I need to re-investigate how lipsync works.

ivan-mogilko:
@messengerbag do you have any game project with lipsync files that I could experiment with?

messengerbag:
I'll PM you.

Issue #996, opened 2019-12-01
Fonts: allow font instancing with different properties


ghost:
Fonts have attached properties that may be changed without altering the source font data itself. For instance: point size / scaling, or line spacing. But currently AGS requires you to create a new copy of a font file if you want to use it with different setting. While fonts files aren't large, this is still unnecessary copying of source data.

My proposal is to change this and treat AGS fonts as "instances" that may refer to same source file, and only separated in memory (even sharing same font data in memory when possible).

ericoporto:
Maybe fontDef or fontConfig class where you add this Settings and pass along with the font itself to be consumed by the functions that draws text and objects (eg. Labels) .

ivan-mogilko:
> Maybe fontDef or fontConfig class where you add this Settings and pass along with the font itself to be consumed by the functions that draws text and objects (eg. Labels) .

Um, no, I thought in API "font" becomes font instance. What changes is only how engine and editor handle source data internally.

ivan-mogilko:
I'd like to propose slightly different approach to this. Originally this ticket was suggesting "instancing fonts with the same source file". I want to suggest other way around: fonts freely link to the source file.

In other words: we have Font entity which is not strictly linked to the actual glyph data. Instead it has a "source file" property, which may be filled or not. Or changed at runtime.

This will allow to link multiple Fonts to the same asset, but also to not link a Font to any asset. The latter may make sense if you expect this font to be drawn by a plugin, or even have glyph data be created by a script.

What will this change immediately in engine and editor?

The engine should provide a dummy font renderer that does nothing, in case no source is linked.

The editor should change a way a font is "imported". When creating a new font it will be "empty". The user will have an option to link a source file to it, or unlink from a file.
But then there's a question on how to make importing font files convenient. Either there has to be an explicit command for that, or "font files" should have their own node in a project tree.

ericoporto:
In the project tree, the font we have could have two subnodes, one for files and the other for the instances.

[...]

PR #2545, opened 2024-10-05
AGS 4: Separated Fonts and FontFiles, let freely change Font's source file and size

ivan-mogilko:
For #996

This is a preliminary draft; it is functional but there's still a doubt about choice of item arrangement in the project tree, as mentioned in the comments to #996, so that's potentially a subject to change.

Summary:
- Separate FontFile and Font items, stored and added each on their own.
- Font item's SourceFilename, which previously pointed to the original imported file, now works as a reference to a FontFile.
- Font's SourceFilename and Font Size may be freely modified in the editor. When modified, font is reloaded using specified settings. SourceFilename is a dropdown list, updated whenever a list of FontFiles changes, and Font Size is edited by pressing "..." button to the right, on which it displays a dialog, because there are 2 ways to set TTF size (this part remains from older versions).
- It is possible to have a Font without a source filename, in which case it won't draw anything, unless substituted by a plugin.
- FontFile properties are not viewable at this point, I'd have to make more changes for this, and also make a preview work for them.
- Standard font name "agsfntX.*" is deprecated, font files now may have any name. When compiling the game, Editor will package only files mentioned in the project tree.

Screenshots:
ags4--newfonts1
ags4--newfonts2
ags4--newfonts3


ivan-mogilko:
I suppose it will also help if there was some visual cue for each different kind of node... different icon, or text font or style used...

ivan-mogilko:
@AlanDrake could you please post the new icons that you made previously here, for easier reference?

AlanDrake:
immagine

[newfonts.zip](https://github.com/user-attachments/files/17299027/newfonts.zip)

ivan-mogilko:
I think this is ready for review.

Other things that were mentioned in #996 were calling font items "Font Styles" and moving font files into "Fonts" subfolder on disk.

Editor/AGS.Editor/Components/FontsComponent.cs
```diff
@@ -40,71 +53,63 @@ public override string ComponentID
...
         }
 
+        private FontFile FindFontFileByName(string fontFileName)
+        {
+            return _agsEditor.CurrentGame.FontFiles.Where(ff => ff.FileName == fontFileName).FirstOrDefault();
+        }
+
+        private void ShowOrAddPane(FontFile chosenFontFile)
+        {
+
```
ericoporto, line 110: If this function is meant to be empty but needs to exist, I suggest adding a comment here, something like "panes only exist for Font for now, but we may call this method for a node that may be a FontFile, just do nothing".

ericoporto:
I am not a super fan of not renaming the font files, I see a few `EndsWith(".ttf")` and similar, these all need to be case insensitive if we are keeping the original font name.

ivan-mogilko:
> I am not a super fan of not renaming the font files, I see a few `EndsWith(".ttf")` and similar, these all need to be case insensitive if we are keeping the original font name. If we renamed on import then these would be guaranteed to work in case sensitive.

Please elaborate, are there other arguments against not keeping original name, or just that I used incorrect comparison function in code? When you think that they should be renamed, do you mean only the letter case, or you think that font files should keep old names "agsfntN.ext"?

Personally, I try to assume that the game may be constructed by gathering the files by hand, without use of editor.

ivan-mogilko:
> I see a few `EndsWith(".ttf")` and similar, these all need to be case insensitive if we are keeping the original font name. If we renamed on import then these would be guaranteed to work in case sensitive.

BTW, I searched for `EndsWith(".` and there are numerous cases like this throughout the Editor code.
I suppose that we need to replace these with a helper function, like TestExtension, that compares case insensitive.

ivan-mogilko:
Per @AlanDrake 's suggestion (posted elsewhere, maybe on Discord?), changed font file location to a "Fonts" subdirectory.

Fonts are copied over on project upgrade, and Editor even makes a folder backup if it existed previously, complying to how RoomComponent does this when upgrading rooms.

I had to modify "--runfromide" engine's command, because fonts are not packed when running a test game, editor must tell an extra directory to the engine. Existing arg list for "runfromide" is rather inconvenient, with hardcoded meaning for each entry. So I modified it to be a custom list of dirs paired with "asset filter" list (asset filters are used to narrow search of certain asset types).

PR #2547, opened 2024-10-09
GUI Buttons can wrap & pad text, GUI Labels full alignment

### LINKED ISSUE #1876: GUI Buttons Text Offsets
**Describe the problem**
I would like to be able to set the offset, in pixels, for text on GUI Buttons. I find the current options of TopCenter, MiddleCenter, etc. too limiting. The button style that I would like to use just does not look good with the current settings.
This is incovenient and ugly. I could do an image for each button, but then I can't use the same in-game font for my dialogs, since the graphics applications don't understand AGS fonts.

**Suggested change**
I would like to be able to set the offset, in pixels, for text on GUI Buttons. E.g. 10 pixels from the left and 3 pixels from the top.

### LINKED ISSUE #2546: Word wrap GUI Button texts
**Describe the problem**
This was originally reported in the [forums here](https://www.adventuregamestudio.co.uk/forums/editor-development/feature-request-word-wrap-for-buttons/msg636666119/#new)

> Currently, the text on buttons doesn't word wrap. If I want to wordwrap text on a button, I have to use a text label and place it on top of the button. Normally, no big deal. But this can be a bit annoying when I want to add a lot of buttons like this:
> 
image
> 
> For each of those buttons, I had to create individual text labels and place them on top of each button. Very time consuming and a bit of a faff to keep track of which label is on top of which button. If the text on the button could word-wrap, that would save a lot of time and effort.

GUI Button text doesn't wrap, the workaround is to use Labels on top of buttons but this requires tracking which label belongs to which button and keep this in mind when moving them.

**Suggested change**
So the idea is to add support for word wrapping. A few notes
- Button MACRO supports a few things that label doesn't, like `(INV)` for showing current active inventory
- If we want a similar word wrapping as Labels, I wonder if there's some way to share the code (for drawing and text size calculation)
- Backwards compatibility: could have some property to turn the behavior on or off, and just default it to false and either expose it as a property or automatically default it to true if above a certain version (but I wonder if the text calculation may make things differently enough to be better to have this as a property)

ivan-mogilko:
Resolves #1876
Resolves #2546

1. Added a GUIControl flag kGUICtrl_WrapText that may be applied to any text-based control, in theory.
GUIButton supports this flag, and in case it is set:
 - splits text into lines;
 - aligns the text block vertically using vertical part of text alignment;
 - aligns each text line horizontally using horizontal part of text alignment;

Wrapping and drawing code is mostly shared with the GUILabel class (picked out as a separate commit).

Added WrapText property to a Button in the Editor.

2. Added TextHorizontalPadding and TextVerticalPadding properties to GUIButton. These specify left-right and top-bottom space between button's frame and aligned text.

Button placeholder text supposedly should not be affected by this.
This has to be tested more.

3. GUILabels support full Alignment (both horizontal and vertical).

Possible TODOs:
* Add new properties to script API.

### DIFF
--- a/Engine/ac/button.cpp
+++ b/Engine/ac/button.cpp
@@ -308,6 +308,48 @@ void Button_SetTextAlignment(GUIButton *butt, int align)
     }
 }
 
+bool Button_GetWrapText(GUIButton *butt)
+{
+    return butt->IsWrapText();
+}
+
+void Button_SetWrapText(GUIButton *butt, bool wrap)
+{
+    if (butt->IsWrapText() != wrap)
+    {
+        butt->SetWrapText(wrap);
+        butt->MarkChanged();
+    }
+}
+
+int Button_GetTextPaddingHorizontal(GUIButton *butt)
+{
+    return butt->TextPaddingHor;
+}
+
+void Button_SetTextPaddingHorizontal(GUIButton *butt, int pad)
+{
+    if (butt->TextPaddingHor != pad)
+    {
+        butt->TextPaddingHor = pad;
+        butt->MarkChanged();
+    }
+}
+
+int Button_GetTextPaddingVertical(GUIButton *butt)
+{
+    return butt->TextPaddingVer;
+}
+
+void Button_SetTextPaddingVertical(GUIButton *butt, int pad)
+{
+    if (butt->TextPaddingVer != pad)
+    {
+        butt->TextPaddingVer = pad;
+        butt->MarkChanged();
+    }
+}
+
 //=============================================================================
 //
 // Script API Functions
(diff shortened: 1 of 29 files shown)


ericoporto:
This appears to work. The `(INV)` macro also still works. I haven't tested many fonts here though.

I noticed that the comment of the text property is "Whether button will wrap long text", which is true but it's more than this, if this property is false the text is considered to be a single line and if this is true it will wrap long text but also take into account new line characters (`\n` and `[`), so I would amend that comment to "Whether button will wrap long text and accepts new line character" or just "When true, wraps text when it's long or has new line characters".

ivan-mogilko:
I had a suspicion that after adding wrapped text, users may also ask for extra padding to better accomodate this text on a button. And we already have #1876.

So, I decided to also add two more properties called TextHorizontalPadding and TextVerticalPadding.

ivan-mogilko:
Added support for full text Alignment for Labels. This seems trivial to make at this point: as the calculating and drawing code is shared with button text, and HorizontalAlignment enum constants have values equal to the matching FrameAlignment enum.
(i.e. Left == TopLeft, and so on).

ericoporto:
Uhm, I am getting an error when loading a save now

> An internal error has occurred. Please note down the following information.
> If the problem persists, contact the game author for support or post these details on the AGS Technical Forum.
> (Engine version 3.6.2.1)
> 
> Error: Unable to restore the saved game.
> Failed to restore the savegame component.
> (`#4`) GUI, version 3060200, at offset 17661.
> Inconsistent format, or file is corrupted.
> Mismatching tag: GUILabels.

<details>

image

</details>

Engine/ac/button.cpp
```diff
@@ -477,6 +549,14 @@ void RegisterButtonAPI()
         { "Button::SetText^1",            API_FN_PAIR(Button_SetText) },
         { "Button::get_TextAlignment",    API_FN_PAIR(Button_GetTextAlignment) },
         { "Button::set_TextAlignment",    API_FN_PAIR(Button_SetTextAlignment) },
+        { "Button::get_WrapText",         API_FN_PAIR(Button_GetWrapText) },
+        { "Button::set_WrapText",         API_FN_PAIR(Button_SetWrapText) },
+        { "Button::get_TextHorizontalPadding", API_FN_PAIR(Button_GetTextHorizontalPadding) },
+        { "Button::set_TextHorizontalPadding", API_FN_PAIR(Button_SetTextHorizontalPadding) },
+        { "Button::get_TextVerticalPadding", API_FN_PAIR(Button_GetTextVerticalPadding) },
+        { "Button::set_TextVerticalPadding", API_FN_PAIR(Button_SetTextVerticalPadding) },
+        { "Button::get_WrapText",         API_FN_PAIR(Button_GetWrapText) },
```
ericoporto, line 558: I thin these two `"Button::get_WrapText"` and `"Button::set_WrapText"` are duplicate?

ericoporto:
Well, as far as I can tell, it all works now, pretty cool, saving and loading and setting/getting properties appear to work too. Not too much sure of what else to test now, but it looks fine for merge. There is just an additional comment I added above in the `RegisterButtonAPI` function

ivan-mogilko:
Would it be better if properties were called `TextPaddingHorizontal` and `TextPaddingVertical`? I think that has an advantage of having "TextPadding" first in both cases which keeps them together name-wise, and in autocomplete list.

ericoporto:
> Would it be better if properties were called TextPaddingHorizontal and TextPaddingVertical? I think that has an advantage of having "TextPadding" first in both cases which keeps them together name-wise, and in autocomplete list.

Yeah, that groups better for typing.

> Also, in hypothetical case we add more paddings, or replace this with a single property TextPadding which contains a struct, the new names will be easier to find.

Right, let's say we have TextPaddingLeft, TextPaddingRight, TextPaddingTop, and TextPaddingBottom, setting TextPaddingHorizontal internally sets both Left and Right, and TextPaddingVertical sets both Top and Bottom. That works, but it can't be read back I think? So they wouldn't be properties if we intend to read them - or they give always say the value of left or top. I think the struct would have the same considerations as direct properties.

I had an idea for a property in the editor that takes comma separated values, so perhaps if it was a single property in the editor it could be like, one number means all text padding is equal, two numbers mean horizontal and vertical and four numbers mean the 4 directions. ([this is from CSS rules, it has even a case for 3 numbers!](https://developer.mozilla.org/en-US/docs/Web/CSS/padding))

Other approach would be something like WalkSpeed, with an additional property to set if the padding is uniform, symmetrical or "four sides" (I lack a word for the 4 dirs), and then setting this would make the proper entries appear. There may be other ways too.

PR #2548, opened 2024-10-10
Common: Fix lzwexpand on strict-alignment platforms

dwatteau:
Hi!

I'm upstreaming this after ScummVM's [Trac#15179](https://bugs.scummvm.org/ticket/15179) and related scummvm/scummvm#6185 PR of mine.

The following commit fixes Maniac Mansion Deluxe on mips64el (OpenBSD/loongson, more precisely) where a fatal SIGBUS was triggered because of an unaligned load in `lzwexpand()`.

`memcpy()` is guaranteed to be safe, for this.

The unaligned load was also reported by UBSan (see the Trac ticket above for logs, if necessary).

(This may need to be cherry-picked for your ags4 branch as well.)

### DIFF
--- a/Common/util/lzw.cpp
+++ b/Common/util/lzw.cpp
@@ -18,6 +18,7 @@
 #include "util/lzw.h"
 #include <stdlib.h>
 #include "util/bbop.h"
+#include "util/memory.h"
 #include "util/stream.h"
 
 using namespace AGS::Common;
@@ -213,7 +214,7 @@ bool lzwexpand(const uint8_t *src, size_t src_sz, uint8_t *dst, size_t dst_sz)
           break;
 
         short jshort = 0;
-        jshort = BBOp::Int16FromLE(*(reinterpret_cast<const int16_t*>(src_ptr)));
+        jshort = Memory::ReadInt16LE(src_ptr);
         src_ptr += sizeof(int16_t);
         j = jshort;
 


ivan-mogilko:
Hello, I'd prefer to keep such workarounds hidden in memory helper functions, and under `MEMORY_STRICT_ALIGNMENT` flag.
We supposedly have this handled in utils/memory.h:
https://github.com/adventuregamestudio/ags/blob/master/Common/util/memory.h

It looks like BBOp::Int16FromLE() may be replaced by Memory::ReadInt16LE() which accepts void* pointer, calls ReadInt16 internally, where it does special handling under MEMORY_STRICT_ALIGNMENT condition.

Could you try if the following code works?
```
jshort = Memory::ReadInt16LE(src_ptr);
```

dwatteau:
> Hello, I'd prefer to keep such solutions hidden in memory helper functions, and under `MEMORY_STRICT_ALIGNMENT` flag. You may find these here: https://github.com/adventuregamestudio/ags/blob/master/Common/util/memory.h
> 
> It looks like BBOp::Int16FromLE() may be replaced by Memory::ReadInt16LE() which accepts void* pointer, calls ReadInt16 internally, where it does special handling under MEMORY_STRICT_ALIGNMENT condition.
> 
> Could you try if the following code works instead?
> 
> ```
> jshort = Memory::ReadInt16LE(src_ptr);
> ```

Ah, thanks for this pointer, your suggestion works as well, yes. I've updated the PR.

ivan-mogilko:
@dwatteau you have to add "#include "util/memory.h"".

dwatteau:
> @dwatteau you have to add "#include "util/memory.h"".

Oops, sorry for that, since the patch is very small I just based it on its ScummVM equivalent. But ScummVM rearranges the header dependencies a bit, so it didn't cause a build failure there.

Thanks! I'll do a local build the next time ;)

PR #2552, opened 2024-10-18
AGS 4: remake ExtendedCompiler setting into ScriptCompiler selection (2)

ivan-mogilko:
A (hopefully) better variant of #2407.

1. Create a IScriptCompiler interface, which lets Editor to communicate with any script compiler instead of calling random NativeProxy methods. This interface provides info about compiler's name, humand-readable description, some properties (like if it can do multiple parallel works), and finally CompileScript method.
**IMPORTANT NOTE:** I had to make this interface inside AGS.Native for the time being (this is still a managed interface though, so Editor can use it directly). There's already a IScriptCompiler interface in AGS.CScript.Compiler managed library, but it's connected to some other code there. Although I don't think it was ever used for real (the managed script compiler looks incomplete), I decided to not bother hacking this at this time. I left an extensive TODO comment above both new and existing interfaces with some thoughts about how they may be merged and used in the future.

2. Added 2 IScriptCompiler implementations: AGS3ScriptCompiler and AGS4ScriptCompiler. These are simple wrappers that hide calls to respective libraries.
**NOTE:** IScriptCompiler's implementation does not have to contain actual compiler inside. It might as well wrap a run of standalone executable, in case we use standalone compilers. I did not yet figure out how to receive compilation result in such case. Possibly the interface will have to get adjusted for that; and that's another reason to keep it inside AGS.Native for now, until we figure that out.

3. Editor no longer calls NativeProxy.CompileScript() and other methods, instead it calls a new NativeProxy.GetEmbeddedScriptCompilers() that returns a collection of IScriptCompiler interfaces. It then can choose any of those and use directly.

4. Finally, replaced "Use extended compiler" property in General Settings to "Script Compiler" with a list of compilers to choose from. This list is filled from GetEmbeddedScriptCompilers(). The only annoying caveat there is that I had to hardcode "default" compiler names second time in the AGSEditor in order to setup this property on project upgrade. But other than that, AGSEditor always uses selected compiler's name to get desired compiler and use it.


ericoporto:
I see you are modifying things in the compiler command line front end, a note is I think I haven't fixed it properly for the new compiler running there to properly get the headers like it's possible when running the front end with the new compiler - I was trying to think a good interface to put there because when I wrote the code the first time for the front end long ago I thought that each compiler - the new and the old - would at some point have their own command line interface instead of trying to write code that works for both. So it has some ugly ifdefs there and all.

ivan-mogilko:
> I see you are modifying things in the compiler command line front end

The only thing I did there was commenting out one compiler option (SCOPT_SHOWWARNINGS).
I also noticed that it does not build with MSVS solution in AGS 4, because new compiler lib is not added to the project, but LEGACY_COMPILER flag is not set.

ericoporto:
Just to confirm it, the idea here is old games build with the old compiler after upgrade and newer games build with the new compiler if we make the Templates have the new compiler set as default? (I see no default in `Settings.cs`)

Is there any macro that is generated to differentiate between compilers?

The old compiler has new features too - it has the RTTI stuff added which allows the managed struct in managed struct - so it isn't exactly the ags3 compiler. Would the name `AGS SCOM Legacy` be too weird to use in the dropdown box?

AlanDrake:
But how long are we gonna keep the old compiler?
I thought we'd remove it once the new one felt good enough.

ericoporto:
I think having the old is nice in terms of checking and comparing things - I usually at least try to compare performance, mostly to pickup if a regression has happened. 

The new one is assumed as the one people actually will use since any script API that returns arrays assume people will use the Length property.

ivan-mogilko:
> But how long are we gonna keep the old compiler?

Good question, I've been thinking about removing it, but maybe keep at least until there's a first "release candidate" for AGS 4.
At the moment old compiler helps in two cases:
- if new one gets broken after some update
- if you load old project, and dont have time to fix script for the new compiler right away (new compiler is more strict)

Other than that, I still think that a compiler *selection* is not a completely redundant thing, it may be useful in some special cases. Like if we experiment with stand-alone tools, or if somebody writes their custom compiler.

> Just to confirm it, the idea here is old games build with the old compiler after upgrade and newer games build with the new compiler if we make the Templates have the new compiler set as default? (I see no default in Settings.cs)

On project upgrade it is supposed to select old compiler for imported 3.6.* projects, and earlier 4.0 projects which had ExtendedCompiler = false. New compiler for 4.0 projects which had ExtendedCompiler = true.
I noticed that my upgrade code is a little wrong in this regard though, so I'm going to correct it.
As for default value, I cannot set any default in Settings class, because AGS.Types has no idea about compiler names: these are gathered by checking available compilers, and that's what Editor app does. I think that I will make it set first available compiler in case selection is empty, and the list of compilers is gathered in order of priority, new compiler first, so that will be default in such case.

> Is there any macro that is generated to differentiate between compilers? (this is more a question so I can make modules that work with both ags4 compilers)

Each compiler returns a list of "extensions", and these extensions are declared as macros:
https://github.com/adventuregamestudio/ags/blob/ef95f57405627e87195ea517406323fae492c1d1/Editor/AGS.Editor/AGSEditor.cs#L848-L855

Macro name is generated as `SCRIPT_EXT_<NAME>`.

Extensions are declared inside each compiler, old compiler:
https://github.com/adventuregamestudio/ags/blob/ef95f57405627e87195ea517406323fae492c1d1/Compiler/script/cs_compiler.cpp#L31-L39
new compiler:
https://github.com/adventuregamestudio/ags/blob/ef95f57405627e87195ea517406323fae492c1d1/Compiler/script2/cs_compiler.cpp#L29-L37

More could be added if necessary.

> The old compiler has new features too - it has the RTTI stuff added which allows the managed struct in managed struct - so it isn't exactly the ags3 compiler. Would the name `AGS SCOM Legacy` be too weird to use in the dropdown box?

Umm... "legacy" sounds bit off. I've been thinking, maybe I could title it "3+" or "3 EXT" (like, v3 extended)?

ericoporto:
3 EXT sounds nice, that would be a good name!

PR #2554, opened 2024-10-21
Script API: implement File.ReadBytes, WriteBytes

ivan-mogilko:
I wonder if this may be useful?

Adds File.ReadBytes and FileWriteBytes that accept a dynamic `char[]` array and index,count positions:

```
/// Reads up to "count" number of bytes and stores them in a provided dynamic array, starting with index. Returns actual number of read bytes.
int File.ReadBytes(char bytes[], int index, int count);
/// Writes up to "count" number of bytes from the provided dynamic array, starting with index. Returns actual number of written bytes.
int File.WriteBytes(char bytes[], int index, int count);
```

The use case is writing and reading long sequences of byte values, instead of using ReadRawChar/WriteRawChar in a loop.

Simple test example:
```
  char bytes[] = new char[100];
  for (int i = 0; i < 100; i++)
    bytes[i] = i;
  int wrote_bytes = f.WriteBytes(bytes, 40, 20);
  System.Log(eLogDebug, "wrote_bytes = %d", wrote_bytes);

  char bytes2[] = new char[100];
  int read_bytes = f.ReadBytes(bytes2, 0, 20);
  System.Log(eLogDebug, "read_bytes = %d, %s", read_bytes, bytes2);
  // Note that %s works here because char[] array is implicitly cast as `const char*`
  // and byte values are printable characters in this example
```

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -1347,6 +1347,10 @@ builtin managed struct File {
   import float  ReadRawFloat();
   /// Writes a raw 32-bit float to the file.
   import void   WriteRawFloat(float value);
+  /// Reads up to "count" number of bytes and stores them in a provided dynamic array, starting with certain index. Returns actual number of read bytes.
+  import int    ReadBytes(char bytes[], int index, int count);
+  /// Writes up to "count" number of bytes from the provided dynamic array, starting with certain index. Returns actual number of written bytes.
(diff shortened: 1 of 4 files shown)


ericoporto:
I have zero uses for this right now but it looks pretty neat. Overall things in the API that can be sped up by either taking an array or returning an array are a good idea I think.

Only thing from looking at the code is, if count is bigger than array size, perhaps it could also log a warn.

PR #2555, opened 2024-10-22
Fix nested `#else` in a script preprocessor

ivan-mogilko:
I did a stupid mistake in #2178 , missing the nested `#if/#else` possibility. As a result, a nested `#else` inside a disabled parent block enables the following code, even though the parent is still disabled. E.g. code like:

```
#define DEFINEME

#ifdef DEFINEME
  import void Func(int a);
#else
#ifdef DEFINEME_NESTED
  import void Func(int a, int b);
#else
  import void Func(int a, int b, int c);
#endif // DEFINEME_NESTED
#endif // DEFINEME
```

In the above example `import void Func(int a, int b, int c);` will suddenly get enabled.

This PR fixes this problem.

### DIFF
--- a/Compiler/preproc/preprocessor.cpp
+++ b/Compiler/preproc/preprocessor.cpp
@@ -130,7 +130,7 @@ namespace Preprocessor {
 
         bool includeCodeBlock = true;
 
-        if ((!_conditionalStatements.empty()) && !_conditionalStatements.top())
+        if (_negativeCounter > 0)
         {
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
> Does this also apply to #elif ?

Elif was not implemented in AGS preprocessor.

ivan-mogilko:
Added preprocessor tests for nested `#ifdef`s

PR #2556, opened 2024-10-25
Editor: test missing event handlers as a post-compilation step

### LINKED ISSUE #1363: Editor: Check event name mismatch at compile time instead of runtime
**Describe the problem**
When you edit a room,  you need to configure what function will handle an event such as "Walks off bottom edge", or "Examine object".  If you configure a function that isn't actually defined in the room file, this will only be detected at runtime.

**Suggested change**
This should be already detected after the project has compiled without programming errors. 

- Compiling entails compiling the rooms. When a room is compiled, the Editor knows what events are configured for the room proper and its hotspots, and objects.  One of the results of a compiling run is the list of exports. So the editor can check whether all the event function names are in the exports list and error out when they aren't.

- In the same vein, character and inventory item events must have functions in the GlobalScript. So the editor can check whether the compiled GlobalScript exports the functions that are configured for character and inventory item events.

ivan-mogilko:
Resolves #1363

This PR's additions may be separated in 3 parts:

1. Improved function searching when pressing on "..." button on events pane. I found that the existing search was not fully reliable, as it used direct string search of a sequence `function functionName(`, and thus could break by simply an extra space. I changed that to using regex pattern, which finds both `function` and `void`, and handles whitespace. This keeps working even if you split function header on multiple lines, like:
```
void
cEgo_Lookat
(Character* c, CursorMode mode)
{
```
In addition, I tried to make it skip found matches if they are inside comments.
I also considered testing preprocessor conditions too, but on second thought decided to leave that be, as that would become tough to do without a full script's preprocessing, which seemed like an overkill for "..." button in events (I'd have to cache results and run preprocessing in background thread, or something...).

2. If a function's name is inserted into event property, but that function is not actually exist in script, then pressing "..." will create that function anyway.

3. Test missing event functions as a script compilation post-step. For this editor keeps a dictionary of preprocessed script texts at compilation stage, so that components could searching for regex patterns there and find matching functions.


ivan-mogilko:
Hmm, maybe I should switch to searching preprocessed scripts variant. There's another reason to do this: if we like to have this scan as a distinct operation, which may be run on its own without compiling scripts.

ericoporto:
It could preprocess, and then search on it using a regex pattern, since ags script doesn't support nested functions it should be possible to find the functions, possibly with some checks to allow both void and function (maybe int too?) event handlers.

ivan-mogilko:
@ericoporto this is literally what I did in this pr.

ivan-mogilko:
Made it so that, if function's name was in the event property, but function itself did not exist, then pressing "..." button would create that function.

ivan-mogilko:
No, this is not right, I was not thinking straight apparently.

The function match should not be done over a script text, because that means regex search done over full script text for *each* event function name. Even if script is preprocessed, that will be very slow.

The function match must be done over prepared *list of functions* (or specifically Set, or ordered list for more optimization).
What this list is created by is a separate question.
If we run this check after compilation, then we might as well use the result of a compilation, because the parsing job was already done by compiler.
If we run this check unrelated to compilation, then we will have to utilize script parser somehow. Either ask compiler to do a "dry run" over a script, when it only gathers symbols but does not create a bytecode etc. Or even just write a primitive sort of parser than searches for all entries matching standard event function regex patterns.

But in the end this must be a set of function names, which in turn we use to match event properties.

ericoporto:
About the function list, it's there already in the script autocomplete cache and used to fill the combobox at top of the script. Perhaps the script autocomplete cache should be kept somewhere when the script is unloaded so it can still be looked up?

About actually testing this, what I noticed is that in the case of a room, if I have an invalid room event handler and click the play button for testing, it appears that it isn't checked - I guess because the room isn't rebuilt? Not sure. But if I click rebuild all scripts then the handler errors actually appear in the output panel.

ericoporto:
About the parser, I am not completely sure, but I think that the one for autocomplete data can pickup if you haven't finished writing some code instead of giving error and also understands that you are local to a function - so it shows local stuff, from picking up where the cursor is.

The one in the compiler errors when the code is unfinished, and I don't know if it has any sort of interface to be able to pickup local symbols too - which are necessary for autocomplete.

ivan-mogilko:
> About the parser, I am not completely sure, but I think that the one for autocomplete data can pickup if you haven't finished writing some code instead of giving error and also understands that you are local to a function - so it shows local stuff, from picking up where the cursor is.
> 
> The one in the compiler errors when the code is unfinished, and I don't know if it has any sort of interface to be able to pickup local symbols too - which are necessary for autocomplete.

I suppose that might be a matter of making parser have its own interface exposed, and work in certain way, like, ignore unfinished parts, ignore all function bodies except one and gather local symbols from a function found at the given line, and so forth.

Given that logic of parsing the code and gathering symbols seems to be the same, I think it is worth investigating this approach, and see how feasible it is to share same tool for multiple purposes.

That's not something I want to try right away, but something to consider as an option in the future.

ivan-mogilko:
I changed it to use Autocomplete, and also put this process into a function called `RunGameScriptTests`, so it may be run from another place if anyone would like to do so later.

ivan-mogilko:
Okay, I merged this, as this works overall.
Any missing issues may be addressed separately.

ivan-mogilko:
> Hey, I was catching up with this code, would it be alright to rename all the Test in here to Check? I think Test feels like a different thing. I would commit this on master if it's alright ([ericoporto@036cb74](https://github.com/ericoporto/ags/commit/036cb74662142e25bb04ac31e274e959ee0429fc))

Alright.

ericoporto:
done! :)

PR #2569, opened 2024-11-07
Engine: use safe strcpy function instead of snprintf where truncation expected

### LINKED ISSUE #2567: Truncation warning in Common/game/main_game_file.cpp
**Describe the bug**

In line 651 in UpgradeCharacters:
warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
snprintf(chars[i].scrname, LEGACY_MAX_SCRIPT_NAME_LEN, "c%s", namelwr);

This is quite obvious since it adds a "c" to the string and both source and dest are the same size (20).

ivan-mogilko:
Fix #2567

This replaces uses of `snprintf` in the code where truncation of the source string is *intended*.
Added a "ags_strncpy_s", using "strncpy_s" from the C11 extension as a reference (see https://en.cppreference.com/w/c/string/byte/strncpy)

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -639,16 +639,19 @@ void UpgradeCharacters(GameSetupStruct &game, GameDataVersion data_ver)
     }
 
     // Fixup character script names for 2.x (EGO -> cEgo)
+    // In 2.x versions the "scriptname" field in game data contained a name
+    // limited by 14 chars (although serialized in 20 bytes). After reading,
+    // it was exported as "cScriptname..." for the script.
     if (data_ver <= kGameVersion_272)
     {
-        char namelwr[LEGACY_MAX_SCRIPT_NAME_LEN];
+        char namelwr[LEGACY_MAX_SCRIPT_NAME_LEN - 1];
         for (int i = 0; i < numcharacters; i++)
         {
             if (chars[i].scrname[0] == 0)
                 continue;
-            memcpy(namelwr, chars[i].scrname, LEGACY_MAX_SCRIPT_NAME_LEN);
+            ags_strncpy_s(namelwr, sizeof(namelwr), chars[i].scrname, LEGACY_MAX_SCRIPT_NAME_LEN - 2);
             ags_strlwr(namelwr + 1); // lowercase starting with the second char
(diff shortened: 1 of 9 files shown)


ivan-mogilko:
@ericoporto i copied error values and some parts from your `ags_strncpy_s` implementation in ags4, but copy characters in a for loop in place instead of calling snprintf, because extra call to a formatting function seems unnecessary in this case.

Also added asserts to help catch errors.

ericoporto:
Looks fine, there is a change to `Engine/main/update.cpp` I can’t tell if it was intended or not.

ivan-mogilko:
> Looks fine, there is a change to `Engine/main/update.cpp` I can’t tell if it was intended or not.

Oops, that was a fix for another warning, it's already in master.

ericoporto:
This appears to work fine but I found an error when loading a save - I was testing playing [mags version of The Shivah](https://archive.org/details/mags_the-shivah). I also tried running it with master and got the same error - created a new save with the master version and then tried to load it. Because it's on master too, it's not related to this PR, so I would say the PR appears to work alright, and would be fine to merge it. I will leave the unrelated error reported below.

image

```
An internal error has occurred. Please note down the following information.
If the problem persists, contact the game author for support or post these details on the AGS Technical Forum.
(Engine version 3.6.2.2)

Error: Unable to restore the saved game.
Failed to restore the savegame component.
(#12) Script Modules, version 3060200, at offset 56866.
Saved content does not match current game.
Save is missing a script module that exists in the game: Internal built-in header.
```

ivan-mogilko:
> I also tried running it with master and got the same error - created a new save with the master version and then tried to load it.

This may be related to the merged changes from #2489, or earlier changes which save script modules with their names...

ivan-mogilko:
In regards to this PR, I tested building using GCC 9.4.0, and this fixes 3 warnings with snprintf truncations.

ivan-mogilko:
@ericoporto I pushed 2 fixes to the master, this problem was introduced at an earlier change to save format: 
06363d796b02ffee65de45e60073dbff17ecdaa0

But apparently this does not happen in newer compiled games, due to the difference in how script data is filled.

ericoporto:
I confirm that the fix on master indeed works! Can save and load fine when playing mags version of The Shivah.

PR #2572, opened 2024-11-10
Editor: fix pressing alt doesn't select File in menu

### LINKED ISSUE #2566: Editor: focusing Menu with Alt key is broken since the first open-source release
This is probably one of those things that are rarely used, that's my explanation why it was never reported in all those years.

**Problem**
Pressing Alt key is supposed to set a keyboard input focus on the Main Menu strip.
Instead it seems to set a keyboard focus on "system menu" (the one which is also called by clicking on a window icon in top-left corner). Pressing up/down arrow keys will open it instead of main menu's first submenu.
OTOH if you have any pane undocked and in floating mode, then Alt sets focus to *that floating pane's* window menu instead.
I do not think any of that was intended.

**Steps to reproduce**
1. Open the AGS Editor
2. Press Alt. Notice the Main Menu changes the look by underlining the first letters in its options. But the first menu item is not selected.
3. Press up or down arrow keys on keyboard. Notice that system menu opened instead of the main menu's submenu.
4. Make sure some pane is open, e.g. even if a "Welcome" pane. Make that pane floating by dragging it by the tab.
5. Now press Alt and then up or down keys. Notice how a this floating pane's window menu opens.

**AGS Versions**
Any version past 3.2.1 (3.2.2 and onwards).
You can check v3.2.1 to see how it was supposed to work.

**Expected behavior**
Pressing Alt should set a keyboard focus to the Main Menu.

ericoporto:
fix #2566 

From [the MSDN docs on the property](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.mainmenustrip?view=netframework-4.6&devlangs=csharp&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Windows.Forms.Form.MainMenuStrip)%3Bk(TargetFrameworkMoniker-.NETFramework%2CVersion%253Dv4.6)%3Bk(DevLang-csharp)%26rd%3Dtrue), it appears it was just something it was not set by accident.

### DIFF
--- a/Editor/AGS.Editor/GUI/frmMain.Designer.cs
+++ b/Editor/AGS.Editor/GUI/frmMain.Designer.cs
@@ -78,6 +78,7 @@ namespace AGS.Editor
             this.FormClosing += new System.Windows.Forms.FormClosingEventHandler(this.frmMain_FormClosing);
             this.KeyDown += new System.Windows.Forms.KeyEventHandler(this.frmMain_KeyDown);
             this.IsMdiContainer = true;
+            this.MainMenuStrip = this.mainMenu;
             this.Controls.Add(this.mainContainer);
             this.Controls.Add(this.toolStrip);
             this.Controls.Add(this.mainMenu);


ivan-mogilko:
There's still an issue with floating panes: when there's an active float pane, its window menu is focused instead.

ericoporto:
I am not sure how to fix that in AGS, it appears that issue is in the dockpanel. I saw a commit online that I think fixes this in it: https://github.com/tsanie/dockpanelsuite/commit/ac583863655c3dbd96b876843272e7e5df84fb38#diff-3278c81de04fbf296b216c4dc761a9a57ae9b5cf17d75c599160b21d0f8eec33R291

ivan-mogilko:
> Edit: uhm, perhaps if I can do it in Editor Content Panel it would work.

Since it is Docking.FloatWindow that implements this form which has its window menu selected, we might look into AGS.Editor.FloatingWindow class, which serves a wrapper over the above, or AGS.Editor.DockingPane, which constructs it.

This is where we have an access to the library's form (FloatWindow type).

Maybe this is where it could be reconfigured, or have some handler attached to it. I know that there has to be ways to intercept window messages. This is not a exact solution, only taking notes for now.

ivan-mogilko:
> Edit: uhm, perhaps if I can do it in Editor Content Panel it would work.

Since it is Docking.FloatWindow that implements this form which has its window menu selected, then it's the one which has to be looked into. Maybe it could be reconfigured upon creation, or have some handler attached to it. I know that there has to be ways to intercept window messages. This is not a exact solution, only taking notes for now.

I also usually have hard times understanding connections between AGS.Editor gui objects and docking classes, especially because these hide beyond custom interfaces (i suspect that this was done for the sake of editor plugins).

ivan-mogilko:
I will merge this, as this might fix the problem for the majority of layouts (not sure if anyone is using floating panes often).

I tried few things, but it looks like it's a standard behavior for the child popup windows inside a parent window to steal focus and alt key as well. Note that it's not the *presence* of a floating pane that steals the Alt key, but the focus: if you click on background, the main window gets focus, and then Alt key goes to Main Menu as usual.

Furthermore this remaining issue does not seem like worth spending more time on it.

PR #2573, opened 2024-11-10
Editor: fix exception when saving at close

ericoporto:
AfterRecentGamesChanged has it's event removed (the update of the Open Recent menu), when the Editor is closed, but later Settings_LimitRecentGames events triggers. This prevents an exception at null in this specific case.

I think this is one of those things I should commit directly on master but I was a bit unsure if it was something wrong on my side since this wasn't reported yet and I thought it would be something common to see.


ivan-mogilko:
Strictly speaking you should be always testing event for being null, as no subscribers is a valid case.

PR #2574, opened 2024-11-10
Backport Editor Template info preview

ericoporto:
This was the PR #1903 , which fixed the issue #564 in AGS 4.

image

I wanted to backport it to ags3 as I really like it. I also wanted to fix a small annoyance I had with the change of the wizard dialog height, I tried to keep it the same, which is my last commit. I also did some small uses of layout in it using table and some docking.

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -23,7 +23,6 @@ namespace AGS.Editor
         public const string FILE_MENU_ID = "fileToolStripMenuItem";
         public const string HELP_MENU_ID = "HelpMenu";
         private const string CONTROL_ID_SPLIT = "^!^";
-        private const string TEMPLATE_INTRO_FILE = "template.txt";
 		private const string ROOM_TEMPLATE_ID_FILE = "rtemplate.dat";
 		private const int ROOM_TEMPLATE_ID_FILE_SIGNATURE = 0x74673812;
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
Is this size adjustment because of the first page with the image? Something has to be done with it, maybe it can be scaled dynamically, replaced with another image, or completely redone into a form that is easier to scale.

ivan-mogilko:
I tested this, and something is off with text layout.

On the first page the text is now moved to the right, occupying maybe rightmost half of the blank space.

game-wizard-layout


On next pages, the title texts are moved further right and closer to the bottom of a white space, and have greater spacing between the lines. I cannot tell why was this necessary.

game-wizard-layout2

ericoporto:
Uhm, I didn't notice those changes, I will fix those.

What I actually want to get is use more of Layouts  (for safe scaling) but try to keep as close as the current Wizard as possible.

> Is this size adjustment because of the first page with the image?

Yeah, I don't like the blank part, but at the same time for that image specifically I can't think of a way to make it dynamically adjust. I am still understanding how the dialog panel actually gets the wizard pages in - layout wise.

ivan-mogilko:
> Yeah, I don't like the blank part, but at the same time for that image specifically I can't think of a way to make it dynamically adjust.

I think it's worth opening a separate issue for this, and maybe ask on forums for help.
In the past I've changed the splash window from a single fixed image to dynamically constructed image with an icon in the middle, following an advice from the forum members.

PR #2577, opened 2024-11-13
Editor: multiline text editing in labels and buttons

ericoporto:
This is just an idea for a multiline text editor for property grids

image

https://github.com/user-attachments/assets/abffa5c1-3169-4992-aa7b-9fa8d2202592

When editing properly tagged strings, they will show a three dots and then hitting it they show this custom modal dialog that allow multiline editing.

[I had this idea after this forum post](https://www.adventuregamestudio.co.uk/forums/beginners-technical-questions/how-to-create-a-large-gui-that-would-hold-a-lot-of-text/msg636666904/?boardseen#new)

In a far future of ags perhaps it has an option to have a folder of txt files and you can assign one to a label or whatever, but I don't know the ui for this.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -210,6 +210,12 @@
     <Compile Include="GUI\LogPanel.Designer.cs">
       <DependentUpon>LogPanel.cs</DependentUpon>
     </Compile>
+    <Compile Include="GUI\MultilineStringEditorDialog.cs">
+      <SubType>Form</SubType>
+    </Compile>
+    <Compile Include="GUI\MultilineStringEditorDialog.Designer.cs">
+      <DependentUpon>MultilineStringEditorDialog.cs</DependentUpon>
+    </Compile>
     <Compile Include="GUI\NumberEntryWithInfoDialog.cs">
       <SubType>Form</SubType>
     </Compile>
@@ -441,6 +447,9 @@
     </EmbeddedResource>
     <EmbeddedResource Include="GUI\LogPanel.resx">
       <DependentUpon>LogPanel.cs</DependentUpon>
+    </EmbeddedResource>
+    <EmbeddedResource Include="GUI\MultilineStringEditorDialog.resx">
+      <DependentUpon>MultilineStringEditorDialog.cs</DependentUpon>
     </EmbeddedResource>
     <EmbeddedResource Include="GUI\NumberEntryWithInfoDialog.resx">
       <DependentUpon>NumberEntryWithInfoDialog.cs</DependentUpon>
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -1465,9 +1465,9 @@ namespace AGS.Editor
             return CustomResolutionDialog.Show(currentSize);
         }
 
-        private String ShowMultilineStringDialog(String text)
+        private String ShowMultilineStringDialog(string title, String text)
         {
-            return MultilineStringEditorDialog.ShowEditor(text ?? string.Empty);
+            return MultilineStringEditorDialog.ShowEditor(title, text ?? string.Empty);
         }
 
         private Color? ShowColorDialog(Color? color)
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -891,6 +891,7 @@ namespace AGS.Editor
                 RoomMessagesUIEditor.ShowRoomMessagesEditor = new RoomMessagesUIEditor.RoomMessagesEditorType(ShowRoomMessageEditorFromPropertyGrid);
                 CustomResolutionUIEditor.CustomResolutionSetGUI = new CustomResolutionUIEditor.CustomResolutionGUIType(ShowCustomResolutionChooserFromPropertyGrid);
                 ColorUIEditor.ColorGUI = new ColorUIEditor.ColorGUIType(ShowColorDialog);
+                MultiLineStringUIEditor.MultilineStringGUI = new MultiLineStringUIEditor.MultilineStringGUIType(ShowMultilineStringDialog);
                 AudioClipSourceFileUIEditor.AudioClipSourceFileGUI = new AudioClipSourceFileUIEditor.AudioClipSourceFileGUIType(ShowAudioClipSourceFileChooserFromPropertyGrid);
             }
         }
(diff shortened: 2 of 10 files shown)


ericoporto:
I was thinking if I should trim the text too. [I also added a topic in the forums](https://www.adventuregamestudio.co.uk/forums/editor-development/idea-editing-multiline-text-in-labels-in-the-gui-editor/msg636666957/#msg636666957) to see if someone there gives other ideas

ericoporto:
I like E as a key, I think I can put it and signal it nicely in the context menu. 
<details>
F2 could work too, but I think for text in a GUI it may be easier - I may be a little selfish here that my laptop also has the f1-12 keys hidden behind an fn key, which is a bit hard to remember, plus for some unknown reason it's in place of ctrl Key, which I reversed in software, which means I need to press Ctrl+F2 to actually type F2... But plain letter keys are just a single press, so I would go with that.
</details>

A couple of little things I could use some help to think about

- The window position, I made it load at center of the AGS Editor, I don't know if there's a better way, I don't think my monitor is big enough to make a difference. Also I was balancing if you load by key press vs clicking in the three dots in the property grid. Would it be better to do the initial location differently?
- In a text editor, I sometimes hit an extra enter, this is an unfortunate habit (due to very old IDEs that cared about the last line). I was thinking about trimming automatically when hitting OK, but don't know if it's a good idea
- I did a very simple two-way replace of `\n` to `\\n`, I don't know if I need to care about other special characters there (text in multiline editor vs in the property grid field)

ericoporto:
E is actually forbidden as shortcut , was getting a weird exception when I tried to set it, it's not documented, but the code is open nowadays...: https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/ToolStripMenuItem.cs,480
https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/ToolStripManager.cs,54f03ea0326af803

I guess I will assign f2, ugh....

ivan-mogilko:
Finally got to test this. I think this is a wonderful addition.

The biggest problem I found so far is that for some reason the text editor stops displaying linebreaks after opening it for a text that has ones. That is, if I have a text property "first line \n second line", then text editor will display this in a single line, although at the same time it seems to keep this "\n" hidden, because when I apply the changes, the property field has "\n"s retained.

Besides the above, if you start editing by a menu command, then the property value is not visually refreshed right after closing the dialog (you have to click around to make it update). There was this function `Factory.GUIController.RefreshPropertyGrid();`, maybe it will be useful here.

Last thing, would it be possible to also make Shift + Enter act as OK, along with Ctrl + Enter? Not sure where, but I got used to this combo too.

ivan-mogilko:
Ah, and I just realized, this window will definitely benefit from saving its position and size, using the window config.

Like in the SpriteChooser:

https://github.com/adventuregamestudio/ags/blob/a26ef7d49743686862b2f58d5f6396fcd03a21b0/Editor/AGS.Editor/GUI/SpriteChooser.cs#L40-L45

https://github.com/adventuregamestudio/ags/blob/a26ef7d49743686862b2f58d5f6396fcd03a21b0/Editor/AGS.Editor/GUI/SpriteChooser.cs#L149-L154

ericoporto:
> Besides the above, if you start editing by a menu command, then the property value is not visually refreshed right after closing the dialog 

I literally never get this right the first time lol! 

> The biggest problem I found so far is that for some reason the text editor stops displaying linebreaks after opening it for a text that has ones

I will debug around and see where I got it wrong, thanks!

ericoporto:
OK, I learned that apparently the textbox uses CR+LF for linebreaks (hitting return there puts these two characters), so I accounted for that, this is why it wasn't working before, and it was leaking a `\r` character.

I added support for Shift and also am saving the position - I used the suggested helper from SpriteChooser!

The property grid refresh though, it was refreshing for me somehow, but I put a refresh on the OnClose event anyway, hopefully this works.

ivan-mogilko:
> The property grid refresh though, it was refreshing for me somehow, but I put a refresh on the OnClose event anyway, hopefully this works.

I think it should be outside of the dialog class, after the new property value is set:
```
textProperty.SetValue(_selectedControl, newText);
```

ivan-mogilko:
I confirm that it's working overall, and all the problems I reported earlier are fixed.

The only remaining problem is editing a text when multiple controls are selected. Editing Text property directly works, but pressing a "..." button throws an error (and I noticed that Ctrl+E is disabled in this case).

The correct behavior for this scenario is to assume the input (old) property value is "default value" (empty string in case of Text), and the new value will be applied to all selected objects. That's standard Property Grid behavior, and how all other properties work. I had to fix a number of TypeConverters recently to make it work for properties that require opening a dialog (Color and Sprite selection). This text property's converter may be fixed same way.

Here are example of my past fixes:
d8a90fa538b7c36e254f40758a72d29a17c31061

From what I see, the `MultiLineStringUIEditor` class has everything right, but maybe MultilineStringEditorDialog needs to be fixed to check if the input "text" argument is null, in which case it's replaced by string.Empty. Like
```
MultilineString = text ?? string.Empty;
```

---

As for "Ctrl + E" and context menu case, I believe that this may also be allowed for multiselection.
But in any case, currently there's an inconsistency: Ctrl + E combo is ignored in case of multiselection, but context menu option is still enabled and works. Except it gets input string and applies new value to the last selected object only.

ivan-mogilko:
> Erh, does editing text in multi-selection makes sense? I can only imagine this done by accident, is there an use-case someone really wants to apply text to multiple items selected in a GUI?

I am in doubts myself, so I think it's fine to just leave the context menu command to work only for the last selected control for the time being.

`_selectedControl` contains the last selected control, iirc.
`_selected` list contain all selected controls, in the order they were selected.

ivan-mogilko:
> Edit3: out of curiosity, does shift + click should in theory support multi-selection of things in the gui editor? I could only do it by clicking and dragging a rectangle, I guess palm rejection also kills shift + click on my notebook in case this is implemented.

It's Ctrl + Click currently, maybe because Ctrl means "select single" and Shift usually means "select a range", but it's hard to tell what "range" would mean in this case.

ivan-mogilko:
From testing again, it looks like the only remaining issue is to make context command work consistently, for both context menu item and Ctrl + E combination. That is either:
- Works for multiple selected controls, or
- Is enabled, but edits last selected control's text, or
- Is disabled when multiple controls are selected.

ivan-mogilko:
> I have an idea that I don't know how to do, what if the title text of the dialog instead of "Edit text..." read "Edit label1 text..." 
> I am not sure how to actually get the name of the controls being edit through the property grid though.

TypeConverter should receive edited object (or collection of objects) in `EditValue` method, through the `ITypeDescriptorContext context` parameter.

See `ITypeDescriptorContext.Instance`:
https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.instance?view=net-8.0

Testing this, in case of a single control the Instance can be cast to GUIControl, and in case of multiple controls it is cast to an array of GUIControl[]

ericoporto:
I think I figured it out. I put in a separate commit.

There is some minor code duplication but the way I had to handle to add the title text that I wanted was slightly different from the Property Grid vs from the GUI Editor itself, plus some minor things to ensure that the multiple selection was handled correctly.

ivan-mogilko:
This looks alright.

PR #2579, opened 2024-11-15
Script API: Prescan saves and test for validity

ivan-mogilko:
This is a logical continuation of #2489

Adds a script function called `Game.ScanSaveSlots()`:

```
static void Game.ScanSaveSlots(int valid_slots_dest[], int min_slot, int max_slot,
    SaveGameSortStyle saveSortStyle = eSaveGameSort_None, SortDirection sortDirection = eSortNoDirection, int user_param = 0);
```

Game.ScanSaveSlots() runs a prescan over a range of slots, checks them for validity and compatibility using all available means (including calling "validate_restored_save" callback in script, if one is available), then fills the provided destination array with found valid slot numbers. Array's length is taken into account, limiting number of slots that may be stored there. Any remaining array elements are initialized with -1 - this is to help user to determine that no more slots have been found.

UPDATE:
Also added couple of auxiliary functions:
```
/// Fills the list box with the current user's saved games using the array of slot indexes.
void ListBox.FillSaveGameSlots(int save_slots[], SaveGameSortStyle saveSortStyle = eSaveGameSort_None, SortDirection sortDirection = eSortNoDirection);

/// Returns a dynamic array, containing indexes of found save slots in the range between "min_slot" and "max_slot"
static int[] Game.GetSaveSlots(int min_slot, int max_slot, SaveGameSortStyle saveSortStyle = eSaveGameSort_None, SortDirection sortDirection = eSortNoDirection);
```

NOTES:
    1. Because save validation may require to run "validate_restored_save" callback in script, the scan process may only be performed post-script (similar to functions like SaveGameSlot, RestoreGameSlot, etc). The scanning action is not executed immediately, but scheduled to run after the current script is done.
    2. User is reported about the completion of scanning by sending a `eEventSavesScanComplete` event, which may be handled in a "on_event" script callback. "user_param" is an optional argument in ScanSaveSlots, which is sent along with `eEventSavesScanComplete`, and lets distinguish multiple calls to ScanSaveSlots.
    3. In order to safely handle the destination array, which is a managed object, we add an internal reference when scheduling the scan, and release that reference after scan is complete. This ensures that the array object will be present at the time of scanning even if user looses its reference by mistake. (We can't simply check for array's existence, because same handle may be used by a different object.)
    4. "validate_restored_save" callback may be run as many times as there are saves found. User may find which slot is tested using `RestoredSaveInfo.Slot` property.
    5. In case "validate_restored_save" is present, scanning process will use the resulting "RestoredSaveInfo.Cancel" property value to know whether to add this slot to destination array or not.

Primitive script example:

```
int prescan_slots[];

function on_key_press(eKeyCode keycode, int mod)
{
  if (keycode == eKeyP)
  {
    gScanSaves.Visible = true;
    prescan_slots = new int[100];
    Game.ScanSaveSlots(prescan_slots, 0, 100);
  }
}

function on_event(EventType evt, int data)
{
  if (evt == eEventSavesScanComplete)
  {
    // Variant 1
    listSaves.Clear();
    for (int i = 0; i < 100; i++)
    {
      if (prescan_slots[i] < 0)
        break; // no more valid slots

      String s = Game.GetSaveSlotDescription(prescan_slots[i]);
      if (s != null)
        listSaves.AddItem(s);
    }

    // Variant 2
    listSaves.FillSaveGameSlots(prescan_slots); // simple like that
  }
}
```

### DIFF
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -706,6 +706,32 @@ void GUIMain::ReadFromSavegame(Common::Stream *in, GuiSvgVersion svg_version, st
     }
 }
 
+void GUIMain::SkipSavestate(Stream *in, GuiSvgVersion svg_version, std::vector<ControlRef> *ctrl_refs)
+{
+    if (svg_version < kGuiSvgVersion_350)
+        in->Seek(15 * sizeof(int32_t));
+    else
+        in->Seek(18 * sizeof(int32_t));
+
+    if (svg_version >= kGuiSvgVersion_36200)
+    {
+        if (!ctrl_refs)
+        {
+            in->Seek(in->ReadInt32() * sizeof(int32_t));
+            return;
+        }
+
+        uint32_t ctrl_count = in->ReadInt32();
+        ctrl_refs->resize(ctrl_count);
+        for (uint32_t i = 0; i < ctrl_count; ++i)
+        {
+            const uint32_t ref_packed = in->ReadInt32();
+            (*ctrl_refs)[i].first = (GUIControlType)((ref_packed >> 16) & 0xFFFF);
+            (*ctrl_refs)[i].second = ref_packed & 0xFFFF;
+        }
+    }
+}
+
 void GUIMain::WriteToSavegame(Common::Stream *out) const
 {
     // Properties
--- a/Common/gui/guimain.h
+++ b/Common/gui/guimain.h
@@ -201,6 +201,7 @@ public:
     // Perhaps review this later (maybe when there's a split runtime GUI class).
     void    ReadFromSavegame(Stream *in, GuiSvgVersion svg_version, std::vector<ControlRef> &ctrl_refs);
     void    WriteToSavegame(Stream *out) const;
+    static void SkipSavestate(Stream *in, GuiSvgVersion svg_version, std::vector<ControlRef> *ctrl_refs);
 
 private:
     void    DrawBlob(Bitmap *ds, int x, int y, color_t draw_color);
--- a/Common/util/directory.h
+++ b/Common/util/directory.h
@@ -236,14 +236,6 @@ struct FileEntryCmpByName
     }
 };
 
-struct FileEntryCmpByNameDsc
-{
-    bool operator()(const FileEntry &fe1, const FileEntry &fe2) const
-    {
-        return fe2.Name.Compare(fe1.Name) < 0;
-    }
-};
-
 struct FileEntryCmpByNameCI
 {
     bool operator()(const FileEntry &fe1, const FileEntry &fe2) const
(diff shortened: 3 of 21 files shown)


ivan-mogilko:
@NPatch poking you in case you're interested to review and/or test this.

ivan-mogilko:
Expanded function, now it is:
```
static void Game.ScanSaveSlots(int valid_slots_dest[], int min_slot, int max_slot, FileSortStyle fileSortStyle = eFileSort_None, SortDirection sortDirection = eSortNoDirection, int user_param = 0);
```

Sorting should be obvious (sort by name means by slot number in practice),
"user_param" is sent along with the `eEventSavesScanComplete`, and lets to distinguish multiple calls to ScanSaveSlots. Acts like a operation "tag" basically.

ivan-mogilko:
Note that I might also add a second function that does not do any validation and only fills existing saves into array of slots.
Such function may be useful if you don't want to have to fill a listbox with saves, but keep things hidden.

NPatch:
I just compiled AGS editor. But trying to reload the previous testing game, It says "Could not build target platform _Debug, because the file 'acwin.exe' is unavailable." Can't find sth about this.

ivan-mogilko:
Added `Game.GetSaveSlots()` function that is declared as:
```
/// Returns a dynamic array, containing indexes of found save slots in the range between "min_slot" and "max_slot"
static int[] Game.GetSaveSlots(int min_slot, int max_slot, FileSortStyle fileSortStyle = eFileSort_None, SortDirection sortDirection = eSortNoDirection);
```

This is not relevant to the save validation, but is rather a complementary function. It executes instantly and returns a dynamic array of ints. I suppose that since we have array.Length ported to 3.6.2 now, it's okay to return new dynamic array instead of requiring user to pass the one they created.

ivan-mogilko:
I'd like to clarify something, about the timing of scanning, and "on_event" callback.

When `Game.ScanSaveSlots` is called the scanning is scheduled, and it is executed after the current script has finished, in a *post-script step*. This is where it queues the event `eEventSavesScanComplete`.

But then, the queued "on_event" calls are also run in the same post-script step, right after all the scheduled actions are finished.
As a result, the whole thing normally completes *before any other script runs*.

For example, if you call Game.ScanSaveSlots in "game_start", "room_Load" and "room_AfterFadeIn":

- **game_start**: call Game.ScanSaveSlots, scanning is scheduled
- after game_start exits, scanning is run
- on_event is run
- **room_Load**: call Game.ScanSaveSlots, scanning is scheduled
- after room_Load exits, scanning is run
- on_event is run
- fade-in performed (or another room transition effect)
- **room_AfterFadeIn**: call Game.ScanSaveSlots, scanning is scheduled
- after room_AfterFadeIn exits, scanning is run
- on_event is run
- normal game loops begin to run.

ivan-mogilko:
As a practical experiment, here's a modified script for Sierra template, which uses Game.ScanSaveSlots instead of ListBox.FillSaveGameList():

[GlobalScript.txt](https://github.com/user-attachments/files/17852863/GlobalScript.txt)

May be compared with original here: https://github.com/adventuregamestudio/ags-template-source/blob/master/Sierra-style/GlobalScript.asc

ivan-mogilko:
After a discussion with @NPatch, I added `ListBox.FillSaveGameSlots()` function that accepts a dynamic array of slot indexes.

This lets avoid keeping the external array filled with prescanned save slots, and make use of ListBox.SaveGameIndex property again.

PR #2580, opened 2024-11-15
Editor: new game wizard project name changes filename

ericoporto:
In the Start New Game wizard, if you type the project name, it will now automatically fill the filename too. If you then type a different filename there it will stop automatically following what is typed in project name. If you then delete what is in the filename AND later types in the project name it goes back to automatically tracking the project name. Here is a video to make it easier to understand

https://github.com/user-attachments/assets/bef088c8-4b7a-48fe-bdd3-4b4b5f249a4c

This is something I have wanted after creating a gazillion of ags projects for testing things, because I always have to type things twice.

I can see people may have different opinions on how the filename from project name rules should be but I gave a good default that feels sensible for me. I don't think it has to work for everyone, people can still type manually like it always been if they have a specific taste.

### DIFF
--- a/Editor/AGS.Editor/GUI/StartNewGameWizardPage2.Designer.cs
+++ b/Editor/AGS.Editor/GUI/StartNewGameWizardPage2.Designer.cs
@@ -92,6 +92,7 @@ namespace AGS.Editor
             this.txtFriendlyName.Size = new System.Drawing.Size(286, 21);
             this.txtFriendlyName.TabIndex = 0;
             this.txtFriendlyName.Text = "New game";
+            this.txtFriendlyName.TextChanged += new System.EventHandler(this.txtFriendlyName_TextChanged);
             // 
             // label2
             // 
@@ -113,6 +114,7 @@ namespace AGS.Editor
             this.txtFileName.Size = new System.Drawing.Size(286, 21);
             this.txtFileName.TabIndex = 1;
             this.txtFileName.TextChanged += new System.EventHandler(this.txtFileName_TextChanged);
+            this.txtFileName.KeyPress += new System.Windows.Forms.KeyPressEventHandler(this.txtFileName_KeyPress);
             // 
             // label1
             // 
--- a/Editor/AGS.Editor/GUI/StartNewGameWizardPage2.cs
+++ b/Editor/AGS.Editor/GUI/StartNewGameWizardPage2.cs
@@ -13,6 +13,9 @@ namespace AGS.Editor
 {
     public partial class StartNewGameWizardPage2 : WizardPage
     {
+        private bool _typedInFileName = false;
+        private readonly Debouncer _afterFileNameClearDebouncer = new Debouncer(150);
+
         public StartNewGameWizardPage2(string baseDirectory)
         {
             InitializeComponent();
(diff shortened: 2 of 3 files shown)


ivan-mogilko:
Hmm, interesting idea itself, although would not it be easier to just make it so that it takes the game name as a filename if nothing is written there? For example, General Settings support empty GameFileName, in which case it uses Game Name.

Anyway, in regards to the functionality, I am very much confused by the conversion rule. Why does it capitalize words even if the game name is all in small letters, and why does it remove hyphens and underscores, if these are pretty much supported for a folder name? For example, I have a habit of calling my test games by names consisting of several parts separated with a "--", but with this rule I will have to rename the filename anyway.
TBH this rule looks like it would rather suit an opposite task: getting game title from a game file name.

ericoporto:
OK, full disclosure, I partially stole the idea from PyCharm

https://github.com/user-attachments/assets/63307338-fb65-4d24-92a3-ff7f6c3aaca2

PyCharm doesn't do the replacing of spaces, and the other thing it doesn't do is that once you type in the filename it's over, there is no going back - and even if you close this window, if you try to create a new project it still remembers where you left off this window, so to get it clean you need to kill it and start it over.... So it's not perfect.

> Why does it capitalize words even if the game name is all in small letters?

I was going for PascalCase as the work around spaces, but I could go for other approach. We can leave the spaces but I really don't like spaces in filenames, specially when calling things from command line, which happens often with ags games.

> Then, removing whitespaces is maybe okay (still not sure), but why does it remove hyphens and underscores, if these are pretty much supported for a folder name?

I can remove the rule to pick those out.

ericoporto:
OK, I now instead only replace Spaces for hyphens ( `-` ) and the only valid character I am killing is the apostrophe (`'`).

ivan-mogilko:
Mmmm, okay, this *seems* like a method suitable for a wider range of users now.

Earlier I was intending to propose adding an editor preference to choose the folder & file name style, but idk if it's worth it, as majority of users probably don't create new projects often for this to matter...

ericoporto:
https://en.m.wikipedia.org/wiki/Filename , check in the table `Comparison of filename limitations` the row that is named `POSIX "Fully portable filenames"`, there is a list that has `A–Z a–z 0–9 . _ -` as the list of characters, and a note that they can't start with an hyphen (my guess is because a program will then interpret as a command line flag). This is sort of what I had in mind, to do an auto-coversion that would generate a safe filename across platforms.

ivan-mogilko:
The way I see this, first of all there's a question of purpose.

If the purpose is to *enforce* certain filename form, then this has to be enforced in every case, including when a user types it here, and also for the value of GameFileName property in general settings, and when upgrading older projects. Otherwise it will be inconsistent and loose a meaning.

If the purpose is to automate certain naming style, then IMO the most sensible default is to copy name as-is, and anything else (capitalizing first word letters, remove spaces or replace with hyphen/underscore, atc) would be a matter of personal preference, and should rather be a application setting.

Finally, if the purpose is to suggest a "universally safe" name, then IMO this should be an option too, and emphasized (made very clear). Because it was never a thing in AGS, and will confuse people if done unconditionally. Users will simply not understand why they don't get a name they expect, and complain. Not that I suggest that right now, but to give an example, there could be a small button next to this field that "suggests" such safe name.

On another hand, are there known cases when the game name was a real problem to run on a platform that we support?
In the worst case user may rename the game file by hand before distributing, or if we support building for such problem, then it may be converted when building for it.

ericoporto:
I don't really want to do it differently, so I prefer to close this - forums are down and I don't like discord.

PR #2581, opened 2024-11-15
Editor: adds version in Welcome Screen

### LINKED ISSUE #2576: Show AGS version in "Welcome to AGS!" screen
**Describe the problem**
I am running multiple AGS Editors all the time and when it's open editing a game I can easily check the version in the About menu. But when I have multiple AGS Editors in the "Welcome to AGS!" screen, I have no way to figure the version I am in.

**Suggested change**
It would be nice to have the version of AGS somewhere in this screen. It could be anywhere. Here's an example

image

ericoporto:
fix #2576 

adds a subtle version label to the Welcome Screen.

image

this screen in 3.6.1 for reference

image

<details>
<summary>I also redid this screen using layouts, in case we ever want to change it's size or for future scaling support.</summary>

image

</details>

### DIFF
--- a/Editor/AGS.Editor/GUI/WelcomeScreen.Designer.cs
+++ b/Editor/AGS.Editor/GUI/WelcomeScreen.Designer.cs
@@ -43,23 +43,23 @@ namespace AGS.Editor
             this.btnExit = new System.Windows.Forms.Button();
             this.panelLogo = new System.Windows.Forms.Panel();
             this.lblLogo = new System.Windows.Forms.Label();
+            this.tableLayoutPanelAll = new System.Windows.Forms.TableLayoutPanel();
+            this.tableLayoutPanelBottom = new System.Windows.Forms.TableLayoutPanel();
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Hm, what about aligning it to the right side instead?
I've got a feeling like this line sort of catches the eye, because it's aligned with everything else: title and options.

ericoporto:
Uhm, that is a great point! And I do want to make it subtle, let me try that

ericoporto:
OK, I do want it to be subtle, so here is three versions,

1. first it goes to the right of the box

image

2. This one I am trying to push right but align with the right of the box that shows the existing games

image

3. This third approach is to align it to the right of the Adventure Game Studio text

image

ivan-mogilko:
Hmm, I would vote for the first variant. I do not understand the second, and the third looks like something that could be on a splash screen, or if the main title would occupy most of the horizontal space, because right now it emphasizes an empty area at the top-right corner.

PR #2582, opened 2024-11-18
New compiler: support constructors

ivan-mogilko:
CC @fernewelten 
This is a remake of #2529, with proposed changes, based on top of a single original commit.

Managed structs can have an constructor:

* Constructor is defined as a member function of a *managed* struct that has name *identical* to the struct's name.
* The constructor must return `void`. In #2529 I suggested it should be completely typeless, but I did not try that still, as I'm not fully certain whether this is necessary, and and have concerns that implementing it this way may require extra hacks to the parser.
* Can have any parameter list, just like a regular function.
* Cannot be `static`. Having `static` type constructors would technically require to run some sort of a autogenerated script prior to "game_start" callback, but that is far outside current feature.
* Cannot be `protected`. Supporting `protected` will only make sense if we support function overloading, and if there's a way to call one constructor from another.
* Cannot be declared as extender function.
* Structs inherit constructors of their parent(s), but declaring their own constructor will completely "hide" parental one(s).

Constructor is called automatically whenever there's a `new T` or `new T( <...> )` expression.
Old-style `new T` without parameter list is still supported, but only if there's no constructor in T and neither its parent.

Calling constructor function directly in script is *forbidden*.

```
managed struct Stx
{
    int Payload;
    import void Stx(int payload);
};

void Stx::Stx(int payload)
{
    Payload = payload;
}

managed struct Sty
{
    short Payload;
};

function game_start()
{
    Stx *st1 = new Stx(7); // ← okay
    Stx *st2 = new Stx(7.0); // ← syntax error, parameters don't match
    Stx *st3 = new Stx(); // ← syntax error, parameter expected but not found
    Stx *st4 = new Stx; // ← syntax error, parameter list missing

    Sty *st5 = new Sty(); // ← empty list is okay, nothing will be called
    Sty *st6 = new Sty; // ← okay
    Sty *st7 = new Sty(4); // ←syntax error, no constructor declared, so cannot pass parameters
}
```

**TODO:**
- ~Using constructors from parent classes when extended class instance is created, and in case extended class does not define their own constructor.~
- ~Fix compiler tests.~

### DIFF
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -736,6 +736,21 @@ AGS::Symbol AGS::SymbolTable::FindStructComponent(Symbol strct, Symbol const com
     return kKW_NoSymbol;
 }
 
+AGS::Symbol AGS::SymbolTable::FindConstructorOfTypeOrParent(Symbol strct) const
+{
+    if (!IsVartype(strct) || !entries.at(strct).VartypeD->Flags[VTF::kStruct])
+        return kKW_NoSymbol; // not a struct
+
+    // Check for the constructor in the current struct and in all its parents
+    while (strct)
+    {
+        if (entries.at(strct).VartypeD->Constructor != kKW_NoSymbol)
+            return entries.at(strct).VartypeD->Constructor;
+        strct = entries.at(strct).VartypeD->Parent;
+    }
+    return kKW_NoSymbol;
+}
+
 AGS::Vartype AGS::SymbolTable::GetFirstBaseVartype(AGS::Vartype vartype) const
 {
     // TODO: extra safety checks?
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
The essential difference from the initial @fernewelten 's implementation is that this actually registers constructor functions as constructors, by tagging the suitable member functions as "is constructor", and by assigning a "Constructor" field to the Vartype definition.

ericoporto:
Hey, I don't know if you know, but for helping understanding the generated code in the Googletests, @fernewelten made a disassembler [here](https://github.com/fernewelten/agsdiss).

ivan-mogilko:
@fernewelten could you please review this if you've got spare time?

ivan-mogilko:
Alright, judging by my tests this works, maybe I'll merge this, and it is going to be tested along with ags4 wip releases.

Issue #2300, opened 2024-01-18
Script compiler: register static function imports as "name^argnum" symbol


ivan-mogilko:
Script compiler registers imported struct member functions using a "name^argnum" notation. This helps engine to link different versions of API function with extended argument list.

But for some reason same is not done for static functions, which do not belong to a struct. Although it is a common idea that we should move towards having all functions belonging to some type, like under a namespace, there are still a lot of those which are not, and sometimes there's a need to extend their arg list.

Need to investigate if there's any technical reason not to use same registration method for static functions, and if not then register them as "name^argnum".

This problem refers to both new and old compiler, new one is a priority but old should be adjusted to if possible.

fernewelten:
Could you please check this again?

Because as far as I can see, both the compilers do export all functions, static and otherwise, with the expected number of parameters appended (after a `$`, not a `^`).
- For the new compiler, cf. function `AddExport()` in file `Compiler\script2\cc_compiledscript.cpp`.
- For the old compiler, cf. function `add_new_export()` in file  `Compiler\script\cc_compiledscript.cpp`.

It seems that this might be a purely Engine matter, pertaining to how predefined (`C++`) functions are linked. The compilers seem to be out of it.

It might even be that at the time that the `function$parametercount` convention was introduced, nobody felt like touching the signatures of the functions that already existed at that time, and so it's simply the ‘old’ functions that don't have the parameter count, not the ‘static’ functions.

ericoporto:
Looks like this is only used here

https://github.com/adventuregamestudio/ags/blob/6baad09f555d9dd98ea26c48cc02b54f7515f4e9/Engine/script/systemimports.cpp#L83

ivan-mogilko:
@fernewelten these are not *exports*, these are *imports*, compilers treat these separately. They append ^N during some kind of a post-step, after all imported functions have been gathered.

Here's what the old compiler is doing:
https://github.com/adventuregamestudio/ags/blob/282cf917e151450dc271c6a26a7e38ba1c957fad/Compiler/script/cs_parser.cpp#L1095-L1101

Here's what the new compiler is doing:
https://github.com/adventuregamestudio/ags/blob/282cf917e151450dc271c6a26a7e38ba1c957fad/Compiler/script2/cs_parser.cpp#L1401-L1407

As you may see, for some reason (which I cannot tell) both only add "^N" part if this function is a member of a struct.

ivan-mogilko:
Okay, I just fixed it,
[...]

PR #2583, opened 2024-11-19
New compiler: Named function arguments, classic char array arguments

fernewelten:
This PR makes it possible (again) to have classic character arrays as function arguments. 

As far as I can see, the only way that this feature can be used is with _variadic_ arguments in _variadic_ functions such as `Display(const string format, ...)`. When the compiler finds a classic array in a variadic argument, it simply pushes a pointer to the first byte and hopes that the receiving function will find a good use for it. 
As far as I know, the "old" compiler did the same. 

The compiler can't check variadic arguments, anyway, because it doesn't know what types to expect. When I realized that, I gave up all misgivings about implementing "decaying classic arrays" as variadic parameters.

```
char ch[100];
Display("%s", ch);
```

Note that it isn't possible in AGS to define an _explicit_ parameter that receives a classic array. However you go about it, you end up defining a parameter that receives a _dynamic_ array. So it currently isn't possible from the get-go to pass a "classic" array to an _explicit_ parameter. 

------

I've taken this as an opportunity to revise and streamline the function calling code in general. And as a result, the code now offers the possibility to use __named arguments__, in `C#` style. 

```
void Heal(this Character *, 
    int added_health, int added_strength, int added_mana = 0)
{ … }

int game_start()
{
    player.Heal(
        added_strength: 30, 
        added_health: 20);
    player.Heal(20, 30); // is equivalent to the code above
}
```

* The way that the functions are __defined and declared__ hasn't changed in any way.
* The "normal" way of calling functions can still be used, and it hasn't changed in any way. Cf. the example above.
* When you use named arguments, each argument has the form `parametername: expression`. The `parametername` determines which parameter gets which expression, respectively. The sequence in which the named arguments are given is immaterial. 
* Currently, there is a restriction: You can't mix the calling conventions. Within one function call, either _all_ the arguments must be named or _none_ of the arguments can be named. 
* This restriction means that you can't use named arguments when calling variadic functions that have variadic arguments (e.g., `Display()`, `String.Format()`, etc.) : Those variadic arguments don't have parameter names, so there is no way to pass them as named arguments.  
* However, functions with optional parameters are fine. When you call a function with optional parameters, your call can name all the arguments that you want to pass, in any sequence, and the compiler will fill in the defaults for all the other parameters. Cf. the example given above.

### DIFF
--- a/Compiler/script2/cs_parser.h
+++ b/Compiler/script2/cs_parser.h
@@ -414,7 +414,7 @@ private:
     Symbol MangleStructAndComponent(Symbol stname, Symbol component);
 
     // Skip over the next symbol in 'src' which _must_ be 'expected', or else an internal (!) error is given.
-    void SkipNextSymbol(SrcList src, Symbol sym);
+    void SkipNextSymbol(SrcList &src, Symbol sym);
     // If the symbol 'actual' isn't in the list 'expected', give an error.
     // 'custom_msg', if given, replaces the "Expected ..." part of the message
     void Expect(SymbolList const &expected, Symbol actual, std::string const &custom_msg = "");
--- a/Compiler/script2/cs_parser.h
+++ b/Compiler/script2/cs_parser.h
@@ -505,7 +505,7 @@ private:
     void ParseParamlist_Param(Symbol name_of_func, bool body_follows, TypeQualifierSet tqs, Vartype param_vartype, size_t param_idx);
 
     // Process a function parameter list
-    void ParseFuncdecl_Paramlist(Symbol funcsym, bool body_follows);
+    void ParseFuncdecl_Parameters(Symbol funcsym, bool body_follows);
 
     void ParseFuncdecl_MasterData2Sym(TypeQualifierSet tqs, Vartype return_vartype, Symbol struct_of_func, Symbol name_of_func, bool body_follows);
 
@@ -543,9 +543,9 @@ private:
     // Check whether there is a type mismatch; if so, give an error. 'msg' for specializing the error message
     void CheckVartypeMismatch(Vartype vartype_is, Vartype vartype_wants_to_be, bool orderMatters, std::string const &msg = "");
 
-    // 'current_vartype' must be the vartype of AX. If it is 'string' and
-    // wanted_vartype is 'String', then AX will be converted to 'String'.
-    // then convert AX into a String object and set its type accordingly
+    // 'current_vartype' must be the vartype of AX. 
+    // If it is 'string' and 'wanted_vartype' is 'String', 
+    // then convert AX into a 'String' object and set its type accordingly
     void ConvertAXStringToStringObject(Vartype wanted_vartype, Vartype &current_vartype);
 
     static int GetReadCommandForSize(int the_size);
@@ -556,20 +556,22 @@ private:
     void EvaluationResultToAx(EvaluationResult &eres);
 
     // We are processing a function call. Emit the actual function call
-    void AccessData_FunctionCall_EmitCall(Symbol name_of_func, size_t params_count, bool func_is_import);
+    void AccessData_FunctionCall_EmitCall(Symbol name_of_func, size_t args_count, bool func_is_import);
 
     // Generate Length attribute for the parsed dynarray type 
     // and add it to the symbol table
     void AccessData_GenerateDynarrayLengthAttrib(EvaluationResult &eres);
 
-    void AccessData_FunctionCall_Parameters_Named(Symbol const name_of_func, std::vector<FuncParameterDesc> const &param_desc, bool const is_variadic, SrcList &params, std::vector<SrcList> &param_exprs);
-    void AccessData_FunctionCall_Parameters_Sequence(Symbol const name_of_func, std::vector<FuncParameterDesc> const &param_desc, bool const is_variadic, SrcList &params, std::vector<SrcList> &param_exprs);
+    void AccessData_FunctionCall_Arguments_Named(Symbol name_of_func, std::vector<FuncParameterDesc> const &param_desc, bool is_variadic, SrcList &arguments, std::vector<SrcList> &arg_exprs);
+    void AccessData_FunctionCall_Arguments_Sequence(Symbol name_of_func, std::vector<FuncParameterDesc> const &param_desc, bool is_variadic, SrcList &arguments, std::vector<SrcList> &arg_exprs);
 
-    // Parse the parameters 'param_list' of a function call
-    void AccessData_FunctionCall_Parameters(Symbol name_of_func, bool func_is_import, std::vector<FuncParameterDesc> const &param_descs, bool is_variadic, SrcList &param_list, size_t &params_count);
+    void AccessData_FunctionCall_Arguments_Push(Symbol name_of_func, bool func_is_import, size_t args_count, std::vector<SrcList> arg_exprs, bool named_args, std::vector<AGS::FuncParameterDesc> const &param_descs);
 
-    // Process a function call. 'param_list' doesn't encompass the surrounding '()'
-    void AccessData_FunctionCall(Symbol name_of_func, SrcList &param_list, EvaluationResult &eres);
+    // Parse the arguments 'args_list' of a function call
+    void AccessData_FunctionCall_Arguments(Symbol name_of_func, bool func_is_import, std::vector<FuncParameterDesc> const &param_descs, bool is_variadic, SrcList &arguments, size_t &args_count);
+
+    // Process a function call. 'arguments' doesn't encompass the surrounding '()'
+    void AccessData_FunctionCall(Symbol name_of_func, SrcList &arguments, EvaluationResult &eres);
 
     // Evaluate 'vloc_lhs op_sym vloc_rhs' at compile time, return the result in 'vloc'.
     // Return whether this is possible.
--- a/Compiler/script2/cs_parser.h
+++ b/Compiler/script2/cs_parser.h
@@ -495,7 +495,8 @@ private:
 
     // We're accepting a parameter list. We've accepted something like 'int'.
     // We accept a param name such as 'i' if present
-    Symbol ParseParamlist_Param_Name(bool body_follows);
+    // 'param_idx' ist the 1-based index of the parameter in the parameter list, for error messages
+    Symbol ParseParamlist_Param_Name(size_t param_idx);
 
     // Additional handling to ParseVardecl_Var2SymTable() that is special for parameters
     void ParseParamlist_Param_AsVar2Sym(Symbol param_name, TypeQualifierSet tqs, Vartype param_vartype, int param_idx);
@@ -506,10 +507,10 @@ private:
     // Process a function parameter list
     void ParseFuncdecl_Paramlist(Symbol funcsym, bool body_follows);
 
-    void ParseFuncdecl_MasterData2Sym(TypeQualifierSet tqs, Vartype return_vartype, Symbol struct_of_function, Symbol name_of_function, bool body_follows);
+    void ParseFuncdecl_MasterData2Sym(TypeQualifierSet tqs, Vartype return_vartype, Symbol struct_of_func, Symbol name_of_func, bool body_follows);
 
     // There was a forward declaration -- check that the real declaration matches it
-    void ParseFuncdecl_CheckThatKnownInfoMatches(std::string const &func_name, SymbolTableEntry::FunctionDesc const *this_entry, SymbolTableEntry::FunctionDesc const *known_info, size_t declared, bool body_follows);
+    void ParseFuncdecl_CheckAndAddKnownInfo(Symbol name_of_func, SymbolTableEntry::FunctionDesc *known_info, size_t known_declared, bool body_follows);
 
     // Enter the function in the 'imports[]' or 'functions[]' array of '_script'; get its index   
     void ParseFuncdecl_EnterAsImportOrFunc(Symbol name_of_func, bool body_follows, bool func_is_import, size_t params_count, CodeLoc &function_soffs);
--- a/Compiler/script2/cs_parser.h
+++ b/Compiler/script2/cs_parser.h
@@ -552,35 +552,24 @@ private:
 
     static int GetWriteCommandForSize(int the_size);
 
-    // Handle the cases where a value is a whole array or dynarray or struct
-    void HandleStructOrArrayResult(EvaluationResult &eres);
-
     // If the result isn't in AX, move it there. Dereferences a pointer
     void EvaluationResultToAx(EvaluationResult &eres);
 
-    // We're in the parameter list of a function call, and we have less parameters than declared.
-    // Provide defaults for the missing values
-    void AccessData_FunctionCall_ProvideDefaults(int func_args_count, size_t supplied_args_count, Symbol func_symbol, bool func_is_import);
+    // We are processing a function call. Emit the actual function call
+    void AccessData_FunctionCall_EmitCall(Symbol name_of_func, size_t params_count, bool func_is_import);
 
-    void AccessData_FunctionCall_PushParams(SrcList &params, size_t closed_paren_idx, size_t func_args_count, size_t supplied_args_count, Symbol funcSymbol, bool func_is_import);
-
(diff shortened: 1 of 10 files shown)


Compiler/test2/cc_bytecode_test_1.cpp
```diff
@@ -3558,3 +3558,62 @@ TEST_F(Bytecode1, Linenum02)
...
+       8,   16,   23,  -999
+    };
+    char fixuptypes[] = {
+      1,   3,   4,  '\0'
+    };
+    CompareFixups(&scrip, numfixups, fixups, fixuptypes);
+
+    int const numimports = 1;
+    std::string imports[] = {
+    "Display",      "[[SENTINEL]]"
```
ivan-mogilko, line 3605: The import here will be registered as "Display^101" now, after the fixed #2300 .
"101" means a variadic function with 1 explicit parameter.

Compiler/script2/cs_parser.h
```diff
@@ -552,35 +552,24 @@ class Parser
...
+    // We are processing a function call. Emit the actual function call
+    void AccessData_FunctionCall_EmitCall(Symbol name_of_func, size_t params_count, bool func_is_import);
 
-    // We are processing a function call. General the actual function call
-    void AccessData_GenerateFunctionCall(Symbol name_of_func, size_t args_count, bool func_is_import);
-
-    // Generate the function call for the function that returns the number of elements
+    // Emit the function call for the function that returns the number of elements
     // of a dynarray.
     void AccessData_GenerateDynarrayLengthAttrib(EvaluationResult &eres);
```
ivan-mogilko, line 563: I did a change to how Array.Length is implemented not a long time ago: #2559

In short, this method no longer generates *the function call*, instead it dynamically generates and registers *attribute and getter entries*, and then lets parser proceed as usual.

Apparently I forgot to fix the comment here. It should be something like "Generate Length attribute for the parsed array type and add it to the symbol table", for example.

ivan-mogilko:
In regards to the named arguments in function call: that's an interesting feature.
It seems to work as expected, and also reports errors if I try to pass something that is not expected by the function (wrong parameter name, or wrong value type).
My big concern about this is that it may be difficult (if at all possible) to adjust our Autocomplete to match this functionality.

---

In regards to passing an array of chars into functions that accept `const string`.
This appears to work in some cases, but not others.
Here are results of my test:

```
function printme(const string text)
{
  Display("%s", text);
}

function game_start()
{
  char c[10];
  c[0] = 'a';
  c[1] = 'b';
  c[2] = 'c';
  c[3] = 0;

  Display(c); // <--- prints "abc"
  Display("%s", c); // <--- prints "abc"
  Display(String.Format("%s", c)); // <--- prints "abc"
  printme(c); // <--- prints empty string (?)
}
```

It looks like "array decay" gets broken in cast between `char[n]` and `const string` in the user script. It works as expected if this cast occurs between user script and engine's function.


Testing dynamic array:
```
  char dc[] = new char[10];
  dc[0] = 'a';
  dc[1] = 'b';
  dc[2] = 'c';
  dc[3] = 0;
  
  //Display(dc); // syntax error, cannot convert char[] to const string (as expected, and matches old compiler)
  Display("%s", dc); // prints "abc"
  //printme(dc); // syntax error, cannot convert char[] to const string (as expected, and matches old compiler)
  Display(String.Format("%s", dc)); // prints "abc"
  printme(String.Format("%s", dc)); // prints "abc"
```
These test results do match old compiler.

Compiler/script2/cs_parser.h
fernewelten: Will enter the comment.

Compiler/test2/cc_bytecode_test_1.cpp
fernewelten: This is an issue!

The bytecode googletests call the parser as-is and then check whether the compiler outputs are exactly what the googletest expects. So the googletest should either have the correct string `"Display^101"` or fail when the compiler calculates `"Display"` erroneously. 

I hope that I haven't mucked up something by mistake. I'm looking into why the function isn't registering correctly at the moment.

fernewelten:
> function printme1(const string text)
> {
>   …
>   printme1(c); // <--- prints empty string (?)
>   printme2(c); // <--- prints empty string (?)
> }

Okay, I'll make a googletest from it and check just what the compiler does with the input.

fernewelten:
> My big concern about this is that it may be difficult (if at all possible) to adjust our Autocomplete to match this functionality.

It would certainly be nice to have the Autocomplete find and autocomplete the parameter names in function calls with named arguments. But I don't see how this could be achieved, either, since the ‘normal’ calling scheme will stay around, too. 

So the user starts typing `i = func(p1:`, and only when the Editor sees the `:` it can conclude that `p1` was meant to be a parameter name instead of a part of an expression. By then it's too late for autocompleting the parameter name. 

The compiler can differentiate between the calling schemes.

ivan-mogilko:
Well, potential Autocomplete's problems is not the compiler's fault, I just mentioned that for the reference.
Dynarray.Length is also not in autocomplete yet (although it could be easier to add).

Compiler/test2/cc_bytecode_test_1.cpp
ivan-mogilko: Maybe there's a misunderstanding, the compiler does register this function correctly as "Display^101" (after #2300).
It's this new google test that has it wrong. Simply changing to "Display^101" here would solve the problem.

fernewelten: Just did that (fix the googletest)

fernewelten: I'm confused.

I've run the following program as a Googletest:
(Named, `CharArrayAsString01` at the end of file `cc_bytecode_test_1.cpp`: 

```
int printme1(const string text)
{
    Display(text);
}

int game_start()
{
    char c[10];
    c[0] = 'A';

    printme1("ab");
    printme1(c);
}

import void Display(const string, ...);
```

(Note that the compiler emits a `zeromem` for `c`, so this array is guaranteed to be `\0` terminated.)

This is what the compiler actually makes of the call of `Display():`
```
   4:   sourceline   3
   6:   load.sp.offs   8      ; move MAR to the location of 'text'
   8:   mov     mar,   ax    ; load AX with that location
  11:   farpush   ax
  13:   setfuncargs   1
  15:   movl    ax,   0    ; import number of 'Display()'
  18:   farcall   ax
  20:   farsubsp   1
```
That looks okay to me.

This is what the compiler actually makes of the calls of `printme1()`:
```
  50:   sourceline   11
  52:   movl    ax,   0    ; constant text block start of 'ab'
  55:   push    ax
  57:   movl    ax,   0    ; address of 'printme()' 
  60:   call    ax
  62:   subi    sp,   4
  65:   sourceline   12
  67:   load.sp.offs   10    ; move MAR to the location of 'c'
  69:   mov     mar,   ax    ; load AX with this location
  72:   push    ax
  74:   movl    ax,   0
  77:   call    ax
  79:   subi    sp,   4
```
Both calls look OK to me, too. 

At which point would the Engine ‘eat’ the contents of `c`?

fernewelten:
I'm confused.

I've run the following program as a Googletest:
(Named, `CharArrayAsString01` at the end of file `cc_bytecode_test_1.cpp`: 

```
int printme1(const string text)
{
    Display(text);
}

int game_start()
{
    char c[10];
    c[0] = 'A';

    printme1("ab");
    printme1(c);
}

import void Display(const string, ...);
```

(Note that the compiler emits a `zeromem` for `c`, so this array is guaranteed to be `\0` terminated.)

This is what the compiler actually makes of the call of `Display():`
```
   4:   sourceline   3
   6:   load.sp.offs   8      ; move MAR to the location of 'text'
   8:   mov     mar,   ax    ; load AX with that location
  11:   farpush   ax
  13:   setfuncargs   1
  15:   movl    ax,   0    ; import number of 'Display()'
  18:   farcall   ax
  20:   farsubsp   1
```
That looks okay to me.

This is what the compiler actually makes of the calls of `printme1()`:
```
  50:   sourceline   11
  52:   movl    ax,   0    ; constant text block start of 'ab'
  55:   push    ax
  57:   movl    ax,   0    ; address of 'printme()' 
  60:   call    ax
  62:   subi    sp,   4
  65:   sourceline   12
  67:   load.sp.offs   10    ; move MAR to the location of 'c'
  69:   mov     mar,   ax    ; load AX with this location
  72:   push    ax
  74:   movl    ax,   0
  77:   call    ax
  79:   subi    sp,   4
```
Both calls look OK to me, too. 

At which point would the Engine ‘eat’ the contents of `c`?

Here's the array of the compiled bytes:
```
int32_t code[] = {
  36,    2,   38,    0,           36,    3,   51,    8,    // 7
   3,    2,    3,   34,            3,   39,    1,    6,    // 15
   3,    0,   33,    3,           35,    1,   36,    4,    // 23
   6,    3,    0,    5,           36,    7,   38,   28,    // 31
  36,    8,   51,    0,           63,   10,    1,    1,    // 39
  10,   36,    9,    6,            3,   65,   51,   10,    // 47
  26,    3,   36,   11,            6,    3,    0,   29,    // 55
   3,    6,    3,    0,           23,    3,    2,    1,    // 63
   4,   36,   12,   51,           10,    3,    2,    3,    // 71
  29,    3,    6,    3,            0,   23,    3,    2,    // 79
   1,    4,   36,   13,            2,    1,   10,    6,    // 87
   3,    0,    5 
};
```

ivan-mogilko:
@fernewelten please dont merge ags4 into your pr, but rebase the pr on top of ags4 branch instead.

fernewelten:
(had to force-push)

ivan-mogilko:
As an update, latest commits fix existing tests, but the problem with passing char array into a script function cast to `const string` still remains. The likely reason is explained in my [previous comment](https://github.com/adventuregamestudio/ags/pull/2583#issuecomment-2498583764).

fernewelten:
> but the problem with passing char array into a script function cast to `const string` still remains.

Yes, I'm still onto that.

fernewelten:
So, I've tried to call up the Editor from within MSVS and compile and run a game from there – something that I haven't done for a very looong time. No luck so far: 
* I've rebuilt Solution/Engine.sln and Solution/AGS.Editor.Full.sln. 
* I've loaded Solution/AGS.Editor.Full.sln and started it in the MSVS debugger (F5)
* The Editor fired up, as it should. I was able to configure a very minimal game.
* But when I started this game within the Editor (F5), I got a Windows message that the procedure entry point `SDL_RenderSetVSync` wasn't found in `…\GAMEFOLDER\_Debug\GAMENAME.exe`. 

image

So this might mean 
* that the Engine doesn't find the library file `sdl2.exe` – although I've checked that it actually _is_ in `…\GAMEFOLDER\_Debug\` as it should 
* or that `sdl2.exe` doesn't contain `SDL_RenderSetVSync` for some reason, maybe due to a version mismatch 
* or that the program only wants to call `SDL_RenderSetVSync` from its _own_ file, where it isn't provided.

So what am I missing to make it work?

ericoporto:
Probably just an old version of the SDL dll. Did you update the libraries to the latest from WinVSdependencies ? Also if I remember correctly the Editor looks for it in it's directory to copy it to the Compiled/Windows dir of the game, so you need to put it there.

fernewelten:
> Also if I remember correctly the Editor looks for it in it's directory to copy it to the Compiled/Windows dir of the game, so you need to put it there.

That seems to be the issue exactly: 
* The `SDL2.dll` from the `_Debug` directory of the test game has 1.273 kB and is from December 2020; 
* my current `SDL2.dll` in `…\agslib\WinDevDependenciesVS\SDL2\lib\x86` has 2.226 kB and is from December 2024.

I'm compiling the Editor in MSVS and calling it up by hitting F5 in MSVS. 
It seems that in this case, this "Editor's directory" is `AGSREPOSITORY\Solutions\.build\Debug\`. I moved file versions into that location that were up-to-date, and that got me in business. 

Thanks!

fernewelten:
Okay, I'm beginning to see what the underlying problem is. We'd have the exact same problem if `c` was a local ‘oldstyle’ `string`, as long as the string is a local variable with local storage – in contrast to a global variable or some literal in quotes. Trouble is, the area especially of ‘oldstyle’ `string`s has never been tested much, the more so since this is a deprecated data type. 

So here we are with a bug that has been in the compiler for at least four years. Oh well. 

I'll get back with a fix.

fernewelten:
So here's the story. I might reword it later if I find a clearer way of expressing myself.

At runtime, memory adresses look like integers, but this is a simplification: Actually each address is a sum of the start of some memory block and an offset into it. For instance, the address of a string could be the start of the block of string literals plus 15, or the start of the block of global variables plus 40. 

In the Bytecode of functions, address literals are given as the offset (e.g., `15`) and a marker that specifies what the offset should be relative to (e.g., the block of global variables). In AGS lingo, this marker is called a ‘fixup’.

Whenever a function is set up to execute, a ‘linker’ assigns specific blocks of memory that the function should use. In other words, it declares: ‘This function should use the block starting at `0x11223344` as global space, the block starting at `0x2211DDEE` as the block of string literals, etc.’ 

Then it runs through the code of that function, and whenever it encounters a cell  that has a marker, it modifies the content of that cell and adds the start of the respective block to it. For instance, when a code cell contains `40` and a marker that says that this is relative to the block of string literals, it modifies the cell so that it contains `0x2211DDEE + 40`, i.e., `0x2211DE16`.

Then afterwards, when the function actually runs, it doesn't need to concern itself with the different blocks, not can it: All its code cells already contain the addition result. When during runtime, a register contains the value, e.g., `0x2211DE16`, then this register tracks whether or not it is loaded with an address. If it _is_ loaded with an address, however, then the information is lost what the start of the address block is and what the offset is. 

So this is why you mustn't push or farpush a ‘naked’ address onto the stack without special provisions. The _called_ function has memory blocks assigned for global variables, stack space etc. that may be different from the blocks assigned to the _calling_ function. The pushed or farpushed value is an addition of ‘start of memory block’ and ‘offset’, but it's potentially the _wrong_ memory block, i.e., the memory block that is correct for the _calling_ function, not the memory block that is correct for the _called_ function. So when the _called_ function pops the value from the stack, it potentially gets an address that points into the wild.

However, when the address is a _string literal_ specifically, then the chance is zero that the called function has a different memory block assigned than the calling function. The reason is, string literals are constant throughout the whole game, they may never be modified. So why should the called function work with a different address room than the calling function? It's more memory efficient to give all functions the same block of constant strings. 

Currently, the compiler emits: 
```
   6:   load.sp.offs   8      ; move MAR to a location relative to the stack block
   8:   mov     mar,   ax      ; load AX with that location, i.e., relative to the stack
  11:   farpush   ax      ; farpush AX to serve as a parameter of a call
```

There it is! All addresses that are relative to `SP` are relative to the stack block in particular. The far function will probably have a different stack block assigned than the calling function. And then the far function won't find the start of the pushed string.

ericoporto:
I don't understand much of this, but from [previous comment](https://github.com/adventuregamestudio/ags/pull/2583#issuecomment-2498583764), I got the impression it was about being able to get an absolute memory location that was figured at runtime. If I understood the issue too is that strings are all saved together in their own block that will never be the same as the function running.

fernewelten:
So I hope that I've squashed the bug now.

ivan-mogilko:
I confirm that the bug is fixed, and judging by the engine, the bytecode contains correct instructions.

Tested named parameters once more, and they still work.

Variadic functions: the first post sais that:
> Currently, there is a restriction: You can't mix the calling conventions. Within one function call, either all the arguments must be named or none of the arguments must be named.
> This restriction means that you can't use named arguments when calling variadic functions that have variadic arguments

To clarify just in case, you can call variadic functions by passing all the fixed parameters, but it's not possible to add variadic parameters after:
```
String.Format(format: "variadic print"); // works
String.Format(format: "variadic print %s", "printme"); // error, cannot mix positional and named arguments in one call
```

Functions with optional parameters: if ALL parameters are optional (have default values), then you can pass any limited number of parameters from any position, e.g. only 4th parameter, or only last parameter.

ivan-mogilko:
Unfortunately this PR caused a new regression: #2610.

PR #2587, opened 2024-11-21
Script API: SendEvent() function, calling "on_event"

ivan-mogilko:
Adds a very simple function:

```
/// Calls the standard "on_event" function, passing EventType and a number of optional parameters
void SendEvent(EventType, int data1=0, int data2=0, int data3=0, int data4=0);
```

This lets to trigger "on_event" callback, passing any arbitrary event type and up to 4 integer parameters (currently).
Declared a `eEventUserEvent` constant inside EventType enum, assigned to a large number. This is meant as a starting index for user scripts to declare their custom events (i.e. eEventUserEvent + 1, eEventUserEvent + 2, and so forth).

As usual, `on_event` is not executed right away, but only after current script finishes running (as a post-script step).

On another hand, since "on_event" is called in every script module, in a regular order (room, then each script from top to bottom), this is also a way to pass commands to scripts that have no direct connection to the current script.

Note: this has no effect on the engine itself, only causes "on_event" to be called with these args. For instance, if you call `SendEvent(eEventRestoreGame)` - that won't actually restore a game. But will make other scripts that handle such event *think* that the game was restored. In other words, this may also be useful to simulate that something happened in the engine.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -537,6 +537,37 @@ enum SortDirection
 };
 #endif // SCRIPT_API_v362
 
+enum EventType {
+  eEventLeaveRoom		= 1,
+  eEventEnterRoomBeforeFadein = 2,
+  // 3 is reserved by an obsolete "death" event
+  eEventGotScore		= 4,
+  eEventGUIMouseDown	= 5,
+  eEventGUIMouseUp		= 6,
+  eEventAddInventory	= 7,
+  eEventLoseInventory	= 8,
+  eEventRestoreGame		= 9,
+#ifdef SCRIPT_API_v36026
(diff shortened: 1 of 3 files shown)


ericoporto:
This is really useful, I have a custom event module I use for some of my games that use the GiveScore function and the score event. Although I used to use them for dialog events that is now done in the engine, so I don't readily have a use, but I could use this in my own modules like in Controlz when the player stops moving, I could emit a eEventPlayerStopped of my own.

ericoporto:
This works, if there is a blocking event, it runs the on_event call right after, which is expected and mentioned in the PR.

In the `///` comment, instead of "Calls", I would use "Triggers", because if there are multiple `on_event` (in different script modules), all of them gets called, in script order, and "Calls" sounds like a specific "on_event" will be called instead.

PR #2588, opened 2024-11-23
Adds Character.Following readonly property

ericoporto:
In Future Flashback, I use the CustomFollow module I showed in the forums, I also have a similar thing in some other unreleased games. In Don't Give Up the Cat I have this line: https://github.com/ericoporto/dont-give-up-the-cat/blob/fc2810d9d580f6e1b51742489fa388fbd96cdbd0/dont-give-up-the-cat/room1.asc#L164

This PR is to be specific to this need. It's not a major need, I can live without it, as I have a workaround, but it feels useful.

The issue I am not sure though is what happened in #2296 , I don't know how to properly return a Character* here that won't fail equality comparisons, Should I return `&game.chars[charid]` directly or the dynamic memory handle? Is ags4 the only version that has a right answer for this?


ivan-mogilko:
You should return simply `&game.chars[charid]`, as that's how Characters are being exported for scripts.
Check `Character.GetAtScreenXY` code for example.

PR #2591, opened 2024-11-26
Upgrade Windows projects to VS 2019

### LINKED ISSUE #2241: Upgrade Windows projects to VS 2019
VS 2015 free Community Edition is deprecated by Microsoft, and is no longer available for download from their official page (although there are still "unofficial" ways to acquire it).

At the same time the latest VS 2022 does not install on Windows 7, which I still use unfortunately. I might have to upgrade to Win10 or 11 soon, as more and more apps stop supporting it in their updates, but that will take some time.

Therefore I suggest to first move to VS 2019, as a temporary measure. According to [this Microsoft's page](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2019) the VS 2019's lifecycle is scheduled to end in 2029.

~As we have a 3.6.1 coming towards a full stable release, I would not like to apply these changes to ags3 branch just yet. Instead I propose to begin with ags4 branch, and apply this change to ags3 only after 3.6.1 is fixed in a separate "release-3.6.1" branch.~ EDIT: 3.6.2 is still in beta, so theoretically we may switch it to 2019.

The changes to the project might include:
1. Upgrade the VS solutions and projects.
2. Fix any new compilation errors and outstanding warnings.
3. Switch to the newer "VC runtime" dependency, make certain that it runs on Windows 7 at least.
4. Replace VC runtime dependency in the AGS Editor's installer, readmes etc.
6. **Low priority:** check if it's still possible to build engine for Windows XP, using a separate configuration.

Task list, for keeping the record:
- [x] Update to VS 2019 in ags3 branch
- [x] Update to VS 2019 in ags4 branch
- [ ] Update projects and make/cmake scripts to C++14.

ivan-mogilko:
Fix #2241

What's done:

1. Updating MSVS solutions is a trivial action, the solution and project files have same format, only few values change inside.
2. Updated PlatformToolset from v140 to v142, Windows SDK version from 8.1 to 10.0. I can at least tell that the compiled programs still work on Windows 7 with these settings.
3. As a separate change, I modified the env variable names in prop files to have "_VS19" postfix. The reasoning for this is to be able to keep both VS2015 and VS2019 version of the project, at minimum while we have to go back to fix bugs in previous release, or compare them. It may be convenient to have separate variables pointing to separate static library paths, where libs are built either with VS15 or VS19.
Examples: `AGS_SDL_SOUND_LIB` -> `AGS_SDL_SOUND_LIB_VS19`, `AGS_SDL_SOUND_LIB_X64` -> `AGS_SDL_SOUND_LIB_VS19_X64`.
And alternative to this would be to use separate library names with "_vs19" prefix, for example.

TODO:
1. Find out which VC Redist should be included into installer.
2. Find out if "_xp" build configurations still work. The last PlatformToolset that has support for WinXP (it seems) is v141_xp related to VS2017, so that's one I set there. I have got a feeling that this may be removed in the newer/future versions of VS.
OTOH, I do not think that winxp support is important at this point, as in all the time since it was moved to a separate config, not a single user have asked for this build.
3. Update CMake for Windows.
4. Update CI scripts (or should be done after this PR?)

### DIFF
--- a/Editor/AGS.Native/NativeDLL.vcxproj
+++ b/Editor/AGS.Native/NativeDLL.vcxproj
@@ -24,32 +24,33 @@
     <RootNamespace>AGS.Native</RootNamespace>
     <Keyword>AtlProj</Keyword>
     <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <UseOfAtl>false</UseOfAtl>
     <CharacterSet>MultiByte</CharacterSet>
     <CLRSupport>true</CLRSupport>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <UseOfAtl>false</UseOfAtl>
     <CharacterSet>MultiByte</CharacterSet>
     <CLRSupport>true</CLRSupport>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <UseOfAtl>false</UseOfAtl>
     <CharacterSet>MultiByte</CharacterSet>
     <CLRSupport>true</CLRSupport>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <UseOfAtl>false</UseOfAtl>
     <CharacterSet>MultiByte</CharacterSet>
     <CLRSupport>true</CLRSupport>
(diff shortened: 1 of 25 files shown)


ericoporto:
> As a separate change, I modified the env variable names in prop files to have "_VS19" postfix

Those are binary compatible - from VS 2015 to VS 2022. There is no need for this.

> The last PlatformToolset that has support for WinXP (it seems) is v141_xp related to VS2017

I don't think this toolset is available in VS2022

> Update CMake for Windows

I don't think these change at all

> Update CI scripts

These have to be made before or this will break the builds.

ericoporto:
Uhm, that is weird, I don't yet until I test more things. Maybe is something C++ specific? Like, SDL2 we use in Windows comes from SDL binary releases and they use the latest MinGW - which I have no idea how it relates to the MSVC stuff, other than that it works?

ivan-mogilko:
SDL2 is dynamically linked, that's a different thing compared to statically linking. When we link dynamically then only function prototypes must match afaik.

ericoporto:
Hey, I asked a person that has more insight on VS development and they gave me the following information

> You can mix VS 2015/2017/2019/2022 toolsets across OBJs, LIBs, DLLs, and EXEs, as long as you obey a couple of important restrictions:
> The toolset used to perform the final link must be the newest toolset involved (or newer).
> The VCRedist you run against (whether installed system-wide or app-local) must be from the newest toolset involved (or newer).
> Full documentation: https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170
> If you can upgrade all the way to VS 2022, that's ideal. Regardless of which version you upgrade to, the more libraries you can rebuild with the toolset, the better (as they'll receive correctness and performance fixes). Upgrading your WinSDK shouldn't be a problem for bincompat. (You will probably have to deal with source compat issues from the compiler upgrade, even with our /std:c++14 and permissive mode mitigations, as well as possibly from the WinSDK upgrade.)
> VS supports targeting old systems, to a certain extent. VS 2022 has dropped support for XP/Vista targeting, as they are totally unsupported by Microsoft and not even receiving security updates for many years. (Chrome, Firefox, Edge, and Steam have all done the same thing.) I believe XP was actually dropped by VS 2019 mid-cycle when we stopped being able to sign SHA-1 certs. VS 2022 will continue to support Win7 for the remainder of its lifecycle but the next major version "Dev18" will increase the minimum system requirement to at least Win8 and possibly Win10 (no further for sure though).

I guess the Dev18 is the future VS2025. I asked about the use of multithread flag (`/MT`). I also mentioned about the desire of Win7 compatibility

> Yeah, static linking (/MT) is totally self-contained, and is a good option if your application is monolithic.
> To clarify, VS 2022 17.x will support Win7 throughout its entire lifecycle, so that’s not a reason to stick to VS 2019.

I guess I can use VS2022 on the CI and it should work with the VS2019 projects while keeping the desired compatibility. But I think WinXP and Vista are out.

ericoporto:
@StephanTLavavej (the person I mentioned in the previous comment) gave me a nice reference for the Win7 support being kept in VS2022 (17.x) and removed only in next version (Dev18): https://github.com/microsoft/STL/pull/4857

ericoporto:
I picked these commits, rebased them, and put them in a new PR, #2640, with a new windows container and CI adjustments that build with the changes from here.

ivan-mogilko:
Superceded by #2640

PR #2592, opened 2024-11-26
Engine: Fix memory leak in ccInstance::CreateEx when it fails

criezy:
This is a regression from commit 1f6abf4. Before that commit it used a std::shared_ptr which means the pointer was automatically deleted.


ivan-mogilko:
That is silly. I wonder why did not I replaced it with unique_ptr in ccInstance::CreateEx, and make it return unique_ptr too. That would make more sense; and same for the Create / Fork methods.

ivan-mogilko:
I fixed this differently, using std::unique_ptr: 30167df0690e0a771a8eb4f14a4d6a72da526b49

PR #2596, opened 2024-11-30
Editor: can drag and drop folders in the Sprite Manager

### LINKED ISSUE #2595: Editor: support dragging folders in the Sprite Manager
This is a very old request, but somehow never had a dedicated issue opened in this tracker.

The SpriteManager currently allows to drag sprites between folders, but not folders. This makes it impossible to rearrange folders or moved folders into other folders.

Therefore it's proposed to support these two operations in the SpriteSelector's window:
- reorder folders among themselves on the same nesting level
- move folders inside other folders.

On implementation.

Similar control scheme has already been implemented in #2362 for Project Explorer.
After a quick glance on the code, it seems possible to pick the minimal necessary drag'n'drop related code out (and any other code that may be shared, although that would be optional for this task), and create a custom control class derived from the TreeView. That class would handle no other data than its folders and items, expanding, collapsing and dragging them around.
EDIT: example of class name: `TreeViewWithDragDrop`.

Any behavior that depends on dragged data should be dealt with either
* by subclassing this custom control into more specialized controls (e.g. ProjectTreeView, SpriteFolderView), or
* by providing events, to which the parent gui (i.e. ProjectTree, SpriteSelector) could hook up and deal with dropping logic.

ivan-mogilko:
Resolve #2595

1. Pick out a drag & drop implementation from ProjectTree into the TreeViewWithDragDrop control class. TreeViewWithDragDrop expands TreeView and provides support for dragging items within itself. It has 3 events: `ItemTryDrag` and `ItemDragOver` and `ItemDragDrop`, which may be used to customize the behavior. They tell whether to start dragging action, if currently dragged item is allowed to be dropped in the current location, and finally handle the drop.
2. Adjust ProjectTree to use TreeViewWithDragDrop type of control. Only 3 drag-related event handlers remain here: `ItemTryDrag`, `ItemDragOver` and `ItemDragDrop`. Everything else was moved to the TreeViewWithDragDrop class.
3. Make SpriteSelector use TreeViewWithDragDrop control for folders, and handle their dragging and dropping among themselves. It's important to keep the previously existing sprites drag n drop as well. So there currently are 2 possible drops: sprites dragged from the sprite list, and folders.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -153,6 +153,7 @@
     <Compile Include="Entities\CharacterIDChangedEventArgs.cs" />
     <Compile Include="Entities\CharacterRoomChangedEventArgs.cs" />
     <Compile Include="Entities\CompilationStepArgs.cs" />
+    <Compile Include="Entities\TreeItemDragEventArgs.cs" />
     <Compile Include="Entities\ExistingScriptHeaderToAdd.cs" />
     <Compile Include="Entities\ExistingFileToAdd.cs" />
     <Compile Include="Config\IObjectConfig.cs" />
@@ -163,6 +164,8 @@
     <Compile Include="Entities\OutputCreationStepArgs.cs" />
     <Compile Include="Entities\ScriptModuleDef.cs" />
     <Compile Include="Entities\GenericMessagesArgs.cs" />
+    <Compile Include="Entities\TreeItemTryDragEventArgs.cs" />
+    <Compile Include="Enums\TargetDropZone.cs" />
     <Compile Include="GUI\AssignToView.cs">
       <SubType>Form</SubType>
     </Compile>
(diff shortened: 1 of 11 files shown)


ericoporto:
I tested it a bit and can confirm it is working!

> Perhaps, CanDropHere may be renamed into CanDropTreeItem, and DoDropHere removed in favor of using DragDrop with some dedicated type of Data..

I am not sure which way is best, but if DoDropHere is meant to only be something that happens with TreeItem, then it could be filtered and it would only be called in the TreeItem case.

ivan-mogilko:
Had to remake few things, as there were couple of bugs which i did not notice at first.

Now there are 3 new events:
* ItemTryDrag
* ItemDragOver
* ItemDragDrop

ericoporto:
OK, I tested this and overall it works fine. The only little detail I picked is if I select a dir, and drag it, once I release it, the selected dir is not the one I dragged.

https://github.com/user-attachments/assets/c4c7b64d-0001-402a-be56-6c533d3d72b3

ivan-mogilko:
Fixed few bugs:

* Forbid dragging folder into itself (or its descendants), - this check is done in TreeViewWithDragDrop class, regardless of whether subscribed handlers check for that.
* Fixed SpriteSelector not selecting dropped folder after drag'n'drop.
* Fixed SpriteSelector not letting to drop into the root folder.

ericoporto:
I confirmed the fixes work in the sprite folders and also that the project explorer (that today I noticed is actually named Explore Project) is also still working.

PR #2601, opened 2024-12-03
Script API: remove 2048 length limit for the variadic text functions

### LINKED ISSUE #2398: Script API: remove 2048 length limit for the variadic text functions
Historically, all the text printing script functions have a limit of resulting string. It was an arbitrary 3000 bytes around 3.6.0, in current master it's 2048, but that's not the point. Because AGS Strings may contain up to INT32_MAX bytes, there should not be any restriction on a length of a string that a user may create using e.g. String.Format.

Printing script args is done using ScriptSprintf function. I propose to fix this function to make it act in compliance with standard snprintf: that is let pass a null buffer in which case it would only calculate and return necessary size. Formatting will be done in two steps: first calculate the necessary buffer length, then allocate a StringBuffer, print text, and create ScriptString, or use this text elsewhere as necessary.

ivan-mogilko:
Fix #2398

This removes an arbitrary 2k chars limit for result of the script formatting API (such as String.Format).

I think there are still cases in the engine where the formatted string remains to be limited, but in most of them the data is assumed to be short (such as getting "location name). The only other notable case is the core function that creates a textual overlay, still where the final printed text is still limited by 2k chars.

### DIFF
--- a/Engine/script/script_api.cpp
+++ b/Engine/script/script_api.cpp
@@ -25,7 +25,6 @@ enum FormatParseResult
 {
     kFormatParseNone,
     kFormatParseInvalid,
-    kFormatParseLiteralPercent,
     kFormatParseArgInteger,
     kFormatParseArgFloat,
     kFormatParseArgCharacter,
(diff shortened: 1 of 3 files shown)


ericoporto:
I tested this with ImGi and a few other games that I have that make use of String.Format and they appear to work the same as before, so this looks alright.

PR #2602, opened 2024-12-04
Editor: fix ProjectTree's unreliable search for nodes

### LINKED ISSUE #2473: Editor: having multiple project folders which names are different only in letter case breaks the item drag-n-drop
**Describe the bug**
If there are 2+ folders in the project tree, which have names different only by letter case (e.g. "Test" and "test"), the Editor gets confused when dragging and dropping an item into one of them: the item will be dropped into the first folder of such name instead of a chosen one.

**AGS Version**
3.6.1 and eariler (seems like an old bug).

**To Reproduce**
1. Open a game project.
2. Open, for instance, Characters node.
3. Create two folders: "Test" and "test".
4. Drag existing character to the folder "test".
5. Notice Character appears in the upper folder "Test" instead.

**Expected behavior**
Dragged and dropped items should end up where user have dropped them.

On a side note, I noticed that folder names are not required to be unique on a given level. Perhaps that may be considered another issue.

ivan-mogilko:
Fix #2473

ProjectTree was using TreeNodeCollection.Find to get a TreeNode with certain Name, but according to MSDN this method does a case-insensitive search for some reason. In AGS Editor currently folder node names (item or folders) are given names that are case-sensitive. Such search could lead to mismatching nodes in case there's e.g. two folders which are only different in name case.

Wrote an extension method FindUnique that filters out precisely matching node name.

NOTE: For the reference TreeNode's Name is not the text that it displays, but its internal "key".
In case user gave two or more folders the same name, that would make their "text" or "title" be the same, but their Names will be made different.

This is achieved with this ugly but working hack:
https://github.com/adventuregamestudio/ags/blob/4e44e71f8258418006af6d9ba99d75fa4a4fada0/Editor/AGS.Editor/Components/BaseComponentWithFolders.cs#L203-L210

### DIFF
--- a/Editor/AGS.Editor/GUI/ProjectTree.cs
+++ b/Editor/AGS.Editor/GUI/ProjectTree.cs
@@ -161,19 +161,19 @@ namespace AGS.Editor
 
         public void RemoveAllChildNodes(IEditorComponent plugin, string parentID)
         {
-            TreeNode[] results = _projectTree.Nodes.Find(parentID, true);
-            if (results.Length > 0)
+            TreeNode node = _projectTree.Nodes.FindUnique(parentID, true);
+            if (node != null)
             {
-                results[0].Nodes.Clear();
+                node.Nodes.Clear();
             }
         }
 
         public void StartFromNode(IEditorComponent plugin, string id)
         {
-            TreeNode[] results = _projectTree.Nodes.Find(id, true);
-            if (results.Length > 0)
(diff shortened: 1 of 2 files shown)


ericoporto:
If the hack is restricted to folders I think it's alright. I think for the things that are in folders some code relies on their node ID being predictable.

ivan-mogilko:
To clarify: this rename hack exists since AGS 3.2.2 (the version that added folders), it's not added by my pr.

Issue #2589, opened 2024-11-25
Editor: clicking on "missing event function" warning should bring up the related object's events table


ivan-mogilko:
After #2556 Editor throws warnings about events that have function names in them, but no actual function defined in script.
These warnings mention object's name, but it would be more convenient if user could double click on it to bring this object up for editing, similar to how you can double click script error to bring up the line in script.

For this to work, we might need a new subclass of CompileMessage, which has something like following:
- object's script name (it's unique in global/room namespace);
- object's typename and id (because script names are not obligatory, sadly);
- parent's identifier, which may be either a gui script name for the gui control, or a room number of anything in the room.

The "Goto definition" menu command already supports going to a global object, that functionality may be shared to some extent.
Going to the rooms will require opening that room. Currently editor only supports opening 1 room at a time, so preferably it should ask if it's okay to load a new one would another be open at the moment.

ivan-mogilko:
UPDATE: this was done for global objects some time ago, but this still does not work for Rooms and room elements.

Issue #2550, opened 2024-10-11
AGS 4: Make linking events to functions more intuitive


messengerbag:
*Problem**
For many years, one of the most common support requests from new AGS users is that they can't get an event handler to run. This is almost always because they've forgotten (or don't realize that they need) to link it to the corresponding event via the event pane. Even experienced users sometimes forget this step, adding delays as they test the game in order to discover and fix the oversight.

There is also an inconsistency in the fact that [global event handlers](https://adventuregamestudio.github.io/ags-manual/Globalfunctions_Event.html) don't need to be linked, but use a predefined name that is called at the right time if it exists.

**Suggested change**
See [previous forum discussion](https://www.adventuregamestudio.co.uk/forums/editor-development/feature-suggestion-auto-link-event-handlers/), where several alternative solutions were proposed, including:

- Detecting unlinked functions that match the default event handler names at build time and displaying a warning
- Automatically link event handlers by default, without a need to do so manually
- Marking linked event handlers in the script editor (in the left-hand column), and a warning icon if not linked 
- Link event handlers explicitly in script rather than through the event pane, using delegates

Perhaps there are new ideas now. One possible "quick fix" is to simply improve the documentation of this feature/quirk, for example in an FAQ section of the manual.

ivan-mogilko:
For the reference, note that after #2534 is merged it is possible to have the handler functions in any module (except for room and room object events). The module is selected by the special property per global object (gui controls share one with their parent GUI).

ericoporto:
The reverse issue is documented in the ticket https://github.com/adventuregamestudio/ags/issues/1363

ivan-mogilko:
I have a uneasy feeling about automatic linking, because if we implement delegates then linking any event with a function  without clear instruction may be contradicting to user's intent.


I'd like to suggest another option: introduce annotations in script, which may be added before the function and instruct editor to link them. This is not necessarily a only solution, but maybe some will like to have.

AlanDrake:
I've always been told that annotations are evil.
Though you are right to be uneasy, the source of uneasiness in the first place was that users had to manually associate functions, which likely led them to create the wrong type of event and later relink it elsewhere.

Were there no need to link manually from the editor, a user could simply name or rename the function event without extra passages.

ivan-mogilko:
And how do you link same function to multiple events without manual linking (or annotations fwiw)?

AlanDrake:
> And how do you link same function to multiple events without manual linking?

[...]

Issue #1439, opened 2021-10-21
Room's compilation messages are treated differently from the script modules


ivan-mogilko:
In short, when the script compiler issues a warning for the regular script module, it ends up in the editor's output window. But when the room script is compiled the warnings are not shown. This is due the difference in handling the compiler's output.

Script modules are compiled within the [AGSEditor.CompileScripts()](https://github.com/adventuregamestudio/ags/blob/a18ff4d84519fc88b42fc6aa38cbe60c25e83f8a/Editor/AGS.Editor/AGSEditor.cs#L1007), while room scripts are compiled within [RoomComponent.SaveRoomOnThread()](https://github.com/adventuregamestudio/ags/blob/a18ff4d84519fc88b42fc6aa38cbe60c25e83f8a/Editor/AGS.Editor/Components/RoomsComponent.cs#L607). The compilation itself is called as `AGSEditor.CompileScript()`. But the code surrounding this call is different between these two cases.

I think either RoomComponent.SaveRoomOnThread should be amended to correctly pass all the information to the output window, similar to how `AGSEditor.CompileScripts()` does this, or there should be a method that wraps a call to CompileScript and  handles the output. Whatever seems more convenient.

ivan-mogilko:
I think this may be related to rooms being saved as a binary blob with both room data and compiled script inside. Thus if you save the room but not recompile the script, then you have got a "inconsistent" room. But this issue should go away as open room format is merged into ags4 (#1278).

There's another consideration, not having to recompile all the rooms when you test-run the game. To avoid this the editor would have to track last change time of each room script, and only compile those that changed.

ivan-mogilko:
UPDATE

While working on #2556 (not merged yet at the time of writing) I have fixed this at least partially, see commit:
https://github.com/adventuregamestudio/ags/pull/2556/commits/4286ad868c2765bd68e54b175fee02117f9ff433

But I've met another problem: when you save a single room, the warnings stay in the output window, but when you do "rebuild all", apparently the output window gets cleared somewhere in the middle of the process, and any room warnings are removed before compilation is complete.

ericoporto:
I think the issue is ShowOutputPanel sets the new messages. It would have to instead be two commands, one for clearing the output panel, and other for appending messages to it. Then one would clear the messages and then begin the compilation process.

ivan-mogilko:
The problem there is how to decide whether to clear it or not.
[...]

PR #2603, opened 2024-12-04
Editor: report warnings about possibly unlinked event handlers

ivan-mogilko:
As suggested by @messengerbag, as a most primitive solution for #2550.
...and a complementary to #2556

This adds another type of post-compilation warning that reports a script function that *looks like* it's a event handler, but *not linked* to the object events.

The feature uses default function names, as configured in Interaction tables.
For example: `cEgo_Look`, `Room_Load` and so forth.

The reported warning uses following pattern:
```Function "ObjectName_EventName" looks like an event handler, but is not linked on ObjType (ID) ScriptName's Event pane```

Example:
warnings--eventhandlers

**TODO:**
Option in Editor Preferences that turns this on and off?

Need to decide what happens if user double clicks the warning. Do we want it to go to the script, or to the object's Events Pane?
Or have both options, called from a context menu?
See also: #2589

Unfortunately, Autocomplete data, which we use for this kind of scanning, contains the index of the function's *starting character*, rather than a *line number*. I think the autocomplete may be amended to also have line numbers for script tokens.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -153,6 +153,7 @@
     <Compile Include="Entities\CharacterIDChangedEventArgs.cs" />
     <Compile Include="Entities\CharacterRoomChangedEventArgs.cs" />
     <Compile Include="Entities\CompilationStepArgs.cs" />
+    <Compile Include="Entities\CompileWarningEx.cs" />
     <Compile Include="Entities\TreeItemDragEventArgs.cs" />
     <Compile Include="Entities\ExistingScriptHeaderToAdd.cs" />
     <Compile Include="Entities\ExistingFileToAdd.cs" />
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -226,6 +226,7 @@
     <Compile Include="GUI\NumberEntryWithInfoDialog.Designer.cs">
       <DependentUpon>NumberEntryWithInfoDialog.cs</DependentUpon>
     </Compile>
+    <Compile Include="GUI\OutputPanelItem.cs" />
     <Compile Include="GUI\Progress.cs">
       <SubType>Form</SubType>
     </Compile>
--- a/Editor/AGS.Editor/Components/AudioComponent.cs
+++ b/Editor/AGS.Editor/Components/AudioComponent.cs
@@ -322,7 +322,7 @@ namespace AGS.Editor.Components
             }
         }
 
-        public override void ShowItemPaneByName(string name)
+        public override bool ShowItemPaneByName(string name)
         {
             IList<AudioClip> audioClips = _agsEditor.CurrentGame.RootAudioClipFolder.GetAllAudioClipsFromAllSubFolders();
             foreach (AudioClip ac in audioClips)
@@ -331,7 +331,7 @@ namespace AGS.Editor.Components
                 {
                     _guiController.ProjectTree.SelectNode(this, GetNodeID(ac));
                     ShowPaneForItem(GetNodeID(ac));
-                    return;
+                    return true;
                 }
             }
 
@@ -342,9 +342,11 @@ namespace AGS.Editor.Components
                 {
                     _guiController.ProjectTree.SelectNode(this, GetClipTypeNodeID(clipType));
                     ShowPaneForItem(GetClipTypeNodeID(clipType));
-                    return;
+                    return true;
                 }
             }
+
+            return false;
         }
 
         private void ImportAudioFiles(string[] selectedFiles)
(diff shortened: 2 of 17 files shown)


ericoporto:
This is great! I think an option in Editor Preferences to turn this off is fine, just important to make sure the default for this is on. If this check is thought as being a build setting, then perhaps it's best that is done in the general settings in the build section.

ivan-mogilko:
Added following:
* Double clicking on "unlinked function" warning zooms into the script file. I was conflicted here about opening an object's event tab, because maybe user would like to see the function itself first, before deciding what to do with it?
* Double clicking on "missing function" warning zooms (at least tries to) into the event tab where this function is assigned.
In order to do this I had to move the parts of "Go to Definition" command code from ScriptEditorBase to GUIController.
Also changed `ShowItemPaneByName` to have a bool return type.

ivan-mogilko:
On a side note, first I checked if it's going to be quick to add "line number" to function entry in AutoComplete, but no. Unfortunately, AutoComplete parser simply does not have a record of "current line number". So adding this would require more work on its own.

On another hand, apparently the "SpeechCenter" plugin does NOT use AGS Editor's AutoCompleteData, as I suspected previously, but has its own script parser. I don't know many editor plugins, and from the ones I heard of this is probably the only one that does thorough script parsing. So idk if there's any editor plugins that would access AutoCompleteData, maybe not.

There's currently a problem with AutoComplete entry classes (ScriptToken and descendants) not using *properties*. They have bare public fields in them, which is really bad for a public interface.
I'm not an expert in C#, and don't quite know how it works in case of dynamic assembly linking (as in case of editor->plugin link).
By experiment I found that if you have a property used by plugin, then *order* of this property inside a class does not matter. This seems different from, for example, C++, where moving a method inside an interface class could change order of entries in a virtual table, and break things. *Maybe* in C# all of these methods and properties are automatically loaded by names, similar to loading a function address by name in C++.

Anyway, I am saying all this, because I think the AutoComplete token classes must be modified to have readonly properties in them, and have the regular fields made private, to secure them from direct access by plugins.

ivan-mogilko:
I went a little further, and moved message-specific behavior to OutputPanelItem class, which is a descendant of ListViewItem. This class wraps CompileMessage, and implements interaction with it.
Initially I had an idea to have subclasses per a type of CompileMessage, but decided to keep a single "base" class for the time being.

OutputPanelItem returns a list of Actions, which OutputPanel fills into the context menu.
There are 2 methods: Action(string command) and DefaultAction(). First is called when user selects a menu item, and second is called whenever user double-clicks an item.

In practice:
- a "Missing handler" message shows "Go to Object" menu command;
- a "Possibly unlinked handler" message shows both "Go to Script" and "Go to Object" menu commands.

ericoporto:
This works, but I think there is some weird interaction in case of room scripts. If a room script has like an oObject_Interact and it is not linked to the object, it will show the message, double clicking it will go to the script - I understand, I saw the issue to address the problem separately - but the big issue is that it then clears all warning messages. I think this is just some other bug in AGS Editor, I believe this is part of #1439.

So in summary, it looks like this works, but particularly for room scripts, it is hard to follow reported warnings. But this looks very useful, and the issue looks like something that can be fixed.

ivan-mogilko:
> it will show the message, double clicking it will go to the script - I understand, I saw the issue to address the problem separately - but the big issue is that it then clears all warning messages.

Do you mean that it clears instantly after opening a room script, or after some other action from your side?

ericoporto:
Ah, I understand now, it is that I instinctively save after editing. I always ctrl+s after editing a file, which makes the room file rebuild, which then clears all warnings.

https://github.com/user-attachments/assets/41aa0340-da19-4e04-98cd-602dce3af37f

ivan-mogilko:
> Ah, I understand now, it is that I instinctively save after editing. I always ctrl+s after editing a file, which makes the room file rebuild, which then clears all warnings.

Hmm, this behavior on video looks strange.
So you are in a room script of a room which had warnings reported for, you save, but all the warnings disappear, including the warnings related to that room.
I can see how it may be annoying that all the other room warnings disappear. That may a be an issue, but it's related to the output panel behavior.

But I think that *current room* warnings should reappear again, since you did not fix them and saved again. Probably the handler checks are skipped in some case.

ivan-mogilko:
> Do I remember right that in ags3 rooms are recompiled when their scripts are saved?

So, that's where this issues comes from. Saving just the room script does *not* recompile neither the script nor the room, and yet the output pane is cleared.

But furthermore, there's no such thing as "save room script" or "save script" in AGS Editor. There's just "Save" that saves whole game... That's why it clears the output.

ivan-mogilko:
Well, it seems that the remaining issues are general issues of the rooms and output panel, so i'll merge this feature.

PR #2604, opened 2024-12-05
Editor: show custom properties in Property Grid

ericoporto:
This almost works. Still need to figure out why if the properties are set to the default value they don't show at all. May need entirely different strategy to account for this... I imagine we do some crazy dance to not save custom properties that are equal to default value.

What I imagine is this code is going to have to be much more complicate to account for it, it will need to somehow match the object type (Room, Object, ...) with the properties, then get a list of the properties for that object type in the schema, and fakely give a list of properties, then for every set if it matches the default value it needs to behind the scenes remove it from the list of properties and also if it doesn't then it needs to add and set the value. So CustomProperties themselves will also need to have access to the schema.

And there is some bug in GitHub vs Cirrus right now, the CI is running here: https://cirrus-ci.com/build/6162051992649728

This PR is from the idea here: https://www.adventuregamestudio.co.uk/forums/index.php?topic=62107.0


ericoporto:
Apparently there is a code for this already from CJ

https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/CustomPropertyDescriptor.cs

I have no idea how to use that instead yet. I am confused.

ivan-mogilko:
> Apparently there is a code for this already from CJ
> 
> https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/CustomPropertyDescriptor.cs
> 
> I have no idea how to use that instead yet. I am confused.

Hmm, never paid attention that this exists either. Maybe he started it but never finished.

From the looks of it, his CustomPropertyDescriptor accepts Schema Item for the particular property in constructor, which is probably makes it more flexible than if we attach schema as a static property (if that actually can work).

ericoporto:
apparently CJ was doing something different here: https://github.com/adventuregamestudio/ags/blob/master/Editor/AGS.Editor/GUI/CustomPropertyBag.cs

I don't know if this is actually used somewhere. I guess for the property grid I would put the ICustomTypeDescriptor in the CustomProperties themselves (erh, like I did). And I think this all has to be in AGS.Types too.

ericoporto:
I got it working but I don't know what to do with the files in AGS Editor, I think they are maybe actually used in the property grid that is inside the Custom Properties GUI when editing them.

Editor/AGS.Types/CustomProperties.cs
```diff
@@ -1,16 +1,25 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
```
ivan-mogilko, line 4: It looks like System.Runtime.InteropServices is not needed here. I paid attention because it's a specific namespace for marshalling memory data.

Editor/AGS.Types/PropertyGridExtras/CustomPropertyDescriptor.cs
```diff
@@ -1,126 +1,128 @@
...
-            return true;
-        }
-    }
-
-}
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text;
+using System.Threading.Tasks;
```
ivan-mogilko, line 5: System.Threading.Tasks is not needed here also.

ivan-mogilko:
Started testing this, and following are few first observations.

1. Inventory Items display Character's custom properties instead of inventory ones. This is a strange bug, because it seemed to work correctly at first, but then worked for some items but not others, and stopped working completely after I restarted the editor.
2. Adding a property in schema does not update selected object's properties right away - you have to switch to another object and back. Probably may be fixed by doing Factory.GUIController.RefreshPropertyGrid() somewhere.
3. Custom properties with default value are not indicated as such, by having a non-bold font. Is there a way to attach "DefaultValueAttribute" to these generated properties?
4. Since there's a full list of custom properties shown in the grid, I think it's better to change the display name for the Category and main property to "Custom Properties" or "User Properties" etc. Something to clarify the difference between them and all the other properties.


In regards to previously mentioned CustomPropertyBag, - my understanding is that it's used as a "wrapper object" to display custom properties inside the "Edit custom properties" dialog.

Editor/AGS.Types/CustomProperties.cs
ericoporto: must be some left over, removed

Editor/AGS.Types/PropertyGridExtras/CustomPropertyDescriptor.cs
ericoporto: this is probably leftover from some hack, removed

ericoporto:
> Since there's a full list of custom properties shown in the grid, I think it's better to change the display name for the Category and main property to "Custom Properties" or "User Properties" etc. Something to clarify the difference between them and all the other properties.

Uhm, doing this moves it to the top, as C comes before all other category names.

> Adding a property in schema does not update selected object's properties right away - you have to switch to another object and back. Probably may be fixed by doing Factory.GUIController.RefreshPropertyGrid() somewhere.

Done!

> Custom properties with default value are not indicated as such, by having a non-bold font. Is there a way to attach "DefaultValueAttribute" to these generated properties?

I had that correctly set but apparently it uses a weird serialize check for this, adjusted.

> Inventory Items display Character's custom properties instead of inventory ones. This is a strange bug, because it seemed to work correctly at first, but then worked for some items but not others, and stopped working completely after I restarted the editor. Other types of objects seem to display correct set of properties, in accordance to the schema.

It took me a while to figure this, I was setting it in the constructor, moved to set in the declaration so it sets correctly independent of the different constructors available. Done!


> EDIT: and yes, collapsing custom properties it is a very annoying behavior when you want to edit multiple objects in a row.

I haven't figured any idea here yet. Other approach would be to instead figure some hack to have them in the top level instead. Other idea would be to always expand all things in the property grid - it doesn't give a fine control onto what to expand.

ericoporto:
I tried to move the custom properties to the top level, to see if I could avoid the need for expanding them manually, but it doesn't work

https://github.com/ericoporto/ags/commit/63c79216430a8edd9bc2d6d639b7bede06320952

It will instead create a weird new collection under CustomProperties

image

ericoporto:
I can squash the commits if necessary. This ended up being a lot more code changes than I antecipated.

Editor/AGS.Types/Character.cs
```diff
@@ -45,7 +45,7 @@ public class Character : ICustomTypeDescriptor, IToXml, IComparable<Character>
         private bool _adjustSpeedWithScaling;
         private bool _adjustVolumeWithScaling;
         private bool _movementLinkedToAnimation = true;
-        private CustomProperties _properties;
+        private CustomProperties _properties = new CustomProperties(CustomPropertyAppliesTo.Characters);
```
ivan-mogilko, line 48: This fix does not work, because _properties member is re-created inside SerializeUtils.DeserializeFromXML call, and its "applies to" gets reset to the Character once again.

ericoporto: If I understood, the Xml constructor sets the Properties property that has the CustomProperties, so I just added an additional set after to force the appliesTo to be correct in https://github.com/adventuregamestudio/ags/pull/2604/commits/c1ce7723c7f940a5984ed4f5c18dc3a377486094

Editor/AGS.Types/CustomProperties.cs
```diff
@@ -15,6 +15,12 @@ public class CustomProperties : ICustomTypeDescriptor
 
         static public CustomPropertySchema Schema;
 
+        [AGSNoSerialize()]
+        public CustomPropertyAppliesTo AppliesTo
+        {
+            set { _appliesTo = value; }
```
ivan-mogilko, line 23: Please add a getter here too, it's strange to see a property that has only setter.

ericoporto: Added in https://github.com/adventuregamestudio/ags/compare/2f81b98829f5c1d9e45ede5c740e4d7d07f7aec4..1a96c44d261da4dc6c5d884a9287f4690389e8b7 , but no one should never get it afaict.

ivan-mogilko:
> > custom Attribute that would define "applies to"
> afaict this would also be overwritten when things are set.

No, attributes belong to the property, not the assigned object, so they are not overwritten. But then, there's of course a question of how the property attribute would propagate its meaning to the TypeDescriptor.

> The other bad thing is this enum zero index should be UnknownType so we could at least throw an exception instead.

In ags4 I remade this enum to flags:
https://github.com/adventuregamestudio/ags/blob/ags4/Editor/AGS.Types/Enums/CustomPropertyAppliesTo.cs
the numeric values are not serialized, only constant names, so that did not break existing projects.

ericoporto:
The only approach I can think of is if we had templates and then the type of the CustomProperty itself could be something like CustomProperty<Character> or something like this. I don’t have a clear picture on this right now though. But imagine if `CustomPropertiesCharacter : CustomProperties` was a thing and we could get the inherited type from its base.

ivan-mogilko:
Mmm, I doubt this issue alone justifies having separate classes, that might result in lots of excessive types.

Anyway, the PR is working now, so I'll merge this. Any further improvements may be done in a later time.

PR #2607, opened 2024-12-08
Editor: add Length to Dynamic Array autocomplete

### LINKED ISSUE #2599: Add Length to Autocomplete
Length read-only property is now supported for Dynamic Arrays by the AGS Script compiler in v.3.6.2. It needs to be added strictly to dynamic arrays in autocomplete.

When a Script Variable is added to the Autocomplete it contains information if it is or isn't array, but not if it's dynamic

https://github.com/adventuregamestudio/ags/blob/583ff7a0f485de3db8deae689eaee335af1edcdc/Editor/AGS.Editor/AutoComplete.cs#L615

During parsing this information is found and set in the Autocomplete state as DynamicArrayDefinition. It's just that such information isn't stored in the variable after.

The other issue is figuring where/who should add Length as a property to dynamic arrays. If it's in some AutoCompleteFixUps that runs over at the end, with Length hardcoded, it would relatively easier to add any new similar things that may appear. A description text may optionally be added too.

ericoporto:
fix #2599

This is working, please try.

~~this is the lowest effort for having Length in autocomplete for dynamic arrays. It looks like there are some refactoring opportunities both in the Autocomplete parser and int the typing parser in the Scintilla Wrapper, but they weren't clear for me.~~ I guess we switched for the highest effort.

~~One thing I am not doing here is adding the Length property to the call tip that shows up when the mouse over some property, I looked at the code and couldn't figure some nice way to add it there.~~ this was done

~~This adds a method `GetMembersOfDynamicArray()` that returns a hardcoded Scintilla formatted `Length` as property~~  Edit: switched for fakestruct approach.

### DIFF
--- a/Editor/AGS.Editor.Tests/Editor/AutoComplete/AutoCompleteTests.cs
+++ b/Editor/AGS.Editor.Tests/Editor/AutoComplete/AutoCompleteTests.cs
@@ -94,21 +94,56 @@ int a;
             Assert.That(scriptToTest.AutoCompleteData.Variables.Count, Is.EqualTo(1));
         }
 
+
         [Test]
         public void ContainsArray()
         {
             string scriptCode = $@"
-int arrOfInts[] = new int[100];
+int staticArrayOfInts[100];
 ";
             Script scriptToTest = CachedAutoCompletedScriptFromCode(scriptCode);
 
             Assert.That(scriptToTest.AutoCompleteData.Variables.Count, Is.EqualTo(1));
 
-            ScriptVariable variable = scriptToTest.AutoCompleteData.Variables.FirstOrDefault(v => v.VariableName == "arrOfInts");
+            ScriptVariable variable = scriptToTest.AutoCompleteData.Variables.FirstOrDefault(v => v.VariableName == "staticArrayOfInts");
             Assert.That(variable, Is.Not.Null);
-            Assert.That(variable.Type, Is.EqualTo("int"));
+            Assert.That(variable.Type, Is.EqualTo("int")); // static array type is only base type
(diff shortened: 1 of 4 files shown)


ericoporto:
I switched for your branch here, I put the previous approach in other branch: https://github.com/ericoporto/ags/tree/add-length-autocomplete-bkp.

ivan-mogilko:
Hmm, tooltips work both in my branch and your pr for me (ah, that's practically same branch)...

But I found a silly mistake here:
```
ScriptVariable lengthVar = new ScriptVariable("Length", trueType, false, false, false, null, null, false, false, false, false, 0);
```
I made "Length" not "int" but the type of array.

ericoporto:
Not sure what was going on before, but the tooltip is working. I also fixed the previous mentioned issue and squashed it in the first commit, and adjusted the tests to match the new implementation.

> On a side note, that's the first time I realized that Autocomplete does not distinguish "readonly" variables, and tooltip does not mention "readonly" either.

The issue here is that there is no IsReadOnly property in the ScriptVariable class. That class has the biggest list of parameters already so I am not going to add another one now, lol, but I think the approach would be to add that information there, using `IsWordInPreviousList("readonly")` and in case of length just forcing it there through the constructor.

ericoporto:
I fixed the previous mentioned issue and squashed it in the first commit, and adjusted the tests to match the new implementation.

> On a side note, that's the first time I realized that Autocomplete does not distinguish "readonly" variables, and tooltip does not mention "readonly" either.

I made a new commit for this. The way it works is it shows the type name that owns the property, so it will look like this. 

image

ivan-mogilko:
Looks like working.

Issue #2436, opened 2024-05-26
Script API: Reform Debug() function


ivan-mogilko:
There's a Debug() function in script, which purpose is to provide several random information for diagnostic purposes.
The detailed description may be found in the manual: https://adventuregamestudio.github.io/ags-manual/Globalfunctions_General.html#debug

In brief, it can do following:
1. Give all inventory.
2. Display some information about runtime engine and display mode.
3. Display room's mask on a half-translucent overlay (in older versions it displayed an opaque overlay which blocked the rest of the game).
4. Teleport to another room.
5. Toggle FPS counter.
6. Display character's current walking path.

There are 2 major issues with this function:
1. For a diagnostic script command it has a inconvenient and limiting interface, and hardcoded behavior. This makes it difficult to customize, changing the way it displays information, etc.
2. As a script command, this feature relies on game authors actually adding it to their game. This means that we cannot use this feature if we testing/debugging a game where this command is not present in the game scripts.

Considering above makes me think that this command should not be in the script API. Instead, there are two things that may be done in parallel:
1. For a diagnostic script API, we should rather implement replacements that allow more precise information retrieval and fine tuning when displaying it.
2. For the engine's test runs, we should rather implement commands that work regardless of the game script; for instance, hotkey combinations that trigger if game is run in debug mode (`--test`) and display wanted information independent from the rest of the game logic.

---

Let's begin with the scripting first. If we look at existing options of `Debug()` function, these are available alternatives:
Debug option | Alternative
---|---
Give all inventory | Trivially done by doing `player.AddInventory()` in a loop, iterating all inv items.
Display runtime info | There's `System.RuntimeInfo` for a while now, but it also relies on hardcoding a formatted string. A better option is recently added `System.GetEngineString` (see #2358).
Display room mask | May be done by getting a mask as a DrawingSurface, e.g. see `[GetDrawingSurfaceForWalkableArea](https://adventuregamestudio.github.io/ags-manual/Globalfunctions_Room.html#getdrawingsurfaceforwalkablearea)`, and newer alternatives in AGS 4.
Teleport to another room. | Teleport itself is trivial `player.ChangeRoom()` command, but we're lacking a way to give a list of room names in script. A workaround is to hardcode array of room names in a global script instead, but this is less convenient.
Toggle FPS counter. | None (but there's a command-line arg).
Display character's current walking path. | None (a solution may be related to #1979).

ericoporto:
[...]

PR #2608, opened 2024-12-08
Engine: Add System.DisplayFPS

ericoporto:
This feels easier to use than remembering to do `Debug(4, 1)`.

This is picked from the discussion in #2436

I use `Debug(4, 1)` when evaluating performance of something I am making, with this I would instead do `System.DisplayFPS = true`, which is more readable, or a simple `on_key_press() { System.DisplayFPS = !System.DisplayFPS }`. This is more for testing quick things than actual final games, where one would probably use GetEngineInteger to retrieve proper fps and do a custom display.


ivan-mogilko:
This works fine.

PR #2611, opened 2024-12-11
New compiler: Fix bug: Don't skip the part of an expression after a function call

### LINKED ISSUE #2610: New compiler: chained access expressions broken
CC @fernewelten 

Following kind of expressions appear broken after the recent PR #2583:

```
int id = Hotspot.GetAtScreenXY(0, 0).ID;
```

Compiler reports error:
> Cannot assign a type 'Hotspot *' value to a type 'int' variable

Another case:
```
import String Func();

function game_start()
{
    int x = Func().AsInt;
}
```

> Cannot assign a type 'String' value to a type 'int' variable

We should also add this syntax case into the compiler tests.

fernewelten:
Fixes #2610

When an expression contains a function call, the part after the function call must not be skipped. 

In the following example program, `.ID` must not be skipped.
```
function game_start
{
    int id = Hotspot.GetAtScreenXY(7, 0).ID;
}
```

Add a Googletest that checks for this.


ivan-mogilko:
Thank you, this fixed it.

PR #2614, opened 2024-12-11
Fix bug (2): Attribute of function call result must be callable

### LINKED ISSUE #2610: New compiler: chained access expressions broken

fernewelten:
Fixes #2610

In an expression, function calls can be first in a chain of '.' components or non-first. In either case, the part after the function call must not be skipped.

Also fix a small bug in a Googletest: A compile was supposed to fail due to trailing symbols after a constructor; but it failed because a function call didn't have the parameter specified.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -1988,15 +1988,8 @@ void AGS::Parser::ParseExpression_New_CtorFuncCall(Symbol argument_vartype, SrcL
         _reg_track.SetRegister(SREG_MAR);
         EvaluationResult eres_dummy;
 
-        // Get the arguments of the constructor call
-        size_t const start_of_arguments = expression.GetCursor();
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Alright, this seems to have finally fixed the problem.

I tested using long expressions like
```
int array_length = GetSomeCharacter().ScriptName.Append(".word").Split(".").Length;
```

and compiled a big game project, which helped me to notice this problem earlier.

PR #2619, opened 2024-12-12
AGS 4.0: implement Overlay.Flip

ivan-mogilko:
By user request, this implements Overlay.Flip property that flips overlay graphic in 3 possible ways.

```
/// Gets/sets the flip direction of this overlay.
attribute eFlipDirection Overlay.Flip;
```

For storage and serialization this uses a new field of type "SpriteTransformationFlags", and a position which was reserved in ags4 save format a while ago. It has flip flags now, but more could be added later as necessary. There's a conversion between script parameter and this new flag set.

NOTE: Software renderer has some bugs in ags4 branch, so switching from some flip to "no flip" does not update the overlay image. I'm planning to address that separately, there's a unfinished code brought by one of the older merges, that was never made to work in completion.


ivan-mogilko:
Rebased on top of #2620 in sake of convenient testing. Only one final commit belongs to the Overlay.Flip feature.

PR #2620, opened 2024-12-13
AGS 4.0: Support all 3 flip styles in ViewFrame

ivan-mogilko:
This lets set any kind of Flip style to a ViewFrame, affecting anything that uses a viewframe (including animating buttons since #1944).

(Probably should have based #2619 on top of this...)

* Updated Direct3D and OpenGL renderers to be able to flip texture all ways.
* Added SpriteTransformationFlags that may hold a number of sprite flags together. Currently they only have flip types.
* ViewFrame's "flags" field is now classified as SpriteTransformationFlags. Removed the only VFLG_FLIPSPRITE flag. There's no backwards compatibility issue here, because luckily VFLG_FLIPSPRITE is matching our new FlipX flag. (If we need separate flags for something ViewFrame specific, we may add another field, and there are 2 reserved ints in it, so we won't have to change serialization format in such case)
* Adjusted all drawing functions that had "flipped" or "mirror" boolean to have GraphicFlip parameter instead. There are functions that cast between GraphicFlip and SpriteTransformationFlags, in case we want to "extract" only the flip style. NOTE: I was a little stupid and somehow missed the fact that GraphicFlip is already suitable to be treated as flags with exact same values (by accident). I might adjust the code to reflect that.
* In the Editor deprecated ViewFrame.Flipped property and added "SpriteFlipStyle Flip" property to be used instead. Added conversion for project upgrade.
* ViewFrame.Flipped in script API is converted to eFlipDirection. This is backwards compatible, because old boolean value match "No flip" and "Horizontal flip" respectively. I decided to keep the property name in sake of users convenience, since it's still suitable as it is.

NOTE: In the Editor, "Flip frame" command is kept to flip horizontally only. I was not certain how to deal with it. Vertical flip is not very common, so maybe it should have a separate command and key shortcut.

---

POSSIBLE NEXT THINGS:
Overlay.Flip: #2619
Add Button.GraphicFlip, to mimic this addition from @AlanDrake 's Draconian version: https://github.com/AlanDrake/ags-draconian/commit/58ea65dd39d6a568b792e72f5b44d642d6e9ee96

### DIFF
--- a/Common/ac/view.cpp
+++ b/Common/ac/view.cpp
@@ -17,18 +17,6 @@
 
 using namespace AGS::Common;
 
-ViewFrame::ViewFrame()
-    : pic(0)
-    , xoffs(0)
-    , yoffs(0)
-    , speed(0)
-    , flags(0)
-    , sound(-1)
-{
-    reserved_for_future[0] = 0;
-    reserved_for_future[1] = 0;
-}
-
 void ViewFrame::ReadFromFile(Stream *in)
 {
     pic = in->ReadInt32();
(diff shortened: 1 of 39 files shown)


ivan-mogilko:
Had to rebase after fixing software renderer in the ags4 branch, and also couple of mistakes in this pr.

Issue #1409, opened 2021-08-27
Let's implement delegates, as a callback and function pointer mechanism


fernewelten:
**Describe the problem**
Currently, the only callback mechanism we have is `CallRoomScript()` and the event `on_call`, which is somewhat rustic.

**Suggested change**
I suggest implementing delegates, as in `C#` (except that AGS doesn't have the `public` keyword).

## How it would be used, from a coder's perspective:
The module writer codes, in the header file:

```
// I want to work with callbacks that take String as an argument. 
// Define the type of the thing that holds these callbacks
delegate int DelegateWithOneString (String msg);

// I'll call this whenever the player dies;
import DelegateWithOneString WheneverPlayerDies;
```

The module writer codes, in the code file:

```
// Define the delegate in global space and export it
DelegateWithOneString WheneverPlayerDies;
export WheneverPlayerDies;

function foo()
{
	...
	// Oops! Player has died, better notify everyone by calling the delegate
	WheneverPlayerDies("Has touched a live wire");
	...
}
```

The module user codes, in GlobalScript.ash:
```
// Whenever the player dies, I'll arrange that  function 'MyPlayerDiesCallback()' gets called
import int MyPlayerDiesCallback(String message);
```

The module user codes, in GlobalScript.asc:
```
int MyPlayerDiesCallback(String message)
{
	Display("Well, guess what: %s", message);
	Display("You Have Died.");
	return 0;
}

function game_start() 
{
	// Make the function 'MyPlayerDiesCallback' be called whenever 'WheneverPlayerDies' is called.
	WheneverPlayerDies += MyPlayerDiesCallback;
}
```

## What will happen behind-the-scenes:

- Internally, a delegate is a dynamic array that contains the functions that the delegate must call.
- Whenever a function is added to the delegate, it is linked (FIXUP_IMPORT) and then moved into a managed struct of bookkeeping information. A pointer to this struct is then added to the end of the dynamic array.
- Whenever a delegate is called, all the functions in the dynamic array are called in turn. The result of the last function call is returned to the delegate caller.
- ~~The delegate can be serialized and deserialized in the standard way (the dynamic array is (de)serialized and the respective structs are (de)serialized).~~ __edit:__ See dedicated comment below for suggested data structure

## Limitations that I foresee

[...]

PR #2621, opened 2024-12-14
[WIP] AGS 4.0: rewrite ccInstance into non-forking ScriptExecutor

ivan-mogilko:
This is an experiment in a WIP stage currently.

**Purpose:**
* reorganize ccInstance class, separate "runtime script data" and "script execution";
* make it work without "forks", which are currently made to let run "repeatedly_execute_always" while another script is suspended;
* as a consequence, in theory support any number of nested script calls made through the nested engine API calls, i.e. interleaved engine -> script -> engine -> script -> etc, which may be wanted for delegates/function pointers to work (see discussion in #1409).

**What is done**
1. The former organization ccScript - ccInstance is expanded into: ccScript - RuntimeScript - ScriptExecutor, where
    * ccScript is a loaded script data;
    * RuntimeScript is a script prepared for execution, with fixups and imports resolved, and linked with other scripts;
    * ScriptExecutor is a object that executes RuntimeScripts.
2. Engine no longer keeps ccScripts, only use them for creating RuntimeScripts.
3. RuntimeScripts don't make any "forks", there's strictly 1 per each script, and no need for more.
4. There's a single ScriptExecutor object, which tracks a callstack of RuntimeScripts passed into it, capable of starting nested script runs while there's another/other script execution suspended.

In the situation where a script would call a blocking action, such as Wait() function, for example,
previously engine would do this:
* run base ccInstance for some script event
* call Wait()
* run forked ccInstance for "repeatedly_execute_always"

after this change it will do:
* call ScriptExecutor::Run for some script event
* call Wait()
* call ScriptExecutor::Run for "repeatedly_execute_always"

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -213,7 +213,7 @@ HGameFileError OpenMainGameFileFromDefaultAsset(MainGameSource &src, AssetManage
     return OpenMainGameFileBase(src);
 }
 
-HGameFileError ReadDialogScript(PScript &dialog_script, Stream *in, GameDataVersion data_ver)
+HGameFileError ReadDialogScript(UScript &dialog_script, Stream *in, GameDataVersion data_ver)
 {
     dialog_script.reset(ccScript::CreateFromStream(in));
     if (dialog_script == nullptr)
@@ -221,7 +221,7 @@ HGameFileError ReadDialogScript(PScript &dialog_script, Stream *in, GameDataVers
     return HGameFileError::None();
 }
 
-HGameFileError ReadScriptModules(std::vector<PScript> &sc_mods, Stream *in, GameDataVersion data_ver)
+HGameFileError ReadScriptModules(std::vector<UScript> &sc_mods, Stream *in, GameDataVersion data_ver)
 {
     int count = in->ReadInt32();
     sc_mods.resize(count);
--- a/Common/game/main_game_file.h
+++ b/Common/game/main_game_file.h
@@ -118,9 +118,9 @@ struct LoadedGameEntities
     GUICollection           GuiControls;
     std::vector<DialogTopic> Dialogs;
     std::vector<ViewStruct> Views;
-    PScript                 GlobalScript;
-    PScript                 DialogScript;
-    std::vector<PScript>    ScriptModules;
+    UScript                 GlobalScript;
+    UScript                 DialogScript;
+    std::vector<UScript>    ScriptModules;
     std::vector<PluginInfo> PluginInfos;
 
     // Original sprite data (when it was read into const-sized arrays)
--- a/Common/game/roomstruct.cpp
+++ b/Common/game/roomstruct.cpp
@@ -11,7 +11,6 @@
 // https://opensource.org/license/artistic-2-0/
 //
 //=============================================================================
-
 #include "ac/common.h" // quit
 #include "game/room_file.h"
 #include "game/roomstruct.h"
--- a/Common/game/roomstruct.h
+++ b/Common/game/roomstruct.h
@@ -41,13 +41,10 @@
 #include "ac/common_defines.h"
 #include "game/interactions.h"
 #include "gfx/gfx_def.h"
+#include "script/cc_script.h"
 #include "util/error.h"
 #include "util/geometry.h"
 #include "util/string_types.h"
-
-struct ccScript;
-struct SpriteInfo;
-typedef std::shared_ptr<ccScript> PScript;
 
 // TODO: move the following enums under AGS::Common namespace
 // later, when more engine source is put in AGS namespace and
@@ -316,10 +313,9 @@ public:
     // Event script links
     UInteractionEvents      EventHandlers;
     // Compiled room script
-    PScript                 CompiledScript;
+    UScript                 CompiledScript;
     // Various extended options with string values, meta-data etc
     StringMap               StrOptions;
-
 };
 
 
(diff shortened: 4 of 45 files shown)


ivan-mogilko:
I think we can push these changes further, and introduce a "Script Thread" class. The "Script Thread" will contain its own data stack and call stack, leaving only registers in the ScriptExecutor. ScriptExecutor will be told to run this or that script on a specific "script thread".

Then there will be 2 default "threads": the "main" one, and a "non-blocking" one to run "rep-exec-always" on. This will nicely mimic the previous behavior.

But the "script thread" objects may become more useful somewhen in the future if we try to implement coroutines in AGS.

ericoporto:
I see there are changes to the script compiler, does this change requires rebuild of the auto test game with it or should be able to run the previously built ags4 auto test game ?

ivan-mogilko:
> I see there are changes to the script compiler, does this change requires rebuild of the auto test game with it or should be able to run the previously built ags4 auto test game ?

No, there are no changes to compiler itself, there are only corrections to types in ccScript.

ivan-mogilko:
Just a small note, this pr is working when the engine is run on its own, but it's crashing whenever run from IDE. Possibly the debugger is broken. I did not have time to look into this yet, but will finish this in the new year. Besides the mentioned bug there are couple of modifications that I wanted to make. And then this also will need to be tested for performance. The main execution loop did not change, it's mostly just copied from one class to another, but it's better be safe.

ericoporto:
>  it's crashing whenever run from IDE.

tested and confirmed, weird error though.

> An exception 0xC0000005 occurred in ACWIN.EXE at EIP = 0x00541D4A; program pointer is -199, engine version 4.0.0.12, gtags (0,0)

ivan-mogilko:
The debugger connection is fixed.

As an additional rework - implemented ScriptThread class which contains a saved state of a "script thread": a data stack, callstack and execution position. The ScriptExecutor now has to be told to execute the script on a particular "thread".
If there's any active script thread, and a new script is run on a different one, ScriptExecutor will push previous thread to the threads stack, and pop afterwards, restoring its state.

There are 2 standard threads created by the engine: "Main" and "Non-blocking" (for running "rep-exec-always"), but more could be added if needed.

My belief is that this system potentially allows to implement "coroutines" in AGS at some point in the future, because it allows to pause a script thread, save it, and then resume later at some point. I am referring to [this coroutine idea I once posted on AGS forums](https://www.adventuregamestudio.co.uk/forums/index.php?msg=636665195).

ivan-mogilko:
Not planned as a part of this PR, and strictly as a "proof of concept", I made this dumb coroutine test:
https://github.com/ivan-mogilko/ags-refactoring/tree/experiment--dumb-coroutine

Not done fully right, but works in practice. Normally you'd make a script keyword for async or iterating operations (like `async` or `yield` in C#); here I did not want to bother with compiler so added WaitAsync(loops) function that converts current script into "coroutine".

In principle it works like this:
- WaitAsync is called, it sets "async wait counter" and asks executor to suspend;
- script executor suspends current script thread and returns;
- engine checks execution result, finds that thread was suspended, copies its full state to the special "coroutine thread";
- game continues to run as normal, with all the player input, GUI, etc;
- engine updates "async wait counter" on each game tick;
- as soon as that reaches zero, it tells script executor to run the previously saved "coroutine thread";
- more suspends are possible, coroutine thread may be suspended again and again after each start, until it finishes.

The current implementation cannot have multiple simultaneous coroutines, any new one will overwrite previous completely, but that's just because I was lazy. Normally it should have a list of wait counters mapped to coroutine threads, or something similar.

Example of script:
```
function RunCoroutineTest()
{
    Display("RunCoroutineTest: 1");
    WaitAsync(60);
    Display("RunCoroutineTest: 2");
    WaitAsync(60);
    Display("RunCoroutineTest: 3");
    WaitAsync(60);
    Display("RunCoroutineTest: 4");
}

function on_key_press(eKeyCode key)
{
    if (key == eKeyW) RunCoroutineTest();
}
```

ivan-mogilko:
I think I'm done with the main part, now need to test and compare performance with the ags4 branch, check if debugging from the Editor and watch variables functionality still works.

ivan-mogilko:
Well, performance wise this PR is promising, I've tested with that 3d space sim "wcproto" game, and compared to ags4 branch this PR shows same level of FPS. Maybe even 1-2 fps more on average (although i can't explain why would it, and maybe this is just random fluctuation).

ericoporto:
I ran this with an updated version of the Sandwalker and it worked fine afaict. I can't comment on performance, since my current laptop has too much cpu fluctuation, but I didn't notice any slow downs.

ivan-mogilko:
> (I guess any thorough testing requires backporting to ags3 and let the people who play old games with find the bugs, but this does feel like a lot of work)

Backporting ScriptExecutor to ags3 is not impossible, there's nothing ags4-specific in it (except for new script opcodes). I just don't feel like doing this now, while there's 3.6.2 is prepared for release. Although me or someone else might do this; I believe it's more clear as a system than ccInstance.

ivan-mogilko:
I think i'll delay merging this until after next 4.0 Alpha update gets released, because there are couple of important bug fixes, and i don't want to mix this serious change in.

ericoporto:
That is fine, from the user point of view I think that the color change (the other pr) is much more disruptive (it is good change, but it is change), and having a release do they can focus on catching up and adjusting their project as needed is a good thing.

ivan-mogilko:
@AlanDrake hey, if you've got spare time, could you try this pr branch and playthrough your game, and see if there will be any errors or anything suspicious in behavior? There's not much of games made in AGS 4, so test material is lacking.

ivan-mogilko:
BTW, It's interesting, but git appears to be able to merge changes made to cc_instance.cpp in ags3 branch with the scriptexecutor.cpp in ags4 branch. Probably it makes a guess because the history tells that cc_instance was renamed to scriptexecutor.cpp in the past.

PR #2626, opened 2024-12-20
Script API: add Room.Name and Room.Number

### LINKED ISSUE #1135: Script API: add Room.Name (description) and Number
Add Room.Name (called "description" in the editor) and Room.Number to script api. There are multiple cases of users trying to get "Description" using GetTextProperty, which obviously does not work, so there's no way to get this field. Room's number may in theory be taken from player.Room, but might be good to have explicit Room's property.

Both properties must be readonly.

Also, perhaps display "Name" as property name instead of "Description" in the editor.

ivan-mogilko:
Resolves #1135.

Adds 3 static readonly properties to Game struct:
```
/// Gets the total number of the rooms in game.
static readonly attribute int Game.RoomCount;
/// Gets the existing room numbers by a sequential index (from 0 to RoomCount - 1); returns -1 if index is not valid.
static readonly attribute int RoomNumbers[];
/// Gets the room's name (description) by the room's number; returns null if such room does not exist.
static readonly attribute String Game.RoomNames[];
```

Adds 2 static readonly properties to Room struct:
```
/// Gets the current Room's number
static readonly attribute int Room.Number;
/// Gets the current Room's name (description)
static readonly attribute String Room.Name;
```

The only caveat is that room descriptions were not saved in the room itself, so I had to add them to room format.

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -154,8 +154,9 @@ enum GameDataVersion
     kGameVersion_400_09         = 4000009,
     kGameVersion_400_10         = 4000010,
     kGameVersion_400_11         = 4000011,
+    kGameVersion_400_13         = 4000013,
     kGameVersion_LowSupported   = kGameVersion_360_21,
-    kGameVersion_Current        = kGameVersion_400_11
+    kGameVersion_Current        = kGameVersion_400_13
 };
 
 // Data format version of the loaded game
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -206,7 +206,8 @@ HGameFileError GameSetupStruct::read_audio(Common::Stream *in, GameDataVersion d
 
 void GameSetupStruct::read_room_names(Stream *in, GameDataVersion data_ver)
 {
-    if (options[OPT_DEBUGMODE] != 0)
+    if ((filever >= kGameVersion_400_13) ||
+        (options[OPT_DEBUGMODE] != 0))
     {
         uint32_t room_count = in->ReadInt32();
         for (int i = 0; i < room_count; ++i)
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -210,10 +210,12 @@ void GameSetupStruct::read_room_names(Stream *in, GameDataVersion data_ver)
         (options[OPT_DEBUGMODE] != 0))
     {
         uint32_t room_count = in->ReadInt32();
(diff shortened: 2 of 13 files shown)


ericoporto:
> The only caveat is that room descriptions were not saved in the room itself

I may be misremembering, but I thought the debug command that can teleport (used with Ctrl+x in templates) had the room names in the list.

ivan-mogilko:
> I may be misremembering, but I thought the debug command that can teleport (used with Ctrl+x in templates) had the room names in the list.

That's true, but these are saved in the game data instead of the room...

ericoporto:
Other thing that I remembered is I believe all other game elements have names that are also writeable. I don’t know how useful this is in regards to room names, I imagine most of cases where a location changes enough to change name it also will be an actual different room - unless it’s something like a room for a character you don’t know that later you get to know - say “Apartment” becomes “Jane’s Apartment” or something.

I guess there is also the Name and the scriptName, in theory - where the scriptName is read-only.

ivan-mogilko:
Added Game.RoomCount and Game.RoomNames[] properties.

There's an obvious problem with these two. Rooms are not sequential and RoomNames[] may have gaps with null elements, similar to sprites for example.
Because of that RoomCount has a dubious meaning. It's not the  "length" of RoomNames, but just a report of how many valid elements are there.
I'm not certain if that's acceptable.

ericoporto:
Uhm, we now can retrieve dynamic arrays from functions. Don’t have any idea on this yet but we could avoid the RoomCount name problem by letting the user figure it out from the Length property.

> One option that I might think of is to have a separate RoomNumbers[] property, that returns a sequence of valid room numbers

I don’t have a good idea to name a Room Count that is specifically to numbers, AvailableRooms may be too verbose but it’s more clear in which room count we are talking and perhaps if the type is an int people understand it’s a number.

ivan-mogilko:
> Uhm, we now can retrieve dynamic arrays from functions. Don’t have any idea on this yet but we could avoid the RoomCount name problem by letting the user figure it out from the Length property.

I would not want to return dynamic array here, because it means create new array of elements and copy all data over, while user may want to retrieve only 1 string.

ericoporto:
> ```
> for (int i = 0; i < Game.RoomCount; ++i)
> {
>      String name = Game.RoomNames[Game.RoomNumbers[i]];
> }
> ```

I didn't had any idea better than this other than retrieving these as an AGS dictionary - keys as numbers and values as names.

ivan-mogilko:
Ok, I'll try to add it this way.

PR #2629, opened 2024-12-23
New compiler: When in the body of a static struct function and finding an otherwise unknown symbol, imply static access

### LINKED ISSUE #2067: New compiler: let omit type name when addressing static members from member methods
New compiler already allows to omit the `this` keyword when addressing object's methods  and fields from a object's method, but I found that when calling static methods from object's method you still have to type `Typename.Function()`.

If possible, I propose to also let omit the typename in this scenario.

*Unless this goes against the desired syntax design*, in which case you may cancel this suggestion.

fernewelten:
Fixes #2067

Currently, whenever the new compiler is compiling the body of a non-static struct function and it encounters an unknown identifier `i` in an expression, it doesn't give up yet. It looks up `this.i` first, and if that subexpression is valid, the compiler uses it.

This PR generalises this concept to static struct functions: When the compiler is compiling the body of a static struct function for the struct `str` and it encounters an unknown identifier `i` in an expression, it doesn't give up yet. It looks up `str.i`, and if that subexpression is valid, then the compiler uses it.

```
float FourCos(static Maths, float rd)
{
    return 4.0 * Cos(rd); // ← 'Cos' unknown, but 'Maths.Cos' exists so use that
}
```

This PR also fixes the following bugs:
- When a struct `s` extends another struct `t` and `t` has an attribute or a function `af` then `this.af` (or `af` when otherwise unknown)  should reference `t::af`. This didn't always work but should work now.
- When a struct `s` has a non-static attribute `a` then calling `s::a` from a `static` function should fail because the specific object isn't known to which the attribute refers – only the class is known. This used to be flagged for functions but not for attributes. It should now also be flagged for attributes.

Added googletests for the new use cases and the bugs.

This PR also cleans up comments in the code (in a separate commit). No functional changes in that commit. 
- Especially the comments at the start of the parser, which have partially become out of date by now.
- Sometimes comments preceded a function in the '.cpp' file where the same or a more accurate comment also was in the declaration of the '.h' header. In these cases I deleted the comment in the '.cpp' file to prevent contradictions or mismatches.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -1517,7 +1517,7 @@ void AGS::Parser::ParseFuncdecl_HandleFunctionOrImportIndex(TypeQualifierSet tqs
         return;
     }
 
-    if (imports_idx < 0 || imports_idx >= _scrip.imports.size())
+    if (imports_idx < 0 || static_cast<size_t>(imports_idx) >= _scrip.imports.size())
         InternalError(
             "Imported function '%s' has the imports index '%d' which is invalid",
             _sym.GetName(name_of_func).c_str(),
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
This appears working as documented.

PR #2632, opened 2024-12-25
Script API: remake RestoredSaveInfo.Result into indexed property

ivan-mogilko:
This is to get opinion on the possible change.

When documenting the new ["validate_restored_save" callback](https://github.com/adventuregamestudio/ags-manual/wiki/ValidateRestoredSave) I began to wonder if preseting RestoredSaveInfo.Result property as a set of flags would be convenient to users. Majority of AGS users don't use flag sets very often, and finding out whether a certain flag is set may be confusing to some of them (if not to many of them).

One idea that came to me is to turn this property into a array of bools instead, where each pseudo-element is a flag.

In simple words, instead of doing this:
```
    if ((info.Result & eRestoredSave_MissingData) != 0)
    if ((info.Result & eRestoredSave_Prescan) == 0)
```
users would do this:
```
    if (info.Result[eRestoredSave_MissingData])
    if (!info.Result[eRestoredSave_Prescan])
```

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -3399,7 +3399,7 @@ managed struct RestoredSaveInfo
 {
   import attribute bool Cancel;
   import attribute SaveComponentSelection RetryWithoutComponents;
-  import readonly attribute RestoredSaveResult Result;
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Ok, I don't know, maybe I was confusing myself, but now I wonder why did not I just make a set of boolean properties instead of flags in script API.

PR #2635, opened 2024-12-26
Organize Plugin sources

ivan-mogilko:
The purpose is to tidy plugin sources present in our repository.

1. Moved plugin event constants back into agsplugin.h, removed agsplugin_evts.h. This was an attempt to deal with naming conflicts within the engine, but I have second thoughts now. Instead, created a new enum `PluginEventID` to use inside the engine code. The idea is that agsplugin.h is NOT included by the engine *except* by agsplugin.cpp.

2. Reorganized all plugin sources into the uniform dir structure:
```
    + ags_plugin
        + ags_plugin
             plugin sources
        + msvc
             msvc project files
        makefile
        readme
        any extra files
```
The only exception is "agslua" plugin that is in a submodule.

3. Created MSVC projects for those that did not have them, updated existing to VS2015 (matching the engine, can be updated to newer ones along with the engine later). Configured output to be in ".build" inside "msvc" dir for consistency. Added x64 build configurations.

4. Made sure that plugins don't include engine files, from Engine or Common folders. Instead copied agsplugin.h and platform.h into every plugin, thus making them self-sustained projects. The purpose of this is to ensure that they don't have any extra dependencies, and this will make it easier for people to copy these plugin projects elsewhere, or use as examples for their own.

5. Fixed "agsblend" and "agspalrender" compilation for Windows; these were just minor problems with wrong order of headers, and incorrect winapi call.

6. NOTE: "agstouch" is a bit a puzzling plugin. It does not build as-is standalone, because it requires SDL (probably SDL 1), and does not have any include paths or lib paths configured. It will have to be fixed separately if required.
Also, it uses old-style `#if defined (IOS_VERSION)` instead of new-style `#if (AGS_PLATFORM_OS_IOS)`, so maybe it does not work at all.
From the looks of the code, it is not a "touch input" as one might guess, but used to bring up a virtual keyboard on screen.


ericoporto:
Uhm, I don't see a need for copying the agsplugin.h and the platform.h throughout, they can be downloaded directly from our repository in any plugin (example in https://github.com/ericoporto/agsworks)

ivan-mogilko:
If that's a problem, then I can put them in a shared folder inside Plugins, but I don't want to link from the Engine.
In my opinion, plugin sources have to contain all files necessary for building it.

ivan-mogilko:
>  they can be downloaded directly from our repository in any plugin (example in https://github.com/ericoporto/agsworks)

I read the readme in your repository, where it sais this:

> NOTE: both the VS Project and the Makefile Project assume you have cloned both this repository and ags repositories in the same directory:
> For now, ags repository is only used to include two header files ../ags/Common/core/platform.h and ../ags/Engine/plugin/agsplugin.h

I honestly think that's very wrong. You are requesting to clone whole "ags" repository just for 2 small files... Why not just include them into the plugin source?

ericoporto:
Here is other plugin where I just get AGS instead https://github.com/ericoporto/agsappopenurl

I don't think duplicating these files makes sense at all. These are old plugins that don't even make sense to distribute with the engine, they exist to run old games.

The first commit is fine but the other things I am a bit lost of what the purpose.

ivan-mogilko:
> Here is other plugin where I just get AGS instead https://github.com/ericoporto/agsappopenurl

Why? When I made a spritevideo plugin, I just included the necessary agsplugin.h. That's how it was supposed to be done.

ericoporto:
I am sorry, I was too aggressive. I think I get it now, I will try to pick this and rework the CI and installer to work with these new projects.

ericoporto:
I am not sure I really want to bring these plugins back to life, this is essentially maintaining 6 new projects in all six AGS platforms. It would also mean 6 new standalone CMake projects and Makefiles.

ivan-mogilko:
I did not have intent to have this built with CI; I only needed to have normal up-to-date msvc projects (and Makefiles) in case there's a need to build these as standalone. This is required very rarely, but when it happens, I have to open existing solutions in the older versions of MSVS, or upgrade, or create a new one (if it's missing), and that's annoying.

My view on this is that, since we still need this plugin code in the backwards compatible branch, then it's better to have it organized. I don't want to leave this exactly as it is, because there are obsolete solutions there (from msvs2008), and that's silly. Updated project files won't cause any problems for this repository. We may state that these plugins are "archived" in their current state, as in, not developed in any way further, and purposed only to support existing old games.

ericoporto:
OK, that's fair. I was under the impression that it was meant to have them distributed with the Editor. I think the new projects can be already VS2019 (I haven't checked if they are yet).

PR #2640, opened 2024-12-30
adjust CI and docker for VS2019 compatibility using VS2022

### LINKED ISSUE #1508: Upgrade Visual Studio to a supported version in Windows Docker
**Windows Dockerfile:** [**`ci/windows/Dockerfile`**](https://github.com/adventuregamestudio/ags/blob/master/ci/windows/Dockerfile)

First, VS2019 (or VS2022) C++ builds are binarily compatible with VS2015 builds. So there's no change here. This is to make it easier to modify and upgrade the Windows build environment in use.

VS2015 is out of support and we rely in an old Docker image I had in dockerhub for it. Ideally, we want to upgrade to the current stable Visual Studio, which is VS2019 (erh, or maybe VS2022). But this can be a problem in Cirrus CI.

- VS2019/VS2022 occupies a lot more space
- The Docker Image gets bigger
- Cirrus CI takes so long to download the docker it timeouts the Windows build

The solution is finding out what are the parts (workloads and individual components) of VS we need and install only those. 

**References:**

- What are workloads: https://docs.microsoft.com/en-us/visualstudio/install/modify-visual-studio
- Workloads list: https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools
- Installing build tools in a container: https://docs.microsoft.com/en-us/visualstudio/install/build-tools-container
- Container issues with Build Tools https://docs.microsoft.com/en-us/visualstudio/install/build-tools-container-issues
- Cirrus Windows Docker images: https://github.com/cirruslabs/docker-images-windows
- MS Windows Container repository: https://github.com/microsoft/Windows-Containers
  - Note issues are closed in 30 days even if they aren't solved so there's a ton of problems accumulating there... (example https://github.com/microsoft/Windows-Containers/issues/217)

---

this is a requirement for #1462 

I talked a bit on the problem of unsupported VS2015 in PR #1291

---

Opened a PR to fix this in https://github.com/adventuregamestudio/ags/pull/2640

ericoporto:
fix #1508 

built on top of #2591

---

**Some notes:**

Visual Studio 2022 build tools image did not work with CMake. It seems the big issue is that Visual Studio 2022 Build Tools doesn't correctly installs all VC tools, the 142 (VS 2019, aka VC 14.29) vc tools seems to install most correctly but it doesn't seem to proper have the CLI/C++ tools working, and the VC 143 (VS 2022, aka VC 14.43) doesn't properly create the cl.exe in place.

I used VS 2022 community instead, which worked fine on the Windows container and properly install packages - though this means the image is slightly larger (21.96 GB in the base image). Here is the docker file for the base image

https://raw.githubusercontent.com/ericoporto/min-ags-dev-env/211537671e66f8e6df0c6fcd0826a529b535acd5/Dockerfile

I opted to build this package separately because installation of VS2022 takes a long time - a little more than an hour, enough to fail Cirrus CI. I built this image in my computer too, since the GHA that currently exists for docker fails with Windows Containers. 

I also kept the Windows Server 2019, in theory it could work with a newer windows (and move from the LTSC2019 base), but the cirrus ci official windows container images are still those, so I opted to keep with what is currently tested.

The resulting image has

- platform toolsets
  - v142 (14.29.30133)
  - v143 (14.42.34433)
- .NET Framework functionality to build 4.6.2 but also 4.8.1
- Should be possible to evolve to support newer .NET as needed (nom framework)

**TO-DO:**

- [x] test the windows binaries of editor in different Windows
  - [x] Windows 7
  - [x] Windows 10
  - [x] Windows 11
- [x] verify in a clear machine (VM?) that the selected VC redist is indeed correct and working with the installer
-  ~~maybe adding a `"windows-container"` repository right in the organization so that I can move the base image to there (using the GitHub Packages feature), this would also be an opportunity to document a how-to on windows containers to explain things~~ not now, maybe when GHA can push windows containers.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,29 +8,29 @@ build_windows_task:
     dockerfile: ci/windows/Dockerfile
     os_version: 2019
   env:
-    AGS_LIBOGG_LIB: C:\Lib\Xiph\x86
-    AGS_LIBTHEORA_LIB: C:\Lib\Xiph\x86
-    AGS_LIBVORBIS_LIB: C:\Lib\Xiph\x86
+    AGS_LIBOGG_LIB_VS19: C:\Lib\Xiph\x86
+    AGS_LIBTHEORA_LIB_VS19: C:\Lib\Xiph\x86
+    AGS_LIBVORBIS_LIB_VS19: C:\Lib\Xiph\x86
     AGS_SDL_INCLUDE: C:\Lib\SDL2\include
     AGS_SDL_SOUND_INCLUDE: C:\Lib\SDL_sound\src
     AGS_SDL_LIB: C:\Lib\SDL2\lib\x86
-    AGS_SDL_SOUND_LIB: C:\Lib\SDL_sound\lib\x86
+    AGS_SDL_SOUND_LIB_VS19: C:\Lib\SDL_sound\lib\x86
   build_debug_script: >
-    "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
+    "C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/Tools/VsDevCmd.bat" &&
     cd Solutions &&
-    msbuild Engine.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v140 /p:Configuration=Debug /p:Platform=Win32 /maxcpucount /nologo
+    msbuild Engine.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v142 /p:Configuration=Debug /p:Platform=Win32 /maxcpucount /nologo
   build_tools_debug_script: >
-    "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
+    "C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/Tools/VsDevCmd.bat" &&
     cd Solutions &&
-    msbuild Tools.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v140 /p:Configuration=Debug /p:Platform=x86 /maxcpucount /nologo
+    msbuild Tools.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v142 /p:Configuration=Debug /p:Platform=x86 /maxcpucount /nologo
   build_release_script: >
-    "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86 &&
+    "C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/Tools/VsDevCmd.bat" &&
     cd Solutions &&
-    msbuild Engine.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v140 /p:Configuration=Release /p:Platform=Win32 /maxcpucount /nologo
+    msbuild Engine.sln /m /property:MultiProcessorCompilation=true /p:PlatformToolset=v142 /p:Configuration=Release /p:Platform=Win32 /maxcpucount /nologo
   build_tools_release_script: >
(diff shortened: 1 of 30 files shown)


ericoporto:
image

There is something wrong, I tried to run in a clean environment using the new [Windows Sandbox feature](https://learn.microsoft.com/en-us/windows/security/application-security/application-isolation/windows-sandbox/windows-sandbox-overview), and got the above error. I also noticed that the AGS installer needs to be adjusted in the entry because the VC Redist is named as VC 2015, but it's the VC 2015+ now (also known as  Microsoft Visual C++ 2015-2019 Redistributable (x86) - 14.29.30133)

ivan-mogilko:
I'm testing this now.

Windows/README.md needs to also be modified, changing prerequisites to "Microsoft Visual Studio 2019 or higher".
Following section needs an update:
```
AGS_SDL_INCLUDE - pointing to the location of SDL2 headers;
AGS_SDL_LIB - pointing to the location of SDL2 library files;
AGS_SDL_SOUND_INCLUDE - pointing to the location of SDL Sound headers;
AGS_SDL_SOUND_LIB - pointing to the location of SDL Sound library files;
AGS_LIBOGG_LIB - pointing to the location of libogg library files;
AGS_LIBTHEORA_LIB - pointing to the location of libtheora library files;
AGS_LIBVORBIS_LIB - pointing to the location of libvorbis library files;
```

where the macro names has to be adjusted to _VS19 variants (except SDL which is linked dynamically).
It may have sense to add a comment in brackets that this is a way to distinguish libs built for VS15 in older branches.

Finally, a mention of VC redist, currently is:
```
Redist\vc_redist.x86.exe - a [Visual C++ Redistributable for Visual Studio 2015](https://download.microsoft.com/download/6/A/A/6AA4EDFF-645B-48C5-81CC-ED5963AEAD48/vc_redist.x86.exe) (or C++ Redist corresponding to the MSVS you were building AGS with).
```
Should be updated to the one required by VS2019.

ericoporto:
About the `_VS19` library macronames, if we add the regular ones to the prop files, after the `_VS19`, would it work too? The idea would be if VS would find the other if the _VS19 variant is not specified. This may be easier for someone that doesn't have the old builds of the libraries locally.

ivan-mogilko:
> About the `_VS19` library macronames, if we add the regular ones to the prop files, after the `_VS19`, would it work too? The idea would be if VS would find the other if the _VS19 variant is not specified. This may be easier for someone that doesn't have the old builds of the libraries locally.

Hmm, since these prop files are read only by the matching solution, I suppose that won't cause troubles.

ivan-mogilko:
XP builds do not work in this PR, VS telling that certain linking option is not recognized.
But they do not build in master branch either, because of another problem. Somehow I haven't noticed this earlier.

ericoporto:
Did the previous builds from CI work in Windows 7? I don't have access to a computer with it - and couldn't even figure how to run it in a VM.

ivan-mogilko:
Tested on Windows 7 and Windows 10. Both required to install new VC redist, and worked after that.

ericoporto:
If we want we can later also try moving to .NET Framework 4.8 (not 4.8.1 though) , it has the same compatibility as 4.6 but 4.6 end of life is 2026 and 4.8 is classified currently as indefinitely - has no set end of life.

ivan-mogilko:
I suppose this may be merged. Is there anything else that should be tested or done here?

ericoporto:
Uhm, not really, this should be working.

My only additional suggestion is, if we want, I could maintain a `windows-container` repository to have the base dockerfile (that has the VS install) and the docker image (using GH packages feature) in the adventuregamestudio organization instead of my personal repository. Unfortunately I still need to build the docker images in my computer and push it manually, because neither CI can actually do the VS install and push of the image - in cirrus it takes so much installing VS it times out the build and in GHA there's a different specific issue on docker itself where Windows containers throw an error when trying to push. Because of this, there isn't much to gain - I would still have to build the base image and push it from my own computer.

persn:
> If we want we can later also try moving to .NET Framework 4.8

If there's no compat issues then let's try

PR #2641, opened 2025-01-05
Editor: Character use less function in Script Header

ericoporto:
I have reduced the use of `function` in the Character API in Script Header.

- `SetProperty` and `SetTextProperty` are the only function calls that return something (`bool`s) that aren't a "Get" function.
- The variadic calls (`Say`, `SayAt`, `SayBackground` and `Think`), I actually can't tell if they are meant to always return 0 or be void, so I haven't changed them
- `GetProperty` returns `int`, so I changed from function to int.
- `IsCollidingWithChar` and `IsCollidingWithObject` are meant to be `bool,` so I changed their types to `bool`, but they are using `int` return macros instead of `bool` return macro in `ac/character.cpp`

This is related to #2463

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2690,11 +2690,11 @@ enum StopMovementStyle
 
 builtin managed struct Character {
   /// Adds the specified item to the character's inventory.
-  import function AddInventory(InventoryItem *item, int addAtIndex=SCR_NO_VALUE);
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
Say, SayAt and Think do not have a return value.
SayBackground returns Overlay.

ericoporto:
I remade `Say`, `SayAt` and `Think` as void an [push forced](https://github.com/adventuregamestudio/ags/compare/0ae89182d1e48a23ea96db8793b02d6129239764..ab9e2afefbe6bfa1b0f1a998ad2582d402e8995f)

PR #2642, opened 2025-01-05
Editor: avoid function on Object in script header

ericoporto:
Same as #2641 but for objects. Related  to #2463.

Note that after this there is no more use of `function` in the script header. The remaining use of function is in the templates and handlers created by the Editor.

And it's also used in this file below (I don't know if the editor use this somewhere?)

https://github.com/adventuregamestudio/ags/blob/fd272e21f671c4b17caa8def5460f73b041e727e/Editor/AGS.Editor/Resources/__DialogScripts.asc#L10


ivan-mogilko:
"function _run_dialog_request" is a local helper function used when generating dialog scripts; not exposed to api.

Issue #2525, opened 2024-09-01
AGS 4: support Alpha in Color parameter


ivan-mogilko:
This is written after #1514 and #1980.

Now when AGS supports true 32-bit colors for drawing in game, we may consider supporting Alpha value as well.
In terms of storing Alpha component in a color number, this is already supported by the field format, but there are several issues to resolve before this may be actively used.

The subtasks follow:

- [x] 1. Expand Game.GetColorFromRGB to also have "alpha" parameter. For backwards compatibility, and also because it's less common, add it as a last optional parameter with default value of 255 (full alpha).
EDIT: alternatively, make another function for getting ARGB/RGBA.
- [x] 2. Make sure that all the drawing operations in Allegro support alpha blending (not just blit, bit also draw line etc). I've never investigated that, but this may not be trivial, depending on how they work. **IF** Allegro does not support that, then we might have to implement our own, by cloning standard allegro functions, and more or less passing any pixel assignment in them through a blender (pixel function). These functions will be called if alpha if anything but 0 or 255.
In case alpha is 255 the standard functions are called;
In case alpha is 0, nothing is drawn and whole operation may be skipped.
- [x] 3. If alpha is supported, then all the object Color properties must respect alpha component as well. Such as: GUI colors, speech color, and so forth.
- [x] 4. BTW, there will be a side effect of this, that color 0 which is currently used in some cases meaning "transparent" (even in 32-bit games) will represent a 0,0,0,0 ARGB (so color with alpha == 0), and therefore it will be possible to assign true-black color to GUI background and such using any non-zero alpha. But this has to be double checked.
- [ ] 5. Editor: there's a issue of the color picker. The common color picker dialog which we currently use does not let input alpha. Maybe there's a mode in which it does, then it will simply have to be enabled. But if not, then we'll require a different dialog.
In short term we may leave things as is, but then you won't be able to set alpha in the editor other than by editing a Color value by hand (and choosing a different color in the picker dialog will reset alpha to 255).
EDIT: split into ticket #2656
[...]

PR #2648, opened 2025-01-10
AGS 4: color properties have alpha component

ivan-mogilko:
The first step for #2525. The priority is to have alpha component stored in color properties, both in Editor and runtime, and fix a problem with some GUI parameters that were not able to get assigned a black color after #1980.

This does not yet guarantee that alpha will actually work (probably it won't). The idea is to make opaque colors work while having alpha component, and be able to assign *transparent color* (0).

Editor treats color numbers as ARGB, but forces opaque properties for now.
1. Color properties formally support alpha component.
2. When upgrading older projects, all 32-bit colors are ensured to be opaque (as alpha was not supported before).
3. Property Grid technically allows ARGB property values, but forces them to opaque for now, until we actually make use of alpha value in the engine. Please note that Color picker cannot select "transparent" color yet, you have to type it manually.
4. The one important exception is transparent color: it can be set and is displayed as "0" for simplicity.
5. ColorUIEditor draws "transparent" color as a classic checkered rectangle.
6. Color properties are now serialized as a hex number, for easier reading.

Engine:
1. Color properties and parameters formally support alpha component, but I suspect that either won't work or produce strange effects if you try to use it now.
2. Game.GetColorFromRGB() ensures that returned color is opaque.
~3. Had to refactor the dreadful `asspch` parameter in display_main, which combined text color with extra meaning, which is impossible to do when color value occupies all 32-bits. Introduced DisplayTextFlags for passing extra options. This part should be backported to 3.6 branch, I think.~ this is done by #2649 and ported among active branches.

In Script:
1. COLOR_TRANSPARENT constant is redefined as 0, because -1 translates to "opaque white" ARGB.

I raised format version numbers, but only ensured that project upgrade works at the moment. If it's necessary, a backwards compatible fixes could be done in the engine too, the raised versions leave such backup opportunity in case someone finds that necessary.


ivan-mogilko:
Rebased after #2649 and porting changes to ags4.

Also, made ColorUIEditor display transparent color selection with checkered rectangle.

ivan-mogilko:
The GH actions will fail, because of the test game failing (would need to correct color constants again).

ericoporto:
Hey, how do I get transparent GUI background color now?

image

The upgrade didn't got any of my GUIs - in case it was supposed to.

ivan-mogilko:
> Hey, how do I get transparent GUI background color now?

The only way is to set 0 directly into the property. Standard color picker does not have such option.

> The upgrade didn't got any of my GUIs - in case it was supposed to.

I'm conflicted about how to upgrade from the previous 4.0, because the previous 32-bit implementation did not let distinguish 0 as transparent and 0x000000 as black. If I make it transparent, then all the controls with black text color will have transparent text color, for example. That's why I thought that it will be less annoying to port it as black, and let users changes to transparent background where they need to.

The upgrade from earlier versions (from before 32-bit colors) should not have this problem... i think.

ericoporto:
that would make sense, since they are transparent in background colors. I don't know how hard that would be, but I use the background transparent a lot, both in GUIs but also in buttons - invisible buttons is how I make regions in a GUI clickable.

ivan-mogilko:
Adjusted the upgrade process. Now it treats color 0 in background and foreground colors separately: back color 0 becomes transparent, fore color 0 becomes opaque black.
GUI.BorderColor is treated as "background" for normal GUI and "foreground" for TextWindow gui.

When upgrading games made **before** 4.0 Alpha 13 all the colors seem to upgrade properly.
In games made after Alpha 13 the only problem is that GUI with fully black background will turn transparent. But there's no real way to work around that.

ericoporto:
I tested this, and the upgrade process works fine!

Now, the only thing I found that is weird, is in my game using `surface.Clear()` causes the surface to be orange. I fixed this by remaking all of these with `surface.Clear(0)`, and now they properly clear the surfaces.

ivan-mogilko:
> Now, the only thing I found that is weird, is in my game using `surface.Clear()` causes the surface to be orange.

Ugh. This is because it has a uncommon default value:
`import void Clear(int colour=-SCR_NO_VALUE);`

where SCR_NO_VALUE is an arbitrary number 
`#define SCR_NO_VALUE          31998`

ivan-mogilko:
Added [SerializeAsHex] attribute to Color properties, for easier reading of Game.agf.

ericoporto:
Hey , I tested this, and it works great! Question on the hex, I see it's written like
```XML
...
                <TextColor>FF000000</TextColor>
...
              <BackgroundColor>00000000</BackgroundColor>
```
There is no `0x` or `#` or like it before it.

> Reading back is done in two steps (first try hex, if failed then try decimal), for backwards compatibility.

I don't understand how it would fail.

ericoporto:
@ivan-mogilko can you test if the SpeechColor from a character in an ags4 game of your upgrades correctly?

ivan-mogilko:
Right, I did not realize that.

Apparently Convert.ToInt32 also handles `0x` prefix, so I use that instead of TryParse.

ericoporto:
ok, I tested this again, and now it is correctly loading both upgrading previous ags4 projects and loading the ones that are already upgraded by this, properly picking up the hex value of colors and pretty much all that I could think of!

ivan-mogilko:
Alright, I will merge this then.

ericoporto:
@AlanDrake just in case the bit shifts you are playing with are color related, I put a code hidden by details tag in the comment https://github.com/adventuregamestudio/ags/pull/2648#issuecomment-2585535676 that may be useful to you.

PR #2649, opened 2025-01-10
Engine: refactor passing text_color into display_main()

ivan-mogilko:
* Replaced DISPLAYTEXT_x macros with DisplayTextType enum;
* Got rid of confusing `asspch` parameter which combines text color with the "text style" by applying a sign. Instead have separate `text_color` and a `style` parameter of DisplayTextStyle enum.
* Also added DisplayTextShrink enum for the `allow_shrink` parameter.
* Grouped them into `DisplayTextLooks` struct.

### DIFF
--- a/Common/core/types.h
+++ b/Common/core/types.h
@@ -47,8 +47,8 @@
 // Stream offset type
 typedef int64_t soff_t;
 
-#define fixed_t int32_t // fixed point type
-#define color_t int32_t
+typedef int32_t fixed_t; // fixed point type
+typedef int32_t color_t; // AGS color number type (meaning depends on game's setting)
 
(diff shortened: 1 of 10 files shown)


ericoporto:
Great work! The display stuff is one of the parts of the engine where there is still a lot of cruft from the original code so this refactor looks great. I think in similar fashion, only aschar rename to char_id and allowShrink being made an enum (with a rename?) would be things in similar changes I would make to this code.

ivan-mogilko:
Changed to another variant, with enums and a struct to group them instead of flags.

ericoporto:
I tested this with a few of my own test games and also the shivah, and afaict all is still working alright.

PR #2650, opened 2025-01-11
SDL upgrade to version 2.30.11

ericoporto:
This updates to newer release of SDL2, the 2.30.11 release, from 01/01/2025.

I tested it with my games and it seems that this doesn't break anything.


ivan-mogilko:
Looks like this needs to be updated after #2653 is merged.

ericoporto:
Done!

PR #2652, opened 2025-01-11
Fix sprite rotation

### LINKED ISSUE #2651: AGS 4.0: a regression in DynamicSprite.Rotate
**Problem**

A user's game features "tiles" which are drawn on grid on a room's background, some rotated by multiples of 90 degrees (90, 180, 270).
When made in AGS 3.6.1 this results in a clean tile map, but when the same thing is done in AGS 4.0 there is a visible displacement on tiles rotated by 180 and 270 degrees, which results in visible gaps and shifts of tiles by few pixels.

**Step to reproduce**
1. Make an empty game with 1 room in AGS, and import a square "tile" sprite. Make sure the game resolution is enough to accommodate 4 sprites drawn on room bg in a row.
2. In the room's event e.g. "after fade in" script creation of a dynamic sprite from that tile, and paint 4 tiles in a row with no gaps between them, rotating each next tile by 90 degrees.
Clearing room bg to some bright color will help to notice if there are any misplacements.
Example of script:
```
function room_Load()
{
  DynamicSprite *tile = DynamicSprite.CreateFromExistingSprite(1);
  int wid = tile.Width;
  
  DrawingSurface *ds = Room.GetDrawingSurfaceForBackground();
  ds.Clear(Game.GetColorFromRGB(255, 0, 0));
  ds.DrawImage(100, 100, tile.Graphic);
  tile.Rotate(90);
  ds.DrawImage(100 + wid, 100, tile.Graphic);
  tile.Rotate(90);
  ds.DrawImage(100 + wid * 2, 100, tile.Graphic);
  tile.Rotate(90);
  ds.DrawImage(100 + wid * 3, 100, tile.Graphic);
  ds.Release();
}
```
3. In any v3.6.* the above will produce a seamless row of tiles.
4. In v4.0 alpha version some tiles will be visibly misplaced, and have gaps between them. This is likely not because of draw positions, but rather because they got distorted when rotating.
5. NOTE: I keep rotating same dynamic sprite in the above test, but even if I recreate the sprite from the original each time and rotate directly to the final angle, there's still a distortion at angle 180.

**Expected behavior**
Because the tiles are rotated by factor of 90 they should not get distorted even if rotated multiple times. The above test should produce visibly seamless row of tiles.

**Test games**
A game project made in v3.6:
https://www.dropbox.com/scl/fi/fbr8o10b805h3hwvbo6rf/test-360-dynamictilerotate.zip?rlkey=zl3lmkrvfl6zbwo54ua7r9ugq&st=rvkvia0e&dl=0
A game project made in v4.0:
https://www.dropbox.com/scl/fi/rdmwbkmrk93ud4pyvf78h/ags4-dynamictilerotate.zip?rlkey=2exi0wau9yaw8hcwtb4lc5of8&st=9dssf1cc&dl=0

ericoporto:
fix #2651

### DIFF
--- a/Common/util/geometry.cpp
+++ b/Common/util/geometry.cpp
@@ -35,14 +35,14 @@ bool IsRectInsideRect(const Rect &place, const Rect &item)
 float DistanceBetween(const Rect &r1, const Rect &r2)
 {
     // https://gamedev.stackexchange.com/a/154040
-    Rect rect_outer(
+    const Rect rect_outer(
         std::min(r1.Left, r2.Left),
         std::min(r1.Top, r2.Top),
         std::max(r1.Right, r2.Right),
         std::max(r1.Bottom, r2.Bottom)
     );
-    int inner_width = std::max(0, rect_outer.GetWidth() - r1.GetWidth() - r2.GetWidth());
-    int inner_height = std::max(0, rect_outer.GetHeight() - r1.GetHeight() - r2.GetHeight());
+    const int inner_width = std::max(0, rect_outer.GetWidth() - r1.GetWidth() - r2.GetWidth());
+    const int inner_height = std::max(0, rect_outer.GetHeight() - r1.GetHeight() - r2.GetHeight());
     return static_cast<float>(std::sqrt((inner_width * inner_width) + (inner_height * inner_height)));
 }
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
Please fix the inverse operation too (RadiansToDegrees), keeping them consistent.

ericoporto:
yeah, if I understand, the `X.X` type of numbers are double and `X.Xf` are floats. I did the adjustments.

I see in cirrus itself the CI is running but for some reason I don't see neither the ci or GHA rerunning here in the PR, not sure if the issue is in GitHub (or my region of GitHub) or my browser.

ivan-mogilko:
I confirm that this fixes the issue with rotated sprites.

ericoporto:
@ivan-mogilko , there is this header in `Common/util/math.h`, in ags4

https://github.com/adventuregamestudio/ags/blob/cf52f679a22ddde0ccd52e22d635d680ed692c45/Common/util/math.h#L26

This is because of the `ClampAngle360`, but we don't use this function in ags3 - and weirdly in ags4 this is only used for character rotation and camera, but not Overlay, or Objects (??)... It looks like this could be used everywhere?

ivan-mogilko:
> and weirdly in ags4 this is only used for character rotation and camera, but not Overlay, or Objects (??)... It looks like this could be used everywhere?

It is used everywhere: gui, character, object, overlay and camera.

PR #2653, opened 2025-01-12
SDL_sound: update to latest

### LINKED ISSUE #2244: 3.6.0: Looping sound issue
**Problem**

There have already been a report about looping sound before (#501), but it was closed after considered to be a "false alarm", or not engine's issue.
EDIT: also it seems to be reported long before 3.6.0 was out, so it was not related to the new audio system.

But recently another user reported a problem with the looping sound again. I tested it, and confirm that it's indeed happening in **3.6.0+**, while playing nicely in previous versions **(3.5.1 and lower)**.

Here's the sound file in question:
[VacuumNoise.zip](https://github.com/adventuregamestudio/ags/files/13482629/VacuumNoise.zip)

Here's a user's recording of playing this sound in 3.5.1.14:
[Looping_sound_AGS_3-5-1-14.zip](https://github.com/adventuregamestudio/ags/files/13482634/Looping_sound_AGS_3-5-1-14.zip)

Here's a user's recording of playing this sound in 3.6.0.53:
[Looping_sound_AGS_3-6-0-0-53.zip](https://github.com/adventuregamestudio/ags/files/13482640/Looping_sound_AGS_3-6-0-0-53.zip)

**To Reproduce**

Simply import into AGS, and play repeatedly on start or room enter, e.g.
```
aVacuumNoise.Play(eAudioPriorityNormal, eRepeat);
```

---

There's an obvious question of how to diagnose this problem. Theoretically I see few potential places where the mistake may take place:
1. Sound decoding.
2. Looping playback.
    2.1. Result of our "audio slot" work;
    2.2. Result of mojoAL playback.

My thought is to add a "sound debug" mode in the engine that would write decoded sound into a file, *separately* from the OpenAL playback. This will let see the soundwave of a decoded data itself.
As for playback, the good thing is that SDL supports "Direct-to-disk Audio" device, which will write the engine's audio output to "raw" file (format is 32-bit float PCM), which may then be imported to a sound editing program such as "Audacity". If that would not be enough, perhaps it may be possible to add our own output for data processed in "audio slot" before it's given to MojoAL.

NOTE: I assign this to 3.6.1 Milestone, but if a fix is found soon enough, then it may as well be backported to 3.6.0 batch and another patch released.

~~May or not be related to #1978.~~

---

**UPDATE:**
This is a SDL_Sound issue: https://github.com/icculus/SDL_sound/issues/92

ericoporto:
This is meant to fix #2244 , but the problem is I can't hear the difference between both games that are attached in the bug report

https://github.com/icculus/SDL_sound/commit/474dbf755a1b67ebe7a55467b4f65e033f268aff

possibly fix looping ogg files that have silence appended at the end - the raw file tests using SDL_sound were successful, I just can't hear the issue with the actual ags game that is attached in the issue, so I can't verify the fix actually worked for the actual game.

### DIFF
--- a/CMake/FetchSDL_Sound.cmake
+++ b/CMake/FetchSDL_Sound.cmake
@@ -1,8 +1,8 @@
 FetchContent_Declare(
     sdlsound_content
     DOWNLOAD_EXTRACT_TIMESTAMP TRUE
-    URL https://github.com/icculus/SDL_sound/archive/16e626561f3b496e4a2da98c95d23359eeee3c01.tar.gz
-    URL_HASH SHA1=37d56d695f263bb35708a05131271f41a1d9c01d
+    URL https://github.com/icculus/SDL_sound/archive/474dbf755a1b67ebe7a55467b4f65e033f268aff.tar.gz
+    URL_HASH SHA1=7f6b977a7ebae4cdecd4ac9b9404760a9e9a6b77
 )
 
 FetchContent_GetProperties(sdlsound_content)
--- a/OSX/README.md
+++ b/OSX/README.md
@@ -78,7 +78,7 @@ The Xcode project depends on the prebuilt SDL2 Framework and library sources.
 
 The project also requires sources of library dependencies, either using the `libsrc/download.sh` script or manually downloading them and extracting them as below.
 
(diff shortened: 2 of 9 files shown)


ivan-mogilko:
I can confirm that the new version fixes the problem.
If I use the LoopVacuum.ags attached [to this comment](https://github.com/adventuregamestudio/ags/issues/2244#issuecomment-1978605483), the previous SDL2_Sound was causing audible short jumps, the new version produces a seamless play.

PR #2655, opened 2025-01-14
Android: upgrade target SDK to 34

ericoporto:
- This is meant to comply with current Google Play Target SDK requirements.

- NDK is kept the same, otherwise we would have to upgrade minSDK from 19 to 21.

- We do a few minor upgrades where possible, including the gradlew wrapper version, this causes a few deprecation warnings, but we don't need to act on these yet (and some warnings are from upstream dependencies, so we will just live with them until we can fix all)

I am leaving this as draft until it builds and until I have time to test with my physical Android phone.

### DIFF
--- a/Android/agsplayer/app/build.gradle
+++ b/Android/agsplayer/app/build.gradle
@@ -14,7 +14,8 @@ android {
     defaultConfig {
         applicationId "uk.co.adventuregamestudio.agsplayer"
         minSdkVersion 19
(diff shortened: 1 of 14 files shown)


ericoporto:
I am going to close this as this is probably outdated and look at this again later. If someone wants to take over the Android stuff I wouldn’t mind, but I will eventually go back to this.

PR #2657, opened 2025-01-15
Editor: support loading 8-bit images with Alpha in palette

ivan-mogilko:
This lets Editor to correctly import 8-bit indexed PNGs that have valid alpha component in its palette.
The previous versions of AGS loaded all PNGs as 32-bit images, which solved this problem, but made it impossible to import 8-bit PNGs as actually 8-bit images.
AGS 3.6.2 imports 8-bit PNGs as 8-bit images, but now has opposite problem: if such image has to be converted to a 32-bit sprite (in 32-bit game), the alpha channel becomes lost. First it does not detect the presence of alpha, secondly the process of converting a 8-bit image to a final sprite does not use alpha component at all.
This PR solves these issues.

EDIT: I am not very good at graphic editors, and could not yet find a way to create or export such images.
So i am attaching a sprite provided by the user who got this problem:
[NE-01-SS-000.zip](https://github.com/user-attachments/files/18428839/NE-01-SS-000.zip)

### DIFF
--- a/Common/gfx/image_file.cpp
+++ b/Common/gfx/image_file.cpp
@@ -883,7 +883,7 @@ PixelBuffer LoadPCX(Stream *in, RGB *pal) {
                     pal[i].r = static_cast<uint8_t>(in->ReadInt8()) / 4;
                     pal[i].g = static_cast<uint8_t>(in->ReadInt8()) / 4;
                     pal[i].b = static_cast<uint8_t>(in->ReadInt8()) / 4;
-                    pal[i].filler = 0;
+                    pal[i].a = 0; // filler
                 }
                 break;
             }
--- a/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
+++ b/Editor/AGS.Editor/Panes/SpriteImportWindow.cs
@@ -142,16 +142,16 @@ namespace AGS.Editor
             cmbFilenames.Items.Clear();
             imageLookup = new List<string>();
 
+            OneTimeControlSetup();
+
             foreach (string filename in filenames)
             {
                 imageLookup.Add(filename);
                 cmbFilenames.Items.Add(Path.GetFileName(filename));
             }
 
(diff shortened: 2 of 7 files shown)


ericoporto:
I tested this and confirms it works - I was curious if it would impact #1771 but it doesn't.

The only thing I noticed in particular is that in the sprite selection, the ones that were transparent alpha 8-bit at import time got the transparent pixels to become magenta, and the ones that are 32-bit regular alpha got actual transparent pixels for the transparent pixels.

image

I also noticed that while they are 8-bit at import time they get converted to 32-bit - I think this is correct, and expected and this has always been the case before, and only ags4 has the option to keep the original bitness, but I thought I should mention anyway just to confirm this doesn't change.

ivan-mogilko:
> The only thing I noticed in particular is that in the sprite selection, the ones that were transparent alpha 8-bit at import time got the transparent pixels to become magenta

I did not notice this, can you clarify the sprite format and import settings?

ivan-mogilko:
@ericoporto 
> The only thing I noticed in particular is that in the sprite selection, the ones that were transparent alpha 8-bit at import time got the transparent pixels to become magenta,

Could you attach a image that causes this? I was not able to find how to create such indexed sprites with the graphic software that I'm used to.

ericoporto:
I tried with my own image attached in the masks test issue, here it is again

https://github.com/adventuregamestudio/ags/files/9541552/mask_test_3.zip

The other one is from Don't Give Up the Cat

https://github.com/ericoporto/dont-give-up-the-cat/blob/main/dont-give-up-the-cat/Sprites/DandelionFx/dandelion_particle_02.png

ivan-mogilko:
So, these both images are recognized as having no alpha component, that's why they have their transparency turned into "magic pink".

ericoporto:
It's weird that before #1772 they didn't even import because of these alphas, but that is alright, I think this doesn't change things at all in-game.

PR #2658, opened 2025-01-15
Refactor Character and Object Move functions, optimize WalkStraight

### LINKED ISSUE #1879: Player can move into non walkable area and gets stuck

ivan-mogilko:
Fixes #1879, along the way.

This is a backport of a code refactor from ags4, that tidies up Character and Object Move and Walk functions, bringing them to consistency of parameter checking and delegating actual work to the internal implementation.
Except here I also did this for deprecated script functions.

The second change is, this optimizes WalkStraight function by not running a pathfinder with "anywhere" parameter. Instead makes it use result of previously called CanSeeFrom method, which is already supposed to tell a walkable path.

As a side effect, this makes WalkStraight also use the position fixup in StopMoving, where engine calls PlaceOnWalkableArea in case character ended up on a non-walkable area by mistake.

NOTE: I was considering writing a different algorithm to use here instead of PlaceOnWalkableArea, but decided to leave this for later. And there's a small concern about backwards compatibility, because it's known that changes in pathfinding may break some old games relying on exact character positions. Maybe we could still do that in AGS 4. I left a comment there about my idea.

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -990,6 +990,10 @@ void Character_StopMovingEx(CharacterInfo *chi, bool force_walkable_area)
     // end on a non-walkable area and gets stuck.
     if (force_walkable_area && (chi->walking < TURNING_AROUND) && (chi->room == displayed_room))
     {
+        // TODO: don't use PlaceOnWalkable, as that may result in moving character
+        // into the random direction. Instead, consider writing a "backtracing"
+        // function that runs along MoveList's current stage back, looking for
+        // the first pixel on a walkable area.
         Character_PlaceOnWalkableArea(chi);
     }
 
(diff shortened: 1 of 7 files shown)


ericoporto:
>  I was considering writing a different algorithm to use here instead of PlaceOnWalkableArea, but decided to leave this for later.

My understanding of your backtracking was it would pick the previous point of the pathfinder, and the current one, and then trace a line between the points, starting from the current one and then stopping in the first walkable pixel it finds. Is this the correct understanding?

ivan-mogilko:
> My understanding of your backtracking was it would pick the previous point of the pathfinder, and the current one, and then trace a line between the points, starting from the current one and then stopping in the first walkable pixel it finds. Is this the correct understanding?

Yes.
The purpose of that such method would guarantee that character stops on a point within the path, and not some random point around the stopping location.

ericoporto:
This really sounds like a better approach. Do we have this last point of the walkable area still available or has it been already popped when we need it to do this algorithm.

ivan-mogilko:
> Do we have this last point of the walkable area still available or has it been already popped when we need it to do this algorithm.

There's a MoveList struct which contains information of the path, speed vector (xpermove, ypermove), and current stage index.
Current character's x,y position is where it is. Knowing this information is enough to write the algorithm.

Ags3 has an extra complication, where MoveList has an optional "final segment" which is enabled when character reaches one coordinate sooner than another, so that has to be taken into account. AGS 4 does not have that, because it's all in floats.

PR #2662, opened 2025-01-18
New compiler: Initializers for global classic arrays and structs

### LINKED ISSUE #2152: Scripting: support initializer lists
It's been a long time suggestion to be able to initialize arrays using initializer lists. I suppose this may also be applied to plain structs (PODs), but not managed structs (these would require C++/C# like constructors or factory methods).

Proposed syntax matching C/C++ style:
```
int myArray[5] = { 3, 7, 21, 9, 73 };
```
```
struct MyStruct
{
    int a;
    float b;
    String s;
};

MyStruct ms = { 10, 5.5, "test" };
```

I suppose that this syntax may be translated to a sequence of assignments for runtime, but any type and boundary checks should be done at compile time, therefore no bound check ops are necessary to generate.

---

There's a curious question of whether this operation may be done fully compile-time, and when this makes sense to do.
From a quick thought, because only global memory is generated at compile time, and local vars are allocated by stack ops, then compile time assignment only makes sense for global data, and obviously only if the assigned values are const (or "constexpr", if we want this term in ags script).

fernewelten:
Fixes #2152

I've implemented the possibility to initialize global structs and global classic arrays, using a similar syntax as we (now) have for parameters.

## Background 
Global variables can be initialized at compile time: The compiler essentially prepares in a byte buffer an image of all the global variables with the initial values already set so that the variables can be used at run time without any further processing. (Managed variables can only be initialized to `null` this way.) 

Of course, this can only work when the compiler knows what initial values those global variables should have.  Previously, there was no way of specifying the initial values for `struct` variables and for classic arrays. Now there is a way.

I've implemented the following initializations:
## Initializer for non-managed `struct` variables 

```
struct Car
{
    bool HasAirbags;
    float MaxSpeed;
    float MaxAcceleration;
};
Car OldBugatti = { MaxAcceleration: 50.0, MaxSpeed: 199.8, }; // ← Comma in front of '}' is allowed but optional
```
Notes: 
- The ordering of the fields is immaterial. What matters is how the fields are named in the list.
- All the fields are set to binary zeros where the init value isn't specified. In, e.g., the above example, `HasAirbags` is set to binary zeros, i.e., to `false`
- The `struct` __may__ have managed fields such as `String`s, but the only value that you can initialize them to is `null`.
- If a `struct A` extends a `struct B` and you initialize a variable of type `A`, then you can initialize all the fields from `A` and `B` by just naming them:
```
struct Vehicle 
{
    float MaxSpeed, MaxAcceleration;
};
struct Car extends Vehicle
{
    bool HasAirbags;
};
Car OldBugatti = { MaxAcceleration: 50.0, HasAirbags: false, }; 
```
- You __must__ use the notation `fieldname : value`. You cannot simply list the values without the fieldnames in some sequential order (I haven't implemented that for now).
- A `struct` variable can have a field that is another `struct`. In this case, the initializers are nested:
```
struct Location 
{ 
    float Longitude, Latitude; 
};
struct PointOfInterest
{
    int AdmissionFee;
    Location WhereItIs;
};

PointOfInterest EiffelTower = {
        WhereItIs: { Longitude: 48.8584, Latitude: 2.2945 }, 
        AdmissionFee: 20 + 3,    // ← Simple 'int' or 'float' expressions are possible
};
```

## Initializers for classic (i.e., non-dynamic) array variables
I provide several ways:

### Sequence initialization
```
int Primes[10] = { 2, 3, 5, 7 };
```
Notes:
- If your list contains fewer values than the dimension of the array, then the rest will be filled up with binary zeros. If your list contains more values, then the compiler will balk.
- You may use simple `int` and `float` expressions that can be evaluated at compile time
- You may not mix Sequence initialization and Named initialization (cf. below, including examples) within the same list
- You may not initialize multi-dimensional arrays (cf 'multi-dimensional arrays' below) by just listing all the values without inner braces (cf. below for an example)

### Named initialization
```
bool IsPrime[10] = {
    [2]: true,     // ← means, IsPrime[2] is 'true'
    [3]: true,     // ← means, IsPrime[3] is 'true' etc.
    [5]: true,
    [7]: true,
};
```
Notes:
- All the non-mentioned indices are set to binary zeros
- The compiler will balk if you attempt to set an index more than once
- The sequence in the list is immaterial. 
- This variant is useful if most of the values in your array are zero and only a few values are different from zero (so-called sparse arrays)
- You may not mix Named initialization and Sequence initialization within the same list. For instance, the compiler will balk at `{ 1, 2, [5]: 99 }` or `{ [0]: 2, [1]: 3, 5, 7, }`

### Multi-dimensional classic arrays
An `int Foo[2][3]` is treated as `2` arrays, each of which has `3` values. So this is how the array is initialized:

```
int Spreadsheet[2][3] = {
    { 1, 2, 3, },
    { 4, 5, 6  }
};
```

Notes:
- `C++` offers the possibility to omit all the inner braces and simply list all the values. I have __not__ implemented that. For instance, you __cannot__ define `int Spreadsheet[2][3] = { 1, 2, 3, 4, 5, 6);`
- You can use both Named and Sequence initialization for multi-dimensional arrays, but you may not mix them within the same list. So you __can__ do, e.g.,
```
// The outer list is Named, both the inner lists are Sequence
int Spreadsheet[2][3] = {
    [0]: { 1, 2, 3, },
    [1]: { 4, 5, 6  },
};
```

### Special case: Classic one-dimensional `char` arrays

These can also be initialized by a string literal. 
```
char SafeCode[10] = "AABAACAAD";
```
Notes:
- The string literal __including the terminating `\0`__ must be at most as long as the array, otherwise the compiler will balk. This is a safeguard attempt for when this array is passed as an argument to a function (e.g., `Display(SafeCode);`)
- If the string literal is shorter than the array, then the rest of the array is filled up with binary zeros.

## Local `struct`s and local classic arrays

Unfortunately, not supported for now. If you define an array within a function, you'll have to initialize its values the old-fashioned way. These initializations need to happen at runtime, anyway.

### DIFF
--- a/Compiler/script2/cc_compiledscript.cpp
+++ b/Compiler/script2/cc_compiledscript.cpp
@@ -215,7 +215,7 @@ bool AGS::Snippet::IsEmpty()
 
 void AGS::RestorePoint::Cut(Snippet &snippet, bool const keep_starting_linum)
 {
-    int32_t const orig_codesize = _scrip.Codesize_i32();
+    CodeLoc const orig_codesize = _scrip.Codesize_i32();
 
     // Reset the code to the remembered location.
     CodeLoc rcl = _rememberedCodeLocation;
--- a/Compiler/script2/cc_compiledscript.cpp
+++ b/Compiler/script2/cc_compiledscript.cpp
@@ -82,7 +82,7 @@ AGS::GlobalLoc AGS::ccCompiledScript::AddGlobal(size_t const value_size, void *v
 {
     assert(globaldata.size() + value_size <= INT32_MAX);
     if (0u == value_size)
-        return static_cast<int32_t>(globaldata.size()); // nothing to do
+        return static_cast<GlobalLoc>(globaldata.size()); // nothing to do
 
     // The new global variable will be moved to &(globaldata[offset])
     size_t const offset = globaldata.size();
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -551,6 +551,10 @@ std::string const AGS::SymbolTable::GetName(AGS::Symbol symbl) const
 
 AGS::Vartype AGS::SymbolTable::VartypeWithArray(std::vector<size_t> const &dims, AGS::Vartype vartype)
 {
+    if (dims.empty())
+        // A zero-dimensional array is a "no-op".
+        return vartype;
+
     // If 'vartype' is an array, the new array index must be spliced into the vartype name
     // in front of the first '['
     std::string pre = GetName(vartype);
@@ -598,6 +602,24 @@ AGS::Vartype AGS::SymbolTable::VartypeWithArray(std::vector<size_t> const &dims,
         entries[array_vartype].VartypeD->Dims = dims_to_use;
     }
     return array_vartype;
+}
+
+AGS::Vartype AGS::SymbolTable::ArrayVartypeWithoutFirstDim(Vartype const vartype)
+{
+    if (!IsVartype(vartype))
+        return kKW_NoSymbol;
+
+    auto &dims = entries[vartype].VartypeD->Dims;
+    size_t const dims_count = dims.size();
+    if (0u == dims_count)
+        return vartype;
+
+    Vartype const base_vartype = entries[vartype].VartypeD->BaseVartype;
+    if (1u == dims_count)
+        return base_vartype;
+
+    std::vector<size_t> const new_dims{ dims.cbegin() + 1u, dims.cend() };
+    return VartypeWithArray(new_dims, base_vartype);
 }
 
 AGS::Vartype AGS::SymbolTable::VartypeWithConst(AGS::Vartype vartype)
@@ -753,7 +775,8 @@ AGS::Vartype AGS::SymbolTable::GetFirstBaseVartype(AGS::Vartype vartype) const
 {
     // TODO: extra safety checks?
     for (const auto *ste = &entries[vartype]; ste->VartypeD && (ste->VartypeD->BaseVartype > 0);
-        vartype = ste->VartypeD->BaseVartype, ste = &entries[vartype]) { }
+        vartype = ste->VartypeD->BaseVartype, ste = &entries[vartype])
+    { }
     return vartype;
 }
 
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -787,12 +787,15 @@ bool AGS::SymbolTable::IsOldstring(Symbol s) const
     Vartype const s_without_const =
         VartypeWithout(VTT::kConst, s);
 
-    // string and const string are oldstrings
+    // 'string' and 'const string' are oldstrings
     if (kKW_String == s_without_const)
         return true;
 
-    // const char[..] and char[..] are considered oldstrings, too
-    return (IsArrayVartype(s) && kKW_Char == VartypeWithout(VTT::kArray, s_without_const));
+    // 'char'[..] considered oldstring, too
+    return
+        IsArrayVartype(s) &&
+        1u == ArrayDimensionsCount(s) &&
+        kKW_Char == VartypeWithout(VTT::kArray, s_without_const);
 }
 
 AGS::Symbol AGS::SymbolTable::Add(std::string const &name)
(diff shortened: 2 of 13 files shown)


ivan-mogilko:
This sounds great! I shall test this soon.
I have a question regarding two unimplemented things though:

1. On struct initializer:
> You must use the notation fieldname : value. You cannot simply list the values without the fieldnames in some sequential order (I haven't implemented that).
2. On local structs and arrays:
> Unfortunately, not supported so far. If you define an array within a function, you'll have to initialize its values the old-fashioned way. These initializations need to happen at runtime, anyway.

Are these technically possible and may they be planned for the future, or not at all?

Personally, I believe that having a ordered struct initializer will be very convenient for simple structs and also consistent with the function argument list syntax (where you can pass arguments either ordered or named).

Local initialization: in theory these can be done by preallocating a struct and then filling them with results of expressions inside initializer. But also, since they are executed at runtime, these may contain non-constexpr expressions, like function calls and `new` operators.

fernewelten:
>> You __must__ use the notation, `fieldname : value`. You cannot simply list the values without the fieldnames in some sequential order (I haven't implemented that).

Yes, it's doable: The fields are ordered in the symbol table:  They come in the order in which they were defined. If the struct has been extended, then the ancester fields and perhaps the fields of _their_ ancesters etc. must be considered, too. We can define a sequence from that, e.g., ‘first the fields of the struct proper and then the field of its ancester, then those of _it's_ ancester … ’, and then a list `{ field1, field2, field3, … }` can be matched to this sequence. 

From a practical perspective of the user-programmer, a sequence list saves some typing in comparison to a named list, but when the fields in the struct get moved around afterwards for some reason then it's _very_ error-prone to find and change around the respective initializations.

> Local initialization: in theory these can be done by preallocating a struct and then filling them with results of expressions inside initializer. But also, since they are executed at runtime, these may contain non-constexpr expressions, like function calls and new operators.

Yes, that's also doable. It would be a different piece of code that would have to be added to another place in the compiler than where the globals are initialized, but it can definitely be done.

I'm not against doing those in principle, it's just that I haven't implemented them for now.

ivan-mogilko:
There are few things that I found while testing this, both related to the new changes and not.

1. My proposal is to change the error message for trying to assign something to a managed pointer.
Right now it sais:
> Expected 'null', found '"abc"' instead

I'd suggest to say something elaborate, such as:
> Global pointers may only be initialized with null.

2. By accident I found that compiler does not like initialization of global dynamic arrays to null, in any case:
```
int dynamicarr[] = null;
```
or
```
struct MyStruct { int dynamicarr[]; };
MyStruct s = { dynamicarr: null; }
```

Both of those make it throw error:
> Type 'int[]' cannot be initialized

I think that's a mistake, and it should be able to init them to null, just like any regular pointer initialization.

3. Another thing not related to the new changes, there's this old error message, when you have `=` or an arithmetic operator somewhere in a random place:
> '=' is illegal outside a function

This is not correct, because you can use `=` and some others (+, -, /, * etc) when initializing variables. I think this message should be adjusted somehow, not sure how to phrase this though.

fernewelten:
> There are few things that I found while testing this, both related to the new changes and not.

Starting to fix those

ivan-mogilko:
I don't know how long will it take to implement local initializers, perhaps this part may be merged, if no one sees any issue with the syntax and finds no bugs?

ivan-mogilko:
There's just one thing that bothers me in the suggested syntax here, it's array initialization where you can define array indexes:

```
bool IsPrime[10] = {
    [2]: true,     // ← means, IsPrime[2] is 'true'
    [3]: true,     // ← means, IsPrime[3] is 'true' etc.
    [5]: true,
    [7]: true,
};
```

you have to write indexes in brackets. Would it be better, for simplicity of syntax, to have them without brackets instead? like
```
bool IsPrime[10] = {
    2: true,     // ← means, IsPrime[2] is 'true'
    3: true,     // ← means, IsPrime[3] is 'true' etc.
    5: true,
    7: true,
};
```

What do everyone think about this?

AlanDrake:
I'd rather the second, mostly becasue it reminds me of JSON and I'm used to it.

fernewelten:
> Would it be better, for simplicity of syntax, to have them without brackets instead?

We can have it that way, no problem. It's deleting something like two lines of code.

ivan-mogilko:
If it takes too long to implement this kind of initialization for local variables, then we might just merge the global ones for now.
The PR only has to be rebased on top of ags4, because there are some code conflicts again.

ivan-mogilko:
Merged a rebased version of this: #2715

PR #2665, opened 2025-01-22
AGS 4: implement dynamic cast instruction in the engine

### LINKED ISSUE #2018: Support script object pointer downcast
CC @fernewelten 

After [RTTI was added](https://github.com/adventuregamestudio/ags/pull/1922) in ags4, it is now theoretically possible to find out the managed object's parent(s), knowing its type. This opens a possibility to support downcasting, that is - a conversion from the base type pointer to a child type pointer. As opposed to the upcasting, which may be calculated at compile time, downcasting has to be performed at runtime, as compiler cannot know which type will be stored in the base pointer.

What would be required for this?

1. **From compiler's side**: a pointer cast syntax. The two obvious options here are:
* C style: `Child* child_ptr = (Child*)parent_ptr`;
* C# style: `Child* child_ptr = parent_ptr as Child`;
* there's also C++ style, with `cast<T>(parent_ptr)` or `cast<T*>(parent_ptr)`, but it looks more complicated.
* EDIT: an approach not requiring a bytecode, but another instance of a "psuedo-function" (like array_Length): a static method that casts from parent to the given type, something like
```
child_ptr = ChildType.CastFrom(parent_ptr);
```
2. A new bytecode instruction, which requests a dynamic cast. This instruction likely should take a ptr from a reg, analyze if it's possible to cast, cast, and store back in a reg (or reg1->reg2 if it's more convenient for any reason). If cast fails, a null pointer is written.
3. **From the engine's side**: an ability to know object's type. The user managed structs are already storing typeid in their headers. The builtin objects do not; they may get one using the "manager"; but I have not thought this through yet. This also may be related to resolving #1228 first.
4. There's an interesting question about types exported by plugin. Script compiler may see their relationship, and they should normally be a part of RTTI, but I am not sure how to match an object received from the plugin with RTTI yet. Something is missing in this picture.

ivan-mogilko:
Resolves #2018, this implements the dynamic cast on the engine side; compiler has it implemented by #2693.

This implements a dynamic cast instruction in the engine, meaning a cast between two managed pointers performed at runtime, when the engine is capable to know which type exactly given pointers refer to.

I had to do a number of other changes just to make this work, both in compiler and engine.
It will take some time to explain everything, so please be patient reading this... Here's what was done and why.

### Matching runtime types in engine with RTTI

The types in RTTI have bare names, and fully qualified names, where fully qualified name has a form of "location::type" ("location" is just a term for a namespace). That is done because in theory there may be multiple types of same name but in different scripts (if they are inside scripts, and not in headers); in which case fully qualified name lets distinguish those. Normally "location" is taken from a script file's name (header or script body), since there's no explicit way in AGS to define a namespace.

Now, if we want to match types known by the engine and plugins with types from RTTI generated by the script, we meet a problem: at compilation time these types may be declared in a script header of any arbitrary name. This means that if we want to find these types in RTTI we cannot rely on the script name at all.
One may assume that engine API is always included, so if you find any type of matching name, then it must be engine's type. But firstly that's not 100% guaranteed (it's potentially possible to compile a script without adding any API header at all). And secondly, that does not solve the problem for plugin types, as we simply do not know which types did plugins declare, because plugin API does not provide any means to tell that to the engine (which may be an oversight in plugin api design, but that's another story).

This brought me to idea, that the script should have means to explicitly declare that certain script contains types from the engine or plugin.

How did I try to solve this here:
I introduced a new preprocessor command called `#module`. This command lets specify the "module", or in other words a "group", where all the types declared within the given script should be assigned to. When preprocessor finds such command `#module modulename`, it updates the `__NEWSCRIPTSTART_` marker in the script's beginning, appending the module's name to the script's name, separating them with semicolon, e.g. `__NEWSCRIPTSTART_ScriptName.ash;modulename`.

When the script compiler later parses the script and finds this marker, it saves this appendix as a "module name" along with the "section name" when compiling the script. Section name is used for reporting compiler messages, as usual, and module name is currently used only when generating RTTI table. When writing a list of types in RTTI, compiler would use module's name as a type's "location", if it's available, instead of a section name.

The engine API header has a new directive: `#module ags`, where "ags" is going to be a standard name for anything declared by the engine. This makes all types declared within the builtin header to get into "ags" group.

The plugin API headers have following directive prepended: `#module pluginname`, where pluginname is a plugin's file name without extension, e.g. `#module agsspritefont`. This name can be easily matched with the plugin name used by the game data and the engine at runtime.

When the engine loads a game up, and generates joint RTTI table, it will create lookup aliases for all types from group "ags" and groups related to plugins (using known plugin names). These aliases are simply bare names of types, without any "location::" prefix, and they will be used during dynamic casting, as explained in a following section.

### Dynamic cast

Introduced a new script opcode  SCMD_DYNAMICCAST = 76, which has 1 argument. This opcode assumes that AX register contains a managed object pointer, and arg1 (literal) is a local typeid of a type which you want to cast to. (Local typeid is script-relative, these are mapped to global typeids in joint RTTI table.) It is supposed to keep same pointer in AX if cast is successful, or write 0 if it was a failure, simple like that. (There's no need to write another pointer, because pointer is always same, regardless of how script "sees" the variable.)

The implementation of SCMD_DYNAMICCAST step-by-step is this:
* take object pointer from AX register;
* if pointer is null, then do nothing (null always casts to null);
* get respective object manager from the managed pool; if none is found, then fail, writing null ptr to AX.
* ask manager about the object's typeid. Only "user structs" currently can return that, others return failure. (I suppose this may be improved at some point, it's strictly implementation problem.)
* on failure, ask manager about the type name, and use this name to find global typeid in a lookup table. That's what these aliases were meant for.
* if nothing was found (that's weird), fail, writing null ptr into AX.
* when we got real object's typeid, this way or another, we check whether the real typeid matches the type we cast to, OR if any of that type's parent typeid matches the type we cast to. If any of that is correct, then the operation is a success.
* otherwise it's a failure, and we write null ptr into AX.

### Changes to dynamic managers

As may be seen from above explanation, the crucial part of the cast for builtin type is being able to get typename from dynamic manager. This also requires that our dynamic managers in the engine return typename exactly matching their counterparts in script. And not only that, but also we must make sure that each builtin object is registered under precise typename, not some "generic" type name.

Many of the existing managers have this right, and their GetType() method return matching strings. But there were 2 kinds of managers that do not:

1. Some managers had historically slightly different names (like "GUIObject" vs "GUIControl"), and some were renamed when their save format changed at some point in the past ("Camera" and "Viewport"), which was a dumb move on my part.
2. GUI control subclasses all uses same GUIObject manager.

Therefore I had to make changes to this (keeping in mind to also handle old names in serializer, in case an older save is loaded).
1. All managers made to return typenames precisely matching the struct names in script API.
2. Created subclass managers for each subclass of GUI and GUIControl, and register gui objects with proper manager.

### Testing

After #2693 ags4 compiler supports following syntax: `pointer as Type`. For example:
```ags
Label* label = gMyGUI.Controls[0] as Label;
```

This operation may be used with any managed type, whether builtin or declared by user.

Need to test:
- [x] Bultin managed classes that have parents.
- [x] User managed structs.
- [x] Plugin script structs.

### TODO

- [x] - Compiler syntax for casting (separate task?).
- [x] - Fix dynamic manager for subclasses (gui controls) when restoring an older save. It's possible to do, although may be annoying to code.


fernewelten:
Okay, so I'll provide an 'as' operator in the compiler.

I'm wondering about the technical side of that. My poor puny git-fu isn't up to such a task.

What we'd need is a kind of a PR for your PR so that I can send some compiler changes over to you and if you like them, you can integrate them into your PR and then at a later stage integrate your PR into the AGS repository.

There's probaby a way for this, but I've never been involved in such a case, and I'm unsure what I have to do to make it happen on my side.

ivan-mogilko:
> What we'd need is a kind of a PR for your PR so that I can send some compiler changes over to you and if you like them, you can integrate them into your PR and then at a later stage integrate your PR into the AGS repository.

First, there's a "patch" feature in git that creates a file with code diff from a commit or range of commits:
https://git-scm.com/docs/git-format-patch

Second, I think on github there's a way to give a permission to push more commits to PR to somebody else, but I've never used that myself.

ivan-mogilko:
> So I'd suggest to make the operation work with AX instead of MAR. Or perhaps change the operation to take two arguments, a register to work with and the local type (analogous to SCMD_NEWUSEROBJECT2).

I am not opposed against any of these suggestions, so let's do one that's convenient.

ivan-mogilko:
Rebased again to resolve conflicts.

ivan-mogilko:
Rebased after merging #2693. Applied necessary changes (use AX register instead of MAR), updated PR description.

This is ready for testing now.

ivan-mogilko:
I tested all 3 variants of structs:
- builtin exported by engine
- declared in user scripts
- declared and exported by plugins
For all 3 kinds dynamic cast works.

For engine and plugins, this requires dynamic object manager to implement GetType() which returns not an arbitrary name, but a name of the type matching the name declared in script.

PR #2666, opened 2025-01-22
New compiler: incorrect compile time left bitshift overflow

AlanDrake:
During compile time, attempting to left shift with `int a = 0xFF << 24;` was causing an overflow error where none is expected. 

One test was catching a bogus overflow for `536870912 << 2` but that's just `0x80000000` which is valid, so I amended it to `<< 3` to cause a real overflow.
I also added an extra test to make sure it can shift to the highest bit unimpeded.

Was this "overflow" done on purpose to avoid touching the sign bit? If that's the case I don't agree, as it's a too big departure from runtime, and prevents from filling valid bit positions.


ivan-mogilko:
This looks correct, and compile-time shift calculation matches runtime one
(NOTE: since new compiler does compile-time arithmetics when values are constants, you have to use variables in shift operation to force it to be performed at runtime).

PR #2667, opened 2025-01-23
New compiler: Fix bug where naming a parameter like its function caused a crash

### LINKED ISSUE #2647: New compiler: erroneously reports that parameter "hides" a function
CC @fernewelten 

New compiler has a bizzare mistake, when it will report that certain parameter "hides" a function if their names match.

Example of a script that causes this:

```
import void func(int param);
import void param();
```

Compiler reports a warning:
``` GlobalScript.asc(1): Parameter #1: This hides the function 'param()'. See line 2 ```

### PR DESCRIPTION
**Pull request by fernewelten**
Fixes #2647.

Parameters act as variables within the function for which they are defined. This leads to trouble inside the compiler when a parameter is named like its own function. Added code that forbids this.

Typical code:
```
function foo(int foo, int x) 
{}
```

I took this as an opportunity to streamline the function defining code in general.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -872,18 +872,18 @@ Symbol AGS::Parser::ParseConstantExpression(SrcList &src, std::string const &msg
     return eres.Symbol;
 }
 
-Symbol AGS::Parser::ParseParamlist_Param_DefaultValue(size_t idx, Vartype const param_vartype)
+Symbol AGS::Parser::ParseFuncdecl_Parameters_ParamDefaultValue(Vartype const param_vartype, size_t const idx)
 {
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
I'm actually not certain if it is correct to disable parameters of same name, because local variables of same name are allowed, so this makes things not consistent.

ivan-mogilko:
@fernewelten do you think it will make sense to support using function name in arguments for consistency with the local variables?

If not, then I'd merge this as is, since this fixes a crash.
It only has to be rebased on top of the latest ags4 first.

ivan-mogilko:
Cherry picked commit and pushed to the ags4 branch: 0bb827ae1ce20204de902c9ffdc1cc43790addcb

PR #2668, opened 2025-01-23
AGS 4: Add BlendMode to DrawingSurface

ivan-mogilko:
This adds DrawingSurface.BlendMode property, that is used by all operations that also use DrawingColor.

Adjusts custom blenders to work with 32-bit ARGB colors, and fixes the incorrect color shifts in them, which switched R and B parts. This change might affect how the object blend mode work in software mode, so that has to be tested too.

NOTE: not working for DrawString & DrawStringWrapper atm, because I need to figure out how to pass blend mode parameter into the internal font renderers.

### DIFF
--- a/Common/gfx/blender.cpp
+++ b/Common/gfx/blender.cpp
@@ -286,3 +286,39 @@ uint32_t _blender_masked_multiply32(uint32_t x, uint32_t y, uint32_t n)
     return _blender_mask_alpha24(_blender_multiply24(x, y, n), x, y, n);
 }
 // ===============================
+
+#include "gfx/gfx_def.h"
+
+namespace AGS
+{
+namespace Common
+{
+
+// Array of blender descriptions
+// NOTE: set NULL function pointer to fallback to common image blitting
+typedef BLENDER_FUNC PfnBlenderCb;
+static const PfnBlenderCb BlendModeSets[kNumBlendModes] =
+{
+    _argb2argb_blender,             // kBlend_Alpha
+    _blender_masked_add32,          // kBlend_Add
+    _blender_masked_darken32,       // kBlend_Darken
+    _blender_masked_lighten32,      // kBlend_Lighten
+    _blender_masked_multiply32,     // kBlend_Multiply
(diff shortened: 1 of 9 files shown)


ericoporto:
Hey, I noticed that internally it can be set a null function to fallback to blit instead of blend, but I don't understand how to set it in the function.

ivan-mogilko:
> Hey, I noticed that internally it can be set a null function to fallback to blit instead of blend, but I don't understand how to set it in the function.

DrawingSurface does not have a explicit differentiation between blit and blend, it always blends so long as it's 32-bit and color alpha is translucent. For blitting you simply pass an opaque color.

ericoporto:
No, I think you are correct that it's fine, a blit as blendmode may be confusing when the same type is used outside of drawing operations - like as a property of an object.

ivan-mogilko:
My apologies, I was meant to add a test game which i've been using to a previous PR, but forgot.
Attached one to this ticket now:
[ags4--alphadrawing.zip](https://github.com/user-attachments/files/18527738/ags4--alphadrawing.zip)

ivan-mogilko:
Made text output work with the blend mode.

But this does not work properly with *automatic outlines*, and in fact translucent alpha does not work with them in general. The reason is that auto-outline is done by painting text with back color on an intermediate bitmap, and then drawing actual text above it. As a result, the text's color gets blended not with the underlying drawing surface, but with the outline, producing completely unexpected effect.
This will have to be fixed separately, I added a FIXME comment in the code.

ivan-mogilko:
Tested regular objects in software render mode, and BlendMode still works, at least visibly so, and colors are correct now.

If there's no objection to this feature, I'd merge the pr, and then make a new 4.0 alpha update with new color logic.

ericoporto:
I haven't had the time to test but a new AGS 4 release is something I have been waiting for since the colors change started so I am all in for it - plus it should also receive a merge from the master branch to get the fix for `/compile` showing the "save" alert window, so I can update the test games projects too.

Issue #2575, opened 2024-11-11
Make engine work reliably with multiple monitor systems


ivan-mogilko:
**Problem**

Existing graphics mode code assumes primary monitor in many places, whenever the desktop size is retrieved, calculated window size adjusted for the working area, window created, etc. While the game may work on any monitor in windowed mode (after player dragged it from the primary one), and even switch to "borderless desktop window" and back, some calculations are still wrong.

In addition, exclusive fullscreen mode is not working reliably in this case, depending on which graphic driver was used. From what I've read, Direct3D9 cannot make a fullscreen mode on another monitor, only one the one which IDirect3D9 object was created for. Here we need to make a decision: either we want to support it switching monitors or not.
- If we do, then it needs to be taught to recreate itself, which in turn requires to fully reset the texture cache, as all the objects tied with the IDirect3D9 interface will be invalidated.
- If we don't, then we need to resolve a situation where a player switches between a window on secondary monitor and exclusive fullscreen on a primary monitor. Judging by users reports, this causes some glitches, probably because fullscreen mode is created without moving a window to the primary monitor first (not fully certain if that's the reason).

**Tasks**

1. Any code that relies on display index should have display index as parameter instead of assuming "default display". We might need a function that returns "current display" from the window position. An example may be found in this temporary commit which I reverted afterwards (it was not enough, so I did not want to keep dangling changes which i'm unsure about): 4fff528e34c724f042f6b68193f0bf5aa6f5001f . But the code changes in that commit are not complete for this task. Note: besides general code, there's also a code in AGSWin32::ValidateWindowSize() that also assumes primary monitor.
2. DisplayMode struct, used to init a mode in graphic driver classes, should contain "DisplayIndex" value, telling the gfxdriver which monitor to init a mode on. The driver class may override that, if it's aware that it cannot perform this request (speaking primarily of Direct3D9 here). If driver creates a window (because it does not exist yet), then it creates and positions it on the requested display. If window already exists, then it moves it to requested display.
3. Optionally we might have "display" selection in winsetup, if this seems like a good idea, if not then this will be always defaulted to 0.
4. There's a code that remembers last settings for each mode (windowed, fullscreen). It should also remember resulting display index, because if window was moved to another display by a user, then old settings may not longer be applicable (resolution).
[...]

PR #2669, opened 2025-01-27
Implement initial display selection

### LINKED ISSUE #2280: Support some way to set the starting monitor
**Describe the problem**
AGS by default enters full-screen in the primary monitor (on Windows), but it may be useful to have it full-screen in a different monitor - either because the user has the primary monitor in vertical, or the secondary monitor is a TV intended for gaming, or even if the user is developing the game in the AGS Editor and has it open in the primary monitor and wants to debug the game in a secondary monitor.

**Suggested change**
Some config in acsetup.cfg/command-line/winsetup that tells the game to prefer a different monitor, in a way that if such monitor doesn't exist, it still fallback to the default behavior of using the primary monitor.

This is only meant for it to be the initial monitor, currently it's possible to move the window to a different monitor and when it enters in full-screen the window does full-screen in the same monitor as it was.

If a command-line option is added, it would be nice to have a setting in the Editor Preferences that you can set and would pass the preference when testing the game.

Note:  I am only aggregating previous suggestions and questions from others, I only have 1 monitor and will take me a bit to test things for this if someone comes up with a solution.

ivan-mogilko:
Resolve #2280, #2575

1. Engine allows to specify display index in DisplayMode struct. All the code that dealt with getting current display's size or available modes have been adjusted to respect a selected display index.
2. Config has a new option: "[graphics] display = N", where N is a display index.
3. Special handling is done in Direct3D renderer, and Software renderer in case SDL created any of the directx drivers: when switching between exclusive fullscreen and windowed mode they will always set exclusive fullscreen on a display where the window *was created* (regardless of a initial mode). That's a Direct3D requirement (and possible new DirectX too). This may only be resolved by recreating a device, which in turn requires full recreation of all resources - textures etc, - and that's just too much of a hassle in my opinion.
4. Added a "Display" dropdown list to WinSetup.

winsetup--displayselect

TESTING:
- [x] - Display mode selection works normally, with all the fullscreen and windowed mode kinds initializing correctly (this may be tested on default display).
- [x] - Switching windowed/fullscreen works normally.
- [x] - Dragging a window to another display and switching to a non-exclusive fullscreen and back should keep window on that display.
- [x] - Dragging a window to another display and switching to a exclusive fullscreen with Direct3D should init exclusive fullscreen on a display where the window was first created.
- [x] - Winsetup lets select display index. When changing display index, resolution selection will be updated to retrieve available modes from the selected display.

POSSIBLE TODOs:
I noticed in #2280 there's a extra suggestion to add a Editor preference that forces testing on a certain display.


ivan-mogilko:
I've tested this on two-monitor system, but please test even if you have 1 monitor, to make sure that basic configuration keeps working, because I had to modify several things in graphics mode init.

ericoporto:
I can test this sometime tonight for 1 monitor in macOS and two monitors in Windows.

ericoporto:
Testing on Windows 11, just using winsetup to change the initial display, it seems it works. I select 0 and get my primary display and I select 1 and get my secondary display. I think it would make more sense if it was 1 and 2, like it shows in Windows multiple displays settings, where the monitors get assigned a number from 1 to how many they are.

About using the feature, the only curious thing is, if there is an application on the secondary display, the AGS game doesn't seem to load in "focus", while on primary display it seems it always loads in "focus" - by focus here I mean it loads the window or fullscreen below the application that is on secondary display, but also not focused so that the game doesn't start until one clicks the game icon in the start bar.

ivan-mogilko:
So, this is going to be complicated. If the display index is 1-based in config, then there will have to be a point of conversion between 1-based and 0-based index. There are two most natural choices for such point: loading/saving of a config, and passing this index between engine and SDL.

In the first case, there's an issue, that the GameConfig struct will need to support a distinct definition for "use default monitor", otherwise it won't be able to *save* it in config.

In the second case, all calls to SDL will need to be wrapped into our functions that convert from 1-based to 0-based index. And there will be an extra issue of storing an actual display index somewhere, because the engine must remember exact index on several occasions. This will make using this index very bug prone, and code will be littered with conversions.

So probably the first approach will be better.

ivan-mogilko:
> EDIT: If I had to nitpick, the default desktop resolution, and available ones, should change when selecting another monitor.

It is supposed to, is it not? My both displays support same list of resolutions, so I cannot notice the difference.

AlanDrake:
> The list of resolutions is supposed to be refilled, is it not? My both displays support same list of resolutions, so I cannot notice the difference.

I see a repaint when changing Display index, but the select values remain unchanged from what was loaded at start.

ivan-mogilko:
Pushed a fix, please try again.

AlanDrake:
> Pushed a fix, please try again.

It works now. All good.

ericoporto:
I had a thinking that once this is merged, the multi-display is tied to the desktop platforms. While I don't think this will change soon, I wonder if it would be better to have a macro explicitly for this in `core/platform.h` and then have it set to the AGS desktop macro for now - in this way it would be clear which parts of the code are the way they are because of multimonitors and not just as a generic desktop thing.

ericoporto:
Thanks, that's pretty good. I think this looks fine and we will probably only get larger testing once this hits a game in production.

PR #2670, opened 2025-01-27
Editor: update properties referencing GUIs on GUI removal or id change

### LINKED ISSUE #2542: Deleting a GUI doesn't update Custom text-window setting
**Describe the bug**
Parts of the editor reference to GUIs by number.

**AGS Version**
3.6.0.53

**To Reproduce**
1. Create a game with three GUIs.
2. In general settings, set Custom text-window GUI to 2
3. Delete the first GUI.
4. Note that Custom text-window setting is still 2, whereas it should be 1. This will crash the game when it displays a textbox.

**Expected behavior**
If a GUI is deleted, references to GUIs should take that into account. I don't think this is a priority since it has an obvious workaround.

 - Windows 11

ivan-mogilko:
Fix #2542

1. Properties that let select GUI display a dropdown list of gui script names instead of being raw number fields. The displayed names are mapped to GUI number internally, so no data upgrade is necessary.
2. Whenever a GUI is removed, or has ID changed by command, SettingsComponent class updates all of its properties and fixes GUI selection. If selected GUI got removed, property is changed to 0 ("None").

The feature is purely in UI and does not change data format.

### DIFF
--- a/Editor/AGS.Editor/Components/GuiComponent.cs
+++ b/Editor/AGS.Editor/Components/GuiComponent.cs
@@ -86,6 +86,7 @@ namespace AGS.Editor.Components
                         GUI newGUI = ImportExport.ImportGUIFromFile(fileName, _agsEditor.CurrentGame);
                         newGUI.ID = _agsEditor.CurrentGame.RootGUIFolder.GetAllItemsCount();
                         AddSingleItem(newGUI);
+                        GUIIndexTypeConverter.SetGUIList(_agsEditor.CurrentGame.GUIs);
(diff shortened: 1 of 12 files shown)


ivan-mogilko:
I used existing Game.GUIAddedOrRemoved event, and had to add a special event for gui ID changing to GuiComponent.
But I think that these events (both old and new) are a hack. IMO we need some sort of a IGameListener (made up name) interface that would contain all similar events, and through which various managers would notify others about changes they made. Added comments with this idea into the code.

PR #2673, opened 2025-01-30
Improve PlaceOnWalkableArea

### LINKED ISSUE #163: Character.PlaceOnWalkableArea does not work at all times
Character.PlaceOnWalkableArea (obsolete script name: MoveToWalkableArea) script API function is supposed to bring character to nearest walkable area, but sometimes fails.

So far I found this occuring in "Duty and Beyond", room 49.

**UPDATE**

There are 2 issues with PlaceOnWalkableArea:

When searching for walkable area engine does not check every pixel on the mask, but does so in 5 pixel intervals. For example, "Duty and Beyond" features very thin walkable zones (1 or 2 pixel thick), which makes them susceptible to be overstepped.

Another - undocumented - exception is that PlaceOnWalkableArea ignores walkable areas lying beyond Room Edges. Probably that is because Edges are usually used to define exit zones.

ivan-mogilko:
Fix #163

This makes Character.PlaceOnWalkableArea use 1-pixel step, in order to not skip thin areas.

Refactored searching algorithm and moved to Pathfinding::FindNearestWalkablePoint(). This function still accepts search "step" parameter, so in theory we could use it if a need arises.
Renamed original function to FindNearestWalkableAreaForCharacter(). Made it clear which restrictions to the scanning area do we make, according to historical AGS rules (don't move character beyond room edges, etc).
Removed the obscure hard-coded restriction which did not let move character to the topmost 14 pixels of the room. I think that may be related to the Sierra-style Statusbar GUI from the default template.

Optimized scanning algorithm for speed (see explanation below).
Would need to make performance tests to see the actual difference in very big rooms.

**Optimization:**

The original algorithm took whole room (or rather section limited by edges), and scanned it *awhole* from left-top to right-bottom. When finding a walkable point, it would calculate a distance from the character, save the nearest found point, and continue. In the end, it would scan all the room regardless, even if it may be certain that there cannot be any point in a smaller radius already. This of course is quite suboptimal.

The new algorithm that I'm trying here does a *outwards spiral scanning*, checking pixels which lie on a rectangle border, where a rectangle is built with certain radius around the starting position.
That is: it first takes 3x3 rectangle and scans pixels on its border. Then it takes 5x5 rectangle and scans its border. Then it takes 7x7 rectangle, and so forth.
In this algorithm I still keep the original act where it would increment along x axis in the outer loop and increment along y axis in the internal loop. This means that pixels located towards top-left have higher priority among walkable pixels with the same distance. Idk if that matters much, but I kept it like that just in case some old game relies on that (call that a paranoia).
In order to not duplicate work, it actually skips the pixels inside rectangle, because these were already processed when scanning previous rectangles. This is done simply by adjusting internal step: when first and last rect line are scanned, we step by 1 pixel, but when any other line is scanned, we step by radius * 2, which results in only two pixels on the opposite rectangle's side checked.
Here's a illustration:
outwardscanning

Whenever a walkable point is found, it calculates a distance to it, checks if that's the smallest found distance, and also reduces the scanning range to that distance, in order to stop early, since we won't be needing any pixel at a larger distance anymore.
NOTE: that it's still necessary to keep testing for some time even if we find a walkable point, because we scan in rectangles and not in circles. Would we scan in circles, we could stop as soon as any walkable pixel is found, but with rectangles we cannot be immediately certain if that pixel is really the closest. For example, a pixel on rectangle's corner is further away from center than a pixel in the middle of a rectangle's edge.

### DIFF
--- a/Common/gfx/allegrobitmap.h
+++ b/Common/gfx/allegrobitmap.h
@@ -193,6 +193,12 @@ public:
     // bitmap itself and should rather be a part of the game data logic.
     color_t GetCompatibleColor(color_t color);
 
+    // Tells if the given point lies within the bitmap bounds
+    inline bool IsOnBitmap(int x, int y) const
+    {
+        return x >= 0 && y >= 0 && x < GetWidth() && y < GetHeight();
+    }
+
     //=========================================================================
     // Clipping
     // TODO: consider implementing push-pop clipping stack logic.
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -748,7 +748,12 @@ void Character_PlaceOnWalkableArea(CharacterInfo *chap)
     if (displayed_room < 0)
         quit("!Character.PlaceOnWalkableArea: no room is currently loaded");
 
-    find_nearest_walkable_area(&chap->x, &chap->y);
+    Point dst;
+    if (FindNearestWalkableAreaForCharacter(Point(chap->x, chap->y), dst, play.debug_mode > 0))
+    {
+        chap->x = dst.X;
+        chap->y = dst.Y;
+    }
 }
 
 void Character_RemoveTint(CharacterInfo *chaa) {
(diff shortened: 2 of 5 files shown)


ivan-mogilko:
Hmm, the preliminary performance test results are following.
Used a 4000x4000 room (I tried using a larger one starting with 16k x 16k, but Editor kept crashing with out of mem and other errors...) with randomly scattered walkable areas.

Time to find a nearest walkable area:
Old algorithm with 5-pixel step: 50-70 ms on average.
Old algorithm with 1-pixel step: 900-930 ms on average.
New algorithm with 1-pixel step: 0-2 ms (probably depends on where the character is positioned related to walkable areas).

ericoporto:
That is more than 10x performance improvement, so on that alone it looks really good!

PR #2674, opened 2025-02-01
Editor: "Start New Game" Wizard with scalable intro image

### LINKED ISSUE #2622: Editor: adjust the intro page in the New Game Wizard, make height higher
After #1903 (after #2574 in 3.6.2) the template selection page has a "Description" field underneath, but since the page height remains the same, the controls are cramped together, which makes it look bad:

<details>
<summary>Click for screenshot</summary>

image

</details>

IMHO it needs at least 1 more row for the list of templates, and increase the Description field maybe x1.5 in case the text is longer.

The problem is that the first page of the Wizard contains a bitmap, and if we just increase the dialog height, that will look bad too without image adjustment:
https://raw.githubusercontent.com/adventuregamestudio/ags/refs/heads/master/Editor/AGS.Editor/Resources/wizard.bmp

We need to either adjust this image for another, bigger height, or change the design of this page altogether with some form that will allow us to adjust the window size easier on need.

ivan-mogilko:
Fix #2622

1. Remake the WizardDialog's title screen into the separate WizardIntroPage control class.
2. Adjust "start new game" pages to scale to the dialog's size, where makes sense. Dialog is relying on each page's "MinimalSize" property when calculating necessary window size.
3. Make "Templates" page high enough to fit at least 2 rows of template icons.
4. Replaced the old intro image with a combination of 2 images found attached in #2622: background pic that is scaled to fit the page's height, and a cup pic of a fixed size.

I was not exactly sure how to position the cup on the background, so just placed it how seemed okay. But this is trivial to modify.

Example of looks:
new-newgamewizard

Here's how it looks for a smaller wizard:
new-newgamewizard-small

TODO:
~1. I do not like the fact that the image is duplicated as a resource encoded in base64. I'd rather have it embedded as a png and loaded, since we have it in repo anyway.~
2. Add new pic author to credits.


ivan-mogilko:
> I think this issue with displays may be related to how we store the position of either windows or layouts and use it to reload the editor, without checking for actual desktop boundaries. I think this is related to https://github.com/adventuregamestudio/ags/commit/e9a9e0b3e21cfe30e5d769a4ff52843a999f14c9 .

"WindowsConfig" does not save main editor window. It has only few handpicked dialogs, and I thought that the positions should be relative to the main form (although that has to be double checked).

PR #2675, opened 2025-02-05
When mouse is locked in window, confine its position to the game viewport

### LINKED ISSUE #2672: Dead zone in mouse area when mouse locked
**Describe the bug**
If a game does not fit the game area, I can end up moving the cursor beyond, resulting in having to travel back far enough before the game recognize that I'm moving in the other direction.

This is what I expect when the cursor is not locked, but it should not happen when the mouse is locked to the window, or on exclusive fullscreen.

**AGS Version**
AGS 3 and 4

**To Reproduce**
1. Start a game with black bars around (or increase the window size manually if windowed)
2. Make sure the mouse is locked to the window (alt gr+ctrl toggle)
3. Move beyond game area
4. Have to travel all the way back before the game cursor recognizes I'm moving in another direction

**Expected behavior**
When the mouse is locked to the window, it should be locked to the game area, allowing to immediately move back no matter how far the mouse moved from the game area, otherwise it may feel like the cursor got stuck.
For all other cases, the current behavior is what expected.

 - OS: Windows 10

ivan-mogilko:
Fix #2672

### DIFF
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -80,7 +80,7 @@ extern int displayed_room;
 extern CharacterInfo*playerchar;
 extern int eip_guinum;
 extern int cur_mode,cur_cursor;
-extern int hotx,hoty;
+extern int mouse_hotx, mouse_hoty;
 extern int bg_just_changed;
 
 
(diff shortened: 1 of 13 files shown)


AlanDrake:
Works for windowed mode, but not for fullscreen

AlanDrake:
Looks good to me. All works as expected.

Issue #2683, opened 2025-02-18
4.0: Change range of Random()


messengerbag:
**Describe the problem**
In AGS, `Random(N)` returns an int between 0 and N, inclusive. So `Random(6)`, for example, has 7 possible return values.

This is both counter-intuitive and non-standard behavior. Compare:

- Java: `Random.nextInt(int bound)`
- C#: `Random.Next(int bound)`

(Things are different for functions that allow you to specify both the min and max values; in those APIs it is common for both endpoint values to be inclusive.)

It is also unexpected to users. See recent discussion on Discord:

> **Kastchey** — 14/02/2025, 21:45
> Yes. Random(15) means 0 to 14.
> […]
> 
> **Snarky** — 15/02/2025, 00:00
> In AGS, Random(15) means 0 to 15. (I've long argued that this is a mistake in the API design.)
> 
> **Scavenger** — 15/02/2025, 00:10
> It WHAT
> […]
> 
> **Scavenger** — 15/02/2025, 00:53
> And here I was programming all my games assuming it was 0 to X-1

**Suggested change**
It is dangerous to simply change the behavior of the existing API, so my suggestion is to create a new function, `Maths.Random(int bound)` that will exclude `bound` from the range. Plain `Random()` can then be deprecated. The documentation should warn users about the change in behavior.

AGS 4.0 seems like the only good opportunity to fix this mistake.

ivan-mogilko:
Opened PR: #2684

PR #2684, opened 2025-02-18
AGS 4: Script API: add eRound options and new Maths.Random()

### LINKED ISSUE #1983: Script API: add a equivalent of "truncate" to FloatToInt modes
FloatToInt currently supports 3 modes:
- eRoundDown - this is eq to [C `floor`](https://en.cppreference.com/w/c/numeric/math/floor) (or C++ std::floor): it finds the closest integer literally <= than the given float.
- eRoundUp - this is eq to [C `ceil`](https://en.cppreference.com/w/c/numeric/math/ceil): it finds the closest integer >= given float.
- eRoundNearest - this is eq to [C `round`](https://en.cppreference.com/w/c/numeric/math/round): it rounds to the closest integer.

What is missing here is ["truncate"](https://en.cppreference.com/w/c/numeric/math/trunc): the one that finds a closest integer which is *less by magnitude*. We may call this eRoundTruncate (unless someone suggests a better name).

The difference between `floor` and `trunc` is that for the floor (eRoundDown):
+1.5 = +1
-1.5 = -2
and for trunc (eRoundTruncate):
+1.5 = +1
-1.5 = -1

~NOTE: also, the FloatToInt implementation in the engine should be rewritten to use libc or std C++ functions.~ done

EDIT: In fact, we might also add a mode opposite to "truncate", which finds nearest integer greater by magnitude. The C/C++ do not have the ready function for this, but it may be implemented like
```
return x < 0 ? floor(x) : ceil(x);
```

ivan-mogilko:
Resolve #1983, #2683

Contents:
1. Expands eRound option with 2 new modes:
    * `eRoundTowardsZero` - rounds towards a closest integer *less by magnitude* (absolute value);
    * `eRoundAwayFromZero` - rounds towards a closest integer *larger by magnitude* (absolute value);

2. Add `float Maths.Round(float, RoundDirection)` - which rounds a float value to another float value (without making it int).

3. Add `Maths.Random(limit)`, a proper variant of `Random`, which treats limit as an upper exclusive boundary. So `Maths.Random(10)` would choose a number between 0 and 9, but never 10.

4. Add `Maths.RandomFloat()` which produces a float in range [0.0, 1.0) (excluding 1.0).

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -140,7 +140,11 @@ enum MouseButton {
 enum RoundDirection {
   eRoundDown = 0,
   eRoundNearest = 1,
-  eRoundUp = 2
+  eRoundUp = 2,
+#ifdef SCRIPT_API_v400
+  eRoundTowardsZero = 3,
+  eRoundAwayFromZero = 4
+#endif // SCRIPT_API_v400
 };
 
 enum RepeatStyle {
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2022,6 +2022,9 @@ builtin struct Maths {
   import static float Tanh(float radians);
   /// Gets the value of PI
   readonly import static attribute float Pi;
+#ifdef SCRIPT_API_v400
+  import static float Round(float value, RoundDirection);
+#endif
 };
 
 builtin managed struct DateTime {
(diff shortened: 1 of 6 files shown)


AlanDrake:
I've never seen a round direction in random algorithms, but I can see it could come in handy.

The only weakness is relying on `rand()`, we should perhaps implement a xoshiro/xoroshiro/lehmer/PCG algorithm, which are fast and have better statistical distributions (could be another PR).

messengerbag:
Thanks, that's grand, @ivan-mogilko!

Only thing is `Maths.RandomFloat()`, which – just as with `Maths.Random()` – I think should exclude the upper limit, i.e. [0.0, 1.0). This would be consistent with `Math.random()` in Java and Javascript, C#'s `Random.NextDouble()`, Python's `random.random()`, etc.

ivan-mogilko:
> Only thing is Maths.RandomFloat(), which – just as with Maths.Random() – I think should exclude the upper limit, i.e. [0.0, 1.0).

I was not sure if it should or not, but forgot to check existing examples.

PR #2693, opened 2025-02-24
New compiler: `as VARTYPE` clause in expressions

fernewelten:
Provide syntax to dynamically cast a pointer to a managed type to a pointer to another managed type. 

The compiler will generate bytecode to check at runtime whether the object actually __has__ the new type. (opcode 76). The engine must be able to interpret that bytecode, of course; work which is provided in #2665 .

~~Currently, the compiler assumes that opcode 76 expects a managed pointer in MAR. If this were changed to AX then the compiler could generate more compact code.~~

Typical code: 

```
managed struct Me
{
    int Me;
};

managed struct Child extends Me
{
    int Child;
};

managed struct Grandchild extends Child
{
    int Grandchild;
};

int game_start()
{
    Me m = new Grandchild as Child;
    Grandchild *g = m as Grandchild *;
}
```

As concerns the third line from the bottom: 
- `new Grandchild` yields a `Grandchild *`, i.e., a dynpointer. It wouldn't make sense to convert a pointer to a non-pointer, so the compiler implies a `*` in the clause `as Child`. 
- `m` is a variable that accepts `Me *` where `Me` is an ancester of `Child`, so the compiler can prove statically that the expression can go into `m`. It does __not__ generate a dynamic check.

As concerns the second line from the bottom:
- `Me` is an ancester of `Grandchild`, so the compiler can't know at compile time whether `m`, which is a `Me *`, can be converted to a `Grandchild *`. So it __does__ generate a dynamic check (opcode 76). If that check fails at runtime, then `g` will receive a `null`.


Compiler/script2/cc_internallist.h
```diff
@@ -40,8 +40,8 @@ class SectionList
     {}
     SectionList(const std::vector<std::string> &&sections,
                 const std::map<size_t, size_t> &&off2sec)
-        : _sections(std::move(sections))
-        , _off2sec(std::move(off2sec))
+        : _sections(sections)
+        , _off2sec(off2sec)
```
ivan-mogilko, line 44: The way this is changed now defeats the purpose of this constructor, it was intended to "move" the passed objects, as opposed to making it copy.

The warning was probably because the constructor arguments are declared with "const", which is incorrect.
Should be:
```
SectionList(std::vector<std::string> &&sections,
                std::map<size_t, size_t> &&off2sec)
```

ivan-mogilko:
> Currently, the compiler assumes that opcode 76 expects a managed pointer in MAR. If this were changed to AX then the compiler could generate more compact code.

Should I change the engine code in #2665 to use AX instead?

Compiler/script2/cc_internallist.h
fernewelten: OK, I'll fix it.

Compiler/script2/cs_parser.cpp
```diff
@@ -2557,6 +2562,89 @@ void AGS::Parser::ParseExpression_Ternary(size_t const tern_idx, SrcList &expres
...
+            _sym.GetName(new_vartype).c_str());
+
+    if (FlagIsSet(_options, SCOPT_RTTIOPS))
+    {
+        EvaluationResultToAx(eres);
+        PushReg(SREG_MAR);                                      // TODO: remove when SCMD_DYNAMICCAST processes AX
+        WriteCmd(SCMD_REGTOREG, SREG_AX, SREG_MAR);   // TODO: remove when SCMD_DYNAMICCAST processes AX
+        CodeCell const scmd_dynamiccast = 76;               // TODO: remove when SCMD_DYNAMICCAST is defined
+        WriteCmd(scmd_dynamiccast, element_vartype); // TODO: change to SCMD_DYNAMICCAST when that is defined
+        WriteCmd(SCMD_REGTOREG, SREG_AX, SREG_MAR);   // TODO: remove when SCMD_DYNAMICCAST processes AX
```
ivan-mogilko, line 2634: I suspect there's a mistake here, it should be reverse
```WriteCmd(SCMD_REGTOREG, SREG_MAR, SREG_AX);```
otherwise the SREG_AX never changes, and it looks like any cast succeeds unconditionally.

**Comment by ivan-mogilko** [comment]
I gave this a quick test, and this seem to work, if merged with #2665.
I only had to fix few things, one mistake in the generated compiler code (commented a commit above), and another change was to the interpreter, where it should treat register's value as a ready pointer rather than a handle.

Please tell, how should I modify the engine to use AX instead? is the logic same, just using AX instead of MAR, or there should be more arguments?

**Comment by fernewelten** [comment]
> Please tell, how should this instruction be modified using AX instead? is the logic same, just using AX instead of MAR, or there should be more arguments? I can do necessary changes to #2665 right away, then merge your PR first, and rebase 2665 after.

Let's go with SCMD_DYNAMICCAST works on the AX register instead of the MAR register, other logic unchanged.

**Comment by fernewelten** [comment]
Implemented the changes: Assume that SCMD_DYNAMICCAST works on AX; shut up the compiler warnings about std::move by dropping the `const` keywords instead of dropping the `std::move`s

(Had to force-push)

**Comment by ivan-mogilko** [comment]
Tested again, and the cast itself is working well (so long if I modify the engine correspondingly), which is great.

I found that there's a problem with chaining calls though. Consider following script:
```
GUIControl control = gGui1.Controls[1];
String text = (control as Label).Text;
```
I expect this to work, but compiler cannot handle this expression and throws error:
> Expected an operator, found '.' instead

(is not '.' an operator?)

**Comment by fernewelten** [comment]
> (is not '.' an operator?)

No, unfortunately it isn't. This is a left-over from the old compiler.

The compiler treats whole concatenations of literal, identifier, `.` and `[…]` as atomic building blocks, or "extended variable names", as it were, and parses these building blocks separately. Only complete building blocks can be part of an expression and joined with operators such as `+`. The underlying, unspoken assumption seems to be that only __numeric__ expressions can be expressions. (The old compiler has a similar approach and moreover an upper limit on how long the building blocks can be.)

This is not the usual way of handling `.` and  `[…]` AFAIK. So I've thought several times of changing the compiler and bring it in line with what is more common. I've even tried out some tentative code for this in GIT branches.  But then I thought, well the code works even if it is quirky, so let's concentrate on other issues.

This might be the occasion to change this at long last. But l'd suggest making a separate PR for it.

**Comment by ivan-mogilko** [comment]
Alright then.

I've been testing for syntax problems further, and found something else:
using a non-declared type in `as` expression leads to unexpected error, in the sense that the error message is not reporting an undefined type. For example:
```ags
Label label = lblGameName as UNDEFINED_TYPE;
```
Compiler reports:
> Expected ',' or ';', found 'as' instead

I would expect it to say that `UNDEFINED_TYPE` is not a declared type.

PR #2709, opened 2025-03-25
OGL: Fixed blendmodes not respecting transparency

AlanDrake:
Solved the OpenGL renderer not respecting alpha transparency on several blend modes.

It was broken since SDL2 was merged, because the hacks that darken/lighten the textures don't work along the _transparencyShader shader. 

Unlike D3D, there's always a shader enabled for rendering sprites, so I rewrote those texture ops into two new shaders to match D3D output (which is not perfect, but at least it's consistent again).

### DIFF
--- a/Engine/gfx/ali3dogl.cpp
+++ b/Engine/gfx/ali3dogl.cpp
@@ -403,6 +403,8 @@ void OGLGraphicsDriver::TestRenderToTexture()
 bool CreateTransparencyShader(ShaderProgram &prg);
 bool CreateTintShader(ShaderProgram &prg);
 bool CreateLightShader(ShaderProgram &prg);
+bool CreateDarkenByAlphaShader(ShaderProgram& prg);
+bool CreateLightenByAlphaShader(ShaderProgram& prg);
 bool CreateShaderProgram(ShaderProgram &prg, const char *name, const char *vertex_shader_src, const char *fragment_shader_src);
 void DeleteShaderProgram(ShaderProgram &prg);
 void OutputShaderError(GLuint obj_id, const String &obj_name, bool is_shader);
@@ -422,6 +424,8 @@ bool OGLGraphicsDriver::CreateShaders()
   shaders_created &= CreateTransparencyShader(_transparencyShader);
   shaders_created &= CreateTintShader(_tintShader);
   shaders_created &= CreateLightShader(_lightShader);
+  shaders_created &= CreateDarkenByAlphaShader(_darkenbyAlphaShader);
+  shaders_created &= CreateLightenByAlphaShader(_lightenByAlphaShader);
   return shaders_created;
 }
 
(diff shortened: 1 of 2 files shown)


AlanDrake:
Quick test project (slider adjusts transparency):
[BlendTest_ags4.zip](https://github.com/user-attachments/files/19454821/BlendTest_ags4.zip)

ivan-mogilko:
To be honest, I do not like this situation at all. Unfortunately, I overlooked blend modes in the past, and never knew how they used to work (better or worse than now).

First of all, the attached game is not ideal for testing this problem, because it uses random pictures with too many colors. It would be better IMO to have an image with a few stripes in basic colors over a simply colored background. That way one at least could guess what happens when seeing resulting colors, and which effect to expect, knowing formula used for that blend mode. It would also be possible to make screenshots and check precise color values in a graphic editor.

Second, I ran the test game and made screenshots at transparency 0 and 50.
At transparency 0 (alpha 1.0) I can see significant difference between Software implementation and Direct3D/OpenGL implementation in couple of blend modes: I think they are Burn and Dodge.
At transparency 50 (alpha 0.5) I can see more differences between Software and D3D/OGL, and then there are also differences between D3D and OGL alone.

This is worrisome, as that means that we may have principally wrong implementations of certain modes.
Another problem in *my* case is that I personally don't know which variant can be used as an anchor in this comparison. Would that be software mode, since it uses explicit math in our code?

Screenshots attached as an archive:
[blends-test-pr2709.zip](https://github.com/user-attachments/files/19483820/blends-test-pr2709.zip)


---

Looking at the PR itself, I do not like the math in second shader ("blend hack 2: lighten  to simulate transparency"), where you add alpha value to a color, because imo that does not make sense mathematically to sum up color and alpha.

In summary, this PR does not make different graphic drivers match each other, all of them have different results. Meaning, it does not really fix the problem.

Because I did not follow the development of blend modes, I miss some parts here. But I'd rather step back, and do following:
1. First of all, establish the correct result for each blend mode. Ensure that at least software renderer has it, since it's easiest to code blend math in the engine.
2. Check which blend modes don't work in D3D9/OpenGL, and why (are there mathematical reasons, or something in the settings).
3. Find a method that would match math required for blend modes.

ivan-mogilko:
EDIT: Alright, I tested the current ags4 branch, comparing OpenGL results, and I can see what this PR is trying to fix now. Many blend modes did not follow a change of a sprite's transparency.

It's quite possible that these blend modes would require a premultiplied alpha to do the blend operation correctly. This is what your "darken" shader hack does, where it multiplies color by alpha. Unfortunately, the second "lighten" shader hack still does not make much sense to me with summing alpha and color. I believe there must be other, better ways.

In any case, this is such a mess overall, I think we may merge this as a temporary measure, but then all this has to be re-investigated and resolved properly.

AlanDrake:
Aligned the lightnening hack using the `Arg1 + Arg2 - Arg1*Arg2` which is the D3DTOP_ADDSMOOTH op from D3D, so the results at 128 alpha should be matching now.

But yes, I agree with what you say. These hacks were and still are meant to be temporary until we figure out how to apply the correct formula directly.

ivan-mogilko:
Please, give these shaders descriptive names instead of "Hack1/2Shader". Perhaps, something like "LightenByAlpha/DarkenByAlpha", or anything similar that will explain what they are doing.

AlanDrake:
That might confuse people in thinking these are legitimate shaders and not ugly hacks to replicate the original glTexEnvi hack... but if you insist I will.

PR #2711, opened 2025-03-27
AGS 4: Replace Alfont with SDL_ttf

### LINKED ISSUE #1528: Switch to a supported version of FreeType
We use a very old (20years old) version of FreeType. It's very hard to work with it, it doesn't do well with newer platforms, it's not reasonable to support fonts for languages with ligatures (Persian, Hebrew, ...) and it appears to use some hacks on our end.

If we switched to a newer version of FreeType, we could also switch to use something like SDL_ttf instead of Alfont.

For newer games, it doesn't seem that anything blocks it, so we could just switch in AGS4 and call it a day. If we want to have backwards compatibility, then things get a little harder.

Also, we could have two versions of FreeType, now I get it that FreeType is written in C, and as such it doesn't support namespaces, but it should not be that hard to use a refactoring tool to add some prefix (ex: `oldft_`) in all symbols that are needed and just go on, after all the source is shoved in this repo, so it should be something we are ok doing.

Though if we do manage to fix so that AGS renders correctly using a newer FreeType other platforms may benefit - last time I checked ScummVm used a newer version of FreeType and also some Linux distros have their own builds that simply use whatever is the latest FreeType, people playing on these platforms (although not our builds/recommended way of playing) will face the font rendering woes that comes with using our Alfont with a newer version of FreeType.

**References**

- https://github.com/ericoporto/ags-old/commits/ags3_sdl_ttf_utf8
- https://github.com/ericoporto/ags-old/commits/upgrade-freetype

ivan-mogilko:
Drafting a change required to resolve #1528.

What is done:

1. Replaced FreeType 2.1.3 with contemporary FreeType 2.13.2. Moves library sources to libsrc dir in the project root.
2. Replaced Alfont library with SDL_ttf 2.24.0. SDL_ttf is embedded as source into the engine, this is done for two reasons:
    * It seems small enough to place in our repository;
    * We may require modifications, that are not necessarily getting applied upstream.
3. As a consequence AGS.Native and thus Editor now also requires SDL2.dll.
4. Done couple of additions to SDL_ttf source code, namely had to add:
    * TTF_SetFontCharSpacing
    * TTF_GetFontCharSpacing
    * TTF_GetFontBBox
    * TTF_GetFirstFontChar
    * TTF_GetNextFontChar
All of the above have been proposed as addition to SDL_ttf. Some were merged into SDL3 based branch, others were not added still at the time this PR is made.
5. Deprecated Font's "TTF font adjustment" property. This property was meant for compatibility with old version of Alfont (and/or old version of FreeType) and is no longer needed. As a result some previously used AGS-compatible fonts may be positioned differently. But there are still means to fix these using custom font height definition and vertical offset property.

What will not be part of this PR, but should be planned for the future:

1. Harfbuzz and proper rendering of specific languages (Arabic, Persian etc). I think it's best leave this for the next update, after this transition is proven good.
2. Optimize text blitting. SDL_ttf has a peculiarity: it does not draw on a ready surface, but creates a new one containing rendered text, which then should be drawn on destination. AFAIK this was done to let use specially allocated buffers which enable use of faster instructions (SSE?). But I cannot tell which way would be faster in our case, considering we must always blit the received SDL_Surface to another bitmap before creating a texture from that. Need to experiment with this, and decide if we may want to modify SDL_ttf to let it render text onto our bitmap, or as a inverse solution, perhaps return a Bitmap object wrapping SDL_Surface to be used further in game.
3. Wrapped text optimization. First of all, given the above problem, if we keep current implementation, then for a multiline text there will be a new SDL_Surface allocation per line. Idk how much of a performance issue that will be in practice, as text is not redrawn each frame, but only once when initializing a textual overlay or a gui control (or a new text is assigned). Another consideration is rendering multiline text in particular languages, the ones that might require their own word-wrapping rules. Or rendering text containing unicode control characters (text direction, etc). In the end we might have to let SDL_ttf handle wrapping anyway. But using SDL_ttf's own text wrapping will require to change few bits in the engine, as currently text wrapping is done prior to calling Font Renderer interface. Likely, a IAGSFontRenderer interface will have to be expanded with RenderTextWrapped method, and same done in plugin API.
4. I noticed that some of the commonly used AGS fonts have "jagged" letters, like pixels either missing or extra pixels present. Don't know yet what's causing this. I have vague memories experimenting with contemporary version of the FreeType before, and recall that this problem can be fixed by either enabling or disabling a "font hinting" for particular fonts. If so, we might need to add more properties to the font that would allow users to fine-tune their rendering.


ivan-mogilko:
@ericoporto if you have some spare time, could you take a look at the CMake scripts in this pr?
There are following things that concern me.
1. Currently I left original CMakeLists.txt in freetype, but i am not sure if that's a right thing to do, because it may be not building what i think it should (i.e. maybe it builds a separate static library which is not linked to the engine). Should we replace that with our own script, similar to what we have in other lib sources?
2. Afaik, freetype must have a FT2_BUILD_LIBRARY compiler flag enabled.
3. As SDL_ttf is going to be a part of the Common lib, it needs to include SDL headers. How do I add such instruction in CMake scripts?

I will read about CMake scripts and figure this out too.

ivan-mogilko:
It seems like I figured out most things, using existing scripts as an example (although few things still are strange to me).
Haven't actually tested this on anything but Windows and Linux yet.

Following immediate issues remain with building:
1. Emscripten fails to compile FreeType, because it does not know "open"/"read" etc posix specific functions in unix/ftsystem.c file. These are declared in "fnctl.h". FreeType has a condition under which it includes "fcntl.h", but it does work here. Need to find out whether Emscripten was supposed to use this file at all, or a generic ftsystem.c instead.
2. Game auto-tests are failing. Haven't checked these yet, maybe this is something wrong in ags4 branch.

ericoporto:
This looks alright, but I can look into this next week when I am back home, I have a variant of this in one of my older branches where Freetype and SDL_ttf are built as static libraries (in VS it gets added to VS dependencies built on the server), if it’s desirable. The advantage of having Freetype built alongside the engine instead is to take advantage of miniz + libpng to get colorful emojis, but perhaps this is not necessary. And also having the sources makes it much easier to build with different build systems. I can’t tell the issues with the things that aren’t building.

ericoporto:
This will take me a little while but my idea is to put SDL_ttf like we have SDL_sound as a dependency, and then put in the docker images (an attempt here https://github.com/ericoporto/ags/commit/364655b7a4112c068303c8e136bdc4097d874663) and add to `sdl2.props` too. The pure CMake route will be using find ot [fetch it in CMake](https://github.com/ericoporto/ags-old/commit/938befea79b16b332761a9af722cd7209a27d23d), and I will need to [fetch dependencies too](https://github.com/ericoporto/ags-old/commit/7295d921562a2284aee0dbd64760c520eb3631b7) if SDL_ttf is being fetch due to it using dynamic linking to its dependencies.

ericoporto:
I think you got it right. Overall it’s easier to have the sources in the directory directly so updating or changing something can be done more easily - in parts. But having the sources not there and have the dependencies outside is less code we are committing in our main repository.

My feeling is we are never going to edit Freetype's code directly because we haven’t done so in a long time. So I think this could be done in a sort of middle ground where the current old Freetype is removed, and we get the new one as an external dependency and build SDL_ttf from code directly in the repo. The “get Freetype as an external dependency” could be done as we have for other dependencies right now or alternatively as a submodule.

ericoporto:
Hey, I think if you add `add_compile_options(-Wno-implicit-function-declaration)` inside the if in CMakeLists.txt for the Emscripten stuff, it will build, because it seems that these errors are from new warnings that are added in a new CLang version that FreeType hasn't got around to addressing yet.

ericoporto:
I think your fork of SDL_ttf should be under AGS organization to make it easy to sync changes from upstream. This is my learning from syncing changes manually from mojoAL.

ericoporto:
I tried to throw away the Freetype CMakeLists.txt because it was too complex and replaced with our own... It seems things are building. https://github.com/ericoporto/ags/commit/cd19e67dc61ff716ecfdbd5fd85f202f02c0a817 , I also had to rebase because I was getting a weird error at link time in my computer.

ericoporto:
I changed my mind on fetching things vs having in the repository. Having in the repository is easier to adjust for different build systems, specifically VS and Xcode (for iOS). Plus If there's something that isn't building due to minor issues (like for Emscripten), it's easier. I think if we supported only regular desktop systems and nothing else I would trust upstream, but as soon as iOS, Android and Emscripten join in, things may break. I have been playing also with Windows Arm64, which is another thing that upstream libraries usually rarely test. Also we want SDL_ttf on the Editor too, and this way it's easier to build.

About the PR I believe it should be done now and we can move on with it as is and then fix anything else that comes up - I think like minor stuff like license mentions.

ericoporto:
@ivan-mogilko  this needs to be rebased and fixed again before merging. Other than this, it looks done, any reason for being a draft still?

ivan-mogilko:
There are two things bothering me:
1. Some common ags fonts look "garbled", maybe caused by a different freetype lib. I recall seeing in the past that enabling or disabling "font hinting" may fix this, but did not try here yet.
2. Text measurement and wrapping done in the engine may become sub-performant with SDL_ttf, because it's caching a number of things itself. We might need to let SDL_ttf do the text wrapping. I suspect that this may be required if we use harfbuzz too, so maybe can be postponed.

ericoporto:
Hey, I think this was one of the last things that we wanted to move in ags4 before moving from alpha to a beta from looking into all ags4 discussions.

ivan-mogilko:
Yes, this should be done.

ivan-mogilko:
Started updating this PR. I noticed that the last released version of SDL_ttf based on SDL2 is still 2.24.0. There are few fixes made after it, but no new release made yet, so I decided to keep at 2.24.0 for now.

There's a number of new font operations added in 3.x branch since the last attempt, so I still will need to update these to use SDL_ttf or FreeType now. Also, I added GetBBox and character spacing to SDL_ttf, but iirc these were merged into SDL3 branch, and never backported to SDL2 branch. So likely I will have to keep patching our local version here.

ivan-mogilko:
Something that annoys me in this draft, I put SDL_Surface ptr inside our Bitmap, but not because it's used in any way directly, but rather because it wraps its internal pixel data, and this was the easiest way to have its ownership and free the surface when the Bitmap is deleted. At the same time it's annoying, because it adds this separate surface field alongside allegro's BITMAP, and also brings SDL_Surface dependency in bitmap class without a very good reason, as it's not used in any other way at the moment. I thought about either replacing this with a unique_ptr which stores unknown object type and a custom deleter, which can be passed as a parameter, or having a different container that pairs Bitmap with SDL_Surface for the particular scenario.

ericoporto:
I think it may be possible to have the bitmap in the SDL_ttf instead. When I did this implementation, I remember I had SDL surface wrap the bitmap or the other way around, meaning there was no copy, it used the same memory/pixels layout, including stride and other things. Other approach may be to have the interface be a pixel array or some other interface to separate this dependency - I think we only have 32-bit and 8-bit needs and all other possibilities supported by SDL surface could be ignored?

ivan-mogilko:
> I think it may be possible to have the bitmap in the SDL_ttf instead. When I did this implementation, I remember I had SDL surface wrap the bitmap or the other way around, meaning there was no copy

SDL_ttf draw text functions create and return their own surfaces, there was no way to pass a precreated surface there last time I checked.

ivan-mogilko:
Ok, back to this, I replaced SDL_Surface pointer in Bitmap class with "abstract" pointer with a custom deleter. Now Bitmap does not depend on SDL, and also is capable of containing any external "image" object, if necessary.

ivan-mogilko:
I think this PR is nearly done. There are still issues that should be solved, but not as a part of this PR (I updated their list in the PR's description).

There's one problem with Editor unit tests, where it fails to load AGS.Native necessary for one test. AGS.Native requires SDL2.dll after these changes. I don't know how to fix this yet.

ivan-mogilko:
@ericoporto because of these changes, and because Editor still draws fonts in native code, the Editor now requires SDL2.dll to run. How will this affect CI that requires to run the Editor (compiling the game, creating the template)?

Editor's unit tests also need SDL2.dll copied somewhere, I suspect they require it in their intermediate directory where they copy all of their dependencies (`Editor\AGS.Editor.Tests\bin\Release`?). How may this be achieved in GHA?

ericoporto:
Haven’t looked at this yet, but could this be a post event in VS that copies the dll itself if it isn’t in the Release/Debug dir from VS? Then it’s just a project thing and nothing really changes in the CI.

I think we did this for something (the image magick dependency perhaps?)

ivan-mogilko:
I think the main question is, where to copy from? Meaning, SDL2.dll has to be prepared somewhere first.

ericoporto:
It is in `C:\Lib\SDL2\lib\x86` or `C:\Lib\SDL2\lib\x64` no? Like not directly but through the props. We have `AGS_SDL_LIB` there, in sdl2.props.

We also have some instruction in the csproj, for Native, I wonder if we can use something similar in vcxproj, using a Copy and CopyToOutputDirectory tags.

image

I can make a PR for this for the Engine.App.vcxproj file and then we can later transport this approach to here if you want?

Edit: made an idea in https://github.com/adventuregamestudio/ags/pull/3043

ericoporto:
ok, now I have no idea why it doesn’t work. Erh, wonder if the AGS Native vcxproj actually builds somewhere else so it needs an additional copy step in the Editor csproj file? The copy step in my screenshot above suggests this is only needed when building from a UNIX environment (I think that was authored by @persn?).

persn:
> I think that was authored by @persn?

I don't recognize it. Wasn't it introduced here? #2197

ivan-mogilko:
For some reason, applying the same code into Native.dll.vcxproj does not make SDL2.dll item appear under AGS.Native project in MSVS, and neither SDL2.dll gets copied to the AGS.Native's output folder.

(I've been testing this locally)

ericoporto:
How does irrklang work? Did we had a copy step for it somewhere? (Edit: we manually copy the irrkmp3 and other dependencies, using a copy step in the csproj!) Perhaps csproj works differently somehow… Is AGS.Native output dir the Reference directory ?

Edit: uhm, I copied that from somewhere then, sorry @persn 😬

ivan-mogilko:
Before anything else, for some reason the "SDL2.dll" item does not appear in AGS.Native project inside MSVS. It does appear for the Engine project, but not the AGS.Native. What could be a reason for this? Maybe i applied this change incorrectly there?

ericoporto:
Found the issue, in your branch, there is an extra sdl2.props for some reason, here: https://github.com/ivan-mogilko/ags-refactoring/blob/ags4--sdlttf/Editor/AGS.Native/sdl2.props, I guess if needs to exist you would need to also add the dll modification from my commit there?

Edit: ok, copying the sdl2.props modification will make the SDL2.dll appear in `ags\Editor\AGS.Native\Debug` (or `Release`). I guess it does also need an extra copy step.

Edit2: ok, proper fix is the commit below

https://github.com/ericoporto/ags/commit/2e373325fe83c59f21c3f0475edb782a17a8ab28

The github actions for the Editor tests in my branch appear to complete fine too!

This extra sdl2.props will probably be removed in #2796, once SDL_Sound is used in the Editor, so the dependencies will be the same.

> I noticed that some of the commonly used AGS fonts have "jagged" letters, like pixels either missing or extra pixels present. Don't know yet what's causing this. I have vague memories experimenting with contemporary version of the FreeType before, and recall that this problem can be fixed by either enabling or disabling a "font hinting" for particular fonts. If so, we might need to add more properties to the font that would allow users to fine-tune their rendering.

I have no idea what it means or how to do, but ScummVM solved by including the "auto hinting" from our version of FreeType by cutting out just it from it. I think I left a specific link for this in the long conversation this issue points to. I guess these extra things you mention may either go to a single issue of things to do or go to individual issues for working out each of these things.

ivan-mogilko:
Alright, this PR seems ready then.

ericoporto:
Yeahp, forgot to mention, I tested it a little bit and it seemed to work, actually, I couldn’t visually find any difference from before.

PR #2712, opened 2025-03-30
Editor: let select palette index for the sprite's transparent color

### LINKED ISSUE #805: Editor: Import Sprite: let select transparent palette index
Some graphic adventure games use color index 255 as the background colour, a real example is "Drascula: The Vampire Strikes Back". You can see one of the sprite sheets i extracted from the game files:
sprite sheet

All the sprite sheets from the game must be remapped to have the background colour at index 0 because in AGS you cannot set it at colour index 255.

This is the original "import sprite" window:
original

And this is what i propose, to let the user select the colour index (by default it can be set at 0 to preserve compatibility):
new

ivan-mogilko:
Long overdue, resolves #805

In the sprite import dialog replaces "Palette index 0" with "Palette index" selection and a numeric input field (default to value 0). This allows user to select exact palette index to use as a transparent color, in case the sprite source has it on other index than 0.
"Palette index 0" is still available in the sprite's properties for backwards compatibility.

Images for the test:
8-bit PNG: 
Bar_Door1
8-bit GIF (multiframe):
Bar Door

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -112,7 +112,8 @@ namespace AGS.Editor
          *                  GlobalVariable may be of array type.
          * 3.6.2.2        - Button.WrapText, TextPadding.
          * 3.6.2.6        - Settings.GameFPS.
-         * 3.6.3          - Settings.GUIHandleOnlyLeftMouseButton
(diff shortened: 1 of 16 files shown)


ivan-mogilko:
I had to add another change, that retrieves original palette from GIF image file and uses it as a reference when importing multi-framed images.

This is necessary, because when GIF is imported, each frame is converted to a Bitmap object, which may have a different palette, usually containing only used colors.

PR #2714, opened 2025-04-04
CMake compatibility

ericoporto:
It seems GitHub Actions builder had an update in CMake that broke building in PR #2712. Made a quick fix that worked in the pipeline in my fork.

I will need to joing together the fetch content scripts as now it's expected to do a single MakeAvailable call later that contains all libraries in it - so, I can't separately call to build SDL2 and then SDL Sound anymore, I need to tell cmake to build SDL2 and SDL_Sound in sequence, so it figures the link between then, when fetching them. This will take a little more time to figure it out.

### DIFF
--- a/CMake/FetchGoogleTest.cmake
+++ b/CMake/FetchGoogleTest.cmake
@@ -11,8 +11,8 @@ else()
     FetchContent_Declare(
         googletest_content
         DOWNLOAD_EXTRACT_TIMESTAMP TRUE
-        URL https://github.com/google/googletest/archive/4ec4cd23f486bf70efcc5d2caa40f24368f752e3.tar.gz
-        URL_HASH MD5=b907483a9045a2edda15ee7d2a68aaa5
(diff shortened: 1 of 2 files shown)


ericoporto:
Just noting two things for myself here so I don't forget. Since CMake 3.24, `OVERRIDE_FIND_PACKAGE` has been introduced in [`FetchContent_Declare`](https://cmake.org/cmake/help/latest/module/FetchContent.html), which enables that later calls to `find_package` trigger the `FetchContent_MakeAvailable`, so something like SDL_Sound could trigger it if needed. The other magical (apparently undocumented thing?) is you can pass `CMAKE_ARGS` in the `FetchContent_Declare` call to pass CMake arguments - [example here](https://github.com/joaquinbejar/simple-polygon-io/blob/e6f402719589d13ae1c42df0aba79683e3d2d35a/cmake/curl.cmake#L1-L7). With this, it should be possible to trim things down and redo things to not depend on FetchContent_Populate in the CMake fetches we do.

PR #2715, opened 2025-04-05
[Remake] New compiler: Initializers for global classic arrays and structs

ivan-mogilko:
This is a rebased version of @fernewelten 's #2662 PR, which has a applied change to the initialization syntax, as suggested in the previous PR discussion: when we use "named" initialization for array elements we don't need to use `[N]:` syntax, but `N:` syntax instead.

Example:
```
bool IsPrime[10] = {
    2: true,
    3: true,
    5: true,
    7: true,
};
```

Everything else should stay as stated in #2662 description, so please refer to it.


ivan-mogilko:
Well, nobody seems to be interested in trying this out, so I'll merge this to not let the work be for naught.

Issue #2705, opened 2025-03-22
Support for pixel shaders


alkhimey:
**Describe the problem**

Currently to be able to perform per pixel operations, there are two options:

**Option 1**
1. Get drawing surface of the desired background, object, area etc.
2. Use [`GetPixel`](https://adventuregamestudio.github.io/ags-manual/DrawingSurface.html#drawingsurfacegetpixel)/`SetPixel` on the drawing surface.

**Option 2 (did not try it yet):**
1. Write a plugin.
2. Request event hook for  [`AGSE_PREGUIDRAW `](https://adventuregamestudio.github.io/ags-manual/EnginePluginRun-timeAPI.html#iagsengine-event-agse_preguidraw)
3. Implement `AGS_EngineOnEvent`:
4. [If the renderer](https://adventuregamestudio.github.io/ags-manual/EnginePluginRun-timeAPI.html#iagsenginegetgraphicsdriverid) is Direct X: the `data` argument will be an `IDirect3DDevice9`. Attach HLSL shader to it. I am not sure about this actually, as well as what will happen with the default HLSL shader AGS uses.
5. If the renderer is software, then call `GetVirtualScreen` to get the bitmap and call `GetRawBitmapSurface` to get the surface. Together with the knowlage of the pixel format, pixels can be read and written individually. This is faster than GetPixel/SetPixel from option 1. The Blend plugin does this for all renderers.
6. If the renderer is OpenGL, then the documentation does not tell what is expected in `data` argument.

Both options are impractical. Option 1 is very slow and option 2 is complex and difficult to develop and debug.

**Why is it important**
Efficient per pixel operations can be very useful. It is possible to implement:

1. Accesibility settings such as color blind mode or high contrast mode.
2. Image filters using lookup tables can set atmosphere during special sequences (eg grayscale on flashbacks)
3. Weather and day night cycles.
4. Display sprites with blending modes.
5. Characters or objects appearing/disappearing with transitions (eg [dissolve](https://godotshaders.com/shader/2d-burn-dissolve-from-point-v-1-0/))

There are endless possibilities.

**Suggested change**
AGS already have infrastructure for running shaders. Transparency, tint and lighting are implemented as shader/s in graphically accelerated drivers.

[...]

PR #2716, opened 2025-04-11
Custom pixel shaders support + Shaders script API

ivan-mogilko:
Addresses #2705

### Changes to graphic drivers:

1. Direct3D and OpenGL graphic drivers support creating and registering custom pixel shaders. Currently OpenGL supports creating shaders from GLSL source by compiling them, and Direct3D supports only loading precompiled shader objects created by Microsoft's fxc utility.
2. Shaders are identified by both string name and uint, string is used as a lookup and may be useful for serialization. A uint id is used for a quick reference when rendering.
3. DDB objects in the engine let assign a custom shader id to them.

### Changes to Script API:

Added `ShaderProgram` and `ShaderInstance` structs:
```
builtin managed struct ShaderProgram {
  /// Creates a new ShaderProgram by either loading a precompiled shader, or reading source code and compiling one.
  import static ShaderProgram* CreateFromFile(const string filename); // $AUTOCOMPLETESTATICONLY$
  /// Creates a new shader instance of this shader program.
  import ShaderInstance* CreateInstance();
  /// Gets the default shader instance of this shader program.
  import readonly attribute ShaderInstance* Default;
};

builtin managed struct ShaderInstance {
  /// Sets a shader's constant value as 1 float
  import void SetConstantF(const string name, float value);
  /// Sets a shader's constant value as 2 floats
  import void SetConstantF2(const string name, float x, float y);
  /// Sets a shader's constant value as 3 floats
  import void SetConstantF3(const string name, float x, float y, float z);
  /// Sets a shader's constant value as 4 floats
  import void SetConstantF4(const string name, float x, float y, float z, float w);
  /// Gets this instance's parent Shader
  import readonly attribute ShaderProgram* Shader;
};
```

Here `ShaderProgram` represents a compiled shader itself, and `ShaderInstance` is shader's setup with certain constant values. You may think of ShaderInstance as a kind of a limited "material" type.

Added `ShaderInstance* Shader` attribute to all drawable game objects, Camera, Viewport and Screen (static property).

Basic idea is that you create a ShaderProgram using `CreateFromFile` static method, and then either assign ShaderProgram.Default to objects, or create new ShaderInstances and assign them to objects.
ShaderInstance may be assigned to multiple objects, in which case they all will share same shader setup.

Because not every graphics driver supports shaders, and for general convenience, CreateFromFile always returns a valid ShaderProgram object, but if there was error creating one, then using this "shader" will make renderer fallback to default shader (or no shader in case of software renderer).

ShaderProgram.CreateFromFile supports all locations that File.Open supports (including files packaged into game pak).

ShaderProgram.CreateFromFile accepts following file extensions:
* "glsl" - for OpenGL shader sources
* "hlsl" - for D3D9 shader sources
* "fxo" - for D3D9 compiled shaders

To make things more convenient, user does not have to make a switch in the code: engine does it for them. If you pass "myshader.glsl", and engine is running a Direct3D driver, then it will look for "myshader.fxo" and "myshader.hlsl" instead. You may even skip extension completely.

### Shader def files for Direct3D

Although engine can now compile Direct3D shaders from source code, using D3D_compile library (standard part of DirectX since Windows 7), it cannot know its constants, because Direct3D9 interface lacks the necessary abilities (we'd have to link a deprecated utility library for that, and I doubt that's a good idea).
For that reason, I decided to introduce a separate helper file that may optionally accompany shader file (whether precompiled or source code). The file should be called `"<shadername>.d3ddef"` (extension means "d3d shader definition"). This file is simply an ini format (because we already use ini format for several things), which contains a number of optional entries:

```
[compiler]
target = compilation target ("ps_2_0", and so on)
entry = entry function name (e.g. "main")

[constants]
<constant_name> = register index (a number >= 0)
```

About compiler targets:
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/specifying-compiler-targets#direct3d-91-92-and-93-feature-levels

If "compiler" options are not present, Direct3D will use defaults.
If "constants" are not present, Direct3D will use default hardcoded register values.

### Built-in shader constants

When writing shaders you are allowed to use a number of constants (or uniforms in GLSL) provided by the engine.
In GLSL these must match the name and type, but their order is not important (they may be not present if not used too).
In HLSL, because engine currently only uses precompiled shaders, and is missing a utility library to get shader info, these constants must match the type and the *register* (but name may be any). This may change in future updates of this draft.

Here's a list of constants:
* float iTime - current time in seconds;
* int iGameFrame - current game frame index (NOTE: must be float in HLSL, because apparently it D3D9-compatible mode does not support integer constants? at least that's what I read somewhere);
* sampler2D iTexture - sprite's texture reference;
* vec2 iTextureDim - texture's dimensions in pixels (type `float2` in HLSL);
* float iAlpha - sprite's general alpha (object's transparency);

Other predefined input parameters:
* vec2 vTexCoord - is predefined for GLSL only, gets current texture coordinate for the pixel. HLSL should use TEXCOORD0 input parameter (see shader examples below).

Example of declaring parameters in GLSL:
```
uniform float iTime;
uniform int iGameFrame;
uniform sampler2D iTexture;
uniform vec2 iTextureDim;
uniform float iAlpha;

varying vec2 vTexCoord;
```

Example of declaring parameters in HLSL (names do not have to match if you precompile, but notice the order of registers!):
```
// Pixel shader input structure
struct PS_INPUT
{
    float2 Texture    : TEXCOORD0;
};

// Pixel shader output structure
struct PS_OUTPUT
{
    float4 Color   : COLOR0;
};

sampler2D iTexture; // is in sampler register 0

const float  iTime:         register( c0 );
const float  iGameFrame:    register( c1 );
const float2 iTextureDim:   register( c2 );
const float  iAlpha:        register( c3 );
```

FXC utility:
https://github.com/user-attachments/files/19633366/fxc.zip
(courtesy of @AlanDrake )

Simple demo project:
https://www.dropbox.com/scl/fi/ifgmbhnmevzlv0jw0z1f6/ags4-shadertest.zip?rlkey=yxvrbna340ftyzzeuef7n787m&st=447hyl5j&dl=0
Example shaders:
[colorwave_shader.zip](https://github.com/user-attachments/files/19714581/colorwave_shader.zip)

It took me some time to learn how to convert GLSL to HLSL, some things are trivial, others require to read through docs and online discussions... Still, there are some mistakes, where HLSL and GLSL variants seem to have an opposite Y offset in  calculations. I was not able to understand where the mistake is yet.

### PLANNED TODO

1. **[DONE]** Support compiling HLSL shaders from source.

For that we'd need to link to `d3dcompiler.lib` and `d3dcompiler_47.dll`.
Related documentation: https://learn.microsoft.com/en-us/windows/win32/api/d3dcompiler/nf-d3dcompiler-d3dcompile

2. **[DONE]** Support custom shader parameters.

This problem is split between script API and passing parameters along with the DDBs into graphic driver.
In #2705 it was suggested that shaders should have a number of predefined user parameters. This may be the easiest way to go if we only support precompiled shaders for Direct3D. If we solve p1 and compile shaders at runtime, then this is still a option to consider.

In regards to the script API, I've been thinking about introducing a separate struct, working title is `ShaderInstance` (may be a subject for change). Such struct would hold a reference to `ShaderProgram` and user parameter values, as well as functions for setting them. And then it will be `ShaderInstance` that is assigned to game objects.

3. **[DONE]** Make a good rule for switching between shader assets. E.g. CreateFromFile may require to pass only filename without extension, and engine would add extension depending on the active gfx driver, and look for that file.

4. Add shaders right in the Editor. Make a "Shaders" node in the project tree, where user may add shader items. Editor will generate ShaderProgram variables in script for these, and these will be initialized automatically on game launch.

5. As you may notice, I added Shader attribute to Viewport struct, so you can have a full-room shader. There's a related issue with this: having a shader applied to room viewport requires viewport or camera to be a texture. Currently this breaks "Render sprites at screen resolution" setting. I believe that it's possible to fix this, but would require certain rewrite in gfx drivers code.

6. **[DONE]** Similarly, we could add "fullscreen" shader. And then it will also require that aforementioned rewrite in order to keep "Render sprites at screen resolution" option working.

7. Optionally, we could support shaders with additional textures attached to them. For example a shader that blends certain preset sprite to a game object. I think that may be possible if `ShaderInstance` would let assign sprite numbers as "texture reference" parameters, and these sprites are converted to textures and used within gfx drivers when such shader is run. That's just a vague idea for now.

8. Switch to using a modern GLSL syntax with explicit input and output parameters in the "main" function.

9. **[DONE]** Serialize shader references in game saves.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -161,6 +161,7 @@ target_sources(common
     util/smart_ptr.h
     util/stdio_compat.c
     util/stdio_compat.h
+    util/stl_utils.h
     util/stream.cpp
     util/stream.h
     util/string.cpp
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -110,6 +110,7 @@ enum CharacterSvgVersion
     kCharSvgVersion_400_13  = 4000013, // compat with kCharSvgVersion_36205
     kCharSvgVersion_400_14  = 4000014, // proper ARGB color properties
     kCharSvgVersion_400_16  = 4000016, // motion path exposed to script, separate runtime flags, turns
+    kCharSvgVersion_400_18  = 4000018, // shaders
 };
 
 
--- a/Common/ac/gamesetupstructbase.cpp
+++ b/Common/ac/gamesetupstructbase.cpp
@@ -146,6 +146,8 @@ const char *GetScriptAPIName(ScriptAPIVersion v)
     case kScriptAPI_v400: return "4.0.0-alpha8";
     case kScriptAPI_v400_07: return "4.0.0-alpha12";
     case kScriptAPI_v400_14: return "4.0.0-alpha18";
+    case kScriptAPI_v400_16: return "4.0.0-alpha20";
+    case kScriptAPI_v400_18: return "4.0.0-alpha22";
     default: return "unknown";
     }
 }
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -197,7 +197,8 @@ enum ScriptAPIVersion
     kScriptAPI_v400_07 = 4000007,
     kScriptAPI_v400_14 = 4000014,
     kScriptAPI_v400_16 = 4000016,
-    kScriptAPI_Current = kScriptAPI_v400_16
+    kScriptAPI_v400_18 = 4000018,
+    kScriptAPI_Current = kScriptAPI_v400_18
 };
 
 const char *GetScriptAPIName(ScriptAPIVersion v);
--- a/Common/gui/guidefines.h
+++ b/Common/gui/guidefines.h
@@ -160,6 +160,7 @@ enum GuiSvgVersion
     kGuiSvgVersion_40010    = 4000010, // compat with kGuiSvgVersion_36200-36202
     kGuiSvgVersion_40014    = 4000014, // proper ARGB color properties
     kGuiSvgVersion_40016    = 4000016, // extended graphic properties for controls
+    kGuiSvgVersion_40018    = 4000018, // shaders
 };
 
 // Style of GUI drawing in disabled state
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -792,6 +792,19 @@ void GUIMain::ReadFromSavegame(Common::Stream *in, GuiSvgVersion svg_version, st
             _scale.Y = 1.f;
     }
 
+    if (svg_version >= kGuiSvgVersion_40018)
+    {
+        _shaderID = in->ReadInt32();
+        _shaderHandle = in->ReadInt32();
+        in->ReadInt32(); // reserved
+        in->ReadInt32();
+    }
+    else
+    {
+        _shaderID = -1;
+        _shaderHandle = 0;
+    }
+
     MarkChanged();
     UpdateGraphicSpace();
 }
@@ -824,6 +837,10 @@ void GUIMain::SkipSavestate(Stream *in, GuiSvgVersion svg_version, std::vector<C
     if (svg_version >= kGuiSvgVersion_400)
     {
         in->Seek(15 * sizeof(int32_t));
+    }
+    if (svg_version >= kGuiSvgVersion_40018)
+    {
+        in->Seek(4 * sizeof(int32_t));
     }
 }
 
@@ -875,6 +892,11 @@ void GUIMain::WriteToSavegame(Common::Stream *out) const
     out->WriteInt32(0); // sprite pivot y
     out->WriteInt32(0); // sprite anchor x
     out->WriteInt32(0); // sprite anchor y
+    // kGuiSvgVersion_40018
+    out->WriteInt32(_shaderID);
+    out->WriteInt32(_shaderHandle);
+    out->WriteInt32(0); // reserved
+    out->WriteInt32(0);
 }
 
 
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -132,9 +132,10 @@ void GUIMain::SetBlendMode(BlendMode blend_mode)
     _blendMode = blend_mode;
 }
 
-void GUIMain::SetShader(int shader_id)
+void GUIMain::SetShader(int shader_id, int shader_handle)
 {
     _shaderID = shader_id;
+    _shaderHandle = shader_handle;
 }
 
 void GUIMain::SetScale(float sx, float sy)
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -801,7 +801,7 @@ void GUIMain::ReadFromSavegame(Common::Stream *in, GuiSvgVersion svg_version, st
     }
     else
     {
-        _shaderID = -1;
+        _shaderID = 0;
         _shaderHandle = 0;
     }
 
--- a/Common/gui/guimain.cpp
+++ b/Common/gui/guimain.cpp
@@ -132,6 +132,11 @@ void GUIMain::SetBlendMode(BlendMode blend_mode)
     _blendMode = blend_mode;
 }
 
+void GUIMain::SetShader(int shader_id)
+{
+    _shaderID = shader_id;
+}
+
 void GUIMain::SetScale(float sx, float sy)
 {
     _scale = Pointf(sx, sy);
--- a/Common/gui/guimain.h
+++ b/Common/gui/guimain.h
@@ -141,7 +141,7 @@ public:
     void    SetTransparencyAsPercentage(int percent);
     BlendMode GetBlendMode() const { return _blendMode; }
     void    SetBlendMode(BlendMode blend_mode);
-    int     GetShader() const { return _shaderID; }
+    int     GetShaderID() const { return _shaderID; }
     int     GetShaderHandle() const { return _shaderHandle; }
     void    SetShader(int shader_id, int shader_handle);
     Pointf  GetScale() const { return _scale; }
@@ -273,7 +273,7 @@ private:
     int     _popupAtMouseY = -1; // popup when mousey < this
     int     _transparency = 0;  // "incorrect" alpha (in legacy 255-range units)
     BlendMode _blendMode = kBlend_Normal; // render blend mode
-    int     _shaderID = -1;
+    int     _shaderID = 0;
     int     _shaderHandle = 0; // // runtime script shader handle
     Pointf  _scale = Pointf(1.f, 1.f);; // x,y scale
     float   _rotation = 0.f;    // rotation, in degrees
--- a/Common/gui/guimain.h
+++ b/Common/gui/guimain.h
@@ -141,6 +141,8 @@ public:
     void    SetTransparencyAsPercentage(int percent);
     BlendMode GetBlendMode() const { return _blendMode; }
     void    SetBlendMode(BlendMode blend_mode);
+    int     GetShader() const { return _shaderID; }
+    void    SetShader(int shader_id);
     Pointf  GetScale() const { return _scale; }
     void    SetScale(const Pointf &scale) { SetScale(scale.X, scale.Y); }
     void    SetScale(float sx, float sy);
@@ -270,6 +272,7 @@ private:
     int     _popupAtMouseY = -1; // popup when mousey < this
     int     _transparency = 0;  // "incorrect" alpha (in legacy 255-range units)
     BlendMode _blendMode = kBlend_Normal; // render blend mode
+    int     _shaderID = 0;
     Pointf  _scale = Pointf(1.f, 1.f);; // x,y scale
     float   _rotation = 0.f;    // rotation, in degrees
     int     _zOrder = 0;
--- a/Common/gui/guimain.h
+++ b/Common/gui/guimain.h
@@ -142,7 +142,8 @@ public:
     BlendMode GetBlendMode() const { return _blendMode; }
     void    SetBlendMode(BlendMode blend_mode);
     int     GetShader() const { return _shaderID; }
-    void    SetShader(int shader_id);
+    int     GetShaderHandle() const { return _shaderHandle; }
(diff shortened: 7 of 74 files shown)


ivan-mogilko:
Add a support for Direct3D compiling shaders from source. This requires to link to d3dcompiler.lib, and a presence of d3dcompiler_xx.dll at runtime. From what I learnt, this library is a part of a standard DirectX package, starting at least since Windows 7. For WinXP configuration I disabled this feature, but engine can still use precompiled shader object (not that I expect ags 4 to be used on win xp).

Some interesting information over shader compilation library and its history:
https://walbourn.github.io/whats-up-with-d3dcompiler_xx-dll/
https://walbourn.github.io/hlsl-fxc-and-d3dcompile/

I also investigated doing reflection over a compiled shader using same library, in order to find location of shader variables dynamically. But I found that such reflection can only be done using Direct3D 10 and higher interfaces.
https://learn.microsoft.com/en-us/windows/win32/api/d3d10shader/nn-d3d10shader-id3d10shaderreflection
Judging by the article, this interface only requires same d3dcompiler_xx.dll, but I did not want to risk mixing D3D9 and 10 interfaces for now. Also I read somewhere else that this method won't work if shader is compiled using older syntax.

In the end, I decided to introduce a separate helper file that may optionally accompany shader file (whether precompiled or source code). The file should be called "<shadername>.d3def" (extension means "d3d shader definition"). This file is simply an ini format (because we already use ini format for several things), which contains a number of optional entries:

```
[compiler]
target = compilation target ("ps_2_0", and so on)
entry = entry function name (e.g. "main")

[constants]
<constant_name> = register index (a number >= 0)
```

If "compiler" options are not present, Direct3D will use defaults.
If "constants" are not present, Direct3D will use default hardcoded register values.

This approach also lets define custom variables, when we finally get to them.

ivan-mogilko:
Revamped shader API, added a new class called `ShaderInstance`, which is now assigned to game objects directly, instead of a "shader id" number.

"ShaderInstance" is a sort of a limited kind of "material" type (which you may meet in modern engines), it has only a reference to parent ShaderProgram it was created from, and a set of values for custom constants. Having Shader Instances allows to apply different shader values to different objects, while still using the same shader.

Here's how this new API looks like:

```ags
builtin managed struct ShaderProgram {
  /// Creates a new ShaderProgram by either loading a precompiled shader, or reading source code and compiling one.
  import static ShaderProgram* CreateFromFile(const string filename); // $AUTOCOMPLETESTATICONLY$
  /// Creates a new shader instance of this shader program.
  import ShaderInstance* CreateInstance();
  /// Gets the default shader instance of this shader program.
  import readonly attribute ShaderInstance* Default;
};

builtin managed struct ShaderInstance {
  /// Sets a shader's constant value as 1 float
  import void SetConstantF(const string name, float value);
  /// Sets a shader's constant value as 2 floats
  import void SetConstantF2(const string name, float x, float y);
  /// Sets a shader's constant value as 3 floats
  import void SetConstantF3(const string name, float x, float y, float z);
  /// Sets a shader's constant value as 4 floats
  import void SetConstantF4(const string name, float x, float y, float z, float w);
};
```

As you may see, ShaderProgram provides a "default instance" by its property. This is done for simple case when your shader does not have any custom constants, or when you dont want multiple instances. In this case the assignment would look like:

```ags
Screen.Viewport.Shader = myShaderProgram.Default;
```

If you need multiple instances of the same shader, here's what you may do:

```ags
cCharacter1.Shader = myShaderProgram.CreateInstance();
cCharacter2.Shader = myShaderProgram.CreateInstance();
cCharacter3.Shader = myShaderProgram.CreateInstance();
```

and then you may setup these instances separately:

```ags
cCharacter1.Shader.SetConstantF("CustomConstant", 1.0);
cCharacter2.Shader.SetConstantF("CustomConstant", 1.0);
cCharacter3.Shader.SetConstantF("CustomConstant", 1.0);
```

Shader instances may be stored in variables of course, and shared among multiple objects. So you may have, for example, 10 objects with one instance, and 10 objects with another instance. These first 10 objects will all be affected by first instance constant values, and next 10 will be affected by second instance constant values.

ivan-mogilko:
Added shaders to save games.
The game state should save the shader instances and constant values.
Graphics driver deletes all shader instances when restoring a save, and recreates new ones from the save. But the shaders themselves are not disposed, as they may get reused.

I'd like to achieve an independence on current graphics driver, where loading a game with e.g. software renderer and restoring a save made with opengl will keep the shader data, although shaders won't be working. This is done by caching some data inside script objects (ShaderProgram and ShaderInstance), from which it could be reapplied after restoring a save.
Probably this is not fully working right yet though.

ivan-mogilko:
Had to do a larger rewrite: moved shaders storage outside of the gfx driver. Gfx driver class now serves only as a factory, and the collection of shaders is managed by the engine. This allows to make their storage indexes independent from gfx driver, and whether it's capable or not of creating shaders overall.

Reworked shaders unserialization, and achieved that changes to shader constants are carried over correctly regardless of which gfx driver was used when the save was made and which is used when it's restored. Now you can e.g. run with software renderer, assign some shader constants (although no shaders are present in software mode), save game, switch to OpenGL, reload this save, and the shaders will have correct values.

ivan-mogilko:
I believe that this is practically ready for review, and possibly merging into ags4, as the first iteration of the feature.
There are following things that remain to be done:

1. Windows build is failing in Github Actions, because d3dcompiler.lib is missing. Not sure whether this has to be fixed in CMake scripts, or this library is not present in the environment used by Actions?
2. Because currently Viewport shader prevents "render sprites at screen resolution", I think to make rendering switch between two different sprite node setup depending on whether the shader is set or not. In other words, when it's present then the viewport will be rendered to a surface, and when not present - rendered directly to a backbuffer. This will remain until the problem is fixed (it requires some rewrite in d3d and ogl renderers, which will take some time).
3. Might add separate shaders for Camera (the difference from Viewport's shader is that multiple viewports can display same camera), and a shader for Screen (which will be applied to the whole game). These will have same consequence as the Viewport's shader, in that "render at screen resolution" will get disabled, for now.

ivan-mogilko:
@ericoporto do you know, what is required for Github Actions Windows builds to succeed? It looks like the D3DCompiler.lib is not found during a linking stage. Do I have to fix CMake script (or other project settings), or this needs an update of the enviroment in which the build is run?

ericoporto:
Uhm, since it builds just fine on the CI in Cirrus which was moved to the new docker with VS2022 and less things I am inclined to think it must be something missing to link that needs to be added in the CMake script, and that it should exist in the GitHub actions environment already. I need to think a bit on this, but I believe we had some directx stuff in cmake script that was removed in the past that I think can be added back to solve this.

ivan-mogilko:
> For D3D we can allegedly use https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dxgetpixelshaderprofile to get the highest supported shader model dynamically (max `ps_3_0` for D3D9).

For which purpose would that be used?

AlanDrake:
> For which purpose would that be used?

It was before I read about d3ddef. It could still be useful to dynamically compile to the current target... in theory it may be possible to catch the shader model version with define macros and make the shader code do something different. But I guess that's an optional.

Anyway, yesterday I was not able to use Shader model 3, despite a successful compilation I only got a black screen. Perhaps I'm missing something. I need to do more testing.

AlanDrake:
Would this be the correct way to scale the texture sizing for a "destination size" parameter?
```
    data[program->OutputDim.Off + 0]    = static_cast<float>( _scaling.X.ScaleDistance(bmpToDraw->GetWidth()) );
    data[program->OutputDim.Off + 1]    = static_cast<float>( _scaling.Y.ScaleDistance(bmpToDraw->GetHeight()) );
```
Results looks fine, I'm only unsure if I'm using the right methods.

ivan-mogilko:
> Would this be the correct way to scale the texture sizing for a "destination size" parameter?

Because you do not explain, I have to *guess* that you mean the size that this texture will be taking in window?
What if the texture is drawn on another texture, which in turn will be scaled before rendering etc?

AlanDrake:
> I guess that you mean the size that this texture will be taking in window after the game is scaled?

Yes

> What if this texture is drawn on another texture, which in turn will be scaled before rendering etc, will this parameter keep its meaning in that case?

So far it's necessary for `Screen.Shader`, but I don't know what other effects may need knowing both texture size and scaled size.
And maybe `_scaling` is relative to overall screen scaling, and I need the sprite own scaling? The code on the renderers has changed seevral times, I'm not sure where to get the current sprite scaling from RenderTexture().

ivan-mogilko:
@AlanDrake I implemented your suggested "OutputDim" constant, but did it differently. For now it's only applied to the final surface render. The values will depend on whether this is a output to the window or a output to a texture (in case of making a screenshot).

If we find that this constant may be useful for game sprites as well, then this implementation may be adjusted.

Attaching your CRT shaders:
[crt-shaders.zip](https://github.com/user-attachments/files/20125306/crt-shaders.zip)


NOTE: I used following d3ddef file for Direct3D shader, because it requires a specific minimal compiler target:
```
[compiler]
target = ps_2_a

[constants]
iTextureDim = 2
iOutputDim = 4
```

Maybe we could raise the default compiler target to something else, but latest D3D9 targets (4 and higher) have a different syntax for the main function afaik), and i am not familiar with that. That's a subjective issue, of course. I am open for suggestions here.

Previously mentioned D3DXGetPixelShaderProfile function requires another utility library D3dx9.lib, which we do not link atm.

ivan-mogilko:
Rebased, and added Mouse.CursorShader property.

Is there anything that needs an update here before merging this into ags4? Have people had a chance to test this out?

ivan-mogilko:
@ericoporto I'm confused by cmake, because it has a addition of "shlwapi.lib" for Windows build, but not "d3d9.lib". How does the engine get linked with "d3d9.lib"?

ivan-mogilko:
@ericoporto looks like the cmake is fixed now.

AlanDrake:
I think it's good enough for merging.
But add a note that if you define a constant on d3ddef, then you *have* to define all of them, not just the extra one.

PR #2717, opened 2025-04-13
AGS 4 Editor: Cleanup delegate calls and declarations

edmundito:
This change cleans up the event handlers using a simplified syntax and replaces delegates without arguments with [System.Action](https://learn.microsoft.com/en-us/dotnet/api/system.action-1?view=netframework-4.6) to simplify the syntax.

### DIFF
--- a/Editor/AGS.Editor/Panes/ScriptEditor.cs
+++ b/Editor/AGS.Editor/Panes/ScriptEditor.cs
@@ -17,7 +17,7 @@ namespace AGS.Editor
         public delegate void AttemptToEditScriptHandler(ref bool allowEdit);
         public static event AttemptToEditScriptHandler AttemptToEditScript;
 
-        private delegate void AnonymousDelegate();
+        private Action _anonymousDelegate;
 
         // Custom Edit menu commands
         private const string TOGGLE_BREAKPOINT_COMMAND = "ToggleBreakpoint";
@@ -57,8 +57,8 @@ namespace AGS.Editor
             _roomNumber = -1;
             Init(scriptToEdit);
 
-            this.Load += new EventHandler(ScriptEditor_Load);
-            this.Resize += new EventHandler(ScriptEditor_Resize);
+            this.Load += ScriptEditor_Load;
+            this.Resize += ScriptEditor_Resize;
         }
 
         public string GetScriptTabName()
@@ -344,7 +344,7 @@ namespace AGS.Editor
             {
                 if (this.IsHandleCreated)
                 {
-                    this.Invoke(new AnonymousDelegate(UpdateFunctionList));
+                    this.Invoke(new Action(_anonymousDelegate));
                 }
             }
         }
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -356,7 +356,7 @@ namespace AGS.Editor
 
             _game = new Game();
             _debugger = new DebugController(_engineComms);
-            _debugger.BreakAtLocation += new DebugController.BreakAtLocationHandler(_debugger_BreakAtLocation);
+            _debugger.BreakAtLocation += _debugger_BreakAtLocation;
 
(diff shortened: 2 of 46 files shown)


ivan-mogilko:
There's one thing that bothers me here. When you have a delegate declared, you may also give names to parameters, for example:
```
public delegate void AfterLabelEditHandler(string commandID, ProjectTreeItem treeItem);
delegate void PropertyChangedHandler(string propertyName, object oldValue);
```
When you have an Action, you see only parameter types:
```
public event Action<string, ProjectTreeItem> OnAfterLabelEdit;
public event Action<string, object> OnPropertyChanged;
```

Sometimes it's possible to deduce what these are from the function's name and parameter type, but not always.

Then, when adding a new function to the event, if you type a new function name after `+=`, and press TAB, with delegates Visual Studio would generate the function with parameter names for you, for example:
```
private void mainForm_OnPropertyChanged(string propertyName, object oldValue)
{
}
```
With Action<> or Function<>, it will instead generate this:
```
private void mainForm_OnPropertyChanged(string arg1, object arg2)
{
}
```

I do not see how that's more beneficial. Is there any way around this when using Action<> and Func<> declarations?

edmundito:
Good point. I did look this up now and unfortunately I can confirm  this is limitation. This is something I got used to while working on Unity games and had no issues but if we want more clarity, I can close this PR and leave things as they are.


In researching this issue I also found that the codebase may not be using the "proper" convention for events declaration anyway: https://stackoverflow.com/a/2282683

persn:
> One question I do have is whether allocating the delegate is necessary over just adding the callback directly. Any ideas? Is what we have older syntax?

It's older syntax. If you're in doubt you can type your code into https://sharplab.io/ and see what the C# compiler interpreted the differences to be. In this case the newer syntax will generate the older syntax so they should be functionally equivalent

edmundito:
Ok, I did another pass on this and make a cleaner commit with only cleaning up what was necessary in #2727. Closing this one.

ivan-mogilko:
@edmundito you could actually do a force push on this PR's branch, replacing commits, while keeping the same PR open.

ivan-mogilko:
@edmundito these are still old commits in this pr? Or are you planning to change them later?

edmundito:
@ivan-mogilko Whoops, looks like there was some mixup with push. I was able to resolve it and now the correct branch has replaced it.

Editor/AGS.Editor/Panes/ScriptEditor.cs
```diff
@@ -344,7 +344,7 @@ private void AutoComplete_BackgroundCacheUpdateStatusChanged(BackgroundAutoCompl
             {
                 if (this.IsHandleCreated)
                 {
-                    this.Invoke(new AnonymousDelegate(UpdateFunctionList));
+                    this.Invoke(new Action(_anonymousDelegate));
```
ivan-mogilko, line 347: This looks like a mistake; you are replacing a invoke to UpdateFunctionList with anonymousDelegate.

ivan-mogilko:
I apologize for delay; I've scanned through the changes, and it looks fine, except there's a strange replacement in ScriptEditor.cs which I noted above.

Editor/AGS.Types/View.cs
```diff
@@ -9,8 +9,7 @@ namespace AGS.Types
     [DefaultProperty("Name")]
     public class View : IToXml, IComparable<View>
     {
-        public delegate void ViewUpdatedHandler(View view);
-        public event ViewUpdatedHandler ViewUpdated;
+        public event Action<View> ViewUpdated;
```
ivan-mogilko, line 12: I suggest to keep events in interfaces (such as AGS.Types library) unchanged, and later remake them into proper events that take 2 arguments: `(object sender, XEventArgs args)`

edmundito:
> I apologize for delay;

No problem, I am delayed with this too. I will address the comments soon and need to do another pass to see if there is anything else.

edmundito:
The is rebased and did another check to ensure that all handlers were covered.

One thing to keep in mind that any event handlers in the `*.Designer.cs` files need to have `new EventHandler` left as-is because this is how the VS UI Editor can identify the event handler and show it in the properties panel. Otherwise, the event will show as blank.

PR #2720, opened 2025-04-26
libsrc: upgrade GLM to 2d4c4b4

ericoporto:
glm upgrade to https://github.com/g-truc/glm/commit/2d4c4b4dd31fde06cfffad7915c2b3006402322f

There is no need to actually upgrade GLM to the newest as possible other than to keep it in sync with its development, but I am doing it here in case it causes any issue we pick. It builds ok and apparently works, and looking on its open issues it appears there shouldn't be any new major flaw introduced. Leaving it as draft until further tests.


ivan-mogilko:
Why do we need to keep in sync with its development; and in any case, why not sync with the official releases?
Their latest published release [1.0.1](https://github.com/g-truc/glm/releases/tag/1.0.1) is of commit https://github.com/g-truc/glm/commit/0af55ccecd98d4e5a8d1fad7de25ba429d60e863

ericoporto:
That is a good point, I switched instead to point to version 1.0.1. I used the version instead of the commit because it looks like that commit was in a branch that doesn't exist anymore? (I get a message that the commit isn't in the repository in the github website)

ivan-mogilko:
This is a draft now, what would be required to merged this?

ericoporto:
> This is a draft now, what would be required to merged this?

The only thing would be testing. It probably there isn't any significant change, but its nice to keep dependencies updated.

PR #2721, opened 2025-04-27
Optional saves compression

### LINKED ISSUE #2718: Compress savegames
**Problem**

Because AGS saves practically contain a dump of all objects and script data in game (both statically and dynamically allocated), they may become quite large. Having numerous saves may occupy disk space comparable with the size of the game (or even a multiple of that, if a game is small).

**Suggestion**

Write saves compressed using a zip algorithm. We already use miniz for compressing sprites, same could be used to compress and decompress saves.

This should be controlled by a option in the game config (not a game setting) being enabled by default.

ivan-mogilko:
Resolves #2718

This implements an ability to compress save data. The engine does not compress the whole save file, but the parts with the game state data within. Any meta data, such as save header, user description, and individual components meta-data - is kept uncompressed. The components internal data is compressed.
Each component's header will have a flag telling that the following data is compressed, in which case engine will know to run decompression or not.

Having this as a per-component switch leaves a possibility to disable a generic compression over certain component and / or delegate the compression to the component handler, would that will be required for convenience or technical reasons.

Compression uses "deflate" algorithm (with the use of miniz library that we already have).

---

Unfortunately seeking is not possible while compressing or decompressing, and we were using Seek in a few places when saving, in order to return back and write down calculated number of items, or similar. I had to change the algorithm there to run a item count precalculation first.

The one case where I was not able to fix this trivially is plugin data, because there's no way to precalculate it without asking plugins at least, and there's no way to do so in plugin API. For now I had to write plugin data into memory first, and then compress into the file stream. I'm thinking that we could limit amount of the data that plugins save with some safe amount, like 32 or 64 MB (not that I know of plugins that write anything close to that...).
Reading plugin data back is done right from the uncompression stream, but if plugins will need to seek around their data, then we'll have to also read everything to memory first.

I'd like to note that allocating extra memory is likely less a problem when loading rather than saving, because all dynamic data is disposed before loading a save.

---

Added unconditional compression of the game screenshot in saves. It's compressed always, regardless of the config option.

---

Added a config option:
```
[misc]
compress_saves = 0/1
```

NOTE: engine will be able to load both compressed and not compressed saves regardless of this setting. This setting only tells whether NEW saves will be compressed.

Added this option into WinSetup, on "Paths" tab, because there is a custom save path. Renamed the "Paths" tab to "Game Files".

**RESULTS**

Need to run tests on multiple games, preferably long games that create big saves.

Quick test on a dummy game made from a basic template shows decrease in save size by 6.5 times: from ~48 KB to ~7 KB.

The "Old Skies" saves went down from 17.8 MB to 3.2 MB (18%). AFAIK half of that gain is from compressing a screenshot alone.

"Rosewater": a user reported very large saves, but probably they become big only in mid game. Saves made in the first scene are around 1.2 MB, and they compress down to 127 KB, so like x10 times.


ericoporto:
> write game state into a temp file,
> open the temp file, read data out, compress and write into the real file,

Uhm, since this writes and then reads back, is there any reason this can't just write in memory directly? Is this because the readback is done in chunks to not use too much ram memory?

ivan-mogilko:
> Uhm, since this writes and then reads back, is there any reason this can't just write in memory directly? Is this because the readback is done in chunks to not use too much ram memory?

That's because we don't know how big the saves may be, and what is current memory requirement of the game.
If there were a way to calculate, we could have it do this in memory if the save is small enough. That may be realistic to do when restoring a save, because it's decompressed size is mentioned in the header. What's left is to get current memory capacity reliably on each platform. But this won't work when writing a save, because the total size is unknown. Or rather it will require running around game data and making an approximate guess. That seems rather bothersome.

AlanDrake:
I never tried myself, but miniz should be able to chunk compression into multiple blocks. The heavy parts are usually gonna be screenshots and dynamic sprites.
When we deprecate 32bit builds, memory is likely not going to be a problem anymore,

As for compressing images, miniz offers a  `void *tdefl_write_image_to_png_file_in_memory()` which we could use until a better option is available (can be used as well to add png to normal screenshot saving).

ivan-mogilko:
> I never tried myself, but miniz should be able to chunk compression into multiple blocks. The heavy parts are usually gonna be screenshots and dynamic sprites.

Screenshot is a non-problem, as that's a single full-screen image, so it may as well be compressed in memory. Engine currently decompresses sprites in memory, and individual sprites may be as large.

What is a problem is save components, but not just because they may be big, but because when writing save components program has to return back to the component's header to save its size. This means that *we cannot start compressing a component chunk before its fully complete*. This may be an issue with dynamic sprites component (as you pointed out) and sometimes script's managed object pool (in case user have allocated a lot). The issue here is not only that they may require a lot of memory, but that the rest of the game is still in memory too when this process takes place. If there are alot of dynamic objects present, compressing them right in memory might raise game's memory requirement by, say, x1.5.

> When we deprecate 32bit builds, memory is likely not going to be a problem anymore,

64-bit builds are still going to be restricted to the user's system memory.

The way I see this currently, using temporary files may be relatively slower, but more stable than using RAM. And, unlike streaming sprites, for example, restoring/saving a game is not something that happens frequently enough to cause performance problems. That's my assumption; the actual performance should be known from testing this build on games which make big saves.

> As for compressing images, miniz offers a void *tdefl_write_image_to_png_file_in_memory() which we could use until a better option is available (can be used as well to add png to normal screenshot saving).

When you say "images" (plural), do you refer to the screenshot only or something else? Are you suggesting to save screenshots as an actual PNG image format? At first I was going to simply compress the raw data same way as the sprite file compresses sprites. Is your idea to let extract screenshots by a external tool easier?

AlanDrake:
> 64-bit builds are still going to be restricted to the user's system memory.
And swap file, hopefully.

> When you say "images" (plural), do you refer to the screenshot only or something else? Are you suggesting to save screenshots as an actual PNG image format? At first I was going to simply compress the raw data same way as the sprite file compresses sprites. Is your idea to let extract screenshots by a external tool easier?

Right, sorry. This method also writes the png headers which you wouldn't need for dynamic sprites and modified room backgrounds.
It's just one could compress these elements one by one, if that helps.

Main argument against dumping on disk directly is that SSDs have limited write cycles, MicroSDs even less so (eg: on Switch)
Normally the writes are not significant, but if one of our 1080p users introduces an autosave on each room change, that could potentially become a scary amount of writes in the worst possible cases.

ivan-mogilko:
> Right, sorry. This method also writes the png headers which you wouldn't need for dynamic sprites and modified room backgrounds.
It's just one could compress these elements one by one, if that helps.

Well, same could be done by a basic compression function, which does not add any extra image headers. Engine is currently using it when compressing sprites.

What does a SSD "write cycle" imply? I read a bit about it, but did not understand fully yet. Will it help, for instance, if a temp file will always be the same, and overwritten?

I might look into precalculating memory space, or compressing components separately, although the latter will complicate the process.

ivan-mogilko:
Actually, if not the whole game state, but each component's contents are compressed individually, then that might solve the issue with seeking. Supposing, the component's header and meta data remain uncompressed, then the process would be like:

- for each component:
- write header placeholder
- open compression stream into the same file and write component's contents into that stream
- close compression stream but keep the file stream
- calculate component's size, seek back and write meta data

The issue here is that we must enforce a rule to not have a Seek anywhere in "Save" functions around the game objects etc.
The Seek forward may be allowed, and reimplemented as read or write N dummy bytes in the compression stream (although it may be better to have explicit "SkipBytes" method).

AlanDrake:
> What does a SSD "write cycle" imply? I read a bit about it, but did not understand fully yet. Will it help, for instance, if a temp file will always be the same, and overwritten?

SSDs generally use wear leveling, with algorithms that shuffles data around depending on how often it changes, which are the freshest cells, etc.
The result is that it will last much longer than the cheap flash drives, which once a block fails it's over (good ones would use wear leveling as well).
It also means that the more space the SSD holds overall, the longer it will last, since the wear is distributed. A small 128GB will have much shorter life compared to a 2TB one.

The TBW indication is how many terabytes you can write before things start to go bad, and it's affected by the quality of the components (some have higher endurance than others) and the overall space available (thanks to wear leveling).

ivan-mogilko:
Alright, here is the new variant: only individual component data is compressed, and compression is done right into the same file stream.

This change restricts us from using Seek anywhere in game object and save functions and component save functions. Although deflate stream could implement reading out bytes in case of forward seeking, so letting to use Seek to skip some parts. But it will be not possible to seek back to prepend a size of something, so everything will have to be calculated beforehand.

I had to fix few save functions where the program would seek back to prepend some calculated value. Instead it has to precalculate these values first, so instead of 1 object iteration loop there will be 2 loops (on saving). There were 2 cases where this happens:
- finding number of topmost count of dynamic sprites
- finding number of valid overlays

The one case where I was not able to fix this trivially is plugin data, because there's no way to precalculate it without asking plugins at least, and there's no way to do so in plugin API. For now I had to write plugin data into memory first, and then compress into the file stream. I'm thinking that we could limit amount of the data that plugins save with some safe amount, like 16 or 32 MB (not that I know of plugins that write anything close to that...).
Reading plugin data back is done right from the uncompression stream, but if plugins will need to seek around their data, then we'll have to also read everything to memory first.

I'd like to note that allocating extra memory is likely less a problem when loading rather than saving, because all dynamic data is disposed before loading a save.

---

In addition, this compresses game screenshots now too. I made it do so unconditionally (always), because in high-res games screenshots may take around 50% of the save at the start of the game.

ivan-mogilko:
I think this is mostly done now. Compression of the game state is optional, done on a per-component basis, not using any temporary files. Screenshot's compression is made non-optional, but is controlled by a flag, so may be disabled too if required.

PR #2724, opened 2025-04-28
DRAFT AGS 4 Editor: Add double-buffer support for TreeView

edmundito:
(Will update the description once I get feedback/create a PR ready for review)

## Changes

This fixes something that has bothered me for the longest time. When clicking/hovering over the tree views, they flicker. This is due to how they are set up by default to accommodate Windows XP limitations, which the editor no longer supports. See this article for details: https://stackoverflow.com/a/10364283/20494

## Preview

TODO

Before:

After:


ivan-mogilko:
Applied as commit b0b199703e965c023917034a881251595df66afe.
Only had to move the tree view function to AGS.Controls.TreeViewExtensions.

PR #2728, opened 2025-05-10
AGS 4 Editor: Update .NET framework from 4.6 to 4.7

edmundito:
This updates the .NET framework used for the AGS Editor from 4.6 to [4.7](https://devblogs.microsoft.com/dotnet/announcing-the-net-framework-4-7) for a couple of good reasons:

1. Enable improved support for high DPI scaling. This has been attempted before and had to be rolled back (See #2087), but I'd like to take advantage of the [improved support](https://devblogs.microsoft.com/dotnet/announcing-the-net-framework-4-7/?WT.mc_id=dotnet-35129-website#take-advantage-of-high-dpi) and [new API introduced in 4.7](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.control.logicaltodeviceunits?view=netframework-4.7) onward to make changes to the controls to support it.
2. It's still compatible with Win 7 SP1: https://learn.microsoft.com/en-us/dotnet/framework/get-started/system-requirements#installation-requirements
3. Editor seems to run with no issues with this updated framework

## TODOs

- [x] Update builds to use an updated framework
- [x] Update Editor installer to use new framework (if needed
- [x] Update dev environment READMEs

### DIFF
--- a/ci/windows/Dockerfile
+++ b/ci/windows/Dockerfile
@@ -6,8 +6,8 @@ RUN IF exist %TEMP%\nul ( echo %TEMP% ) ELSE ( mkdir %TEMP% )
 
 # .NET Framework 4.7 Developer Pack
 RUN pushd %TEMP% && \
-  curl -fLO https://download.microsoft.com/download/3/5/1/351f9afb-0505-47dc-a42e-34935707f300/NDP47-DevPack-KB3186612-ENU.exe && \
-  start /install /quiet /norestart && \
+  curl -fL https://download.microsoft.com/download/3/5/1/351f9afb-0505-47dc-a42e-34935707f300/NDP47-DevPack-KB3186612-ENU.exe -o dotnetsetup.exe && \
+  start /wait dotnetsetup.exe /install /quiet /norestart && \
   popd && \
   mkdir empty && \
   robocopy empty %TEMP% /MIR > nul & \
--- a/ci/windows/Dockerfile
+++ b/ci/windows/Dockerfile
@@ -4,10 +4,10 @@ FROM ghcr.io/ericoporto/min-ags-dev-env:2.0.16
 # if no temp folder exists by default, create it
 RUN IF exist %TEMP%\nul ( echo %TEMP% ) ELSE ( mkdir %TEMP% )
 
-# Windows 10.0.10240 SDK (but only install the .NET 4.6 SDK)
+# .NET Framework 4.7 Developer Pack
 RUN pushd %TEMP% && \
-  curl -fLO https://download.microsoft.com/download/E/1/F/E1F1E61E-F3C6-4420-A916-FB7C47FBC89E/standalonesdk/sdksetup.exe && \
-  start /wait sdksetup /ceip off /features OptionID.NetFxSoftwareDevelopmentKit /quiet /norestart && \
+  curl -fLO https://download.microsoft.com/download/3/5/1/351f9afb-0505-47dc-a42e-34935707f300/NDP47-DevPack-KB3186612-ENU.exe && \
+  start /install /quiet /norestart && \
   popd && \
   mkdir empty && \
   robocopy empty %TEMP% /MIR > nul & \
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -315,7 +315,7 @@ build_editor_task:
     set "UseEnv=true" &&
     copy C:\Lib\irrKlang\x86\*.dll Editor\References\ &&
     cd Solutions &&
-    cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6\Lib\um\x86;!LIB!" &&
+    cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\NETFXSDK\4.7\Lib\um\x86;!LIB!" &&
     msbuild AGS.Editor.Full.sln /p:PlatformToolset=v142 /p:VCToolsVersion=14.29.30133 /p:Configuration=Debug /p:Platform="Win32" /maxcpucount /nologo"
   test_debug_script: |
     curl -fLOJ https://github.com/nunit/nunit-transforms/raw/master/nunit3-junit/nunit3-junit.xslt
@@ -328,7 +328,7 @@ build_editor_task:
     set "UseEnv=true" &&
     copy C:\Lib\irrKlang\x86\*.dll Editor\References\ &&
     cd Solutions &&
-    cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6\Lib\um\x86;!LIB!" &&
+    cmd /v:on /c "set "LIB=C:\Program Files (x86)\Windows Kits\NETFXSDK\4.7\Lib\um\x86;!LIB!" &&
     msbuild AGS.Editor.Full.sln /p:PlatformToolset=v142 /p:VCToolsVersion=14.29.30133 /p:Configuration=Release /p:Platform="Win32" /maxcpucount /nologo"
   test_release_script: |
     curl -fLOJ https://github.com/nunit/nunit-transforms/raw/master/nunit3-junit/nunit3-junit.xslt
--- a/Changes.txt
+++ b/Changes.txt
@@ -20,6 +20,7 @@ Common:
  - Removed support for '[' as a linebreak character in text.
 
 Editor:
+ - Editor requires .NET Framework 4.7 to run.
  - Open room project format: rooms are now saved as subfolders in "Rooms" folder, where each
    component has its own separate file (room data, backgrounds, masks and scripts).
  - "Fonts" node in the Project Explorer now contains separate "Font Files" and "Fonts" sublists.
--- a/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
+++ b/Editor/AGS.CScript.Compiler/AGS.CScript.Compiler.csproj
@@ -23,7 +23,7 @@
     <OldToolsVersion>3.5</OldToolsVersion>
     <UpgradeBackupLocation>
     </UpgradeBackupLocation>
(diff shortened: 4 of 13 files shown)


edmundito:
I'm going to downgrade this PR to draft because there are some things that need to be upgraded to 4.7 and I'll add them to a todo list soon.

In the meantime I'd like to know if we are OK to move forward to 4.7 or what concerns need to be addressed.

ericoporto:
The Windows docker has currently the .NET Framework 4.6 SDK installed. You would need to instead install the .NET Framework 4.7 Developer pack (see here https://dotnet.microsoft.com/en-us/download/dotnet-framework/net47 , the download should be the offline installer like [here](https://dotnet.microsoft.com/download/thank-you/net47-developer-pack)). Then you need to run it in your machine from command prompt passing `/?` and it will inform the flags it supports.

image

Now you need to use this to redo this line on the docker here:

https://github.com/adventuregamestudio/ags/blob/8bf6bb8378301fb2221cecfd6c441b2fd32566ed/ci/windows/Dockerfile#L7-L14

Then you should be able to install .NET Framework 4.7 on the docker image we use when building.

ci/windows/Dockerfile
```diff
@@ -4,10 +4,10 @@ FROM ghcr.io/ericoporto/min-ags-dev-env:2.0.16
 # if no temp folder exists by default, create it
 RUN IF exist %TEMP%\nul ( echo %TEMP% ) ELSE ( mkdir %TEMP% )
 
-# Windows 10.0.10240 SDK (but only install the .NET 4.6 SDK)
+# .NET Framework 4.7 Developer Pack
 RUN pushd %TEMP% && \
-  curl -fLO https://download.microsoft.com/download/E/1/F/E1F1E61E-F3C6-4420-A916-FB7C47FBC89E/standalonesdk/sdksetup.exe && \
-  start /wait sdksetup /ceip off /features OptionID.NetFxSoftwareDevelopmentKit /quiet /norestart && \
+  curl -fLO https://download.microsoft.com/download/3/5/1/351f9afb-0505-47dc-a42e-34935707f300/NDP47-DevPack-KB3186612-ENU.exe && \
```
ericoporto, line 9: Instead of using the `O` flag, you can use `-o` flag to rename the output for something (say `netsetup.exe`), so you could use `curl -fL https://download.microsoft.com/download/3/5/1/351f9afb-0505-47dc-a42e-34935707f300/NDP47-DevPack-KB3186612-ENU.exe -o netsetup.exe` instead.

I mention this because in the line below you are calling `start` without passing the name of the executable you want to start. It should be like

```
start /wait NDP47-DevPack-KB3186612-ENU.exe /install /quiet /norestart && \
```

or if you renamed it

```
start /wait netsetup.exe /install /quiet /norestart
```

edmundito: Whoops, didn't quite get the start command. Fixed it.

edmundito:
Updated this to be ready for review.

This change does not offer anything new to the Editor until all the high-dpi issues can be corrected. Depending on availability, this may take some time to complete and should not block the eventual release of 4.0. If accepted and merged, each problem can be addressed individually by taking advantage of the new APIs in 4.7 and releasing them with current builds. Later, high-dpi support can be enabled once all the support has been completed.

ericoporto:
Hey, when it makes the installer there is a `DOTNET_FRAMEWORK_RELEASE_VERSION` with a set version. I never knew from where that number should come from but it looks like there was something wrong with the number before and you fixed. Can you tell me from where you picked that number? I never knew from where it should come from.

edmundito:
> Hey, when it makes the installer there is a `DOTNET_FRAMEWORK_RELEASE_VERSION` with a set version. I never knew from where that number should come from but it looks like there was something wrong with the number before and you fixed. Can you tell me from where you picked that number? I never knew from where it should come from.

The numbers are provided right in the doc in the comment above the line:
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version

edmundito:
I pushed one final change to provide a way to enable DPI scaling through the Editor's app config so that DPI fixes can be worked out without having to re-do the line. I removed the old way of doing it, which is no longer recommended.

https://learn.microsoft.com/en-us/dotnet/desktop/winforms/high-dpi-support-in-windows-forms#configuring-your-windows-forms-app-for-high-dpi-support

image

ericoporto:
I checked this runs fine in Windows 11, and I don't have any objection. I would maybe adjust the indentation in `app.config` for the Editor, but that is the only thing I could find.

I guess a Windows 7 check would be nice from @ivan-mogilko

ivan-mogilko:
Well, Editor starts up. Do I need to test something specific?

I may propose to add the linked version of .NET Framework to the About dialog here:
https://github.com/adventuregamestudio/ags/blob/5c352b7a84d35f51e1c7fd3b59f6d13ceaf996a7/Editor/AGS.Editor/GUI/AboutDialog.cs#L31-L39

edmundito:
@ivan-mogilko I added showing the .NET runtime version running the app, and noticed that Windows 11 was displayed as 10, so I fixed that too:

image

ivan-mogilko:
My understanding is that registry tells the installed version, does that always correspond to the one that the application is actually using?

edmundito:
Checking the registry seems to be the recommended way to determine this.

I started from here: https://learn.microsoft.com/en-us/dotnet/api/system.environment.version

Which took me to here: https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed

ivan-mogilko:
I've read these pages too, but I could not understand if "installed version" = "version that the application uses".

There's also `RuntimeInformation.FrameworkDescription property` mentioned on that page, which "Gets the name of the .NET installation on which an app is running.":
https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#runtimeinformationframeworkdescription-property

ivan-mogilko:
Another reason to double check this: if we care about potential for porting the Editor to something crossplatform, then we should not be using anything Windows-specific, and afaik Registry is (it also comes from namespace Microsoft.Win32).

edmundito:
> I've read these pages too, but I could not understand if certain "installed version" = "version that the application uses", because in theory one may have multiple .NET frameworks installed on a system.
> 
> 
> 
> There's also `RuntimeInformation.FrameworkDescription property` mentioned on that page, which is described as "Gets the name of the .NET installation on which an app is running." and "The property returns a string that indicates the name of the currently executing .NET implementation":
> 
> https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#runtimeinformationframeworkdescription-property

I can bump the framework version to 4.7.2 which is the last 4.7.x version. This also provides RuntimeInformation which is introduced in 4.7.1. If you feel you're fine with this version, I am happy to make the change.

persn:
My opinion is to not do it, In .NET Core it's much easier to do so but in .NET Framework this is the perfect "Works on my machine, doesn't on someone elses". It's a lot of headache just for the sake of a number that no one cares about.

However if you insist on doing it anyway you can try and check if the code I wrote in 2017 still works, it bypasses everything by checking version directly on the assemblies being used by the runtime. I favored this solution because it doesn't rely on hardcoding OS and such, but I have no idea if it still works https://github.com/persn/ags/blob/master/Editor/AGS.Editor/Utils/Utilities.cs#L43

ericoporto:
I guessed a good name for the function would be `GetNETFramework` and searched it here in GitHub too to see in other projects, and they used the registry strategy too (like [here](https://github.com/nickorzha/roslynedit/blob/44cd38e56e0aadf82a99991e01760a054912c80f/src/RoslynPad.Common.UI/PlatformsFactory.cs#L105-L145)). I also searched for the registry string here too to check how other projects were using and it looked similar, but yes, it does get the installed .NET Framework. (I did found a [weird hack alternative](https://github.com/Milstein/AspxCommerce2.0_Source/blob/36dba6d5c6dd0060cf15d7a654277a5e50de4851/SageFrame.Common/Application/Application.cs#L235-L262) using assemblies...).  I also found dotnet46 source in github somehow, and it looks like the finding out to which [target framework is intended](https://github.com/bigfish1913/Source/blob/18c50da18a55f2abd0c42dd346a1fc68eccc9fc3/dotnet46/Source/ndp/fx/src/xsp/system/Web/Compilation/MultiTargetingUtil.cs#L149) is more an ASP.Net thing.

I can't tell much, but if a "NET version at build time" is something intended, it seems like it requires hardcoding a string there.

persn:
> `RuntimeInformation.FrameworkDescription`

I think this will only tell you whether you got .NET Framework, .NET Core, Mono, or just .NET, it doesn't tell you anything about the version number

edmundito:
The good news is that code is flexible and commits can be undone. I will drop those last changes and focus on the framework update first (and make the Windows 11 support as a separate commit). These changes don't seem to get the result we're looking for

The only confirmation I need is whether I should stay with 4.7 or bump to 4.7.2, mainly because I avoid asking game authors to update twice if we do it later. I'm set on 4.7 because of its compatibility with Mono and the potential to experiment with cross-platform support in the future. Thoughts?

ivan-mogilko:
> These changes don't seem to get the result we're looking for

I cannot tell what this refers to, the existing `Utilities.NetRuntimeVersion` works fine for me, I wanted to know if it works for others.
But if it does not suit this PR, I can add the .net version line to About dialog separately too.

edmundito:
> I have no knowledge of this; did Mono stop its support at 4.7 and won't update further?

That is the last version that is fully supported: https://www.mono-project.com/docs/about-mono/compatibility/ - I was able to confirm somewhere else that it was up to 4.7.2.

However, about a month ago it changed ownership from Microsoft to the Wine foundation so not sure what the future holds.

ivan-mogilko:
I think this may be just merged as it is. If upgrading further is under doubt, then that may be discussed later. I don't see much problem in updating framework again. This is a in-development version, and not too many people are using it currently. Also, as I mentioned above, people likely have higher versions of framework installed.

PR #2731, opened 2025-05-18
AGS4 Engine: fixing movement issues

AlanDrake:
This hotfix should resolve the change in behavior after the last refactor that was causing movement skips, because `do_movelist_move()` needs to process the current stage before increasing onpart (as per original behavior).

So I added a new method to avoid interfering with the MotionPath script API.

There were (and still are) some mismatches in the reported MotionPath coordinates when comparing from script the X/Y of player vs motionpath, apparently related to walkable area zoom.
I suspect it's an issue with the old hacks to slow down movement at smaller scalings. If that's the case, the proper fix could be using float coordinates internally, but that's another matter.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -2469,7 +2469,7 @@ builtin managed struct Object {
   /// Gets/sets whether the object will be drawn and updated during the game update.
   import attribute bool Enabled;
   /// Gets this object's current MotionPath, or null if it's not moving.
-  import attribute MotionPath* MotionPath;
+  import readonly attribute MotionPath* MotionPath;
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
I will merge this as a hotfix, but this solution brings MoveList to a internally inconsistent state, where "current position" does not correspond to "progress". So this cannot stay long-term and will have to be replaced with another solution.
Ideally, MoveList should be always internally consistent, where you can calculate current pos from progress (which is `onpart` variable in the current implementation) and vice versa. Any special behavior should rather be performed by an external code, like character's walking update logic.

I'd like to create a walking test case in [our auto-test game](https://github.com/adventuregamestudio/ags-test-games/issues/10) so to be able to keep an eye on any behavioral differences that the code changes might induce.

PR #2732, opened 2025-05-23
AGS 4: add GUIControl.Rotation, ScaleX, ScaleY, SetScale(x,y)

ivan-mogilko:
This brings gui controls in consistency with the GUI and rest of game objects.
Although working technically, gui controls are currently cannot be affected by the "render at screen resolution" setting, because they are drawn on a intermediate GUI surface first, therefore rotated and scaled controls look worse than rotated and scaled GUIs or Overlays. The only exception is rotating by multiplies of 90 degrees and scaling up by integer factor.


ericoporto:
You wrote `Gets/sets the control's image rotation in degrees.` in the comment for the Rotation property, from what I get from the `IsOverControlImpl` is it doesn't take into account the control rotation (does it take into account the GUI the control is in rotation?). Wouldn't be better to name the property as `GraphicRotation` instead (if it doesn't account for clicks in the rotated graphic...)?

ivan-mogilko:
IsOverControlImpl does account for both scaling and rotation, as it receives coordinates already transformed into the control's local space.

ericoporto:
In that case I think it's better to remove the work `image` from the comment of the `Rotation` property in script, so it reads just `Gets/sets the control's rotation in degrees.`.

ivan-mogilko:
I forgot to apply transformation in mouse interactions, should be fixed now.

ivan-mogilko:
There's still something wrong with how Sliders are drawn; that's because they have an uncommon bounding box. I will need to fix this.

ericoporto:
> I forgot to apply transformation in mouse interactions, should be fixed now.

It is! Thanks! Working great now!

> There's still something slightly wrong with how Sliders are drawn, best visible with standard slider looks. That's because they have an uncommon bounding box. I will need to fix this.

I can't tell on that (they look fine?).

PR #2733, opened 2025-05-23
AGS 4: Support multiple textures in shaders (samplers)

ivan-mogilko:
This adds a support of having up to 3 secondary textures (samplers) to work with in shaders (so a total of 4 samplers, including primary input texture).

In script, added a new function to the ShaderInstance struct:
```
  /// Sets a secondary shader's input texture, using a sprite number. Only indexes 1-3 are supported.
  void ShaderInstance.SetTexture(int index, int sprite);
```

Note that you can only use indexes 1..3. Index 0 means primary texture, so it's not allowed here.
The `sprite` can be any sprite number, either regular or dynamicsprite, there's no difference.

Use in shaders:

1. In GLSL you must use hardcoded `sampler2D` uniforms `iTexture2`, `iTexture3` and `iTexture4`.
2. In HLSL names are not hardcoded per se (although you may use the same), but the registers matter: the `sampler2D` variables has to be appointed to registers `s0`, `s1` and so on.

<details><summary>GLSL example:</summary>

```
#version 130
uniform sampler2D iTexture;
uniform sampler2D iTexture2;
uniform vec2 iTextureDim;

varying vec2 vTexCoord;

void main()
{
    vec2 uv = vTexCoord.xy;
    if (floor(mod(uv.x * iTextureDim.x, 2)) == 0 && floor(mod(uv.y * iTextureDim.y, 2)) == 0)
        gl_FragColor = texture2D(iTexture, uv);
    else
        gl_FragColor = texture2D(iTexture2, uv);
}
```

</details>

<details><summary>HLSLexample:</summary>

```
// Pixel shader input structure
struct PS_INPUT
{
    float2 Texture    : TEXCOORD0;
};

// Pixel shader output structure
struct PS_OUTPUT
{
    float4 Color   : COLOR0;
};

sampler2D iTexture : register( s0 );
sampler2D iTexture2 : register( s1 );

const float2 iTextureDim:   register( c2 );

PS_OUTPUT main( in PS_INPUT In )
{
    PS_OUTPUT Out;                             //create an output pixel

    float2 uv = In.Texture;
    float4 outpixel;
    if (floor(fmod(uv.x * iTextureDim.x, 2)) == 0 && floor(fmod(uv.y * iTextureDim.y, 2)) == 0)
        outpixel = tex2D(iTexture, uv);
    else
        outpixel = tex2D(iTexture2, uv);

    Out.Color = outpixel;
    return Out;                                //return output pixel
}

```

</details>

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -3357,6 +3357,8 @@ builtin managed struct ShaderInstance {
   import void SetConstantF3(const string name, float x, float y, float z);
   /// Sets a shader's constant value as 4 floats
   import void SetConstantF4(const string name, float x, float y, float z, float w);
+  /// Sets a secondary shader's input texture, using a sprite number. Only indexes 1-3 are supported.
+  import void SetTexture(int index, int sprite);
 
   import readonly attribute ShaderProgram* Shader;
 };
--- a/Engine/ac/draw.cpp
+++ b/Engine/ac/draw.cpp
@@ -1056,9 +1056,9 @@ void clear_shared_texture(uint32_t sprite_id)
     texturecache.Dispose(sprite_id);
 }
 
-void texturecache_precache(uint32_t sprite_id)
+std::shared_ptr<Texture> texturecache_precache(uint32_t sprite_id)
 {
-    texturecache.GetOrLoad(sprite_id, nullptr, false);
+    return texturecache.GetOrLoad(sprite_id, nullptr, false);
 }
 
 void add_custom_shader(IGraphicShader *shader, uint32_t at_index)
--- a/Engine/ac/draw.h
+++ b/Engine/ac/draw.h
@@ -31,6 +31,7 @@ namespace AGS
     }
     namespace Engine
     {
+        struct Texture;
         class IDriverDependantBitmap;
         class IGraphicShader;
         class IShaderInstance;
@@ -108,7 +109,7 @@ void update_shared_texture(uint32_t sprite_id);
 // Remove a texture from cache
 void clear_shared_texture(uint32_t sprite_id);
 // Prepares a texture for the given sprite and stores in the cache
-void texturecache_precache(uint32_t sprite_id);
+std::shared_ptr<AGS::Engine::Texture> texturecache_precache(uint32_t sprite_id);
 
 // Adds a custom shader to the collection at the desired index;
 // if there's any shader on that index then it will get disposed of
--- a/Engine/ac/dynobj/scriptshader.cpp
+++ b/Engine/ac/dynobj/scriptshader.cpp
@@ -260,6 +260,13 @@ void ScriptShaderInstance::SetConstantData(uint32_t index, float value[4], uint3
     _constData[index].Val[3] = value[3];
 }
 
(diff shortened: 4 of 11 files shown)


ericoporto:
Does gall supports naming declarations like below?

```glsl
uniform sampler2D iTexture2..4
```

ivan-mogilko:
> Does glsl supports naming declarations like below?
> 
> ```glsl
> uniform sampler2D iTexture2..4
> ```
> 

Yes, that's how it is supposed to work.

> I also think the iTexture that are numbered could be numbered from 1-3 to match the indexes that are passed through the AGS script api.

I thought about this too, but then, would not it look strange if we had names like `iTexture`, `iTexture1`, ...?

ivan-mogilko:
I might rename iTexture to iTexture0, or try both during shader configuration. In HLSL these must use register index anyway.

> Is there some way to retrieve the dimensions like iTextureDim for the other textures in pixels

It may be possible to add these as a built-in variables and set only if a texture is set. I was not certain if that's necessary.

ericoporto:
https://github.com/user-attachments/assets/6a1cf83d-7458-4b51-85bc-69e52247e8d6

I solved my case by hardcoding in the shader

<details><summary>shader code</summary>

    uniform float iTime;
    uniform int iGameFrame;
    uniform sampler2D iTexture;
    uniform sampler2D iTexture2;
    uniform vec2 iTextureDim;
    uniform float iAlpha;
    
    varying vec2 vTexCoord;
    
    
    void main()
    {
        // this is both direction and speed, should be lower than 1.
        vec2 dir = vec2(0.25, 0.25);
    
        // width and height of the texture
        vec2 wh = iTextureDim;
        vec2 wh2 = vec2(120.0, 120.0);
        vec2 uv = vTexCoord;
        vec2 scroll = mod(uv * wh + dir * iTime * wh, wh2);
        vec2 uv_1 = vec2 (scroll.x / wh2.x , scroll.y / wh2.y);
        vec4 color = texture2D(iTexture, uv);
        if(color.w == 0.0)
          color = texture2D(iTexture2, uv_1);
        gl_FragColor = color;
    }

</details>

I think the case I can think I would need I could either hardcode or pass through SetConstant for now.

ivan-mogilko:
Added standard constants for all sampler dimensions.

Now the texture constants are called `iTexture0..3`, but `iTexture` is kept as an alias to `iTexture0`.
The texture dimensions are set in constants `iTextureDim0..3`, and `iTextureDim` is kept as an alias to `iTextureDim0`.

ivan-mogilko:
The pixel shaders cannot change *where* the pixels are drawn, they can only learn their relative position on *texture* and set their color. The position of pixels is defined by the drawn texture or the vertex shader, which is we do not let customize.

ericoporto:
That is alright, what I meant is it looks alright and I can't assess anything beyond saying it looks alright and works alright, but I can't print actual values, so in what I can see things here looks correct.

The case for dimensions for me was to draw a texture in a way it's tiled in another thing (that has a texture with its own dimensions), and it works afaict. In AGS Script I have been logging things to also read numbers, but here on shaders, since I can't see a way to print actual numbers, I can only check in a graphical way.

In summary, what I meant is looks alright, lets merge this.

PR #2737, opened 2025-05-26
AGS 4.0: Support ViewFrame offsets, make most properties settable in script

ivan-mogilko:
This adds ViewFrame.XOffset and YOffset properties in the Editor.
Curiously, ViewFrame always had "xoff/yoff" fields serialized in game data, but they were never used. I do not know the history behind these, maybe they were planned at some point but never implemented.

Support these offsets when the frames with these are applied to Character, Object, and Button animations. Character and Objects apply these offsets scaled along with the object scaling. Button offsets only the image displayed within the button control's rect, but *not* the control's position itself.

In script API added setters to previously existing properties: ViewFrame.Flipped and ViewFrame.Speed.
Added new properties ViewFrame.XOffset and YOffset.

### DIFF
--- a/Common/ac/view.h
+++ b/Common/ac/view.h
@@ -20,10 +20,12 @@
 namespace AGS { namespace Common { class Stream; } }
 using namespace AGS; // FIXME later
 
-struct ViewFrame {
+struct ViewFrame
+{
     int   pic = 0;
-    short xoffs = 0, yoffs = 0;
-    short speed = 0;
+    int16_t xoffs = 0;
+    int16_t yoffs = 0;
+    int16_t speed = 0;
     Common::SpriteTransformFlags flags = Common::kSprTf_None;
     int   sound = -1;  // play sound when this frame comes round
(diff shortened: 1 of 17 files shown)


ericoporto:
Huge ask, it’s ok if it’s not easily doable and can be skipped and perhaps a ticket of its own, but would it be possible to visualize the offset in the view viewer somehow? I guess it would require to pre-calculate the maximum width and height given offsets and width and heights of the entire loop to set where it would draw it.

ivan-mogilko:
I had to rewrite preview drawing for this, but now it accounts for the frame offsets too.

ericoporto:
This appears to work. I am not sure if it was like this or if this a consequence of this change, but if I resize the view pane of the preview by moving the edge of it, in a way the size is smaller, the view preview doesn't get smaller in a proportional way, but instead it gets squished - I added 15 px of offset in the frame below.

image

ericoporto:
Nice, now it scales proportionally! This looks great now!

PR #2745, opened 2025-06-08
4.0: Color finder displays color number as hex

### LINKED ISSUE #2743: AGS 4 Editor: Selecting color with colour finder shows incorrect value.
**Describe the bug**
When selecting a color with the sliders in the Colours Colour Finder panel, the number above is incorrect.

**AGS Version**
4.00.00.17

**Game**
New game from bass template

**To Reproduce**
See description

**Expected behavior**
Positive number

**Screenshots**

Image

 - OS: Win
 - Version 11

ivan-mogilko:
Fix #2743

Color Finder displays color numbers as a hexadecimal number, because it's inconvenient and confusing when the number is a negative integer.

Color properties allow to paste or type a hexadecimal number too now, along with the previously supported 3,4 - component string.

The only caveat here is that when pasting in script you must type "0x" part by hand. I am in doubts whether Color Finder has to print the number with "0x" or not. But technically there's no problem in doing that (then properties grid will have to account for possible "0x" prefix as well).

### DIFF
--- a/Editor/AGS.Editor/Panes/PaletteEditor.cs
+++ b/Editor/AGS.Editor/Panes/PaletteEditor.cs
@@ -21,6 +21,7 @@ namespace AGS.Editor
         private bool _noUpdates = false;
         private List<int> _selectedIndexes = new List<int>();
         private TabPage _colourFinder;
+        private Color _currentColor = Color.Black;
 
         public PaletteEditor()
         {
@@ -29,6 +30,7 @@ namespace AGS.Editor
             Factory.GUIController.OnPropertyObjectChanged += GUIController_OnPropertyObjectChanged;
             _selectedIndexes.Add(0);
             GameChanged();
+            UpdateNumberFromScrollBars();
         }
 
         protected override void OnPropertyChanged(string propertyName, object oldValue)
@@ -93,8 +95,11 @@ namespace AGS.Editor
             if (!_noUpdates)
             {
                 int newVal = 0;
-                Int32.TryParse(txtColourNumber.Text, out newVal);
-                if ((newVal < 0) || (newVal > int.MaxValue))
+                try
+                {
+                    newVal = Convert.ToInt32(txtColourNumber.Text, 16);
+                }
+                catch (Exception)
                 {
                     newVal = 0;
                 }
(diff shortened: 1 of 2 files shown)


AlanDrake:
Maybe just display both at once, this way one can paste from their graphic editor on the # format input, and quickly convert to 0x to copy in script.

ivan-mogilko:
> #112233FF (alpha is last, and could be omitted when full opaque)

Alpha goes FIRST in AGS, we've discusses this before. If alpha is last then you CAN'T omit it, because then the number will change.

> Maybe just display both at once, this way one can paste from their graphic editor on the # format input, and quickly convert to 0x to copy in script.

Please clarify, which both to display?

ivan-mogilko:
> I was thinking one might use the Color finder to paste the #RRGGBB code and convert the value to 0x, or viceversa.

So, your meaning is to have two input fields, one accepting #RRGGBBAA format, and another that outputs 0xAARRGGBB format for user to paste into scripts?

> The Color properties in property grid already support parsing "#RRGGBB", although it can only display a group of number,

Does it? I tried pasting such value, and it fails to parse it and reverts to 0 instead. When you say "display a group of number", do you mean it's wrong, and it should display something else?

ivan-mogilko:
> Ah, I was trying out from the Color finder property grid, but that one is different from other places where it fails to parse the webcolor string.

The Color Finder does not have any properties, but the editor displays the Palette properties while Color Finder tab is displayed.
That should be considered a mistake.

I did not know that it can parse different formats, haven't checked this yet, but maybe these Palette properties are using separate value converter, or have special handling. We will need to merge them with the color properties converter for consistency.

ivan-mogilko:
Very well, changing properties display is easy.

Looking at the Color Finder again, I am getting an impression that the sliders are unnecessary there. Perhaps they may be removed and a colored rectangle could bring a ColorPicker dialog on click instead.

AlanDrake:
The "Find colours..." button already summons the color picker.
I'm really not sure how to make it more useful, beside converting between 0x and #RGB, other stuff usually becomes redundant once you got a color picker window.

ivan-mogilko:
> Ah, I was trying out from the Color finder property grid, but that one is different from other places where it fails to parse the webcolor string.
> web color #RRGGBB(AA)

So, I tested how the Property Grid converts the #xx color, and turns out that it interprets it as #AARRGGBB, alpha first.

My initial intent was to only hotfix the color number in the "finder", so that it did not display negative number.
This is what I am going to do now, and I will leave everything else to those who have a better idea on how to display web colors and whatnot.

AlanDrake:
Hmm, I was wrong about dotnet color convert. As usual Microsoft has to play tricks with people's expectations. They make it able to recognize all colors, including named ones, only to screw up the alpha placement for #...

Issue #2739, opened 2025-06-01
Script compiler: detect string formatting mistakes


ivan-mogilko:
**Problem**

String formatting in AGS script is done in C-style, which is quite bug prone. The current script interpreter allows us to detect type mismatch at runtime, but that's specific to this interpreter implementation which records type info along with data (other implementation variants may not be able to do so), and not always possible (e.g. currently in case of values returned from plugin).

In any case, for the user it would be more convenient if formatting mistakes were detected at compile time when possible. Such feature won't give a 100% guarantee, but will help avoiding bugs in majority of cases.

**Suggested change**

Script compiler supports string formatting validation.

Whenever compiler meets a function that does string formatting, it parses the format string and runs over the arguments, matching arguments and format placeholders. It detects mismatches of two kinds:
- argument type is incompatible with the format;
- too many or not enough arguments;

In either of these cases compiler issues a compilation error.

**Implementation issues:*

1. Compiler will need to receive a list of functions that do string formatting. I see several existing variants of doing this:
    * Hardcode them and pass their list into compiler as an option.
    * Define them using a precompiler "pragma", let the precompiler parse these, make a list of them, which is then passed into compiler as an option.
    * Introduce a new script keyword which defines such function, similar to how "internalstring" defines a string type, and mark each such function. Let compiler parse these itself.

2. Compiler should assume that the formatting string is the last string argument just before the variadic arg list ("...").

3. The compile-time validation may only work if formatting string is passed as a string literal. If user passes a string variable, then it cannot be asserted.

fernewelten:
That makes sense. But of course, we can only check those function calls easily that pass a string literal for the format string. When the user codes, `cEgo.Say(mystring, 3.14);` then they are on their own.

We might mark the relevant functions with some marker where the `...` at the end of the parameter list goes, e.g., `void printf(string f, ... formated)`. 

Or we might put the marker where the format string goes, e.g., `void printf(format string f, ...);` 

In both cases, we can #define the marker to nothing for the benefit of the old compiler so that that compiler can continue to run as it is.

ivan-mogilko:
> Or we might put the marker where the format string goes, e.g., void printf(format string f, ...);

That is a good option, I like that.

fernewelten:
Okay, I'll have a go at it.

ivan-mogilko:
The compiler part is resolved by #2751. The new compiler now supports `__format` keyword used to define format string argument.

[...]

PR #2751, opened 2025-06-19
New compiler: Check arguments of format functions (`String.Format()` …)  at compile time

fernewelten:
This partially addresses #2739.

Introduce a new keyword `__format`. 

This can adorn one string parameter of a variadic function. This will declare the function, such as `Character.Say`, as a _format function_ and the adorned parameter as the _format parameter_ of that function.
```
managed struct Character {
    import void Say(__format string fmt, ...);
}
```

I'm envisioning that end users don't use the keyword, hence the leading `__`. It's very improbable that there is pre-existing code that has used `__format` in some (incompatible) way.

Whenever the compiler sees a call to a format function where the format parameter is passed a string literal, the compiler will check whether the variadic arguments of that call match the specifications in the format parameter.
```
cEgo.Say("Your backpack contains %s items", InvWindow.ItemCount); // ← Error, an 'int' is passed where a kind of string is expected
cEgo.Say("Don't forget your %10s, Red Riding Hood!") // ← Error, not enough variadic arguments passed to satisfy `%10s`

String msg = "Your backpack contains %s items";
cEgo.Say(msg, InvWindow.ItemCount); // ← Sorry, no error, compiler doesn't attempt to check format _variables_
```

**Note** This is the work on the compiler side. To make this work in practice, the system must also be modified:
– The AGS definitions must be changed to mark the respective format parameters of all relevant functions (`String.Format`etc.)
– A way must be provided to differentiate whether the 'old' or the 'new' compiler will compile the code.
– When the 'old' compiler (that doesn't know `__format`)  will run, it must be shown a line `#define __format`.

Nitty gritty details: 
– The code assumes that the format parameter describes the _variadic_ arguments. The first format specification pertains to the first _variadic_ argument, etc. So in a function `int f(__format string fmt, int i, ...)`, the parameter `i` is not considered to satisfy the first spec in the format parameter `fmt`.

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -1046,9 +1046,29 @@ void AGS::Parser::ParseVarname0(bool const accept_member_access, Symbol &structn
 
 void AGS::Parser::ParseFuncdecl_Parameters_Param(Symbol const name_of_func, bool const body_follows, bool const read_only, size_t const param_idx)
 {
+    auto &parameters = _sym[name_of_func].FunctionD->Parameters;
+    size_t const declared = _src.GetCursor();
+
+    bool const is_format_param = (kKW_Format == _src.PeekNext());
+    if (is_format_param)
+    {
+        SkipNextSymbol(_src, kKW_Format);
+        if (_sym[name_of_func].FunctionD->IsFormat)
+            UserError(
+                "Parameter #%u: Can only declare at most one parameter per function as '%s'",
+                param_idx,
+                _sym.GetName(kKW_Format).c_str());
+        _sym[name_of_func].FunctionD->IsFormat = true;
+    }
+
     Vartype param_vartype = ParseVartype(_src);
 
-    size_t const declared = _src.GetCursor();
+    if (is_format_param && !_sym.IsAnyStringVartype(param_vartype))
+        UserError(
+            "Expected a string type for the '%s' parameter #%u, found '%s' instead",
+            _sym.GetName(kKW_Format).c_str(),
+            param_idx,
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
I started testing this, and there appears to be a memory access error somewhere introduced by this change.
At first it was causing all kinds of seemingly random errors where compiler will find non-existing mistakes in unrelated places. But in the end the compiler crashed with "read access violation" exception, so that's likely the main culprit.

It's difficult to pinpoint the necessary combination of code, but this snippet triggers random compilation errors and sometimes "access violation" for me, if compiled in the full Editor (so including the script API header):

```ags
import void test(const string fmt);

function x()
{
    String.Format("%x", 10);
    String.Format("%d", 10);
}
```

Compiler/script2/cs_parser.cpp
```diff
@@ -3191,25 +3207,48 @@ void AGS::Parser::AccessData_FunctionCall_Arguments_Push(Symbol name_of_func, bo
...
                     UserError(
                         (ctfp + "Argument has type '%s' and doesn't match the format specifier '%s' for type 'float'").c_str(),
                         _sym.GetName(arg_vartype).c_str(),
                         format.c_str());
                 break;
 
+            case 'p':
+                if (!_sym.IsDynVartype(arg_vartype))
+                    UserError(
+                        (ctfp + "Argument has type '%s' and doesn't match the format specifier '%s' for dynamic types").c_str(),
```
ivan-mogilko, line 3244: There's a typo in error message, it checks 'p' but sais '%s'

ivan-mogilko:
Alright, this works, I shall merge this now, then add necessary changes to the script API header.

PR #2760, opened 2025-06-27
Editor: shutdown splash screen if app initialization failed

### LINKED ISSUE #2752: When the Editor fails to startup, app hangs at the splash screen indefinitely
**Describe the bug**
When the Editor fails to startup because one of its libraries did not load (common case is when AGS.Native did not load because required VC Redist is not installed on the system) the application does not shut down properly, but remains frozen at the splash screen.

**AGS Version**
Any version starting with 3.0 (i think).

**To Reproduce**
1. Go to the Editor's program folder and move / rename AGS.Native.dll to something else.
2. Run AGSEditor.exe.
3. Observe the error message, and press OK.
4. Observe splash screen hanging indefinitely.

**Expected behavior**
Editor should shut down properly in case when it could not startup.

If possible, there have to be user friendly error message telling which library could not load, and what the user may try to fix this. Because right now it just shows .NET exception text.

ivan-mogilko:
Fix #2752

This does a number of changes. First I fixed the splash problem, but then also adjusted how the CommandLineOptions are handled, because I thought that application exceptions need to be also printed in console if the editor is launched from cmd line with `/compile` and similar commands. So with this change the StdConsoleWriter will be enabled earlier.

IMHO it's reasonable for the CommandLineOptions to only parse args and declare options, and it should be up to the application (GUIController or ApplicationController in this case) to decide how to handle them, so I moved enabling of the StdConsoleWriter to the ApplicationController code, and bad parameters reports (missing project path, or missing game file) to GUIController instead.
The actions are still executed by GUIController as before, I did not change anything to those.

More details are in commit description.

How to test the Splash issue: remove/rename AGS.Native.dll (for example).


ericoporto:
It stopped happening now, but I didn't do anything special afaict.

PR #2763, opened 2025-06-30
Editor: Welcome pane links to AGS 4 help topics

ericoporto:
Very minimal change, the idea here is just to have links to the proper AGS 4.0 help topic in the Welcome Pane.

- [What's new in AGS 4.0](https://github.com/adventuregamestudio/ags-manual-source/blob/ags4/WhatsNewIn40.md)
- [Upgrading to AGS 4.0](https://github.com/adventuregamestudio/ags-manual-source/blob/ags4/UpgradingTo40.md)

image

Because we are already building from the ags4 manual source branch and getting it here from the CI, this should just work.

### DIFF
--- a/Editor/AGS.Editor/Panes/WelcomePane.Designer.cs
+++ b/Editor/AGS.Editor/Panes/WelcomePane.Designer.cs
@@ -170,7 +170,7 @@ namespace AGS.Editor
             this.pnlRight.Margin = new System.Windows.Forms.Padding(10);
             this.pnlRight.MinimumSize = new System.Drawing.Size(443, 362);
             this.pnlRight.Name = "pnlRight";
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
I have a related question, I can see that the ags4 manual is built as ags4-help.chm. Should we perhaps use this name in ags4 editor, for convenience?

ericoporto:
I rename it when downloading in the CI, the idea was to not need to do anything here.

I could alternatively name it as ags-help.chm and store it zipped so instead of renaming it would be unzipped.

If the convenience is about local builds the manual name isn't different when building locally, this was only done in CI of ags-manual repository.

ericoporto:
The advantage of moving to ags4-help.chm name would be to have both in the build dir of VS, so it would make it possible to test help issues without when switching between ags4 and "ags3" branches.

PR #2765, opened 2025-06-30
Agspak: merge with agsunpak, add "list" and "export" commands

ivan-mogilko:
1. Merges agspak and agsunpak into a single utility. Unpacking is done with '-u' or '--unpack' command argument, e.g.:
```agspak -u template.agt export_dir```
2. Add List command ('-l', '--list') which prints library contents, e.g.:
```agspak -l template.agt```
3. Add Export command ('-e', '--export') which lets unpack selected assets; assets are provided as a comma-separated list of filenames e.g.:
```agspak -e template.agt export_dir "agsfnt0.wfn, agsfnt1.wfn, agsfnt2.wfn"```


ivan-mogilko:
I put this to draft, because I am not myself convinced of the chosen way to pass a "operation" command. With this syntax they are easily mixed with options, and also reduce the possible variants that options may use.

One way to resolve this is to not use `-N` syntax, but something else like simply 'N' syntax for a "operation".
There's remaining problem that the "packing"goes without explicit operation, and so this will introduce a risk for misinterpreting a command argument, for example, if a file's name will match a "operation" name (unlikely, but still).

ivan-mogilko:
After looking in the [tar description](https://man.archlinux.org/man/tar.1), what about '-c' for "create" a pack?

I would also like to add commands for adding (or "importing") and removing (or "deleting") distinct files in pack.

ericoporto:
> what about '-c' for "create" a pack?

That seems fine.

> I would also like to add commands for adding (or "importing") and removing (or "deleting") distinct files in pack

This doesn't seem useful for building but it could be useful for poking or patching a file.

Just in the random ideas, but no need to do this here. One cool thing could be to list all files in a pack ordered by size, with their size in kB and percentage of the total size of the pack. Other idea could be a find command that would return if the file is in the pack and all occurrences of it - since the "dirs" in the pack allow multiple files inside with the same name in different folders. I don't have a ready use for these it was just ideas.

ivan-mogilko:
> I imagine that in reality the command here is

No, actually, the tool seeks for the pattern file in the given "work directory". IIRC that's how it worked before?

ericoporto:
Yeah, I agree. I think this can be merged as is, the provided commands seem fine.

PR #2767, opened 2025-07-02
Engine and Editor: Add support to set character spacing in fonts

### LINKED ISSUE #2766: Adjust Kerning/Character Spacing for Fonts?
**Describe the problem**
Currently, there is no way to manually adjust the spacing between characters for fonts.

**Suggested change**
Add the ability to set a character spacing in the font properties so that characters can render more spaced out the default.

edmundito:
## Overview

Resolves #2766 

This change enables game authors to set character spacing in fonts. Supports both TTF and WFN fonts, as well as outlines (for WFN fonts, the spacing must match the outline). I did this in master because I have someone interested in using this for a project this year, but they are not ready to take the risk with 4.0.x 

## Screen Shots
TTF
Screenshot 2025-07-01 221150

WFN
Screenshot 2025-07-01 221323

Editor
image

## TODO
- [ ] "Character Spacing" or "Letter Spacing" (to avoid term overloading?)
- [ ] Test with RTL fonts, and CJK fonts
- [ ] Changes.txt

### DIFF
--- a/Common/ac/game_version.h
+++ b/Common/ac/game_version.h
@@ -124,7 +124,7 @@ OPT_SAVESCREENSHOTLAYER, CHF_TURNWHENFACE. Button's WrapText and padding.
 Few minor behavior changes.
 3.6.3:
 OPT_GUICONTROLMOUSEBUT
-
+Font.CharacterSpacing
 */
 
 enum GameDataVersion
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -100,8 +100,6 @@ void GameSetupStruct::read_font_infos(Common::Stream *in, GameDataVersion data_v
             fonts[i].Outline = in->ReadInt32();
             fonts[i].YOffset = in->ReadInt32();
             fonts[i].LineSpacing = std::max(0, in->ReadInt32());
-            // Character spacing added in a future version, use 0 for older games
-            fonts[i].CharacterSpacing = (data_ver >= kGameVersion_Current) ? in->ReadInt32() : 0;
             AdjustFontInfoUsingFlags(fonts[i], flags);
         }
     }
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -100,6 +100,8 @@ void GameSetupStruct::read_font_infos(Common::Stream *in, GameDataVersion data_v
             fonts[i].Outline = in->ReadInt32();
             fonts[i].YOffset = in->ReadInt32();
             fonts[i].LineSpacing = std::max(0, in->ReadInt32());
+            // Character spacing added in a future version, use 0 for older games
+            fonts[i].CharacterSpacing = (data_ver >= kGameVersion_Current) ? in->ReadInt32() : 0;
             AdjustFontInfoUsingFlags(fonts[i], flags);
         }
     }
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -310,6 +310,8 @@ struct FontInfo
     int           YOffset;
     // Custom line spacing between two lines of text (0 = use font height)
     int           LineSpacing;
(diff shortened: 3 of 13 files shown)


edmundito:
> Could you elaborate on why do you think this should be done by the engine, and not by e.g. finding another font or modifying existing one?

I need to investigate more about this and talk to the game author. The change was trivial so I figured I push this for now and discuss/review its feasibility.

edmundito:
> We had a plan to move to SDL_ttf library eventually, and I'd like to know if it has the same ability, because if it does not, then this will increase compatibility problems that we will have to solve.

Ah was not aware of this plan and I would also be concerned that 3.6.x and 4.x would diverge in features. I did some basic research on SDL_font after your comment and looks like the closest solution may be to render each char separately which I am not even sure about the perf impact.

ivan-mogilko:
> I did some basic research on SDL_font after your comment and looks like the closest solution may be to render each glyph separately which I am not even sure about the performance impact.

Performance is one thing, but there's also a issue of multi-glyph sequences (ligatures) that have to be rendered together in certain way. There's a full extra library that does this in addition to SDL_ttf itself.

I will try asking the SDL_ttf developers if character spacing may be hypothetically added in the future, out of curiosity.

ericoporto:
> There's a full extra library that does this in addition to SDL_ttf itself.

A few years ago I made a version of the engine which included this library, Harfbuzz, with SDL_ttf and newer freetype over the at the time unicode translation support and put in the forums here

https://www.adventuregamestudio.co.uk/forums/engine-development/ags-high-priority-improvement-needs/msg636637035/#msg636637035

Unfortunately, compatibility was added to engine to work with some broke fonts (I would much prefer if the game authors would just fix their fonts instead...). This required support for things not in SDL_ttf which made it pretty hard for me to go on, and well, SDL_ttf project moves at glacial pace... The ligature library beyond persian and other ligature based languages also gives support for complex emojis. (note natural persian would still need BIDI https://github.com/libsdl-org/SDL_ttf/issues/135, but unfortunately I didn't understood how to implement in SDL_ttf).

The new PR in https://github.com/adventuregamestudio/ags/pull/2711 is in a way a newer version of this, with a modified SDL_ttf required for it. There is a PR upstream to SDL_ttf (https://github.com/libsdl-org/SDL_ttf/pull/538), to give this addition to support these old issues. This is necessary, people will still build AGS with upstream libraries anyway and open bug reports later (see #1528 and #2768 as examples).

So, I believe at least an issue should be opened upstream in SDL_ttf for this. Additionally, in AGS 4, font files and font instances are separate. Here the spacing is an absolute number, which should make things easier, but I am unsure if people who use will use it like this or expect that it resizes when the text changes size.

ivan-mogilko:
@edmundito , given there's a way to insert this functionality into SDL_ttf, which I found earlier (see my previous comment), I suppose that we may accept this as a engine feature. If we move to SDL_ttf in ags4, there are options to:
* suggest the feature to the upstream version of library;
* maintain our own SDL_ttf fork with couple of extra necessary features (if lib authors won't agree to add these). We may introduce a "hint" macro, or use a runtime check method to know which library the engine is linking to, in case someone wants to use an official version installed on their system.
* temporarily disable this feature (in any case).

What needs to be done to this PR in order to merge it now (I've mentioned some of these in one of the [earlier replies](https://github.com/adventuregamestudio/ags/pull/2767#issuecomment-3033951170)):
* some earlier commits squashed, because they contain incorrect serialization method (i think you may redo commits).
* rebase to the latest master branch (there seem to be code conflicts).
* do not scale CharacterSpacing with SizeMultiplier (afaik you only do that in WFN renderer): that's not consistent with other settings such as LineSpacing or YOffset, and that would let more precision when selecting spacing for the multiplied font sizes.
* for serialization: actually use existing "v360_fonts" extension to save CharacterSpacing, as it has few reserved integers which are 0 by default. This will save 1 extra extension.

ivan-mogilko:
Superceded by #2813

PR #2768, opened 2025-07-02
add support for building against more system libraries

rathann:
This enables building against system glm, tinyxml2, miniz and freetype.

Additionally, switch to using system OpenAL instead of embedded mojoAL on Linux.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -73,7 +73,7 @@ build_linux_cmake_task:
     arch=$(dpkg --print-architecture)
     mkdir build_debug_$arch && cd build_debug_$arch
     cmake .. \
-      -DAGS_USE_LOCAL_ALL_LIBRARIES=1 \
+      -DAGS_USE_LOCAL_ALL_LIBRARIES=0 \
       -DAGS_USE_LOCAL_SDL2_SOUND=0 \
       -DCMAKE_BUILD_TYPE=Debug \
       -DAGS_BUILD_TOOLS=1 \


ericoporto:
The CMake is missing clearly adding the variables.

- they should be added here: https://github.com/adventuregamestudio/ags/blob/e2ae947f95fa5aa849105a419c9cb76276d69320/CMakeLists.txt#L36-L42
- they should additionally be documented here: https://github.com/adventuregamestudio/ags/blob/master/CMAKE.md#using-locally-installed-libraries

Our linux builds cannot use these changes by default since they are meant to be self contained to be easy to ship as a binary to be available in game stores - they are not meant to link to system libraries. Any change related to this would just be as alternative and not supported - and cannot be the default. So these should all be optional and the code has to be made sure that the defaults are kept as they are now.

System freetype is not supported on AGS3 (master branch). Fonts will look weird and we will get hard to follow on bug reports. It may be possible in AGS4 (after #2711).

The miniz find should just rely on PkgConfig instead of having hardcoded paths. I guess these changes are specific changes for the Fedora open source distribution of the AGS, but other systems may have different expectations. We do have currently wrongly set local hardcoded paths for other systems (Windows and macOS), but that's just because these are configurations that we don't use normally (untested by the CI) and should be adjusted too, just didn't had time to do it yet.

rathann:
> The CMake is missing clearly adding the variables.
> 
>     * they should be added here: https://github.com/adventuregamestudio/ags/blob/e2ae947f95fa5aa849105a419c9cb76276d69320/CMakeLists.txt#L36-L42
> 
>     * they should additionally be documented here: https://github.com/adventuregamestudio/ags/blob/master/CMAKE.md#using-locally-installed-libraries

Will do, thanks for the pointers.

> Our linux builds cannot use these changes by default since they are meant to be self contained to be easy to ship as a binary to be available in game stores - they are not meant to link to system libraries. Any change related to this would just be as alternative and not supported - and cannot be the default. So these should all be optional and the code has to be made sure that the defaults are kept as they are now.

Understood.

> System freetype is not supported on AGS3 (master branch). Fonts will look weird and we will get hard to follow on bug reports. It may be possible in AGS4 (after #2711).

Ok, I'll consider dropping that part. I think the main reason for using system freetype, apart from general mandate to use system libraries, was that I could get the bundled freetype-2.1.3 to compile.

> The miniz find should just rely on PkgConfig instead of having hardcoded paths.

Interesting. It looks like the Fedora package is missing the `.pc` file. I'll get that fixed in Fedora and change that part accordingly.

> I guess these changes are specific changes for the Fedora open source distribution of the AGS, but other systems may have different expectations.

Yes, the patch is what I'm carrying downstream in Fedora.

> We do have currently wrongly set local hardcoded paths for other systems (Windows and macOS), but that's just because these are configurations that we don't use normally (untested by the CI) and should be adjusted too, just didn't had time to do it yet.

I'm not sure what you're asking of me above.

Thanks for the review!

ericoporto:
> was that I could get the bundled freetype-2.1.3 to compile

Uhm, do you mean you can't get it to build? I could spin a VM to figure it out, can you tell me the version of Fedora I should try?

> I'm not sure what you're asking of me above.

Sorry, just talking with myself, it's not for you. I am not always on the computer so sometimes I just write things out so eventually when I am on the computer I can pick it up and work things out.

Any trouble just ping me here, and also if you get part of it working but not everything you want, just comment and I can merge and workout the rest. :)

rathann:
Now with working support for miniz upstream CMake files (they have `add_library(MiniZ::MiniZ ALIAS miniz::miniz)`)

ericoporto:
I think you can just turn that off and I will later adjust. I used local libraries in that mostly because of the particular fetch that occasionally caused error when downloading when there were many CI triggers (for some reason this is not a problem in GitHub Actions, but that CI run uses Cirrus CI which uses Google Cloud Community instances which have some finicky issues).

You can explicitly set it to zero in the line below and I will later fix.

https://github.com/adventuregamestudio/ags/blob/076dc50398d39b41d86fcdbf947eaad8c8afcac6/.cirrus.yml#L76

I see you are switching to default using OpenAL on Linux but I would like to keep the default on mojoAL.

We use mojoAL which uses SDL in the backend so we can relay settings to be able to switch the backend through all the ones supported by it. Usually in desktop platforms having this flexibility of switching the backends is desired by users.

In the past, the Android port worked alright with it's OpenAL so I kept it, and the Emscripten port used to just work with the OpenAL.js in it while the mojoAL at the time used to require user action at a specific time to actually play sounds, this may have been fixed since but I haven't checked.

If you can drop the commit I can readjust the scripts later, or alternatively, have a explicit use local OpenAL variable, which I would ignore in iOS, Android and Emscripten.

rathann:
[...]
> You can explicitly set it to zero in the line below and I will later fix.
> 
> https://github.com/adventuregamestudio/ags/blob/076dc50398d39b41d86fcdbf947eaad8c8afcac6/.cirrus.yml#L76

Done.

> I see you are switching to default using OpenAL on Linux but I would like to keep the default on mojoAL.
[...]
> If you can drop the commit I can readjust the scripts later

Done.

rathann:
[...]
> We use mojoAL which uses SDL in the backend so we can relay settings to be able to switch the backend through all the ones supported by it. Usually in desktop platforms having this flexibility of switching the backends is desired by users.

Does that mean you can switch audio backend at runtime? openal-soft supports all the usual backends, too. Most desktop users on Fedora will have it installed, so avoiding it has little benefit.

By the way, mojoAL has moved to SDL3 since https://github.com/icculus/mojoAL/commit/ba0932742f9bf22041f43278127bd6f9d6ceb8d5 . Are there any plans to switch AGS to SDL3, too?

PR #2769, opened 2025-07-03
Tools: add include_utils test

ericoporto:
this adds tests for include_utils (and some way to test tools related code). This ports some of the C# tests for the cpp version of it.

This is a draft because I have reached a roadblock I can't figure it out yet.

Screenshot 2025-07-02 at 22 10 16

The `"exclude_me_anyway.dat"` from the patterns list, will be translated to the string `"(^|/)exclude\\_me\\_anyway\\.dat($|/)"`, which will then give me the error below:

```
unknown file: Failure
C++ exception with description "The expression contained an invalid escaped character, or a trailing escape." thrown in the test body.
```

The error also only happens in macOS.


ivan-mogilko:
Could you also create a Tools.Test project and add this to MSVS Tests solution?

ericoporto:
I don't have a VS2019 on the computer I was using now. I will boot one of my old ones that still has it so I can create these.

ericoporto:
I added in both, I couldn't think why it wouldn't be a good idea, but we can remove from one of them later.

ericoporto:
I went back on my idea, as I understand by having it in the Test solution it will build when we build it. If it becomes too weird we can go back and switch to the Tools.sln.

PR #2770, opened 2025-07-03
SKETCH: a build generator for ags command line tools

ericoporto:
this is just a sketch, the idea is to output a makefile or ninja-build that can be run with make/ninja and build an ags game using the command line tools and the game Game.agf project file. My goal for ninja is it is a nicely self contained executable that could be shipped easily, but I guess some people may prefer make so I thought to come up with some general "architecture" with the idea that there could be some code reuse but this didn't happen for real yet.

Note this can't actually build a game yet since some necessary tools don't exist yet.

Consider the code a huge sketch. Most of the comments are from my thought process. 

We can decide we don't want this, but writing this helped me understand all the necessary steps to build a game with command line tools and exercise what is written in the wiki (see https://github.com/adventuregamestudio/ags/wiki/AGS-Game-Build-process-(3.5.*)).

edit: one of the references was a little build script here: https://github.com/adventuregamestudio/ags/pull/1370#issuecomment-887920148


Tools/agfbuildgen/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+    std::sort(rooms.begin(), rooms.end());
+    for (const auto &r: rooms)
+        opt.RoomFileList.push_back(String::FromFormat("room%d.crm", r));
+
+    opt.HasDialogScripts = false;
+}
+
+bool tool_exists_in_path(const char* tool_path)
+{
+    String tool_exe = String::FromFormat("%s.exe", tool_path);
```
mgerhardy, line 251: this would be windows only.

Tools/agfbuildgen/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+    opt.OutputDir = ExistingPathAsAbsolute(parsedOptions.OutputDir); // we never mkdir this
+    opt.GameProjectFile = ExistingPathAsAbsolute(parsedOptions.GameAgf);
+    opt.GameProjectDir = Path::GetDirectoryPath(opt.GameProjectFile);
+    opt.TempDir = Path::ConcatPaths(opt.GameProjectDir, "temp"); // need to mkdir somewhere...
+    // we don't delete the temp after or we will have to rebuild the dependencies everytime.
+    // maybe call this a different name?
+
+    // all the entries below should be the default full paths
+    // but I don't know any easy way to get them...
+    // ideally it would look in the path THIS executable is (assuming there is a Tools dir)
```
ivan-mogilko: We have this in Directory::GetCurrentDirectory()

ericoporto: It gets the working dir and not the directory of this executable.

I am imagining by default all tools would be in the same directory, like a Tools or Tools/bin directory .

I assume the working directory is simply the path from where this tool is invoked?

ivan-mogilko: Oh right, I see, i got mistaken.

I think that the first argument in command line (argv[0]) should be the executable path, but I can't remember if it's given always as a absolute path or not. There should be a way to get the current process's executable location though, in both Windows and Unix systems that we support.

Tools/agfbuildgen/main.cpp
ericoporto: It just checks for either the name without exe or with exe, it is correct.

ericoporto:
Did a lot of work here, still a lot to go too, but I managed to fix many things

Tools/agfbuildgen/main.cpp
ericoporto: I added a function for this here

https://github.com/adventuregamestudio/ags/pull/2770/changes/f0661ab24d548f46b9b93dc6402ffd74ca44399c

Because I was a in a hurry I just stole it from someone else that had one for the same platforms than us, did some adjustments for AGS codebase and then used an Codex AI to fix remaining issues with it

ivan-mogilko: This should be in Common/utils/directory.cpp, I suppose.

ericoporto: Maybe, yeah. I was thinking this approach of a tools dir could be useful to your version of agsbuild too, using the build orchestration tool directory as a default place to look for other tools.

ericoporto:
Hey, I’m going to favor your approach merged in #3057 , but I will keep mine here for a little because I still need to read things more carefully there and there are some notes I made in my code here that I think will also apply to your approach.

PR #2772, opened 2025-07-04
ags4: new compiler solutions with x64 target, and ci builds and test runs

ericoporto:
I just noticed that the new compiler project didn't yet include x64 targets. It has been built and run under x64 bit in test under the CMake builds - in the CI these are run under GitHub Actions currently.


ivan-mogilko:
shwlapi.lib is required by stdio_compat.cpp (win api path-related functions).

I noticed that your PR is building, does that mean that you've solved the problem?

ivan-mogilko:
Hm, I tried the AGS Editor in ags4 branch, and it builds so long as I place necessary dlls in References/x64 folder, namely:
- irrKlang.NET4.dll
- ikpMP3.dll

> So it would be required by all builds of common lib and anything that use it directly - perhaps some tools. If something only links to common.lib, then it wouldn't require because Common already linked to it ?

I suppose so.

ericoporto:
Wait, AGS.Native is building in your computer? I guess I need to clean up my project in VS and see if it builds then.

ericoporto:
ok, deleted all vs related stuff and now I can build the AGS Editor fine in all configs afaict.

Issue #2729, opened 2025-05-12
Engine: Customize developer and support information for crash alert dialog


edmundito:
**Describe the problem**
Despite all the testing, games will have errors that will crash, and the developer must know about them. Currently, the way the message displays to the user is very AGS-specific:

![Image](https://github.com/user-attachments/assets/583b9dd0-1d23-4f21-850a-61ce62f728ec)

* Title is Adventure Game Studio
* "Contact the author for support or post in the *AGS Technical Forum"
 
**Suggested change**
The way I would solve it would be to have:

* Use the existing game metadata set in the game settings.
* Add a custom field for a support URL

> **Old Skies**
> 
> An internal error has occurred. Please note down
> the following information. If the problem exists, please
> [contact Wadjet Eye Games](https://wadjeteyegames.com/support) for support.
>
> Error: Lorem ipsum
> Game Version: 1.0.5
> Engine Version: 3.6.1.30

ivan-mogilko:
This is a duplicate of #1232, but that was never resolved and buried in the list of issues now, so we may leave this now.

What has to be done here:
1. Decide how the message is constructed, which kind of parameters (meta data) would be required for the message, and what functionality does the message dialog has (like a clickable link).
2. Actually save this meta data in the compiled game, because right now it's only stored in the project source. Meaning we need another block of data appended to game format.
3. Decide what message to display instead if this meta data is not available.
4. Devise the way of displaying the message. We have multiple supported platforms. There's apparently a method in SDL2 api that displays a system message dialog; but there's another issue that I mentioned in #1232: message is displayed in a simple message box as a single label, and there's no way to customize by adding a link to author's website.

ivan-mogilko:
Another thing (which I also mentioned in #1232) is that the engine lacks a error type for the bad saves format.
In short, there are "internal errors" and "script errors", where the "script error" adds a note that this error might be a problem with the game scripts. But the save error is neither. So we need another error type, expressed differently in this message.

PR #2773, opened 2025-07-05
Add game description properties into the compiled game, available in script

ivan-mogilko:
This is an intermediate step for the #2729, but also may be used e.g. in savegame upgrading process, and perhaps for other things.

Editor writes the "game information" properties as a bundle into the compiled game, as a single extension.
Added following properties in script:
```
  /// Gets the game's description
  static readonly String Game.Description;
  /// Gets the game's developer's name
  static readonly String Game.DeveloperName;
  /// Gets the game's developer's URL string
  static readonly String Game.DeveloperURL;
  /// Gets the game's genre description
  static readonly String Game.Genre;
  /// Gets the game's release date as a DateTime instance
  static readonly DateTime* Game.ReleaseDate;
  /// Gets the game's release version, represented as "X.Y.Z.W" string
  static readonly String Game.Version;
```

### DIFF
--- a/Common/ac/gamesetupstruct.h
+++ b/Common/ac/gamesetupstruct.h
@@ -83,6 +83,9 @@ struct GameSetupStruct : public GameSetupStructBase
     int               numGameChannels = 0;
     // backward-compatible channel limit that may be exported to script and reserved by audiotypes
     int               numCompatGameChannels = 0;
+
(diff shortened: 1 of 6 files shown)


ericoporto:
Once this made into a released version, we could update the about of one of the templates to use one of these to point out those exist,

ericoporto:
Uhm... Being a separate struct makes it easier to check all of it's properties.

PR #2775, opened 2025-07-07
Upgrade Googletest submodule

ericoporto:
This upgrades the Googletest submodule to the latest commit.


ivan-mogilko:
I really do not want the project to be part in C++14 and half in C++17. Either whole project should be updated to use C++17, or we should use a version of the library that does not require it.

ericoporto:
I will close this and do the other change to 1.8.1 release.

PR #2777, opened 2025-07-07
AGS 4: reimplement Character waiting 1 frame before moving to let animation advance

### LINKED ISSUE #2744: 4.0: proper fix for character going to animation frame 1 before walking
In AGS it seems that the characters are making 1 frame advance in walking animation before actually moving.
In the past this was a consequence of MoveList changing its progress *after* returning current position.
I've changed that MoveList logic when rewriting it in #2685, so this behavior was broken unintentionally.
#2731 was a hotfix, but it hacked the MoveList logic exclusively for the character walk, in a way that makes it inconsistent, where "current position" does not correspond to "progress". So this cannot stay long-term and will have to be replaced with another solution.
Ideally, MoveList should be always internally consistent, where you can calculate current pos from progress (which is onpart variable in the current implementation) and vice versa. Any special behavior should rather be performed by an external code, like character's walking update logic.

I did only a quick look before, but it seems possible to achieve using one of the "waitxxx" variables in character struct, the one that works as a walking timer. Presumably, if we set it to 1 when starting to walk (as opposed to 0), then it will let animation advance once before motion begins.

ivan-mogilko:
Fix #2744

This reimplements a default AGS behavior where character will keep at the previous position for 1 frame more before starting to move along the path.
Previously this worked because MoveList returned a *previous* position when updating. It is now returning a *new* position after a MoveList refactor and a introduction of MotionPath api. Therefore the "animation first" behavior has to be implemented in another way.
IMHO it's reasonable to not have this logic inside MoveList, as that makes MoveList having inconsistent state; instead this logic shall be a part of the Character which decides when to animate and when to move along the path.
Users of the MotionPath api can do according to their desire in script.

@AlanDrake , please try this and tell if this seems working to you. I checked that it does 1 walk frame before starting to move, and right after turning, but I could have missed something.

### DIFF
--- a/Common/ac/characterinfo.h
+++ b/Common/ac/characterinfo.h
@@ -142,7 +142,8 @@ struct CharacterInfo
     int16_t walkspeed_y = 0;
     int16_t pic_yoffs   = 0; // this is fixed in screen coordinates
     int     z           = 0; // z-location, for flying etc
-    int     walkwait    = 0;
+    int     walkwait    = 0; // number of frames to wait before advancing a move;
+                             // also used as a turning counter
     int16_t speech_anim_speed = 0;
     int16_t idle_anim_speed = 0;
     int16_t blocking_width = 0;
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -1972,27 +1972,25 @@ void move_character_impl(CharacterInfo *chin, const std::vector<Point> *path, in
         return;
     }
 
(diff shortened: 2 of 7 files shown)


ivan-mogilko:
Adjusting "walk_wait" in this case won't work, because it will cause the "walking on same spot" effect.

But I noticed that disabling copying a "step fraction" solves this issue, while not causing any visible negative effects. That's strange, I need to investigate more.

Issue #1596, opened 2022-04-01
AGS 4: The project upgrade to open rooms may leave game in the inconsistent state


ivan-mogilko:
**Problem**

If the user opens an older project in ags4 editor, the editor will automatically convert rooms to the open format. This includes *moving room scripts* to the Rooms/n/ folders. However, the project file itself won't be saved automatically.
If the user then will close the editor without saving, the project file will remain with older version index in it. Hence the project awhole will enter an inconsistent state, where the index tells one version, but the file structure belongs to another.

Note that if user opens same project second time, ags4 editor will try to upgrade again, but won't find room scripts this time, because they were already moved earlier. The previously created "Rooms" folder will also be copied to "backup".

**Steps to reproduce**

1. Create a game in a previous version of AGS (3.5.0 - 3.6.0).
2. Make sure there's some code into the room script(s).
3. Load the project into ags4 editor, wait until the conversion is complete. 
4. Exit the editor, refuse to save the game.
5. Load the project into ags4 editor again; try to open any room script, notice it is blank.

**Suggested solution**

Because conversion to AGS 4 will likely require lots of major changes (open rooms is only one of them), IMO it's best to act similar to importing 2.* games into the 3.* editor, that is -
1. Before doing any conversion, warn user about major changes to the project, and suggest an automatic backup.
2. If they agreed, copy full project (or at least files that will be affected by conversion) into the separate folder (e.g. "Backup").
3. Do the conversion, and *save the project file* with the new version index.

PR #2778, opened 2025-07-07
AGS 4: implement Room upgrading process

ivan-mogilko:
A while ago I've been trying to fix certain historical issue in ags4, but had to revert, as there appeared to be no room upgrade process in AGS 4, similar to the one existing for the game project awhole. This PR resolves that. But it does bit more.

1. Changed the project XML back to using a X.Y.Z.W string version instead of the decimal "index". Apparently CJ thought that index is easier to compare, but it's also more difficult to maintain (even though we switched to XXYYZZWW encoding around 3.6.0). String versions are arguably easier to have, and you may compare them by converting to the System.Version object.
2. I changed Game.SavedXmlVersion property to a Version object. It's a public member, so I wonder if that may have any consequences with editor plugins. But given this is ags4, plugins will likely have to be updated anyway.
3. Modified Room's XML version accordingly.

Finally, added the Room upgrade process. Currently it does nothing (an empty method), but things may be added there in time.
The upgrade is run in two cases:
1. When you open a new game, editor may run a "all rooms upgrade" in a similar style to how the conversion from crm to open format is done. This may be required if the new editor version requires all rooms to have a fix unconditionally. Currently this is never done, but I left a place in the code with a comment where one may set a variable signalling that this process should be performed.
2. Whenever you load an individual room.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -55,21 +55,18 @@ namespace AGS.Editor
         public const string SPRITE_FILE_NAME = "acsprset.spr";
         public const string SPRITE_INDEX_FILE_NAME = "sprindex.dat";
 
-        /* 
-         * LATEST_XML_VERSION is the last version of the Editor that used 4-point-4-number string
-         * to identify the version of AGS that saved game project.
-         * DO NOT MODIFY THIS CONSTANT UNLESS YOU REALLY WANT TO CHANGE THE IDENTIFICATION METHOD.
-        */
-        public const string LATEST_XML_VERSION = "3.0.3.2";
-
         /*
-         * LATEST_XML_VERSION_INDEX is the current project XML version.
+         * LATEST_XML_VERSION is the X.Y.Z.W string which defines current project format.
          * DO increase this number for every new public release that introduces a new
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
But I came to a realization that this "all rooms upgrade" process is not done right. It has the same problem as the room conversion from binary format (see #1596).

In other words, for a correct behavior, the rooms should not be upgraded at least until the user explicitly saves the project first time after loading.

The reason why I added this is because the Editor won't know if the room has to be upgraded until it's loaded. Which may leave the project in a partially non-upgraded state.

Perhaps there are other ways to do this. I think there was a trick to mark the project as "Must Rebuild All" (or whatever that setting was called), which would trigger all rooms rebuild, which in turn would require all rooms to be loaded, and when each is loaded it's also get upgraded. Maybe that's what I do instead.

ivan-mogilko:
Also added current room data version to the Game.agf, as an attribute in "Rooms" node.
This is used to quickly check which room versions are expected to be found with the loaded game.

ivan-mogilko:
Just a small note: I'd like to release another 4.0 alpha update without too big changes, because there have been a number of important fixes in it. So I will postpone merging this until then (and #2779 too).

PR #2779, opened 2025-07-08
AGS 4: Fix historical 1-pixel mistake of character and object draw position

ivan-mogilko:
WARNING: applied over #2778 temporarily, because this requires to upgrade the rooms.

Some time ago a user brought my attention to this annoying fact: it appears that AGS always have had a math mistake when calculating draw position for characters and objects, which, as known, are aligned to their position with their sprite bottoms (i.e. their origin is at the middle-bottom and left-bottom respectively).

The math mistake is simple: it was doing e.g. `pos.y - height` instead of `pos.y - height  + 1`.
Because of that every character and room object in every existing AGS game is actually drawn 1 pixel higher than their position demands. This is true for any renderer, and for the Room Editor as well. I don't know if many people noticed that, but assume that most just aligned object positions in the room editor judging by their looks without precisely checking their coordinates.

But this is very easy to notice if you e.g. have a low-resolution tile-based room, where objects and characters must precisely match the grid. (That's what the user who reported this had.)

AGS 4 seems like a good place to fix this.

However, because users will be upgrading existing projects, there's a question whether the room objects' positions should be adjusted. Personally, I am in big doubts, because objects have several uses in AGS. Sometimes it matters how they visually relate to the room background, and sometimes their positions are dictated by a game logic. Furthermore, we may update the properties, but we may not upgrade scripts.

@AlanDrake told me to do their positions upgrade, so I had in this PR. But I'm leaving this on draft until we gather more opinions.


messengerbag:
I say do the fix, warn people about it in the release notes, and let game makers with ongoing projects decide whether (a) this is a problem for them, and (b) they want to update to the new engine nevertheless and tweak the positions themselves.

ivan-mogilko:
> We could add a yes/no dialog to prompt the user whether to proceed on translating the coordinates, and that should cover all needs.

I'm concerned that this may require a time and thought to consider, so demanding user to answer right away when opening a project may create a bad experience. It would be more convenient if such decision could be postponed.

But postponing an action would require having it as a menu command, and not just that, but its pending state has to be remembered by the project.

Following is just a fantasy for now, but I imagine we could have a "UpgradeComponent" in the Editor that handles pending optional data updates/conversions. It would work like this:
- each pending update is a item in a list, which user may look onto anytime.
- user may command to either perform update or cancel it, in the latter case it's removed from this list forever.
- the updates that relate to rooms also give a checkable list of rooms to apply this update to.

If something like that existed, then
- user could postpone deciding whether they need or not this automatic conversion in their project
- user may decide which rooms to apply this to and which not.

messengerbag:
That does sound very fancy. A perhaps simpler alternative is to just have a menu option (probably still under some kind of upgrade menu) to adjust all Character and Object y-coordinates by a user-set value. When the project updates, there could be a prompt asking whether to apply it with +1, and letting users know that there is a menu option to do/undo it.

AlanDrake:
We should remove the teeth and be done with it.
AGS4 is meant to break compat, make the code simpler with less legacy cases to handle.

About pending updates. We could warn the user that if they go ahead with opening the project it will be updated and saved. Then no more problems about inconsistent states and the such.
A reminder to the user to only go forth if they made a backup, and then apply the format upgrades without looking back.

ivan-mogilko:
> We should remove the teeth and be done with it.

If I'd look at that this way, then I would choose to not do any automatic adjustment at all, and let users handle this by hand on per case basis.

> AGS4 is meant to break compat, make the code simpler with less legacy cases to handle.

Yes, but that does not mean we should not make something to aid people migrate from one version to another.

AlanDrake:
That's why I suggested to make it clear and give the user an option when upgrading the project.
As you say, this is only necessary once. It's not an tool that could be useful down the line.

With removing "the teeth" I merely mean that we don't spend too much trying to make revertible settings and special cases that complicate internal logic. 
 

I'm against "apply or cancel" because it sounds like it would add complexity. The simple solution to deal with inconsitencies is saving the project right upon upgrading.
We just have to make it very clear this is a one way process and that the user should not proceed unless a backup was made.

I'm pretty sure a scary prompt, expressly urging users to backup before opening the project, will cut down user regrets more efficiently than any other system we can think about.

messengerbag:
> This operation is only necessary once for particular purpose, and if adjusting by 1 pixel. What could be a reason to let users choose arbitrary value, or apply one repeatedly (if the state is not saved, then there's no way to know if users did this already)?

It means that the user can do it and undo it, as many times as they like, to test and see the effect. And this way there is no need to keep track of whether they have done it or not, thereby simplifying things.

For this use-case there is no particular need for the ability to set any value other than +1 and -1, but if building it anyway, why not? It might come in handy in cases where a redesign shifts the whole room, or all rooms, up or down by some fixed amount (for example if there is a GUI redesign).

ivan-mogilko:
Alright, scrap the "Upgrade component".

What about "Upgrade Wizard"? AGS already has one when upgrading from 2.* projects. We could have a new one for upgrading to 4.0 from an older version (or older 4.0 versions).

1. It shows right on project opening so you cannot skip it by mistake.
2. Each page may contain a separate update selection or warning.
3. It looks more formidable than merely a dialog, and lets users to think about it before making a decision.

AlanDrake:
> What about "Upgrade Wizard"? AGS already has one when upgrading from 2.* projects. We could have a new one for upgrading to 4.0 from an older version (or older 4.0 versions).

I guess it could work. I didn't know there was one for 2.*.
How did that guy from discord manage to upgrade without making the backup, there's even bold text and a checkbox...

 
Anyway I doubt command line needs extra options to force, it should just upgrade. I don't see normal users using it, an advanced ones should know what they're doing. Might change in the very far future, but there's no use case for now.

ivan-mogilko:
I made a Upgrade Wizard here: #2794

My plan is that this 1-pixel fix should add a new page in the wizard, where user could select whether to adjust room objects positions in all rooms, and maybe characters too, or skip that and handle by hand later.

ivan-mogilko:
Grmmm.... This took a lot of extra work compared to fixing the problem itself, but I added a new page to the Upgrade Game wizard, corresponding to this room objects fix.

I'm still glad I had to do this, because now there's a proper Upgrade wizard that explicitly tells to users what is going to happen with their project and allows to back up (or cancel).

Anyway... this PR draft is now based on #2794 (until its merged), and provides a wizard page with "optional room updates".
Currently there's only one option that sais "Adjust ALL Objects in ALL Rooms by 1 pixel". It's ON by default (is that right?).

The looks of it is quite basic (as is the look of the Game Upgrade wizard overall). But I am not a GUI designer... I suppose that anyone who can do better may improve it later, since there's a "framework" for this now.

AlanDrake:
> Currently there's only one option that sais "Adjust ALL Objects in ALL Rooms by 1 pixel". It's ON by default (is that right?).

Should be a good default.

PR #2782, opened 2025-07-09
AGS 4: support creating ShaderProgram from a string in script

ivan-mogilko:
Implements following api:

`static ShaderProgram.CreateFromString(name, shaderScript, shaderDef)`

creates a ShaderProgram compiled from the given shaderScript string, optionally using a "shader definition" config in INI format (currently only used by Direct3D).

`static ShaderProgram.CreateShaderStub(name)`

creates a ShaderProgram that does nothing (no-op).

`static readonly ShaderLanguageType System.ShaderLanguage`

returns the shader language type used by the active gfx driver

The ShaderLanguageType is defined as:
```
enum ShaderLanguageType {
    eShaderNotSupported = 0,
    eShaderGLSL = 1,
    eShaderHLSL = 2
};
```

**NOTES:**

The ShaderProgram created from string will keep the whole shader script in memory and serialize one into game saves in order to recreate itself after save's restoration.
However, since it has only a string in language used at a time when it was created, so unlike the shaders created from file, it won't be able to automatically re-initialize if the gfx driver changes.
That is something that the user has to resolve in script. I suppose, they might save the ShaderLanguageType in a variable and check that variable's value after the save is restored.

In summary, on upside this api lets to create custom shaders programmatically in script, but on a downside user has to also script switching shaders if a gfx driver have changed in between the game sessions.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -663,6 +663,15 @@ enum StringSplitOptions {
 };
 #endif // SCRIPT_API_v400
 
+#ifdef SCRIPT_API_v400_18
+enum ShaderLanguageType {
+    eShaderNotSupported = 0,
+    eShaderGLSL = 1,
+    eShaderHLSL = 2
+};
+#endif // SCRIPT_API_v400_18
+
+
 internalstring autoptr builtin managed struct String {
   /// Creates a formatted string using the supplied parameters.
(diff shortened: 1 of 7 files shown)


ericoporto:
> but on a downside user has to also script switching shaders if a gfx driver have changed in between the game sessions.

Uhm, had an idea for this, what if it was possible to pass the `ShaderLanguageType` as an input and somehow add a few and then they would be stored so when loading a session, AGS would be able to restore from the proper one? 

I am fine with the current approach as this is I think an advanced feature, but I was just thinking if there was a safe approach to this.

Issue #2706, opened 2025-03-22
Better filtering for Log Panel


tarnos12:
**Describe the problem**
Currently it takes 3-4 clicks to change filtering options for the Log Panel:
1. Open hidden settings
2. Click on a log type(Game, Engine, Script, Plugin)
3. Click the arrow to open a drop down
4. Select filter level

**Suggested change**
There are many way to achieve it I am sure, but I am not familiar with the process.

The main point is to reduce the amount of clicks required.

One solution:
Add buttons to the Log Panel bar where you have pause, clear all, copy, open settings buttons and align it to the right side of the bar.

Similar to how Unity has 3 buttons which you can enable/disable by clicking on them for Info, Warning, Error logs.
Since in AGS there are 6 different log levels, it might require to either bundle them up or to have 6 buttons.

Buttons should act as filters for each separate log level, example: I want to see Info and Warnings only, but hide Debug/Fatal/Error etc.
Those buttons should also be applied on top of hidden filters to act as a filter layer on top of everything.
That means if I set Game, Engine, Plugin to filter up to the Warning level and Script up to Debug then filtering out Warning and Error would only show Info, Alert, Debug, Fatal from all 4 sections(Game, Engine, Plugin, Script)
Since the filters would apply to everything(it's basically another layer on top)

From a game dev perspective I rarely care about Plugin/Engine error/fatal/warning logs.
I mostly want to see my own logs from a Script.

If you decide to bundle 6 log levels into 3 for simplicity then this could work:
**Info**: Info + Debug
**Warning**: Warning
**Error**: Fatal, Error, Alert

If I dont want to see Debug from the Engine I can just open hidden settings and change the filter option for the Engine to something like Info or Warning.
This is something you'd do rarely so it's fine for it to be behind a settings menu.
But from day to day work I'd mostly care about generic filters that filter everything with easy access.

The end goal is to have easy to access filter buttons without having to go through hidden settings to enable them for each type of output(engine, game, plugin, script)

The settings are still useful to decide what is shown IF all filter buttons are turned ON.

I hope that makes sense, I had a short discussion with @ivan-mogilko on discord:
https://discord.com/channels/221047797292597249/221048075207180288/1353115894384955422

Leaving it here in case someone considers this to be worth doing.

ericoporto:
[...]

PR #2784, opened 2025-07-09
Doc: Create RFCs doc structure with template and example

edmundito:
## Background

Significant changes to the AGS source code are currently a bit convoluted to discuss. Sometimes it happens in GitHub issues, the AGS Forum, Discord, or directly in the PR. The PR can be the most dangerous because the change can be rejected, causing the most expensive work—writing actual code — to be lost.

As one example, I'd like to significantly change the Editor code to make it more flexible, maintainable, and adaptable for future cross-platform development or potential UI library changes. However, I'd like to get some feedback and talk about my plans before making code changes, so I don't surprise anyone when the PRs start appearing.

## Enter the RFC

Request For Comments, or RFCs, are a good way to accomplish this. The user creates a document that can be commented on via the regular PR review process, and then the outcome can be tracked and logged for historical reasons. This also allows developers to gather feedback and gauge the appetite for a change.

## This change

This provides a template based on best practices, along with an example of how the Log Panel could have been proposed. The initial submission is extensive and is open to feedback and changes to adjust the requirements for the RFC, as well as the process.

## AI Disclosure

I used Anthropic's Claude Sonnet 4 to generate this change, and then manually reviewed and made some adjustments. I've reviewed the generated structure and agreed that this is a typical pattern I've seen in other RFCs; I did my best to ensure that it didn't introduce any hallucinations. If you notice anything unusual, I can always edit it out or make changes.

### DIFF
--- a/RFCs/QUICK_REFERENCE.md
+++ b/RFCs/QUICK_REFERENCE.md
@@ -1,0 +1,151 @@
+# AGS RFC Quick Reference
+
+## TL;DR - How to Submit an RFC
+
+1. **Copy the template**: `cp RFCs/templates/rfc-template.md RFCs/drafts/0000-my-feature.md`
+2. **Fill it out**: Replace placeholder text with your proposal details
+3. **Choose component type**: Mark as `[EDITOR]`, `[ENGINE]`, `[EDITOR AND ENGINE]`, `[INFRASTRUCTURE]`, or `[DOCUMENTATION]`
+4. **Submit PR**: Open a pull request with your RFC
+5. **Iterate**: Respond to feedback and refine your proposal
+
+## RFC Types & Examples
+
+| Type | When to Use | Example |
+|------|-------------|---------|
+| `[EDITOR]` | Changes to AGS Editor interface, tools, or functionality | New room editor features, plugin system changes |
+| `[ENGINE]` | Changes to AGS runtime, scripting, or game execution | New script functions, graphics features, platform support |
+| `[EDITOR AND ENGINE]` | Changes affecting both Editor and Engine | File format changes, cross-component architecture |
+| `[INFRASTRUCTURE]` | Build system, CI/CD, project structure | CMake changes, new deployment processes |
+| `[DOCUMENTATION]` | Major documentation initiatives | API documentation overhaul, tutorial series |
+
+## Common Mistakes to Avoid
+
+❌ **Too vague**: "Make AGS better"  
+✅ **Specific**: "Add real-time lighting system to AGS Engine"
+
+❌ **No motivation**: Jumping straight to technical details  
+✅ **Clear motivation**: Explain the problem you're solving
+
+❌ **Missing alternatives**: Only one solution considered  
+✅ **Consider alternatives**: Compare different approaches
+
+❌ **No backwards compatibility plan**: Breaking changes without migration path  
+✅ **Migration strategy**: How existing users adapt to changes
+
+## Template Checklist
+
+### Required Sections
+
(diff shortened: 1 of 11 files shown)


ericoporto:
I really don't like this much formality and this much bureaucracy.

ericoporto:
That is the wrong way to go in my opinion, get one prototype of what's wanted, discuss in the forums where the gamedevs are, or an issue if it's structural, then either put a PR or move it out of draft.

RFCs as a document must be humanly readable and writeable and not that noisy, I recall the last time I saw something like actually good were the IETF documents which are a plain text file with a limited 80 char width and that's it.

The concept of the two columns in C++ papers aren't totally bad either - "we have X, change proposes Y".

But overall I don't think this is useful simply because this type of process assumes a ton of people involved and the nearly infinite manpower to achieve a goal. They are a drain of human hours and resources. They also just don't survive the realities, I barely recall any time I thought about something and got it implemented the way I had in mind when I started, there are too many complexities in real life.

I also challenge the notion that conversation isn't the way to go to discuss things.

If this approach was actually feasible in real life we would be writing documentation (as in the ags-manual sense) before doing anything, but I don't believe this is a valid way to approach actually getting things working, you know, building for all compilers, running in all platforms. We put some effort, get it working in some way, and then move to something else.

ericoporto:
If the planning is important then a PR is certainly the wrong tool for this discussion. About personal experience, sure, I am from a research background, mostly of my work results in failure - but what does work is new, and sometimes cool too.

>Another thing, having something described in detail helps others understand what do you have in mind.  

Any real life example of anything game related will probably require images (or videos) and I don't want to have these here as they will make the repository bigger quick. If you want to run an experiment with this, it is better to use a different repository.

ivan-mogilko:
> > Another thing, having something described in detail helps others understand what do you have in mind.

> Any real life example of anything game related will probably require images (or videos) and I don't want to have these here as they will make the repository bigger quick. If you want to run an experiment with this, it is better to use a different repository.

That's not precisely what I meant.

I like having a feature "specification", that explains what I should *expect* from it. How it works, what it is supposed to do, and what not. How it deals with problems. Ideally I like having this before starting to implement things. Although I realize that's not always easy to write one, but it's good to have such outline.

* this helps multiple developers to understand the idea better;
* this lets to know whether and when the goal is achieved or not;
* this lets to test the feature, because you know what is a right behavior and what is not.

ericoporto:
I am not convinced that this needs to be in this repository. And both the format and the process are way too complexly written for a person to read. Things like `Success criteria` is just robot thinking, the success criteria for 99% of things I guess will be "people happily use feature x", which cannot be measured, just not objectively.

> this lets to know whether and when the goal is achieved or not; and if not, then what is remaining and should be planned for later.

Usually what's done is the minimum that allows one to achieve what's necessary even if in a convoluted way, and the pending issues are doing this nicely integrated in the editor. Support authoring translation, `template.files` or shaders from outside is fine, but if it's not in the Editor it's like if they don't exist in the project explorer as a node that you click and it pops an editor, for the majority of users I guess it's like if it doesn't exist. But it's better to get some functionality rather than none at all.

edmundito:
We need to have a more in-depth discussion to agree on how to plan for larger features that may require multiple steps. I agree with @ivan-mogilko that something needs to be available, maybe not this.

> I am not backing this particular RFC format, I was speaking about planning in general.

I'm not super attached to this either, but what should I do?  Should I close this PR and move the discussion to an issue?

edmundito:
See #2706

PR #2785, opened 2025-07-09
New compiler: Fix bugs around `null` as an argument for a function parameter

### LINKED ISSUE #2781: New compiler: can't pass "null" as a "const string" argument
CC @fernewelten 

**Problem**

In AGS script `const string` is interpreted as a `const char*`, therefore it should be permitted to pass `null` in.
However, the latest version of the new compiler denies that and throws an error:

> Cannot convert 'null' to 'const string'

**Expected behavior**

Compiler should permit to pass `null` for a function argument of type `const string`.

fernewelten:
Fixes #2781

Fix a bug where the compiler wouldn't accept `null` as an argument for a `string` or `const string` parameter.

Typical code:
```
int func(const string s)
{
    func(null); // ← Erroneous error message here
}
```

I'm taking this as an opportunity to also fix a bug where the compiler wouldn't accept `null` as a variadic argument in a format function for a `%s` format.

Typical code:
```
import int func(__format String s, ...);

void game_start()
{
    func("%s", null);
}
```

### DIFF
--- a/Compiler/script2/cs_parser.cpp
+++ b/Compiler/script2/cs_parser.cpp
@@ -1729,9 +1729,10 @@ bool AGS::Parser::IsVartypeMismatch_Oneway(Vartype vartype_is, Vartype vartype_w
         return false;
 
 
-    // Can convert 'null' to dynpointer or dynarray
+    // Can convert 'null' to dynpointer or dynarray or 'const string'
     if (kKW_Null == vartype_is)
         return
+            _sym.VartypeWithout(VTT::kConst, vartype_wants_to_be) != kKW_String &&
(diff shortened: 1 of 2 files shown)


fernewelten:
(Force-pushed.)

ivan-mogilko:
I confirm that the bug is fixed. But while testing I also found another format specifier that is not recognized by the format check: `%p`. It used to work with the managed handles (because they convert to real addresses before being passed into the function). Not certain how useful that is, but could you also add that specifier to the list?

fernewelten:
> `%p`, which is supposed to print raw pointer value.

Not sure why this doesn't work, it ought to. I'll look into it.

PR #2787, opened 2025-07-14
Engine: fix check for asset existence

mausimus:
The check for asset existence can return too early without checking all libs.

To reproduce, start from BASS template, add a Music file, and check if it's Available. The problem only appears in compiled game, not in editor.

```
function room_FirstLoad()
{
 if(aMusic1.IsAvailable)
 {
  player.Say("It's available.");
 }
 else
 {
  player.Say("It's NOT available.");
 }
}
```

Looks like this got broken here:
https://github.com/adventuregamestudio/ags/commit/5cb40d3b371ffc31e3ec6eb71656716ff11f63d2

### DIFF
--- a/Common/core/assetmanager.cpp
+++ b/Common/core/assetmanager.cpp
@@ -175,7 +175,8 @@ bool AssetManager::DoesAssetExist(const String &asset_name, const String &filter
         }
         else
         {
-            return lib->Lookup.count(asset_name) > 0;
+            if (lib->Lookup.count(asset_name) > 0)
+                return true;
         }
     }
     return false;


ericoporto:
Uhm, it seems this happens in this case because the music is in a separate box file in this template 

https://github.com/adventuregamestudio/ags-template-source/blob/6531b767ff48df3ccbe4926eb6624b1c7e92e485/BASS/Game.agf#L4628

And because it returns there instead of following to the next for iteration it doesn’t gets to find that asset.

ericoporto:
@mausimus  thanks for the fix!

Had time to finally sit on the computer and check, I was curious how this wasn't noticed before but it appears it only affects that check and not if the music will actually play or not, and I guess people don't check before playing, at most they check for null for the resulting AudioChannel in case they were going to access that AudioChannel later and that's it.

mausimus:
No problem! Yes I stumbled upon this by chance, wasn't sure if there would be any other impact, the check itself does sound quite rare.

PR #2793, opened 2025-07-23
ci: add Arm64 builds for Linux and Windows MinGW

ericoporto:
This adds some cool experimental ci runs from the public beta arm64 partner images.

I think these are fun, and also I have a raspiberry pi arriving tomorrow I want to play with. But the real reason I am doing this PR here (and leaving it in draft) is the error that will happen in the Windows Arm64 MinGW build, [on ci here](
const int AccessibilityPageDialog::TextReadSpeedMin;
const int AccessibilityPageDialog::TextReadSpeedMax;), highlighted below:

```
ld.lld: error: undefined symbol: AGS::Engine::AccessibilityPageDialog::TextReadSpeedMin
>>> referenced by C:/a/ags/ags/Engine/platform/windows/setup/advancedpagedialog.cpp
>>>               libengine.a(advancedpagedialog.cpp.obj)

ld.lld: error: undefined symbol: AGS::Engine::AccessibilityPageDialog::TextReadSpeedMax

>>> referenced by C:/a/ags/ags/Engine/platform/windows/setup/advancedpagedialog.cpp

>>>               libengine.a(advancedpagedialog.cpp.obj)

c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I don't understand what that error is, but I had the same error locally when I was trying to build @ivan-mogilko ags4-SDL_ttf branch until I did a rebase and it disappeared. So my guess is whatever is this error we (may?) have already fixed it in the ags4 branch. But I don't know what that is.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,10 +17,12 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows-mingw64, os: windows-latest, artifact: 'ags-mingw64', shell: 'msys2 {0}', runbin: "./ags",      bindir: "",       msystem: mingw64, msys-env: mingw-w64-x86_64 }
-        - { name: Windows,         os: windows-latest, artifact: 'ags-windows', shell: cmd,         runbin: ".\\ags.exe", bindir: "Release" }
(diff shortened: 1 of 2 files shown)


ericoporto:
Amazingly [this SO suggested fix](https://stackoverflow.com/a/5508201) works.

PR #2794, opened 2025-07-24
AGS 4: UpgradeGameManager and Wizard dialog

### LINKED ISSUE #1596: AGS 4: The project upgrade to open rooms may leave game in the inconsistent state

### LINKED ISSUE #2791: AGS 4: Upgrade Game Wizard
Because when AGS 4 upgrades the older games some things may change drastically, or require attention from user, I came to a realization that we need a new Upgrade Game Wizard dialog, which would be displayed when loading an older game, and guide the user through any significant changes which are going to be applied to their game. Some of these changes may also be made optional, and wizard dialog would explain the available choice and possible consequences of each option. The upgrade may be combined with a project backup, similar to how a backup is done when importing pre-3.* games into any 3.* Editor (in a "import old game wizard" dialog). Finally, the project will be saved unconditionally after the upgrade, thus preventing any inconsistencies.

I think this will resolve or help resolve issues #1596 and #2779.

Couple of things that should be kept in mind when designing this wizard:
* it should be simple to add additional steps if ones are required in the future;
* it should be possible to automate the upgrade process in case the Editor is run from command line with /compile, /maketemplate or similar command, where it's forbidden to display any modal GUIs.

Following is a general idea of the process and implementation that I have.

1. Make a class that manages the game upgrade. Let's call it UpgradeManager for simplicity. There's an existing example called OldGameImporter, although it's quite trivial, and we might need something more sophisticated. BTW, OldGameImporter should likely be removed, as it deals with 2.* games, which we no longer support importing in AGS 4.
2. The upgrade consists of "steps". The "upgrade step" is likely a base class that may have specializations. For each "step" subclass we have description and some parameters. For example, "Backup project files" is a step, and its parameter is a backup destination folder. "Step" may be defined as optional, in which case it has "Enabled" check. Each step has a WizardPage control subclass connected to it. Finally, each step class has a method that performs the operation over the project.
3. When UpgradeManager is tasked with a game upgrade, it must either have Game already loaded, or preload game's XML and find out its saved version. Judging by the game version, UpgradeManager decides which steps are required. Probably the "step" classes should also have this in their definition (or the manager can map each step to the latest version in history that requires it). Based on this, UpgradeManager gathers a collection of steps.
4. If we work in GUI mode, then UpgradeManager creates a "Upgrade Game Wizard" dialog, inserts pages from steps, and shows it on screen. After the user walked through the pages, and finally pressed "Finish" (or "Go", whatever), the user choices from the wizard's pages are applied to steps, and UpgradeManager runs these steps one by one.
5. If we work in auto mode, then UpgradeManager can possibly receive a list of options from command line, and assign step parameters based on these options. Then steps are executed as usual.
6. After the upgrade completes with success, the game project is saved unconditionally, this will prevent any inconsistencies. If the upgrade process included rooms, then each room is also re-saved.
7. **NOTE: this wizard is only shown when the changes are complicated and game/room upgrade cannot be done automatically on load.** In any other cases the UpgradeManager does the update silently. Perhaps the step class may have a marker telling this.

ivan-mogilko:
Resolves #2791 and #1596

UpgradeGameManager is a class that acts as a controller for all the tasks performed when upgrading an older game project.
It gathers a collection of Upgrade Tasks, some of which may require to display a cue to user about a bigger change to the project, others are meant to be executed implicitly and without condition.
If there are any tasks that imply a big change, then UpgradeGameManager creates and shows a "Upgrade Wizard" dialog.

IUpgradeGameTask is a interface of a single upgrade task (although a single task may perform various updates, how they are organized is a matter of convention). This interface provides means to know which game versions should have this task applied, and instructions on how to handle this task (whether it may be canceled by user, skipped in case of a failure, and so forth).
IUpgradeGameTask also has a method that creates a dialog page for the "Upgrade Wizard", with information and related options (if necessary).

UpgradeGameManager has its own list of "common" upgrade tasks which it creates itself, but also fires a EditorEvents.GamePrepareUpgrade event, which lets any Editor's component to add its own Upgrade Task.

Following Tasks are implemented by this PR:
* UpgradeGameIntroAndBackupTask - (optionally) does a game files backup, this task is added by the UpgradeGameManager when there's at least "non-implicit" upgrade task in the gathered list. This task also displays an introductory wizard page that tells user that making backup is important, and suggests to do that on their own, or choose to let Editor to do it.
* UpgradeGameCommonTask - an implicit task that is run always and contains all the random Game updates. The update code is picked out from the Tasks.UpgradeFeatures() method (which is now removed).
* UpgradeGameRoomsOpenFormatTask - the task that converts rooms from CRM to open format. For this I actually left the conversion code in RoomsComponent, because it felt like having more sense if it's there, and also it's calling a number of private RoomsComponent's methods. The Task object itself simply delegates work to RoomComponent when run, through a ... delegate.
* UpgradeGameFontsToFontFilesTask - the task that converts fonts to a new representation (font files + fonts). Similarly, the conversion code itself remains in FontsComponent, and the task object delegates the work.

On a side note, this PR also removes "import 2.72 game" dialog, because that no longer works in AGS 4.0 anyway.

---

KNOWN ISSUES SO FAR

[FIXED]~There's something wrong with Wizard's size, it's too short and about 1/3 of the text is cut off. I assumed that it will resize itself automatically, but it does not, or maybe I did something wrong.~

[FIXED]~Also, I wanted to make parts of the text **bold**, for that reason I used "rich text box" there, but formatting did not work either, so looks like I did something wrong again.~

---

WHAT HAS TO BE TESTED

Quite a lot of things... from the top of my head:

1. Loading current version should not cause any effect.
2. Loading slightly older version (that does not require upgrading rooms, etc) will be upgraded correctly without showing any Wizard dialog. It's not saved when loaded, and gives that standard warning that if you save then you won't be able to open in an older editor.
3. Upgrading a 3.6.* game: should display a Upgrade Wizard. If Wizard is canceled anytime, then no changes are made to the project and Editor returns to the "Choose game to load" dialog. If Wizard is finished, the upgrade process begins. It should display BusyDialog with all kinds of progress messages (best tested with big games with lots of rooms, because it's too fast in case of small games). NOTE: the game project *will be saved unconditionally* after this upgrade (this should fix #1596).
4. You can choose to do a backup in the Wizard. In which case a folder named Backup0 (1,2,3... etc) will be created inside the project folder, and the old files moved there. ~This does not move literally all files though, but only standard game files that are usually modified during update: Game.agf, scripts, rooms, fonts, translations, sprite file.~


ericoporto:
> best tested with big games with lots of rooms, because it's too fast in case of small games

Haven't tested it yet, but if a message happens too fast to be read perhaps they can be accumulated and later copied to the Output panel - it only has "errors" (red) and "warnings" (yellow) but perhaps there could be some "information" icon for this?

ivan-mogilko:
Adjusted sizes using layout tables. Switched to assigning Rtf text. Added a readonly variant of richtextbox that does not display a caret when you click on it (this was bugging me much).

ivan-mogilko:
Remade "backup" task, now it copies all the files, except for Compiled and _Debug folder. I figured out that we cannot make guesses which files need to be backed up and which do not.

Also the wizard page now lets choose a destination folder for backup.

PR #2795, opened 2025-07-24
View Editor: "Copy/Paste" frames and "Reverse Frames" commands

### LINKED ISSUE #2792: Editor: expand Copy/Paste operations in View Editor
View Editor did not allow to select multiple frames before, so likely that's why there was no proper "Copy/Paste" command for selected frames, only for the whole loop; but once multi-selection has been implemented there's no reason not to have these. Unfortunately I did not realize they need to be added earlier.

Anyway, I suggest adding following commands:
* Copy selected frames (Ctrl + C)
* Cut selected frames (Ctrl + X)
* Paste after (pastes copied frames *after* selected frame) (Ctrl + V). This complements existing "Paste over the loop" which replaces all existing frames with copied ones.
* Paste before (?).
* Reverse selected frames (R) - reverses an order of selected frame range
* Reverse all frames in the loop.

Another thing, there's "Paste over the loop flipped" menu command, which combines 2 operations. I wonder if it should be removed if we add more ways of pasting frames and more possible operations, because the number of possible combinations will increase, and we might have to add too many menu commands to cover them all for consistency.

If not, then perhaps there could be just "Paste over the loop reversed".

ivan-mogilko:
Resolve #2792

Adds following commands to the ViewEditor:
* Copy frame(s) (Ctrl+C) - works if selection is across multiple loops too
* Cut frame(s) (Ctrl+X) - works if selection is across multiple loops too
* Paste frame(s) before - pastes copied number of frames *before* the selected frame in a single loop
* Paste frame(s) after (Ctrl+V) - pastes copied number of frames *after* the selected frame in a single loop

Also
* Paste over loop - now works if clipboard has copied frames, not whole loop.

More commands:
* Reverse frames (R) - reverses an order of selected frames within a single loop. Selection may have gaps in it (i think).
* Reverse all frames - reverses all frames in a loop.
* Paste over loop reversed - pastes a copied loop or frames over the loop and then reverts frames.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -295,6 +295,7 @@
     <Compile Include="Panes\ScriptEditorBase.cs">
       <SubType>UserControl</SubType>
     </Compile>
+    <Compile Include="Panes\ViewEditClipboard.cs" />
     <Compile Include="Panes\ViewLoopContextMenuArgs.cs" />
     <Compile Include="Panes\ZoomTrackbar.cs">
       <SubType>UserControl</SubType>
--- a/Editor/AGS.Editor/Panes/ViewEditClipboard.cs
+++ b/Editor/AGS.Editor/Panes/ViewEditClipboard.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using AGS.Types;
+
+namespace AGS.Editor
+{
+    /// <summary>
+    /// ViewEditClipboard shares copied data among the ViewLoopEditor instances.
+    /// </summary>
+    public class ViewEditClipboard
+    {
+        /// <summary>
+        /// Last sprite selected to be assigned to the view's frame.
+        /// </summary>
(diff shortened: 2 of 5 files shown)


ericoporto:
Hey, I tested under debugger and it really doesn't trigger the `OnCopyFrames`, or `OnCutFrames`, using the shortcuts. I noticed that there is no Edit menu while on the ViewEditor. Since it only add the elements to the menu when the OnContextMenu is triggered, I think it really wouldn't work to trigger the shortcuts.

I think the fix should be something like this https://github.com/ericoporto/ags/commit/38f2d4951f7ab705bdcfd7ebfe4f698fdb54e488

ivan-mogilko:
I did it through the "OnKeyPressed" instead, along with the existing handling for F, R and Delete.

ericoporto:
ok, I tested and the shortcuts works fine now! 

I played with it a bit, and if you have multiple views opened and want to copy paste frames between done it all works much better now. Great work!

PR #2797, opened 2025-07-29
AGS 4: Script API: expand animation RepeatStyle

ivan-mogilko:
In script, expand enum RepeatStyle with following modes:
* eOnceAndReset - same as eOnce, but resets back to the first (or last, in case of backward direction) frame. This mode was already working for Room Objects if using a legacy global Animate function, but it was never implemented consistently in the contemporary API.
* eOnceAndBack - goes all the way in the initial direction, then goes all the way back to the first (last) frame.
* eRepeatAlternate - repeats indefinitely, changing direction whenever reaching first/last frame of a loop (or multiloop sequence).

NOTE: "direction" parameter is considered a "initial direction". Actual direction might switch while the animation is playing, depending on RepeatStyle.

Support all the RepeatStyle modes in Character.Animate, Object.Animate and Button.Animate.

Support all the RepeatStyle modes in MotionPath.Create, so that moving along a path could be done same ways (alternating directions, etc).

In code refactored animation fields in the character, roomobject and animatingbutton structs, moved all fields related to the current animation to a struct called ViewAnimateParams. Created two enums: AnimFlowStyle and AnimFlowDirection to use instead of macro constants.
Refactored CycleViewAnim function, make its code easy to understand.

Updated saves formats for the affected objects, ensure that they save animation parameters consistently.

---

NOTE: RepeatStyle is also used in AudioClip.Play(), but I suppose that it makes no sense implementing other types than Once and Repeat there, and we can just say that it does not support others...

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -162,7 +162,7 @@ enum RoundDirection {
 enum RepeatStyle {
   eOnce = 0,
   eRepeat = 1,
-  eOnceReset = 2,
+  eOnceAndReset = 2,
   eOnceAndBack = 3,
   eRepeatAlternate = 4
 };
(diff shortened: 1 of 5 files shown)


AlanDrake:
I think it's fine as it is, but a possible alternative to `eOnceAndBack` could be `eAlternateOnce`.

PR #2803, opened 2025-08-05
Editor: hide certain properties when multiple objects are selected for editing.

### LINKED ISSUE #2802: Editor: must prevent certain properties from being changed when multiple objects are selected
Since 3.6.2 Editor supports setting properties to multiple selected gui controls and view frames.
(Previous versions also support same for Sprites.)
There may be a case when certain properties should not be let set to the identical value in separate objects, therefore they must not let edit themselves when multiple objects are selected. The good example is a (Script)Name.

An existing mechanism of hiding own properties is inheriting ICustomTypeDescriptor and implementing GetProperties(). But you cannot detect that multiple objects are selected this way (ICustomTypeDescriptor cannot know anything about the context it's being called from).
There's another method: OnPropertyChanged implemented in certain Editors, but that does not seem to work here either, because it does not let retrieve multiple "old values" per each object. Instead, when there are multiple objects selected, the property grid can report "null" as a "old value" (when values are different).

This means that we need something else to solve this problem.
Ideally, I think, there would be an Attribute which we may assign to each property that should not be visible/enabled on multiple object selection, and property grid would not display these in such case.

ivan-mogilko:
Fix #2802

Implemented BrowsableMultieditAttribute that marks properties that should or should not be displayed whenever multiple objects are selected for edit in PropertyGrid.

When PropertiesPanel.SelectedObjects is set, check whether there are multiple objects, and if yes then assign this custom attribute to the propertiesGrid.BrowsableAttributes collection. If a single object is selected, then reset BrowsableAttributes to defaults.

This allows to hide properties that must have unique values, such as Name, so that user won't try to edit them.

### DIFF
--- a/Editor/AGS.Editor/GUI/PropertiesPanel.cs
+++ b/Editor/AGS.Editor/GUI/PropertiesPanel.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Windows.Forms;
 using WeifenLuo.WinFormsUI.Docking;
 using System.Windows.Forms.Design;
+using AGS.Types;
 
 namespace AGS.Editor
 {
@@ -103,6 +104,7 @@ namespace AGS.Editor
             {
                 propertiesGrid.SelectedObject = value;
                 propertiesGrid.ExpandAllGridItems();
+                propertiesGrid.BrowsableAttributes = null; // reset to defaults
             }
         }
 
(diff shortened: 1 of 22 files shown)


ivan-mogilko:
> I am impressed that it appears that properties that are different also change to some empty value in the property grid?

That is a standard PropertyGrid behavior, it actually passes value as "null" in such case.

PR #2806, opened 2025-08-11
Editor: add Go To Room by Number dialog

ericoporto:
I've been upgrading a big game that has a lot of rooms to ags 3.6.2, so I can then test the upgrade wizard in ags 4.

There are a lot of new warnings in the rooms that shows up, like room 3 warning and something, I copied the warnings to a text editor to be able to fix things one by one, and I noticed I wanted a quick way to jump to a specific room number in the project explorer. I simply copied the Go To Line dialog and made it a Go To Room number, which can be opened by right clicking in the room top level node.

1. image

2. image

3. image

### DIFF
--- a/Editor/AGS.Editor/GUI/ProjectTree.cs
+++ b/Editor/AGS.Editor/GUI/ProjectTree.cs
@@ -46,6 +46,7 @@ namespace AGS.Editor
 			_projectTree.ItemTryDrag += projectTree_ItemTryDrag;
             _projectTree.ItemDragOver += _projectTree_ItemDragOver;
             _projectTree.ItemDragDrop += _projectTree_ItemDragDrop;
+            _projectTree.PriorityKeyDown += _projectTree_PriorityKeyDown;
         }
 
         private void _projectTree_BeforeCollapse(object sender, TreeViewCancelEventArgs e)
@@ -395,14 +396,8 @@ namespace AGS.Editor
             }
         }
 
-        private void projectTree_KeyDown(object sender, KeyEventArgs e)
+        private bool _projectTree_PriorityKeyDown(object sender, KeyEventArgs e)
         {
-            if (e.KeyCode == Keys.F2)
-            {
-                if (_projectTree.SelectedNode != null)
-                    _projectTree.SelectedNode.BeginEdit();
-            }
-
             // We are going to pass down keys in case a context menu entry has shortcuts
             if (_projectTree.SelectedNode != null)
             {
@@ -421,10 +416,21 @@ namespace AGS.Editor
                             if (command.ShortcutKey == e.KeyData && command.Enabled)
                             {
                                 component.CommandClick(command.ID);
+                                return true;
                             }
                         }
                     }
                 }
+            }
+            return false;
+        }
+
+        private void projectTree_KeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.KeyCode == Keys.F2)
+            {
+                if (_projectTree.SelectedNode != null)
+                    _projectTree.SelectedNode.BeginEdit();
             }
         }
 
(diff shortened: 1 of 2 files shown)


ivan-mogilko:
There are two multipurpose dialogs: NumberEntryDialog and NumberEntryWithInfoDialog, may these be used instead of creating a new similar one?

ericoporto:
I made a specialization 

image

Typing the number selects on the list, you can also directly select on the list, and if the number of the room doesn't exist the ok button becomes disabled.

ericoporto:
Made an example use of it for Inventory Items.

ericoporto:
I think I added everything that can be go to by number. I can squash everything in a single commit to make it cleaner.

Please check if there's something more to add.

ericoporto:
I can drop the two latest commits and try again later, but this was an idea to support context menu shortcuts in the project panel. Unfortunately, if a Script Editor is open, it picks up the key presses even if the script is not on focus, and I have no idea how to fix it. But if it was fixable, it would be pretty nice to quickly navigate the project tree when you need something.

ivan-mogilko:
This seems like a good addition overall.

Just to clarify: overriding main menu is done in order to override context-depending Edit menu?
In general, it's not a good thing to override global commands that are meant to work everywhere.

I started testing this, and if a parent node does not contain any items, then trying to "Go to ..." causes Unhandled exception.

A minor thing: in "Audio" node I suggest ordering "goto" command above "Properties", because "Properties" traditionally is latest in context menus.

ivan-mogilko:
Suggestion: treat double-click on the item in the list as confirmation (same as if user clicked on OK).

ericoporto:
> could you add "Ctrl+G" shortcut to the menu item? (I know it won't be working on its own, but that will hint users about key shortcut.) This may be done using a utility function in ToolStripExtensions which accepts more arguments.

You mean make it show there in text? I don't know how to do this, I did pass it correctly in the menu command but I don't know why it doesn't show.

> I noticed that command sais "Go to Dialog", "Go to Character", but then "Go to room" (room starts with a small letter).

I can fix this, yes. Do we want the bigger letter in the type or prefer small letter? I understand the issue is to pick one as standard but now I am not sure which way to go.

> Just to clarify: overriding main menu is done in order to override context-depending Edit menu?

The Edit should not appear if the script editor is not focused but it stays there, in theory if you click the project explorer, the edit would change to be relative to the project explorer, but instead it stays there? I don't know how to change this and am not sure how to work on it, but if you click the project explorer, it feels expected that things from it would have priority, including handling key presses. This doesn't happen though and the key down event won't trigger if there is a script editor pane open. Also, it seems neither of the other panes does this and since the script editor felt way out of my league to understand, this workaround makes it work without much trouble. I still need some workaround anyhow since the shortcut of a context menu can only work after it exists, so after a right click by default.

ericoporto:
I can squash all commits except the two last ones, so we get one commit that adds the functionality and the four last ones, so we have

- big commit that adds Go To to Inventory Item, Views, Character, Dialog, GUI, Rooms and Audio Clip.
- add non-working shortcut
- make shortcut show up
- make shortcut work
- make shortcut work when project tree is focused even if something not focused has the same shortcut.

ivan-mogilko:
> I started testing this, and if a parent node does not contain any items, then trying to "Go to ..." causes Unhandled exception.

This is fixed, but now it looks weird, I hit the menu command and nothing happens.
Either the menu item should be disabled, or some message displayed if no items are present.

ericoporto:
I chose to disable it in the context menu here https://github.com/adventuregamestudio/ags/pull/2806/commits/ed538f7916838a28db120c2daa9f010e7874c42e

It will still do nothing if you are hitting the shortcut though, since it will just call nothing.

Can I do the commit squashing I mentioned before?

ericoporto:
Also, this is not related to this PR, but I was thinking about adding Ctrl+N for the "New Character",... and Ctrl+Shift+N for the "New Folder" where applicable in the other context menus... But just an idea for now.

ivan-mogilko:
> Can I do the commit squashing I mentioned before?

Of course.

ericoporto:
Did the necessary squashing! Kept the later commits because they make easier to understand why some stuff was done. I think this PR is done.

PR #2808, opened 2025-08-15
Editor: Font preview redesign

### LINKED ISSUE #2708: Editor: redesign Font preview pane
**Problem**

The font preview panel was initially purposed to display a ASCII range of characters, and the only change since was that the range was expanded to 256 chars. This means that it does not let evaluate non-latin languages, or any chars outside of this small range.

Very recently I've rewrote the way it draws font, letting it draw any number of chars without allocating huge bitmaps: 4fb3399c9892958201c04324b81bef3e873ff86d
but even if we let draw all available glyphs there, it will still be not very practical, as what user sees is only a grid of all characters with a scrollbar to navigate them.

**Suggestion**

Redesign the font preview from ground up. Think of what it should display and provide to the user.
As an example, I might propose following:
1. Keep a full glyphs grid, but add a control that lets to jump to particular unicode char by inserting it into a text field or typing its index.
2. Add a separate control (above or below) that lets user input a test string and would display it in this font.

ivan-mogilko:
Resolve #2708

Redesigned Font preview panel, in order to help users see the font when used to draw an actual text, and be able to see all the font's characters when there are more than 256 of them.

1. Font panel is divided (uses splitter) into two parts:
* text preview
* font preview

2. Text preview allows you to type any text and see how it looks like with this font. This allows text wrapping, in order to test line spacing.

3. Font preview displays all the characters available in the font, with a scrollbar if they don't fit into the panel.
4. You can enter either character code or character itself into the input fields, and press "Goto" button to jump to wanted character.
5. You can select a character in the preview and it will be printed in the input fields.
6. Optionally the preview may display character codes right under the characters, there's a checkbox that toggles this mode.

<details>
<summary>Click for Screenshots</summary>

font-preview-new1

font-preview-new2

</details>

---

TODO:
1. Should refactor coordinate conversion and place into helper methods. Preview pane uses "scaling" factor, which currently depends on the game resolution. This is not entirely good, but regardless, the scaling may be wanted, so I keep it. The calculation code becomes a bit more complicated though.
2. There's a strange glitch where the preview control's bottom is always cut off slightly by the splitter's panel. Could not understand why this happens.
3. Need to figure out if we want to share the text preview and display modes among editors, and/or save them in workspace data between sessions. Idk if that's necessary.

### DIFF
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -298,7 +298,7 @@ struct FontInfo
         kRounded = 1,
     };
 
-    // General font's loading and rendering flags
+    // General font's loading and rendering flags (FFLG_*)
     uint32_t      Flags;
     // Nominal font import size (in pixels)
     int           Size;
--- a/Common/font/agsfontrenderer.h
+++ b/Common/font/agsfontrenderer.h
@@ -117,6 +117,8 @@ public:
     virtual void GetFontMetrics(int fontNumber, FontMetrics *metrics) = 0;
     // Perform any necessary adjustments when the AA mode is toggled
     virtual void AdjustFontForAntiAlias(int fontNumber, bool aa_mode) = 0;
+    // Get a range of supported character codes (first and last present in the font)
+    virtual void GetCharCodeRange(int fontNumber, std::pair<int, int> *char_codes) = 0;
 
 protected:
     IAGSFontRendererInternal() = default;
--- a/Common/font/agsfontrenderer.h
+++ b/Common/font/agsfontrenderer.h
@@ -119,6 +119,8 @@ public:
     virtual void AdjustFontForAntiAlias(int fontNumber, bool aa_mode) = 0;
     // Get a range of supported character codes (first and last present in the font)
     virtual void GetCharCodeRange(int fontNumber, std::pair<int, int> *char_codes) = 0;
+    // Get a list of supported character codes
+    virtual void GetValidCharCodes(int fontNumber, std::vector<int> &char_codes) = 0;
 
 protected:
     IAGSFontRendererInternal() = default;
--- a/Common/font/agsfontrenderer.h
+++ b/Common/font/agsfontrenderer.h
@@ -14,6 +14,7 @@
 #ifndef __AC_AGSFONTRENDERER_H
 #define __AC_AGSFONTRENDERER_H
 
+#include "util/geometry.h"
 #include "util/string.h"
 
 struct BITMAP;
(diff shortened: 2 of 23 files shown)


ericoporto:
The scroll of the font is like it's missing a `Clear()` if this was a DrawingSurface in AGS Script, see video below


https://github.com/user-attachments/assets/7f5bda76-459f-4086-8d67-05cab12c10a9

ivan-mogilko:
> Uhm, should this be able to preview things like outline? (It never did in the past). It probably should not if in ags4 this panel is tied to fonts and not the instances.

I can make it display outlines too, but then I will have to move automatic outline function to the Common, or copy this code to the AGS.Native.

In AGS 4 this panel displays Fonts, which are instances of font files.

ericoporto:
> I can make it display outlines too, but then I will have to move automatic outline function to the Common, or copy this code to the AGS.Native.

Perhaps let's skip this here then.

ivan-mogilko:
> You can compare with this website rendering: https://fontdrop.info/#/?darkmode=true 

I can see that this preview skips missing glyphs, which I do not. I suppose that this can be done too.

ericoporto:
> @ericoporto I cannot understand what were you doing on that video. Please elaborate. Were you scrolling with keys, or wheel, and then with mouse? I only tested scroll with the mouse, and neither arrows keys nor wheel were working for me at all, so I assumed that they do not affect this scroll.
> What happens if you scroll only with mouse by clicking on the scrollbar?

I downloaded the latest build and it still reproduces. If I scroll using the mouse wheel it's like it doesn't scroll and it's buggy, and the characters repeat in different positions. Here is the scroll using only the mouse wheel.

https://github.com/user-attachments/assets/c70c4c29-06b6-45d4-97b2-9028b4cdd79b

If I click and hold the scroll with the mouse and move it, it looks like it's scrolling but the characters don't make any sense?


https://github.com/user-attachments/assets/8ed54e89-1b6d-4c42-81a8-79bdccd8bca1

Here is clicking in the empty area of the scrollbar to scroll

https://github.com/user-attachments/assets/8dc550a4-cbb0-4f09-a725-5de8650ab5fc

I think the general problem is you are printing characters that aren't available in the font so the preview doesn't look like it makes sense.

ivan-mogilko:
@ericoporto I suddenly got another idea. Cannot be that you have a ANSI text format set for your game?
Because I tried that right now, and it actually begins to display same 128-range of characters repeatedly when scrolling.

ivan-mogilko:
> There is still only a minor issue with the yellow border thing that appears to be not enclosing around the letter correctly, but I can't tell if it matters or not, made a video below.

Please tell what the font is this, and which are its settings?

>  My feel is that it would make sense to either not have that selection or the default selection be a third "Auto" option which would use the game default. I can't really tell why one would use a different mode than the game mode with that panel.

Okay, I added "Auto" selection which is on by default.
The reason why I wanted to give this selection is in case user wants to try if the font matches ascii or unicode mode.

ivan-mogilko:
> It happens with the agsfnt0.ttf I shared in the zip file above. Here's it's config

Alright, this happens when "TTF font adjustment" = "Resize ascender...".
The preview printing uses this setting, but the selection box does not account for it.

PR #2813, opened 2025-08-21
Support custom Font character spacing

ivan-mogilko:
This is a remake of #2767.
@edmundito told me that he's too busy now, so I based on his work with few modifications:
* Added SetCharacterSpacing to the IAGSFontRendererInternal interface, to avoid redundant casts to WFN and TTF renderers.
* Do not scale CharacterSpacing with SizeMultiplier in WFNFontRenderer, for consistency with LineSpacing and VerticalOffset, and to let more precise spacing setting.
* Serialize character spacing property in the existing "v360_fonts" data extension, because it had reserved integers.

Overall the feature is the same:
1. Fonts have CharacterSpacing property.
2. This property defines a global spacing value, in pixels, which is *added* to the default spacing between any pair of characters (kerning).
3. This property is NOT scaled with the font's point size, neither with "size multiplier" (used for bitmap fonts). This lets to have a more precise value.
4. This property may be negative, in which case the spacing is reduced (if it's reduced too much it may actually cause letters to appear in backwards order...)
5. This option appears to be already supported in AlFont library, so no changes were necessary there.
6. This option was recently applied to SDL_TTF library (https://github.com/libsdl-org/SDL_ttf/pull/578), so potentially available there too. (Although the SDL3 version of it; idk if they will backport to SDL2 version, but if not then we might do a custom fork).

### DIFF
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -857,9 +857,9 @@ HError GameDataExtReader::ReadBlock(Stream *in, int /*block_id*/, const String &
             finfo.AutoOutlineThickness = in->ReadInt32();
             finfo.AutoOutlineStyle =
                 static_cast<enum FontInfo::AutoOutlineStyle>(in->ReadInt32());
-            // reserved
-            in->ReadInt32();
-            in->ReadInt32();
+            // since kGameVersion_363
+            finfo.CharacterSpacing = in->ReadInt32();
+            in->ReadInt32(); // reserved
             in->ReadInt32();
             in->ReadInt32();
         }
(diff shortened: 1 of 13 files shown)


Common/game/main_game_file.cpp
```diff
@@ -857,9 +857,9 @@ HError GameDataExtReader::ReadBlock(Stream *in, int /*block_id*/, const String &
             finfo.AutoOutlineThickness = in->ReadInt32();
             finfo.AutoOutlineStyle =
                 static_cast<enum FontInfo::AutoOutlineStyle>(in->ReadInt32());
-            // reserved
-            in->ReadInt32();
-            in->ReadInt32();
+            // since kGameVersion_363
+            finfo.CharacterSpacing = in->ReadInt32();
+            in->ReadInt32(); // reserved
```
edmundito, line 862: Didn't know if this reservation was special hence why I originally I did not changei t.

Issue #2823, opened 2025-08-29
Hotspot.Name returns Hotspot.Description


brushfe:
**Describe the bug**
When I call a hotspot's name in a system.log report, I am given the hotspot's description instead.

**AGS Version**
Build 3.6.2.12

**To Reproduce**
1. Draw a hotspot in a room.
2. Give it unique text in the name and description fields.
3. Request the hotspot's name in by calling system.log("%s", hHotspot0.Name);

**Expected behavior**
The hotspot's name should appear where the hotspot's description is given.

**Screenshots**
[Screenshot of result and code](https://drive.google.com/file/d/1N9ydInORotlq4SsIyhb6udf8Fw7KSoEU/view?usp=sharing)

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 10

**Additional context**
It's my first bug report here, so please forgive any missing info or errors!

ivan-mogilko:
This is not a bug, there's no "Description" property in script, it's been always called "Name".

brushfe:
Thank you for investigating! I think I could've been more clear in my report, apologies.

I meant to say that hotspot.Name returns the text in the "Description" property, instead of the text in the "Name" property. The manual states that it returns the Name property, as is identical to ScriptName.

I'm not sure what the consequences are of different fixes, but changing the "description" property to "name" would certainly align with the code. It does seem weird to have a name and script name, but I think that sounds like the simplest way.

ericoporto:
The Name ("Script Name") is the name you do that makes sense in your code, you usually want something short, and it also is valid in script (so no spaces or weird characters). The ScriptName property is there if you want to give logs or do some special checking using strings directly to select something - probably using Custom Properties for any tag like behavior could be better, but anyway you could check it ScriptName contains something (e.g. "green") and then react accordingly.

 The description on the other hand is usually player visible. 

I guess these properties are similar to winforms Name and Text properties when adding something to a form using the designer.

ericoporto:
I wasn't explaining, I was thinking out loud. I am not sure what to do since it's been with these quirky differences between properties in Editor vs Script API for quite some time, and I have no idea what to do in a backwards compatible way.

I kinda am not sure if it's a good idea to change this under the 3.6.2 release since this seems like it could cause confusion for people with ongoing projects that aren't expecting major changes when upgrading their Editors.

ivan-mogilko:
[...]

PR #2824, opened 2025-08-30
Editor: consistent Script Name and Description/Name property labels

ivan-mogilko:
There's a long time issue that the properties that tell "script name" and "human text name" are called differently in the Editor and in the script.

In the old Editor the settings were labelled "Name" and "Script O-Name" (maybe it means "object name", as in OOP).
In the new Editor they got different labels: "Description" for a human name, and "Name" for the script name.
But the script commands stayed the same.

This is an endless source of confusion for beginners, and maybe not only beginners (for example, see recent #2823).

This PR assigns consistent labels to "script name" and "description" kind of properties in all the game entities in the Editor. This also makes them consistent with how they are called in script api. Only *labels* of properties are changed, leaving property identifying names as they are, so there should not be any compatibility issues and the project data format is not changed. The effect is purely in how the users will see these properties are called in the editor panel.

### DIFF
--- a/Editor/AGS.Types/Character.cs
+++ b/Editor/AGS.Types/Character.cs
@@ -12,7 +12,7 @@ namespace AGS.Types
     public class Character : ICustomTypeDescriptor, IToXml, IComparable<Character>
     {
         public const string PROPERTY_NAME_SCRIPTNAME = "ScriptName";
-        public const string PROPERTY_NAME_DESCRIPTION = "RealName";
+        public const string PROPERTY_NAME_DESCRIPTION = "Name";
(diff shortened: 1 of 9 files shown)


ivan-mogilko:
> actually it appears everything got a GetByName added which will make things more confusing if the Name now gets renamed as ScriptName,

It's already ScriptName in script, and this discrepancy existed before.

They both were added in this PR:
https://github.com/adventuregamestudio/ags/pull/2011

Somehow nobody mentioned that it's bad.

ivan-mogilko:
Plan to redo this for ags4 instead.

PR #2828, opened 2025-08-31
Write game version into saves, let read it when scanning/validating a save

ivan-mogilko:
Someone (I think this was @NPatch ?) asked me about this back when I was working on a save upgrade / prescan system, but I postponed this and then forgot, so this piece was not included in 3.6.2.

Earlier we have "game meta information" serialized in game data, and available in script (see #2773).
This PR adds this information into the save. When the save is restored this info does not overwrite the current game's info of course, but is read into RestoredSaveInfo struct, and may be accessed in "validate_restored_save" and used to check the save's compatibility.

Perhaps there may be other uses for this, since this data will be in saves it will be possible to implement more with it in the future.

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -283,6 +283,16 @@ void StrUtil::ReadStringMap(StringMap &map, Stream *in)
     }
 }
 
+void StrUtil::SkipStringMap(Stream *in)
+{
+    size_t count = in->ReadInt32();
+    for (size_t i = 0; i < count; ++i)
+    {
+        StrUtil::SkipString(in);
+        StrUtil::SkipString(in);
+    }
+}
+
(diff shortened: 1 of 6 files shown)


NPatch:
Yeah, I was looking into deleting the save games that were from a previous version of the game to avoid current version users from getting the classic old save crash. 

I honestly don't remember the full details, but I know that even though some template games don't show them (I think they were using a custom save/load menu or sth), any other game that uses the default way of handling save/load, will. 

To the best of my recalling the issue, I had written some code in AG that would read the headers of a save game and would try to figure out whether to delete or not but there was no game version saved within the save so we couldn't really compare. The comparison could only work for engine version, which was serialized into the save, but if the previous build was also produced with the same engine version, you can't differentiate between builds for the saves.

PR #2829, opened 2025-08-31
Editor: preview font and GUI texts with proper outlining

ivan-mogilko:
Complementary to #2808.
Ported a minimal text outline implementation to the Editor.
This lets see font preview and GUI texts with proper outlining, according to the font's properties.

For the font preview, which draws white text on black background, I chose a dark gray outline color for now. But maybe someone will have a better idea?

GUI texts always have black outline color. This seems to be the default engine behavior. I did not pay much attention to this, but apparently there's no way to configure outline color per GUI/control at all. There's a single setting that defines a color for ALL outlined texts in game. Perhaps this may be improved at some point, but that's a separate topic.

<details>
<summary>Screenshot:</summary>

font-preview-new3-outlines

</details>


ivan-mogilko:
>  Would it be too weird to have it use black background and white text, but when it has outline it instead use grey background, black outline and white text? I guess this complicates the code though.

That's not a lot of code, but I think it may be unexpected if preview colors change on their own with outline option.

PR #2841, opened 2025-09-11
AGS 4: View Frame events

ivan-mogilko:
This is an attempt to implement an idea that we discussed with @AlanDrake on AGS Discord some time ago.
The feature lets user bind "event triggers" to the selected View Frames, which would trigger "on frame event" in a object *that plays animation*. Currently the objects that may play a View-based animations are: Characters, Room Objects and Buttons.

Remade this PR after #2860.

1. For ViewFrame added a new "EventName" property of type String. It's an empty string by default. User may enter any arbitrary value there. This value does not correspond to real script events, but is rather a custom "tag". View editor displays a "thunderbolt" icon for the frames that have events.
2. Character, Room Object and Button have new OnFrameEvent events, that may be bound to a script function, just like any other events. The script function prototype for this event is:
```
function Character_OnFrameEvent(Character *theCharacter, int view, int loop, int frame, String eventName);
```
(Objects and Buttons have Character argument replaced with them of course.)
3. While in game, when an animation is playing, whenever there's a frame that has "event name" set, it will trigger a respective event in the object that plays this animation, calling a linked script function. If there's no script function, then it does nothing.
4. Character, Room Object and Button have `RunFrameEvent` function that lets trigger arbitrary frame's event (if it has one) for this object. The function is declared as:
```
// Triggers custom frame event, if this frame has one, and runs registered event handler. Returns if this frame has an event.
bool Character.RunFrameEvent(int view, int loop, int frame);
```
5. As an accompanying change: compile GUI events into the game the same way as Characters etc compile these since #2860. Similarly the engine handles these events internally following the same route as character etc events now. This is done to make it easier to add new events, and to have the same logic behind events firing.
6. OnFrameEvent callback can be run during blocking animations too, using "non-blocking" script thread, the one commonly used by the "repeatedly_execute_always". That's the only callback that support this behavior at the moment.

Trivial example of use:
```
function AnyCharacter_OnFrameEvent(Character *theCharacter, int view, int loop, int frame, String eventName)
{
    if (eventName == "step")
    {
        Region* region = Region.GetAtRoomXY(theCharacter.x, theCharacter.y);
        String surface = region.GetTextProperty("surface_type");
        if (surface == "wood") aStepWood.Play();
        else if (surface == "brick") aStepBrick.Play();
        // etc
    }
}
```

TODOS and ISSUES:

1. [RESOLVED] In the View Editor add a new icon for the frames that have a "event name" set. I think the good candidate is the "lightning" icon from events tab, or some derivative of it.
2. [RESOLVED] Add ViewFrame.EventName to be able to read it. Do we need to be able to set one? In which case it will have to be written to the game saves...
3. [RESOLVED]  Consider the `TriggerFrameEvent` function (as suggested in the comments below) to let users force it from script, as LockViewFrame apparently does not process the "frame change", and besides neither Object nor Button have ability to directly set a frame (edit: hmm, objects have SetView(view, loop, frame) though). Alternatively consider functions that set a frame to process the frame event & sound as if it was set by animation.
4. [RESOLVED] The biggest issue that I see currently is **blocking animation**. The current event calling mechanism in the engine works so that all events are scheduled during blocking action to be run after blocking action completes. That's very inconvenient, and makes this callback unusable in case of a blocking animation. Need to figure out how to solve this.


ivan-mogilko:
I remade this PR after #2860.
Now it supports OnFrameEvent for all 3 types: Characters, Room Objects and Buttons.
Updated the list of TODOs and known issues.

The biggest issue that I see currently is **blocking animation**. The current event calling mechanism in the engine works so that all events are scheduled during blocking action to be run after blocking action completes. That's very inconvenient, and makes this callback unusable in case of a blocking animation. Need to figure out how to solve this.

AlanDrake:
Can events be run on the repeatedly_excute_always thread?

ivan-mogilko:
Alright, I made it run on a non-blocking script thread when necessary. As noted above, this is not a problem directly, but the issue is to not break anything, in the convoluted events/callbacks logic of this engine. Also, this is going to be the first callback in the engine that has such behavior. Which makes things more inconsistent, unfortunately. 

I believe that there should be a plan to completely redo the script callback system in the future. Right now it's overcomplicated with duplicated entities, restrictions and strange rules.

ericoporto:
Does the difference between calling and execution in ChangeRoom, which also is scheduled and ran after is related to the mentioned issue or is that a different thing? 

Also the issues you described seems to affect more the RunInteraction call than the main goal here which were the events. Perhaps either leave as is to fix later or just hide the RunInteraction from the API for now?

ivan-mogilko:
I've added readonly ViewFrame.EventName property.
Do we need this to be settable in script? That would require adding event names to saves too.

ivan-mogilko:
> I am in doubts about this function's design and purpose. Does it need to trigger this for the current object's frame, or for any arbitrary frame? And then, does it only run the event if the selected frame has a EventName, or lets pass any arbitrary event name?

Alright, I added this:
```
bool Character.RunFrameEvent(int view, int loop, int frame);
```
which lets trigger event for any arbitrary frame, but only if it has event assigned.

PR #2843, opened 2025-09-11
Don't use alpha channel in colored gui controls in 2.*  games

### LINKED ISSUE #2819: Diamonds in the Rough - Background issue
**Game**
Diamonds in the Rough Demo

**AGS Version**
v2.71.894NMP vs. v3.6.3.0

**Download**
https://www.adventuregamestudio.co.uk/play/game/1006/

**Desktop**
OS: Windows 10

**Describe the bug**
Start a new game and click on the "T" on the upper left corner and the menu has no background in the newest ags version. The old one is fine, see below.

@[ivan-mogilko](https://github.com/ivan-mogilko) says this about the bug:

> Looked into "Diamonds in the Rough" issue. That's a very strange one. It appears that there are 2 GUIs, one with a "wooden board" background image, and then another with notes stamped to it. But original game was drawing the upper GUI with a transparent background, even though it has non-transparent solid background and border colors set. I could not figure out this behavior yet.
> Also, I have a strong feeling that I've seen this problem before, long time ago, and either tried to fix it or was discussing this with somebody.

**Screenshot**

Image

Old version:
Image

New version:
Image

ivan-mogilko:
Fix #2819

Investigating a strange issue with a 2.7 game "Diamonds in the Rough" (see #2819),
I found by experimentation that 2.72 and earlier engines did not fixup drawing color values with opaque alpha when painting primitives on 32-bit surfaces in 32-bit games.
This includes anything that is not a sprite: raw drawing, but also default gui controls (with no image assigned).
As a result, whenever a GUI background was considered to contain alpha channel (i.e. have a sprite with alpha as a background), then anything painted as simple colored primitive will become fully transparent.

This commit implements a backwards compatible mode for emulating this.

@Max833, this is a second attempt which uses same sort of fix, but narrows it down to guis. Last time it was global, and affected things that should not be changed, so I decided to make it safer. Did this as a PR too, so that it could be easier to test without breaking master branch...

### DIFF
--- a/Editor/AGS.Native/agsnative.cpp
+++ b/Editor/AGS.Native/agsnative.cpp
@@ -135,10 +135,6 @@ std::vector<GUISlider> guislider;
 std::vector<GUITextBox> guitext;
 std::vector<ViewStruct> newViews;
 
-// A reference color depth, for correct color selection;
-// originally was defined by 'abuf' bitmap.
-int BaseColorDepth = 0;
-
 
 struct NativeRoomTools
 {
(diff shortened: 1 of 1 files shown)


i30817:
superjazzman "works" (it's slow and I'm not quite sure it's correct) in software renderer mode.

PR #2860, opened 2025-09-25
AGS 4.0: Improve object events table: separate cursor & non-cursor events, allow unlimited cursor events

### LINKED ISSUE #2848: AGS 4: improve object event table, allow unlimited handlers for cursor modes
**Problem**

I noted that currently the event table for characters, room objects, etc is stored inconveniently, where the events are identified by a numeric index, and the same list mixes cursor (verb) events and non-cursor events. An example of the latter are Hotspot's "Stand on" and "Mouse over" events, while Regions have only non-cursor events. It seems like some of those were added in varying order historically, which led to a mismatching indexes. Indexes mismatch not only between event table and the list of game cursors, but also between different types of objects. (Don't remember by heart, but for example, a "Look at" event for character may be at index A, and same event for object may be at index B.)
This is quite inconvenient, and makes it difficult to keep things consistent when expanding the events (so this is technically related to #2841).
Also, this makes it difficult to let have events for each custom game cursor. Right now AGS only supports events for 2 custom cursor modes, and there have been user requests to support more.

**Proposal**

1. Split event table into 2 sections:
* - cursor (verb) events;
* - other events (maybe there's a good name for this, idk... "normal events"?)
2. Cursor events are identified by the cursor index internally, but are using cursor names when displayed in property grid. They generate script functions using actual cursor names too. There should be a event for each cursor in game, which is marked as "Standard cursor" (this property has a historical meaning of cursors that can be cycled to select a action verb).
3. Other events are identified by name, similar to how events are ided in gui controls ("OnClick", etc).
4. This distinction is used when serializing events in a project file.
5. In the compiled game the events are written also as two separate tables, each with its own identification.
6. RunInteraction script function can only run verb-related events (well, that's how it's done now).
7. Have a rule to add script functions for triggering any other event for object types that support ones.

ivan-mogilko:
Resolve #2848, nice to have for #2841

### Purpose:

1. Have a more convenient event table for adding more non-cursor related events in objects, because right now they are mixed with cursor interaction events, and in non-consistent order.
2. Allow any number of custom cursors generate events in objects.

### Changes:

1. In the Editor the events are now divided into two parts: cursor interactions and common events. In the code InteractionSchema is now a single static instance, used for all the game objects, and it stores events matched with the respective cursor ID (previously was created for each object type, and their indexes were all messed up). Common events (not tied to a cursor) are defined as object's properties, similar to how it's done in GUI classes.
2. Moved out all the non-cursor events from InteractionSchema, and made them class properties. For example, all the Room events are now plain properties. Editor makes necessary conversion when loading old projects or old rooms.
3. Interaction events are serialized as before, having a numeric index, but now it matches cursor ID, rather than an arbitrary made up index. Common object events are serialized as event-name/script function maps. Although they are resolved to a indexed list in the engine at runtime, for a simpler lookup.
4. OnAnyClick is now a explicit event (in related objects), and will exist even if there are no cursor events.
5. InventoryItem now has "OnAnyClick" event, similar to Characters etc. It had only "Other click" event before, which did not make much sense to have.
6. Cursors have new properties: CreateEvent (boolean), EventLabel and EventFunctionName. If CreateEvent is set to true, then InteractionSchema will generate event for it, EventLabel is used as a "display name" for event, and EventFunctionName is used as a script function suffix when creating a function. If either of these are left empty, then Cursor's Name is used instead (converted to a script-acceptable name). *Any* cursor may create event.
7. Making more (or less) cursors with CreateEvent updates the event table immediately.
8. In the engine: modified the event table structures and code accordingly.

Example screenshot:
neweventstable2

### Known Issues and Possible TODOs

1. There's a discrepancy in how different event properties display function names in them. The events that are generated by InteractionSchema do display in normal font, while events that are declared explicitly as properties make them **bold**, as if they are different from a "default value", although neither have default value attribute. I'd like to figure out how it's possible to not use bold font there, because that looks annoying IMHO.
2. In regards to "Any click", apparently right now the rule is that it runs for ANY cursors except Walk when *clicking* on object, and even for Walk when using RunInteraction function in script. I wonder if we need another property for the Cursor telling whether to be included into "Any click" list or not.
3. I think that if you rename a Cursor, then any function assigned to its event in objects will be lost. Need to find a way to keep that value reassigned to the new event name. NOTE: in theory we could even go further and rename the function in script using a new suffix.
4. Update GUI to have a similar events table in all aspects in code. This involves how these are written to the compiled game, stored in engine structs, etc.
5. Update Editor Tests.
6. It might make sense to plan script functions for triggering non-cursor events directly. RunInteraction normally runs only cursor events, except for Regions, where it allows to pass undeclared constants 0,1,2 corresponding to special Region events. However, other objects have events that cannot be run directly, such as Hotspot (stand on, mouse over) and of course Room (idk if it makes sense to do it for the Room though). This should not be a part of this PR, but maybe open a feature ticket for the future.
7. An issue of **standard cursor roles**. There's a number of standard roles hardcoded in the engine, that trigger actions. Note I am speaking of automatic built-in behavior, not ones scripted in templates. These roles are:
    - Walk - makes player character walk;
    - Look - triggers look in inventory (when game setting is set); also triggers "walk to hotspot in look mode";
    - Interaction - triggers interact in inventory; maybe triggers "walk to hotspot" (need to double check);
    - Use Inventory - used to display active inventory item;
    - Pointer - used in some states, such as Dialog Options;
    - Wait - used during blocking actions.
<br>
Right now these modes are sort of "protected", since Editor not allows you to delete them (and few others, which is no longer necessary). But you still make edit their names, thus making things look weird.
My thinking is that there are two ways this situation may be made more organized:
    - a) make these cursor modes completely fixed, that is - forbidden to delete AND their name properties readonly. This seems to be the fastest solution.
    - b) introduce new cursor property, say, Role, with this hardcoded list of roles. This will be more complicated to implement (would require engine to lookup for a cursor of certain role to set a mode instead of using a enum). But on a good side this will let user full freedome in the cursor list, and remove even Walk if they dont need it in their game.
I'd like to do either of this in this PR.

### Testing

1. When loading old projects, Cursors have their CreateEvent property set according to the standard old behavior (only those have it that normally corresponded to object interactions). Character and Inventory item events are remapped accordingly on game load. Room events, and room object / hotspot / etc events are remapped when the old room is loaded.
2. Changing CreateEvent in cursors should update event table of all objects that have cursor events.
3. Events should work in game just like before: call a script function that is assigned to them. This relates to both cursor events and common events. The list of events is rather big though... but would be nice to double check all of them.
4. "Any click" event should be run on any cursor interaction, except Walk when clicking. (Might double check how the old engine behaved when RunInteraction is run with eModeWalk or eModeWait.)


ivan-mogilko:
I found a nasty problem, the "room version" field is kept 16-bit all this time, therefore it cannot fit our new version number format (values like 3060200 or 4000000). Fortunately, there were no room updates that would require a version up in 3.6.* branch so far, but there have been in 4.*, and it's a pure luck that this did not cause any real trouble so far.

This got to be fixed...

ivan-mogilko:
I have not tested the runtime thoroughly yet.

ivan-mogilko:
@AlanDrake I've fixed as many problems that I could find in a simple test game.

There's one clear issue that is not resolved, it's "unhandled event" function. As may be seen [in the manual], this function has different interpretation of the cursor mode number depending on a object type. But in this PR all objects use same interaction indexes internally, so "unhandled event" will receive "wrong" numbers.

Either I need to return the mapping to these special numbers exclusively for "unhandled event", or we agree that "unhandled event" receives correct cursor modes.

AlanDrake:
> @AlanDrake I've fixed as many problems that I could find in a simple test game.
> 
> There's one clear issue that is not resolved, it's "unhandled event" function. As may be seen [in the manual](https://adventuregamestudio.github.io/ags-manual/Globalfunctions_Event.html#unhandled_event), this function has different interpretation of the cursor mode number depending on a object type. But in this PR all objects use same interaction indexes internally, so "unhandled event" will receive "wrong" numbers.
> 
> Either I need to return the mapping to these special numbers exclusively for "unhandled event", or we agree that "unhandled event" receives correct cursor modes at all times.

I feel unhandled_event should be using the enums everyone expects, LocationType and CursorMode.
Just looking at the inconsistent table in the manual makes me angry.

ivan-mogilko:
> Edit: right... LocationType doesn't have anythign for inventory

I suppose that could be added, although whether to return one from GetLocationType remains a separate question.

AlanDrake:
It no longer crashes when hovering, but there's still something wrong with `room_leave` events, which are not being executed for some reason.
Other room events appear to be working fine.

PR #2871, opened 2025-10-09
AGS 4: AnimatedOverlay

### LINKED ISSUE #2811: Animated Overlays
There have been multiple user requests to make animating overlays possible with a scripting command. Such ability would be quite useful, as Overlays are currently the only dynamically created graphic objects, and it seems unfair to force users to script animation themselves (with all its options). But it somewhat conflicts with the idea to have a "minimal graphical object" in the engine.

There are two concerns here: object design and performance.

1. In regards to performance (I'll touch this first), if Overlays can optionally animate, then we need to add extra state check for them during the game update. It's hard to predict exactly, but assuming there may be thousands of overlays in extreme cases, when they are used to display a complex scene, then extra unnecessary checks may slow things down a bit.
This may be solved in ECS-style way: by having a separate list with references to the overlays that animate. When overlay is ordered to animate we place its reference into this list, when it finishes animating we remove its reference from this list. The engine passes the main overlays collection only when drawing them, and passes this "animating" list when updating overlay animations.

2. In regards to the object design, previously there was stated an idea that we need some kind of a "basic graphic object", which could be used either as a parent class or other objects component (part of an actual object type) in some future class hierarchy. Having an animated object would require to have Animate command, as well as (readonly) View, Loop, Frame properties, and maybe more. Thus an object becomes more than a graphic, but also something having a logical state and behavior.
If we still need that "primitive graphical object", then I see two options here:
a) make Overlay animateable, but then extract another, more basic type from it whenever we need that. That's the most straightforward and simple solution now, but may require additional changes later.
b) add a new class called e.g. AnimatedOverlay which extends Overlay and have Animate controls.

Internally there are various ways this may be implemented; i.e. in either case it may be still stored separately as an Overlay, and something that controls its animation (similar to how there's a separate Button animation controller) - this way there will be no need to change scene drawing code.

ivan-mogilko:
Resolve #2811

Implement a new type called AnimatedOverlay which is an extension of Overlay. I decided to go with a separate type, because there's a concern about keeping existing Overlay simple. Also, there's a number of cases where overlay is returned for a built-in object (such as speech text, portrait, etc), which probably should not have any extra functionalities in them.

AnimatedOverlay is declared as:

```ags
builtin managed struct AnimatedOverlay extends Overlay
{
  /// Creates a AnimatedOverlay object on the screen layer, optionally using a initial sprite
  import static AnimatedOverlay* CreateAnimated(int x, int y, int slot = 0, bool pauseWithGame = true); // $AUTOCOMPLETESTATICONLY$
  /// Creates a AnimatedOverlay object on the room layer, optionally using a initial sprite
  import static AnimatedOverlay* CreateRoomAnimated(int x, int y, int slot = 0, bool pauseWithGame = true); // $AUTOCOMPLETESTATICONLY$
  /// Animates the overlay using the specified view loop.
  import void Animate(int view, int loop, int delay, RepeatStyle=eOnce, BlockingStyle=eNoBlock, Direction=eForwards, int frame=0, int volume=100);
  /// Gets whether the overlay is currently animating.
  import readonly attribute bool Animating;
  /// Gets the current frame number during an animation.
  import readonly attribute int  Frame;
  /// Gets the current loop number during an animation.
  import readonly attribute int  Loop;
  /// Gets whether the overlay's animation is paused when the game is paused.
  import readonly attribute bool PauseWithGame;
  /// Gets the current view number during an animation.
  import readonly attribute int  View;
};
```

The `Animate` command and properties are mostly mimicking existing Button example.
The standing out parameter is "PauseWithGame". While Button animation do not pause when the game gets paused, because the GUI is not paused at all, but Overlays may be used for various purposes, and be created in room or screen layer. So I decided to make this a parameter passed when creating animated overlay.

Internally, AnimatedOverlay consists of a normal overlay itself, and a additional object which describes animation and controls its flow, connecting to overlay. The connection between these two is done using reference tables. This ensures that overlays own update is not affected performance-wise. Overlay animations are updated separately, and they just change referenced overlay's graphic value.

I'd like to hear any comments on the API. For instance, I might perhaps make "pause with gui" a enum, for better clarity.

### DIFF
--- a/Editor/AGS.Editor/Resources/agsdefns.sh
+++ b/Editor/AGS.Editor/Resources/agsdefns.sh
@@ -669,6 +669,28 @@ enum StringSplitOptions {
 };
 #endif // SCRIPT_API_v400
 
+#ifdef SCRIPT_API_v350
+enum GUIPopupStyle {
+  eGUIPopupNormal = 0,
+  eGUIPopupMouseYPos = 1,
+  eGUIPopupModal = 2,
+  eGUIPopupPersistent = 3
+};
+#endif // SCRIPT_API_v350
+
+enum BlockingStyle {
+  eBlock = 919,
+  eNoBlock = 920
+};
+enum Direction {
+  eForwards = 1062,
+  eBackwards = 1063
+};
+enum WalkWhere {
+  eAnywhere = 304,
+  eWalkableAreas = 305
+};
+
 internalstring autoptr builtin managed struct String {
   /// Creates a formatted string using the supplied parameters.
   import static String Format(__format const string format, ...);    // $AUTOCOMPLETESTATICONLY$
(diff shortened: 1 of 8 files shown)


ivan-mogilko:
> About the API, the pause sensitivity is with the PauseGame () function and with the GUIs that have the config "when visible pause game" only?

Yes. This parameter can be used to keep animated overlay on while the "pausing" guis are displayed, because normally animations are not updated when the game is paused.

Issue #2869, opened 2025-10-09
Editor: Custom Properties node in the Project Explorer, and import/export commands


ivan-mogilko:
1. Add a Custom Properties node into the project explorer, clicking on which opens schema editor. This will make it faster to access the schema editor, because right now one has to go to some object, open its Custom Properties editor, and press "edit schema" button there.
2. Implement "export schema" / "import schema" that write schema into a standalone xml file and read it back from such file.

In regards to the schema file: the schema should be written exactly how it's written into the Game.agf, with a header having format version in it (equals the current project version).

ericoporto:
Do we need to add a `CustomPropertyAppliesToAny` or redo the enum as flags, to be able to open a generic schema editor?

ericoporto:
Yeah, I am, sorry. I am trying to understand why the last line of this function exists

https://github.com/adventuregamestudio/ags/blob/9793dedbffebe83cb317cd1b5c4a92e3b336cbb5/Editor/AGS.Editor/GUI/CustomPropertiesEditor.cs#L28-L34

I guess this is some way to refresh the property grid after the schema has been edited?

AAHH, crap, it's the property grid inside the custom property editor for that type. I got it.

ivan-mogilko:
Initially the assumed use case is to copy schema from one game to another, as a "template", but as I think more about your question, perhaps it may make sense to have a schema prepared for a certain module too. In which case such imported schema must not override other things.

OTOH I do not think that's useful to have properties with modified names (like added numbers), because the name is the most important part of the custom property, and having their names changed will defeat their purpose in most times.

Therefore, I suppose, there can be 2 (or 2.5) import modes, and user makes a choice between them when importing:
- overwrite all schema with an imported one
- append imported schema, adding new properties; and for the properties that have matching names:
    - skip existing
    - replace existing

Note that "replacing" means overriding type and description, as these are only things that a property definition has after its name.

About the export; initially I assumed full export. Of course ideally it would be nice to let users select properties which have to be exported, if that's not too much work. I suppose that in such case there could be a dialog with a check-list of custom properties, and user may check out those that it does not want to export.

Then program makes a new temporary custom properties schema with a copy of selected property definitions, and serializes that instead of the one in game.

PR #2897, opened 2025-10-19
Editor: add Custom Properties node in Explore Project

ericoporto:
it uses the schema editor modal instead of a pane for simplicity for now.

Partial of #2869

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.csproj
+++ b/Editor/AGS.Editor/AGSEditor.csproj
@@ -156,6 +156,7 @@
     <Compile Include="Entities\CharacterRoomChangedEventArgs.cs" />
     <Compile Include="Entities\CompilationStepArgs.cs" />
     <Compile Include="Entities\CompileWarningEx.cs" />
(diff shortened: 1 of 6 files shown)


ericoporto:
@AlanDrake I am here in need of a better icon. My original idea was “table from properties grid with a settings gear in the front”, but not sure if I did it well.

AlanDrake:
The icon looks fine, works both for light and dark themes

ericoporto:
@ivan-mogilko hey, sorry, a lot suddenly came up and I couldn't finish this. If you want to finish this I wouldn't mind, please pick the first two commits into a PR of yours. Feel free to close this PR.

The last one commit I was trying to add some diff viewing panel but couldn't figure it out how to go about it, I wanted to enable importing not everything regarding custom properties but at the same time I couldn't come up with a design that was both generic but still worked. Disregard that commit. I will leave it here in case its an useful reference in the future.

ivan-mogilko:
Implemented as #2931

PR #2917, opened 2025-12-15
Basic frame model and colored looks support for all GUI Controls

ivan-mogilko:
This adds a number of properties that configure gui controls in terms of looks and colors.

### Parent GUIControl class
has new properties that define the basic frame model:
* BackgroundColor
* BorderColor
* BorderWidth - border thickness
* SolidBackground - tells to fill or not fill control's rect with background color
* ShowBorder - tells to draw or not draw a border around control
* PaddingX - padding is a distance (in pixels) between control's left and right borders and internal contents.
* PaddingY - distance (in pixels) between control's top and bottom borders and internal contents.

All the above properties can be set for any control. Effect of Padding depends on control type, because each control type defines what is "internal contents" on its. own. Slider is different from the rest in this regard, at least in current implementation, it does not refer to the padding at all.

### GUIButton class
has new properties that define the button colors in 3 different states:
* ColorStyle - determines which colors are actually going to be used, and how (see below)
* MouseOverBackgroundColor - background color when the button is hovered by a cursor
* PushedBackgroundColor - background color when the button is being pushed down
* MouseOverBorderColor - border color when the button is hovered by a cursor
* PushedBorderColor - border color when the button is being pushed down
* ShadowColor - second color used for border to emulate 3D looks
* MouseOverTextColor - text color when the button is hovered by a cursor
* PushedTextColor - text color when the button is being pushed down

ColorStyle's values are defined by enum called ButtonColorStyle, and are following:
1. Default (eGUIButtonDefault in script) - All colors are fixed at their "normal" states, button has a 3D effect, using BorderColor and ShadowColor.
2. Dynamic (eGUIButtonDynamic in script) - Background and Text colors change along with the state, border color uses BorderColor and ShadowColor for a 3D effect.
3. Dynamic Flat (eGUIButtonDynamicFlat in script) - Background, Border and Text colors change along with the state, and border uses only 1 color, so there's no 3D effect.

For buttons with images you can still use Dynamic style if you keep Text drawn by button itself (not a part of an image), and you like the text color to change with state.

### GUISlider class
has couple of the new properties:
* HandleColor - main color of the handle
* ShadowColor - secondary border color, used for a 3D effect.

### DIFF
--- a/Engine/ac/button.cpp
+++ b/Engine/ac/button.cpp
@@ -423,6 +423,9 @@ int Button_GetTextPaddingHorizontal(GUIButton *butt)
 
 void Button_SetTextPaddingHorizontal(GUIButton *butt, int pad)
 {
+    // Original Button.TextPadding included button's border
+    if (loaded_game_file_version < kGameVersion_363_04)
+        pad--;
     butt->SetPaddingX(pad);
 }
 
@@ -433,6 +436,9 @@ int Button_GetTextPaddingVertical(GUIButton *butt)
 
 void Button_SetTextPaddingVertical(GUIButton *butt, int pad)
 {
+    // Original Button.TextPadding included button's border
+    if (loaded_game_file_version < kGameVersion_363_04)
+        pad--;
     butt->SetPaddingY(pad);
 }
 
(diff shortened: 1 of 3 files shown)


ivan-mogilko:
Just for the reference, here is how the similar properties are called in WinForms (since we emulated .NET framework naming sometimes in the past):
winforms-buttonflatcolors

PR #2918, opened 2025-12-26
Added ability to detect mouse click on edges

craigharman:
## Overview

I wanted a quick consistent way of adding "fast room" changes by double clicking on a room edge. Although there are some workarounds they involve a lot of code and/or work on each room. This allows processing of the mouse click event on the room and calling the appropriate edge function.

## Summary of Changes

  1. Engine Constants (Engine/ac/runtime_defines.h)

     - Added LOCTYPE_EDGE = 5 constant

  2. Core Detection Logic (Engine/ac/game.cpp)

     - Added edge detection in __GetLocationType() after all other location checks
     - Edges are only detected when no character, object, or hotspot is found (lowest priority)
     - Returns edge index in getloctype_index: 0=left, 1=right, 2=bottom, 3=top

  3. Click Processing (Engine/ac/global_game.cpp)

      - Added edge handling in RoomProcessClick() to route edge clicks to run_unhandled_event()

  4. Unhandled Event Support (Engine/script/script.cpp)

     - Added LOCTYPE_EDGE case mapping to evtype = 6
     - Added filter rule for edge clicks

  5. Script API (Editor/AGS.Editor/Resources/agsdefns.sh)

     - Added eLocationEdge = 5 to the LocationType enum

  6. Tests (Engine/test/loctype_test.cpp + Engine/CMakeLists.txt)

     - Created unit tests for edge detection logic
     - Tests verify constant value, each edge detection, inside-room detection, and corner priority

  Script Usage Example

```c#
  function on_mouse_click(MouseButton button) {
      LocationType loc = GetLocationType(mouse.x, mouse.y);

      if (loc == eLocationEdge) {
          // Edge detected - can be handled here or via ProcessClick
      }
  }

  function unhandled_event(int evtype, int evnt) {
      if (evtype == 6) {
          // evnt is the edge index: 0=left, 1=right, 2=bottom, 3=top
          Display("Clicked edge %d", evnt);
      }
  }
```

### DIFF
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -647,6 +647,7 @@ endif()
 if(AGS_TESTS)
     add_executable(
         engine_test
+        test/loctype_test.cpp
         test/scsprintf_test.cpp
         test/systemimports_test.cpp
     )
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1494,6 +1494,27 @@ int __GetLocationType(int xxx,int yyy, int allowHotspot0) {
     if ((winner == LOCTYPE_HOTSPOT) && (!allowHotspot0) && (hsat == 0))
         winner = 0;
 
+    // Check edges (lowest priority - only if nothing else found)
+    if (winner == 0) {
+        int edge_detected = -1;
+
+        // Check edges using room coordinates (xxx, yyy already converted)
+        if (xxx <= thisroom.Edges.Left)
+            edge_detected = 0; // Left edge
+        else if (xxx >= thisroom.Edges.Right)
+            edge_detected = 1; // Right edge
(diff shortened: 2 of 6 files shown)


craigharman:
Thanks for your response @ivan-mogilko - let me put an example together and send something through to you which hopefully can address a majority of your queries.

craigharman:
Thanks for the detailed review, I don't think this will work given the changes you reference for 4.x. I still have the issue of how to trigger the edge event from the code you provided but that's a separate issue.

ivan-mogilko:
After writing the above example, I realized that we forgot to document SendEvent, so it was not present in the manual (it was only noted in Changes list).

Here's my recent fix to the manual source, if needed:
https://github.com/adventuregamestudio/ags-manual-source/commit/5b826493b6ba789ef5cc07154db13aefff6494c7
(this will be included along with the next engine release)

PR #2928, opened 2026-02-05
Editor: implement options for including/excluding certain script lines in Translation source

ivan-mogilko:
Sometimes users wish that particular text lines from script do not end up in translation source. While it's not easy to create fully customizable rules for gathering translations in the existing system, but I wanted to give users few options covering most common cases.

This PR adds following 3 options to the General Settings:
* **"Include script texts with prefix"** - includes ONLY script text lines which contain this prefix. This refers to both normal scripts and dialog scripts. Other texts (from object properties) are included by default. This is the most radical option, it requires user to add a prefix to virtually any text in speech, etc, to make it included into translation. I am not convinced myself whether this will be much useful, but to give an example, one can use prefix "&" to ensure that only voiced speech lines are going to be added to TRS.
* **"Exclude script texts with prefix"** - excludes any script text lines which contain this prefix. Easy to use option if you want to skip debugging lines, just prefix these with some kind of "[DBG]" tag.
* **"Exclude script function calls"** - excludes any texts found within a script function call. In theory this allows to exclude both standard and custom debug and logging functions, such as System.Log.

Also, added **"Add Text property values to the game's translation file"** checkbox to Custom property editor. This tells if the Editor should add this property's values (default value and individual values from objects) into the translation. NOTE: this does *not* affect whether the property value may be translated at runtime, as that's not done automatically anyway.

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -120,8 +120,9 @@ namespace AGS.Editor
          * 3.6.3.5        - Renamed ShadowColor -> BorderShadeColor.
          * 3.6.3.6        - Character.BlockingRectangle, RoomObject.BlockingRectangle.
          *                  expose Character.Transparency, Baseline, RoomObject.Transparency.
+         * 3.6.3.8        - New Translation settings.
         */
-        public const int    LATEST_XML_VERSION_INDEX = 3060306;
+        public const int    LATEST_XML_VERSION_INDEX = 3060308;
         /*
          * LATEST_USER_DATA_VERSION is the last version of the user data file that used a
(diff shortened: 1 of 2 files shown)


ericoporto:
Do you do anything to clear the difference between Windows and Linux line endings? I think it may be something like this but I would need to test it locally - when it is parsing the `TestScript` contents, the line ending depends on the git config. I would either avoid using `@"` or always pass something on the string using it to ensure the line endings are homogeneous.

ericoporto:
Cool that it worked! I think some other tests turn the lines in an array of strings somewhere when parsing and this disambiguates this - at least I think the preprocessor did this in its tests. Anything that comes up with the CI please tag me so I can take a look, it creates a notification to me!

PR #2930, opened 2026-02-09
Tool: spritepak

ivan-mogilko:
Implements a spritepak standalone tool, meant for unpacking and (re)packing sprite file.
The tool is used from a command line, and has following parameters:

```
spritepak <COMMAND> <SPRITEFILE> [<WORK-DIR> | <DEST-SPRITEFILE>] [OPTIONS]
```

Commands are:
```
  -c, --create           create a spritefile, gathering the files from the
                         input directory.
  -e, --export           export (extract) files from the existing spritefile
                         into the output directory.
  -l, --list             print spritefile's table of contents.
  -q, --info             print quick info about spritefile.
  -w, --rewrite          rewrites spritefile contents into another spritefile,
                         optionally using different storage options.
```

Command options:
```
  -n, --index <indexfile>
                         specifies the index file to read or write, depending
                         respectively on the current command.
  --out-index <indexfile>
                         when rewriting a spritefile, specifies new index file.
  -p, --pattern <file pattern>
                         when creating the new spritefile, or exporting one,
                         use the given pattern for individual image files,
                         either with or without file extension. The pattern
                         may contain following placeholders:
                           * %N% - sprite number
                         If no pattern is provided, the program will use
                         "spr%N%" pattern by default. If neither pattern
                         nor extension is provided, then ".bmp" will be used.
  -s, --storage-flags <flags>
                         when creating the new spritefile, use additional
                         storage options, defined using a hexadecimal bitset:
                           * 0x01 - optimize storage size when possible;
                             e.g. write 16/32-bit images as 8-bit images with
                             palette (only when this achieves less space).
  -z, --compress <type>  when creating the new spritefile, use compression:
                           * none
                           * rle
                           * lzw
                           * deflate
  -v, --verbose          print operation details
```

Although formally working, this tool's functionality has limited uses, because
1. It currently does not support packing sprite file by reading import instructions (e.g. from Game.agf), only packs numbered image files as-is. Therefore it cannot substitute the process integrated into the Editor. NOTE: i'm not yet sure if the full import feature should eventually be the part of this tool, or a separate one, meant strictly for that purpose.
2. ~There's no support for exporting image formats with alpha channel, because we don't have available C++ code in the engine. While the BMP format should theoretically support alpha channel, there's some restriction in this regard in the LoadBMP/SaveBMP functions, which afaik originally come from Allegro 4 library. Maybe that restriction can be lifted by adjusting the code~. **Partially fixed by supporting 32-bit ARGB BMP files** But the best option is adding PNG load/save support (and this was requested by users for the engine before).

In its current state this tool is best used for diagnostic and test purposes:
- checking the contents of the existing game's sprite file
- recreating the existing spritefile with different compression parameters (this works well, because there's no intermediate serialization of image files, only processing of plain pixel data in memory).

### DIFF
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -395,7 +395,7 @@ void SpriteCache::InitNullSprite(sprkey_t index)
     _spriteData[index] = SpriteData();
 }
 
-int SpriteCache::SaveToFile(const String &filename, int store_flags, SpriteCompression compress, SpriteFileIndex &index)
+HError SpriteCache::SaveToFile(const String &filename, int store_flags, SpriteCompression compress, SpriteFileIndex &index)
 {
     // Gather a list of sprites;
     // the list contains pairs, where first element tells whether the sprites
--- a/Common/ac/spritecache.cpp
+++ b/Common/ac/spritecache.cpp
@@ -337,9 +337,9 @@ Bitmap *SpriteCache::LoadSprite(sprkey_t index, bool lock)
         return nullptr;
     assert((_spriteData[index].Flags & SPRCACHEFLAG_ISASSET) != 0);
 
-    Bitmap *image{};
-    HError err = _file.LoadSprite(index, image);
-    if (!image)
+    PixelBuffer pxbuf;
+    HError err = _file.LoadSprite(index, pxbuf);
+    if (!pxbuf)
     {
         Debug::Printf(kDbgGroup_SprCache, kDbgMsg_Warn,
             "LoadSprite: failed to load sprite %d:\n%s\n - remapping to placeholder.", index,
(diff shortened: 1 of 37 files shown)


ericoporto (dismissed):
There is just a typo in the CMakeLists.txt in Tools directory, it should be `../Common/gfx/bitmapdata.cpp` without the underscore.

ericoporto:
Hey, I am still struggling with [what I mentioned in this comment](https://github.com/adventuregamestudio/ags/pull/2930#issuecomment-3890985243), for whatever reason, it doesn't build with CMake in my IDE (CLion) when I try it. I know it somehow builds in the CI, but I don't know why.

Our cut-down allegro library is really small, I would like to just use it in our CMake build, because then this would work fine, at least when I build it locally here.

PR #2958, opened 2026-04-07
Use lexographical comparison for script strings and sortable objects

ivan-mogilko:
Following a user request, this implements a lexographical comparison for script strings.

By default, when comparing strings for sorting purposes (so not for equality, but to get which string gets earlier order), the character codes are used, but this is not always suitable for characters in unicode range, because there may be alphabets which use part of the standard latin range (A-Z) and then few more letters with much higher codes (easiest example is letters with accents such as ''À"). For the game made in non-English language or a translation it may be desired to have a sorting based on the language's alphabet instead.

For example, if we have following strings:
* Atext
* Btext
* Ztext
* Àtext
* Čtext

the above would be the character-code-reliant order, but the alphabetically correct order would be:
* Atext
* Àtext
* Btext
* Čtext
* Ztext

What this PR does:
1. Added "GameTextLanguage" property in General Settings (actually, backported one from AGS 4.0, where it was used for generating PO translation source files). It's supposed to be initialized with a language name used for locales, such as "en", "en_US" and so forth.
2. Support "#Language" option in TRS as well, with values set by the same principles.
3. At runtime, the engine will set a locale name for game texts by combining the language setting and ".utf8" postfix, so e.g. "en_US.utf8".
4. This locale will be used to do string collation for game strings. Currently this is implemented in:
* String.CompareTo function
* Dictionary object (when it's created with "sorted" style)
* Set object  (when it's created with "sorted" style)
5. This locale changes when game uses a translation, using translation's own setting.
6. For backwards compatibility, or when the setting is not present, the "default locale" will be used instead.

### DIFF
--- a/Common/ac/gamesetupstruct.h
+++ b/Common/ac/gamesetupstruct.h
@@ -30,10 +30,10 @@
 #include "game/main_game_file.h" // TODO: constants to separate header or split out reading functions
 
 
-using AGS::Common::UInteraction;
-using AGS::Common::UInteractionEvents;
-using AGS::Common::InteractionVariable;
-using AGS::Common::HGameFileError;
+using Common::UInteraction;
+using Common::UInteractionEvents;
+using Common::InteractionVariable;
+using Common::HGameFileError;
 
 
 // TODO: split GameSetupStruct into struct used to hold loaded game data, and actual runtime object
@@ -56,9 +56,9 @@ struct GameSetupStruct : public GameSetupStructBase
     GameDataVersion   filever = kGameVersion_Undefined;
     Common::String    compiled_with; // version of AGS this data was created by
     char              lipSyncFrameLetters[MAXLIPSYNCFRAMES][50] = {{ 0 }};
-    AGS::Common::PropertySchema propSchema;
-    std::vector<AGS::Common::StringIMap> charProps;
-    AGS::Common::StringIMap invProps[MAX_INV];
+    Common::PropertySchema propSchema;
+    std::vector<Common::StringIMap> charProps;
+    Common::StringIMap invProps[MAX_INV];
     // NOTE: although the view names are stored in game data, they are never
     // used, nor registered as script exports; numeric IDs are used to
     // reference views instead.
@@ -86,7 +86,9 @@ struct GameSetupStruct : public GameSetupStructBase
     int               numCompatGameChannels = 0;
 
     // A dictionary of semi-arbitrary game info properties: title, developer's name, etc
-    AGS::Common::StringMap GameInfo;
+    Common::StringMap GameInfo;
+    // Game text language definition, in the form of locale name ('en', 'en_US' etc)
+    Common::String GameTextLanguage;
     
     // TODO: I converted original array of sprite infos to vector here, because
     // statistically in most games sprites go in long continious sequences with minimal
--- a/Common/game/main_game_file.cpp
+++ b/Common/game/main_game_file.cpp
@@ -535,6 +535,11 @@ void UpgradeGame(GameSetupStruct &game, GameDataVersion data_ver)
     {
         game.options[OPT_GAMEFPS] = 40; // default to 40 fps in the older games
     }
+
+    if (game.GameInfo.count("text_lang") > 0)
+    {
+        game.GameTextLanguage = game.GameInfo["text_lang"];
+    }
 }
 
 void UpgradeDialogs(GameSetupStruct &game, LoadedGameEntities &ents, GameDataVersion data_ver)
--- a/Common/util/directory.h
+++ b/Common/util/directory.h
@@ -18,9 +18,11 @@
 #ifndef __AGS_CN_UTIL__DIRECTORY_H
 #define __AGS_CN_UTIL__DIRECTORY_H
 
+#include <locale>
 #include <memory>
 #include <regex>
 #include <stack>
+#include <stdexcept>
 #include <vector>
 #include "platform/platform.h"
 #include "util/string.h"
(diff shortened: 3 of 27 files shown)


ericoporto:
I don't know if the "locale" names are the same between C# and C++, but I wonder if there is a way to check if the `GameTextLanguage` is set to something that actually is a valid locale or an invalid value (to catch a typo). And I can't quite get which property to set to make it instead use the "character-code-reliant" approach.

ivan-mogilko:
> I don't know if the "locale" names are the same between C# and C++

The values set in the editor is the leftmost part the local name, in the form of "en_US". These are afaik pretty much standard. We are already using this exact property in AGS 4 when generating PO source files for translation: it is required to set up "Base Language" field.

I may search if there's any method to check valid and invalid language values in C#.
EDIT: found this: https://stackoverflow.com/questions/480425/asp-net-validating-culture-names

The full locale name is generated by the engine for its needs only.

The more interesting question is to ensure that the full locale name can be found by the runtime C++ library, as I noticed in the past that some use ".utf8", others ".UTF-8", and I dont know if these are interchangeable. If we find that some platforms do not recognize the current variant, we might have to make a switch using platform type as a condition.

> And I can't quite get which property to set to make it instead use the "character-code-reliant" approach.

This is done simply by resetting this property to empty string. That's also how backwards compatibility works.

The engine also falls back to "standard locale" if the requested locale is not found.

ericoporto:
> The more interesting question is to ensure that the full locale name can be found by the runtime C++ library, as I noticed in the past that some use ".utf8", others ".UTF-8", and I dont know if these are interchangeable. If we find that some platforms do not recognize the current variant, we might have to make a switch using platform type as a condition.

Uhm, I had a problem on this at work, but it was specifically with Python, which I don’t know what it uses behind the scenes, but it was that on Linux and Windows both worked but the macOS Python only accepted the upper case variant (“.UTF-8”). So “en_US.UTF-8” worked everywhere, but not “en_US.utf-8”.

Oh, and the other issue we had is once we fixed this, because the failure was silent, this fix that was meant to fix a text issue changed people’s decimal number output that was using comma (`,`) to dots  (`.`), which was unintentional.

ivan-mogilko:
So I did two changes compared to the previous variant:

1. In the Editor, the GameTextLanguage property is now a drop-down list, with all language names taken from the CultureInfo list. The names should be matching locale names, except they use hyphen `-` where in C/C++ (and elsewhere) they use underscore `_`, but that is easy to fix by replacing one character with another when writing game data, or loading data in the engine.
2. Engine will try several variants of the locale name form (with ".utf8", "UTF-8" etc) when initializing, trying to find one compatible with the current backend.

ericoporto:
Awesome, and I think this approach of trying different similar locales is a good way to circumvent the platform differences!

ivan-mogilko:
I made a big mistake here, but did not realize it until making tests for the sorting functions from another PR (#2959).

It appears that lexographical comparison always treats small and capital variants of the same letter as being close to each other in a sorted order (idk about others, but "en_US" locale puts small letters first). This is not entirely like a "case insensitive" comparison, because these are still treated as different letters, and wont be matching each other as dictionary keys, for example. But it's similar in a sense that strings which only differ in letter case will be positioned together in a sorted list.

This means that it may not be correct to replace existing comparison methods with lexographical compare, because that might change results. Not for old games, probably, since these dont have a "locale" setting, therefore will use a default C locale. But for the new games, where the results of sorting may appear unexpected.

I need to find a way to fix this, design-wise, and the only idea that I have so far is to introduce new StringCompare styles, like
* eCompareLexographical
* eCompareLexographicalCaseInsensitive (or eCompareLexographicalNoCase for short)
(existing compare styles are called eCaseX, but eCaseLexographical does not sound right).

EDIT: or find a better term than Lexographical, which would be easier to understand for users.

The reason to have both of them, including "case insensitive" variant, is that there still be a difference when testing for equality, thus affecting Dictionary keys, for instance.

ericoporto:
On the naming of things…

Searching for Lexographical the first result for me is a Wikipedia article that is very confusing: https://en.wikipedia.org/wiki/Lexicographic_order

I think what is meant here is perhaps something like Language Aware Alphabetical ordering.

Googling this then gives me directions to the word “collation” and also gave me another article on Wikipedia https://en.wikipedia.org/wiki/Unicode_collation_algorithm

I guess it would be like Locale Aware Comparison or Locale Aware Sort? (Trying to use other words to find a better meaning, Lexographical sounds like scarier word).

Unfortunately LocaleAware is huge

```
eCompareLocaleAware
eCompareLocaleAwareNoCase
```

but perhaps below isn't too bad?

```
eCaseInsensitive
eCaseSensitive
eCaseInsensitiveLocaleAware
eCaseSensitiveLocaleAware
```

or not sure how you are using in the other api but making a different thing entirely, with like `eCompareDefault` or ‘eCompareBinary` and a new 'eCompareLocaleAware`.

ivan-mogilko:
@AlanDrake have proposed to use term "Collate", like
> eCollatedCaseSensitive, eCollatedCaseInsensitive

I suspect that whatever term we use here, the users who dont know these differences will be equally confused and will require a documentation.

EDIT:
> 
```
eCaseInsensitive
eCaseSensitive
eCaseInsensitiveLocaleAware
eCaseSensitiveLocaleAware
```
Yes maybe.

EDIT2: making new names entirely may also have sense. We may keep old names marked as deprecated, but still available for some time.
BTW I've already had to deprecate name Character.AnimationVolume and replace with Character.AudioVolume in 3.6.3, because i've been adding more sound-related properties for controlling frame-linked sounds, and did not want that become confusing.

Issue #2964, opened 2026-04-16
Script Dictionary and Set types dont work in unicode + case-insensitive modes


ivan-mogilko:
**Problem**

The Dictionary and Set appear to use 1-byte string methods for key-matching in case insensitive modes.

The sorted variants use common stricmp to compare strings, that does not work with multi-byte characters in utf-8 strings.

The unsorted variants run hash algorithm over a lower-cased character sequence. But the hash algorithm is written for 1-byte character sequences and uses ASCII variant of "tolower". This will not work properly for multi-byte characters in utf-8 strings, leading to wrong key matching results.

**AGS Versions**
3.6.0 -> 3.6.3 (since Unicode support)

**Expected behavior**

Case-insensitive variants of Dictionary and Set must properly match uppercase variants of keys with lowercase variants (and vice versa) in the unicode mode.

ivan-mogilko:
Pushed a fix for release-3.6.2: ec617eaee743312a74615bf56e05a6b3e740f5a2
Will have to do a separate fix for 3.6.3 on master, also accounting to the problem mentioned in [this comment](https://github.com/adventuregamestudio/ags/pull/2958#issuecomment-4239764105).

ivan-mogilko:
This is fixed in both 3.6.2 and 3.6.3 branches now.

PR #2965, opened 2026-04-18
Re-implement lexographical compare in script under separate compare style

ivan-mogilko:
This is complementary to #2958 (and in effect to #2959).

I've made a mistake with #2958, which only realized later. Quoting my [comment](https://github.com/adventuregamestudio/ags/pull/2958#issuecomment-4239764105):

*It appears that lexographical comparison always treats small and capital variants of the same letter as being close to each other in a sorted order (idk about others, but "en_US" locale puts small letters first). This is not entirely like a "case insensitive" comparison, because these are still treated as different letters, and wont be matching each other as dictionary keys, for example. But it's similar in a sense that strings which only differ in letter case will be positioned together in a sorted list.*

*This means that it may not be correct to replace existing comparison methods with lexographical compare, because that might change results. Not for old games, probably, since these dont have a "locale" setting, therefore will use a default C locale. But for the new games, where the results of sorting may appear unexpected.*

This PR solves above by adding separate string compare style to script API:
```
eCaseInsensitiveLocaleAware,
eCaseSensitiveLocaleAware
```

When passed into respective functions, these will order a lexographical (locale-aware) string collation. The old styles will keep using default string comparisons where it's defined as a difference of character codes.

NOTE: the values of the string comparison style constants are suited for checking case sensitivity and locale-awareness as flags. This happened by accident, but is useful, so I added comments noting this fact.

The PR also contains fixes for #2964 (these were already pushed to release-3.6.2 branch).


ericoporto:
Does this makes any difference in [String.CompareTo](https://adventuregamestudio.github.io/ags-manual/String.html#stringcompareto) ? I kinda expect the case insensitive or case sensitive being indifferent for actual equality comparison vs being locale aware or not.

ivan-mogilko:
You are right that equality does not depend on this mode, only sorting position does. Meaning, that if you use CompareTo to sort strings (check if return value < 0 for example), the result may change depending on whether you use Lexographical compare style or not.

This is also the reason why StartsWith, EndsWith and Replace functions dont get any practical change with these new modes.

ivan-mogilko:
PR for locale-aware compare game tests: https://github.com/adventuregamestudio/ags-test-games/pull/26

PR #2966, opened 2026-04-21
Support translating Text Parser + unicode support

ivan-mogilko:
This does essentially three things that may be reviewed (and added) separately:
1. Refactored Text Parser code in the engine and made it support unicode strings.
2. Added support for translating Text Parser (this needs further explanation).
3. Allows to input multiple comma-separated words into Text Parser items in the Editor, this is another way to add synonyms, instead of having each synonym as a separate item in the list. That's just to speed things up when adding lots of things there.

### How does Text Parser translation works?

The main idea is that we cannot translate its dictionary word-to-word, because each language may require *different number of synonyms* to cover things nicely, and there may be no direct equivalents to all the base language words either. Therefore, instead of translating each word, we rather need to translate (replace) the whole list of synonyms.

For achieving this, my proposal is to add special entries into the translation source, annotated with a parser word id. The syntax is this:
```
//$PARSERWORD=N
comma-separated list of the base language synonyms
comma-separated list of the translation language synonyms
```
For example:
```
//$PARSERWORD=5
apple,fruit
pomme,fruit
```
**Nor the number, nor order of the synonyms does not have to match.**

Special words like "anyword" and "rol" probably should not be translated.

When compiled, these "parser words" are saved in another dictionary, separate from the main translation lines. They are loaded by the engine into a separate "translation" dictionary. This "translation" dictionary is used to parse the Parser.ParseText input (which is usually a player's input).

Other Parser functions, such as Said, and FindWordID, use base dictionary, and dont require translation by default. Since the parsed sentences are converted into the list of word IDs, then the result of parsing player's input (using translated dictionary) and result of parsing Said pattern may match together.

### Optional Parser.Said translation

The above scheme works so long as Parser.Said patterns suit the translation too. This is normally the case when the order of the words is the same both in base language and translation. However, certain language may require a different word order. In this case Parser.Said's pattern must be translated, and a translation dictionary used to parse it instead.

This feature provides an option for this. In TRS the new global option is defined as:
```
//#AutoTranslateParserSaid=ON or OFF
```

This is off by default, but when it's on, engine will require all Parser.Said lines to be translated in TRS as well. When translating these lines user must using exact same terms as in translated dictionary.

For example:
```
//$PARSERWORD=3
take
prendre
//$PARSERWORD=4
eat
manger
//$PARSERWORD=5
apple,fruit
pomme,fruit
```
```
// This is something typed in Parser.Said
take,eat apple,fruit
prendre,manger pomme,fruit
```


ivan-mogilko:
Ideas after discussing this on Discord:

* It's theoretically possible to not translate Said function's input, and instead parse the original input using base game's words dictionary. That way player's input is parsed using translated dictionary, and Said's pattern is parsed using base dictionary, then the parsed results are compared together (these word ids should match). The benefit of such approach would be that translators may skip translating Said input strings, which may save them alot of work.
The potential problem here is that some languages may require a different composition of the sentence (e.g. a different order of words). This case, perhaps, may be solved by adding another setting that tells whether Said input should be automatically translated or not. This setting should be per translation (because it depends on language and translator's choice).

PR #2970, opened 2026-04-30
Editor: automatically resolve UI questions in console mode

### LINKED ISSUE #2921: Editor: auto reimport assets from sources when doing automatic compilation
If any of the pre-compiled resources are not present in the project, such as sprites file and files in AudioCache, then in the manual mode user may issue a full sprites and audio reimport from sources. When doing a automatic game compilation with a "/compile" command arg, there's no way to control this, which may lead to a faulty game data. I think that the same could be done during automatic compilation, but automatically. We might add a new command arg for that.

This would let have automatically recompiled projects without sprites file and AudioCache folder, for the minimal repository size.

Also, by default, if such option is not passed, then the question about missing sprite file must be skipped, as if user replied "No".

ivan-mogilko:
Resolves #2921

This adds an extra 'defaultResult' parameter to GUIControler.ShowQuestion(), which is returned automatically if editor is working in "console" mode. This parameter is "Yes" by default, in order to force current operation to proceed, but may be provided on per case basis if necessary.

I've made a quick search throughout the project, and I think that in practice this may affect only one question so far:
- Whether user wants to recreate a missing sprite file from available sources.

PS. I might also backport this to 3.6.2 or earlier versions, since this is a small change that only affects automatic game building by the editor.

### DIFF
--- a/Editor/AGS.Editor/GUI/GUIController.cs
+++ b/Editor/AGS.Editor/GUI/GUIController.cs
@@ -195,13 +195,30 @@ namespace AGS.Editor
             ShowMessage(message, icon);
         }
 
-        public DialogResult ShowQuestion(string message)
+        public DialogResult ShowQuestion(string message, DialogResult defaultResult = DialogResult.Yes)
         {
-            return ShowQuestion(message, MessageBoxIcon.Question);
+            return ShowQuestion(message, MessageBoxIcon.Question, defaultResult);
         }
 
-        public DialogResult ShowQuestion(string message, MessageBoxIcon icon)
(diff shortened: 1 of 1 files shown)


ericoporto:
This looks alright. I haven’t tried the current behavior of this, does it show a question pop-up even on console mode right now? So it would hang-up the pipeline?

I wonder if it does and it relates to this: https://github.com/ericoporto/agstoolbox/issues/52 (but it is ags4)

ivan-mogilko:
> This looks alright. I haven’t tried the current behavior of this, does it show a question pop-up even on console mode right now? So it would hang-up the pipeline?

Yes, unfortunately it does, which means that a user must have a ready spritefile prepared for the automatic building.

And yes, the user who opened that issue reminded me of this problem.

PR #2973, opened 2026-05-03
Add section markers into translation source

### LINKED ISSUE #2413: Headers missing in translation source
**Describe the bug**
When creating a TRS file, in older versions of AGS, the engine would add section headers such as:
`// **** Room0.crm (Title screen) MESSAGES ARE BELOW ****`
`// **** GUI LABELS AND BUTTONS ARE BELOW ****`
The current version doesn't do that. I have no idea why this was changed, but it makes translating larger games inconvenient.

**AGS Version**
3.6.0

**Game**
n/a

**To Reproduce**
1. Create a translation

**Expected behavior**
Include the section headers

 - OS: Windows
 - Version 11

ivan-mogilko:
Resolves #2413

This adds section markers into TRS, thus dividing entries into groups by context, script, room, etc. The addition is purely meant to help TRS reading and editing, and has no impact on translation compilation, or on how it works at runtime.

The section markers are not ordinary comments, I had to make them special "tags" with "$SECTION" keyword. This is necessary for reading these sections back from TRS, because when Translation gets updated, it first parses TRS and reads its contents, and then merges only the new lines in.

Section's marker syntax is
```
//$SECTION = NAME; COMMENT
```
where NAME is a unique section identifier, and comment is any human-readable string. This separation is made to allow have descriptions, such as room names, which may be changed along with the development, without changing the section id.

Internally, the section names are gathered along with the text lines by IGameTextProcessor. For the global script texts the section name is the script filename. For rooms and their scripts it's "RoomN" and room's name is a comment. For dialogs - "DialogN" and dialog's script name is a comment.

### DIFF
--- a/Editor/AGS.Editor/Components/RoomsComponent.cs
+++ b/Editor/AGS.Editor/Components/RoomsComponent.cs
@@ -1752,7 +1752,10 @@ namespace AGS.Editor.Components
             {
                 Room room = LoadRoomAsTemporary(unloadedRoom, errors, doLoadScript: true);
 
-                room.Script.Text = processor.ProcessText(room.Script.Text, GameTextType.Script);
+                string roomContext = $"Room{room.Number}";
+                string roomContextComment = room.Description;
+
+                room.Script.Text = processor.ProcessText(room.Script.Text, roomContext, roomContextComment, GameTextType.Script);
                 if (processor.MakesChanges)
                 {
                     room.Script.SaveToDisk();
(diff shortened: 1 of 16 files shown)


ericoporto:
I see all GUIs are grouped in a single section, and this looks fine, but small question, if someday someone in the forums asked to have a section header per GUI that has actual lines to put there, would there be any problem - I mean if this version existed then and someone was loading an already filled trs file in the editor.

ivan-mogilko:
With the current code, if a section gets lost or renamed, the translated entries are kept where they are, in the old section.
User may manually move the lines to the correct section.

It might be possible to add a code that detects such problem beforehand (prior to merging old and new lines), and does something about this.

Also, it's worth mentioning, that in AGS 4.0 we have a PO translation source, where things must be done differently. We have to comply to PO syntax, which does not allow any "sections", but has "context" annotation per each individual entry. If we support context update, then all lines have to be compared with their new versions.

I think that all this requires to change the direction of the "merging". Right now it loads existing lines into Translation, and then adds only new lines that are not found there. The new algorithm should instead be: take the lines found in game, and copy existing translations from file into them, then put resulting dictionary into Translation.

ericoporto:
I haven’t used the translation feature much in the past but this looks like it’s doing what is saying it is doing. I see not all my rooms appears in generated translation from a test game, but looking into my rooms it looks like it is because the rooms that don’t appear just don’t have anything that could be translated in them.

PR #2975, opened 2026-05-20
agspak: add commands dealing with data appended to game exe

ivan-mogilko:
This adds a number of commands which let deal with game data appended to exe (common case with the older games).
* `-a, --attach` -  (syntax is `-a PAK_FILE DEST_FILE`) appends a input standalone pack to the existing file. Replaces the previously attached pack data, if there was any.
* `-d, --detach` -  (syntax is `-d SRC_FILE DEST_PAK`) copies appended asset pack from the source file to the standalone pack file, and removes the attachment in source file.
* `-x, --cut` - cuts an attached pack data from the file. If pack data occupies whole file, then its size will become zero.
* `-A, --create-attach` - (Yes, it uses capital 'A') - similar to "create", but appends pack data to the existing file, gathering the files from the input directory. Replaces the previously attached pack data, if there was any.


ericoporto:
Are any of these command one meant to be the opposite of the other? Like attach and detach, would it be append and split?

ivan-mogilko:
> Are any of these command one meant to be the opposite of the other? Like attach and detach, would it be append and split?

Hm, there's no direct opposite to split. I can add one, if it's necessary? Except this may be confused with the current "append".

Initially I thought about adding "append"/"attach" as a parameter to "create" command. I could remake append command as a create's parameter, and then add "attach" command for attaching existing package to exe.

ericoporto:
Sorry, I am dumb, I understand attaching to the exe and detaching from the exe, but what is the third command meant to do?

I am not a super fan of the append word because it doesn’t have an obvious antonym (“reverser”) to it.

I would expect a command that takes an exe file and create an exe file and an AGS file and another that takes an exe and an AGS file and create an exe file - going from the description. But I haven’t had the time to look at the code - GitHub is not loading the diff for me where the commands are actually added and I didn’t pull your branch from this PR yet because I am not on my computer.

ivan-mogilko:
Because attachment to exe is a narrow case, which is not recommended for use today, perhaps a better variant would be to designate only long names for the related commands, and reserve one-letter commands for something more frequently used.

>  I understand attaching to the exe and detaching from the exe, but what is the third command meant to do?

- cut = delete attachment
- split = detach attachment into the separate file.

So I am now thinking the list of commands should be different:
- attach operation that takes game.ags and appends to game.exe, removing game.ags (optionally?)
- detach operation that splits game.exe into pure game exe and game.ags
- delete attachment operation that just deletes attachment from exe.
- optionally, a combined command that *creates* and then *attaches* to exe.

ivan-mogilko:
Another thing, i begin to see now that i've made a mistake when rewriting agspak last time, when i made commands be designated using `-x` syntax. It may be more convenient to have them designated without dash, similar to how e.g. git has its commands, and also how you did in [agfexport](https://github.com/adventuregamestudio/ags/blob/master/Tools/agfexport/main.cpp).

Then `-x` syntax will be used purely for command options, which will give more freedom in designating option letters. As one command's options do not have to be strictly matching another command's options.

ericoporto:
Oh, yeah, the `-x` approach was more modeling towards `tar` but it isn’t a good option I think when you have many many options (erh.. like tar), the other approach even allows using `-h` in the subcommands which is good to get more help when running things. The things I use more in day to day usage on command line is definitely git, so I used similar ergonomics in agstoolbox atbx command line tool (that even has sub sub commands, but perhaps that was a bit too much)

ericoporto:
@ivan-mogilko , Hey, I took a stab at rewriting as command based, can you take a look and see what you think?

https://github.com/ericoporto/ags/commit/1c76bc55f4844148ee0121c80c382554e0304c23

I copied from agfexport most of the code. Frankly I had forgotten I had written it until you mentioned! I remember now it was to join some previously separate tools.

I am unsure a bit if the initial help message should have all the optional flags or if they should only be in each specific command help.

ivan-mogilko:
I will return to this later, right now i have couple other priority tasks.

ivan-mogilko:
So, I have remade the list of commands:
* `-a, --attach` -  (syntax is `-a PAK_FILE DEST_FILE`) appends a input standalone pack to the existing file. Replaces the previously attached pack data, if there was any.
* `-d, --detach` -  (syntax is `-d SRC_FILE DEST_PAK`) copies appended asset pack from the source file to the standalone pack file, and removes the attachment in source file.
* `-x, --cut` - cuts an attached pack data from the file. If pack data occupies whole file, then its size will become zero.
* `-A, --create-attach` - (Yes, it uses capital 'A') - similar to "create", but appends pack data to the existing file, gathering the files from the input directory. Replaces the previously attached pack data, if there was any.

I will leave the idea of bringing commands to a format without dashes for later time. I think if that's done, then some other tools should have this too, spritepak for instance.

PR #2976, opened 2026-05-24
Tool: CMake uses allegro library

ericoporto:
The allegro library is not too big and it will go away when optimizing.

Without this fix I get a ton of undefined symbol erros in macOS  building spritepak, like:

```
: && /usr/bin/c++ -g -arch arm64 -mmacosx-version-min=10.9 -Wl,-search_paths_first -Wl,-headerpad_max_install_names Tools/CMakeFiles/spritepak.dir/spritepak/main.cpp.o Tools/CMakeFiles/spritepak.dir/spritepak/commands.cpp.o -o Tools/spritepak Tools/liblibtools.a libsrc/miniz/libminiz.a libsrc/tinyxml2/libtinyxml2.a && :
Undefined symbols for architecture arm64:
"_geta32", referenced from:
_geta_depth in liblibtools.a[38](color.c.o)
"_getb15", referenced from:
_getb_depth in liblibtools.a[38](color.c.o)
"_getb16", referenced from:
_getb_depth in liblibtools.a[38](color.c.o)
"_getb24", referenced from:
_getb_depth in liblibtools.a[38](color.c.o)
"_getb32", referenced from:
_getb_depth in liblibtools.a[38](color.c.o)
"_getb8", referenced from:
_getb_depth in liblibtools.a[38](color.c.o)
"_getg15", referenced from:
_getg_depth in liblibtools.a[38](color.c.o)
"_getg16", referenced from:
_getg_depth in liblibtools.a[38](color.c.o)
"_getg24", referenced from:
_getg_depth in liblibtools.a[38](color.c.o)
"_getg32", referenced from:
_getg_depth in liblibtools.a[38](color.c.o)
"_getg8", referenced from:
_getg_depth in liblibtools.a[38](color.c.o)
"_getr15", referenced from:
_getr_depth in liblibtools.a[38](color.c.o)
"_getr16", referenced from:
_getr_depth in liblibtools.a[38](color.c.o)
"_getr24", referenced from:
_getr_depth in liblibtools.a[38](color.c.o)
"_getr32", referenced from:
_getr_depth in liblibtools.a[38](color.c.o)
"_getr8", referenced from:
_getr_depth in liblibtools.a[38](color.c.o)
"_makeacol32", referenced from:
_makeacol_depth in liblibtools.a[38](color.c.o)
"_makecol15", referenced from:
_makecol_depth in liblibtools.a[38](color.c.o)
_makeacol_depth in liblibtools.a[38](color.c.o)
"_makecol16", referenced from:
_makecol_depth in liblibtools.a[38](color.c.o)
_makeacol_depth in liblibtools.a[38](color.c.o)
"_makecol24", referenced from:
_makecol_depth in liblibtools.a[38](color.c.o)
_makeacol_depth in liblibtools.a[38](color.c.o)
"_makecol32", referenced from:
_makecol_depth in liblibtools.a[38](color.c.o)
ld: symbol(s) not found for architecture arm64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I tried many other alternatives, but most of the time I get something working in my macOS computer it apparently breaks some other platform. This way is safe though, and actually works.

### DIFF
--- a/Tools/CMakeLists.txt
+++ b/Tools/CMakeLists.txt
@@ -91,7 +91,8 @@ endif()
 
 target_include_directories(libtools PUBLIC ../libsrc/allegro/include)
 set(TOOLS_MINIMAL_ALLEGRO_SOURCES
-        ../libsrc/allegro/src/color.c)
+        ../libsrc/allegro/src/color.c
+        data/minallegro_inline.c)
 target_compile_definitions(libtools PRIVATE ALLEGRO_STATICLINK)
 
 target_sources(libtools
--- a/Tools/data/minallegro_inline.c
+++ b/Tools/data/minallegro_inline.c
@@ -1,0 +1,33 @@
+//=============================================================================
+//
+// Adventure Game Studio (AGS)
+//
+// Copyright (C) 1999-2011 Chris Jones and 2011-2026 various contributors
+// The full list of copyright holders can be found in the Copyright.txt
+// file, which is part of this source code distribution.
+//
+// The AGS source code is provided under the Artistic License 2.0.
+// A copy of this license can be found in the file License.txt and at
+// https://opensource.org/license/artistic-2-0/
+//
+//=============================================================================
+
+
+// similar to own allegro's inline.c
+// but minimal so it doesn't pull all allegro's inline functions
+// the idea here is to do minimal inline when building without optimizations (e.g. Debug)
+
+#define AL_INLINE(type, name, args, code)    type name args code
+
+#include "allegro/color.h"
+
+// unless we define ALLEGRO_DOS, color.inl needs a symbol for set_color from gfx.c
+// we don't use set_color at all in spritepak which is the reason we are writing this.
+// we fix the linker error with symbol not found for _set_color in libtools by having a dummy implementation
+
+void set_color(const int idx, AL_CONST RGB * p)
+{
+    // this should neve be called
+    (void) idx;
+    (void) p;
+}


ivan-mogilko:
I put an effort into keeping dependencies low in the tools, because all they require from Allegro at the moment is reading and writing pixel buffers (geta/getb/getr, rgb shifts). Even moved Allegro code around a little, to avoid having seemingly unrelated stuff in color.c.

Unfortunately, I did not pay much attention to this macOs problem.
What about the tool Makefiles, they use the "minimal allegro" too. Do they work or also no?

Maybe I could check this? I have had a mac os installed on a vm. Although never had any compilers there.

Alternatively, I could see if it's viable to reimplement color functions outside of Allegro.

ericoporto:
It’s not that you didn’t pay attention, it’s just that it does work in the CI - it literally builds alright here with CMake in all platforms including macOS here through GHA. The thing is, I don’t understand why it works.

But it doesn’t work on my local PC. It appears that AL_INLINE just resolves weird on my pc to something unexpected. And also that Clang defines the GCC macros for things it supports which makes sense, but I think this wasn’t the case when these were written in the header that defines AL_INLINE.

I tried throwing an llm at it, but it’s solution was adding a new c file that was like inline.c but it had only specific color.h and some other header instead of all and add it to the minimal allegro you made, but I thought it was cleaner to just add allegro with how small it is.

my other alternative idea was to rewrite how AL_INLINE is defined, but after reading a bit on Clang, Gcc and MSVC I didn’t feel like I totally understood what inline is actually meant to do in this use case.

ivan-mogilko:
> It’s not that you didn’t pay attention, it’s just that it does work in the CI - it literally builds alright here with CMake in all platforms including macOS here through GHA. The thing is, I don’t understand why it works.

Right, i mean, i did not follow your comments on this problem, being too busy with other stuff, and then completely forgot about this. I wish this was documented as an issue.

I'd either revise how AL_INLINE is implemented, or how color.c is included.

> But it doesn’t work on my local PC. It appears that AL_INLINE just resolves weird on my pc to something unexpected

Do you know what it resolves to precisely? That is, which of the conditional preprocessor branches does your compiler take there?

ericoporto:
It looks like for me I get here

https://github.com/adventuregamestudio/ags/blob/ac9e2e362c1bef53b4b45a17e2375ff627398518/libsrc/allegro/include/allegro/internal/alconfig.h#L100-L102

My compiler is Apple clang version 17.0.0 (clang-1700.0.13.5). The Clang docs (but not necessarily this Apple version) [has some info on this](https://clang.llvm.org/docs/AttributeReference.html#gnu-inline).

I know that if I force it to be what is on line 92 then it magically works (but this is probably wrong elsewhere).

```diff
diff --git a/libsrc/allegro/include/allegro/internal/alconfig.h b/libsrc/allegro/include/allegro/internal/alconfig.h
index a3c9a2529..1f9afe511 100644
--- a/libsrc/allegro/include/allegro/internal/alconfig.h
+++ b/libsrc/allegro/include/allegro/internal/alconfig.h
@@ -88,7 +88,7 @@
    #define ALLEGRO_GCC
 
    #ifndef AL_INLINE
-      #ifdef __cplusplus
+      #ifdef __GNUC_STDC_INLINE__
          #define AL_INLINE(type, name, args, code)    \
             static inline type name args;             \
             static inline type name args code
```

ericoporto:
I got it now, the CI CMake test only builds for Release, and locally I only test things for Debug. The inline.c file has the following sentence.

https://github.com/adventuregamestudio/ags/blob/ac9e2e362c1bef53b4b45a17e2375ff627398518/libsrc/allegro/src/inline.c#L11-L12

So, I don't understand how this file works, and while there is an `#UNDEF AL_INLINE` in the allegro msvc header, there isn't one in the unix variants.

Anyway, if I build for Release locally, it works, the error only happens when I build for debug - which I guess disable the optimizations, which are needed for that inline to work properly.

ivan-mogilko:
Thinking logically, 

1. The linker errors point to color.c, meaning that when color.c is compiled, it somehow cannot properly get these functions are inlined, and instead gets them purely as external function declarations.

2. There are several variants of AL_INLINE definitions in alconfig.h. First there's a big switch, that you reference above, but the interesting thing is that every single one of these variants end up with having a `code` part in them. E.g. for `__GNUC_STDC_INLINE__` part:
```
 #define AL_INLINE(type, name, args, code)    \
    extern __inline__ __attribute__((__gnu_inline__)) type name args;         \
    extern __inline__ __attribute__((__gnu_inline__)) type name args code
```
Here's `code` included. Which means, in theory, that function is declared with its body.

3. But, there's one exception for this, it's another AL_INLINE declaration, at line 251 below:
```
#ifndef AL_INLINE
   #define AL_INLINE(type, name, args, code)       type name args;
#endif
```
If I'm reading this right, then this definition must be enabled in case neither of the above worked. As you may see, it does not have the `code` part. Meaning there's no function body, and such declaration will require inline.c (or equivalent) linked to the program (which spritepak tool does not do).

Could you double check that your macos Debug build is getting (or not getting) under this final `#ifndef AL_INLINE` when compiling color.c?

ivan-mogilko:
In the worst case, we might add an equivalent for inline.c purposed for the minimal allegro sources (including only necessary headers). This will cover the case when the function cannot be inlined for some reason.

ericoporto:
> Could you double check that your macos Debug build is getting (or not getting) under this final #ifndef AL_INLINE when compiling color.c?

I am not on my pc right now, but I tried to add a `#error` macro there to see if it would trigger in the CI, but I think it didn’t work and I am still seeing the linker errors on the CI. I am on my phone though, will be able to think better at home.

> In the worst case, we might add an equivalent for inline.c purposed for the minimal allegro sources (including only necessary headers).

I wasn’t sure of where to put it, should it be in tools data directory?

ivan-mogilko:
> I wasn’t sure of where to put it, should it be in tools data directory?

Yes, i think so, because Allegro itself does not know which of its parts are going to be used.

ivan-mogilko:
If you think that you are spending too much time on this, then i suppose it's better to go with your original straightforward solution of linking allegro library. I'd only ask to keep Makefile as it is for now, as a alternative with minimal dependency.
I noticed that CMake already links all Common files used in all tools as one lib, regardless of whether particular tool needs them all or not.

About Allegro dependency, it should be resolved somehow in the future, just dont have exact plan yet, and too busy to attend this task at the moment.

ericoporto:
Managed to make a minimal file for inline and made it work. I have no idea what set_color does and I was having build errors with it, but I noticed it is conveniently in a ifdef condition below

https://github.com/adventuregamestudio/ags/blob/886e8fa001fa3ab0833f86406ba0be7e78bd71f9/libsrc/allegro/include/allegro/inline/color.inl#L163-L170

So I just defined so it just go away. Without this I still got a single linker error as below

```
[6/9] Linking CXX executable Tools/spritepak
FAILED: Tools/spritepak
: && /usr/bin/c++ -g -arch arm64 -mmacosx-version-min=10.9 -Wl,-search_paths_first -Wl,-headerpad_max_install_names Tools/CMakeFiles/spritepak.dir/spritepak/main.cpp.o Tools/CMakeFiles/spritepak.dir/spritepak/commands.cpp.o -o Tools/spritepak Tools/liblibtools.a libsrc/miniz/libminiz.a libsrc/tinyxml2/libtinyxml2.a && :
Undefined symbols for architecture arm64:
"_set_color", referenced from:
__set_color in liblibtools.a[38](minallegro_inline.c.o)
ld: symbol(s) not found for architecture arm64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

It looks like this would break `file.c` because then it would use the wrong conventions in an unix system due to ALLEGRO_DOS definition, but since we don't use `file.c` there in the Tools, this should be fine.

> I'd only ask to keep Makefile as it is for now, as a alternative with minimal dependency.

They have O2 hardcoded, so this is not a problem there.

ivan-mogilko:
Sorry for delay. I am checking this out now.

There's one error in "Test Windows" action:
> D:\a\ags\ags\Tools\data\minallegro_inline.c(27,6): error C2055: expected formal parameter list, not a type list

That refers to `void set_color(int, AL_CONST RGB *) {}`

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2055?view=msvc-170

Another failing action is mingw-x86_64 which fails at locale-aware game tests, but that's not relevant here.

ericoporto:
I added the names of the parameters but not an implementation as I think we don't use it at all.

ivan-mogilko:
> I added the names of the parameters but not an implementation as I think we don't use it at all.

I forgot to mention, in C the "unused parameter" warning may be prevented by adding following lines on top of the function:
```
(void)param;
```

ericoporto:
This will take two more lines of code, just to check, I can also implement the function in two lines of code.

I think it is an if the index is between 0 and the palette size, then it sets the color to that palette index.

https://github.com/adventuregamestudio/ags/blob/4af4283fc87b7e8b16077896be4b68bc5f340134/libsrc/allegro/src/gfx.c#L189-L225

The advantage of going for silencing the warnings is if we add something to print/throw that it should not reach that and we reach we know to add gfx.c too in the future.

edit:
Made the void stuff for now

https://github.com/adventuregamestudio/ags/compare/c3d0248a103aa65b37b675d382ccf8399fb7dfe3..021de8b47af8d092947e075dcabe9d1200fc55fd

PR #2986, opened 2026-06-10
AGS 4.0: add Pivot and PivotOffset for game objects and Camera

ivan-mogilko:
This adds following properties:
```
float Character.GraphicPivotX
float Character.GraphicPivotY
int Character.GraphicPivotOffsetX
int Character.GraphicPivotOffsetY
```
```
float Object.GraphicPivotX
float Object.GraphicPivotY
int Object.GraphicPivotOffsetX
int Object.GraphicPivotOffsetY
```
```
float Overlay.GraphicPivotX
float Overlay.GraphicPivotY
int Overlay.GraphicPivotOffsetX
int Overlay.GraphicPivotOffsetY
```
```
float Camera.PivotX
float Camera.PivotY
int Camera.PivotOffsetX
int Camera.PivotOffsetY
```

PivotX,Y are pivot positions in normalized relative coordinates (0.0 - 1.0). This is useful when you need to align a pivot to the sprite's edge, or fraction of its size.
PivotOffsetX,Y are relative offsets in pixels. This is useful when you need to align a pivot to exact pixel within the sprite.
These two may be used simultaneously (they are combined internally).

Furthermore added 2 functions to all the graphic objects: Character, GUI, GUIControl, Object and Overlay:
* GetGraphicPosition() - returns array of Points with coordinates of the object graphic's corners
* GetGraphicBoundBox() - returns array of Points with coordinates of the object graphic's AABB (axis-aligned bounding box).

### DIFF
--- a/Common/ac/characterinfo.cpp
+++ b/Common/ac/characterinfo.cpp
@@ -165,7 +165,7 @@ void CharacterInfo::WriteToFile(Stream *out) const
     out->WriteInt8(0); // alignment padding to int32
 }
 
-void CharacterInfo::ReadFromSavegame(Stream *in, CharacterSvgVersion save_ver)
+void CharacterInfo::ReadFromSavegame(Stream *in, LegacyFields &old_fields, CharacterSvgVersion save_ver)
 {
     defview = in->ReadInt32();
     talkview = in->ReadInt32();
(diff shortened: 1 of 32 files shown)


ericoporto:
Does it need to also be supported in the software renderer?

ivan-mogilko:
Eventually, but I intentionally skip software renderer for now, as it takes additional effort to add transformations, and is probably already falling behind in some features.

ivan-mogilko:
I got to say, unintentionally, but this allows to *visually* rotate an object around an external point, which is located virtually anywhere, even far away offscreen. Which may be useful, but seems very weird design-wise. Because object's "position" (x,y properties) can stay the same, while the object's graphic is circling all around the screen.

I am now wondering if we should restrict pivot and pivot offset to the object sprite's bounds.

ivan-mogilko:
It appeared to me, that we need functions that would return sprite's position on the screen. This becomes ever more important with rotation pivot, as that may shift sprite's visual position very far away from the object's formal position, defined by its x,y properties.

I am thinking about two functions:
- GetGraphicPosition()
- GetGraphicAABB(), another variants: GetGraphicBox, GetGraphicBoundBox
both returning a dynamic array of 4 Points. First returns points of 4 actual sprite corners (may be shifted by rotation), and second returns a Axis-Aligned bounding box of a sprite.

For room objects these are positions in room coords, for screen objects they are in screen coords.

PR #2988, opened 2026-06-14
Support loading/saving images as PNGs

ivan-mogilko:
This adds a minimal support for loading and saving PNGs image:
- by the engine (for DynamicSprite.CreateFromFile / SaveToFile, and similar commands)
- by the spritepak tool.

The implementation idea is actually based on SDL3 code, where PNGs are loaded using stb_image, and written using available simplistic function in miniz (since we already have miniz).

This PR adds a stripped version of stb_image.h, which only contains code necessary for loading a PNG file. Potentially, we could re-add more formats support there, if it will be necessary. stb_image itself is coded with numerous compiler switches, that allow to enable only particular formats, so engine might use some of them, and tools use all of them, for example.

The PR is currently a draft, as there's number of issues remaining.
1. [DONE] PNGs have a different order of RGB compared to AGS sprites, so we need to either adjust the pixel data on save/load, or adjust the stb_image code and let its function to receive rgb shifts.
2. [DONE] Did not test 16-bit sprites yet, dont know if they will work properly.
3. [DONE] stb_image does not support loading image and return palette in one go. In SDL3 they have a extended function that does that. We may copy it over, but I'd like to find out where the code comes from first (is it written by SDL3 team or another version of stb_image).
4. [DONE] miniz's PNG writing function does not support palette as parameter. In SDL3 they have an extended implementation which does. Similarly, we may copy it over.
5. [CHANGED MY MIND] ~Our PixelBuffer object does not support attaching pre-allocated array. That may be easy to do (support custom deleter), then the engine may keep stb's allocated buffer without making extra copies.~
EDIT: i think it's better to mod the library and make it allow accepting our allocated buffer instead.
6. [NOT SURE] stb_image.h has its own deflate implementation necessary for decompressing PNGs. I've been wondering if we could remove it and wire miniz instead.


ericoporto:
I forgot how to use these stb libraries, since they are only a header, I think they require at least one cpp file that has a macro definition in it to say that’s where the implementation should be. I am not sure how that works if you keep it without this empty cpp file in the stb::stb CMake library.

my suggestion is to add a file like below

**`libsrc/stb/stb.cpp`**
```C++
#define STB_IMAGE_IMPLEMENTATION
#include “stb_image.h
```

ivan-mogilko:
I already include stb_image.h with implementation in gfx/image_png.cpp

The errors I was getting were not linker errors, they were CMake errors, because "stb" library is not found. I dont know CMake well, so just tried to mimic how miniz is included. Maybe this is not what I was supposed to be doing.

EDIT: the way CMake deals with libraries is confusing for me, so I just followed your suggestion for now...

ericoporto:
> This PR adds a stripped version of stb_image.h

Perhaps it would be a good idea to add a readme.md in it’s directory that mentions this, and points from which upstream and commit hash it was created from, so we can update in the future if necessary.

ivan-mogilko:
Added "version.txt", similar to the one in miniz directory.

ericoporto:
<details><summary>Edit: apparently, closing AGS Editor and opening it again magically fixed both issues.</summary>

Hey, two things that aren't specific to this PR...

- does the event `room_AfterFadeIn` runs in the room that the game started? (it looks like it is not running for me...)
- when I click the `Enters room after fade-in` entry in the room properties, it fills the box with the text `room_AfterFadeIn`, and jumps to the room1.asc file, but it doesn't auto write there this event. I seem to recall this used to auto write the event there...

</details>

On this PR, I noticed that it is missing adjusting the agsdefns.sh

https://github.com/adventuregamestudio/ags/blob/0b8f996ae8eaa43f7ddffa528c154684b26f9016/Editor/AGS.Editor/Resources/agsdefns.sh#L1592-L1593

It should probably list PNG as a possible file there too.

I tried reading PNG files and it looks like it is working!

In AGS4 the dynamic sprites can be 8-bit too, I am not sure if something has to change when merging there in case it opens a 8-bit png in a 32-bit game.

PR #2989, opened 2026-06-16
AGS4 Engine: remove CursorRole Look condition when evaluating OPT_WALKONLOOK

clarkjoe:
The `OPT_WALKONLOOK` config property is tied to the usage of Cursor Roles. This means if I want to use the `WalkTo` for a hotspot for an Interact but NOT a Look, then I have to use Cursor Roles.

This PR decouples that condition and makes `OPT_WALKONLOOK` more absolute, and more reflective of the Editor description:

> Automatically walk to hotspots in Look mode
Whenever the player clicks somewhere in Look mode, the player character will automatically be moved there

### DIFF
--- a/Engine/ac/hotspot.cpp
+++ b/Engine/ac/hotspot.cpp
@@ -161,7 +161,7 @@ void RunHotspotInteraction(int hotspothere, int mood)
         play.usedinv = playerchar->activeinv;
     }
 
-    if ((game.options[OPT_WALKONLOOK] == 0) && (game.HasCursorRole(mood, kCursorRole_Look)));
+    if (game.options[OPT_WALKONLOOK] == 0);
     else if (play.auto_use_walkto_points == 0);
     else if ((!game.HasCursorRole(mood, kCursorRole_Walk)) && (play.check_interaction_only == 0))
         MoveCharacterToHotspot(game.playercharacter, hotspothere);


ivan-mogilko:
I need to clarify the situation: this "Walk on Look" that triggers automatic walk for Look mode, is a remnant of a very old engine logic when lots of stuff was simply hardcoded, and I'd rather plan on removing this setting and this mechanic from ags4 altogether, leaving users to script their own. Having this done in script is much more flexible than trying to configure such behavior inside the engine.

clarkjoe:
Okay. Is your vision just to remove the setting altogether and all references to it? I can do that in another PR. 

I assume it should be removed from the editor as well.

ivan-mogilko:
In the Editor the property itself should be kept but marked with [Obsolete] attribute, that's a way to avoid errors when loading the older projects we have at the moment. Any underlying variable should be removed, and also the property removed from any use, such as writing into the compiled game data.

In the engine code, the OPT_x constant is to be renamed to OPT_OBSOLETE_x, this is just to have a information that this option index was used by a deprecated option.

(There should be existing examples of such options, they may be used as a reference.)

I would really remove automatic walking to Hotspots completely (there's a play.auto_use_walkto_points variable that controls this in the engine), because AFAIK it's inconsistently works only for Hotspots and nothing else, so if a user wanted to have it with characters and room objects, they'd have to script this anyway.
(Also this probably won't work with any non-standard character walking.)

EDIT: checking this now, there's also a obscure hardcoded behavior in RoomProcessClick, that if player uses Walk cursor on a hotspot with assigned "WalkTo" point, then this point will be used instead. I frankly think this should be removed as well.

PR #2991, opened 2026-06-19
Advanced game text direction, based on U control chars

ivan-mogilko:
This implements a rather basic text direction handling based on Unicode control characters such as LRE (U+202A) and RLE (U+202B). Whenever one of these chars are met, the next section of the text will be applied a respective direction. This allows to have sections with different directions even in one string.

In this mode the OPT_RIGHTLEFTWRITE option serves only as the starting direction, but if any control char is met in the string, they will override the direction.

This rule is applied only when the game is in Unicode mode, naturally, as ASCII/ANSI does not have such control characters.

There are following common use cases:
* Combine languages which have different directions in the same game, simultaneously.
* When using RTL game texts, or translation, still allow some parts of the text to be LTR, because not everything is reversed in RTL languages. Primary example is numbers.

This feature is likely superseded by the use of text rendering libraries, such as HarfBuzz, in the future AGS 4.x.

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -449,6 +449,82 @@ size_t StrUtil::ConvertWstrToUtf8(const wchar_t *wcstr, char *out_mbstr, size_t 
     return len;
 }
 
+String StrUtil::ApplyTextDirection(const String &text, bool default_rtl)
+{
+    const Utf8::Rune LRE = 0x202A;
+    const Utf8::Rune RLE = 0x202B;
+
+    // We may only cut control characters, but not add any new ones,
+    // therefore we allocate a buffer which is at least as large as the input.
+    // NOTE: we don't allocate a buffer unless meet control chars (or RTL section);
+    // in case whole string is LTR, we just return original string.
+    char *dst_buf = nullptr;
+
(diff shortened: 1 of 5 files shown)


ericoporto:
Just a minor nitpick in the description above, HarfBuzz is about shaping, and this handling that is about bidirectional text isn’t supported by HarfBuzz, their recommended approach is to instead rely on a BiDi library. Even with SDL_ttf this isn’t handled because it doesn’t integrate a bidi library 

https://github.com/libsdl-org/SDL_ttf/issues/135

Edit: actually I think they added something similar to what you did, check issue below

https://github.com/libsdl-org/SDL_ttf/issues/472

I am also not a 100% sure, but I think HarfBuzz can change the width of the word once the ligatures are applied - but it has an api like the one from FreeType to retrieve the width if I remember correctly and this is reflected in sdl_ttf too, from memory.

ivan-mogilko:
The current pr is a minimal and pretty naive implementation, not aiming at supporting all control chars. Just to let translators be able to mix LTR and RTL text parts. Because right now they have trouble with things like non-translated english names or numbers in format strings (`%d`) which must *not* be reversed.

PR #2995, opened 2026-06-26
4.0: Fix Joystick button events breaking input processing

### LINKED ISSUE #2979: AGS 4 Engine Bug: `Mouse.Click` stops working after pressing Joystick button.
**Describe the bug**
When testing Joystick or even AGS-Controller support in AGS 4, it looks like `Mouse.Click()` stops working after pressing a button in the gamepad. First I was using AGS-Controller plugin, but when I migrated to Joystick it also was causing the same problem, so I suspect it's an SDL-related issue. I created a test project below to show the issue.

**AGS Version**
4.0.0.28 (May 2026)

**Game**

[AGS4TestGame.zip](https://github.com/user-attachments/files/28565164/AGS4TestGame.zip)

Look at repeatedly_execute_always and on_key_press in the GlobalScript file for example code.

**To Reproduce**

1. Connect a joystick/gamepad to your computer
2. Run the game
3. Move the cursor with your mouse/trackpad and press Return/Enter (this is mapped to mouse.Click(left))
4. Press a button in your joystick/pad (A, Cross, whatever)
5. Try pressing enter again. Nothing happens!

**Expected behavior**
mouse.Click() should trigger events in `on_mouse_click`, and player walks near where the cursor is located.

**Screenshots**
n/a

**Additional context**
I was looking at this commit to see if it caused the problem: https://github.com/adventuregamestudio/ags/commit/9fb068fa6ed6a2bd6274d23e663eb2ddaf41f817

ivan-mogilko:
Fix #2979

1. Fix Joystick button events not registering as a proper input type (kInputGamepad). Skip them if the joystick is "gamepad controller", because in that case we also receive gamepad button events. This will avoid duplicates.
2. Remake input event processing loop, don't rely on the loop ending when there's kInputNone met, because in theory it might appear in the queue by mistake, leading to never flushing queue.

### DIFF
--- a/Engine/ac/dialog.cpp
+++ b/Engine/ac/dialog.cpp
@@ -1372,7 +1372,7 @@ bool DialogOptions::Run()
 bool DialogOptions::RunControls()
 {
     bool state_handled = false;
-    for (InputType type = ags_inputevent_ready(); type != kInputNone; type = ags_inputevent_ready())
+    for (InputType type = kInputNone; ags_hasinputevent_ready(type);)
     {
         switch (type)
         {
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -349,7 +349,7 @@ std::unique_ptr<Bitmap> create_textual_image(const char *text, const DisplayText
 bool display_check_user_input(int skip)
 {
     bool state_handled = false;
-    for (InputType type = ags_inputevent_ready(); type != kInputNone; type = ags_inputevent_ready())
+    for (InputType type = kInputNone; ags_hasinputevent_ready(type);)
     { // NOTE: must handle them all in case there were engine's hotkeys too
         switch (type)
         {
--- a/Engine/ac/sys_events.cpp
+++ b/Engine/ac/sys_events.cpp
@@ -309,6 +309,14 @@ int sys_keydown_count = 0; // counter of keys pressed down
 int sys_modkeys = 0; // saved accumulated key mods
 bool sys_modkeys_fired = false; // saved mod key combination already fired
 
+bool ags_hasinputevent_ready(InputType &type)
+{
+    if (g_inputEvtQueue.empty())
+        return false;
+    type = ags_inputevent_ready();
+    return true;
+}
+
 InputType ags_inputevent_ready()
 {
     if (g_inputEvtQueue.empty())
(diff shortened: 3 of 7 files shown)


edmundito:
Tested with a Nintendo Switch Joy-Con. A, B, Return key, and mouse clicks work as expected. 👍🏼

ivan-mogilko:
No, this is still a incomplete fix.

We do receive joystick button events in both cases: whether this is joystick and if this is a game controller with joystick interface. But engine only handles game controller button events further, in `run_service_gamepad_controls`. So joystick own events will get ignored, even if it's a "pure" joystick and not a "game controller".

We need to determine:
* whether we handle joystick button events at all, and if yes then how do we do this in `run_service_gamepad_controls`.
* can we handle *only* joystick button events, as they are received both for joystick and game controller? in this case we could just ignore game controller button events. If not, then we need to handle two kinds of button down in `run_service_gamepad_controls`.

@ericoporto do you remember why SDL_JOYBUTTONDOWN was not handled in your code?

EDIT: BTW I tried to filter SDL_JOYBUTTONDOWN event instead using SDL events filter function, thinking that it will be more "proper", but it turned out that if we filter out SDL_JOYBUTTONDOWN on SDL2 level, then SDL_CONTROLLERBUTTONDOWN is not received at all.

ericoporto:
> [ericoporto](https://github.com/ericoporto) do you remember why SDL_JOYBUTTONDOWN was not handled in your code

I am not sure, but I think it was because I had not figured out an API to set the joystick buttons for things (skip display, skip speech, skip video), so the events only existed for gamepad, meanwhile the joystick interaction was more through querying the joystick API itself.

Dualnames1:
image

NES controller for the win. Tested with a joystick.

ivan-mogilko:
Alright, I see, the controller has button codes designating their roles, while joystick buttons have only index.

Guess I'll leave this at how it is now then, since i'm only fixing the input queue breaking here. This joystick button issue should be resolved before final 4.0 release.

ivan-mogilko:
@Dualnames1 , Are you sure the above code works properly? You are accessing SDL_Event's `cbutton` member under JOYSTICKBUTTON event, but afaik the joystick events have `jbutton` member set instead,

In any case, that's a separate matter.

PR #2996, opened 2026-06-26
Joystick Skipping dialog fix

Dualnames1:
Engine: 4.0.0.29
-Fixing the issue where a joystick button couldn't skip dialog.
-Changed a few naming conventions and mostly made several functions use GamepadInput instead of an int, to be able to adjust things better down the line

### DIFF
--- a/Engine/ac/display.cpp
+++ b/Engine/ac/display.cpp
@@ -395,15 +395,15 @@ bool display_check_user_input(int skip)
             GamepadInput gbut;
             if (!run_service_gamepad_controls(gbut) || play.fast_forward || state_handled)
                 continue; // handled by engine layer, or fast-forwarded, or resolved
-            eAGSGamepad_Button gbn = gbut.Button;
-            if (check_skip_cutscene_gamepad(gbn))
+            
+            if (check_skip_cutscene_gamepad(gbut))
             {
                 state_handled = true; // stop display
             }
             else if ((skip & SKIP_GAMEPAD) != 0 && !play.IsIgnoringInput() &&
-                    is_default_gamepad_skip_button_pressed(gbn))
+                    is_default_gamepad_skip_button_pressed(gbut))
             {
-                play.SetWaitSkipResult(SKIP_GAMEPAD, gbn);
+                play.SetWaitSkipResult(SKIP_GAMEPAD, get_gamepadkey(gbut));
                 state_handled = true; // stop display
             }
             break;
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -1606,10 +1606,11 @@ bool check_skip_cutscene_mclick(int mbut)
     return false;
 }
 
-bool check_skip_cutscene_gamepad(int gbut)
+bool check_skip_cutscene_gamepad(const GamepadInput &getinput)
 {
     CutsceneSkipStyle skip = get_cutscene_skipstyle();
-    if (gbut>0 && skip == eSkipSceneAnyKey)
+    
+    if (getinput.IsAnyButtonDown() && skip == eSkipSceneAnyKey)
     {
         start_skipping_cutscene();
         return true;
--- a/Engine/ac/game.h
+++ b/Engine/ac/game.h
@@ -26,6 +26,7 @@
 #include "gfx/bitmap.h"
 #include "main/game_file.h"
 #include "util/string.h"
+#include "joystick.h"
 
 // Forward declaration
 namespace AGS { namespace Common { class AssetManager; class Stream; } }
@@ -178,7 +179,7 @@ CutsceneSkipStyle get_cutscene_skipstyle();
 void start_skipping_cutscene ();
 bool check_skip_cutscene_keypress(int kgn);
 bool check_skip_cutscene_mclick(int mbut);
-bool check_skip_cutscene_gamepad(int gbut);
+bool check_skip_cutscene_gamepad(const GamepadInput &getinput);
 void initialize_skippable_cutscene();
 void stop_fast_forwarding();
 
--- a/Engine/ac/joystick.cpp
+++ b/Engine/ac/joystick.cpp
@@ -22,12 +22,20 @@
 #include "script/script_api.h"
 #include "ac/dynobj/dynobj_manager.h"
 
-bool is_default_gamepad_skip_button_pressed(eAGSGamepad_Button gbn)
+bool is_default_gamepad_skip_button_pressed(const GamepadInput &getinput)
 {
-    return (gbn == eAGSGamepad_ButtonA) ||
-    (gbn == eAGSGamepad_ButtonB) ||
-    (gbn == eAGSGamepad_ButtonX) ||
-    (gbn == eAGSGamepad_ButtonY);
+    if (getinput.isGamepad)
+    {
+        eAGSGamepad_Button gbn = getinput.GamepadButton;
+        return (gbn == eAGSGamepad_ButtonA) ||
+            (gbn == eAGSGamepad_ButtonB) ||
+            (gbn == eAGSGamepad_ButtonX) ||
+            (gbn == eAGSGamepad_ButtonY);
+    }
+    else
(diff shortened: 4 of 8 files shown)


ivan-mogilko:
@Dualnames1 please make proper commit descriptions that explain what this commit does. A short description on the first line (72 chars max), and then unlimited comment after two linebreaks if necessary.

See guidelines here: 
https://github.com/adventuregamestudio/ags/blob/master/CONTRIBUTING.md#commits

Note that you can force-push the same branch with modified commits, and PR will update automatically.

ericoporto:
Is get_gamepadkey meant to convert between AGS and SDL numbering of buttons? Or the reverse? I am curious if either of these already existed in the code…

Also, `const GamepadInput &gamepad`, I think `GamepadInput` is actually a button, so perhaps the better var name is button or what it was before.

> Fixing the issue where no joystick button would be able to skip dialog

This was intentional, it needs some way to configure the button for skipping in the script api.

Joystick should not have a default otherwise you need to remove your usb attached stuff or risk randomly skipping a game dialog.

Dualnames1:
> Is get_gamepadkey meant to convert between AGS and SDL numbering of buttons? Or the reverse? I am curious if either of these already existed in the code…
> 

Not that I could find.

> Also, `const GamepadInput &gamepad`, I think `GamepadInput` is actually a button, so perhaps the better var name is button or what it was before.
> 

Fine to change that, if you think so.

> > Fixing the issue where no joystick button would be able to skip dialog
> 
> This was intentional, it needs some way to configure the button for skipping in the script api.

But this works fine now.

> 
> Joystick should not have a default otherwise you need to remove your usb attached stuff or risk randomly skipping a game dialog.

There is no joystick to my knowledge that doesn't have a single button, but happy to adjust to remove the default.

ericoporto:
My fear was more that someone would had a joystick that would have an always pressed button connected to their pc - like a driving wheel or some of the aviation stuff. Unsure if that could be a real issue. Anyhow, if you can come up with an api to set the button/buttons for these skipping stuff it would be nice. I had no idea on this, which is why I didn’t add, but perhaps you had some idea when experimenting with this stuff.

ivan-mogilko:
@Dualnames1 there's some misunderstanding about commit naming. We do not need a current engine version in the commit name. We simply use a "Engine:" prefix to tag commits related to the engine code, but the commit name should mention its purpose. In your case, it could be something like: "Engine: fixed joystick not able to skip dialog".
You may check the commit history and see examples of how commits are named.

Looks like I must add some examples to CONTRIBUTING.md.

I will review the code itself later today.

ivan-mogilko:
@ericoporto 
> Anyhow, if you can come up with an api to set the button/buttons for these skipping stuff it would be nice.

Yesterday I've been thinking about opening a proposal for this, based on your Actions idea, where we introduce a Action struct in script and a number of built-in Action instances for the built-in skipping behavior (speech, display, video, cutscene).
Then I thought that maybe that's an overkill for this particular problem, as, with exception of Display command (which may be replaced by a custom overlay or gui) and blocking video (which is obsolete in ags4), user may script skipping of everything using available commands.

This leaves only a case when game developer did not script anything for joystick, but we still want joystick to work. In which case this seem to be a task for engine config, rather than script api.

EDIT: what we might need though, is "Skip*" properties of type "InputType", which would let users to tell if they want something to be skipped by a gamepad or not.

thadguidry:
Curious, do any of you foresee issues with accessibility or button remapping programs or Voice Attack in this issue?

ivan-mogilko:
> Curious, do any of you foresee issues with accessibility or button remapping programs or Voice Attack in this issue?

Please elaborate?

ivan-mogilko:
I gave this some thought, and to wrap up my view on this, there are 2 possible situations related to skipping a speech with joystick:

**Situation 1:** the game developer accounts for joystick controls in their game.

In this case they should not rely on any "default"  hard-coded behavior in the engine, but instead script the reaction to joystick themselves. Even if they use the standard speech they can still script the skipping, if they do so in "repeatedly_execute_always", tracking joystick buttons. That should already be possible to achieve, no changes to the engine necessary (afaik). For this case *there is no need in any hard-coded reaction to the joystick in the engine whatsoever*.

Same goes for skipping cutscenes.

Display is the only text message that cannot be skipped with a script, because no script is run while its on screen. But this means that developer should not use Display but use a custom scripted message instead, easily done.

PlayVideo cannot be skipped like that because  no script is run during a blocking video. The solution is to use new VideoPlayer api.

**Situation 2:**  the game did not account for joystick controls, but someone still wants to use a joystick to play it.

This is the situation where we need to make joystick buttons skip built-in states such as speech etc without script. But in this case, naturally, we cannot configure joystick behavior with script api. So the only solution seems to be a mapping setup, where engine would be configured to either emulate certain other input on joystick presses, or perform certain action directly. To some degree a similar thing probably may be achieved with a external remapping program.

**Conclusion:** it seems like we should not hardcode joystick buttons in the engine at all. The necessary script api for wiring joystick to skipping things already exists (this may be double checked and a proof of concept created). If we are speaking about playing a previously made game, then we should rather implement a input->action map in engine's config.

If it's *really* necessary, we may discuss some minimal script api for setting joystick buttons for skipping things like Display and standard speech. But this should mean that game developer explicitly sets things in script for this to work. Still nothing should be hardcoded in the engine.

EDIT: In fact, even the existing hardcoded reaction to controller buttons may be a bad thing. We'd need to do something of the above for gamepad buttons too (a action mapping in script api and/or in engine config).

Dualnames1:
The hardcoding is a separate issue though. But regardless, let's release the hold on 0-3 button ids on the joysticks, and do the same for controller? (releasing the hold on A,B,X,Y)

thadguidry:
Yes, hardcoded reactions to buttons is not ideal.  By not having hardcoded reactions in the Engine, then this allows a developer more flexibility via scripting. My use case would be scripting for custom button mapping to address accessibility concerns, i.e. overwriting some of SDL's defaults with wired and Bluetooth controllers via scripting.

Dualnames1:
> Yes, hardcoded reactions to buttons is not ideal. This allows a developer more flexibility via scripting. My use case would be scripting for custom button mapping to address accessibility concerns, i.e. overwriting some of SDL's defaults with wired and Bluetooth controllers via scripting.

It is most common in games from my knowledge to be able to assign a specific key (or 2 or any) to skip text for controller. But that should be up to the developer and perhaps a different PR altogether.

ivan-mogilko:
> The hardcoding is a separate issue though. But regardless, let's release the hold on 0-3 button ids on the joysticks, and do the same for controller? (releasing the hold on A,B,X,Y)

I am not sure if we understood each other. By "hardcoding" i meant hardcoded reaction to the gamepad/joystick button press that skips a speech and other things. I think that joystick should not be hardcoded to skip things in the engine, at all. Your most recent commit simply removes the button id range check; i think it makes things worse in case of gamepad, and makes things more unreliable overall.

> It is most common in games from my knowledge to be able to assign a specific key (or 2 or any) to skip text for controller. But that should be up to the developer and perhaps a different PR altogether.

You should already be able to do so in script, no additional changes to the engine is necessary for this, afaik.

Dualnames1:
I did misunderstand. So if joystick shouldn't skip things in the engine, then this PR is kinda useless entirely no?
(Removed last commit that removed the ABXY things in is_default_skip )

ivan-mogilko:
> So if joystick shouldn't skip things in the engine, then this PR is kinda useless entirely no?

Yes, I am afraid so. It took me a while to remember the situation and gather thoughts, so I could not say this earlier.

ericoporto:
One thing about the Gamepad, through an environment variable you can remap it - and perhaps we could also add a script API for remapping it. But, I think that Steam controller mapping (if you configure it in their ui) will set this environment variable and pass along for games run there, and SDL2 can then pickup this environment variable value when the joystic/gamepad (SDL Controller) system starts up. I haven’t retested this in recent years though.

PR #2997, opened 2026-06-27
Joystick IsAnyButtonDown

Dualnames1:
Engine: Added bool IsAnyButtonDown which works per joystick. 
             Fixed a possible exception with IsGamepad


ericoporto:
If you hold down any button and start the game, does this solution actually works?

In case doing this does break the above solution, my suggestion would be to not trust it directly and instead if it’s different than zero just check by going through all buttons of the joystick. I guess you are not doing so for some performance reason?

other two cases to check: alt+tab, button press, alt+tab, button release; press down to load a game, keep holding while the game loads, release and check if it is working.

ivan-mogilko:
@ericoporto good point, but I was the one who suggested this solution, because i used the same for keyboard (recently added function for keyboard): 9f2735daf38d6bfc0bdaadbd0ae10788b7a459ea
This was of course done to not iterate whole keyboard. I thought it makes sense to do the same for the joystick buttons.

The mouse button state has a similar variable since 3.6.0, except it's not a counter, but a bit-set.

AGS clears the remembered input state on enter/exit game window (lost focus, got focus).

Either this particular thing should not be cleared, or it has to be reinitializing full iteration of keys on the window switch-in.
I don't remember for certain if SDL2 tracks the input events when the window is out of focus. Maybe not.

EDIT: BTW, does it track joystick connection/disconnection events when the window is out of focus?

ivan-mogilko:
Looks like my mistake is worse than that, and maybe older. `ags_clear_input_state` is called on multiple occasions, such as: new room loaded, save restored etc. Initially, the idea was that this resets some things that are supposed to be ignored after certain game state changes, such as queued input events, so that they are not carried over.

But this does not make sense in case of recording control states for the purpose of reporting them to the script.
So we need to rework this function, maybe split it up into two... one that resets a input event queue on a game state change, and another that resets/reinitializes the state when the window looses/receives a focus.

ericoporto:
I think the keyboard has GetKeyboardState for this

> EDIT: BTW, does it track joystick connection/disconnection events when the window is out of focus?

good point, I don’t remember

ivan-mogilko:
Of course, I know that SDL returns an array of key states with SDL_GetKeyboardState. My meaning is that I did not want to iterate this whole array whenever a script function is called, thought the counter would work better.

ivan-mogilko:
@Dualnames1 , I think it will be easier for now to use your original variant, where you just check each joystick button state when this function is called. i would have to fix few things in controls when i get spare moment. The "any button" check could be optimized later, if necessary.

Dualnames1:
But this works fine, i checked it? The clearinput part is not even required tbh. And I can remove it. When a device is disconnected the joystick is removed entirely from the thing anyway.

Imho, because of the likely chance from personal experience to use isAnyButtonDown on a repeatedly execute, i'd prefer it if it was performant. Anyway, if you insist i'll do my rendition.

ericoporto:
About performance in script, my experience is script is faster than the API, so what I do is I run once per frame and maintain my own array in script with a copy of the button states. This is usually faster than asking the api twice or more. I can share my script code later. Edit: actually, a GetButtons that would retrieve an array with all buttons states would be nice - and same for the other joystick stuff.

Dualnames1:
I checked your scenarios and a button is released properly when I remove joystick_clearinput. @ericoporto
My suggestion is to remove the joystick_clearinput part for the time being, it's not needed anyway.

> EDIT: BTW, does it track joystick connection/disconnection events when the window is out of focus?

Yes it will return false once the joystick is disconnected.

ivan-mogilko:
Alright, maybe do this way, but I will need to double check possible edge cases and do fixes to this "clear input" function that i mentioned. There's still a problem with IsAnyKeyPressed (and maybe mouse equivalent too).

> My suggestion is to remove the joystick_clearinput part for the time being, it's not needed anyway.

Yes, it's likely not needed.

Dualnames1:
Did a rendition that's closer to what @ericoporto suggested.

ivan-mogilko:
If you're sure going with this implementation for now, could you squash the commits into a single one please?

Also, about the commit descriptions, try having not more than 72 characters in the first line. You can write more details after two linebreaks.

ericoporto:
Hey, not sure if you ever tried Duals, but GitHub Desktop has a good ui to squashing, reordering e and other stuff when authoring commits. :)

Dualnames1:
> Hey, not sure if you ever tried Duals, but GitHub Desktop has a good ui to squashing, reordering e and other stuff when authoring commits. :)

I'm an old man, I will not abandon gitbash. I'm just bad at understanding guidelines, we should be good now.

Engine/ac/joystick.cpp
```diff
@@ -252,6 +252,34 @@ ScriptJoystick* Joystick_GetiJoysticks(int i)
...
+        {
+            if (SDL_GameControllerGetButton(joy_impl.sdlGameController, static_cast<SDL_GameControllerButton>(butt)))
+            {
+                return true;
+            }
+        }
+    }
+    else
+    {
+        int button_count = SDL_JoystickNumButtons(joy_impl.sdlJoystick);
```
ivan-mogilko: I have a thought that we may save number of buttons in JoystickImpl when the joystick is opened. My understanding is that number of buttons cannot change while device is connected, is that correct?

Engine/ac/joystick.cpp
```diff
@@ -385,6 +414,12 @@ RuntimeScriptValue Sc_Joystick_IsConnected(void *self, const RuntimeScriptValue
     API_OBJCALL_INT(ScriptJoystick, Joystick_IsConnected);
 }
 
+// int (ScriptJoystick *joy)
+RuntimeScriptValue Sc_Joystick_IsAnyButtonDown(void* self, const RuntimeScriptValue* params, int32_t param_count)
+{
+    API_OBJCALL_INT(ScriptJoystick, Joystick_IsAnyButtonDown);
```
ivan-mogilko: Should use `API_OBJCALL_BOOL` here, since the function returns bool.

Engine/ac/joystick.cpp
Dualnames1: Correct.

Engine/ac/joystick.cpp
Dualnames1: Ah, you right.

Engine/ac/joystick.cpp
Dualnames1: Did a thing. I think this is good, if u disagree let me know, if not, I'll resolve the conversation.

Dualnames1:
I took @ivan-mogilko 's idea and implemented it to keep keys on a variable when opening a joystick/controller. And fixed that return [int to bool]

Engine/ac/joystick.cpp
```diff
@@ -150,6 +150,7 @@ struct JoystickImpl {
     SDL_Joystick* sdlJoystick = nullptr;
     SDL_JoystickID sdlJoystickID = -1;
     int scriptJoystickHandle = 0;
+    int sdljoybutton_count = 0;
```
ivan-mogilko: The implementation look fine to me, but the naming style is inconsistent in this struct, and also all of it is not compliant to our preferred naming style where public struct members should be called using CamelCase style.

Could you fix it please?
I.e. `sdljoybutton_count` should be `SDLJoyButtonCount`, or just `ButtonCount` for simplicity. Other members may be fixed accordingly.

Dualnames1: Proposed name then?

ivan-mogilko: ButtonCount is a good name, there's no need to have "SDLJoy" in front, as we know it's joystick buttons since it's a joystick struct.

Dualnames1: Thank you!

ivan-mogilko:
About tracking up and down events and alt+tab, I did a test, and found that:
1. If you keep joystick button down while in game, switch out, and release, then SDL2 sends the button up event.
2. If you switch out, press joystick button, and switch into the game, then SDL2 does NOT send the button up event.

This means that it may be safe to count the button up/down events, as SDL2 only sends them when the press was "belonging" to our window, so "up" event comes if "down" event was triggered earlier.

I also confirmed that SDL2 sends joystick connect/disconnect events when window is not focused.

EDIT:
With keyboard it's little different. If I hold down a key elsewhere and switch in to the game window, SDL will immediately triggers "keydown" event. Double checked that it's not a key repeat (there's a field in key event that allows to read that).
Maybe SDL sends this event whenever the focus is received and some key were already pressed.

ericoporto:
About keyboard, I usually don’t use on_key_press event, instead I use IsKeyPressed(), because then if I am using to control moving something on the screen, then I don’t depend on the “speed” the OS reissues repeat key presses (Windows is usually configured to send a lot more), and then I have more adjustments I can do that rely only on the game fps.

If there was a GetPressedKeys() that returned an array of keycodes, I could use a for and switch clause through script to parse through all the pressed keys, which could make the code slightly nicer, but I don’t know if such things exists on SDL side.

ivan-mogilko:
> If there was a GetPressedKeys() that returned an array of keycodes, I could use a for and switch clause through script to parse through all the pressed keys, which could make the code slightly nicer, but I don’t know if such things exists on SDL side.

Alternatively we could add separate key_down/key_up events in script, and then one could store states of the keys they need in game script.

ericoporto:
I am not super fan of these events because I keep second guessing if I know for sure the current state, and I am more interested in the state than the transition.

ivan-mogilko:
Very well, I suppose we may record a list of pressed keys, and update it with these events in the engine.
AFAIK that's how SDL2 updates its own keystates array.

Issue #1997, opened 2023-05-03
Allow acessing DynamicSprite data as array


ericoporto:
**Describe the problem**
I have previously had the need to access data of individual pixels in sprites in AGS, either because images are convenient when storing information ([I rented a boat Height Maps](https://github.com/ericoporto/i_rented_a_boat/blob/main/I_rented_a_boat/Sprites/Maps/area1/area1_heightmap_ags.png) or [Galleys maps](https://github.com/ericoporto/galleys/blob/ca3fbd659609324204bac49d21977e10db4b6b1f/DungeonCraw/MapData.asc#L9-L22)), or because I wanted some particular effect ([normal maps](https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/using-normal-maps-demo/)). GetPixel and SetPixel methods are unfortunately slow for this - even if they didn't had boundary checks or vtables, it's still a per pixel call to an "external" function, so there will be some overhead - so you can only do few thousand (~2048) GetPixel per frame, without causing slowdowns, requiring one to divide the load. Writing pixel information is even slower. Pixel manipulation could also  produce graphic effects (a better [distfx](https://www.adventuregamestudio.co.uk/forums/index.php?topic=60380.0))!

**Suggested change**
It would be nice to retrieve the bitmap data from inside the DynamicSprite (or Surface?) directly as an array. Though, in AGS, dynamic arrays must have headers, and cannot point to any random memory.

- If we want to have only reads, maybe as a copy, optionally copying them back on "Release" call, or some kind of SetPixelsBack call.
  - this may be easier to code
- if we want it to be read and written, and written fast, the array could be the data itself, with shared ownership.
  - it is possible to avoid extra dereferencing, and possible to give out raw pointer to script vm too, as dynamic objects are already stored in pairs of data address / manager, and manager can be an object itself.
  - there will be a little extra memory for allocation
   - there may be ways of optimizing this by using memory pools or allocating dynamicobject class and array data on the same buffer.

The API for this could be, depending on either convenience or going for the more generic solution, something like

```AGS Script
import char[] GetPixelArray(); // if we are talking about 32-bit color, this retrieves per color component (4 * w * h)
import short[] GetPixelArray16(); // classic ags color is 16-bit
import int[] GetPixelArray32();
```

I talked previously about t his with @ivan-mogilko , I hope I added the things mentioned correctly! It's possibly necessary to observe #1980 when implementing this.

ivan-mogilko:
[...]

PR #2998, opened 2026-06-27
4.0: implement DrawingSurface.GetPixelsCopy, SetPixels

ivan-mogilko:
Suddenly realized that we never implemented the functions for working with arrays of pixels from DrawingSurface, discussed as an alternative to #1997, so doing this now (#1997 is still under question, i might revisit it once i get more spare time, and see if there are opportunities to implement it in AGS 4).

This adds two script functions:
```
char[] DrawingSurface.GetPixelsCopy(int x = 0, int y = 0, int width = -1, int height = -1);
void DrawingSurface.SetPixels(char pixels[], int x = 0, int y = 0, int width = -1, int height = -1);
```

`GetPixelsCopy` copies a surface region and returns as a dynamic array of bytes (remember, in AGS script `char` is a *unsigned byte*, so capable of storing full range value 0-255). The meaning of bytes depends on the image format. In 8-bit images each byte is a pixel which value equals a palette index. In 32-bit images each byte is a component of RGBA (so 4 bytes make a pixel). We do not support 16-bit images in AGS 4.0 anymore, but if we did their case would be complicated, as RGB are encoded in 2 bytes as R5G6B5 and must be retrieved using bitwise operations.

`SetPixels` does the opposite, it copies the contents of the dynamic array of bytes onto the requested position in the image.
The `width` parameter there is very important as engine must know how much to copy per line. Height may be provided, but the height's limit is also calculated from the total length of the array, so that array is not exceeded when copying data.

Couple of simple examples:
```
void noloopcheck DoPixels()
{
  DrawingSurface* ds = Room.GetDrawingSurfaceForBackground();
  char pixels[] = ds.GetPixelsCopy(100, 100, 50, 50);
  for (int i = 0; i < pixels.Length; i++)
    pixels[i] = pixels[i] + 10;
  ds.SetPixels(pixels, 150, 100, 50);
  ds.Release();
}
```
this will make a copy of a 50x50 rect and paste it next to the original position, while slightly adjusting color values.

```
void noloopcheck DoPixels32()
{
  DrawingSurface* ds = Room.GetDrawingSurfaceForBackground();
  int pixels[] = ds.GetPixelsCopy32(100, 100, 50, 50);
  for (int i = 0; i < pixels.Length; i++)
    pixels[i] = pixels[i] & 0x00FF0000;
  ds.SetPixels32(pixels, 150, 100, 50);
  ds.Release();
}
```
this will make a copy of a 50x50 rect and paste it next to the original position, while turning all pixels into hues of red.

@ericoporto do you think we may need variants of functions that work with 32-bit int arrays, so that you can handle each pixel as a int number? These could be implemented as wrappers, calling base ones internally.

PS. I might actually backport this to 3.6.3 too, since this is a simple enough function.

### DIFF
--- a/Common/gfx/allegrobitmap.cpp
+++ b/Common/gfx/allegrobitmap.cpp
@@ -584,6 +584,13 @@ int AGSColorToBitmapColor(int color, int /*color_depth*/)
     return color;
 }
 
+int BitmapColorToAGSColor(int color, int /*color_depth*/)
+{
+    // no conversion necessary, we assume that "ags color" is matching
+    // palette index in 8-bit mode and 32-bit A8R8G8B8 in 32-bit mode
+    return color;
+}
+
 void AGSColorToRGB(int color, int color_depth, RGB &rgb)
 {
     if (color_depth == 8)
--- a/Common/gfx/allegrobitmap.h
+++ b/Common/gfx/allegrobitmap.h
@@ -304,6 +304,8 @@ namespace BitmapHelper
 {
     // Remaps AGS color number to a color value compatible to a bitmap of certain color depth
     int AGSColorToBitmapColor(int color, int color_depth);
+    // Remaps a bitmap pixel of certain color depth to AGS color number
+    int BitmapColorToAGSColor(int color, int color_depth);
     // Remaps AGS color number in certain color depth mode to a actual RGB
     void AGSColorToRGB(int color, int color_depth, RGB &rgb);
     // TODO: revise those functions later (currently needed in a few very specific cases)
--- a/Common/gfx/bitmap.cpp
+++ b/Common/gfx/bitmap.cpp
@@ -104,8 +104,9 @@ Bitmap *CreateBitmapFromPixels(int width, int height, int dst_color_depth,
     if (!bitmap)
         return nullptr;
 
-    if (!PixelOp::CopyConvert(bitmap->GetDataForWriting(), ColorDepthToPixelFormat(dst_color_depth),
-            bitmap->GetLineLength(), width, height, pixels, ColorDepthToPixelFormat(src_col_depth), src_pitch))
+    if (!PixelOp::CopyConvert(pixels, ColorDepthToPixelFormat(src_col_depth), src_pitch,
+            width, height, bitmap->GetDataForWriting(), ColorDepthToPixelFormat(dst_color_depth),
+            bitmap->GetLineLength()))
         return nullptr;
 
     return bitmap.release();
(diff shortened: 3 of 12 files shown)


ericoporto:
I think as int would be easier to work with since one can use `#801040ff` like colors, but my intention for using this was mostly to recover things I encode as sprites (like maps), because today I use GetPixel and it takes a long time to retrieve all pixels of an image.

The good thing of your as byte implementation is I guess for retrieving the masks (walkable area and all that) as an array if you need it for some reason.

ericoporto:
This looks like it is working afaict. I ressurected my Mode7 code from the issue, and it appears it is indeed slightly faster than GetPixel - I get it running at 13 FPS from 11 FPS using GetPixel, so the improvement may not look like much, but I verified it separately and the getting and setting of the array is really super fast. I haven't played with the byte stuff, but I think they may be cool for simple effects one could make for a sprite - although not meant to be repeatedly applied.

[ArrayMode7.zip](https://github.com/user-attachments/files/29950031/ArrayMode7.zip)

ivan-mogilko:
Idk if that's what you were doing, but I think that if you're editing an image continuously, then you may precreate that dynamic array, and skip GetPixelsCopy entirely, only edit your array and call SetPixels.

One last thing that bothers me is big-endian systems, it's possible that they will need a ARGB swap both when returning and setting array to do correct result with the array of ints. Unfortunately i dont have a big-endian system to test this, so I need to think this over a bit.

ericoporto:
> Idk if that's what you were doing, but I think that if you're editing an image continuously, then you may precreate that dynamic array, and skip GetPixelsCopy entirely, only edit your array and call SetPixels.

Yeah, I was doing that, but thanks! It looks like the for clause that goes by every pixel of half the screen was the biggest cost, that and the mapping made things slow - like removing most of the math but setting each pixel of the array with one pixel of the other array before setting still only bumped to around 21fps. Anyway, but that is not related to this, it’s on AGS for clauses. Also, I discovered I can bump the fps slightly by adding the noloopcheck attribute to the function.

Edit: ah, discovered that reading from a static array is faster in a separate specific test, this is useless for the mode7 one, but it does speed up my other test which is a simpler for like the examples you added above in your PR description.

<details>

```AGS Script
// room script file
char src[320*200*4];
char dst[];
int w, h;

function hGlowingOrb_Look(Hotspot *thisHotspot, CursorMode mode)
{
  player.Say("It is the second best glowing orb that I've seen today.");
}

int kk;
function noloopcheck room_RepExec()
{
  kk++;
  int len = dst.Length;
  
  for (int i = 0; i < len; i++)
    if(i%4!=3)
    dst[i] = src[i]/kk;
  
//  for(int x=0; x<w; x++) {
//    for(int y=0; y<h; y++) {
//      dst[4*(x+w*y)  ] = src[4*(x+w*y)  ]/kk;
//      dst[4*(x+w*y)+1] = src[4*(x+w*y)+1]/kk;
//      dst[4*(x+w*y)+2] = src[4*(x+w*y)+2]/kk;
//      //dst[4*(x+w*y)+3] = src[4*(x+w*y)+3]/kk;
//    }
//  }
  if(kk>60) kk = 0;
  
  DrawingSurface* ds = Room.GetDrawingSurfaceForBackground();
  ds.SetPixels(dst);
  ds.Release();
}

function noloopcheck room_Load()
{
  Game.Speed = 60;
  System.DisplayFPS = true;
  DrawingSurface* ds = Room.GetDrawingSurfaceForBackground();
  System.Log(eLogError, "depth: %d", ds.ColorDepth);
  w = ds.Width;
  h = ds.Height;
  dst = ds.GetPixelsCopy();
  for(int i=0; i<dst.Length; i++)
  {
    src[i] = dst[i];
  }
  ds.Release();
}
```

</details>

PR #3000, opened 2026-06-30
Script API: implement GetPressedKeys()

ivan-mogilko:
Implements a function:

```
/// Gets a dynamic arrays of keys that are currently held down.
eKeyCode[] GetPressedKeys();
```

The returned array allows to iterate all pressed keys. Example:
```
function repeatedly_execute_always()
{
  eKeyCode keys[] = GetPressedKeys();
  String s = "";
  for (int i = 0; i < keys.Length; i++)
  {
    s = s.Append(String.Format("%d ", keys[i]));
  }
  Label1.Text = s;
}
```

The array is a copy of the current state, it does not update automatically, so storing it for more than 1 game frame is meaningless.

Possible issues:
1. Made this a function in a global scope, because other key-related functions such as IsKeyPressed and IsAnyKeyPressed are also in global scope. We might wrap these functions in some struct, but idk which one. Mouse's IsButtonDown is in a Mouse struct, but there's no Keyboard struct atm; would that be convenient if they are in Keyboard struct?
2. While testing, i noticed that some key combinations with 3+ keys dont work. Other combinations work even if 5 keys are pressed at the same time. Could be keyboard driver issue, no idea.

### DIFF
--- a/Common/ac/keycode.h
+++ b/Common/ac/keycode.h
@@ -326,7 +326,8 @@ inline bool IsAGSServiceKey(eAGSKeyCode keycode)
     return keycode >= eAGSKeyCode_FirstServiceKey;
 }
 
-// Converts eAGSKeyCode to script API code, for "on_key_press" and similar callbacks
+// Converts eAGSKeyCode to script API code, for script callbacks (e.g. on_key_press),
+// and matching the user-set key variables.
 eAGSKeyCode AGSKeyToScriptKey(eAGSKeyCode keycode);
 // Converts eAGSKeyCode to ASCII text representation with the range check; returns 0 on failure
 // Not unicode compatible.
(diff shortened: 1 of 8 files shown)


ericoporto:
> While testing, i noticed that some key combinations with 3+ keys dont work

I think this may be Hardware+Driver limitations, the issue is called “key ghosting”, and it appears the solution is called “rollover”, I found a website that lets one test this: https://www.typequicker.com/tools/keyboard-rollover-test

also I will test this, thanks for it (I am making the test game).

ericoporto:
Hey, I am having an issue here, it looks like the letter keys are wrong? Here is the table:
 https://adventuregamestudio.github.io/ags-manual/v3/Keycodes.html

If I press A in the keyboard, I don't get 65, I get like 97?

I started a small test project, but I only managed to add the numbers and the letters in the logic (it's in Keyboard.asc file), and noticed it only works for the numbers. **[get_keys_test.zip](https://github.com/user-attachments/files/29580257/get_keys_test.zip)**

ivan-mogilko:
> If I press A in the keyboard, I don't get 65, I get like 97?

Looks like somehow this returns lowercase letters.

ivan-mogilko:
@ericoporto does it work for you now, after the fix?

ericoporto:
[get_keys_test.zip](https://github.com/user-attachments/files/29668882/get_keys_test.zip)

Yes it does! I finished my keyboard test game and attached above if you want to try.

PR #3001, opened 2026-07-02
Add new "AGS Quest" demo game to the AGS Installer

### LINKED ISSUE #2994: 3.6.3: add new Demo Game to Editor distribution
The new Demo Game was created by a community effort, and stored here:
https://github.com/adventuregamestudio/ags-demo-game/

Need to add this as a *optional component* to the AGS installer (but probably not zip archive?).

The CI task should probably take the demo game from the latest tag or release tag.

ivan-mogilko:
Resolve #2994

My first time writing Github Actions, not sure how successful...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -740,6 +740,7 @@ jobs:
       - windows-msvc-engine
     env:
       TEMPLATES_REPOSITORY: adventuregamestudio/ags-templates
+      DEMOGAME_REPOSITORY: adventuregamestudio/ags-demo-game
       AGS_SDL_LIB: C:\Lib\SDL2\lib\x86
     steps:
       - uses: actions/checkout@v6
@@ -827,6 +828,12 @@ jobs:
           call Script\setvar.cmd ACI_VERSION_STR
           tar -f _android_tmp\AGS-%ACI_VERSION_STR%-android-proj-release.zip -xvzC Windows\Installer\Source\Android
           rd /s /q _android_tmp
+
+      - name: Download Demo game
+        shell: cmd
+        run: |
+          mkdir Windows\Installer\Source\AGSQuest
+          curl -fLSs https://github.com/%DEMOGAME_REPOSITORY%/releases/latest/download/AGSQuest.zip | tar -f - -xvzC Windows\Installer\Source\AGSQuest --strip-components 1
 
       #   NOTE: Inno Setup 6.7.1 comes preinstalled in Windows runners.
       #   It looks like it is also in PATH already, but our script expects its
--- a/Windows/Installer/ags.iss
+++ b/Windows/Installer/ags.iss
@@ -16,7 +16,10 @@
 #else
 #define AgsVerNameStr AgsName + ' ' + AgsFriendlyVersion + ' ' + AgsSpVersion
 #define AgsOutputFile 'AGS-' + AgsFullVersion + '-' + AgsSpVersion
-#endif 
+#endif
+
+; Registry key that contains last installation path for the demo game
+#define DemoGamePathKey "DemoGamePath_36x"
 
 [Setup]
 ; NOTE: The value of AppId uniquely identifies this application.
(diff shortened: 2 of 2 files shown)


ivan-mogilko:
So, there's a interesting problem: we have AGS Quest without spritefile in the repository, so it's not available, which makes AGS Editor show a warning when opening this project, and suggest to recreate the file from sprite sources. This may be confusing to users.

There are two options which I see here:
- Regenerate this file as a part of the GHA work, when gathering the sources for installer. This means that it must run the editor, but i think it's okay, since editor is prepared in this task anyway? Also the installer will become slightly bigger because of this file.
- Regenerate this file as  a part of installation, which means that user will see the editor window spawning and recompiling something. Could be better if editor could run without main window being visible when launched from command line...

BTW regardless of when this is done, this will upgrade the game project to the current version, which is a good thing.

ericoporto:
Hey, I will look at it in detail, but two questions come to mind - more about the original issue.

- The Demo Game today only has one branch, is the idea to keep it for now AGS 3 only or should support multiple branches there too?
- About the building it to regenerate the sprite cache, in the current GHA we eventually build the editor, so we could run this after to use it to regenerate (but it would do so everytime) or alternatively we could use a CI in the demo quest own repository and have it package something there, but then we need to make releases there to have somewhere to attach the project. Which one is preferred?

ivan-mogilko:
> The Demo Game today only has one branch, is the idea to keep it for now AGS 3 only or should support multiple branches there too?

I suppose, there must be a version made for ags4, because it takes certain effort to upgrade a project, a task not suitable for the target audience of this demo game.
At the same time, I wish there are not several 3.x versions, because that will quickly become tedious to maintain them all. The purpose of this demo game is to display the core concepts of AGS, not every single feature.

> About the building it to regenerate the sprite cache, in the current GHA we eventually build the editor, so we could run this after to use it to regenerate (but it would do so everytime) or alternatively we could use a CI in the demo quest own repository and have it package something there, but then we need to make releases there to have somewhere to attach the project. Which one is preferred?

Perhaps the second one, also because some users may need to be getting this game releases separately from the Editor releases.

ivan-mogilko:
Changed to simply downloading and unpacking prepared zip.

I created a project zip by hand for now:
https://github.com/adventuregamestudio/ags-demo-game/releases/tag/v0.1.0.0

ericoporto:
This looks great, and it works. I like that the installer asks where it should go so one can open later too. And I think this approach of relying on the project being packed there in the AGS Quest repository is better - it could eventually have a ags4 version, using the approach of having a single ags4 and ags3 release if needed.

I did found a weird issue when playing with the demo game, where clicking far for the character to walk causes the character to keep walking in place.

https://github.com/user-attachments/assets/83eb7db2-70ff-44c9-8ab8-18126cdf1ef0

ericoporto (approved):

ivan-mogilko:
> I did found a weird issue when playing with the demo game, where clicking far for the character to walk causes the character to keep walking in place.

This is either a bug in the game (idk if it may have any related script), or a bug in the engine, please open a dedicated ticket in either repository.

ivan-mogilko:
@ericoporto 
> I did found a weird issue when playing with the demo game, where clicking far for the character to walk causes the character to keep walking in place.

I just accidentally triggered a similar bug in another person's game.
The culprit is a `onpart` value (walking progress), which somehow becomes NaN at some point.

I will try making a small test game and open a bug issue.

ericoporto:
I didn’t open an issue because I couldn’t isolate it. 
Notice that it keeps going forever! I was curious if it was related to the smooth transition recent change.


I did found an old issue, but it kept going and stopped. And it was already fixed. https://github.com/adventuregamestudio/ags/issues/935

ivan-mogilko:
It's true, this is caused by the very recent commit:
master: 86cf6bf02e21da602e50a2222ad1e2997499f2aa
release-3.6.2: 0c6f8b0b1c5da06757fd6db78e27b60c88a29b51

PR #3003, opened 2026-07-03
Editor: Copy/Paste commands for global game items

### LINKED ISSUE #601: Duplicate/Clone GUI

ivan-mogilko:
Resolve #601

Implement Copy/Paste context menu commands for global game items.

This allows to copy/paste:
* Characters
* Cursors
* Dialogs
* GUIs
* Inventory Items
* Views

### DIFF
--- a/Editor/AGS.Editor/Components/CharactersComponent.cs
+++ b/Editor/AGS.Editor/Components/CharactersComponent.cs
@@ -1,3 +1,5 @@
+using AGS.Editor.TextProcessing;
+using AGS.Types;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
@@ -5,8 +7,6 @@ using System.Linq;
 using System.Text;
 using System.Windows.Forms;
 using System.Xml;
-using AGS.Types;
-using AGS.Editor.TextProcessing;
 
 namespace AGS.Editor.Components
 {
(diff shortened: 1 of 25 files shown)


Editor/AGS.Editor/Utils/ClipboardUtils.cs
```diff
@@ -0,0 +1,76 @@
+﻿using AGS.Types;
+using System;
+using System.Collections.Generic;
+using System.Net.NetworkInformation;
```
ericoporto: This using `System.Net.NetworkInformation`  is probably a typo 

Image

Editor/AGS.Editor/Components/CharactersComponent.cs
```diff
@@ -1,21 +1,24 @@
+using AGS.Editor.TextProcessing;
+using AGS.Types;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
 using System.Text;
+using System.Web.UI;
```
ericoporto: You probably don't need System.Web.UI

Image

ivan-mogilko:
Sometimes MSVS inserts these using namespace when i choose wrong autocomplete item.

ericoporto:
Yes, I get that too, but sometimes it’s a collision between the name of something in AGS.Types and something from somewhere else.

I did try this and it looks like it works.

PR #3006, opened 2026-07-07
Feature/ags commandline mcp

Verbari93:
## Summary

Adds `AGS.CommandLine`, a new console application that implements a
[Model Context Protocol (MCP)](https://modelcontextprotocol.io/) server
over stdio. This allows AI coding assistants (Kiro, Claude Desktop,
Cursor, etc.) to build and inspect AGS games programmatically through a
structured JSON-RPC 2.0 API.

## What it does

The server exposes 47 tools across five categories:

- **Read** — query characters, rooms, hotspots, regions, walkable areas,
  dialogs, inventory, GUIs, views, audio clips, and script modules
- **Write** — create/update all the above entities; import audio files
- **Events** — bind interaction handlers to entities; write and update
  AGS Script functions in `.asc` files; declare global variables
- **Build** — validate project file integrity, report compiled output
  state, launch the game executable, export a distribution zip.
  Note: the MCP server cannot invoke the AGS compiler itself — script
  compilation still requires the AGS Editor.
- **Room** — read/write hotspots and objects in `.crm` binary files via
  the existing `AGS.Native` C++/CLI bridge

## Architecture

- Pure .NET Framework 4.6 console app — no new framework dependencies
- Reuses `AGS.Types` (game data model) and `AGS.Native` (room binary I/O)
- MCP JSON-RPC 2.0 implemented in ~200 lines — no external MCP SDK
- `Newtonsoft.Json 13.x` (already used by the solution)
- Embedded `blank_room.crm` template for new room creation without
  requiring an existing compiled room

## Changes to existing Editor files

**`AGSEditor.csproj`**
- irrKlang references made conditional: `Exists()` guards on `<Reference>`
  and `<Content>` entries mean the DLL is only linked when present in
  `Editor\References\`. `IRRKLANG_AVAILABLE` is auto-defined via MSBuild
  when the DLL is detected, so contributors without irrKlang build cleanly
  and contributors with it retain full audio preview unchanged.
- Added missing `<Compile>` entry for `Utils\ClipboardUtils.cs` (file was
  present on disk from a recent merge but absent from the project file).

**`IrrklangPlayer.cs`**
- Stub in the `#else` branch was missing `GetMediaInfo()` required by
  `IAudioPreviewer`; added returning `new MediaInfo(0, 0, 0)`, matching
  the existing pattern in `MidiPlayer.cs`.

## Documentation

Full build instructions and MCP tool reference included under
`Editor/AGS.CommandLine/docs/`.

## Testing

Built and tested on Windows with VS 2022 Community, .NET Framework 4.6.
All 47 tools exercised against a real AGS 3.6.x game project.

## Notes

This was developed with AI assistance (Kiro). I am not deeply fluent in
C# and welcome feedback and suggestions on code style and conventions.

I did not open a tracker issue beforehand — this project started as a
personal experiment and the decision to submit it upstream came only once
it was working. I am happy to open an issue first if the maintainers
prefer that workflow before reviewing this PR.


Verbari93:
@ivan-mogilko @ericoporto Would appreciate your review when you have time.
This adds an MCP server for AI-assisted game development — happy to discuss
the approach or make changes based on your feedback.

ericoporto:
I would leave to @ivan-mogilko , but my perception is this can be made as a plugin that is maintained separately by whoever is the plugin author in their own repository. In ags4 you can just edit the files (due to open rooms) directly afaict and isn’t necessary, and one can just use skills, so again, don’t see any gain there.

Verbari93:
> I would leave to @ivan-mogilko , but my perception is this can be made as a plugin that is maintained separately by whoever is the plugin author in their own repository. In ags4 you can just edit the files (due to open rooms) directly afaict and isn’t necessary, and one can just use skills, so again, don’t see any gain there.

@ericoporto thanks for the quick reply, makes sense!

Yeah, keeping it as a standalone tool is probably the right call — I didn't really
think about the maintenance side when I submitted this.

The MCP concept still feels useful to me even for ags4, mostly because structured
tool calls are much cheaper on tokens than having an agent read/write large files
directly. But you're right it doesn't need to be this complex — a simpler parser
would do the job once the 4.0 formats settle down.

For now I'll just keep using this version on the side and see how things evolve.
Closing this for now, thanks again!

ivan-mogilko:
This is closed before I could post a comment, but I will just add that this better be in a separate repository, because our main repository already has much content in it which we must maintain, with Editor, Engine, and standalone game compilation tools, so it's best not overload it.

Whether to make this a editor plugin or a standalone application is a question of effort, I believe it's possible to do both.

If there will be a real interest from potential users we could even add a fork of such project into our github organization in the future (just saying).

Verbari93:
> This is closed before I could post a comment, but I will just add that this better be in a separate repository, because our main repository already has much content in it which we must maintain, with Editor, Engine, and standalone game compilation tools, so it's best not overload it.
> 
> Whether to make this a editor plugin or a standalone application is a question of effort, I believe it's possible to do both.
> 
> If there will be a real interest from potential users we could even add a fork of such project into our github organization in the future (just saying).

Sorry for closing it before you could reply, got a bit trigger-happy there!

Thanks for taking the time anyway — really appreciated. Makes total sense to keep 
it out of the main repo, I honestly should have thought about that before submitting.

I'll try to spin it out into a proper separate repo at some point, just to keep 
things cleaner and easier to follow for anyone who might be interested. We'll see 
how it goes!

PR #3010, opened 2026-07-13
4.0: add optional script names to Regions, WalkableAreas and Walkbehinds

### LINKED ISSUE #2951: AGS 4.0: add ScriptName to all Room Area types which dont have one
Per user request, have all area kinds feature a ScriptName, similar to how Hotspots do have that.

There are two use cases for this:
1. Obviously, script will have generated variables using these names, so them could be used instead of say "region[10]".
2. Since all areas are still allocated at once when the room is created, having them named would let user to know which are actually in use.

What is to be done:
1. Region, WalkableArea and Walk-behind have an optional ScriptName property in the editor. Default value is empty string.
2. These have a readonly ScriptName property in script API.
3. Add GetByName() function to these types in script API, similar to existing function in all the other types.
4. For those areas that have script name assigned by user, Editor generates variables of corresponding types in the room script.

ivan-mogilko:
Resolve #2951

1. In the Editor: adds ScriptName property to Region, WalkableArea and Walkbehind, meaning a script name. This is an optional property, initialized with empty string by default. They follow same rule as other script name properties in the project: they must be unique and suitable for a variable name.
2. Editor generates script variables of respective pointer types using these names (for those which have one set).
3. Engine exports the respective room elements as script variables (when the room is loaded).
4. Script API: added ScriptName readonly property and GetByScriptName() static function for Region, WalkableArea and Walkbehind types.

### DIFF
--- a/Common/game/room_file.cpp
+++ b/Common/game/room_file.cpp
@@ -468,6 +468,43 @@ HError ReadExt_400_ObjectOptions2(RoomStruct* room, Stream* in, RoomFileVersion 
     return HError::None();
 }
 
+HError ReadExt_400_AreaNames(RoomStruct* room, Stream* in, RoomFileVersion data_ver)
+{
+    // NOTE: Object names and script names were historically written in
+    // blocks kRoomFblk_ObjectNames and kRoomFblk_ObjectScNames, but limited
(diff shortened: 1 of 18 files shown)


ivan-mogilko:
Updated after #3014 to bring property names to the new style.

PR #2936, opened 2026-02-24
C++ ImGui based editor

mgerhardy:
Bildschirmfoto vom 2026-02-23 15-32-49
Bildschirmfoto vom 2026-02-23 15-32-34
Bildschirmfoto vom 2026-02-23 15-29-18

# TODO

- [ ] Shortcuts
- [ ] Docking version of imgui was used - let's use it and replace the fixed layout
- [ ] some editors allow to scroll to the right side and some buttons are hidden there in the tool bar
- [ ] Debugger isn't working on linux

PR #3011, opened 2026-07-13
Tools: add native agsbuild game compiler

mgerhardy:
Add a headless native build pipeline that compiles scripts, writes `game28.dta`, packages `Compiled/Data/<GameFileName>.ags` and generates VOX/TRA/config outputs without copying engine binaries and without `AGSEditor /compile`.

Extracted from imgui editor branch: https://github.com/adventuregamestudio/ags/pull/2936


ivan-mogilko:
I can see that a big effort was put into this, but after the first glance, I am getting an impression that this is not exactly the tool that we'd wish to have.

Please correct me if I'm wrong, but I notice that you've made a single tool that does everything at once at one go, from Game.agf to the final game distributable?

The thing is, we were planning to have separate command-line tools *per each separate task*, which is: one tool for building game28.dat from Game.agf, one tool to compile script, a tool to generate spritefile from source image files, a tool for packaging, a tool for compiling translation, a tool for packaging assets into `*.ags`, and so forth, all of these as *separate standalone executables* that could be run one by one (and some of those already exist, as you can see in Tools folder).
We wanted something that could be run from the Editor, as a replacement for the code that is currently inside the Editor, but also could be run by a user, or custom script written by a user, perhaps with some custom building rules.

Some of the code that i see in this PR should rather be part of the common library: for example, the Write* game data functions are better be paired with the Read functions (placed together), for consistency, since we have to keep them in sync when expanding game data. Some things clearly duplicate the existing code from Common library, such as constant declarations. This is rather a question of the code organization, which may be fixed over time, so not a critical issue.

But I can also see the code which presence confuses me very much, as I did not expect to see that in this building tool. For a few examples: there's a Debugger (for communicating with the engine), some kind of AutoComplete structure (?), 2.72 game importer, speech lines collector, a sort of the "new game" generator, and perhaps more... It looks like a part of a bigger application got copied over. I suppose this could make sense if we were rewriting whole IDE from scratch, but this is too much for the task of compiling game28.dta alone.

At the moment, I am in doubts what to suggest. I think we may begin discussing what's your goal, and compare them to our goals. As we had certain plans set for standalone tools. We may review these plans, or suggest which part your work could cover, but only after filtering out strictly necessary bits.

mgerhardy:
yes, i copied most of the code from my imgui editor branch - might still need to clean it up - also have some fixes left. idk if anyone is interested in this, but my goal was to mimic what the ags editor /compile does - but on the command line and portable - so i can use it on linux. if this is nothing that would be accepted - that's totally fine, i am just trying to compile games on linux and in github actions on linux images.

ivan-mogilko:
I understand it so that it may be easier for you to put everything in one program, that makes sense.
For our use case we'd rather have separate tools for separate tasks, as I mentioned above. Another reason which I forgot to note is that it's easier to maintain and share between 3.x and 4.x branches when they are separate, as we may replace one tool in the middle of a pipeline while keeping the others.

IIRC our idea was to be able to configure the game building pipeline as a script (such as batch, sh, python, or other), where user could customize certain steps according to their needs.

There are certain things that we dont have implemented as standalone tools at the moment, and game28.dta compilation is one of the core parts which is missing. Such tool's job would be to parse Game.agf into the structs in memory, convert those to Common data structs and write these into game28.dta. If you could pick out just the code responsible for this single job and make a PR, then we might accept that.

OTOH I recall that @ericoporto also had such tool drafted somewhere, but I dont know what is its current state. Perhaps he could also post his opinion here.

EDIT: But I also suppose that this `agsbuild` tool that you've written, may also be useful to people. I remember already proposing you to make a separate repository with your IDE tool; i may propose that same for this tool as well, as you may find users for it.

mgerhardy:
I will try to split them up.

ericoporto:
I can comment in more details later when I am on the computer, but for the editor in particular and GitHub actions, you can just run the editor on windows (which is what we use), but if you want to run on Linux for whatever reason, you can also run on Wine. I have a GitHub actions variant that just runs on wine, I need to find it. 

For the tools, once they exist individually, my idea is to use my generator (https://github.com/adventuregamestudio/ags/pull/2770) to orchestrate them. Then just running either ninja or make would build the game.

mgerhardy:
agdta split https://github.com/adventuregamestudio/ags/pull/3012 - this branch also contains a agdta tool now - let me know which approach you like more and i will go into that direction

ivan-mogilko:
> agdta split #3012 - this branch also contains a agdta tool now - let me know which approach you like more and i will go into that direction

I will check this out soon, thank you.

In regards to the building process, here's the idea that we had. There's a tool called [`agfexport`](https://github.com/adventuregamestudio/ags/tree/master/Tools/agfexport), which purpose is to parse Game.agf and provide instructions to other tools. It exports things such as: list of game scripts, list of rooms, generated script header with object names, etc (these options may be expanded if necessary).

This tool is a start of the pipeline, whatever launches the build calls `agfexport` first, and then uses its outputs to run other tools (script compiler, asset packager, and other).

*What* launches this is still a matter of choice. And I *suppose* that, as a first solution, we could have something similar to your `agsbuild` either in form of a script, or in form of a C++ executable. Except, unlike existing agsbuild variant, such executable would not have everything in itself, but call other tools instead, either using their program api, or by spawning processes (the former may be easier to code, but the latter may serve as a test that these tools work properly as standalone utilities).

ivan-mogilko:
The variant of a building tool that delegates tasks to other tools was merged as #3057.

PR #3012, opened 2026-07-14
Tools: add native agdta game28.dta writer

mgerhardy:
Introduce a CLI that parses Game.agf and writes game28.dta from compiled script objects, with libgamebuild backing and an integration test.

project.cpp belongs in libtools (imo) next to agfreader; it’s in libgamebuild now because agfreader only covers script-oriented refs, not full DTA metadata. 

A possibe follow-up: extend agfreader or add LoadGameDataFromAgf() in libtools - what do you think?


ericoporto:
This is ignoring all discussion on https://github.com/adventuregamestudio/ags/issues/1188

ivan-mogilko:
Alright, I'll begin with noting down the actual functionality implemented in this PR, for the sake of information, because the PR's description is minimal.

This adds a tool named `agdta`, which purpose is to build game28.dta (main game asset) from Game.agf.
The tool's usage is:

    `agdta --open <Game.agf> [options]`

The `--open <Game.agf>` part is required, and accepts the absolute path to Game.agf.
The additional options supported are (as printed by `--help` command):

```
  --project-dir <dir>     Project root (default: directory containing Game.agf)
  --output <dir>          Build output root (default: <project>/Compiled)
  --scripts-dir <dir>     Directory with *.o files (default: <output>/_temp)
  --dta <path>            Output .dta path (default: <output>/_temp/game28.dta)
  --verbose               Print build log to stderr
  -h, --help              Show this help message
```


When run, the tool does following steps:

1. Opens and parses Game.agf, saves the parsed data into the GameData class. GameData is simply put a collection of all game-related data read from the project file, mimicing respective data classes from the Editor. It is further wrapped into the Project class, that, besides GameData itself, also has reference to the project directory (i guess it was meant also for working with the other project files).

2. The loaded game project data is then passed to the BuildSystem manager class, which mimics Editor's game compilation functionaly to some degree.
2.1. BuildSystem prepares the output directories, including the "output" directory, and also a "output/temp" directory. The tool expects certain files be present in the "output/temp" directory, and will also write game28.dta into "output/temp" by default as well.
2.2. Does pre-build checks, which is mostly check that a data was loaded, and that it contains at least a single script module.
2.3. Generates the list of compiled script infos, by scanning their directory (output/temp by default), looking for "*.o" files. This step is obligatory, the tool will fail if no compiled script o file was found.
2.4. Finally writes game28.dta, passing the previously generated GameData + list of script modules into the function `WriteGameDataFile` that does the job.

3. `WriteGameDataFile` (found in game_data_writer.cpp) is pretty much an equivalent of the Editor's DataFileWriter class, converted from C# code.


On file structure. The tool itself is in Tools/agdta directory. The main code that does the parsing and building is in Tools/libgamebuild directory.
The tool uses parts of the Common library and some parts of the Tool/data utilities. AGF helper functions are used to parse the xml nodes, however the parsing is mostly done outside of the existing system. To elaborate, our original idea was that there's a base parsing class that has to be extended by subtypes, each responsible for parsing certain type of the node. Existing `agdta` tool makes a different approach where it uses ValueParser class as a helper to parse standard value types (bool, int and string), while the field reading is done by the functions in Tools/libgamebuild/project.cpp. (For this reason ValueParser was changed to have its static methods public instead of protected, as it has to be used by an external code.)

The data writing is done by a collection of functions in `game_data_writer.cpp`.

The current implementation does not contain a ready MSVS project, only relevant CMake scripts.

---

Overall, I have to admit, that this is quite different from the code structure that we had planned earlier. We will have to either alter this variant eventually to make it compliant with our existing code, or alter our existing code to be more in line with this code, so that everything is consistent. I'm referring mostly to the Game.agf parsing here.
For the reference, this is roughly what we had in mind:
https://github.com/ericoporto/ags/blob/add-tool-agf2dta/Tools/data/agfreader.h
https://github.com/ericoporto/ags/blob/add-tool-agf2dta/Tools/data/agfreader.cpp
in simple words: have a separate "parser" class for parsing each object type from the xml.

Game data structs: we have started shaping these up in Tools/data/game_utils.h, and i'd rather have the full versions merged together somehow, but suppose that's mostly a matter of moving the code around.

The writing of the compiled data: i hoped to have it separated into functions within respective Common classes as well, so that writing is conveniently paired with reading together in the same place in code. This should make it easier to maintain and keep reading & writing in sync. Of course, this may be done as an additional step later.

There are things that probably should not be there, like the Project class, which is superfluous for the task (except the code that reads xml of course), and also the directory structure assumptions, which should rather be made by a build system on a "higher level" than this tool. (So further simplifications are required, in order to make the approach more "direct".)

Adding compiled scripts should be optional, as we have an alternate way of packaging them as separate assets instead of embedding into game28.dta, and this method will probably be preferred in the future.

ivan-mogilko:
Very quickly - on code style. This code uses std::strings mostly, while we tend to use our Common::String in our code. As your code still uses parts of our Common library, it must have helper functions to bridge between both, sometimes back and forth.

I'd prefer tools to use Common::String though, since it's how the rest of our tools code is written, and because we share the code with Common library. Common::String may still be converted to std::string if using certain std utilities (we may add conversion function or operator to the String class for easier use).
(The script compilers, both old and a new one in ags4, are so far the only tools that uses std::string, but they are mostly self-contained programs.)

Then, we have a naming guidelines here
https://github.com/adventuregamestudio/ags/wiki/AGS-Coding-Conventions-%28Cpp%29#naming-conventions
unfortunately, the naming style that you use in code is very different from ours. Specifically, I noticed that:
1. Public struct fields are all lowercase with underscores dividing words (e.g. `cursor_image`), while ours is CamelCase without underscores (e.g. `CursorImage`).
2. Private struct fields are lowercase with underscore suffix (e.g. `game_data_`), while ours have underscore prefix, begin by lowercase word, but next words start with a capital letter (e.g. `_gameData`);

ivan-mogilko:
@mgerhardy It appears that @ericoporto wants to finish his previous work on the similar tool: #3016. It's based on our older discussion in #1188, and is more in line with our existing codebase both in terms of code style and architecture, so I am more inclined to accept that variant.

I realize that you put time and effort into your code, I hope it won't go waste since you are using your own programs, if I understand correctly. But at the same time, if you are actually interested in contributing to our repository, I strongly suggest discussing things first (you can open a issue for that), so to keep our intents and plans aligned, and not duplicate work or cause any wasted time and effort.

On another hand, I figured you are more interested in having cross-platform cli build tools than we are (since we are mostly using the Editor). It is true that having these is overdue. Several years ago I started writing them, but stopped for some reason (probably carried away with other tasks), which makes the current situation partially my own fault. And I would not want these to get postponed for unknown duration again.
Therefore, if you have idea of which other build tools, or features in existing tools, might be necessary, then please mention that, by opening feature proposals.

As soon as I get some spare time (maybe on this weekend) I will check how close our existing tools collection is in regards to building the complete game from the AGS 3.x project files, and what else is required to fill functional gaps.

ivan-mogilko:
#3016 implementation merged instead.

Issue #2735, opened 2025-05-24
Unhandled event could pass ID as third parameter (or some alternative)


ericoporto:
**Describe the problem**
When an event is unhandled we get a call to [unhandled_event](https://adventuregamestudio.github.io/ags-manual/Globalfunctions_Event.html#unhandled_event), that tells us the type of the thing that was unhandled and the interaction used, but unfortunately no identification of what specifically was - which hotspot or object. So if you want to handle something through it, to somehow trigger through it, it may not be trivial 

**Suggested change**
A third parameter could be provided so the object could be retrieved through ID or a string with its name or some other form of identification.

See https://www.adventuregamestudio.co.uk/forums/advanced-technical-forum/get-hotspot-by-reference/msg636683547/#msg636683547

We may alternatively decide not do anything since it seems there is a workaround solution for this.

ericoporto:
Erh, in theory you could pass the object itself through Anything and let the user cast to whatever it is?

ivan-mogilko:
Hmm, NOW that there's a feature pr, we could do that, indeed.
If Anything is merged, we can do that whenever there's a need for a generic callback for multiple types of objects.

PR #3015, opened 2026-07-15
4.0: Anything - universal implicit parent for managed structs

### LINKED ISSUE #2992: Use Anything as ultimate managed class
**Suggested change**

> Recently, I've been thinking about adding at least 1 ultimate managed class called `Anything` (yes) with only 1 function `GetTypeName()`, that could be a base class for every non-static managed class in script, and even supporting being used as user class parent. 
> Because we now support dynamic pointer downcast in scripts, such base class could be used to at least store multiple types in one array.

_Originally posted by @ivan-mogilko in [#1298](https://github.com/adventuregamestudio/ags/issues/1298#issuecomment-4769955923)_

Hey, I decided to bring this into its own issue to move the discussion here. This looks like a different thing from https://github.com/adventuregamestudio/ags/issues/1228, as the idea to me feels "like" the `object` type in C# (but different, more on that later).

I will collect a few things I got from the discussion here.

- We want to declare this object in script, because it will be easier to edit it if necessary (say we want to add ToString or Equals someday or whatever).
- We have something in the String object in the agsdefn.sh for this, `internalstring` we may reuse this approach in someway in this `Anything` declaration
- I think this would also help avoiding `Anything` inheriting from itself.
- It would be nice to have it inherited automatically (implicit) without requiring adding extends in everything (we may need to forego this in an initial version if it is too much work...)
- Internally, TypeName can be implemented by calling existing GetType() for engine's and plugins's dynamic objects, and retrieving the object's typeinfo from RTTI for user script classes.

Below is one idea I had on implicit inheritance on the compiler

> Just thinking about this, supposedly somewhere in the compiler it parses the extends keyword and handles inheritance from whatever type it is being inherited.
> 
> Now somewhere else in the compiler it reads a managed type declaration and does... Something. So if it was a nice function (doubt, but let’s pretend the code is organized), we could then always also run the function that handles the extends keyword with the type “Anything” hardcoded there. Edit: I think the place is in the if clause that handles the SYM_EXTENDS, we would add an else clause to it and then it could be worked out there, if it doesn't extend anything, it would at least extend Anything if this is a managed struct.

I think a question I have is if `String` is an `Anything` or not (I guess yes, but a `string` isn't?). I understand here the idea is a bit less like C# object, because Anything would be Any-Thing, so any managed class. Something like C# object would probably be `any` keyword.

**Describe the problem**

I think this is a simple alternative to generics that also has it's own use cases.

ivan-mogilko:
Resolves #2992

This introduces `Anything`, a simple managed struct which can be used as a parent by any managed type, both built-in and declared by user:
```
managedbase struct Anything {
    import readonly attribute String TypeName;
};
```

The keyword `managedbase` must be supported by the script compiler in order for this feature take full effect. If it does, then ALL managed structs that do not have a direct parent class (via `extends` keyword) will automatically inherit the type declared as `managedbase`.

This provides 2 effects:
1. Minor: each managed struct will have TypeName attribute, allowing to check its type name.
2. Major: we can now have a managed pointer type that can refer to virtually any managed object, hence allowing a fully generic storage. Given we support `as` downcast keyword, one can store anything in a `Anything*` variable, and later cast it down to the actual type.

On implementation:
1. New script compiler supports `managedbase` keyword. The rules of using this keyword are:
    1.1. Must be the first fully declared managed struct in script. Preceding forward declarations of managed structs are allowed.
    1.2. There can be only one 'managedbase' type.
    1.3. Cannot inherit (extend) anything else itself, must not have a parent.
    1.4 `managedbase` automatically makes the struct `managed`, so you may omit `managed` keyword.
    1.5 It is not forbidden to directly inherit `managedbase` struct, this will have exact same effect as if it was automatically inherited.
    1.6. Cannot be inherited by non-managed struct.
2. Old script compiler: TODO, not done yet.
3. Compilers must declare "MANAGEDBASE" extension, indicating that they support this keyword. `SCRIPT_EXT_MANAGEDBASE` macro may be used to test if it's supported. For the case if it's not supported i added a redefinition of `managedbase` to `managed` into the script header.
4. Engine: TypeName attribute returns Type string for all the built-in, and plugin's classes (where it should be provided by the author). For user-declared structs it retrieves the actual name from RTTI, if it's available. If it's not available, then will return empty string (never returns null pointer).
5. AutoComplete: properly supports `managedbase` keyword. After all structs are gathered, finds if there's a "managedbase", and applies as a parent to all structs that don't have a parent.

### DIFF
--- a/Common/ac/gamesetupstructbase.cpp
+++ b/Common/ac/gamesetupstructbase.cpp
@@ -149,6 +149,7 @@ const char *GetScriptAPIName(ScriptAPIVersion v)
     case kScriptAPI_v400_16: return "4.0.0-alpha20";
     case kScriptAPI_v400_18: return "4.0.0-alpha22";
     case kScriptAPI_v400_24: return "4.0.0-alpha28";
+    case kScriptAPI_v400_30: return "4.0.0-alpha34";
     default: return "unknown";
     }
 }
--- a/Common/ac/gamestructdefines.h
+++ b/Common/ac/gamestructdefines.h
@@ -208,7 +208,8 @@ enum ScriptAPIVersion
     kScriptAPI_v400_16 = 4000016,
     kScriptAPI_v400_18 = 4000018,
     kScriptAPI_v400_24 = 4000024,
-    kScriptAPI_Current = kScriptAPI_v400_24
+    kScriptAPI_v400_30 = 4000030,
+    kScriptAPI_Current = kScriptAPI_v400_30
 };
 
 const char *GetScriptAPIName(ScriptAPIVersion v);
--- a/Compiler/script2/cc_symboltable.cpp
+++ b/Compiler/script2/cc_symboltable.cpp
@@ -49,6 +49,7 @@ AGS::SymbolTableEntry::~SymbolTableEntry()
 
 AGS::SymbolTableEntry &AGS::SymbolTableEntry::operator=(const SymbolTableEntry &orig)
 {
+    this->Sym = orig.Sym;
     this->Name = orig.Name;
     this->Declared = orig.Declared;
     this->Scope = orig.Scope;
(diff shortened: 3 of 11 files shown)


ericoporto:
This looks to pretty much adhere to what was discussed in the issue and the use of the redefinition of managed base as managed if the compiler doesn’t support is a good strategy to make it work in the legacy compiler (but still gives a way to do that later). Hope to test it later.

ivan-mogilko:
If nobody sees any problems with this addition, I would merge this, and maybe as a test of its usefulness add it to "unhandled_event" as suggested in #2735 ?

ericoporto:
My limited test making a function that takes an Anything and just passing whatever is under the mouse, it looks like it works and that the cast can successfully get back whatever it was. I would say let's merge.

ivan-mogilko:
Ah, I forgot to implement this in the old script compiler. Will have to look into that separately. Although, I believe that at this point everyone should be using new one.

PR #3016, opened 2026-07-15
Tools: add agf2dta

### LINKED ISSUE #1188: Tool: Build main game data from Game.agf

ericoporto:
```
agf2dta v0.1.0 - AGS game main data file build tool
Copyright (c) 2026 AGS Team and contributors
Usage: agf2dta <INPUT-GAME.AGF> <OUTPUT-DIR>
Creates a "game28.dta" file in the OUTPUT-DIR from data in INPUT-GAME.AGF.
Options:
  -h, --help             Show this help message
```

fix #1188 

**NOTE:** highly recommend for now to check [files changed](https://github.com/adventuregamestudio/ags/pull/3016/changes) instead of commits, since there is a lot of horsing around...

Leaving this as a draft. 

🚨**AI DISCLOSURE:** This PR is highly authored using Codex for code writing and Claude for reviewing, I am using it through JetBrains Clion which is... Not good... But I can more manually control their urges. Except the older commits I had previously and fully hand written, this is roughly 6 hours of AI work and 20 USD of credits, made today. I still had to use around the same amount of time later to clean up some things by hand...🚨

**NOTE:** In the end this had to use many more hours of work, I guess I am probably averaging 4-5 hours per day, in the seven days I spent revising/writing this PR. There was not only direct code generation, but also a lot of manual authoring too, and rethinking of things, discussions here in the comments and a lot of time I just spent thinking about things. 

Edit: my old commits are in https://github.com/ericoporto/ags/tree/add-tool-agf2dta-bkp, I have reworked the commits as something that just adds/changes the files sequentially simple.

**EDIT2:** I will probably just merge everything in a single commit once this is done...

~~I am still also writing things by hand too, so both human and machine mistakes are expected. My idea to check things is to add a CI step that uses the Editor to build something (auto test? AGS Quest?) and then builds with this tool too and somehow compare, excluding the script and dialogue stuff. I have absolutely no idea how to do that yet, and in my head I will probably have to build a small python script helper to make such binary comparison test.~~ Changed my mind on this, I will revisit this in a later PR.

### DIFF
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -270,6 +270,24 @@ void StrUtil::WriteString(const char *cstr, size_t len, Stream *out)
         out->Write(cstr, len);
 }
 
+void StrUtil::WriteFixedString(const String &s, size_t len, Stream *out)
+{
+    if (!out || len == 0)
+        return;
+
+    // truncate and pad using raw bytes
+    const String tmp = s.Left(len);
+    if (!tmp.IsEmpty())
+        out->Write(tmp.GetCStr(), tmp.GetLength());
+    for (size_t i = tmp.GetLength(); i < len; ++i)
+        out->WriteByte(0);
+}
+
+void StrUtil::WriteFixedString(const char *cstr, size_t len, Stream *out)
+{
+    WriteFixedString(String(cstr), len, out);
+}
+
 void StrUtil::ReadCStr(char *buf, Stream *in, size_t buf_limit)
 {
     if (buf_limit == 0)
--- a/Common/util/string_utils.cpp
+++ b/Common/util/string_utils.cpp
@@ -279,13 +279,12 @@ void StrUtil::WriteFixedString(const String &s, size_t len, Stream *out)
     const String tmp = s.Left(len);
     if (!tmp.IsEmpty())
         out->Write(tmp.GetCStr(), tmp.GetLength());
-    for (size_t i = tmp.GetLength(); i < len; ++i)
-        out->WriteByte(0);
+    out->WriteByteCount(0, len - tmp.GetLength());
 }
 
 void StrUtil::WriteFixedString(const char *cstr, size_t len, Stream *out)
 {
-    WriteFixedString(String(cstr), len, out);
+    WriteFixedString(String::Wrapper(cstr), len, out);
 }
 
 void StrUtil::ReadCStr(char *buf, Stream *in, size_t buf_limit)
--- a/Tools/data/agfreader.cpp
+++ b/Tools/data/agfreader.cpp
@@ -655,6 +655,7 @@ void GUIControl::ReadButtonData(DocElem elem, DataUtil::GUIButtonData& data)
 {
     data.ClickAction = ReadString(elem, "ClickAction");
     data.ClipImage = ReadBool(elem, "ClipImage");
+    data.WrapText = ReadBool(elem, "WrapText");
     data.Font = ReadInt(elem, "Font");
     data.Image = ReadInt(elem, "Image");
     data.MouseoverImage = ReadInt(elem, "MouseoverImage");
@@ -1272,7 +1273,7 @@ String Font::ReadScriptName(DocElem elem)
 
 void Font::ReadAllData(DocElem elem, DataUtil::FontData &data)
 {
-    data.Name = ReadScriptName(elem); // This is Font Name
+    data.Name = ReadString(elem, "Name"); // This is Font Name
     data.AutoOutlineThickness = ValueParser::ReadInt(elem, "AutoOutlineThickness");
     data.AutoOutlineStyle = ReadFontAutoOutlineStyle(ValueParser::ReadString(elem, "AutoOutlineStyle"));
     data.CharacterSpacing = ValueParser::ReadInt(elem, "CharacterSpacing");
--- a/Tools/data/agfreader.cpp
+++ b/Tools/data/agfreader.cpp
@@ -12,6 +12,8 @@
 //
 //=============================================================================
 #include "data/agfreader.h"
+#include <algorithm>
+#include <cstdio>
 #include <tinyxml2.h>
 #include "debug/out.h"
 #include "util/string_compat.h"
@@ -26,6 +28,138 @@ const int  LOWEST_SUPPORTED_FORMAT = 26; // AGS 3.5.0
 
 using namespace AGS::Common;
 using namespace tinyxml2;
+
+namespace
+{
+
+static const CstrArr<3> kGameColorDepthNames = {{
+    "Palette", "HighColor", "TrueColor"
+}};
+
+static const CstrArr<4> kCompressSpritesTypeNames = {{
+    "None", "RLE", "LZW", "Deflate"
+}};
+
+static const CstrArr<3> kAndroidBuildFormatNames = {{
+    "ApkEmbedded", "AabEmbedded", "Aab"
+}};
+
+static const CstrArr<3> kGuiAlphaStyleNames = {{
+    "Classic", "AdditiveOpacity", "MultiplyTranslucenceSrcBlend"
+}};
+
+static const CstrArr<2> kSpriteAlphaStyleNames = {{
+    "Classic", "Improved"
+}};
+
+static const CstrArr<3> kRenderAtScreenResNames = {{
+    "UserDefined", "True", "False"
+}};
+
+static const CstrArr<5> kRoomTransitionNames = {{
+    "FadeOutAndIn", "Instant", "Dissolve", "BlackBoxOut", "CrossFade"
+}};
+
+static const CstrArr<3> kNumberDialogOptionsNames = {{
+    "None", "KeyShortcutsOnly", "Normal"
+}};
+
+static const CstrArr<7> kSkipSpeechNames = {{
+    "MouseOrKeyboardOrTimer", "KeyboardOnly", "TimerOnly",
+    "MouseOrKeyboard", "MouseOrTimer", "KeyboardOnlyStrict", "MouseOnlyStrict"
+}};
+
+static const CstrArr<4> kSpeechPortraitSideNames = {{
+    "Left", "Right", "Alternate", "BasedOnCharacterPosition"
+}};
+
+static const CstrArr<4> kSpeechStyleNames = {{
+    "Lucasarts", "SierraTransparent", "SierraWithBackground", "WholeScreen"
+}};
+
+static const CstrArr<3> kFontHeightDefinitionNames = {{
+    "NominalHeight", "PixelHeight", "CustomValue"
+}};
+
+static const CstrArr<2> kFontMetricsFixupNames = {{
+    "None", "SetAscenderToHeight"
+}};
+
+static const CstrArr<4> kInterfaceDisabledNames = {{
+    "GreyOut", "GoBlack", "Unchanged", "TurnOff"
+}};
+
+static const CstrArr<4> kGuiPopupStyleNames = {{
+    "Normal", "MouseYPos", "PopupModal", "Persistent"
+}};
+
+static const CstrArr<3> kInventoryHotspotMarkerNames = {{
+    "None", "Crosshair", "Sprite"
+}};
+
+static const CstrArr<3> kCustomPropertyTypeNames = {{
+    "Boolean", "Number", "Text"
+}};
+
+static const CstrArr<9> kFrameAlignmentNames = {{
+    "TopLeft", "TopCenter", "TopRight",
+    "MiddleLeft", "MiddleCenter", "MiddleRight",
+    "BottomLeft", "BottomCenter", "BottomRight"
+}};
+
+static const CstrArr<3> kFontOutlineStyleNames = {{
+    "None", "Automatic", "UseOutlineFont"
+}};
+
+static const CstrArr<2> kFontAutoOutlineStyleNames = {{
+    "Squared", "Rounded"
+}};
+
+static const CstrArr<3> kSpriteImportResolutionNames = {{
+    "Real", "LowRes", "HighRes"
+}};
+
+static const CstrArr<2> kAudioBundlingTypeNames = {{
+    "InGameEXE", "InSeparateVOX"
+}};
+
+static const CstrArr<6> kAudioFileTypeNames = {{
+    "OGG", "MP3", "WAV", "VOC", "MIDI", "MOD"
+}};
+
+static const CstrArr<5> kCrossfadeSpeedNames = {{
+    "No", "SlowFade", "SlowishFade", "MediumFade", "FastFade"
+}};
+
+static const CstrArr<3> kButtonColorStyleNames = {{
+    "Default", "Dynamic", "DynamicFlat"
+}};
+
+static const CstrArr<3> kInheritableBoolNames = {{
+    "Inherit", "False", "True"
+}};
+
+static const CstrArr<6> kAudioPriorityNames = {{
+    "Inherit", "VeryLow", "Low", "Normal", "High", "VeryHigh"
+}};
+
+static const CstrArr<3> kLipSyncTypeNames = {{
+    "None", "Voice", "Text"
+}};
+
+static const CstrArr<2> kPaletteTypeNames = {{
+    "Gamewide", "Background"
+}};
+
+static const CstrArr<4> kTurnOrderPriorityNames = {{
+    "Clockwise", "CounterClockwise", "Random", "FaceDown"
+}};
+
+static const CstrArr<3> kTextBoxKeyClaimStyleNames = {{
+    "All", "Handled", "TextOnly"
+}};
+
+} // namespace
 
 namespace AGS
 {
@@ -52,13 +186,16 @@ HError AGFReader::Open(const char *filename)
         return new Error("Not a valid AGS game project");
 
     const char *attr_filever = _doc->RootElement()->Attribute(XML_ATTRIBUTE_VERSION);
+    if (!attr_filever)
+        return new Error("Game.agf format version is missing");
     if (strcmp(attr_filever, LATEST_XML_VERSION))
         return new Error(String::FromFormat("Unsupported Game.agf format: %s", attr_filever));
 
     const int attr_format = _doc->RootElement()->IntAttribute(XML_ATTRIBUTE_VERSION_INDEX);
     const char *attr_editorver = _doc->RootElement()->Attribute(XML_ATTRIBUTE_EDITOR_VERSION);
+    _editorVersion = attr_editorver ? attr_editorver : "";
     Debug::Printf("AGFReader: opened %s,\n format tag: %s\n format index: %d\n saved by AGS %s",
-        filename, attr_filever, attr_format, attr_editorver);
+        filename, attr_filever, attr_format, attr_editorver ? attr_editorver : "unknown");
 
     if (attr_format < LOWEST_SUPPORTED_FORMAT)
         return new Error(String::FromFormat("Unsupported Game.agf format index: %d", attr_format));
@@ -74,6 +211,7 @@ void AGFReader::Close()
 {
     _doc.reset();
     _gameRoot = nullptr;
+    _editorVersion.Empty();
 }
 
 //-----------------------------------------------------------------------------
@@ -84,10 +222,14 @@ void EntityListParser::GetAllElems(DocElem game_root, std::vector<DocElem> &elem
     const char *folder_elem, const char *list_elem, const char *type_elem)
 {
     DocElem list_root = game_root->FirstChildElement(list_elem);
+    if (!list_root)
+        return;
     DocElem node = list_root;
     if (folder_elem)
     { // Get the main folder
         node = node->FirstChildElement(folder_elem);
+        if (!node)
+            return;
     }
     return GetElemsRecursive(node, elems, folder_elem, list_elem, type_elem);
 }
@@ -96,10 +238,14 @@ void EntityListParser::GetAllElems(DocElem game_root, std::vector<DocElem> &elem
     const char *root_elem, const char *folder_elem, const char *list_elem, const char *type_elem)
 {
     DocElem list_root = game_root->FirstChildElement(root_elem);
+    if (!list_root)
+        return;
     DocElem node = list_root;
     if (folder_elem)
     { // Get the main folder
         node = node->FirstChildElement(folder_elem);
+        if (!node)
+            return;
     }
     return GetElemsRecursive(node, elems, folder_elem, list_elem, type_elem);
 }
@@ -107,6 +253,8 @@ void EntityListParser::GetAllElems(DocElem game_root, std::vector<DocElem> &elem
 void EntityListParser::GetElemsRecursive(DocElem folder,  std::vector<DocElem> &elems,
      const char *folder_elem, const char *list_elem, const char *type_elem)
 {
+    if (!folder)
+        return;
     // First pass subfolders
     DocElem list_node = folder;
     if (folder_elem)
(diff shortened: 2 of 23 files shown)


ericoporto:
Hey, I think I am done with the AI part of this PR and will need some time to clean things up by hand and review the code before putting this as “Ready to Review”, but I think I will only have time for this during the weekend.

The data_file_writer.cpp/h I put in the agf2dta directory instead of the data directory because I assumed this tool will be the only user of such code, but if there is a chance we need something from it (some other clib writing?) then I can move it to the data directory.

edit: I ended up moving it to the data directory

ivan-mogilko:
I did not look too close yet, only scanned the whole thing, but something i noticed: few text writing functions in data_file_writer.cpp may be replaced with existing functions from StrUtils:
* WriteNullTerminatedString = StrUtils::WriteCStr
* WriteCountedText = StrUtils::WriteString (with length parameter)
* WriteFixedText = no equivalent, but may be added into StrUtils

ericoporto:
Thanks, I will use those. Anything that you think can reduce the codebase size is interesting too. 

DataFileWriter.cs is like 2k lines but this PR is adding 3k lines, even though it has to also parse Game.agf, I wonder if there is too much stuff.

In parallel I am building a small python tool that lets me compare game28.dta files, accounting for possible malformations, just to help me iterate a bit quicker comparing the output from the editor and from this tool.

One thing I was thinking about is to add some marking in the produced game28.dta so that if someone builds a game with it while it is experimental and opens a bug report in the Engine we can trace back to it? Obviously someone can fork and remove such marking.

ivan-mogilko:
> One thing I was thinking about is to add some marking in the produced game28.dta so that if someone builds a game with it while it is experimental and opens a bug report in the Engine we can trace back to it?

game28.dta contains "CompiledWith" field which has full editor version in it. You have it present in your code:
https://github.com/adventuregamestudio/ags/pull/3016/changes/c0bf02f20639744aa5b0682733f49c50fbf457b8#diff-7ad5e09bd77c0ac4df7aff392557a3ab099ed75d26e83b362a243f236c78d3feR1114

The problem is how to distinguish separate tools (Editor, cli tools, etc).

EDIT: another option: reserve a data extension for this.

Tools/data/agfreader.cpp
```diff
@@ -833,6 +834,22 @@ void ReadEntityRef(DataUtil::EntityRef &ent, EntityParser &parser, DocElem elem)
...
 
+static String BuildCursorScriptID(const String &name)
+{
+    String cursor_name;
+    // This is from MouseCursor.cs but I think this for part can be skipped because
+    // Game.agf would be in theory authored correctly?
+    for (size_t c = 0; c < name.GetLength(); ++c)
+    {
+        const unsigned char ch = static_cast<unsigned char>(name[c]);
+        if (std::isalnum(ch) || name[c] == '_')
```
ericoporto: This check also exists in [parser common header](https://github.com/adventuregamestudio/ags/blob/09d30a987fb802bafa4881b46484345358fa675e/Compiler/script/cs_parser_common.h#L127) but I think I would need to refactor to somewhere else if I was to put it here

ivan-mogilko: This could go in strutils by default, but there's also data/data_helpers.h, which currently contains only linebreak symbol conversion. The comment sais that this file is for "Helper functions related to reading or writing game data."

On another topic, I noticed a comment just above the code saying
> // This is from MouseCursor.cs but I think this for part can be skipped because
    // Game.agf would be in theory authored correctly?

IMO we should assume that Game.agf could be edited by hand, or a third-party tool too, so may contain mistakes.

ericoporto: Right, but Tools/data is outside the scope of the Compiler. I guess I would add to StrUtils then, with a comment explaining its uses.

About the comment I will remove it then, and put a note on top of agfreader to note that someone may have edited Game.agf by hand, just in case later on we need to make some decision we remember this.

ivan-mogilko: Actually, I misunderstood the purpose of this operation. This is not fixing the bad data, this is converting one (possibly valid) data to another form. The Editor allows cursor names to contain any characters, including spaces, so it has to be converted to a script-compatible name.

My thinking is that, in general, the build tool should not be fixing wrong data, but write it as is, so long as it fits the data format (it may post warnings though). But in the above case it just had to do this fix, because the source property contains a different kind of string.

ericoporto: It’s a bit more confusing when comparing with AGS.Types because I kept EntityRef as the parent of things (in agfreader) but the naming of some things is not consistent. Like in MouseCursor ScriptID is what would be ScriptName (Name is still Name so that is good).

Additionally, ScriptID is NOT serialized (it has `[AGSNoSerializa]`, it is not in Game.agf, so we need to figure it out. It’s a property used by DataFileWriter.

> The Editor allows cursor names to contain any characters, including spaces

I am not seeing this property in MouseCursor.cs, you sure?

I need to also pull all the DataFileWriter comments in because there’s quite a few notes there, specially in the extensions and legacy stuff we are not really using.

ivan-mogilko: If we suggest that the ScriptID is a name *for script*, then for the mouse cursor this field should not be read directly, but *read and converted*, similar way it's converted in this code snippet.

You can see exactly the same is done in ScriptID in MouseCursor class in the AGS.Types:
https://github.com/adventuregamestudio/ags/blob/09d30a987fb802bafa4881b46484345358fa675e/Editor/AGS.Types/MouseCursor.cs#L152-L170

In summary, MouseCursor has a Name field which is serialized in the project, but that's an arbitrary name. ScriptID is a derived value, which takes Name and converts to script-compatible form.


~EDIT: If you dont like this called ScriptID in EntityRef, then we may rename it to ScriptName.~
EDIT2: oh, it's already called ScriptName there. I got confused.

ivan-mogilko:
About tests: i think there's a need to test at least 2 major operations separately:
- parsing of agf into data structures
- writing game28.dta data.

IMO it's possible to achieve these without any external *binary* files (because these are always inconvenient to update, and not good to have in the git repository).

1. Have a Game.agf either in repository, or xml hardcoded in code. Opened as a stream (file or memory) and passed into the operation that must read and return full game data structure. Compare that the values of game data structure match the values from xml.

2. Have "design-time" game data structures, either filled in by hand, or read from Game.agf like in above test. Create a memory buffer, open a memory stream and write game28.dat into that. Then read the same buffer back into the "runtime" data structures (ones from Common). Compare that the values in "runtime" structures match the fields from "design-time" structs.

ericoporto:
So I was thinking in a much simpler approach, which was to download something we already have (templates, AGSQuest or the autotest game), and then build it with the Editor, which would upgrade the Game.agf to and generate the exactly expected binary.

Then I can use agspak to extract game28.dta from the .ags file built by the AGS Editor.

Now to compare these binaries the trick is that the one built by the editor has the game scripts and game dialog script compiled binaries also packed along, so my plan would be to create a small python script to automate the comparison excluding these specific regions.

Then doing so would make the build fail with an error in this step when something changed which would work as a reminder to update the agf2dta tool.

If we add a Game.agf in the repository we will have to keep updating it afaict, because the agfreader would always build the exact version and not update/upgrade things like the editor.

But anyway, we can add, just need to keep this in mind I think.

--- 

The other thing, is data_file_writer.cpp was made with the idea it could have pieces picked up and moved to respective classes in their original objects to pair the file reading functions at some point, but I am rethinking if I should just readjust it to match DataFileWriter code more closely to be easier to compare it one with the other side by side.

ivan-mogilko:
The reasons why I suggested the kind of test above:
- Being able to test separate operations (not one big process at once).
- We dont depend on the state of the files from other repositories, which are not meant for the tests. If we need to test other variants of data, we just edit our test input locally.
- No need to run other software, like AGS Editor, much faster test run.
- And we dont depend on other software to compare the results, which *also* may have mistakes; so such test will be more contained and more reliable.

The whole test may be contained within Tests solution, where we already have tests for tools, if i remember right.

BTW, keep in mind that eventually AGS Editor will use this tool too, instead of its own code, in which case comparing its output for testing this tool will become pointless.

ivan-mogilko:
> If we add a Game.agf in the repository we will have to keep updating it afaict, because the agfreader would always build the exact version and not update/upgrade things like the editor.

Yes, that's true though.

Also, there has to be a way to detect when the tool was not updated in-line with the Editor, so long as Editor has its own code to compile the game.

ericoporto:
I think I need to move data_file_writer into libtools, so into the data directory for easier testing, so I will do that.

Also, I am still on the fence on rewriting all data_file_writer.cpp into something that looks more like DataFileWriter.cs when viewed side by side, right now things are picked into functions that were meant to be refactored at some point to things from main game struct and their objects but I think this will make annoying to update it in the meantime.

ivan-mogilko:
> Also, I am still on the fence on rewriting all data_file_writer.cpp into something that looks more like DataFileWriter.cs when viewed side by side, right now things are picked into functions that were meant to be refactored at some point to things from main game struct and their objects but I think this will make annoying to update it in the meantime.

It is going to be somewhat annoying either way, but being the person who regularly updates the data format, I think it's better to have these C++ "writing" function mimic C++ "reading" functions. Then one can copy the code between them and only have to change write to read or vice versa. Right now it's quite inconvenient to have DataFileWriter in one form and "read" functions in Common in another, I wish it was not done like that from the start.

ericoporto:
I haven't figured a good approach for the agfreader but I added a tiny test just to have the file in place. For the Game.agf in the repository, I noticed it can't be only it, it would have to include a few minimal game files, otherwise it won't be easy to just load it in the Editor for upgrading. Made one from the empty game project [minimal_game.zip](https://github.com/user-attachments/files/30161200/minimal_game.zip), but it will add 116kB on the repository, was unsure about it.

For the data file writer, I managed to reuse some of the reading that was already in the engine. Using all the readers is a bit non-trivial because they are in files that pull a lot of dependencies... Picked the ones I thought would not pull too much stuff. The tests themselves were AI authored, so the things they fill in the pieces of the structs are a bit nonsense, but they do correctly put and read back things (I checked).

ivan-mogilko:
> For the Game.agf in the repository, I noticed it can't be only it, it would have to include a few minimal game files, otherwise it won't be easy to just load it in the Editor for upgrading.

Which files are required, and why is it not easy without them? I thought at this point Editor should recreate missing files itself. If it does not, then we could update the Editor just for this task. I'd rather have only Game.agf in a git repository, and no binary files, at least not for this particular test.

For upgrading Game.agf game compilation is not necessary, it only needs to load and save the project.

EDIT: Of course, for testing the full game build, we'd need to have all game project eventually. In which case we might have to use the "demo game" or another available game repository, as you suggested earlier.

ivan-mogilko:
It's been a while since I coded the agf reader classes, and checking them now makes me wonder why I made them like that, where you have to create an object, but then pass xml element to each of their methods. I cannot remember my reasoning now. Perhaps it would make sense to refactor them and construct a Value/EntityParser class with DocElem as a parameter, which is then saved in a parser object, and used whenever their "read" methods are called. This might also make certain code less, by using constructs like:

```AGF::Characters(elem).GetAll(list_of_chars)```.

Another thing that may be introduced is "ForEach" kind of method to be used instead of GetAll where makes sense. Such method would accept a function pointer, supposed to be run for each element, which would let pass lambdas too.

Idk if you'd like to experiment with this here, or leave this for the later times after this PR is merged. Perhaps it's better to make this tool work first.

**Comment by ericoporto** [comment]
The tool is working I think, I am more working on having tests on the binary writing side right now. The agf reading I am less worried because that one is easier to inspect.

I think my plan is to merge the tool working with SOME tests.

Then my next step would probably be in moving the Writes to whoever own the Reading in the common part, which then would move that pair read/write test into the common tests - which would then simplify the tool tests.

Only then I would like to look into the agf reading stuff, both what you mentioned, but also having proper tests in that side.

In parallel I would like to work on my build generator, but hopefully more slowly on my handmade code, because I think I would be able after this to be able to build a game - even though there are some small tools missing like plugin stuff.

Ah, back to this PR, except by the first string_utils related commit, I’ve been thinking once I am done I should squash everything in a single commit here.

**Comment by ivan-mogilko** [comment]
> In parallel I would like to work on my build generator, but hopefully more slowly on my handmade code, because I think I would be able after this to be able to build a game - even though there are some small tools missing like plugin stuff.

I've been thinking to maybe make a simple tool, even if as a temporary solution, that would implement the basic pipeline, calling other tools one by one, and building full game data from the project.

**Comment by ericoporto** [comment]
In my other tool the way it works if you look into its code is it describes files that are input, output file and command, then make or ninja when run will look into the generated file and figure out the build graph, so if you change something it figures the branch/leafs of the graph that need update and rerun that. The big issue there is that some common command like copy and move, differ considerable between bat (cmd.exe) and bash, so for these it would be nice to have some tool (some name like agsbb for busybox) that had a few of these commands, in very simple form. I put a lot of comments on that tool everytime I had a doubt.

**Comment by ivan-mogilko** [comment]
I started writing this basic dirty program that just spawns processes one by one, because I needed a way to test how existing tools work: https://github.com/ivan-mogilko/ags-refactoring/commits/experiment--agsbuild/
(this branch has merged agf2dta and recent additions to agfexport).

(Maybe I should have written a batch file, or a python script instead...)

There's already a benefit that I've found a number of mistakes in agfexport and agscc, will push them to the master branch separately.

**Comment by ivan-mogilko** [comment]
Hmm, could be this mistake only appears after your changes to agfreader in this PR, but the generated CursorModes enum gets `eMode` part duplicated, so it needs to be modified like:

```
- header.Append(DeclareEntitiesAsEnum(game.Cursors, "CursorMode", "eMode"));
+ header.Append(DeclareEntitiesAsEnum(game.Cursors, "CursorMode"));
```

This might happen if Cursor's script name now includes "eMode" itself.

EDIT: forgot to check, but maybe similar change is required for AudioType and Fonts enums.

**Comment by ericoporto** [comment]
I may have broke the agfexport, I think I tried to skirt around the Ref entities by adding some Data version to avoid changing but may have broken something. :/

> This might happen if Cursor's script name now includes "eMode" itself.

It is the ScriptID and Name difference, need to check on that.

I am working on figuring two specific tests:

- The GUI write stuff, can’t figure out a good way because they have too much runtime stuff, I am almost moving the write stuff into the gui classes and then have those tests happen in the common_test. The only approach I got so far is adding stubs in the tools_test, but I thought they looked ugly. Other idea I had was to have the gui classes extend a gui…_data version of those classes that would only have the properties and the read stuff.
- a big test that writes the entire thing so I can be sure the order of the blocks is correct. Again, the hard thing is figuring out what the minimum amount of hack to bring the function that reads everything from main game file without having too much stuff and stubs.

**Comment by ericoporto** [comment]
ok, I kinda want to squash all commits, except the two that are Common related, and be done with this PR for now. (I am still looking into the Cursor, AudioType and Fonts though)

The full round trip test of the write and read will bring A LOT of the Common to the Tools test, and the alternative that makes sense to me is moving the Write stuff to its Common counterpart stuff and testing in common, but then the agf2dta itself will bring a lot of the Common, because then it needs to link to all things there. This second approach looks more sensible if some refactoring is done to separate the data from the runtime stuff, to reduce a little of the common dependencies. I can't make my mind here between the two approaches.

On the testing the agfreader side, I think I will need to refactor it to the way you mentioned in your previous comment before making the tests otherwise I will need to fix all tests later, beyond also having to change all the reading stuff...

I think the tool is working, but I would like to move to working on my other PR to remind of all the other stuff that still has to be done... And just so I have a bit of time to think about the testing of this after moving my mind to other thing.

**Comment by ivan-mogilko** [comment]
Yes, GUI is annoying because they have all the drawing and calculation.

One quick solution that I might suggest to try, is this: pick out their property fields into separate structs like "ButtonData" etc; and then either inherit each gui class from that struct with `protected` modifier (C++ supports multiple inheritance, so that should be possible to inherit e.g. GuiButton from both GuiObject and ButtonData), or place such struct as their private member (such as `ButtonData _data;`). Both ways may require to modify how these classes access the data, and lots of code changes may be inevitable (with the second way each `_textColor` will become `_data.TextColor` or similar), but it should not be *too bad*, i think.
Reading from file may be done either by delegating from GUI class's member function to the underlying GUIData class's function, or even by reading into GUIData and then initializing new GUI object passing Data struct into copy or move constructor.

Then you could uses these data structs in the tools.

EDIT: oh yes, you've mentioned this solution in your previous comment. I believe. we've already discussed it a while back.

**Comment by ericoporto** [comment]
The reason why I want to delay that approach is because, afaict, this would require new files in common for each of the data versions of the GUIs, which would then require in all the project files. I can do it, it's just that I need to do it with care and calm to properly adjust everything.

> This might happen if Cursor's script name now includes "eMode" itself.
> 
> EDIT: forgot to check, but maybe similar change is required for AudioType and Fonts enums.

I think I fixed the duplicate stuff in the headers properly now.

I want to do the squashing of everything else before this last commit, except the two commits that change Common entries and put this PR in review mode. I am mentioning also in case you have changes built on top of it because I don't know how rebase works after I am done with squashing the commits...

**Comment by ivan-mogilko** [comment]
> afaict, this would require new files in common for each of the data versions of the GUIs, which would then require in all the project files

I think it could be better just 1 file with all gui data structs, since they are plain structs with fields there's not much need to split them into separate files.

**Comment by ivan-mogilko** [comment]
Hmm, but actually, we are currently working on 3.6.3 branch, and this version is in RC stage already, so it's not good to do big refactor there. So, imo, either do a temporary solution with gui stubs for the test, or we split 3.6.3 out into release-3.6.3 branch before this.

OTOH I remembered that we already have stubs for GUI in tests:
https://github.com/adventuregamestudio/ags/blob/master/Common/test/common_stubs.cpp
are these enough for tool tests?

**Comment by ericoporto** [comment]
I tried going for the GUI stubs there, but it still pulls dependency into a lot of other things of the common library, including even allegro, which then is better linking the entire common into the tools test - I haven't actually got it to build, I am in a loop of getting link errors, adding more files, getting link errors, adding more files...

Edit: got there in the CMake, don't know yet how to get there in the vcxproj files yet though.

```diff
diff --git a/Tools/CMakeLists.txt b/Tools/CMakeLists.txt
index acf3ebec3..67a787625 100644
--- a/Tools/CMakeLists.txt
+++ b/Tools/CMakeLists.txt
@@ -223,13 +223,36 @@ endif ()

 if(AGS_TESTS)
    add_executable(tools_test
+      ${CMAKE_SOURCE_DIR}/Common/test/common_stubs.cpp # GUI reader runtime stubs used by data_file_writer_test
       ../Common/ac/gamesetupstructbase.cpp # needed by data_file_writer_test
       ../Common/ac/view.cpp # needed by data_file_writer_test
       ../Common/ac/audiocliptype.cpp # needed by data_file_writer_test
       ../Common/ac/dynobj/scriptaudioclip.cpp # needed by data_file_writer_test
       ../Common/ac/dialogtopic.cpp # needed by data_file_writer_test
+      ../Common/ac/common.cpp # needed by GUI readers in data_file_writer_test
+      ../Common/ac/spritecache.cpp # needed by GUI readers in data_file_writer_test
+      ../Common/data/assetmanager.cpp # needed by Common font renderers
+      ../Common/font/fonts.cpp # needed by GUI readers in data_file_writer_test
+      ../Common/font/ttffontrenderer.cpp # needed by fonts.cpp
+      ../Common/font/wfnfont.cpp # needed by fonts.cpp
+      ../Common/font/wfnfontrenderer.cpp # needed by fonts.cpp
       ../Common/game/interactions.cpp # needed by data_file_writer_test
       ../Common/game/customproperties.cpp # needed by data_file_writer_test
+      ../Common/gfx/allegrobitmap.cpp # needed by GUI readers in data_file_writer_test
+      ../Common/gfx/bitmap.cpp # needed by allegrobitmap.cpp
+      ../Common/gui/guibutton.cpp # needed by data_file_writer_test
+      ../Common/gui/guiinv.cpp # needed by data_file_writer_test
+      ../Common/gui/guilabel.cpp # needed by data_file_writer_test
+      ../Common/gui/guilistbox.cpp # needed by data_file_writer_test
+      ../Common/gui/guimain.cpp # needed by data_file_writer_test
+      ../Common/gui/guiobject.cpp # needed by data_file_writer_test
+      ../Common/gui/guislider.cpp # needed by data_file_writer_test
+      ../Common/gui/guitextbasedcontrol.cpp # needed by data_file_writer_test
+      ../Common/gui/guitextbox.cpp # needed by data_file_writer_test
+      ../Common/util/geometry.cpp # needed by GUI readers in data_file_writer_test
+      ../Common/util/path_ex.cpp # needed by assetmanager.cpp
+      ../Common/util/version.cpp # needed by common_stubs.cpp
+      ../Common/util/wgt2allg.cpp # needed by Common bitmap/font implementations
       test/agfreader_test.cpp
       test/data_file_writer_test.cpp
       test/data_helpers_test.cpp
@@ -243,7 +266,10 @@ if(AGS_TESTS)
       INTERPROCEDURAL_OPTIMIZATION FALSE
    )
    target_link_libraries(tools_test
+      Allegro::Allegro
       libtools
+      AlFont::AlFont
+      AAStr::AAStr
       gtest_main
    )
```

I am tempted to either postpone these tests altogether or going the data refactor approach, but the data refactor I don't think it should be too problematic, other than the issue that it will touch in every project file in the repository, which I would prefer to do in a separate PR focused only so we can more carefully check it.

Edit: I guess I will do it as is, and leave the comments so once a refactor is done these can be more easily removed....

ericoporto:
ok, I would now keep the two Common related commits, and then squash every other commit except the last two. Is it ok? And then put the PR in review mode.

ivan-mogilko:
Oh right, it seems to be impossible to avoid spritecache used in these gui classes. Not sure about font renderers.

Anyhow, I suppose, since these are tests, it's okay to link all libraries.

ivan-mogilko:
> ok, I would now keep the two Common related commits, and then squash every other commit except the last two. Is it ok?

Looking at the file changes, i think that would be okay, since the changes are mostly limited to few files, and are mostly complely new code.

ericoporto:
I redid the commits and rebased on top of the current master, let's see how it builds. I have NOT added a proper test for the Game.agf in agfreader, I would like to do it separately so I can have a bit more time to think about it.

There is one idiotic thing in here in the agfreader that the GameRef reader uses the reading of the entire GameData, it is working, anyhow, but it is an idiotic approach, but the untangling is a bit ugly which is essentially dupplicating the entire read all data function that currently exists or like, having a bool parameter that either Reads All data or just the entity refs, and some overload of the function to that uses the correct types (like the function always use the entire data, but we just don't fill the necessary things in the GameRef only path and later we cast it to GameRef.

Edit: decided to do this one final change before review. I had to duplicate it slightly because there were a few more issues I had forgotten that the two different paths approach didn’t properly work.

Common/util/string_utils.cpp
```diff
@@ -270,6 +270,24 @@ void StrUtil::WriteString(const char *cstr, size_t len, Stream *out)
...
+    const String tmp = s.Left(len);
+    if (!tmp.IsEmpty())
+        out->Write(tmp.GetCStr(), tmp.GetLength());
+    for (size_t i = tmp.GetLength(); i < len; ++i)
+        out->WriteByte(0);
+}
+
+void StrUtil::WriteFixedString(const char *cstr, size_t len, Stream *out)
+{
+    WriteFixedString(String(cstr), len, out);
```
ivan-mogilko: This should be String::Wrapper(cstr) to avoid string copying.

Common/util/string_utils.cpp
```diff
@@ -270,6 +270,24 @@ void StrUtil::WriteString(const char *cstr, size_t len, Stream *out)
...
+{
+    if (!out || len == 0)
+        return;
+
+    // truncate and pad using raw bytes
+    const String tmp = s.Left(len);
+    if (!tmp.IsEmpty())
+        out->Write(tmp.GetCStr(), tmp.GetLength());
+    for (size_t i = tmp.GetLength(); i < len; ++i)
+        out->WriteByte(0);
```
ivan-mogilko: There was a helper function Stream::WriteByteCount that could be used here instead of having a for loop.

Like `out->WriteByteCount(0, len - tmp.GetLength());`

Tools/data/agfreader.h
```diff
@@ -344,6 +362,27 @@ class GlobalVariables : public EntityListParser
...
+    static String ReadName(DocElem elem) { return ReadString(elem, "Name"); }
+    static String ReadDescription(DocElem elem) { return ReadString(elem, "Description"); }
+    static String ReadDefaultValue(DocElem elem) { return ReadString(elem, "DefaultValue"); }
+    static AGS::Common::PropertyType ReadType(DocElem elem);
+    static bool ReadAppliesToCharacters(DocElem elem) { return ReadBool(elem, "AppliesToCharacters", true); }
+    static bool ReadAppliesToHotspots(DocElem elem) { return ReadBool(elem, "AppliesToHotspots", true); }
+    static bool ReadAppliesToObjects(DocElem elem) { return ReadBool(elem, "AppliesToObjects", true); }
+    static bool ReadAppliesToInvItems(DocElem elem) { return ReadBool(elem, "AppliesToInvItems", true); }
+    static bool ReadAppliesToRooms(DocElem elem) { return ReadBool(elem, "AppliesToRooms", true); }
+    static bool ReadTranslated(DocElem elem) { return ReadBool(elem, "Translated"); }
```
ivan-mogilko: Two questions: why do these methods have to be static, and why are they needed here at all, and not a single ReadAllData reading into a data struct?

Tools/data/agfreader.h
```diff
@@ -352,6 +391,7 @@ class Game : public EntityParser
     String ReadScriptName(DocElem elem) override { return ""; }
 
     DocElem GetSettings(DocElem elem);
+    static int ReadPlayerCharacter(DocElem elem) { return ReadInt(elem, "PlayerCharacter", -1); }
```
ivan-mogilko: Same question here, why does this have to be a static method?

Tools/data/agfreader.h
```diff
@@ -360,9 +400,7 @@ class GameSettings : public EntityParser
     String ReadType(DocElem elem) override { return "GameSettings"; }
     int    ReadID(DocElem elem) override { return -1; }
     String ReadScriptName(DocElem elem) override { return ""; }
-
-    String ReadSayFunction(DocElem elem) { return ReadString(elem, "DialogScriptSayFunction"); }
-    String ReadNarrateFunction(DocElem elem) { return ReadString(elem, "DialogScriptNarrateFunction"); }
+    void ReadGameSettings(DocElem elem, DataUtil::GameSettings& s);
```
ivan-mogilko: Let's keep "read-all" functions having same name everywhere, like "ReadAllData".

ivan-mogilko:
I trust the automatic tests work, but also i've completed my [silly tool here](https://github.com/ivan-mogilko/ags-refactoring/tree/experiment--agsbuild), and it manages to produce a actually working game.ags, by calling all the existing tools in a chain :).

ericoporto:
lol, that strategy of the hard linked files in a dir is crazy, glad it works, I had a huge doubt on how to figure it out to author the pattern file to pass to agspak, and how to make agspak use a different internal structure (I think I suggested somewhere adding support for something like `; in/pak/path` to have a way to note the rewritten path inside the pak file).

About the Audio stuff, I guess we need to either rebuild the AudioCache somehow or just use it directly for now?

ivan-mogilko:
>  that strategy of the hard linked files in a dir is crazy, glad it works, I had a huge doubt on how to figure it out to author the pattern file to pass to agspak, and how to make agspak use a different internal structure

That is indeed a problem, which i realized too late. I believe it must be solved somehow in the future.

> About the Audio stuff, I guess we need to either rebuild the AudioCache somehow or just use it directly for now?

AudioCache contents can be packed directly into audio.vox. There's a setting that tells whether particular clips are packed into the game.ags instead, i am not sure how essential that is, but afaik even these may be packed into audio.vox and engine will still find them. If it's wanted to follow editor's instructions exactly, then we'd need to support this extra option later.

ericoporto:
I will create a proper issue to explain later, but the problem is AudioCache is not put in version control, and what it does is put all the sound files that could be anywhere in a single flat directory. I don’t remember if it also renames the sound files to some hash. So something would need to reproduce this if we want to use agspak in files from AudioCache instead of the original source ones.

Tools/data/agfreader.h
ericoporto: No idea, this is part of the AI generated stuff, I will look into it and figure out and get back

ivan-mogilko:
Spritefile may be not be put under source control too, so if we are talking about building a project from source control entirely, then we also need a complete sprite import reimplementation as a cli tool. The existing spritepak is a simpler tool, it does not use sprite definitions from Game.agf, but imports from simple files corresponding to 1:1 to sprite. For a full import support it would need:
- read sprite definitions from Game.agf,
- added GIF support at least (people normally use PNG for simple images, and sometimes GIF for animated sequences).

Common/util/string_utils.cpp
ericoporto: Done

Common/util/string_utils.cpp
ericoporto: Ah, correct. Done!

Tools/data/agfreader.h
ericoporto: I tried to do changes to ReadAllData and get rid of the weird static stuff the AI had generated, I think it looks alright now. Note that not everything is like this, some stuff is used by other agf related tools, and I just didn't want to touch things that were used by those when possible.

ericoporto:
First, added two commits that I think fixes the issues. If you need me to squash them too, let me know, although not sure where to squash the last one.

I guess we can do some version without the spritecache and later comeup with something for it because I think it will be a big tool. But we do have that new stb library in the repository now, so it should make things a bit easier I think, but the gif stuff is very nontrivial and I am not sure how well things for that are in stb_image. There is a version of stb_image maintained by the SDL folks that is usually more quickly patched for issues.

ivan-mogilko:
Reading through the comments in data_file_writer, which note differences from the editor, there's one thing remaining which I could not understand:

```
//  * AGF input also drops some project data that the Editor writer serializes:
//    GUI button WrapText;
```

ericoporto:
It turns out the note was actually something that was missing. Fixed it!

Edit: on my end, this work with AI code generation feels like doing right take a lot of work and I probably didn't even do it proper, reviewing the tests, they feel right but I found it much harder to follow than if I had just written the thing - they are so long! I really put an effort to follow on each decision it took and try to adjust to feel more like our own code, but it felt a lot slipped through the cracks. 

I am curious how this was in your side as reviewer, because the thing just sprouted a lot of code in such short time.

In the end I guess I am happy that this tool (agf2dta) at least came to fruition and this means we worked toward advancing the tools plan we originally set - even though getting it properly materialized was a bit bumpy.

Tools/data/data_file_writer.cpp
```diff
@@ -0,0 +1,1261 @@
...
+// Read by GUIObject::ReadFromFile_Ext363() in Common/gui/guiobject.cpp as the
+// common prefix of each concrete GUI control's extension reader.
+void WriteGuiControlLooks363(Stream *out, const DataUtil::GUIControlData &control)
+{
+    out->WriteInt32(control.BackgroundColor);
+    out->WriteInt32(control.BorderColor);
+    out->WriteInt32(control.BorderWidth);
+    out->WriteInt32(control.PaddingX);
+    out->WriteInt32(control.PaddingY);
+    for (int i = 0; i < 4; ++i) out->WriteInt32(0);
```
ivan-mogilko, line 976: Please add a comment that these trailing bytes are "reserved", here and in similar places below (like is done in read functions for these extensions).

Tools/data/data_file_writer.cpp
```diff
@@ -0,0 +1,1261 @@
...
+{
+    const auto buttons = CollectGuiControls<DataUtil::GUIButtonData>(game);
+    out->WriteInt32(static_cast<int32_t>(buttons.size()));
+    for (const auto &button : buttons)
+    {
+        WriteGuiControlLooks363(out, *button);
+        const bool dynamic = button->ColorStyle == DataUtil::kButtonColor_Dynamic ||
+            button->ColorStyle == DataUtil::kButtonColor_DynamicFlat;
+        out->WriteInt32((dynamic ? 1 : 0) |
+            (button->ColorStyle == DataUtil::kButtonColor_DynamicFlat ? 2 : 0));
```
ivan-mogilko, line 1095: Please replace the numbers with ButtonFlags constants here.

ivan-mogilko:
> I am curious how this was in your side as reviewer, because the thing just sprouted a lot of code in such short time.

I decided to not check the written format thoroughly and instead rely on automatic tests in that. Of course we will have to still check for mistakes or stylistic inconsistencies later.

ericoporto:
Hey, sorry it took me a little more than I expected to find the free time to go back to this

PR #3017, opened 2026-07-16
Editor: support showing context menu by Menu key (aka Application key)

ivan-mogilko:
Done by user's request. I don't know how often, but apparently there are situations where a user may want to open a context menu using a keyboard key known as "Menu"/"Application" key, rather than right-clicking a mouse. So i decided to add this.

I am not fully certain what behavior is expected in regards to menu position. Judging by how MSVS does this (as an example), it looks like if there's a item selected on current control, then the menu should appear at that selected item. Otherwise - it appears where the mouse cursor is. I tried to mimic this behavior in the Editor, where possible. If it's not wanted, then it's easy to change.

To make sharing behavior easier, added a ListViewExtended class, which handles RMB and Apps key, and sends "ContextMenuTrigger" event, which may in turn be handled by another code. This class is used in multiple lists across the Editor.
Other controls and editor panels handle these on their own.

Also used this as an opportunity to move several extended control classes from AGS.Editor to AGS.Controls library.

Which controls and panes are affected?

1. Many list (list view) controls: CallStack, Output, Find Results panels, Custom property schema, Global variables, etc.
2. Project Tree.
3. GUI Editor.
4. Sprite selector.
5. Palette Editor.
6. Room Editor.
7. View Editor.

This is where I found context menus. If I missed any, then they may be amended here, or later.


ericoporto:
That is cool, I have a notebook that has a touchpad that is essentially good for tapping but the buttons are just really bad. So tapping is used as left click. And this notebook it does have [this menu key button](https://en.wikipedia.org/wiki/Menu_key) - actually except MacBooks, I don’t think I have a notebook that doesn’t have this button. But I didn’t knew it wasn’t actually the right click button, so I thought it didn’t work with AGS.

What I tried was an AGS editor that had a three dots button that positions in the project tree rightmost, and clicking this three dots button brings the context menu. I thought it looked weird so I gave up on developing it - edit: found it, it was a single commit: https://github.com/ericoporto/ags/commit/08ac5a13b6fc6fc8d2cbc925a88b6ec8960b0e07

ivan-mogilko:
Supposedly, the Editor's code may be simplified if it used standard ContextMenuStrip, assigned as a ContextMenu property in controls. In which case .NET framework will take care of this behavior. But for some reason (unknown to me) AGS uses mostly its own system. Since 3.6.3 is close to release, I thought I won't try to alter that now. But maybe this can be redesigned in the future.

ericoporto:
First, I tested and it seems to be working. I noticed you had to manually figure out the positions to spawn it, but they seem to be sensible choices as in testing they feel correct.

About your comment, what I actually meant is I didn't knew it had to actually be programmed to work, I thought it was a windows feature that didn't work, lol. But now seeing the code there is no way it could know some stuff there that has to be manually passed.

Now, on the topic of _having to be programmed_, the [wikipedia article](https://en.wikipedia.org/wiki/Menu_key) has an entry like this

> Some Windows public terminals did not have a Menu key on their keyboard with the goal of preventing users from right-clicking; however, in many Windows applications, a similar functionality could be invoked with the ⇧ `Shift+F10` keyboard shortcut

Again, I thought the Shift+F10 was a Windows feature of auto-remapping, but since it doesn't work I suppose this is also some convention that some applications happens to program in. This would mean, everywhere a ` Keys.Apps` comparison is, it would need an or with an ( `Keys.F10` and `Shift` pressed) if we want to adhere to this "convention". In my PC this only seems to be supported by browsers and things like Notepad, but a lot of more complex applications seem to ignore this Shift+F10 combination, so not sure we need that, but just thought to mention.

ivan-mogilko:
Adding Shift + F10 combination seems to be simple enough.

PR #3018, opened 2026-07-17
Editor: fix Room Editor's navbar hiding drop menus when hot-tracking

ivan-mogilko:
This is an attempt to fix the navigation bar in the Room Editor to NOT automatically hide dropdown menus when user hovers the mouse cursor over the other items in the bar (without clicking). This issue was annoying users for a very long time.

I had an attempt couple of years back, but it was not successful, leading to other glitches, so I abandoned it:
https://github.com/adventuregamestudio/ags/issues/2415#issuecomment-2102486304

In a nutshell, it turned out that simply handling Closing event is not enough, as something is not working right in the version of .NET Framework we are using, and it's reporting the incorrect CloseReason all the time. So I had to add few hacks (getting into the .NET internals through reflection) to make it work properly. Details are in comments to the code.

This must be tested a bit, to see if the drop-down menu CLOSES when it should close.

Also, there's a chance that the above issue was fixed in the later versions of framework, in which case we just need to remove these hacks whenever we move to those.

### DIFF
--- a/Editor/AGS.Controls/AGS.Controls.csproj
+++ b/Editor/AGS.Controls/AGS.Controls.csproj
@@ -144,6 +144,9 @@
     <Compile Include="AssemblyInfo.cs">
       <SubType>Code</SubType>
     </Compile>
+    <Compile Include="Controls\ToolStripDropDownButtonEx.cs">
+      <SubType>Component</SubType>
+    </Compile>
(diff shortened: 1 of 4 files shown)


ivan-mogilko:
Alright, i will merge this and let people test more.

PR #3020, opened 2026-07-20
Common: interactions report deserialization errors

ericoporto:
Refactor out the quit call, fix silently ignored error and readjust factory functions to return HError.

When I was trying to make the tests for the data_file_writer.cpp (in #3016), I noticed this approach would make things easier to test, but since it impacts a few things I decided to move it to its own PR - and also because there was a FIX-ME comment in interactions.cpp that gave me this idea in the first place. If this is alright, then I will rebase my other PR on top of this to get on with the tests.

**AI Disclosure:** I used Codex in JetBrains CLion to automate this refactor, I did in part of the other branch and then went back to master branch to apply it.

### DIFF
--- a/Common/ac/gamesetupstruct.cpp
+++ b/Common/ac/gamesetupstruct.cpp
@@ -122,7 +122,7 @@ HGameFileError GameSetupStruct::read_cursors(Common::Stream *in)
     return HGameFileError::None();
 }
 
-void GameSetupStruct::read_interaction_scripts(Common::Stream *in, GameDataVersion data_ver)
+HError GameSetupStruct::read_interaction_scripts(Common::Stream *in, GameDataVersion data_ver)
 {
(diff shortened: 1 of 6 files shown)


ivan-mogilko:
Weird, I had a false memory that this was fixed long time ago.

About factory methods, may I suggest a opposite approach: the function would return an object (or unique_ptr with an object), while HError reference is one of the parameters?

ericoporto:
> About factory methods, may I suggest an opposite approach: the function would return an object (or unique_ptr with an object), while HError reference is one of the parameters?

That is easy to change, I will do this later today when I am back at the computer.

ericoporto:
@ivan-mogilko did the alterations, because some things didn't check/propagate errors before the PR, it is still a lot of code changes.

Issue #2530, opened 2024-09-10
Standard error/message print in command line tools


ericoporto:
When reading error messages from command line tools we may want to be able to understand their errors in ags, so it would be nice to have they print something like `Error: at line X in file 'FILE', ERROR_MESSAGE` so we could pick this up in the Editor using a regex.

Additionally, it may be nice to also print the tool name, like `agfexport:Error: msg`, which would make it easier to get from which tool came which error - imagine you are running many tools at once. 

Finally some standard error messages like `failed to open 'FILE' for reading` could be useful since they happen across all tools. 

I would suggest adding a small helper file to help make these uniform across tools. I wonder if setting the tool name could be made at compile time using either macro or constexpr, this would avoid having a singleton class. Edit: actually I think it may not be possible to set at compile time differently per tool.

ivan-mogilko:
> I wonder if setting the tool name could be made at compile time using either macro or constexpr, this would avoid having a singleton class. Edit: actually I think it may not be possible to set at compile time differently per tool.

It might be possible, for example, if you either pass it as a parameter to a error formatting function, or have it defined as a compilation variable. Although idk if any of that is necessary.

ericoporto:
If it was a singleton I would make something like this class here

https://github.com/ericoporto/agsworks/blob/main/src/misc/awlog.h
https://github.com/ericoporto/agsworks/blob/main/src/misc/awlog.cpp

But instead of where I init it passing the engine instance I would set the prefix. I don't know if this is a problem at all.

And I know having such prefix is not a common thing in general command line tools, but if you run a bunch of tools once one of them errors it becomes a bit hard to figure where the issue was - specially in "can't open file for writing" and any other types of error messages that could be identical across tools.

> It might be possible, for example, if you have it defined as a compilation variable

This is the only working approach I could figure it out.

Anyway, if a singleton isn't a problem at all I think I will go with it instead.

Other approach would be if it would be possible to reuse the log in the engine but I looked at the files of it in `Common/debug` and couldn't understand how I would do that.

ivan-mogilko:
I cannot remember why i did not do this from the start, but i think that it's better to use the log system from the Common in tools, not only because it already exists and covers certain situations (like logging from multiple threads), but also because the tools use parts of Common, which may have calls to its logging system. If there are two separate log systems, then such messages made in Common files will get lost.

[...]

PR #3021, opened 2026-07-21
AgfExport: commands for exporting lists of fonts, translations, and default config

ivan-mogilko:
Adds following commands to the agfexport tool:

* `audioclip-list` - exports list of audio clip files, their sources, and packing location, as comma-separated entries, e.g.:
    `au000001.ogg,C:\\Projects\\Audio\\OriginalMusic.ogg,vox`
* `custom-data-dir` - exports list of custom game data directories to include into final package.
* `game-cfg` - writes default game config.
* `font-list` - exports list of font files. Note that since in 3.x we cannot tell exact file names for fonts, this list is a *suggestion* containing all possible variants (wfn and ttf); a build tool should expect some of these files missing which is not a error case.
* `plugin-list` - exports list of game plugins.
* `tra-list` - exports list of translation sources (`.trs` in 3.x). This list may be used to know which translations to compile and include into game package.

### DIFF
--- a/Tools/agfexport/main.cpp
+++ b/Tools/agfexport/main.cpp
@@ -25,8 +25,9 @@ const char *HELP_STRING = ""
     "  game-cfg               Generate default game config\n"
     "  glvar                  Generate global variables scripts\n"
     "  header-list            Exports ordered list of headers from script modules\n"
-    "  script-list            Exports ordered list of scripts from script modules\n"
+    "  plugin-list            Exports list of game plugins\n"
     "  room-list              Exports list of active rooms\n"
+    "  script-list            Exports ordered list of scripts from script modules\n"
     "  tra-list               Exports list of translations\n"
     "  -h, --help             Show help message for command\n";
 
@@ -65,12 +66,12 @@ const char *HELP_HEADER_LIST = ""
     "  -h, --help             Show this help message\n"
     "  -t, --to-stdout        Instead write the list of scripts to stdout";
 
-const char *HELP_SCRIPT_LIST = ""
-    "Usage: agfexport script-list <INPUT-GAME.AGF> <OUT-FILE>\n"
-    "Writes <OUT-FILE>, a file with an ordered list of scripts from script modules."
+const char *HELP_PLUGIN_LIST = ""
+    "Usage: agfexport plugin-list <INPUT-GAME.AGF> <OUT-FILE>\n"
+     "Writes <OUT-FILE>, a file with a list of game plugins."
     "Commands:\n"
     "  -h, --help             Show this help message\n"
-    "  -t, --to-stdout        Instead write the list of scripts to stdout";
+    "  -t, --to-stdout        Instead write the list of rooms to stdout";
 
 const char *HELP_ROOM_LIST = ""
     "Usage: agfexport room-list <INPUT-GAME.AGF> <OUT-FILE>\n"
@@ -78,6 +79,13 @@ const char *HELP_ROOM_LIST = ""
     "Commands:\n"
     "  -h, --help             Show this help message\n"
     "  -t, --to-stdout        Instead write the list of rooms to stdout";
+
+const char *HELP_SCRIPT_LIST = ""
+    "Usage: agfexport script-list <INPUT-GAME.AGF> <OUT-FILE>\n"
+    "Writes <OUT-FILE>, a file with an ordered list of scripts from script modules."
+    "Commands:\n"
+    "  -h, --help             Show this help message\n"
+    "  -t, --to-stdout        Instead write the list of scripts to stdout";
 
 const char *HELP_TRA_LIST = ""
     "Usage: agfexport tra-list <INPUT-GAME.AGF> <OUT-FILE>\n"
@@ -93,8 +101,9 @@ enum CommandType
     kCmdGameCfg,
     kCmdGlVar,
     kCmdHeaderList,
-    kCmdScriptList,
+    kCmdPluginList,
     kCmdRoomList,
+    kCmdScriptList,
     kCmdTraList,
     kCmdMAX,
     kCmdNone = kCmdMAX
@@ -112,8 +121,9 @@ struct Command
         {"game-cfg",    kCmdGameCfg,    2, HELP_GAMECFG},
         {"glvar",       kCmdGlVar,      3, HELP_GLVAR},
         {"header-list", kCmdHeaderList, 2, HELP_HEADER_LIST},
-        {"script-list", kCmdScriptList, 2, HELP_SCRIPT_LIST},
+        {"plugin-list", kCmdPluginList, 2, HELP_PLUGIN_LIST},
         {"room-list",   kCmdRoomList,   2, HELP_ROOM_LIST},
+        {"script-list", kCmdScriptList, 2, HELP_SCRIPT_LIST},
         {"tra-list",    kCmdTraList,    2, HELP_TRA_LIST},
         {nullptr,       kCmdNone,       0, nullptr}
 };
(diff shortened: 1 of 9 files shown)


ivan-mogilko:
@ericoporto could you tell, was agfexport intended to execute several commands at once?

There's this code that execute list commands here:
https://github.com/adventuregamestudio/ags/blob/caa93b569e965f0e9504651b12c40a218f3c1e92/Tools/agfexport/main.cpp#L106-L134

where you have a sequence of `ifs` rather than `else if` or `switch`, and it sort of looks like intended to run commands in sequence, but i am not sure if that works.

ericoporto:
I don't think I had this intention, I probably just thought things grouped neatly that way, but if this is useful somehow, maybe it could support it.

ivan-mogilko:
I have an idea. What if instead of plain list of strings this tool would print data in INI format? Then it can create a single file filled with multiple different lists, divided into sections. Furthermore, this would also allow to output misc key-value data if necessary.

ivan-mogilko:
Hmm, actually, disregard the above. That will make things more complicated since the external tool will have to parse ini. At first I thought about a ini that contains a list of keys only, not key-value pairs, like:
```
[list id]
file
file
file
[list id]
file
file
file
```

But now when I think more on this, firstly one has to know these hardcoded "list ids", and second the '[' character is allowed in filenames too (although that's a edge case), so it will be not entirely possible to distinguish ini sections from files.

I suppose a better alternative could be to use simply empty lines to split sections from different commands:
```
file <--- first list begins here
file
file

file <--- second list begins here
file
file
```
the lists are ordered by order of commands in command line args, so the caller should know this order.

Of course if we allow this, the tool must not allow to combine "list" commands with non-list commands, like a list of rooms with autogenerated script header.

ericoporto:
Uhm, one idea here is if the produced list is somehow compatible with the pattern file we are using with agspak, and perhaps the sections could use a `#comment` to separate if we want for some reason.

the key value from the ini idea could be used for `path/in/filesystem=path/in/pak/file`, so the pattern file could use something `filesystem/path/pattern; pak/file/path` somehow.

What are all the tools that are mentioned to consume these lists? I don’t remember.

ivan-mogilko:
I am not sure if we may need to define a asset location per each pattern, because usually you just keep the relative directory structure as-is in case you pack a dir with subdirs. The problem with agspak is packing assets from multiple locations that must end up *in the same level* in the asset pack. So, likely, this instruction has to be on a more global level, outside of the patterns list.

I posted more detailed thoughts here: #3022.

---

Completely unrelated note, I am now also trying to rewrite the `crmpak` tool, which is currently terrible and mostly useless (besides replacing a compiled script). My plan is to make a small refactor in Common, which will help to reduce redundant dependencies if the full room structure parsing is added to this tool. Notably have Bitmap instances replaced with PixelBuffer in the loaded room struct, so that it won't drag all the Allegro drawing functionality with itself. After this, crmpak will have import/export commands for individual features that make sense atm: script, background and masks. After this, I think I will merge crm2ash into crmpak, and have it export room script ash as well, instead of having an extra tool for just that.

ivan-mogilko:
@ericoporto my suggestion is to rename "-t / --to-stdout" command to "-s / --stdout", because "to" part is too generic to use it for a command letter imo.

Also, switched regular tool logging from stdout to stderr when printing lists to stdout, similar to how it's done in the engine with "--tell..." commands. Of course this is just a quick fix, normally we should have a proper logger in the tools, which handles this.

ericoporto:
> my suggestion is to rename "-t / --to-stdout" command to "-s / --stdout",

that is alright. I think these and other command line changes can be made freely, and then later perhaps inspect all tools to see if there are things repeated between them so we try to "standardize" some flags across tools. But now I think these can all be change very freely.

> normally we should have a proper logger in the tools

Probably something for #2530, once you try to use all the tools together at once and start getting errors is when it feels more necessary.

ivan-mogilko:
Added a command for exporting a list of audio clips, as comma-separated entries of internal filename, source filepath, and packing location. Example of such list entry:
    `au000001.ogg,C:\\Projects\\Audio\\OriginalMusic.ogg,vox`

This list may be used for either of the two purposes:
1. Rebuilding AudioCache from sources in the game project.
2. Packing imported audio files into the ags/vox.

Tools/agfexport/main.cpp
```diff
@@ -8,31 +8,77 @@
...
+    "  * packing location, which is either 'ags' for the main game pack\n"
+    "    or 'vox' for the separate 'audio.vox'\n"
+#if (AGS_PLATFORM_OS_WINDOWS)
+    "Entry example: au000001.ogg,C:\\Projects\\Audio\\OriginalMusic.ogg,vox\n"
+#else
+    "Entry example: au000001.ogg,~/Projects/Audio/OriginalMusic.ogg,vox\n"
+#endif
+    "Commands:\n"
+    "  -h, --help             Show this help message\n"
+    "  -s, --stdout           Instead write the list of font files to stdout\n";
```
ericoporto: Erh, I think it should be "Instead write the list of **audio** files to stdout\n"? Same for custom data dir below.

Tools/agfexport/main.cpp
```diff
@@ -101,14 +176,45 @@ HError write_to_file(const String &content, const String &file)
...
+            exp_data.AppendFmt("%s,%s,%s\n", clip.CacheFileName.GetCStr(), clip.SourceFileName.GetCStr(),
+                clip.BundlingType == kAudioBundling_InMainData ? "ags" : "vox");
+        }
+    }
+
+    if (cmd == kCmdCustomDataDir)
+    {
+        std::vector<String> dirs;
+        AGF::ReadCustomDataDirectories(dirs, reader.GetGameRoot());
+        // Unfortunately, in AGS 3.x project there's no explicit indication of a filename,
```
ericoporto: I think this comment can be removed.

Tools/agfexport/main.cpp
```diff
@@ -8,31 +8,77 @@
...
+
+const char *HELP_FONT_LIST = ""
+    "Usage: agfexport font-list <INPUT-GAME.AGF> <OUT-FILE>\n"
+    "Writes <OUT-FILE>, a file with a list of font files used by the game\n."
+    "Commands:\n"
+    "  -h, --help             Show this help message\n"
+    "  -s, --stdout           Instead write the list of font files to stdout\n";
+
+const char *HELP_GAMECFG = ""
+    "Usage: agfexport gamecfg <INPUT-GAME.AGF> <OUT-FILE.CFG>\n"
```
ericoporto: Usage is `game-cfg` instead of `gamecfg`.

Tools/agfexport/main.cpp
```diff
@@ -283,11 +441,13 @@ int main(int argc, char *argv[])
...
+        fprintf(StdFile, "%s\n", err->FullMessage().GetCStr());
         return -1;
-    } else if(!stdout_list_print) {
-        printf("Data exported successfully.\n");
     }
-    return 0;
+    else if (!stdout_list_print)
+    {
+        fprintf(StdFile, "Data exported successfully.\n");
+        return 0;
```
ericoporto: I think this return 0 was moved inside the if by accident.

Tools/agfexport/main.cpp
```diff
@@ -79,20 +145,29 @@ struct Command
     const size_t NumArgs;
     const char *Help;
 } Command[] = {
+        {"audio-list",  kCmdAudioClipList, 2, HELP_AUDIOCLIP_LIST},
```
ericoporto: this is `audio-list` here but you mentioned `audioclip-list` in the help and in the commit message.

ericoporto:
Hey, I didn’t comment before but this looks alright afaict, I think it can be merged

PR #3023, opened 2026-07-24
agspak: support multiple input dirs

### LINKED ISSUE #3022: agspak: support gathering assets from multiple places
Opened this for a discussion. The problem in a nutshell is this: there may be situation when we need to gather assets from multiple places, for example, one place is the game project's root directory, another is a temporary directory with intermediate compilation output, and so forth. But at the same time we need to be able to put these files into the same place in the virtual asset filesystem, without correspondence to their relative location on disk.

In other words, we need a mechanism where we say (by using command-line arguments): files from this directory on disk end up in this "directory" in the asset pack. And be able to do so multiple times for multiple separate lists of files.

As of now agspak works like this: there's "working dir" arg which defines the root from where the files are packed, then we can pass a "pattern file" (file made in format similar to .gitignore which specifies which files to include/exclude), and also optionally pass a list of inclusion patterns directly as a command arg. An additional arg tells whether to include all found subdirs recursively. The assets included this way are assigned a virtual path which maintains relative directory structure starting with the "working dir" treated as a root.

Instead we'd need some args syntax that would allow us to define:
- real dir,
- pattern file and/or list of match patterns
- destination dir in the assets table
- recursive subdirs flag (?)

And we need to be able to have multiple of the above in a single command line.

Above is my thought, perhaps there may be alternative ways to organize this.

EDIT: On second thought, possibly, there's no need to tell where to place the assets to, only a way to list multiple different source locations, treating each as going into the pack's root (and it's subdirs go into corresponding pack's subdirs).

ivan-mogilko:
Resolve #3022

Add `-D dir` and `-R dir` options for "create" command, that add more separate input directories, where `-D` instructs to read only this exact dir's contents and `-R` instructs to read this dir and all the nested dirs recursively.
The files found in these directories will be put relative to the asset pack's virtual filesystem's root. So assets generated from the input dir itself will have no parent path, and assets generated from nested dirs will have relative parent paths.

For handling asset duplicates added two more options:
- `--replace-dup` - every next found duplicate overrides the previous one. As the dirs are checked in the same order as they were in the command line, this allows to accurately override one dir contents with another.
- `--skip-dup` - skip duplicates (the first found asset of this name will be used).

If neither of the above two options is provided, then any found duplicate will cause error.

### DIFF
--- a/Tools/agspak/commands.cpp
+++ b/Tools/agspak/commands.cpp
@@ -110,15 +110,12 @@ static HError CutAssetLibrary(const String &pak_file, bool verbose)
     return HError::None();
 }
 
-static HError PrepareAssetLibrary(AssetLibInfo &lib,
-    const String &src_dir, const String &dst_pak,
-    const std::vector<String> &pattern_list, const String &pattern_file,
-    bool do_subdirs, size_t part_size_mb, bool verbose)
+// Makes the list of assets from a single input directory, applying pattern list and/or rules from pattern file.
+// Fills a map of asset-name -> filepath pairs.
+static HError MakeAssetMapFromADir(const String &asset_dir, StringIMap &asset_map,
+    const std::vector<String> &pattern_list, const String &pattern_file, bool do_subdirs, bool verbose)
 {
-    const String &asset_dir = src_dir;
-    const String &lib_basefile = dst_pak;
     const bool has_pattern_file = !pattern_file.IsEmpty();
-
     std::vector<String> files;
     HError err = MakeListOfFiles(files, asset_dir, do_subdirs);
     if (!err)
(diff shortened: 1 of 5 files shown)


ericoporto:
About the asset duplicates, in a regular build from the editor (imagine we are building with the command line tools from there), we would not hit it, right? Just to check if these options are one of those that are added just in case.

ivan-mogilko:
The tool must have them as this is a real chance when having multiple input dir, so user should have a way to instruct the program what to do in such case.

In my case of testing full build with tools this happened because I was making copies of roomN.crm files and cutting or replacing scripts in them, so there was a original roomN.crm in the project dir, and a modified variant in the temp dir.
Editor may do this differently by modifying the ones in the project dir maybe, that's still under question.

Also, in theory, there's always a chance that someone puts unnecessary files in the project dir, which names match the names of the intermediate output assets.

ericoporto:
ok, for the ones that don’t error it would be nice if some warning is produced - I haven’t checked the code!

ivan-mogilko:
> ok, for the ones that don’t error it would be nice if some warning is produced - I haven’t checked the code!

It does:
https://github.com/ivan-mogilko/ags-refactoring/blob/2a9e3de219f261c84e5c61185b3dddf70de62949/Tools/agspak/commands.cpp#L189-L202

ericoporto:
The replace duplicate is meant to keep the last found or the first found? I am trying to understand it.

ivan-mogilko:
> The replace duplicate is meant to keep the last found or the first found? I am trying to understand it.

As said in the description, *"--replace-dup - every next found duplicate overrides the previous one."*. Which means that the last found will be packed.

Tools/agspak/main.cpp
```diff
@@ -212,7 +242,7 @@ int main(int argc, char *argv[])
 {
     printf("%s\n", BIN_STRING);
 
-    CmdLineOpts::ParseResult cmdargs = CmdLineOpts::Parse(argc, argv, {"-p", "-f", "--pattern-file"});
+    CmdLineOpts::ParseResult cmdargs = CmdLineOpts::Parse(argc, argv, {"-p", "-f", "--pattern-file", "-D", "--dir", "-R", "-recursive-dir"});
```
ericoporto: The recursive-dir is missing a hyphen

Tools/agspak/commands.cpp
```diff
@@ -227,31 +266,55 @@ int Command_Cut(const String &src_pak, bool verbose)
...
-                   const std::vector<String> &pattern_list, const String &pattern_file,
-                   bool do_subdirs, size_t part_size_mb, bool verbose)
+int Command_Create(std::vector<std::pair<String, bool>> &src_dirs, const String &dst_pak, bool append,
+                   const std::vector<String> &pattern_list, const String &pattern_file, DuplicateAssetAction dup_action,
+                   size_t part_size_mb, bool verbose)
 {
-    printf("Input directory: %s\n", src_dir.GetCStr());
+    if (src_dirs.size() == 1)
+    {
+        printf("Input directory: %s", src_dirs[0].first.GetCStr());
```
ericoporto: This is missing a new line at the end that we usually add to all printfs.

ericoporto:
Hey, the WriteLibraryFile you changed in the mfl_utils.cpp didn't reflect in the mfl_utils.h, but I think it didn't error because the only use is inside the mlf_utils.cpp itself.

ericoporto:
[paktest.zip](https://github.com/user-attachments/files/30392085/paktest.zip)

Below are my tests here on my computer, I think the `--replace-dup` flag is not working.

Hey, this is a directory named paktest. It has a file named

```
1.txt -> contains text "original"
2/ dir
  - 1.txt -> contains text "duplicate"
empty -> an empty dir
```

Now, because I HAVE to add a directory if I want to only have the flat style I added the empty dir. Anyway, I put all of this in a dir named paktest where my agspak.exe is, and call the following

```
./agspak.exe -c test.pak -D paktest/ paktest/empty/ --replace-dup
```

Now, the last flag doesn't matter because it is not recursive so it only finds the first file, and I can extract to confirm it.

```
./agspak.exe -e test.pak out/
```

Ok, now lets use 

```
./agspak.exe -c test.pak -R paktest/ paktest/empty/ --replace-dup
```

and it also doesn't matter because the output will preserve the directories. OK, now lets for real try this

```
./agspak.exe -c test.pak -D paktest/ -D paktest/2/ paktest/empty/ --replace-dup
agspak v0.5.0 - AGS game packaging tool
Copyright (c) 2026 AGS Team and contributors

Input directories: paktest/empty/
                   paktest/
                   paktest/2/
Output pack file: test.pak
Warning: duplicate asset '1.txt' [replace]: first found: 'paktest/1.txt', new found: 'paktest/2/1.txt'
Pack file(s) written successfully.
Done.
```

Note the message! Now if I extract this generated package, there is a single file there, it is `1.txt`, but its contents is "original" and not "duplicate".

ivan-mogilko:
Turned out i used incorrect insert function, should be insert_or_assign to add a possibly matching key.

ericoporto:
That is C++17… Hey, what do you think about moving on from our very oldold  Debian and going for something more recent and then putting a big notice somewhere for people shipping Linux games on Steam to put a newer base linked there? Then we could move to newer C++ version across platforms.

ivan-mogilko:
Fixed again, using more classic variant.

> That is C++17… Hey, what do you think about moving on from our very oldold Debian and going for something more recent and then putting a big notice somewhere for people shipping Linux games on Steam to put a newer base linked there? Then we could move to newer C++ version across platforms.

Given it will soon be 10 years since C++17 format introduction, i think we should, perhaps right after 3.6.3 release?

Tools/data/mfl_utils.cpp
```diff
@@ -110,44 +110,33 @@ HError MakeListOfFiles(std::vector<String> &files, const String &asset_dir, bool
...
-    for (FindFile ff = FindFile::Open(parent, "*", true, false, do_subdirs ? -1 : 0);
-        !ff.AtEnd(); ff.Next())
+    // Generate the list of AssetInfos, as we must precalculate file sizes
+    // before arranging them among the library partitions.
+    std::vector<AssetInfo> assets;
+    for (const auto &asset_entry : asset_map)
     {
+        soff_t filesz = File::GetFileSize(asset_entry.second);
+        if (filesz < 0)
+            return new Error("Failed to retrieve asset file's size: %s", asset_entry.second);
```
ericoporto: This `Error(` should be `Error(String::FromFormat(`

ericoporto:
I opened an issue about the Linux stuff in #3026 and other than the comment above, this looks alright now.

Tools/data/mfl_utils.cpp
```diff
@@ -128,7 +128,7 @@ HError MakeAssetLib(AssetLibInfo &lib, const String &lib_basefile,
     {
         soff_t filesz = File::GetFileSize(asset_entry.second);
         if (filesz < 0)
-            return new Error("Failed to retrieve asset file's size: %s", asset_entry.second);
+            return new Error(String::FromFormat("Failed to retrieve asset file's size: %s", asset_entry.second));
```
ericoporto: I didn’t noticed before, but second is a String, so it also needs a GetCStr() call.

I checked all the added formats and this is the only place where this call was missing.

ericoporto:
Looks good, let’s merge and use it. :)

PR #3024, opened 2026-07-26
crmpak: complete rewrite, allow to import/export/cut specific room contents

ivan-mogilko:
Completely rewrote crmpak, as the previous implementation was done wrong and almost useless (with exception of replacing compiled script).

Usage: `crmpak <COMMAND> <ROOM-FILE> [<CONTENT-LIST>] [<OPTIONS>]`

Content list is a sequence of key-value pairs separated by spaces, where the key is a content name and the value is a file path:
`CONTENT1 FILE1 CONTENT2 FILE2 ...`
Certain commands (such as '-x') may not require specifying file paths, in which case content list must contain only sequence of content names:
`CONTENT1 CONTENT2 ...`
Certain other commands (such as '-l') do not require a content list.

Available commands:
 * -c, --create - create a room file with specified content; if no content list is provided, then will create a room with minimal default content.
 * -e, --export - export (extract) content from the existing room file into the output file(s).
 * -i, --import - import (add, overwrite) specified content into the existing room file.
 * -l, --list - print the list of content found inside room file.
 * -x, --cut - cut (delete) specified content from a room file.

Supported content to import/export/cut:
 * ash - generated room script header (export only)
 * backN - background image, where N is a 0-based index; supported image formats: BMP, PCX, PNG.
 * mask-hotspot
 * mask-region
 * mask-walkarea
 * mask-walkbehind
 * script - compiled script
 * script-text - script text (commonly present in v2.x rooms)

Command options:
 * -w <out-room.crm> - for import and cut commands: write the resulting room into a new file; otherwise will modify the input file.

Other options:
 * -v, --verbose - print operation details

**Examples:**

export main background:
`crmpak -e room1.crm back0 background0.png`

import (replace) script:
`crmpak -i room2.crm script room2.o`

cut script from the room, but save room into another file
`crmpak -x room3.crm script -w packaged_files/room3.crm`

---

Besides the tool itself this PR contains certain refactor, with aim to avoid dependency on Allegro 4's BITMAP and drawing code (which is the biggest remaining part of the stripped Allegro 4 library source we're still using).

Because historically RoomStruct class had Bitmaps in it (for backgrounds and masks), I have picked a RoomData class, which has the main bulk of the loaded room data, and keeps all images as PixelBuffer objects: much simpler data type which is just a pixel array with meta data and nothing else. RoomStruct inherits RoomData, and has Bitmap objects, in which it transfers the respective pixel buffers after being loaded. RoomStruct is meant to be used by the engine, while RoomData class can be used by the tools.

Other things that I've changed (these may be tracked by reading commits):
- Moved string encryption functions into separate cpp file, avoid including unnecessary text parser dictionary into room tool;
- Removed any reference to AssetManager from room_file.cpp, as that is only needed by the runtime engine so far.
- Merged room_file_base.cpp back into room_file.cpp (it was split earlier as an attempt to reduce dependencies in tools).
- Merged path_ex.cpp back into path.cpp. It was also split a while ago for the same reasons. But I figured out that this is inconvenient, and not necessary now, when Allegro library is stripped, and apparently only select files may be included into the tools (which i did as necessary).

---

TODO:
1. Editor still requires fixing after RoomStruct->RoomData split, some room operations in AGS.Native probably do not work correctly.
2. crmpak needs more testing.
3. Review content-list syntax? I've been also thinking about making it a comma-separated list of key=value pairs.
4. [DONE?] Tools' makefiles need to be updated.
5. [DONE] crmpak needs to support exporting script text for rooms made in AGS 2.x.
6. [DONE] want to merge crm2ash into crmpak, as one of its export options.


Tools/crmpak/commands.cpp
```diff
@@ -0,0 +1,425 @@
...
+#include "util/file.h"
+#include "util/path.h"
+
+using namespace AGS::Common;
+
+namespace CRMPak
+{
+
+const CstrArr<CRMPak::kNumContentTypes> ContentNames = 
+    {{ "undefined", "back", "mask-hotspot" "mask-region" "mask-walkarea" "mask-walkbehind", "script", "script-text" }};
```
ericoporto: There are a few commas missing here and below

ericoporto: Also perhaps this could use the command structure from agfexport? It allows also to associate the number of positional arguments with the commands more easily.

ivan-mogilko: The way these content args are currently done is that they are not limited to a exact number. They can be a sequence of any length, appended after command.

ericoporto:
Hey, because of the path and path_ex merge now everything should require allegro. I think also we had BitmapData before PixelBuffer.

ivan-mogilko:
Yes, I am adding 3 files from allegro everywhere now.

Tools/crmpak/commands.cpp
```diff
@@ -0,0 +1,476 @@
...
+bool SaveRoomFile(const RoomDataExt &room, const String &filename)
+{
+    auto out = File::CreateFile(filename);
+    if (!out)
+    {
+        printf("Error: failed to open room file for writing.\n");
+        return false;
+    }
+
+    HError err = WriteRoomData(&room, out.get(), kRoomVersion_Current);
```
ericoporto: This will cause an error because HRoomFileError is not HError compatible I think

Engine/ac/room.cpp
```diff
@@ -551,6 +551,18 @@ static void init_object_states(size_t start, size_t end)
     }
 }
 
+HError LoadRoom(const String &filename, RoomStruct *room, AssetManager *mgr, bool game_is_hires, const std::vector<SpriteInfo> &sprinfos)
+{
+    auto in = mgr->OpenAsset(filename);
+    if (in == nullptr)
+        return new RoomFileError(kRoomFileErr_FileOpenFailed, String::FromFormat("Filename: %s.", filename.GetCStr()));
```
ericoporto, line 558: Is RoomFileError compatible with HError?

Tools/crmpak/commands.cpp
ivan-mogilko: I am reading logs, and fixing these mistakes. MSVS compiler does not error on everything that other compilers error on, so i have to wait when github actions find anything.

Engine/ac/room.cpp
ivan-mogilko: Yes, any child of Error is compatible with HError, because they are stored as a pointer to the base Error class.

Tools/crmpak/Makefile
```diff
@@ -37,8 +53,29 @@ COMMON_OBJS = \
 	../../Common/util/string_compat.c \
 	../../Common/util/string_utils.cpp
 
+TOOL_OBJS = \
+	../../Tools/data/cc_script_stubs.cpp \
+	../../Tools/data/room_utils.cpp
+
+MINIMAL_ALLEGRO_OBJS = \
+	../../libsrc/allegro/src/allegro.c
```
ericoporto: You need to add a trailing `\` to all lines here except ufile.c

Tools/spritepak/Makefile
```diff
@@ -46,7 +46,11 @@ COMMON_OBJS = \
 	../../Common/util/string_utils.cpp
 
 MINIMAL_ALLEGRO_OBJS = \
+	../../libsrc/allegro/src/allegro.c
```
ericoporto: Here it needs also a trailing `\`

Edit: same on all other Makefiles too.

ivan-mogilko:
Alright, seems like the Editor is the only thing remaining to be fixed here.

ivan-mogilko:
Ok, this should be ready for review and testing now.

There's one thing that I still dont like how it's done now, because editor uses RoomStruct with bitmaps, it need to convert bitmaps back to pixel buffers before saving RoomData, as the writing function expects pixel buffers. Maybe i will figure out a better way afterwards.
(This will look differently in AGS 4, as it has room drawing in C# code, and rooms are only saved to crm in native code, so it can just convert to RoomData directly there.)

Common/game/roomdata.cpp
```diff
@@ -204,6 +339,18 @@ float RoomData::GetMaskScale(RoomAreaMask mask) const
     }
 }
 
+void RoomData::SetMask(RoomAreaMask mask, PixelBuffer &&pxbuf)
+{
+    switch (mask)
+    {
+    case kRoomAreaWalkBehind: WalkBehindMaskBuf = std::move(pxbuf);
```
ericoporto: Hey, I think this switch was made copy pasting the getmask one but then the breaks were forgotten!

Editor/AGS.Native/agsnative.cpp
```diff
@@ -4460,17 +4466,23 @@ void save_room_file(RoomStruct &rs, const AGSString &path)
...
-    rs.BackgroundBPP = rs.BgFrames[0].Graphic->GetBPP();
+    rs.BackgroundBPP = rs.BgImages[0]->GetBPP();
     for (int i = 0; i < 256; ++i)
         rs.Palette[i] = rs.BgFrames[0].Palette[i];
 
     std::unique_ptr<Stream> out(AGSFile::CreateFile(path));
     if (out == NULL)
         quit("save_room: unable to open room file for writing.");
 
+    // FIXME: this is quite ugly, but this is the most straightforward way
```
ericoporto, line 4477: I think I am stupid, why is this Ugly? Is this a problem because of the having only one room open at a time issue? (did we ever had an actual issue ticket for opening multiple rooms in the editor?)

ericoporto: I think I get it, calling that prepare function is really needed and it is also only used here.

ivan-mogilko: I call this ugly, because this changes a "RoomStruct" object state from "can use bitmaps for drawing" to "can save pixel data into file" and then back, which is not something that is done normally.

Common/game/roomdata.cpp
ivan-mogilko: Oh, right. Fixed.

ericoporto:
I think this looks alright, it does quite a few changes that makes me eager to try the GUI data refactor too, and I also now notice I need to add debug builds from CMake and makefile builds to the CI too.

But other than this I don’t have much to say, I guess when using all tools together we will see what happens? I don’t understand much of the need of all the functionality here other than to be like image magick is to someone’s room editor, but this will probably become useful when making the ags4 version that has to build everything from XML too.

There is a thing a noticed is there is quite a few read block functions for the room crm. I thought the stream stuff would be in room data and only anything file specific would go into the room file, so I guess I didn’t quite get the separation. Also I noticed some stuff is separate in data helpers and others, I am curious, is the crm similar somehow to clib or some other binary file format we use?

Anyway, I would suggest merging, the more critical thing here is the entire refactoring which appears to at least pass the CI tests, so I think they are alright.

ivan-mogilko:
>  I don’t understand much of the need of all the functionality here other than to be like image magick is to someone’s room editor, but this will probably become useful when making the ags4 version that has to build everything from XML too.

Yes, for compiling a 3.x project one would require only the command that replaces or removes a compiled script. Other content types can be interesting when used to decompose the compiled games for diagnostic purposes, or restoring lost sources, and maybe third-party editing tools. Indeed in 4.x we'd need full functionality, and also new command that would generate main room data from data.xml.

> There is a thing a noticed is there is quite a few read block functions for the room crm. I thought the stream stuff would be in room data and only anything file specific would go into the room file, so I guess I didn’t quite get the separation. Also I noticed some stuff is separate in data helpers and others, I am curious, is the crm similar somehow to clib or some other binary file format we use?

It goes several years back when I decided to separate room data struct and reading into different files. The reasons were that the room file is a collection of blocks, so it's easier to be done reading and writing by a separate class that deals with this block structure. Also the file can contain extra data that is not necessary at runtime. Today there's also extensions, similar to main game data extensions, and so I find it more convenient to have all this read/write code in a separate file.

At the same time, there may be some remains from old crmpak in data helpers, they serve no immediate purpose, but may be used as a reference later. I had functions that help picking out minimal data from the room file, such as script names for ash header. But since now crmpak allows multiple exported things at once, this program would need a smarter approach, depending on which combination of content is requested. This might speed things up when generating room headers for a big collection of rooms. Perhaps this may be done afterwards.

EDIT: there's a "room block reader" class, that just parses block by block and calls a "reader" function callback for each one of them. I think it may be reorganized to allow custom reading callbacks, which skip unwanted data.

ivan-mogilko:
Alright, so, I added a very basic performance optimization for this tool, which lets skip most data when exporting individual content, such as generated script header with object names, compiled script, or script text from legacy rooms. This is done in a bit clumsy way, but should not affect other uses of room files (in engine); and i'd leave any improvement of this for the latter. My intent here was to optimize at least the most common scenario when the rooms are parsed for generating script headers, so that the tool won't load all the unnecessary images or compiled script along the way (the biggest room data).

ericoporto:
I think it’s fine. I would recommend to merge it since it touch a lot of stuff and anything that goes after will need to wait it merge anyway to be able to properly rebase.

ericoporto:
Hey, I am trying to figure it out how to extract the ash files, I can't understand the help?

Edit: I think I need to use the export command and also the ash parameter, but I am unsure how to pass the name of the resulting ash file.

ivan-mogilko:
Both export and import commands require a name of room file, followed by the content list with pairs of arguments `CONTENT FILE` repeated for any number of times, where `CONTENT` is a name of content and `FILE` is either source or destination file (on import or export respectively).

`crmpak -e roomN.crm ash roomN.ash` - would be a command to export ash header.

`crmpak -e roomN.crm ash roomN.ash script roomN.o` - exports ash header and compiled script into two separate files.

PR #3030, opened 2026-07-31
Tool: spriteimport

### LINKED ISSUE #3029: Tool: spriteimport
Task is to write a C++ cli tool for performing sprite import from the collection of source image files, following specifications from Game.agf.
Previously I thought to add such operation into existing "spritepak" tool, but then came to a conclusion that having a separate tool may be better, because sprite import operation requires different set of dependencies (parsing Game.agf, but also room files for 8-bit games that may require using room palettes), and its own, exclusive set of parameters.

The idea of a "spriteimport" tool is to replicate Editor's operation of "recreating sprite file from sources".

**Inputs:**
- Game.agf
- directory with room files (optional, for 8-bit games only)
- available source images (implicit, their paths are read from Game.agf)

**Outputs:**
- variant 1: a ready compiled spritefile (its internal format options are determined by the command args, similar to the "spritepak").
- variant 2: a collection of image files in a single output directory, each corresponding to one sprite, and named as "sprNNNNNN.ext", where image format is determined by the command arg.

**Operation**
For each sprite spec from Game.agf, import a image file, cut the tile or a separate frame (gif case) as necessary, perform necessary conversions of the color depth and transparency, then save the resulting image to the output (write spritefile or a new image file in output dir).

The above is a straightforward approach, but for a better performance this tool may do a preliminary preparations, where it will group sprites by their source files (as one source image may contain multiple sprites), and then load certain source file only once to cut all sprites from there. Similarly, for 8-bit games it may preload and cache all the required room palettes in memory instead of loading a room data each time a new sprite requires one.

ivan-mogilko:
Resolve #3029

This implements a "spriteimport" cli tool, which purpose is to import sprites from source images, according to the Game.agf specs, and then either output them as individual image files (one image file per sprite), or compile into the spritefile.

Usage: `spriteimport <COMMAND> <GAME.AGF> <OUT-PATH> [<COMMAND-OPTIONS>] [<OPTIONS>]`

Commands:
* -c, --out-pak - outputs a compiled spritefile. Additional options '-n', '-s' and '-z' modify this command's behavior.
*  -f, --out-files - outputs sprites as individual image files into the specified directory. The files's names and format is determined by  the pattern option (see '-p'). If not such pattern provided then uses "spr%6N%.png" by default.

NOTE: the command options mimic ones used for "spitepak"'s import and export commands.

Command options:
* -n, --index <indexfile> - when output is the spritefile: specifies the accompanying sprite index file.
* -p, --pattern <name pattern> - when output is image files: use the given pattern to define their naming and image format. The pattern may be given with or without file extension. The pattern may contain following placeholders: * %xN% - sprite number, where 'x' may be any single digit integer specifying number of padding zeroes, e.g. "%6N%." If no pattern is provided, the program will use "spr%6N%" pattern by default. If no extension is specified in the pattern, then ".png" will be used as an extension.
* -s, --storage-flags <flags> - when output is the spritefile: use additional storage options, defined using a hexadecimal bitset:
    * 0x01 - optimize storage size when possible; e.g. write 16/32-bit images as 8-bit images with palette (only when this achieves less space). Default is "0x01".
* -z, --compress <type> - when output is the spritefile: use compression:
    * none
    * rle
    * lzw
    * deflate
    Default is "deflate".

Other options:
* -v, --verbose - print operation details

---

Implementation details.
1. This tool optimizes importing tiled sprites. It first gathers a map of all sprite sources and the sprites that originate from them, then loads each source image file only once, cutting all respective sprites out of it in one go.
2. Because of the above, when compiling a spritefile itself, it does so in a non-trivial way. As it processes one source file after another, it compiles a temporary spritefile, where sprites are in a "random" order. After this is done, it starts the new, final spritefile, and copies data between temp and final ones, making the order of sprites correct.
3. If this tool is told to output individual image files, these may be further used by "spritepak" to compile a spritefile if needed.

---

TODO:
1. [DONE] This draft PR does not do any image conversions yet, so transparency option does not work.
3. [DONE] 8-bit images optionally require to sync with the game's or room's palette.
4. Makefile.

WON'T DO IN THIS PR:
1. Other image formats, such as GIF. Less commonly used, will require adding GIF loading support (may port from stb_image). I'd rather do this as a separate task.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -251,6 +251,7 @@ if(AGS_TESTS)
         test/inifile_test.cpp
         test/math_test.cpp
         test/memory_test.cpp
+        test/paletteop_test.cpp
         test/path_test.cpp
         test/splitline_test.cpp
 		test/spritecache_test.cpp
--- a/Common/gfx/bitmapdata.cpp
+++ b/Common/gfx/bitmapdata.cpp
@@ -46,18 +46,34 @@ namespace PixelOp
 void CopyPixels(const uint8_t *src_buffer, const int bpp, const size_t src_pitch,
     const int width, const int height, uint8_t *dst_buffer, const size_t dst_pitch)
 {
-    CopyPixelsRegion(src_buffer, bpp, src_pitch, 0u, width, height, dst_buffer, dst_pitch, 0u);
+    CopyPixelsRegion(src_buffer, bpp, src_pitch, 0, 0, width, height, dst_buffer, dst_pitch, 0, 0);
 }
 
 void CopyPixelsRegion(const uint8_t *src_buffer, const int bpp, const size_t src_pitch,
-    const size_t src_px_off, const int width_px, const int height_px,
-    uint8_t *dst_buffer, const size_t dst_pitch, const size_t dst_px_off)
+    const int src_px_off, const int src_py_off, const int width_px, const int height_px,
+    uint8_t *dst_buffer, const size_t dst_pitch, const int dst_px_off, const int dst_py_off)
 {
-    const size_t src_boff = src_px_off * bpp;
-    const size_t dst_boff = dst_px_off * bpp;
+    const size_t src_boff = (src_py_off * src_pitch) + (src_px_off * bpp);
+    const size_t dst_boff = (dst_py_off * dst_pitch) + (dst_px_off * bpp);
     const size_t width_b = width_px * bpp;
     // NOTE: all assertions are done further in Memory::BlockCopy
     Memory::BlockCopy(src_buffer, src_pitch, src_boff, width_b, height_px, dst_buffer, dst_pitch, dst_boff);
+}
+
+PixelBuffer CopyPixelsRegion(const BitmapData &bm_data, const int src_x, const int src_y, const int width, const int height)
+{
+    const int src_bpp = bm_data.GetBytesPerPixel();
+    const size_t src_pitch = bm_data.GetStride();
+    int do_src_x = src_x, do_src_y = src_y;
+    int do_width = width, do_height = height;
+    Math::ClampLength(do_src_x, do_width, 0, bm_data.GetWidth() - 1);
+    Math::ClampLength(do_src_y, do_height, 0, bm_data.GetHeight() - 1);
+    if (do_width == 0 || do_height == 0)
+        return {};
+
+    PixelBuffer out_buf(do_width, do_height, bm_data.GetFormat());
+    CopyPixelsRegion(bm_data.GetData(), src_bpp, src_pitch, do_src_x, do_src_y, do_width, do_height, out_buf.GetData(), out_buf.GetStride(), 0, 0);
+    return out_buf;
 }
 
 bool CopyConvert(const uint8_t *src_buffer, const PixelFormat src_fmt, const size_t src_pitch,
--- a/Common/gfx/bitmapdata.cpp
+++ b/Common/gfx/bitmapdata.cpp
@@ -39,6 +39,38 @@ uint32_t BitmapData::GetPixel(int x, int y) const
     }
 }
 
+void BitmapData::SetPixel(int x, int y, uint32_t value)
+{
+    uint8_t *line = GetLine(y);
+    switch (_bitsPerPixel)
+    {
+    // for 1-bit and 4-bit: clear the bit(s) position to zero and then write new value
+    case 1: line[x / 8] = line[x / 8] & ~(1 << (7 - x % 8)) | ((value & 1) << (7 - x % 8)); break;
+    case 4: line[x / 2] = line[x / 2] & ~(4 << (1 - x % 2)) | ((value & 4) << (1 - x % 2)); break;
+    case 8: line[x] = static_cast<uint8_t>(value & 0xFF); break;
+    case 15: /* same as 16 */
+    case 16: *reinterpret_cast<uint16_t*>(&line[x * 2]) = value; break;
+    case 24: Memory::WriteInt24(&line[x * 3], value); break;
+    case 32: *reinterpret_cast<uint32_t*>(&line[x * 4]) = value; break;
+    default: assert(false); break;
+    }
+}
+
+String PixelFormatName(PixelFormat fmt)
+{
+    switch (fmt)
+    {
+    case kPxFmt_Indexed1:   return "1-bit indexed";
+    case kPxFmt_Indexed4:   return "4-bit indexed";
+    case kPxFmt_Indexed8:   return "8-bit indexed";
+    case kPxFmt_R5G5B5:     return "15-bit R5G5B5";
+    case kPxFmt_R5G6B5:     return "16-bit R5G6B5";
+    case kPxFmt_R8G8B8:     return "24-bit R8G8B8";
+    case kPxFmt_A8R8G8B8:   return "32-bit A8R8G8B8";
+    default:                return "unknown";
+    }
+}
+
 
 namespace PixelOp
 {
@@ -76,6 +108,8 @@ PixelBuffer CopyPixelsRegion(const BitmapData &bm_data, const int src_x, const i
     return out_buf;
 }
 
+// TODO: quite possibly may be *at least partially* reimplemented as a template function,
+// with parameters defining pixel sizes and rgb shifts.
 bool CopyConvert(const uint8_t *src_buffer, const PixelFormat src_fmt, const size_t src_pitch,
     const int width, const int height, uint8_t *dst_buffer, const PixelFormat dst_fmt, const size_t dst_pitch)
 {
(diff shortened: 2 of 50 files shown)


ivan-mogilko:
This is done for the most part, in theory, but I still need to add some tests for the sprite conversion, where the precreated bitmap data is being converted to the game's compatible color depth and applied transparency according to the option.

Also, the spriteimport's code now partially duplicates the code from AGS.Native library (see Tools/data/sprite_utils.cpp), and I've been thinking if it's possible to make AGS.Native link the newer tool's code instead. Interestingly, I was almost certain that I'll have to link all the Allegro 4 into this tool, but it turned out that's not necessary, as all pixel data conversions can already be done without it.

ivan-mogilko:
I think that this is ready for review.

There's a strange issue with 1 test failing in Windows MSVC Tests GHA, although it's not failing for me locally, and it seems not failing in other GHA that run tests. I will investigate this.

Secondly, I decided to postpone replacing the old code in AGS.Native with the new one from tools. I'd like to have it tested more for a while. Also there's a 3.6.3 release coming, perhaps it would be better to do this change afterwards.

ericoporto:
About the failing test Import8RemapToGamePalette and the line

```
EXPECT_TRUE(std::equal(image2.GetData(), image2.GetData() + image2.GetDataSize(), dest.GetData()));
```

I don’t quite get, three things are being checked for equality here? I read [std::equal docs on cpp reference](https://en.cppreference.com/cpp/algorithm/equal) and it seems to apply for a range but I don’t fully understand it yet.

About environment differences, if you are using VS2019 or earlier, I wonder if there is a change of the implementation of std::equal that could affect this behavior in C++20, because the docs on cpp reference mark some things that are from C++20 and few from earlier. VS2022 is the one used in the CI for now (we will probably be forced to move to VS2026 at some point in the next year or the other). My memory is VS doesn’t have earlier C++ standards to build - like even before when we used VS2015, it already forced C++14 ignoring any settings.

I will give a check to the rest of the PR at a later time, I am traveling right now, but I have my computer with me this time so I can check things out during the night.

ivan-mogilko:
> I don’t quite get, three things are being checked for equality here? I read [std::equal docs on cpp reference](https://en.cppreference.com/cpp/algorithm/equal) and it seems to apply for a range but I don’t fully understand it yet.

This compares two arrays byte by byte.

Tools/data/sprite_utils.cpp
```diff
@@ -0,0 +1,709 @@
...
+{
+    // Ensure that transparent color is index 0
+    if (transcol != 0)
+    {
+        pal[transcol] = pal[0];
+        PaletteOp::SetRGB(pal, 0, 0, 0, 0); // set index 0 to black
+    }
+
+    Palette dst_pal;
+    for (int i = 0; i < 255; ++i) // CHECKME: 255th index is ignored?
```
ericoporto: Why it doesn’t go to 256? (or PAL_SIZE)

ericoporto:
@ivan-mogilko the CI builds for release, are your locals builds debug builds?

> This compares two arrays byte by byte.

It is 16 pixels, it's in theory not that much, I think I would prefer to compare pixel by pixel instead so we find out WHERE it fails. In the tap test you once had a check [that became a comment](https://github.com/adventuregamestudio/ags-test-games/blob/854d38aca48cc0f0d03cc20ff536cebd75527667/ags3/auto-test/gfx-compare.asc#L30), this pixel by pixel check makes things easier, couldn't the assert equal be inside a simple for clause and then we have the i or the x and y of the pixel?

Tools/data/sprite_utils.cpp
ivan-mogilko: No idea, this is copied from the editor's code. Could be a typo that remained for years.

Common/gfx/bitmapdata.cpp
```diff
@@ -262,7 +320,100 @@ void CopySwapRGBA(const uint8_t *src_buffer, int src_r_shift, int src_g_shift, i
...
+            {
+                color_mapped_table[i] = bestfit_color(dst_pal, src_pal[i].r, src_pal[i].g, src_pal[i].b);
+            }
+        }
+
+        if (keep_transparent)
+        {
+            // keep transparency
+            color_mapped_table[0] = 0;
+            // any other pixels which are being mapped to 0, map to 16 instead
```
ericoporto, line 410: this 0->16 mapping is also "allegroism" of things that existed or this makes sense in some way?

Common/util/wgt2allg.cpp
```diff
@@ -120,50 +100,3 @@ int my_setcolor(const int color, int color_depth, bool fix_alpha)
...
-      else
-      {
-        color_mapped_table[jj] = bestfit_color(pal2, pal1[jj].r, pal1[jj].g, pal1[jj].b);
-      }
-    }
-
-    if (keep_transparent) {
-      // keep transparency
-      color_mapped_table[0] = 0;
-      // any other pixels which are being mapped to 0, map to 16 instead
```
ericoporto, line 151: ok, it was from here... Wonder if this means any other index can't be transparent?

ericoporto: @ivan-mogilko the for here starts at 1. (I think because of the magical remmaping)

Common/gfx/bitmapdata.cpp
```diff
@@ -262,7 +320,100 @@ void CopySwapRGBA(const uint8_t *src_buffer, int src_r_shift, int src_g_shift, i
...
+            if (geta32(*px) <= alpha_threshold)
+                *px = MASK_COLOR_32;
+    }
+}
+
 } // namespace PixelOperations
 
+namespace PaletteOp
+{
+    void Rotate(PALETTE pal, uint8_t first, uint8_t last, bool forward)
```
ericoporto: I read this and the original `wcolrotate` a gazillion of times and still do not trust they do the same thing. I really would keep the two original branches instead of this swap.

ericoporto: Also in the for, in the original the stop condition switched from `< finish` to `>= start`. I am not sure that stop condition that is fixed is doing the same if things are swapped.

ericoporto:
The refactor of things that happen along here is working on original code that was really hard to get already. I think I will need later to just run an import of varied things here, and get an equivalent version before this PR of the Editor, and do the same import and then compare the generated acsprset.spr files, one with the other, and see if they match. If they do match, then we are good and done - I will try this tomorrow though.

Common/gfx/bitmapdata.cpp
ivan-mogilko: I will add tests to assert that the function works as expected.

Common/gfx/bitmapdata.cpp
ivan-mogilko: In Allegro palette index 0 is a mask color. Idk if it's also allegro or just AGS thing, but AGS palette had 0 and 16 as black colors by default. That's why it maps to 16 here (they are sort of equivalent).

ivan-mogilko:
Curiously enough, the latest changes fixed the problem with GHA tests, no idea what it was. Could be there was something off with how `std::equal` is implemented in MSVC libraries.

Common/gfx/bitmapdata.cpp
ericoporto: I noticed the old function the “for”  started from 1

Common/gfx/bitmapdata.cpp
ericoporto: Thanks!

Common/gfx/bitmapdata.cpp
ericoporto: Ah, you adjusted, sorry

Common/CMakeLists.txt
```diff
@@ -251,6 +251,7 @@ if(AGS_TESTS)
         test/inifile_test.cpp
         test/math_test.cpp
         test/memory_test.cpp
+        test/paletteop_test.cpp
```
ericoporto, line 254: The file name is missing a “s”, it is tests, this is why every CMake based build that builds this is failing

ivan-mogilko: It was wrong other way around, the other files were without 's', but i created new one with 's'.

ivan-mogilko:
I found there are some mistakes,
* certain images fail to import (probably if they have very small size)
* it seems like imported image sizes have a 1 pixel difference compared to the project specs (maybe these two mistakes are related).
* when building sprite pack with spriteimport, the sprites may be saved under wrong numbers.

Also, i forgot to log out a list of imported / failed sprites.

ivan-mogilko:
Found a mistake in BufferedStream which could cause errors in Read/Write stream mode (we don't use that much, so missed it earlier).

But then also found there's a wrong test in game tests, where it tried to read from a file which is opened in Write mode. And that "worked" only by accident, because of this incorrect behavior in BufferedStream basically let it read data back from a buffer prepared for writing (https://github.com/adventuregamestudio/ags-test-games/commit/cfff9324fcefdd2162a122a98e870a2472bb5877).

~Which also makes me realize, we might have a use for the new ReadWrite mode for file api in script.~
EDIT: ok, thinking more, we probably should not have this in script, because it's a non-trivial mode and may confuse people more than help.

ericoporto:
The ReadWrite thing, if ever implemented, has to consider it will make the reading be limited to the same directories the Write is limited to ([see docs](https://adventuregamestudio.github.io/ags-manual/v3/File.html#fileopen)).

ivan-mogilko:
Alright, I think i'll merge this now. This PR has only a tiny effect on AGS.Native, and I've already split the 3.6.3 version into the separate release-3.6.3 branch, so it won't affect that either.

PR #3036, opened 2026-08-09
Editor: fix crash in Recent Games when a long game name exists

### LINKED ISSUE #3027: Editor: possible exception when parsing recent games list
User reported following exception occuring when starting the editor:

```
An unexpected error occurred trying to start up the AGS Editor. Please consult the details below and post the error to the AGS Technical Forum.

System.ArgumentOutOfRangeException: startIndex cannot be larger than length of string.
Parameter name: startIndex
   at System.String.Substring(Int32 startIndex, Int32 length)
   at AGS.Editor.Components.FileCommandsComponent.GetRecentGameMenuText(RecentGame game, Int32 maxPrintSize)
   at AGS.Editor.Components.FileCommandsComponent.GetRecentGamesSubcommands()
   at AGS.Editor.Components.FileCommandsComponent.GetOpenRecentMenuCommand()
   at AGS.Editor.Components.FileCommandsComponent..ctor(GUIController guiController, AGSEditor agsEditor)
   at AGS.Editor.ApplicationController.CreateComponents()
   at AGS.Editor.ApplicationController..ctor(String[] commandLineArguments)
   at AGS.Editor.Program.startupTimer_Tick(Object sender, EventArgs e)
   at System.Windows.Forms.Timer.OnTick(EventArgs e)
   at System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
---------------------------
```

Presumably this has something to do with either game names, and/or game folder names. Unfortunately, user did not provide exact details yet (will update if more is known).

ericoporto:
fix #3027

There is no length limit on the Game Name field, so really long game names caused a crash when maxPrintPathSize became negative.

Instead of just clamping, the idea here is to have minimal path size too, and balancing both path size and the game name size, leaving a bit more characters to the game name.

I tested this with a few different game name sizes and then resorted to directly editing the user.config file in the `AppData/Local/AGS/.../3.6.1.12` directory before loading the Editor. It seems to work, and produce results in the recent games menu that look alright while I hope they are still comprehensible by the game author (they can still read enough to find/understand which are their games).

image

### DIFF
--- a/Editor/AGS.Editor/Components/FileCommandsComponent.cs
+++ b/Editor/AGS.Editor/Components/FileCommandsComponent.cs
@@ -112,19 +112,39 @@ namespace AGS.Editor.Components
             return name + "     (" + path + ")";
         }
 
-        private string GetRecentGameMenuText(Preferences.RecentGame game, int maxPrintSize)
+        private string GetRecentGameMenuText(Preferences.RecentGame game, int maxPrintSize, int minPrintPathSize)
         {
-            int maxPrintPathSize = maxPrintSize - GetRecentGameText(game.Name, string.Empty).Length;
+            if (minPrintPathSize < 0 )
+            {
+                throw new ArgumentOutOfRangeException("minPrintPathSize cannot be negative");
+            }
+            if (maxPrintSize < 0 )
+            {
+                throw new ArgumentOutOfRangeException("maxPrintSize cannot be negative");
+            }
+
+            string gameName = game.Name;
+            int maxPrintPathSize = maxPrintSize - GetRecentGameText(gameName, string.Empty).Length;
+            if (maxPrintPathSize < minPrintPathSize)
+            {
+                // it looks like the game name is too big, let's truncate the game name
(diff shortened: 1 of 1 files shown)


Editor/AGS.Editor/Components/FileCommandsComponent.cs
```diff
@@ -112,19 +112,30 @@ private string GetRecentGameText(string name, string path)
             return name + "     (" + path + ")";
         }
 
-        private string GetRecentGameMenuText(Preferences.RecentGame game, int maxPrintSize)
+        private string GetRecentGameMenuText(Preferences.RecentGame game, int maxPrintSize, int minPrintPathSize)
```
ivan-mogilko, line 115: Strictly speaking, you need to either clamp maxPrintSize and minPrintPathSize to the minimal value of zero here, or throw in case it is negative.

ericoporto: Add one check for each so they throw in case either is negative.

PR #3039, opened 2026-08-09
Editor: remember the time of rebuild all command, don't rebuild newer rooms

### LINKED ISSUE #2357: Editor: reimplement WorkspaceState.RequiresRebuild in a user-friendly way
Since certain version Editor marks an imported old project as "requires full rebuild". This mark is serialized in workspace settings, and is only reset on complete successful compilation. This condition is problematic, because the project may contain numerous issues that have to be resolved after upgrade, but this mark will force Editor to rebuild every single room each time. This makes step by step fix of the project rather tedious, as you have to wait when all rooms are recompiled again each time (and might seem confusing to users as well).

My proposal is to invent a more user friendly method. For instance, Editor could save a list of things needed to be rebuilt instead (rooms etc) and remove items from this list as they are built successfully at least once.

I believe this has to be done both in ags4 and ags3.

ivan-mogilko:
Fix #2357

When the project is upgraded from the last version it's marked for the full rebuild, so that all scripts take newer features into account. Because project may consist of multiple rooms, we need to let editor skip those that have already been built in the new version. For this - remember the time when the rebuild order is set, and test it against each room's compiled file.

Do the same thing when:
- Switched Debug/Release configuration
- User ordered "Rebuild all..."

### DIFF
--- a/Editor/AGS.Editor/AGSEditor.cs
+++ b/Editor/AGS.Editor/AGSEditor.cs
@@ -1271,11 +1271,13 @@ namespace AGS.Editor
 
             Utilities.EnsureStandardSubFoldersExist();
 
+            forceRebuild |= NeedsRebuildForDebugMode();
             forceRebuild |= _game.WorkspaceState.RequiresRebuild;
+            DateTime? requiredRebuildTime = forceRebuild ? (DateTime?)_game.WorkspaceState.RequiredRebuildTime : null;
 
             if (PreCompileGame != null)
             {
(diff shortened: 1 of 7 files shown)


ericoporto:
Hey, I have a weird issue. I edited an old game with the current master, and then restored it from backup, back to it's old version.

Now when trying to open it with the previous version of ags I get the message below

image

Where is this WorkspaceState saved?

Edit: ok, figured it was the Game.agf.user file, deleted and all good now!

PR #3041, opened 2026-08-11
Installer: fix VC++ redist detection

### LINKED ISSUE #2759: VC redistributable install is always asked by installer
**Describe the bug**
I am not sure, but I believe that the `.exe` installer used to ask to install the VC redistributable only if it wasn't installed

**AGS Version**
In current AGS 3.6.2 installs and AGS 4 too.

**To Reproduce**
1. Run the installer
2. Click next until the Select Additional Tasks screen come up

Image

3. Notice the checkbox for the installation of the VC Redistributable is there and checked.

**Expected behavior**
I believe when detected, this checkbox should not appear.

**Desktop:**
 - OS: Windows
 - Version: 10

ericoporto:
Fix #2759 

Broken in https://github.com/adventuregamestudio/ags/commit/5018777e0823c672ef8364d4c768aad5b1b4edcd. Turns out the version in the registry key (the "path") is a fixed value.
Removed the constant VCPP_REDIST_MAJOR_VERSION because last time I erroneuously updated it to reflect reality.

If you are looking through the registry editor the path there is HKLM\SOFTWARE\WOW6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\ because you will be viewing it from 64-bit application. Our installer is 32-bit.

### DIFF
--- a/Windows/Installer/ags.iss
+++ b/Windows/Installer/ags.iss
@@ -204,9 +204,10 @@ const
   // Platform check
   PLATFORM_CHECK_ERROR_MESSAGE = 'This program is only supported on Windows Vista or newer.';
 
-  // Visual C++ runtime
+  // Visual C++ runtime 14.42
   // https://download.visualstudio.microsoft.com/download/pr/5319f718-2a84-4aff-86be-8dbdefd92ca1/DD1A8BE03398367745A87A5E35BEBDAB00FDAD080CF42AF0C3F20802D08C25D4/VC_redist.x86.exe
-  VCPP_REDIST_MAJOR_VERSION = 14.42;
+  // Registry Key ("path"-like) is fixed at "14.0" for Visual C++ 2015-2022 runtimes
+  VCPP_REDIST_MAJOR_VERSION = 14;
   VCPP_REDIST_BUILD_VERSION = 34433;
 
   // .NET Framework 4.6 or newer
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I dont think it's correct to remove the constant though, the use of a constant was a right thing, it worked as a "parameter" to a function VCRedistInstalled, which purpose is to test any arbitrary vc redist.

The [previous fix](https://github.com/adventuregamestudio/ags/commit/5018777e0823c672ef8364d4c768aad5b1b4edcd) was incorrectly changing the "major" version number to "major + minor" ("42" is a minor version, not part of a major version). I suggest to keep the constant and move the new comment up to the constant declaration.

ericoporto:
Is still needs a better name then. The Bld is an Integer that is the VC Redist Build Version, and the Major property also exists as an Integer. It should be like 14 with something like %d.0 as the format so that it would better match what it really is.

This took me a lot of time to figure out because of the other thing, which is the registry key (the things that looks like a path) is actually a different one if you are checking with the registry editor because of the Wow64 stuff. So if you just load it, and try to go to that specific X86 key, you wouldn’t find it.

This is because the inno installer is always 32-bit, it only becomes 64-bit if you put a `SetupArchitecture=x64` or 64 mode if you use `ArchitecturesInstallIn64BitMode=x64compatible` entry under its `[Setup]` section. [From the docs](https://jrsoftware.org/ishelp/index.php?topic=32vs64bitinstalls).

ivan-mogilko:
It seems that existing constant's name (`VCPP_REDIST_MAJOR_VERSION`) would be a proper name if you change it to "14" and formatting string to "%d.0", because in this case constant's value 14 will match the value of the "Major" registry entry.

ericoporto:
OK, I made the change, that still uses the format string. I know this is a lot of comments, but we only change this code very few times and it is a lot of details to remember.

PR #3042, opened 2026-08-11
4.0: GetLocationType returns eLocationInvItem 

ivan-mogilko:
Made this a PR, because I'd like to know if there exist any counter-arguments for this.

Recently i've added `eLocationInvItem` to use in `unhandled_event`, and a user asked me why it's not returned from `GetLocationType` too. I cannot see a reason not to really. Because `GetLocationName` (now it's called `GetDisplayNameAt`) returns inventory item names, so why not also return the type? It's difficult to imagine a scenario in which a user would like GetLocationType to skip through the inventory gui and return some room object below it (and then, there's always GetAtScreenXY, and other tricks).

The PR contains couple of refactor commits which are also suitable for backporting to ags3.

### DIFF
--- a/Engine/ac/game.cpp
+++ b/Engine/ac/game.cpp
@@ -182,9 +182,6 @@ unsigned int load_new_game = 0;
 int load_new_game_restore = -1;
 SaveCmpSelection load_new_game_restore_cmp = kSaveCmp_All;
 
-// TODO: refactor these global vars into function arguments
-int getloctype_index = 0, getloctype_throughgui = 0;
-
 //=============================================================================
 // Audio
 //=============================================================================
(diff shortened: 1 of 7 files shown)


Engine/ac/game.cpp
```diff
@@ -1663,9 +1664,17 @@ int GetLocationTypeImpl(int *locobj_index, int x, int y, bool click_through_gui,
...
+    if (!click_through_gui)
     {
-        return LOCTYPE_NOTHING;
+        int gui_index = GetGUIAt(x, y, kHit_Interactable);
+        if (gui_index >= 0)
+        {
+            // On GUI, test if we're above an inventory item
+            int invitem_index = GetInvAt(x, y, kHit_Interactable, kHit_Interactable);
+            if (invitem_index >= 0)
+                return LOCTYPE_INVITEM;
```
ericoporto, line 1678: I think you need to set locobj_index to invitem_index, from looking into `GetLocationNameAndRef` in `global_game.cpp`

ericoporto:
This looks alright and good to merge.

I don’t know if in ags3 the inventory being over in some old AGS game could have a backwards compatibility issue but perhaps for the case there is, if it happens we will know and then can gate in some version check there?

ivan-mogilko:
> I don’t know if in ags3 the inventory being over in some old AGS game could have a backwards compatibility issue but perhaps for the case there is, if it happens we will know and then can gate in some version check there?

I am not planning to port the inventory "location type" to ags3, I was only thinking to backport the code refactor which is not supposed to change the behavior.

PR #3051, opened 2026-08-18
Solutions, Editor, Native: adjust output of all MSVC, MSBuild projects (vcxproj and csproj) 

### LINKED ISSUE #3048: Revise x64 build output directories for MSVC Solution projects
Win32/x86 builds are incompatible with x64 builds when using the MSVC projects, which is fine, so there may be an interest in keeping the OutDir for things like intermediary libraries (Compiler.lib, Common.lib, ...) and the final exe files (acwin.exe, agscc.exe, ...) different from the directory of the x86 targets.

If this is indeed the case, then the actual OutDir should be harmonized between all the vcxproj (and possibly csproj too?) files so they are produced in whatever directory they are intended to go. Currently, while all Win32/x86 seems to be well resolved and the produced results go to the directories like `.build/Debug`, `.build/Release` or `.lib/Debug` and so on, the x64 builds are going to different places. If we want, we could decide to put then in `.lib/Debug/x64` or `.build/Debug/x64` or whatever place we want that makes sense, but it would be nice to have it organized in a way that is standard across all VS projects.

Essentially, this issue is about a design decision of where to put things (we can discuss here) and once a decision is made, it can be closed by revising these VS projects (Tools, Compiler, Editor, Engine) and ensuring they all agree for their x64 builds.

**Additional context**

I added a job for building the Tools and the standalone Compiler in https://github.com/adventuregamestudio/ags/commit/8cca60267483bf5d2f00319399c6d7b4c4c479fa, but I had not checked before the x64 produced binaries and had to wonder why agscc and agfexport were not there along like they are in the x86 artifacts, turns out things are not as well defined for the x64 builds.

ericoporto:
fix #3048 

Leaving this as draft because I still need to check all the built artifacts in the CI, plus this approach is very maximalist, touching on a lot of stuff. For the vcxproj, csproj and MSBuild related solutions the changes are below:

- All outputs go to Solutions/.build/$Platform/$Configuration, like Solutions/.build/x86/Release or Solutions/.build/x64/Debug
- The same logic is applied to .lib, .obj and .test directories, all build results are now placed in the Solutions directory in the root of the project (Including AGS.Native!) `*`
- I modified the csproj files to use either x86 or x64 (sometimes x86 was missing, so I added), and their solutions were adjusted too, to use either, this avoids creating things in a directory named `AnyCPU`.
- A new props file was created to centralize the variables used in these project files. While this makes things easier to make the same, props files makes harder to modify settings through the VS ui, so there is some loss in flexibility here.
- AGS.Native for some reason has a `_MD` ending directory and I preserved that.
- CI has been adjusted for the new paths

**NOTE:** 🚨 I used Claude Code (Sonnet 5) and Chat GPT Codex (GPT-5.6-Luna) heavily to help me review, bulk modify and all. I did review things, but because it is a lot of lines this has to be reviewed with more attention.🚨

`*` **NOTE2:** I didn't modify the plugins MSVC projects. They are not hard to adjust too, if we want to do changes on them, let me know.


ivan-mogilko:
Interesting solution with using .props file for defining output directories, this makes this setup cleaner.
By the looks of it, the changes are okay, this is mostly one string replaced with another.

One thing I wonder, you've replaced "AnyCPU" config with "x86", but did not remove "AnyCPU" config from the project files, is this intentional, or a oversight?

EDIT:
> props files makes harder to modify settings through the VS ui, so there is some loss in flexibility here.

Hmm, I expected this props file appear in the Property Manager list, which would allow to edit it directly from MSVS, but for some reason it is not listed there.

ericoporto:
> Hmm, I expected this props file appear in the Property Manager list, which would allow to edit it directly from MSVS, but for some reason it is not listed there.

Does the sdl2 one appears there for you? I may have done something wrong with it. (I either don’t see the sdl2 or am looking in the wrong place)

btw there’s a minimal alternative to this which is to only manually fix the x64 stuff and leave everything else as is

> One thing I wonder, you've replaced "AnyCPU" config with "x86", but did not remove "AnyCPU" config from the project files, is this intentional, or an oversight?

Intentional oversight in the sense I wanted to remove but was unsure if I could, if it is indeed alright I will remove it. I can not do the AnyCPU->x86 change that but then I need to hardcode or use some workaround for it to not generate things in a AnyCPU directory.

ivan-mogilko:
> Does the sdl2 one appears there for you? I may have done something wrong with it. (I either don’t see the sdl2 or am looking in the wrong place)

Property Manager lists following items for Engine project (related to the prop files):
- libsdl2
- libvorbis
- libtheora
- libogg

> I can not do the AnyCPU->x86 change that but then I need to hardcode or use some workaround for it to not generate things in a AnyCPU directory.

I'm sorry, but I cannot understand this sentence.

ericoporto:
> > I can not do the AnyCPU->x86 change that but then I need to hardcode or use some workaround for it to not generate things in a AnyCPU directory.

> I'm sorry, but I cannot understand this sentence.

It’s alright. What I mean is the reason I would like to discontinue using the AnyCPU is because it as a platform will with the current variable (`$Platform`) be used to name the output directory as `AnyCPU\`. If we want to maintain the AnyCPU config use then what I can do is not use the variables for those C# projects, and instead hardcode the output directory (to x86).

Because the editor has native parts (like AGS.Native but also Magick, and irrklang) we do actually have to be specific on the editor being x86 or x64 (it currently is x86). So it doesn’t seem that useful to have the AnyCPU config.

ericoporto:
@ivan-mogilko I did only a test, I discovered that when I add the props through the interface it copies once for each config too, because props are per config, can you check for the engine project if it is working for you? I added for every project. The csproj (C#) don't have this feature visible in the VS interface, so for them nothing changes.

Also I guess I was supposed to do the sdl2.props manually too somehow, because it copied the sdl var to vcxproj.filters.

Still, this doesn't appear to change anything functionality wise (it was already working before).

The actual content of the prop file doesn't appear to be possible to edit in the VS IDE.

ericoporto:
It looks like this is working, I can squash all the commits in a single one if necessary.

ivan-mogilko:
I can see these AGS.OutputPaths props in the property manager of MSVS now.
But I also found that likely you dont need to use these special constants `$(AgsLibOutDir)` and `$(AgsObjDir)`. Instead you could do this in AGS.OutputPaths.props:
```
<PropertyGroup>
    <OutDir>$(SolutionDir).build\$(PlatformFolderName)\$(Configuration)</OutDir>
</PropertyGroup>
```
(and same of intermediate dir, etc)
The projects then set their output directories to "Inherit from parent or project defaults". This way you also see how this string is formed in each project settings, instead of a custom constant.

The main difference will be that it will use `SolutionDir` and not `AgsSolutionsDir`. I could not tell, is there a reason to have a `AgsSolutionsDir`, does it point to another directory in certain cases?

ivan-mogilko:
On a side note,
> AGS.Native for some reason has a _MD ending directory and I preserved that.

Common is built in two pairs of configurations: with MT(d) and MD(d) flags respectively. AGS.Native uses MD ones while engine uses MT ones.

ericoporto:
> The main difference will be that it will use SolutionDir and not AgsSolutionsDir. I could not tell, is there a reason to have a AgsSolutionsDir, does it point to another directory in certain cases?

I haven’t touched on them, but if we want to do the same with the plugins their solutions are the only ones that aren’t placed in the Solutions directory.

ericoporto:
> Instead you could do this in AGS.OutputPaths.props

Right, but this forces the exact flat structure to everyone, which is what we have today and is fine, but if the acwin.exe is at some point meant to be under a subdirectory then we won’t be able to be specific for it if we want - or have the tools go into a Tools directory.

ivan-mogilko:
Hmm... i see.

ivan-mogilko:
Sorry, I did not realize this earlier, but i've merged #3030 right now, so spriteimport tool will require this fix to dirs too.

ericoporto:
That is easy to do, no worries. This project/build system stuff is more trouble for design decisions, but making changes is easy, it’s only annoying for the users of it (mainly you in this case, I am more an user of CMake) because it impacts the workflow.

ivan-mogilko:
There's something different in C# projects, they have 3 platform setting now:
* Win32
* x86
* x64

Where "x86" platform targets x86 while "Win32" platform targets "AnyCPU".

I noticed this when browsing "Batch Build" menu, it gives a overview of every configuration combination we have.

EDIT: also Win32 platform has different build settings compared to their x86/x64 counterparts. For instance, Debug/Win32 platform configuration does not have DEBUG constant set.

ericoporto:
Hey, I checked the following csproj files: AGS.Editor.csproj, AGS.Types.csproj, AGS.CScript.Compiler.csproj, AGS.Controls.csproj and AGS.Editor.Tests.csproj. They all only have Debug/Release for x86 and x64, and they all properly define the DEBUG constant mentioned.

I found the mentioned screen

image

But I don't understand yet why Win32 shows there.

I also checked the AGS.Editor.Full.sln, and the only uses of Win32 there in the right side of the config is in the vcxproj files, which is correct, the csproj files all use the x86 naming.

> EDIT: also Win32 platform has different build settings compared to their x86/x64 counterparts. For instance, Debug/Win32 platform configuration does not have DEBUG constant set.

Where did you find this? Is there a project specifically you looked into?

The only solution I can think for this would be to duplicate the x86 propertygroup and stuff and then use it for Win32.

Edit: this is utterly bizarre, but I went into the configuration manager and for each config of a csproj file I clicked edit... and clicked Win32 and then selected remove and then they are gone from the Batch Build

image

But I don't have any actual git diff for this as an actual change (like it inverted the order of the x86 and x64 stuff in the csproj files mentioned in the `.sln` file, but nothing ACTUALLY changed, there is no Win32 that either appeared or disappeared there.

Edit2: perhaps the line in the AGS.OutputPaths.props file is injecting the Win32 possibility in the csproj file? I need to review this with calm and perhaps move this line into the vcxproj files individually.

https://github.com/adventuregamestudio/ags/pull/3051/changes#diff-b60f4909c07be3f314bbadd6f00f7b880c6c9be60026dcfbe74d3f25b4677ff6R27

ivan-mogilko:
There's a problem that I did not notice earlier. All the project files now have the reference to AGS.OutputPaths.props, but I cannot find where it's configured in MSVS. This possibly means that every time someone creates a new project, they must know about adding this reference and all the associated changes, and do that by editing vcxproj file by hand.

This must be documented in our project wiki.

ericoporto:
The props itself I added manually in the property sheet menu in VS (I think it is called Property Manager?). The use of the variables I did editing the vcxproj file by hand, I don’t know if it is possible to write the variables in the VS property window for the project (the regular property window, not the property manager)

PR #3052, opened 2026-08-18
agfexport: add --no-empty option that prevents writing of empty files

ivan-mogilko:
For agfexport tool add `--no-empty` option that prevents writing of empty files.

The primary use of this option is to avoid adding redundant script files into the build process, which matches the Editor's behavior in regards to global variables script at least. But in general, it may be used to detect that some type of content is not available by testing if a output file exists. (The operation error may still be detected by the program's exit code.)

### DIFF
--- a/Tools/agfexport/main.cpp
+++ b/Tools/agfexport/main.cpp
@@ -49,6 +49,7 @@ const char *HELP_AUDIOCLIP_LIST = ""
 #endif
     "Commands:\n"
     "  -h, --help             Show this help message\n"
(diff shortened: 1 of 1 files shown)


ericoporto:
About global variables, if no global variables exists the script and the header are both empty? Which would mean both are not generated.

ivan-mogilko:
> About global variables, if no global variables exists the script and the header are both empty? Which would mean both are not generated.

Yes that's what happens, and with this option neither file will be written.

PR #3057, opened 2026-08-21
agsbuild, a tool running other tools and building game.ags

ivan-mogilko:
### Overview

"agsbuild" is a tool meant for compiling game data from the project source. It's input is a game source project, and output is game.ags, acsetup.cfg + optionally audio.vox and speech.vox packs; basically what you have in Compiled/Data when you build in the Editor.

Compared to the another proposed tool "agfbuildgen" (#2770) which purpose is to generate "makefiles" for the project, "agsbuild" is more simple and straightforward. It executes the build itself, while calling other existing tools (agfexport, crmpak, script compiler, etc) for performing individual tasks.

I do not know which tool will be used to replace the Editor's own compilation code in the end. Both may have their uses. I think the advantage of agsbuild is that it's simple and does not require any additional steps from user, you just run it with some parameters and it does the thing. The disadvantage is that it has most actions hardcoded, so it will be more difficult (if at all possible) to fine-tune and customize the individual steps in build process.

### Idea and implementation

"agsbuild" is a director tool, meaning it does a minimum of work itself, and delegates most work to other tools from our collection.

Originally my intent was for agsbuild to not read or parse any data itself at all, that was the "ideal solution" in my mind. Idea was that agfexport will be used to export all necessary information about the game settings and game files, which can be used to configure other tools. However, I had to diverge from this plan, for several reasons. First, agfexport still does not export everything necessary; for instance it does not export script compilation settings, or game packaging settings. Second, as of now, agfexport exports only one thing at a time, which makes it suboptimal, as you need to run it about 10 times with different arguments to get all necessary file lists. It may not be a big deal, but is annoying. Third, as I was developing the task system, i found that writing one that can actually dynamically process the agfexport's output data, and dynamically generate sub-tasks too based on that data, would take a while, and i started having doubts about whether it's worth to do at this moment.

So, I chose a simpler path for the time being, and made agsbuild parse Game.agf for game settings and file lists, and then setup build tasks directly, based on these settings and lists. Majority of work is still performed by other tools as planned.

The implementation is built around the TaskManager class that operates the array of task threads. The number of threads is limited by the tool's setup. Each planned task has a state (Pending, Running, Success, Failure or Cancelled) and a number of "inputs". The input defines the previous task, and data received from that task (latter is not used in the current implementation, but is reserved for the future use).

After being started, the TaskManager continuously checks for free thread slots, and checks the pending tasks to see if there are any ready to be run. A task is ready when it either does not have any prerequisites, or all of its prerequisites (inputs) have completed with success. In either of these cases, and if there's a free thread slot, TaskManager would run the task on a thread. When the task is finished, the thread slot gets freed, and next tasks are possibly run. If the task is finished with a Failure, then all the dependent tasks are automatically Cancelled and removed from the pending list. The whole process is considered success if all the tasks have completed with success. However, the TaskManager does not stop immediately if any task has failed, but tries to complete as many tasks as possible.

Task objects are referenced via the abstract Task base class, but each kind of task has its own subclass that does the actual job. Most interesting is a TaskPipedProcess class that spawns child process and connects to its stdout and stderr through pipes, reading out their output and pasting it into their "log buffer". Each task has a "log buffer" in memory to where it prints all their log. The accumulated contents of this log buffer is printed to program's own stdout only after the Task is finished, in one go. This is done so because tasks are run on multiple threads, and having them all printing their logs simultaneously could result in a unreadable mess.

### Usage

agsbuild is used as:

```agsbuild <input-game.agf> <out-dir> [<OPTIONS>]```

The "out-dir" is a final directory to deplot the game.ags (and all accompanying files) to.

The "temp-dir" is a intermediate output directory, where all the temporary files are stored. These may be files generated by tools, and copies of the project files, if there's a need to modify them during the compilation for some reason. The location of this directory may be provided with options (see below), or it's deduced as "out-dir/_temp".

The location of other tools may be provided with options (see below), or they are looked up in the same directory where the agsbuild exe is found.

agsbuild requires the script API header ("agsdefns.sh") to be present. It's location may be provided with options (see below), otherwise it's assumed to be present in the same directory as tools.

Options are:
```
  --api-ash <path> - a path to the script API header. If not provided, agsbuild
      will look for the 'agsdefns.sh' file inside the system (tools) directory.
  --build-debug - builds the project in debug mode (override).
  --build-release - builds the project in release mode (override).
  -c, --cleanup - deletes intermediate output on successful completion.
  --import-audio - generate a new audio cache, importing audio files from the
      source files, according to game.agf specs.
  --import-sprites - create a new sprite pack, importing sprites from the
      source image files, according to game.agf specs.
	  Requires a presence of spriteimport tool.
  -s, --sys-dir <path> - a path to the system directory, where all the tools
      are located. If not provided, agsbuild will assume it is the same
	  directory where it's located.
  -t, --temp-dir <path> - a path to the intermediate (temporary) outout
      directory. If not provided, agsbuild will try creating one as a
	  subdirectory inside output directory.
  -p, --threads <N> - limit task threads to this number (default: 8);
      note that tasks may spawn new processes, executing other tools.
  -v, --verbose - print operation details (also affects spawned processes)
```

`--import-audio` and `--import-sprites` is useful if you are building from game's repository, and either sprite file or audio cache is not built.

agsbuild will also automatically import (copy) any missing audio files (they will be imported into the temp dir though).
It does not do anything in case of missing sprites at the moment, although I suppose such task may be added.

### Current state and TODO

The tool seem to be generally working in the most default case.

Missing functionality:
- [DONE] Translations (compile and add to the final output).
- [DONE] Multiple speech vox support (only does the main speech.vox atm).
- Compile lipsync data, need a separate tool (#3065)
- Game plugins are neglected, but we need a separate tool for extracting their script headers (#2771);
also, plugins themselves probably should not be copied by the standard build process, as different platform targets require different plugin binaries, which are located in different places. This is a whole separate story.
- [DONE] Script compilation settings, that need to be passed as flags to the script compiler (backwards switches, api levels, etc).
- [DONE, except partitioning] Game packaging settings, such as sprite compression, data file partitioning, etc.
- Support copying acsetup.cfg provided by user in the source dir.

### CODE REVIEW

**Comment by ericoporto** [comment]
I think the main advantage here vs my approach is that both make and ninja are timestamp based when selecting what needs and what doesn’t need to be rebuilt, which is an issue because Game.agf produces the Global Variables header that is added in everything, and I couldn’t yet figure out how to only build what is changed (other than using a modified ninja that is hash based instead).

I think this isn’t addressed here, and I am not saying it has to be in this PR, but it is something that in theory could be worked on using the approach you introduced here.

> So, I chose a simpler path for the time being, and made agsbuild parse Game.agf for game settings and file lists, and then setup build tasks directly, based on these settings and lists. Majority of work is still performed by other tools as planned

I had the same realization and think it is fine. The main advantage is the parallelism of the task you are using the task manager to run here.

> agsbuild searches for the tools in the "current working dir" at the moment. This is likely a mistake, and it should search in the directory of the executable instead. Also maybe have a explicit option in command line.

About the other tools, one issue is finding them. I have a few ideas (they do NOT have to all be used or any at all).

- have it look them automatically in the directory agsbuild is.
- have an environment variable for AGS_TOOLS_DIR or alternatively individual ones per tool if one must (I think the only one someone may find useful for this is the compiler specifically, if someone wants to replace the compiler by other). Environment variables also can be easily set by a script that configures the terminal itself.
- have a command line option for the tools dir or per tool, these may be less convenient than environment variables (you have to pass them all the time if you need), but may be easier to work with in an environment you are unfamiliar with (say helping debug someone else’s build process).
- and finally, have a command line option that doesn’t run the tool at all, just reports it can successfully find all the required tools, perhaps also printing their paths.

again, they do not have to be worked in this PR but this is something I think I would like to add later to this tool.

**Comment by ericoporto** [comment]
Wait, why I can’t use ; in Linux for header separation?

**Comment by ivan-mogilko** [comment]
> Wait, why I can’t use ; in Linux for header separation?

So, I just found that sh splits commands by semi-colon. Meaning, everything past the semi-colon is treated as another command. This happens both when running agscc from terminal, and when spawning "/bin/sh" process programmatically.

https://stackoverflow.com/questions/20233934/semicolon-on-command-line-in-linux
EDIT: there are two suggestion there, either escape or put in quotes, I did not realize to try these.

I meant to open a issue for this, but maybe I just ask here, why is the colon and semi-colon chosen as separators, and not e.g. comma, that would have same meaning on both windows and unix systems?

**Comment by ericoporto** [comment]
> I meant to open a issue for this, but maybe I just ask here, why is the colon and semi-colon chosen as separators, and not e.g. comma, that would have same meaning on both windows and unix systems?

I didn’t think things through and copied ascc ideas when possible (I wrote this in the original standalone compiler PR, mentioning this because I was now wondering wth I was thinking when I wrote it that way too…). I checked clang and it uses comma. I think comma is a much saner option, I believe we should change it to use commas instead as separator for whatever needed in the agscc (and any other command line tool that requires this actually).

The only other thing that uses these that I can remember is PATH environment variables? I couldn’t find info on what GCC prefers, but my guess is in gcc it depends on what is being separated.

**Comment by ivan-mogilko** [comment]
This is now ready for review. It's generally working, tested on Windows and Linux.
Adjusted command line arguments.
There's still few functionalities missing (see the end of the PR post), but that should be relatively easy to add.

Ironically, because this tool spawns multiple processes in parallel (and because it does not load full room data, with backgrounds and masks), the building time is faster than with Editor.

**Comment by ericoporto** [comment]
Hey, you dropped the agscc change, can I pick that up to switch it for commas later?

File: Tools/agsbuild/main.cpp [hunk 1]
```diff
@@ -0,0 +1,683 @@
...
+    // tasks to complete before deploying game files
+    std::vector<TaskInput> pre_deploy_tasks;
+
+    // Export all the necessary information
+    {
+        const String tool_agfexport = Path::ConcatPaths(sys_dir, ToolAgfExport);
+        const String tool_dlgasc = Path::ConcatPaths(sys_dir, ToolDlgAsc);
+
+        taskmgr.AddTask(std::make_unique<TaskPipedProcess>("autoash", &taskmgr,
+            PipedProcessParams(tool_agfexport, String::FromFormat("autoash %s %s/_AutoGenerated.ash %s", src_agf.GetCStr(), temp_dir.GetCStr(), verbose))));
```
ericoporto: Hey this and all the calls that have an argument where there is a path or a free text, the argument needs to be enclosed in `"`, otherwise any space character in a path or elsewhere will turn a parameter in more parameters. Edit: `’` is not supported on Windows from what I looked, so it does need to be "

ivan-mogilko:
> Hey, you dropped the agscc change, can I pick that up to switch it for commas later?

I have replaced the platform switches with a single switch at the top, where it defines the separator macro. It's used similar to PRI64 macro in format strings.

ivan-mogilko:
I've met a problem regarding the "include" pattern file for agspak. Passing a list of files as a "file filter" argument in command line may not be a good thing, because the list may become quite large for the big games. So I tried to replace this with generated include files, and this is where i've met another issue.

Right now it works like this: agspak accepts only the name of the include file, and then searches for this file in every input directory. This means that if i want the tool to generate such include file, then i must generate it in every directory i am packing the files from, including the project source. This may not be a good thing, because
a) this location may be not writeable, or it may simply be not desired to write anything there,
b) there may already be a file with the same name there, by coincidence (low chance perhaps, but still).
These are the reasons why I am inclined to only allow this tool create new files in "output" and "temp" directories.

I understand the current logic may be meant for the case of persistent repository, where these include files are maintained, and also it seems nicer to have a include file per directory responsible only for that directory.
But I wish there were a way to tell agspak to use include file from a different place in the directory. I.e. create N include files somewhere in a temporary dir, and then tell to use include file Af for input directory Ad, include file Bf for input directory Bd, and so forth.
Not fully certain if that's a optimal approach.

While such option does not exist, I can see two ways:
- go back to creating full package structure in a temporary dir using hardlinks;
- create temporary include file in the project source dir and delete when done working (potentially leave the garbage if the program fails).

Perhaps the hardlinks way was a fine approach, and i should have sticked to that, instead of using multiple input dirs way. I am simply uncertain if they will work fine on every platform.

ericoporto:
The include file won’t work because you can’t change the internal “directory” (the $DATA path) it packages to, which is why I proposed the idea of a `;` as separator in the same line to allow a destination/rewritten internal path to be passed. Perhaps there is some other idea for the internal path but I haven’t came up with anything.

> But I wish there were a way to tell agspak to use include file from a different place in the directory. 

I forgot how that was written, but this looks like a fine command line “instead of searching, use the one passed in `--pattern-file FILE`”.

But about the command line being long it should not be a problem. Another alternative could be to allow adding files to an existing package file - similarly how you can add files to an existing zip file.

> I am simply uncertain if they will work fine on every platform

Hardlinks should work fine in Windows, macOS and Linux, their only restriction is the files should be in the same storage unit (from reading in the internet), so since all files are in the project directory including the build it should be fine. I guess for the case of using hardlinks it could test a hardlink and if it fails emit some sort of warning and use actual file copy instead.

ivan-mogilko:
> The include file won’t work because you can’t change the internal “directory” (the $DATA path) it packages to

I dont need to change internal directory, I have multiple input dirs that all map to the same "root" internal directory.

> Another alternative could be to allow adding files to an existing package file - similarly how you can add files to an existing zip file.

Such function may be added (it's currently reserved as "-i" in agspak), but in this case it will slow things down unnecessarily, as the same file will have to be rewritten multiple times, and with big games there may be gigabytes of data.

Right, I can make it fallback to copying if hardlinks dont work.

ericoporto:
Hey, my regular Windows command line is bash through MSYS2 (git-bash or msys2 standalone, both should really be the same, but here I used git-bash). I tried using with it to build the ags3 auto-test game, but it produced errors and then hanged. I tried again using cmd.exe, but it also produce errors (albeit slowly), and then it exit with the errors.

[msys2_bash_error_log.txt](https://github.com/user-attachments/files/31578430/msys2_bash_error_log.txt)
[cmd_exe_error_log.txt](https://github.com/user-attachments/files/31578431/cmd_exe_error_log.txt)

ivan-mogilko:
> Hey, my regular Windows command line is bash through MSYS2 (git-bash or msys2 standalone, both should really be the same, but here I used git-bash). I tried using with it to build the ags3 auto-test game, but it produced errors and then hanged. I tried again using cmd.exe, but it also produce errors (albeit slowly), and then it exit with the errors.
> 
> [msys2_bash_error_log.txt](https://github.com/user-attachments/files/31578430/msys2_bash_error_log.txt) [cmd_exe_error_log.txt](https://github.com/user-attachments/files/31578431/cmd_exe_error_log.txt)

According to CMD logs:
All errors in these logs are agscc errors, most occur because it does not find old api, and one that is strange:
> Error: compile failed at test-string, line 25 : Error (line 25): undefined symbol '-61'

**EDIT:** checking the script, this is caused by a UTF-8 character literal. Possibly because agscc is working in ASCII mode.

The ones about old api I cannot explain at the moment, because according to the agscc own log, the compatibility level is set to v321. Perhaps agscc should be tested individually on these scripts.

---

According to MSYS2 bash logs, many tools cannot find input files. I noticed that you are using relative paths on agsbuild input, in my tests I used absolute so far. Maybe there's something about that... But they seem to work when run from CMD.

ericoporto:
[msys2_bash_error_log2.txt](https://github.com/user-attachments/files/31603576/msys2_bash_error_log2.txt)
[cmd_exe_error_log2.txt](https://github.com/user-attachments/files/31603577/cmd_exe_error_log2.txt)

OK, tried again. It looks like the bash in MSYS2 still errors, but with new errors, but the cmd.exe now actually runs and finishes building! Below is the successful auto test game!

[Compiled.zip](https://github.com/user-attachments/files/31603599/Compiled.zip)

Now actually running the auto-test, it seems they all pass, but it runs 2063 from a plan of only 2042. I am not sure, perhaps the auto-test is incorrect (I have not redownloaded it again since when I downloaded it yesterday. It looks it built correctly.

[agstest.tap.zip](https://github.com/user-attachments/files/31603655/agstest.tap.zip)

From reading the entire output, I think at the end of a successful build it could write a message like `"Successfully built game GAME NAME at:\nfile/path/GAMENAME.ags"`. I think we have at least a minimal message like this in the Editor that happens at the end.

This looks good overall, I guess if we try to wire it to run from the editor (to leverage the faster build time) a few other things could appear regarding the things that get printed, but this is a different issue entirely (probably #2530). I would say let’s merge and then fix the other things later in newer tickets.

Edit: ha, noticed on reading the first commit changes to `file.cpp` that we used to always overwrite on rename on Linux and others, but fail to overwrite on rename on Windows, and now you also fixed this with a more consistent behavior, nice!

ivan-mogilko:
> Edit: ha, noticed on reading the first commit changes to file.cpp that we used to always overwrite on rename on Linux and others, but fail to overwrite on rename on Windows, and now you also fixed this with a more consistent behavior, nice!

This reminds me, I should split these changes as a separate commit, which may be ported into older branches.

ivan-mogilko:
> OK, tried again. It looks like the bash in MSYS2 still errors, but with new errors

All errors are compiler not being able to open "agsdefns.sh". I noticed that this is the only header which path has backwards slashes `\`, all the other paths have forward slashes `/`.

Does not MSYS2 require unix paths? I wonder why the backwards slashes appear in that particular path. Idk if that's related to the problem.

ericoporto:
[msys2_bash_error_log3.txt](https://github.com/user-attachments/files/31619148/msys2_bash_error_log3.txt)
[cmd_exe_error_log3.txt](https://github.com/user-attachments/files/31619151/cmd_exe_error_log3.txt)

OK, I ran both again. The cmd.exe still finishes correctly, but the git bash build produces an error.

> All errors are compiler not being able to open "agsdefns.sh"

I noticed there are some success AND some failure, when the failure is `Error: failed to open header for reading:` , so I ran the git bash again, but passed the `-p 1` command (threads pool)... And it succeeds?
[msys2_bash_error_log_p_is_1.txt](https://github.com/user-attachments/files/31619264/msys2_bash_error_log_p_is_1.txt)

Are we reading in a locking way somehow?

Edit: we have an [`stdio_compat.c`](https://github.com/adventuregamestudio/ags/blob/master/Common/util/stdio_compat.c) but I think we sort of just use it in our file implementation... But in case of an error, the error information could be made available on Windows through [`GetLastError`](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror), which would help to understand things a bit more, but I think perhaps we could just open an issue to look into this with care later and merge things here as is?

Also, I haven't tried this in macOS or Linux yet, the code path there is very different for the managements of the process.

ivan-mogilko:
This is quite strange, as the file is read but never written by this process. I did not think there can be problem in opening same file multiple times for reading. It does not look like we are using file locks anywhere, unless the C library does this itself internally in some case. I'd need to install MSYS2 and investigate this myself.

ericoporto:
I don’t think it is related, so I edited and removed from my comment, but in case there is relevance, here is MS Docs on CreatePipe

https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe

> CreatePipe also creates handles that the process uses to read from and write to the buffer in subsequent calls to the [ReadFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) and [WriteFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile) functions.

I don’t understand what that means, but I am not sure if it means it somehow modifies something on how files are read.

But on the use of GetLastError if it would be possible to get the reason the ReadFile fails, I think at least it would make more sense. I remember at some point you mentioned ReadWrite stream, is the stream only opened as read indeed?

> I'd need to install MSYS2 and investigate this myself

only running from git bash should be sufficient if you have git for windows installed.

edit: also found related code in ninja itself and it is different but somewhat similar

https://github.com/ninja-build/ninja/blob/master/src/subprocess-win32.cc

but I don’t see anything useful to steal from there.

ivan-mogilko:
The handles created by CreatePipe are identifiers of the pipe's ends itself. ReadFile and WriteFile are using the handles you pass into them, which may be handles of a file, pipe, or anything else that implements reading and writing in the system.
Posix code does exactly same, except it uses its own library functions, and declares handles as "ints".

ivan-mogilko:
Rebased and reorganized commits (squashed most agsbuild changes).

ericoporto:
I am good with merging and adjusting things afterwards if needed. At least the first fix can be ported to 3.6.3 too.

ericoporto:
I read a particular interesting blog post today https://holyblackcat.github.io/blog/2026/09/05/escaping-createprocess-arguments.html

I noticed that the actual application path isn’t escaped when calling a tool, so if the tools are in a path where one of the directories in the path have a space, then agsbuild breaks.

Issue #2600, opened 2024-12-02
Editor: setup initial Character's inventory


ivan-mogilko:
This suggestion is opened after a user's question.

Currently AGS only allows to mark inventory items that are added to player character at game start. This is done by setting each item's "PlayerStartsWithItem" property.

Sometimes it may be desired to have several characters begin with some items in their inventories. The character roles is mostly irrelevant, these may be playable characters, or NPCs. This may be done in script, but would be convenient for users to also have a way to set this up in the editor directly.

The idea is to have Inventory property in Character instead. This property would hold a list of inventory items. In Properties Grid this property would have a "..." button, which opens a dialog window with a list, that lets add and remove items, set their quantity and change their order.

This list would be saved in the game data, and used to initialize each character's starting inventory contents.

ericoporto:
This can be done very easily with code generation afaict.

ivan-mogilko:
Then, where would be the line between things initialized by game data and by generated code?
Every new feature added to characters and other global objects may in theory be put into generated code instead of game data.
(Rooms probably may too, although for that they would have to support extra init script)

ericoporto:
Well, it can be done by adding things in game28.dta too.

ivan-mogilko:
I actually would not mind doing this in script, that sounds rather flexible, and especially suitable if we support dynamic creation of objects. But there may be caveats.

We must keep in mind that anything done in script may have side effects, and that may be not good when we are initializing an initial game state, especially in the auto generated code (that user may not be aware of).

For instance, AddInventory function triggers eEventAddInventory event, so it should not be used. There may be other similar cases, and there's no guarantee that something of that sort is not added in the future.
If we want to secure initialization stage, we'd need to have a flag that enables and disables side-effects. There was something similar in WinForms, which wraps the code in xxxDesigner.cs.

---

In regards to inventory though, I just realized that Character data already has inventory array in game data. It was simply never used when compiling the game:
https://github.com/adventuregamestudio/ags/blob/1d01739c895999729623aded7817748c328d2f47/Common/ac/characterinfo.h#L144

Since it's already there, I suppose it's easier to just put this data there.

ivan-mogilko:
Wait, I was an idiot. This array IS used when compiling the game, that's how the "player starts with this item" works:

https://github.com/adventuregamestudio/ags/blob/1d01739c895999729623aded7817748c328d2f47/Editor/AGS.Editor/DataFileWriter.cs#L1650-L1654

[...]

PR #3059, opened 2026-08-21
4.0: remove limit of 300 inventory items, turn item list into 0-based

ivan-mogilko:
Inventory items are unlimited (technically limited by int32 ID or program memory of course).
Turn 1-based list into 0-based list, so that item IDs begin with 0 as proper.

In CharacterInfo replaced a fixed max-sized inventory array with a map.
In data and save formats removed fixed max-sized array of int16, and serialize map of ints instead (this cuts 600 useless bytes from every NPCs in file data). The new inventory is now written in data extension, potentially for each character, which keeps open opportunity to setup starting inventory for NPCs (#2600).

When upgrading project, I need to turn a 1-based inventory list to 0-based one. If the item IDs will be shifted down, that might break the game scripts where they use ID numbers, or for loops going from 1 (although frankly they will still be broken, because Game.InventoryItemCount now includes the 0th item). I could not invent a better solution so just made it insert a dummy item 0, and let user do whatever they want with it.


ivan-mogilko:
I'm a little confused about the game tests, they failed previously, and still fail if i run latest ags4 branch locally, but at the same time all checks are now successful in this PR. In any case, these failed tests were related to the character walking, which is not touched by this PR.

PR #3066, opened 2026-08-28
ci: add tools to release

### LINKED ISSUE #3061: Add Tools to the release pipeline in CI
I need to wire up the pipeline for the assets that should contain the tools in our release pipeline, and make small decisions to these assets.

Now perhaps at some point the tools are used in the Editor too, so they are included in the regular editor packages too, but here I am focusing on a standalone package, imagining someone could download only it in a CI for building - like when MS used to distribute too the C++ build tools that didn’t include the full graphical VS along, or the Android command line tools when they were all in a single package.

First, I imagine we would have an asset per platform (Windows, Linux and macOS).

Regarding the architecture, I think we could perhaps package all in a single package? Not sure on that, but it would be inside of each a directory named like x86, x64 or however Apple names Arm64, and then tools would actually go inside these. I think for Apple it would only be Arm64, for Linux it could be only x64 (and perhaps at some point we could include Arm64 there too). Now for Windows I was thinking to include a x64 dir with all tools and a x86 dir with only the plugin header extraction tool.

For naming I think the packages could be named something like `AGS-Tools-<ags x.y.w.z version>-<platform>.zip` (or `tar.gz` for Linux and something for macOS).

For build system I am thinking about using the MSVC builds for Windows and either Makefiles or CMake for Linux/macOS.

ericoporto:
fix #3061

Initially my idea is to just add the Windows Tools to release. Note that the archive currently has both the x86 and x64 tools, and I am not sure that is the best way to go. I think we can make one version of the tools for release and later change the shape of the archives as we understand better how to work with them. I imagine we may change again if at some point the tools are added along the Editor.

I added also makefile builds of the tools for Linux and macOS, Linux build is also added to the release.

I didn't add the macOS tools to the release since they will require using `xattr -dr com.apple.quarantine *` on them after they are downloaded since I am not signing and notarizing them - afaict as I can tell this would require a paid license, and the license is tied to a specific dev and not a software, so this means changes by someone else in the ags codebase that apple doesn't like could make I and not ags lose the account.

~~Keeping as a draft until #3057  is merged~~

Merged so it is now for review.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -862,7 +862,7 @@ jobs:
 
       - name: Install bsdtar
         run: sudo apt-get update -y && sudo apt-get install -y libarchive-tools
-        
+
       - name: Log in to GHCR
         uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 #v4
         with:
@@ -874,13 +874,13 @@ jobs:
         run: |
           TAG_NAME=${{ env.BASE_BRANCH }}-${{ matrix.platform.arch }}
           IMAGE="ghcr.io/${{ github.repository_owner }}/ags-linux-builder-image:$TAG_NAME"
-  
+
           docker run --rm \
             -v "$PWD:/workspace" \
             -w /workspace \
             "$IMAGE" \
             bash -c 'find Tools -name Makefile -execdir make -j4 PREFIX=/workspace/dist-tools/usr/local install \; && make -j4 --directory=Compiler PREFIX=/workspace/dist-tools/usr/local install'
-    
+
       - name: Create Linux Tools archive ${{ matrix.platform.archname }}
         run: |
           bsdtar -f "linux-tools-${{ matrix.platform.archname }}.tar.gz" -cvzC dist-tools/usr/local/bin .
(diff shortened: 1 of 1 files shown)


ivan-mogilko:
I think this is fine. I used to give download link to my dropbox to everyone who asked for these tools.
Is there anything else to do before merging this pr? It does not seem like it's possible to test this without making a real release, but the changes must be merged for that anyway.

ericoporto:
I made a release here: https://github.com/ericoporto/ags/releases/tag/v3.6.4.1

You can see it has two new github release assets, shown below

- [AGS-Tools-3.6.4.0.zip](https://github.com/ericoporto/ags/releases/download/v3.6.4.1/AGS-Tools-3.6.4.0.zip)
- [AGS-Tools-3.6.4.0-Linux.tar.gz](https://github.com/ericoporto/ags/releases/download/v3.6.4.1/AGS-Tools-3.6.4.0-Linux.tar.gz)

The first is the tools built for windows, and since windows is sort of our default, I didn't add anything to the name like our other builds. Additionally, the other are the tools built for linux, in a tar.gz package meant to preserve the executable bit of the tools.

After https://github.com/adventuregamestudio/ags/issues/2530, I plan to do https://github.com/adventuregamestudio/ags/issues/3068 and then add the AGS version too when reporting the tool version.

PR #3072, opened 2026-08-31
Engine: fix fatal error on a stale character loop after room change

### LINKED ISSUE #3071: Engine crash "no loop X of view Y" when a room change leaves a stale player loop
Hi all,

I'm porting the game Endacopia to the R36S handheld (Linux ARM64, it runs the AGS 3.6.3.13 interpreter as a PortMaster-style port) and I hit a crash that I believe is an engine bug rather than a problem with the game itself. I spent some time tracing it through the engine source and I think I know exactly what's going on, so here's the whole thing. Sorry in advance for the wall of text.

The game shuts down with this when I enter the main/hub area:

```
Error: The character 'cEgo' could not be displayed because there was no loop 5 of view 20.
***** ENGINE HAS SHUTDOWN
```

The odd part is when it happens. Inside rooms everything runs fine, the crash only fires when I walk back to the hub and the game starts the room transition. It never happened on PC either: I have 19+ hours on the Steam version with keyboard and never saw it. And it's not a broken save: I restored saves and played for minutes without any issue, the crash always comes at that transition, and only if I arrive at it holding the analog stick on a diagonal. That last detail turned out to be the whole story.

What I found in the source: `load_new_room()` (Engine/ac/room.cpp, around line 968) reassigns the player character's view on room entry (the room's PlayerView, or the character's defview) and resets the frame, but not the loop:

```cpp
if (forchar->flags & CHF_FIXVIEW) ;
else if (thisroom.Options.PlayerView==0) forchar->view=forchar->defview;
else forchar->view=thisroom.Options.PlayerView-1;
forchar->frame=0;   // make him standing
```

Whatever loop the character had in the previous room just stays there. Later, when the character is drawn in the new room, `update_character_scale()` (Engine/ac/character.cpp, around line 2577) checks `if (chin.loop >= views[chin.view].numLoops)` and calls quitprintf(), which ends the game.

Diagonal movement lives in loops 4-7 (DownRight/UpRight/DownLeft/UpLeft). With an analog stick you arrive at room transitions holding a diagonal constantly, so the stale loop is usually one of those. If the hub's player view has no diagonal loops, that loop no longer exists in the new view, and that's the fatal error: loop 5 (up-right) of view 20. With WASD you almost never trigger a transition while holding a diagonal, which is why this never showed up on PC. Same engine code, different input device, very different odds.

To reproduce it outside this game: room A's player view has the standard 8 loops with diagonals, room B's PlayerView is a 4-loop view (no diagonals), and you walk through the A to B transition holding a diagonal. The first draw pass in room B ends the engine.

I noticed 3.6.3's Changes.txt has an entry about fixing the "loop range limit used when selecting a suitable loop after changing to another normal view or unlocking a temporary view". As far as I can tell that fix limits the range used inside the loop selection function itself; this is a different path, because load_new_room() switches the view without calling the selection function at all, so the stale loop survives until draw time.

The fix I'm running on my handheld (a patch on 3.6.3.13; the crash is gone and I haven't seen any animation glitches):

1. In load_new_room(), right after that view assignment block, call `FindReasonableLoopForCharacter(forchar, true)` so the loop is re-selected against the view that was just assigned. It's the same canonical function the engine uses everywhere else.
2. In update_character_scale(), treat an out-of-range loop the same way the out-of-range frame a few lines below is already treated: debug_script_warn() and reset to kDirLoop_Default instead of quitprintf. A view with no loops at all stays fatal, which matches what FindReasonableLoopForCharacter would report anyway.

Both parts together cover the state being created and the state surviving any other path. If this approach sounds good to the team I can send a PR against master (the patch applies and compiles there, I checked with gcc 14). Happy to provide the full log or anything else that helps.

SaMeiers:
Fixes #3071.

Short version: load_new_room() reassigns the player character's view on room entry and resets the frame but not the loop, so a loop from the previous room can survive into a view that doesn't have it, and update_character_scale() then quits with a fatal "no loop X of view Y" error. Full analysis and repro in #3071.

The change:

- room.cpp: the room-entry view assignment now goes through the new internal Character_SetViewDirect(), which sets the active view without touching defview and stops idling and the current animation, resets the frame and re-selects a valid loop with FindReasonableLoopForCharacter(). Character_ChangeView() is refactored to call it after its own defview assignment, so the logic is not duplicated. When the view is locked (CHF_FIXVIEW) nothing is touched anymore, frame included.
- character.h: declaration of Character_SetViewDirect.
- character.cpp: if an out-of-range loop still reaches update_character_scale(), log a warning and re-select a valid loop with FindReasonableLoopForCharacter() instead of ending the game -- the same thing this function already does for an out-of-range frame a few lines below. A view with no loops at all stays fatal.

Tested on hardware: I built the previous revision for aarch64 and ran it on a R36S handheld with an Endacopia port based on 3.6.3.13. The transition that used to kill the engine every time (entering the hub while facing up-right on the analog stick) works now, both playing fresh and after restoring a save. Happy to rebuild and re-test this revision on the device before merge.

The patch also applies and compiles against current master (gcc 14, Linux).

### DIFF
--- a/Engine/ac/character.cpp
+++ b/Engine/ac/character.cpp
@@ -319,6 +319,21 @@ void Character_ChangeRoomSetLoop(CharacterInfo *chaa, int room, int x, int y, in
 }
 
 
+// Internal: set the character's active view directly, without changing
+// their default view. Does not validate the view number, and expects
+// it to be 0-based.
+void Character_SetViewDirect(CharacterInfo *chap, int vii, bool is_walk_view) {
+    // if the idle animation is playing we should release the view
+    stop_character_idling(chap);
+    chap->view = vii;
+    stop_character_anim(chap);
+    chap->frame = 0;
+    chap->wait = 0;
+    chap->walkwait = 0;
+    charextra[chap->index_id].animwait = 0;
+    FindReasonableLoopForCharacter(chap, is_walk_view);
+}
+
 void Character_ChangeView(CharacterInfo *chap, int vii) {
     vii--;
 
@@ -329,18 +344,9 @@ void Character_ChangeView(CharacterInfo *chap, int vii) {
     if ((chap->flags & CHF_FIXVIEW) && (chap->idleleft >= 0))
         debug_script_warn("Warning: ChangeCharacterView was used while the view was fixed - call ReleaseCharView first");
 
-    // if the idle animation is playing we should release the view
-    stop_character_idling(chap);
-
     debug_script_log("%s: Change view to %d", chap->scrname, vii+1);
     chap->defview = vii;
-    chap->view = vii;
-    stop_character_anim(chap);
-    chap->frame = 0;
-    chap->wait = 0;
-    chap->walkwait = 0;
-    charextra[chap->index_id].animwait = 0;
-    FindReasonableLoopForCharacter(chap, true);
+    Character_SetViewDirect(chap, vii, true);
 }
 
 enum DirectionalLoop
@@ -2576,8 +2582,10 @@ void update_character_scale(int charid)
     }
     if (chin.loop >= views[chin.view].numLoops)
     {
-        quitprintf("!The character '%s' could not be displayed because there was no loop %d of view %d.",
+        // a stale loop can survive a view change, re-select a valid loop
+        debug_script_warn("WARNING: The character '%s' was to be displayed with loop %d of view %d, which does not exist; resetting to a valid loop.",
             chin.scrname, chin.loop, chin.view + 1);
+        FindReasonableLoopForCharacter(&chin);
     }
     // If frame is too high -- fallback to the frame 0;
     // there's always at least 1 dummy frame at index 0
--- a/Engine/ac/character.h
+++ b/Engine/ac/character.h
@@ -39,6 +39,7 @@ void    Character_ChangeRoomAutoPosition(CharacterInfo *chaa, int room, int newP
 void    Character_ChangeRoom(CharacterInfo *chaa, int room, int x, int y);
 void    Character_ChangeRoomSetLoop(CharacterInfo *chaa, int room, int x, int y, int direction);
 void    Character_ChangeView(CharacterInfo *chap, int vii);
+void    Character_SetViewDirect(CharacterInfo *chap, int vii, bool is_walk_view);
 void    Character_FaceDirection(CharacterInfo *char1, int direction, int blockingStyle);
 void    Character_FaceCharacter(CharacterInfo *char1, CharacterInfo *char2, int blockingStyle);
 void    Character_FaceLocation(CharacterInfo *char1, int xx, int yy, int blockingStyle);
--- a/Engine/ac/room.cpp
+++ b/Engine/ac/room.cpp
@@ -967,9 +967,8 @@ void load_new_room(int newnum, CharacterInfo *forchar)
             play.temporarily_turned_off_character = game.playercharacter;
         }
         if (forchar->flags & CHF_FIXVIEW) ;
-        else if (thisroom.Options.PlayerView==0) forchar->view=forchar->defview;
-        else forchar->view=thisroom.Options.PlayerView-1;
-        forchar->frame=0;   // make him standing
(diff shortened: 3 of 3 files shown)


ivan-mogilko:
Hello.

The idea of a fix makes sense, but the code is not entirely right. In the load_new_room the view is changed only unless the character's view is locked (see the use of CHF_FIXVIEW flag); in case it's locked there's is no need to call FindReasonableLoop. It seems that setting frame to 0 in this case is also wrong (this is what the old code was doing).

I propose to instead replace direct assignments of `forchar->view` with calls to Character_ChangeView. This function already deals with everything that should happen when a new view is assigned, including a call to FindReasonableLoop.

I'd also propose to use FindReasonableLoop in `update_character_scale`, in case the loop number exceeds the view's contents (it also does the "numLoops < 1" check internally). This way we will have consistent loop fixing behavior everywhere.

ivan-mogilko:
On a side note, reading FindReasonableLoop function now, I found that it has a mistake of its own in the following part:
```
if (is_walk_view)
{
    if (((chap->flags & CHF_NODIAGONAL)!=0) && (loop_limit > 4))
        loop_limit = 4;
}
```

The idea here was to limit to orthogonal loops of a walking view, in case character does not use these. But I realize now that it should also limit to diagonal loops in the opposite case. In other words the fixup should look like

```
if (is_walk_view)
{
    if ((chap->flags & CHF_NODIAGONAL) != 0)
        loop_limit = std::min(loop_limit, 4);
    else
        loop_limit = std::min(loop_limit, 8);
}
```

SaMeiers:
Thanks, that's a better way to do it. I pushed an updated version:

- load_new_room now calls Character_ChangeView in both branches, and when the view is locked nothing is touched anymore, frame included.
- update_character_scale calls FindReasonableLoopForCharacter when the loop is out of range, keeping a warning so the situation shows up in the log.

One thing I noticed while rewriting: Character_ChangeView also assigns defview, but the old room-entry code only changed view. With a room that has PlayerView set, defview would end up pointing to that room's view, and the character would keep it in the following rooms (and on ReleaseCharView). So I restore defview after the call in that branch, to keep the room's view temporary like before. If you'd rather handle that differently, tell me and I'll change it.

And I can push that FindReasonableLoop fixup as a second commit here if you want.

ivan-mogilko:
> One thing I noticed while rewriting: Character_ChangeView also assigns defview, but the old room-entry code only changed view. With a room that has PlayerView set, defview would end up pointing to that room's view, and the character would keep it in the following rooms (and on ReleaseCharView).

Ah right, somehow i missed that. Then we'll need a separate function (I'd call it Character_SetViewDirect for example) that sets the active view, but not def view, and does rest of adjustments (stops idling, stops current animation, finds loop). This function can then be called even from Character_ChangeView in order to avoid code duplication.

SaMeiers:
Ah, that's much cleaner than the save/restore I had, thanks. Done - I added Character_SetViewDirect (declared in character.h) with the idling/animation/frame/loop adjustments, and Character_ChangeView now calls it after setting defview. load_new_room calls SetViewDirect directly with the 0-based values, so defview is never touched there. Pushed as a single updated commit.

ivan-mogilko:
This looks good. There's just one thing, could you also add a "bool is_walk_view" parameter to the Character_SetViewDirect, and pass it into FindReasonableLoop, instead of hardcoding that to "true" there? I am thinking that maybe this new function can be used elsewhere later, so wish to avoid accidental mistakes.

SaMeiers:
Sure, done - Character_SetViewDirect now takes is_walk_view and forwards it to FindReasonableLoopForCharacter. Character_ChangeView passes true as it did before, and so does load_new_room. Pushed as a single updated commit.

Also still happy to add that FindReasonableLoop cap from earlier as a second commit here if you want it.

ivan-mogilko:
> Also still happy to add that FindReasonableLoop cap from earlier as a second commit here if you want it.

Yes, please, I believe that will be proper (although not related to this bug).

ivan-mogilko:
Anyway, thank you for the bug fix, merged this now added the loop limit fix separately (6c5032ec2abe47d481d9c775691500c8788dbf73)

Also will copy these fixes to release-3.6.3 branch.

PR #3078, opened 2026-09-04
Editor: Fix VS 2022+ winforms designer in x86 

### LINKED ISSUE #3077: Editor: cannot load AGS.Types assembly when editing forms
I suspect this happens after the recent changes to the project configuration (#3051), but I may be mistaken.
When trying to open any gui form for editing in MSVS, the IDE reports that it "cannot load 32-bit AGS.Types", and fails to open the form designer. It still lets to optionally load a ["out-of-process designer"](https://github.com/dotnet/winforms/blob/main/docs/designer/designer-selection.md) and edit forms in there. TBH I don't know much about this.

Image

Need to figure out why it cannot load AGS.Types when editing forms, and fix the problem.

This problem occurs in master and ags4 branches, and does not occur in release-3.6.3 or ags4--patch-4.0.0.30 branches.

ericoporto:
fix #3077 

I noticed the Editor used to be a x86 and not AnyCPU build, just the other things used AnyCPU, but I made it AnyCPU too here.

I wrote all of the code changes so far using sublime text, and am now revising in Visual Studio.


ericoporto:
@ivan-mogilko , I can't for the life of me build this locally in my Visual Studio, but it somehow builds fine in the CI... Do you have any ideas? None of my `.csproj` files builds anymore in my machine.

Edit: amazing, made a change that actually builds the csproj in my pc but now I broke ags native somehow. Edit2: apparently Compiler and Common HAS to be MD and I forgot.

Edit3: I think I actually fixed this! It is working on my computer, so if it passes in the CI it should be good.

image


Edit4: fixed the non-native solution too.

ericoporto:
I noticed it was missing the mark the `Prefer 32-bit` in the csproj despite the checkbox being marked when editing the project properties of the Editor project. I unmarked and marked again, and it updated the Editor csproj file. This is working afaict in my machine, please test if possible, but I think it can be safely merged.

ivan-mogilko:
I cannot judge the changes in project files, but the problem which i've had in #3077 seems to be gone.

ericoporto:
Unfortunately in the end I had to edit the solutions and project files through VS itself and it did additional changes to project files reordering a few things (specially in solutions), so the git diff is less clean than purely a text edit of those files, but on the other end things are actually working. :)

The build results are still good so I would like to merge this and go on. This PR in the end fixes the issue introduced in #3051, and that PR was important to me to make it easier to change the directories in the AGS Editor directory if we want to introduce the tools there somewhere, or say have acwin.exe built for both 32-bit and 64-bit.

PR #3081, opened 2026-09-06
Editor: fix 64-bit Scintilla malfunction

### LINKED ISSUE #3076: Editor 64-bit: Scintilla control malfunction
**Problem**

When Editor is built as a 64-bit program, Scintilla control has a constant malfunction, where its "line collection" seem to contain less data than expected, which leads to assertion failures (and possibly further bugs).

**Steps to reproduce**
1. Build Editor using x64 config.
2. Open any project.
3. Open any script.
4. If script has more than 2 lines, then place cursor onto the 3rd line. If it has less, then add more lines (hit Enter few times).

The assertion triggers with the following callstack:
```
>	AGS.Controls.dll!ScintillaNET.LineCollection.CharPositionFromLine(int index) Line 87	C#
 	AGS.Controls.dll!ScintillaNET.LineCollection.ByteToCharPosition(int pos) Line 56	C#
 	AGS.Controls.dll!ScintillaNET.Scintilla.CurrentPosition.get() Line 3772	C#
 	AGSEditor.exe!AGS.Editor.ScintillaWrapper.CurrentPos.get() Line 508	C#
 	AGSEditor.exe!AGS.Editor.ScriptEditor.SelectFunctionInListForCurrentPosition() Line 328	C#
 	AGSEditor.exe!AGS.Editor.ScriptEditor.scintilla_UpdateUI(object sender, System.EventArgs e) Line 656	C#
 	AGSEditor.exe!AGS.Editor.ScintillaWrapper.OnUpdateUI(object sender, System.EventArgs e) Line 1228	C#
 	AGS.Controls.dll!ScintillaNET.Scintilla.OnUpdateUI(ScintillaNET.UpdateUIEventArgs e) Line 1842	C#
 	AGS.Controls.dll!ScintillaNET.Scintilla.WmReflectNotify(ref System.Windows.Forms.Message m) Line 2744	C#
 	AGS.Controls.dll!ScintillaNET.Scintilla.WndProc(ref System.Windows.Forms.Message m) Line 2815	C#
```

The immediate culprit is member `perLineData` which has only 2 elements, despite script having more lines.

**Affected versions of AGS**

At least the recent ones: 3.6.3, 4.0.

ericoporto:
fix #3076 

I applied the changes to the SCNotification cs so it matched the C++ counterpart in the spirit from the PR in https://github.com/jacobslusser/ScintillaNET/pull/447 (which fixes the upstream https://github.com/jacobslusser/ScintillaNET/issues/488), then I just adjust each build error due to type mismatch until it built in VS and got pretty much the same code changes from the PR - except one missing member at the end of the struct that I added.

It looks like I don't reproduce the reported issue in the Editor Debug x64 build.

### DIFF
--- a/Editor/AGS.Controls/ScintillaNET/NativeMethods.cs
+++ b/Editor/AGS.Controls/ScintillaNET/NativeMethods.cs
@@ -1836,17 +1836,17 @@ namespace ScintillaNET
         public struct SCNotification
         {
             public Sci_NotifyHeader nmhdr;
-            public int position;
+            public IntPtr position;
             public int ch;
             public int modifiers;
             public int modificationType;
             public IntPtr text;
-            public int length;
-            public int linesAdded;
+            public IntPtr length;
+            public IntPtr linesAdded;
             public int message;
             public IntPtr wParam;
             public IntPtr lParam;
-            public int line;
+            public IntPtr line;
             public int foldLevelNow;
             public int foldLevelPrev;
             public int margin;
@@ -1854,9 +1854,10 @@ namespace ScintillaNET
             public int x;
             public int y;
             public int token;
-            public int annotationLinesAdded;
+            public IntPtr annotationLinesAdded;
             public int updated;
             public int listCompletionMethod;
+            public int characterSource;
         }
 
         #endregion Structures
--- a/Editor/AGS.Controls/ScintillaNET/LineCollection.cs
+++ b/Editor/AGS.Controls/ScintillaNET/LineCollection.cs
@@ -341,10 +341,10 @@ namespace ScintillaNET
 
             // Fake an insert notification
             var scn = new NativeMethods.SCNotification();
-            scn.linesAdded = scintilla.DirectMessage(NativeMethods.SCI_GETLINECOUNT).ToInt32() - 1;
-            scn.position = 0;
-            scn.length = scintilla.DirectMessage(NativeMethods.SCI_GETLENGTH).ToInt32();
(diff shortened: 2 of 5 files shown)


ericoporto:
Hey, I retested this and it is indeed working, can I merge it?

Editor/AGS.Controls/ScintillaNET/NativeMethods.cs
```diff
@@ -1836,27 +1836,28 @@ public struct Sci_TextRange
...
             public int margin;
             public int listType;
             public int x;
             public int y;
             public int token;
-            public int annotationLinesAdded;
+            public IntPtr annotationLinesAdded;
             public int updated;
             public int listCompletionMethod;
+            public int characterSource;
```
ivan-mogilko, line 1860: This `characterSource` member is added, but it's not used anywhere?

ericoporto: I copied it from the cpp struct

https://github.com/adventuregamestudio/ags/blob/cb2c2b28a6ee428dfc3aa862706c0565037d139a/Editor/scintilla/include/Scintilla.h#L1265

I think because it is the last entry in the struct, the `Marshal.PtrToStructure` was just making it disappear when going from that cpp to the cs struct, but I think for correctness it should be there.

ivan-mogilko:
I tested a little, and it seems working, in both 32-bit and 64-bit editor builds.

ericoporto:
Awesome. I am merging to be able to move my mind to other things.

I also checked a bit on the forks, and other than changes to get newer versions of scintilla, the other changes I found were for additional features (in the version we use), but didn’t found additional changes that were solely bug fixes.

PR #3084, opened 2026-09-09
Common, Engine, Tools: error refactor to variadic constructor

### LINKED ISSUE #3028: Common: variadic factory method for Error which lets pass a string format with args
The Error class was originally designed following some example, but judging by its use cases, not everything from its features is useful, and its constructors are not fully convenient.

Suggestion:
1. Remove "comment" member, which seem redundant.
2. Support variadic constructor which lets pass a string format with arguments instead of making a separate call to String::FromFormat.

Because variadic function must have format arguments last, the order of arguments in constructor has to be modified, with error text to be the last, followed by optional args. Possible look of constructors:

```
Error(const String &text, ...);
Error(const char *text, ...);
Error(PError prev_error, const String &text, ...);
Error(PError prev_error, const char *text, ...);
Error(int code, const String &text, ...);
Error(int code, const char *text, ...);
Error(PError prev_error, int code, const String &text, ...);
Error(PError prev_error, int code, const char *text, ...);
```

The "previous error" argument's position changes to be the first, so that the new error code and text are together following each other.

ericoporto:
fix #3028

I redid this PR, it should be ready to review and merge. I hid the old text below.

<details><summary> 🚨NOTE: the previous code was AI authored, while I did pick from the old code, and also used as a baseline check, the new code is human authored (which means it can contain more/different mistakes, lol)🚨 </summary>
 

> I am leaving as a draft because I still need to look into it with more care, I know two places there was a format argument but not `%` pattern to get the argument, which got fixed just because I had to look into it. There is also a few places in mfl_utils.cpp where I got type warnings and the fix is lazy.
> 
> The actual error.h refactor was AI authored with minor hand touches. The actual replacement of each calls used a mix of find and replace, hand authoring and AI.  I also did this on my Mac, so better let the CI catch any possible issues in other systems. Leaving it as a draft for now.
> 
> Gemini wrote that new error_test to when I was trying to reduce/simplify the modified header, because I thought it was too big. It can totally be deleted if we think it is useless.
> 
> 🚨NOTE: These code changes were authored by AI, I used Gemini Flash 3.7, Claude Code and Codex GPT Sol 5.6🚨

</details>

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -245,6 +245,7 @@ if(AGS_TESTS)
         test/cmdlineopts_test.cpp
         test/common_stubs.cpp
         test/datahelpers_test.cpp
+        test/error_test.cpp
         test/gfxdef_test.cpp
         test/gui_test.cpp
         test/indexedobjectpool_test.cpp
--- a/Common/ac/spritefile.cpp
+++ b/Common/ac/spritefile.cpp
@@ -214,8 +214,8 @@ HError SpriteFile::OpenFileImpl(std::unique_ptr<Stream> &&sprite_file,
     if (_version < kSprfVersion_Uncompressed || _version > kSprfVersion_Current)
     {
         _stream.reset();
-        return new Error(String::FromFormat("Unsupported spriteset format (requested %d, supported %d - %d).", _version,
-            kSprfVersion_Uncompressed, kSprfVersion_Current));
+        return new Error("Unsupported spriteset format (requested %d, supported %d - %d).", _version,
+            kSprfVersion_Uncompressed, kSprfVersion_Current);
     }
 
     // unknown version
@@ -424,8 +424,8 @@ HError SpriteFile::RebuildSpriteIndex(Stream *in, sprkey_t topmost,
         if (hdr.BPP == 0) continue; // empty slot, this is normal
         if (hdr.BPP < 0 || hdr.Width <= 0 || hdr.Height <= 0)
         {
-            return new Error(String::FromFormat("RebuildSpriteIndex: invalid sprite metrics %d (%dx%d %d-bit), cannot deduce pixel data size.",
-                i, hdr.Width, hdr.Height, hdr.BPP * 8));
+            return new Error("RebuildSpriteIndex: invalid sprite metrics %d (%dx%d %d-bit), cannot deduce pixel data size.",
+                i, hdr.Width, hdr.Height, hdr.BPP * 8);
         }
         int pal_bpp = GetPaletteBPP(hdr.SFormat);
         if (pal_bpp > 0) in->Seek(hdr.PalCount * pal_bpp); // skip palette
@@ -457,8 +457,8 @@ HError SpriteFile::LoadSprite(sprkey_t index, PixelBuffer &sprite)
 {
     sprite = {};
     if (index < 0 || (size_t)index >= _spriteData.size())
-        return new Error(String::FromFormat("LoadSprite: slot index %d out of bounds (%d - %d).",
-            index, 0, _spriteData.size() - 1));
+        return new Error("LoadSprite: slot index %d out of bounds (%d - %d).",
+            index, 0, _spriteData.size() - 1);
 
     if (_spriteData[index].Offset == 0)
         return HError::None(); // sprite is not in file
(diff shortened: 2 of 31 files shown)


ivan-mogilko:
> I know two places there was a format argument but not % pattern to get the argument, which got fixed just because I had to look into it

Are you speaking of original code, or the new code? If this mistake was in original code, then this fix should be backported to release-3.6.3 and possibly release-3.6.2 branches.

ericoporto:
> Are you speaking of original code, or the new code? If this mistake was in original code, then this fix should be backported to release-3.6.3 and possibly release-3.6.2 branches

It is the two places below, but looking again I am wrong, the second parameter is the “comment” parameter that this PR is removing.

https://github.com/adventuregamestudio/ags/blob/a323b3e8b826b57552e7012a1ec698b38e6f6c0b/Engine/plugin/agsplugin.cpp#L1096 

https://github.com/adventuregamestudio/ags/blob/a323b3e8b826b57552e7012a1ec698b38e6f6c0b/Tools/agspak/commands.cpp#L37

btw, I am not so sure now looking at the changes (talking about the entire PR now) if this is a good idea, since these variadic format stuff are easy to make a mistake and before because the string had to be made through String Format it was obvious what was the format and what was extra stuff.

I think in the end I prefer how it worked before.

ivan-mogilko:
> btw, I am not so sure now looking at the changes (talking about the entire PR now) if this is a good idea, since these variadic format stuff are easy to make a mistake and before because the string had to be made through String Format it was obvious what was the format and what was extra stuff.

Is not it the same situation as with any function that has a string formatting, similar to e.g. logging (Debug::Printf)?

Majority of `new Error` constructors in the code require string formatting, meaning one has to be typing `new Error(String::FromFormat(...` every time.

With the removal of useless "comment" argument, the constructor's arguments is only formatted string, and optionally the previous error as a first argument.

ericoporto:
I haven't tested it yet, but I asked ChatGPT if there was a way to add a compile time check, and it gave me this code for a macro to add to the header where a function that takes a variadic meant to use for formatting could be used, it wouldn't get checked in MSVC, but it would get a compile time check in GCC/Clang when the code ran in the CI.

```c++
// Adds printf-style format checking to variadic functions on GCC and Clang.
//
// The two numbers are 1-based parameter positions. For member functions,
// the implicit "this" parameter is counted as argument 1.
//
// The arguments are:
//   fmt  - parameter number containing the format string
//   args - parameter number of the first variadic argument
//
// For a regular function:
//   void Log(const char *fmt, ...) AGS_PRINTF_FORMAT(1, 2);
//     1 = fmt
//     2 = first argument after ...
//
// For a member function:
//   void Log(const char *fmt, ...) AGS_PRINTF_FORMAT(2, 3);
//     1 = this
//     2 = fmt
//     3 = first argument after ...
//
// This enables compile-time checks such as:
//   Log("Value: %d", 123);       // OK
//   Log("Value: %d", "hello");   // warning/error
//   Log("Value: %d %d", 123);    // warning/error
//
// The macro is empty on compilers that do not support this attribute.
#if defined(__GNUC__) || defined(__clang__)
    #define AGS_PRINTF_FORMAT(fmt, args) __attribute__((format(printf, fmt, args)))
#else
    #define AGS_PRINTF_FORMAT(fmt, args)
#endif
```

I think I won't do this in this PR, but this could be something interesting to add later (I would need to look into the code for all the variadic function calls that use format like strings, plus check how this actually works in all the compiler variants we use (old gcc, newer gcc, mingw, clang, apple clang, emscripten, clang throug gradle for android...)

ivan-mogilko:
I noticed there was a comment about adding `__attribute__(format)` for printf-like functions in the engine (for compilers that support that), but this comment is now removed. In any case, i think that was a good idea.

EDIT: by the way, MSVC also supports a similar thing, although it's more complicated. One has to enable SAL in the build configuration:
https://learn.microsoft.com/en-us/cpp/code-quality/understanding-sal?view=msvc-170
Formatted string annotation:
https://learn.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values?view=msvc-170#format-string-parameters

ericoporto:
Thanks, I wanted to read more on that before and testing but didn’t found the time to sit in the computer to do so yet.

ericoporto:
@ivan-mogilko I think I am ok now with these code changes. On the code format check idea, I found some promising code in SDL:

https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_stdinc.h

Look that they include this SAL approach you mention but also some easier way to apply to gcc and the like, below is an example of how they use

https://github.com/libsdl-org/SDL/blob/26b37f5d1c3518125ff8b269b5bdfcf452870500/include/SDL3/SDL_error.h#L89

I imagine a simplified version of these could be made for our usecase.

ericoporto:
I think I want to have #3087 done before, later this here is easier to rebase on top of that, and readjust the code changes here that are mostly very mechanical.

ericoporto:
I moved the old code to https://github.com/ericoporto/ags/tree/refactor-error-variadic-bkp and redid this PR by hand.

ivan-mogilko:
Did not notice when this happened, but apparently there are new conflicts (maybe after recent crmpak changes?).

ericoporto:
It was a merge conflict in data_ext, where the inner error flipped to be the first parameter, and you had also done the same a little bit of changes I had done to error.h [here](https://github.com/adventuregamestudio/ags/pull/3085/changes#diff-c7686f4c46357a7763ffc8a32da71ba1f308dec25bbd07c3132added441a9ee5). I think I fixed both. But I will need to adjust again after #3088 or you may have to adjust there, whichever PR gets in last. :)

ivan-mogilko:
Hmm, looking at all of these constructors with va_list code makes me think that it can use variadic forwarding to String::FromFormat, similar to this code:
https://github.com/adventuregamestudio/ags/blob/35348239cb9c433cca7785c453f2f24220686664/Engine/debug/debug_log.h#L35-L39

ericoporto:
ok, I did it in https://github.com/adventuregamestudio/ags/pull/3084/changes/182e8f649c84735955ac02d3feb15adb59cf55af, but I honestly think it makes harder to read and the code is not much smaller.

Edit: I dropped because it breaks the compile time checks, so it makes it worse in a second front too. The above also skips the compile time checks if I recall correctly. So I won't do that.

ivan-mogilko:
Hmm, that's unfortunate. Is it because you removed `AGS_FORMAT_STRING_OBJECT` from the constructor declaration?
Okay, I'd leave this as it is, but wish to investigate if it's at all possible to achieve with C++11 variadic style.

I suppose the debug_script_xxx functions will have to be converted to use va_list too, for the sake of having printf checks.

ericoporto:
I am not very good with templates and meta programming yet, I can do some stuff and follow some basic examples but this stuff goes over my head. I will try to learn more.

PR #3085, opened 2026-09-10
crmpak: additional options

ivan-mogilko:
As PNGs do not support 16-bit color depth, I am adding an option `--ibpp N` that lets to specify the destination bits-per-pixel for imported backgrounds. This will allow to import 24 or 32-bit PNG images into 16-bit rooms.

This option is only applied when importing **primary background**. The secondary backgrounds always use existing primary bg's color depth.

If the primary background's color depth changes, as a result of an import, then all the secondary ones will be automatically corrected by the tool. This must be done, because AGS expects room bgs to match same color depth.

~There's still something wrong in this code, because when I test the modified rooms with a 16-bit game, the bg sizes become smaller, and room scripts seem to not work properly, as if the room overwrite have broken something. So leaving this a draft for now until i fix the mistakes.~

While trying to diagnose the problem, I also made `-l. --list` command to print table of blocks, similar to what old version of crmpak did before a rewrite. This lets to compare the sizes of data between original and modified room files.


ericoporto:
> the bg sizes become smaller

erh looking at the code I can’t think of how that could happen. Did you test with different images? The only thing that I think could be wrong is color conversion shifts/mask (also there is a typo in the comment since it’s 16-bit RGB and not ARGB).

ivan-mogilko:
There's also a chance that this has something to do with old game behavior, as I've been testing this with a very old game.
Need to retest with a game made in current version first.

ivan-mogilko:
I found and fixed mistakes in image conversion. The tool works when testing with the "new" games. The previous problem I've seen was likely related to room being saved in a new format, while the game expects old format, so there was some kind of a mismatch.

I dont see other way to solve such issue, rather than add support for saving rooms in old format. But that's a separate matter. I suppose if we do this in crmpak, then we also should do this with every other tool.

ericoporto:
I think it is fine, the PR looks good too, I see you also added two extra color conversions.

In bitmapdata, in the color conversion stuff, that bit rotation pattern with the extra `>> 3` and `>> 2` looks like something that is easy to forget, and hard to understand  with the other runtime defined `_rgb_r_shift_24` and the like (magic numbers would be actually easy to read there), plus hard to test (I mean other than visually)... I don't think this PR is the place to change any of that, but just mentioning so I can remember if new color conversions are authored.

ivan-mogilko:
When i started writing this image conversion function, it was purposed for something that Allegro does not do (converting from 1-bit and 4-bit bitmaps). Then I had to add more conversions, I was conflicted, whether to use Allegro or not, because using functions like getr32, makecolor and similar would drag multiple dependencies along. So at first I tried to avoid that. After more time made a small refactor to uncouple dependencies within Allegro and let tools use color.c without bringing whole bitmap drawing code. So in theory this code now can probably use color get/set functions. Perhaps it should, since it's using shift constants, and tools already include color.c anyway. But eventually, if we remove Allegro completely, these operations and all of their constants will have to be moved to our own code.

Issue #1488, opened 2022-01-08
Configurable error behavior for script API


ivan-mogilko:
### Problem

When AGS engine detects an error in a script API function, caused by the invalid arguments, or simply unable to perform at this time, it reports an error.  The result of such report may vary. Historically, some errors were merely reported as warnings and written in the log file ("warnings.log"), while others would cause the engine to stop, displaying an alert, and quit. The latter method is also used by the Editor to pinpoint actual script line where this happened.

These different ways of report are done using different functions.
For instance, `quit` and `quitprintf` functions are meant to report error and stop the game. `debug_script_warn` and similar are used to merely print the message to log.

Over time there have been multiple requests to replace this or that game stopping error with merely a warning, as closing the game seemed to be too radical. In most cases the script api error can be dealt by the engine simply by supressing the command.

However, the situations vary, and sometimes a developer may want to be more alert to any mistakes in their game.

**More on error reporting methods**

IMHO it's difficult to justify shutting down a game on a script api error even as an option. Such approach may have been easier from the code writing perspective (as you dont have to invent how to handle an exception). But i dont think it's every convenient for the user, would that be a game developer, a tester or a player.
When testing the game it may be important to try the failing action several times and observe the situation; that will become a chore if the game stops complately after every error.
For players that means a progress loss in case of any slightest script mistake. This may damage the player's experience and impression of the game.

Currently the game stopping errors are used by the Editor to halt execution and let user observe the location where the error occured.
But *it does not have to be game shutdown* for this to work. After certain changes in the engine, it may be made possible to trigger such halts without quitting. Or even pass non-blocking warnings to the Editor's output window.

But then, the reaction to particular error is be better configured inside the Editor itself; engine should not care about such things. All it has to do is perhaps to distinguish error "levels". It's possible to leave this for a separate task though, as planning such levels and implementing them to all the existing error messages will take more time.

### Proposed solution

The goal is to remove "quitting the game on script error" as an option completely, and replace it with a log message + error event passed into the debugger.

Add a setting, or a number of settings if necessary, that control the error reporting for script api.
At the very least it should have a selection among two modes:
- stop the game on error;
- print error into the log and continue;

Write a function all the script API errors suppose to go through, that acts according to these settings: that is
- prints into the log;
- passes the error event to active debugger, if one is active;
- if no debugger is active, then (optionally) displays as an alert window on systems that support that (i.e. using `platform->DisplayAlert()`).

[...]

PR #3086, opened 2026-09-11
Uniform object and character assertions in script functions

ivan-mogilko:
Enforce that Object and Character script functions use same Assertxxx function that tests if the object is valid and prints to the log instead of quitting. If assertion fails, the function returns immediately with default result.

Added similar assertions to all the object functions that missed them.
This fixes the multiple potential crashes which could happen if game script tries to get/set object's properties using `object[N]` object access, in a room that does not have a object at particular index.

### DIFF
--- a/Changes.txt
+++ b/Changes.txt
@@ -321,6 +321,11 @@ Engine:
    translate the GUI text property values).
  - Engine does not quit with error when DrawImage is called with a non-existant sprite, only logs
    a warning.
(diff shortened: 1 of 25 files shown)


ivan-mogilko:
As an addition, made all debug_script_xxx functions to end up called debug_script_event, which may decide to quit the game with a alert instead of simply logging a error. Currently the choice is based on whether the game is compiled in DEBUG mode.
Later this may be expanded with some sort of config (see #1488).

PR #3087, opened 2026-09-13
Adds macros for compile time check of variadic format-like calls

ericoporto:
Hey, I am trying to do what I mentioned in a comment

https://github.com/adventuregamestudio/ags/pull/3084#issuecomment-5648702463

This PR adds three macros that when used in the variadic function calls in ags, allow for type checking at compile time. Unfortunately (expectedly?) adding made me find a gazillion number of issues with the debug/log of stuff (mostly 32-bit / 64-bit builds mismatch). I haven't tested the SAL stuff on MSVC btw, I am mostly using the CI and macOS builds.

Edit: noticed I have no idea what these casts below mean, I fixed format errors in this file that I found in a different way.

https://github.com/adventuregamestudio/ags/blob/b443282af7596bea7fed4b729f57c96acf09c640/Common/font/wfnfont.cpp#L69-L70


ericoporto:
Looks like it finally builds everything now. I am curious if I should leave all individual fixes in many commits or if I should squash everything in a single "All: string format fixes"

Engine/debug/debug_log.h
```diff
@@ -27,6 +27,11 @@ void apply_debug_config(const AGS::Common::ConfigTree &cfg, bool finalize);
 void shutdown_debug();
 
 void debug_script_event(AGS::Common::MessageType mt, const AGS::Common::String &msg);
+// overload for calls that have no args
+inline void debug_script_event(AGS::Common::MessageType mt, const char *msg)
+{
+    debug_script_event(mt, AGS::Common::String(msg));
```
ivan-mogilko: This should use String::Wrapper to avoid string copying.

ivan-mogilko:
I have few notes about the macro declarations.

1. The "Usage" comment is misleading where it compares situations with GCC/Clang and MSVS, because at first it looks like it is suggesting to use different variants in code depending on a compiler. I think these differences do not belong to "usage" explanation but rather to macro implementation. If someone wants to know how to use the macros they do not need to worry about how they are figured out by different compiler. So the "Usage" part should state simply: "in case of a static function use this, in case of a non-static member function use that".

2. Macro use for non-static members requires extra index increment. That is easy to forget and may become a source of mistakes. What about having a separate macro for this case? Like, something with THIS, or OBJECT in the name, or else (e.g. `AGS_FORMAT_STRING_HAS_THIS`).

3. A minor suggestion, I would perhaps reserve a compiler option for disabling these, just in case someone has to make a build without these checks for some reason.


About the formatting fixes, I trust the compiler checks are correct to point about these mistakes. You may squash these commits if you like or leave them, as you prefer. I think critical fixes must be backported to earlier branches, like the one with the missing arguments.

ericoporto:
> A minor suggestion, I would perhaps reserve a compiler option for disabling these, just in case someone has to make a build without these checks for some reason

ah, ok, SDL does it too and I mostly copied the rest from them

https://github.com/libsdl-org/SDL/blob/dc05826028d6850ccc48661010ac444942cec0af/include/SDL3/SDL_stdinc.h#L1199

I guess I could name it as AGS_DISABLE_FORMAT_MACRO or AGS_DISABLE_FORMAT_CHECK

Engine/debug/debug_log.h
ericoporto: thanks! Fixed it.

ericoporto:
hey, I tried to move all the stuff that had to do with 64-bit/32-bit format like differences to a single last big commit, and left the other issues in separate commits before, in case you think anything there is worth porting over to some other branches.

I tried also to address the previous comment, and I added an `_OBJECT` variant of the macro for using on things that are member of something. Please review, again - I know this is a lot of stuff, but hopefully the new separate commits makes it a bit easier.

PR #3088, opened 2026-09-16
Change Rooms to use Deflate compression on backgrounds

ivan-mogilko:
This PR makes Rooms use Deflate compression when saving their backgrounds to the compiled room file. Rooms used LZW historically (for most of AGS history, it seems), but Deflate can be roughly ~30% more efficient. Of course this requires a room version increment. Older versions still load with their respective decompression type.

This is something I was planning to do for a while, but kept forgetting. While it's not a good thing to add such changes when the version is so close to become a stable release, but I thought this might come as a useful "bonus", since 3.6.3 has a raised room backgrounds limit, and especially seeing as people make more high-res games in the recent years.

90% of the changes in this PR is moving code around. After some thought, I decided to make rooms use the "sprite format" for their backgrounds, for consistency, and also because it implements further space optimization. But this required to move parts of the SpriteFile and SpriteFileWriter classes into standalone functions.

Unlike the spritefile package, rooms are supposed to always use the best compression and max available "storage optimization" settings.


ericoporto:
This looks sensible and a good change to improve the binary room sizes. This also looks, safe AFAICT.

Ah, this may conflict with #3084, simply because I touched too many files there. It may be easier to merge this before and the other later - or the other way around if you prefer to author the adjustments here.

I know this PR doesn't change it, but a note that I played with a version of this in the past for the Masks (walk mask, region, ...) and I was surprised how well the existing RLE approach performed in them, I think for their particular case of very little color and big areas of the same colors, it beat any other thing I tried.

ivan-mogilko:
By the way, using sprite format leaves an option to switch compression without modifying the save format, because compression type is stored as meta data. Even though there are no way to control this now.

Issue #1232, opened 2021-03-06
Allow custom support message for run-time scripting errors


morganwillcock:
Telling someone to contact the AGS forums when a commercial game won't run probably isn't the best thing to do by default:

https://www.adventuregamestudio.co.uk/forums/index.php?topic=58915

Ideally the game author should be able customise this message with their own support information, if required.

ivan-mogilko:
Would it be a good idea for a quick fix to amend the message with "Contact game developers, OR " and only then go about ags tech forums?

BTW, in the linked case, there's a engine mistake, as this is not an "internal error", so error type for mismatching saves might be fixed too.

morganwillcock:
With these sorts of errors I would imagine that for the most part only the game developer would be able to help, so maybe the forum shouldn't be mentioned at all?

Maybe this is something that should be automatically flipped based on the build type, so the debug builds gives the message to contact the forum, and the release build gives the message to contact the developer, but also have an option to override with custom messages.

rofl0r:
reminds me that i proposed on another issue that maybe it would be a good idea to have actually 2 buildtypes of engine: standalone version that has all kinds of debug features (including cheat functionality) that's associated with the github org or forum, and bundled version which is used by game developers (probably acwin.exe stub prepended to .exe). the engine could probably even differentiate the usage mode depending on whether the game filename was passed as a commandline argument, or argc == 1 and the data attached, in case having 2 different builds is undesired.

ivan-mogilko:
Just for reference, for time being I made a small fix to release-3.5.0 branch, asking to contact author as a first option (d0914896cec805be17bbeda56b1bdca93afeaef1).

Also, I realized there might be missing error "type" among the "types" identified and reported by the engine. There's "script error" - usually when a script command was misused (there is also a feat request for making a setting at which these errors are treated as warnings). And there's "internal error", but it's used both when something completely unexpected had happened, and when something wrong happened but engine could not recover. The example for latter is restoring a save: the only reason engine quits is because currently it is not programmed to backup/restore a previous state in such case. But this is not "internal" error in usual sense.

ivan-mogilko:
So, for the minimal implementation we would need to design a method of setting this message up.

The potential problem that I predict is that the message is displayed in a simple message box as a single label, and there's no way to customize by adding a link to author's website, for example.
And some oses do not display this as a modal message box, but rather in terminal instead, - Linux, and OSX too, I think.

Should the message differentiate between the OSes?

[...]

PR #3092, opened 2026-09-18
Engine: use game name for message boxes

ericoporto:
In the spirit of https://github.com/adventuregamestudio/ags/issues/1232 and https://github.com/adventuregamestudio/ags/issues/2729, but not directly, this change makes the message box use the same title used in main game title.

https://github.com/adventuregamestudio/ags/blob/35348239cb9c433cca7785c453f2f24220686664/Engine/main/engine.cpp#L141-L146

(I had the idea when peeking around https://github.com/pointandorclick/ags/commit/7aa15de02906da707f59b68cd190f71adacfbbc0#diff-3f4d9fe4e91fcb150cc002f44b79c89db1df3220390dd181a611d6a51f9cb1c5)


ericoporto:
Actually, after looking at an error message, I noticed we just write "AGS" in the internal error text an never the full "Adventure Game Studio", so for a random game from the forums, it will lose the title "Adventure Game Studio" and then the error messages become more mystical.

ivan-mogilko:
Thank you for reminding about this. I am not sure why #1232 and #2729 were not done; probably I kept forgetting.
I believe it's should be possible, and not hard, to add this to 3.6.3.

There's a "v363_gameinfo" extension for developer info already:
https://github.com/adventuregamestudio/ags/blob/4129b370a361c5cdf2d7f2737d31daad93022e88/Common/game/main_game_file.cpp#L1005-L1009

it's just a string dictionary, where engine can look up for info using certain keys. The list of the keys is not specified in one place, unfortunately, but may be found by searching for `game.GameInfo` in the engine code. (Perhaps the list of keys could be added as a comment to GameInfo variable.)

What remains is to decide a minimal set of data necessary for these error messages. I dont think it's worth having separate property for each message type. Perhaps we could do with just 1 message, which is used always when assigned in a similar vein to the commit that you linked above, and standard engine messages are used otherwise.

ericoporto:
I believe a part of these messages can be left hard coded with a second part that actually changes - with the defaults pointing to the AGS forum and stuff.

Looking at the fork can give some insights on the messages: https://github.com/pointandorclick/ags/commits/sierra-quest/?author=craigharman

The messages having some differences are relevant to be able to traceback from where the error message was called from.

PR #3093, opened 2026-09-19
Customize game's quit with error text

ivan-mogilko:
Resolves #1232 and #2729

Adds a new property to General Settings called "DeveloperErrorText" which lets user to specify a text which will appear in AGS messages when it quits with error.

This string is *added to translations*, and can be translated if wanted so.

Tidied up engine's fatal error's text generation, making it consistent through all error types.
The message now has a following structure
* Error explanation: uses "developer error text" if defined, or a standard string, explaining the nature of error (script error, aborted by abort key, internal error)
* Engine version.
* Game version (if provided, available only in games made by 3.6.3 and higher)
* The actual error value, with script's callstack if possible.

Examples:
newerror1
newerror2

Replaced error popup's title with game's name (following #3092).


Engine/main/quit.cpp
```diff
@@ -109,65 +109,80 @@ void quit_shutdown_audio()
...
+                "Please contact the game author for support.";
         }
         else
         {
             qreason = kQuit_GameError;
-            alertis.Format("An error has occurred. Please contact the game author for support, as this "
-                "is likely to be an error in game logic or script and not a bug in AGS engine.\n"
-                "(Engine version %s)\n\n", EngineVersion.LongString.GetCStr());
+            error_is = !error_text.IsEmpty() ? error_text :
+                "An game error has occurred. Please contact the game author for support, as this is "
```
ericoporto: I think it is “A game” and not “An game”

ericoporto: Same for “a mistake” and not “an mistake”

ericoporto:
I haven’t run it yet, but a note, even if a developer replaces the error message, it would be nice if somehow it was still possible to know if it’s an internal engine error, a script error (access of a null pointer) or error because abortgame was called.

ivan-mogilko:
Added a short extra string above engine version:
- `Game script requested abort.` (in case AbortGame called)
- `Game script or data error.`
- `Internal engine error.`

Engine/platform/base/agsplatformdriver.cpp
```diff
@@ -120,10 +120,19 @@ void AGSPlatformDriver::WriteStdErr(const char *fmt, ...)
...
+    _windowTitle = title;
+    sys_window_set_title(_windowTitle.GetCStr());
+}
+
 void AGSPlatformDriver::DisplayMessageBox(const char *text)
 {
     if (_guiMode)
-        SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_WARNING, "Adventure Game Studio", text, sys_get_window());
+    {
+        String msgbox_title = !_windowTitle.IsEmpty() ? _windowTitle : "Adventure Game Studio";
```
ericoporto: I think this has to also be done in the Windows version in agswin32::DisplayMessageBox

Engine/main/quit.cpp
```diff
@@ -109,65 +109,92 @@ void quit_shutdown_audio()
...
-            alertis.Format("An error has occurred. Please contact the game author for support, as this "
-                "is likely to be an error in game logic or script and not a bug in AGS engine.\n"
-                "(Engine version %s)\n\n", EngineVersion.LongString.GetCStr());
+            error_is = !error_text.IsEmpty() ? error_text :
+                "A game error has occurred. Please contact the game author for support, as this is "
+                "likely to be a mistake in game logic or script and not a error in AGS engine.";
+            error_kind = "Game script or data error.";
         }
 
-        alertis.Append(cc_get_err_callstack());
```
ericoporto, line 144: It looks like the call stack was appended unconditionally before and now it is inside an if not normal game exit and not user request.

Editor/AGS.Types/Settings.cs
```diff
@@ -125,6 +125,7 @@ public Settings()
 		private string _version = DEFAULT_VERSION;
 		private string _developerName = string.Empty;
 		private string _developerURL = string.Empty;
+        private string _developerErrorText = string.Empty;
```
ericoporto, line 128: The other stuff is being set to string.Empty in FromXml before reading, I think this should be too?

Engine/main/quit.cpp
ivan-mogilko: Yes, this is intentional.
There's no error message whatsoever when it's a normal game exit, and simple "was aborted by player" message when aborted by a key press.

ericoporto: I am unsure when using abort to exit not having the callstack (which is what I understood was the user requested but perhaps I misunderstood it)

ivan-mogilko: kQuit_UserAbort - is abort key (Alt+X).
kQuit_ScriptAbort - is AbortGame script function call.

Engine/platform/base/agsplatformdriver.cpp
ivan-mogilko: I remade this to have this condition in SetWindowTitle instead, and also initialize _windowTitle in constructor.

Engine/main/quit.cpp
ericoporto: Uhm, I made a confusion, but the callstack when exiting with Alt+X may still be useful, I remember we used that when something hanged - hanging hasn’t been a problem for me in a long time, so I hadn’t had the need recently.

ivan-mogilko: Right, I remembered this too now, so restoring the previous behavior for this case.

ericoporto:
Looks good overall, I don’t know if this deserves its own topic in the development board in the forums or if someone would comment on the editor thread when this makes release. I am curious of what the people there and the commercial developers think about.

Engine/main/quit.cpp
```diff
@@ -109,65 +109,97 @@ void quit_shutdown_audio()
...
         }
         else
         {
             qreason = kQuit_GameError;
-            alertis.Format("An error has occurred. Please contact the game author for support, as this "
-                "is likely to be an error in game logic or script and not a bug in AGS engine.\n"
-                "(Engine version %s)\n\n", EngineVersion.LongString.GetCStr());
+            error_is = !error_text.IsEmpty() ? error_text :
+                "A game error has occurred. Please contact the game author for support, as this is "
+                "likely to be a mistake in game logic or script and not a error in AGS engine.";
```
ericoporto: Typo: and not *an* error in AGS

ivan-mogilko:
The error message consists of user's message and then error details which are always important, therefore I doubt there may be something wrong or unnecessary in it. It could be that users will want something extra though. Like, #2729 suggests that the support link is clickable. I dont know how to achieve this with a standard message box, and if this is at all possible to do without coding a custom dialog on a per-platform basis (we'll need this on every dekstop system, and maybe Android/iOS too). But this may potentially be done in the future. If done, then it could just keep using same "error text" property, only support extra tags to define a link (e.g. `<url></url>`). So this feature is extendable to a degree.

ericoporto:
So let’s merge it, I think it is fine. I haven’t tested old games with this runtime before these dev properties existed, I to check what happens. Like, I don’t remember, did those error messages always show Game version?

ivan-mogilko:
They never showed game version, and game version itself was just added to game data in 3.6.3.
If this property is not available (empty string) then it's not supposed to be shown.

PR #3095, opened 2026-09-22
[SKETCH] AGS4 Editor: external Dialog files

ivan-mogilko:
For the #1008

This may be either a remake or a complementary change to #2978, depending on how it goes.
Here I started from a different angle, and first made Dialogs themselves be stored in separate files, as was suggested in the #2978's discussion. The scripts are still saved within the dialog data.

The "Dialogs" node in Game.agf now has a "Version" attribute, which, if present, signifies that the dialogs should be read in the "new" way. If it's abscent, or present but the value is less than minimal expected, then it is read in "old" way.
In the "old" way the Dialogs are read fully from the Game.agf.
In the "new" way the objects called DialogRef are read from Game.agf, and Dialogs are only created and assigned ID and ScriptNames. Full dialog data is read on a GamePostLoad event by DialogsComponent, where it searches for the respective files in "Dialogs" subfolder, and reads their contents into the Dialog objects. DialogsComponent also does the saving of these files now, on a newly added GamePostSave event. Saving is always done in the "new" way.

DialogRef is a helper class, which has ID, ScriptName and Dialog pointer. Their primary purpose is serialization, specifically - ensuring that only minimal data is saved into Game.agf. This is necessary, as saving is done through IToXml interface, which imposes some constraints. List of DialogRefs and Dialogs must be kept in sync at all times.

About the data file names:
1. If dialog has a script name, then it's "DialogName.xml".
2. If dialog does not have a script name, then it's "@dialogN.xml", where N is Dialog's numeric ID. The "@" prefix is necessary to avoid naming conflicts, because in theory *another* dialog may be called "dialogN" (by user's mistake). As "@" symbol cannot be used in script names, it guarantees that such name will not exist.

NEXT MAJOR STEPS:
1. Save Scripts separately from the Dialog data.
3. Add GameUpgrade step.
4. Add file listeners and figure out how to pick up external changes (at least for the scripts).


ivan-mogilko:
Done saving scripts separately. I used @edmundito's suggestion to use "dasc" extension here, although not claiming this final.

Among other things, there's something i have not decided is whether we should create and delete dialog files as soon as they are created, deleted or renamed in the Editor. At the moment it does not, and only (re)writes them when the whole project is saved. This keeps these files synced with the serialized project's state; which may be good if user exit without saving, or system shuts down abruptly. But on a downside, it's impossible to see or edit dialog files externally until the game was saved at least once after adding them. Also, editing a file, which belonged to the *renamed* dialog will likely result in loss of edit, as soon as the game is saved. Not sure if that's important, but I'm still in doubts about this case.

ivan-mogilko:
With the approach in this PR, it seems like there's not much to do during Game Upgrade, except for display some note to the user, and backup "Dialogs" folder if one existed previously.

ericoporto:
Hey, this is probably not related to this PR, but I did a simple thing, I cloned my [dont-give-up-the-cat](https://github.com/ericoporto/dont-give-up-the-cat), and tried to load it with the AGS4 built from CI in this PR and after loading it, when it compiles, it errors with: `_BuiltInScriptHeader.ash(449): Error (line 449): Invalid use of forward-declared struct`

image

Now, on the dialog itself, I wonder if it would be possible to indicate the file names in the properties. This is just a minor request, and it it is difficulty we can leave it. Note this is already inconsistent with the regular scripts, where the script has it shown in the properties but the header has its properties blank.

ivan-mogilko:
> Hey, this is probably not related to this PR, but I did a simple thing, I cloned my [dont-give-up-the-cat](https://github.com/ericoporto/dont-give-up-the-cat), and tried to load it with the AGS4 built from CI in this PR and after loading it, when it compiles, it errors with: `_BuiltInScriptHeader.ash(449): Error (line 449): Invalid use of forward-declared struct`

Please open separate bug issues for unrelated problems.

The above is a problem with old compiler that cannot parse something. New compiler does not have this.

ivan-mogilko:
> Now, on the dialog itself, I wonder if it would be possible to indicate the file names in the properties. This is just a minor request, and it it is difficulty we can leave it.

This is just a matter of adding display attributes.

ericoporto:
Hey, I know it is still draft, but I tried again and I have an issue that if I have a dialog open and I hit build exe, it shows the dialog below saying the dialog has been modified externally... Except I didn't modify it externally, it is something in the build exe that is triggering this. It doesn't seem to happen if I just click the play debug button.

image

ivan-mogilko:
The above happens because I forgot to disable filewatcher while the dialogs are being saved.
But I want to remake other things first, currently the dialog scripts remain in memory when the pane closes, and I thought it should instead behave similar to regular scripts, where they are marked "modified" and ask user if they want to save changes before closing a pane with modified dialog script.

ivan-mogilko:
Still not done, I still need to rewrite dialogs saving on whole project save, because earlier it was intended as the only point of saving them, and now dialog scripts are also saved when closing a dialog pane, just like regular scripts. The file watch is not disabled during project save, which causes them to trigger. That's because file watchers are owned by the editor panes, while saving is done by accessing dialog scripts directly (which is not right).

Also file renaming should be handled when saving an individual pane.

